PR #329, opened 2017-01-01
add numerals classification

jenniew:
Add numerals classification example.


jenniew:
Add finetune example

jason-dai:
Can you create a new pull request for the fine-tune example (instead of merging it with the previous example)?

PR #342, opened 2017-01-05
Seq2seq

JerryYanWan:
Add tests for LabeledSentence, support array label and single label
modify rnn test to select the output with largest probability each recurrent time

### DIFF
--- a/dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/Dictionary.scala
+++ b/dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/Dictionary.scala
@@ -41,7 +41,7 @@ class Dictionary()
 
    def vocabulary(): Array[String] = _vocabulary.toArray
 
-   def discardVocab(): Array[String] = _disvardVocab.toArray
+   def discardVocab(): Array[String] = _discardVocab.toArray
 
    def getIndex(word: String): Int = {
      _word2index.getOrElse(word, _vocabSize)
@@ -57,7 +57,7 @@ class Dictionary()
 
    def getWord(index: Int): String = {
      _index2word.getOrElse(index,
-       _disvardVocab(Random.nextInt(_discardSize)))
+       _discardVocab(Random.nextInt(_discardSize)))
    }
 
    def print(): Unit = {
@@ -66,7 +66,7 @@ class Dictionary()
    }
 
    def printDiscard(): Unit = {
-     _disvardVocab.foreach(x =>
+     _discardVocab.foreach(x =>
        logger.info(x))
    }
 
@@ -85,8 +85,8 @@ class Dictionary()
      _vocabSize = _vocabulary.length
      _word2index = _vocabulary.zipWithIndex.toMap
      _index2word = _word2index.map(x => (x._2, x._1))
-     _disvardVocab = freqDict.take(freqDict.length - length).map(_._1)
-     _discardSize = _disvardVocab.length
+     _discardVocab = freqDict.take(freqDict.length - length).map(_._1)
+     _discardSize = _discardVocab.length
    }
 
    def this(directory: String) = {
@@ -109,9 +109,9 @@ class Dictionary()
      _index2word = _word2index.map(x => (x._2, x._1))
      _vocabulary = _word2index.keys.toSeq
      _vocabSize = _word2index.size
-     _disvardVocab = Source.fromFile(discardFile.getAbsolutePath)
+     _discardVocab = Source.fromFile(discardFile.getAbsolutePath)
        .getLines().toSeq
-     _discardSize = _disvardVocab.length
+     _discardSize = _discardVocab.length
    }
 
    val logger = Logger.getLogger(getClass)
@@ -120,12 +120,12 @@ class Dictionary()
    private var _word2index: Map[String, Int] = null
    private var _index2word: Map[Int, String] = null
    private var _vocabulary: Seq[String] = null
-   private var _disvardVocab: Seq[String] = null
+   private var _discardVocab: Seq[String] = null
 }
 
 object Dictionary {
   def apply(sentences: Array[Array[String]],
-            vocabSize: Int)
+            vocabSize: Int = 10000)
   : Dictionary = new Dictionary(sentences, vocabSize)
   def apply(directory: String)
   : Dictionary = new Dictionary(directory)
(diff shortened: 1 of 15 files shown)


dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/LabeledSentenceToSample.scala
```diff
@@ -60,24 +58,40 @@ class LabeledSentenceToSample(vocabLength: Int,
       // initialize featureBuffer to 0.0
 
       util.Arrays.fill(featureBuffer, 0, featureBuffer.length, 0.0f)
+      util.Arrays.fill(labelBuffer, 0, labelBuffer.length, 0.0f)
 
       // One-Hot format for feature
+      // Expected transformed format should be:
```
JerryYanWan: fixed

JerryYanWan:
Hi, I have split the text data processing part. DocumentTokenizer.scala is a transformer that will tokenize all the sentences from disk file. (a stanford-parser tokenizer) The other people can use this transformer too. A Dictionary.scala that can either create dictionary from Array[Array[String]] (namely tokenized sentences) or from disk files (dictionary.txt that has already been created.)

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/Dictionary.scala
```diff
@@ -0,0 +1,132 @@
...
+     _word2index.foreach(x =>
+       logger.info(x._1 + " -> " + x._2))
+   }
+
+   def printDiscard(): Unit = {
+     _disvardVocab.foreach(x =>
+       logger.info(x))
+   }
+
+   def this(sentences: Array[Array[String]],
```
psyyz10: How about RDD[Array[String]] ?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/Dictionary.scala
```diff
@@ -0,0 +1,132 @@
...
+     _word2index = _vocabulary.zipWithIndex.toMap
+     _index2word = _word2index.map(x => (x._2, x._1))
+     _disvardVocab = freqDict.take(freqDict.length - length).map(_._1)
+     _discardSize = _disvardVocab.length
+   }
+
+   def this(directory: String) = {
+     this()
+
+     val dictionaryFile = new File(directory, "dictionary.txt")
```
psyyz10: Maybe we need to read file from HDFS or with spark `sc.textFile("data.txt")` ?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/Dictionary.scala
```diff
@@ -0,0 +1,132 @@
...
+   private var _discardSize: Int = 0
+   private var _word2index: Map[String, Int] = null
+   private var _index2word: Map[Int, String] = null
+   private var _vocabulary: Seq[String] = null
+   private var _disvardVocab: Seq[String] = null
+}
+
+object Dictionary {
+  def apply(sentences: Array[Array[String]],
+            vocabSize: Int)
```
psyyz10: It is better to set vocabSize a default value, which means the user do not want to add any restrictions to the size of vocabulary?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/Dictionary.scala
```diff
@@ -0,0 +1,132 @@
...
+     _discardSize = _disvardVocab.length
+   }
+
+   val logger = Logger.getLogger(getClass)
+   private var _vocabSize: Int = 0
+   private var _discardSize: Int = 0
+   private var _word2index: Map[String, Int] = null
+   private var _index2word: Map[Int, String] = null
+   private var _vocabulary: Seq[String] = null
+   private var _disvardVocab: Seq[String] = null
```
psyyz10: Is it better to name `_discardVocab`?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/Dictionary.scala
```diff
@@ -0,0 +1,132 @@
...
+       .toMap[String, Int]
+     _index2word = _word2index.map(x => (x._2, x._1))
+     _vocabulary = _word2index.keys.toSeq
+     _vocabSize = _word2index.size
+     _disvardVocab = Source.fromFile(discardFile.getAbsolutePath)
+       .getLines().toSeq
+     _discardSize = _disvardVocab.length
+   }
+
+   val logger = Logger.getLogger(getClass)
```
psyyz10: In my opinion, we offten need the frequency of the words, is it a bad idea to add frequencies as a member here as well?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/Dictionary.scala
psyyz10: Is it better to add arguments such as 'needDiscard: Boolean = true' to set whether the user need discardVocab? I'm not sure :-)

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/Dictionary.scala
```diff
@@ -0,0 +1,132 @@
...
+   }
+
+   def printDiscard(): Unit = {
+     _disvardVocab.foreach(x =>
+       logger.info(x))
+   }
+
+   def this(sentences: Array[Array[String]],
+            vocabSize: Int) = {
+     this()
```
psyyz10: How about use some RDD transformations? which could be much faster

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/Dictionary.scala
```diff
@@ -0,0 +1,132 @@
...
+     val freqDict = sentences
+       .flatMap(x => x)
+       .foldLeft(Map.empty[String, Int]) {
+         (count, word) => count + (word -> (count.getOrElse(word, 0) + 1))
+       }.toSeq.sortBy(_._2)
+
```
psyyz10: How about use some RDD transformations? which could be much faster

yiheng:
Don't commit other's change

PR #344, opened 2017-01-05
fix #343: fix the wrong padding offset

### LINKED ISSUE #343: BGRImgRdmCropper's padding offset is incorrect

fanshiqing:


fanshiqing:
@JerryYanWan

PR #346, opened 2017-01-06
[Issue #370]Word2Vec (WorkInProgress)

psyyz10:
For issue #370


dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/LowerCase.scala
```diff
@@ -0,0 +1,32 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
```
fanshiqing: License: ASF => Intel Coporation
The same as below.

psyyz10: Changed, and thanks a lot :+1:

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/Tokenizer.scala
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * @param sep the delimiting regular expression to split on.
+ *            Defaults to matching all punctuation and whitespace
+ */
+class Tokenizer(sep: String) extends Transformer[String, Seq[String]] {
+  override def apply(prev: Iterator[String]): Iterator[Seq[String]] =
+    prev.map(x => x.split(sep))
+}
+
+object Tokenizer {
```
yiheng: BTW, spark mllib also have a tokenizer...

https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/ml/feature/Tokenizer.scala

psyyz10: Yan's version is not exactly the same as I want. Zhichao's version is similar to what I want, it is possible to combine to one, I will discuss with him. Thx.

psyyz10: Thanks for reminder, spark mllib's RegexTokenizer is exactly what I want, maybe I need to see how to use it and insert to the transformer pipeline

linbojin:
Do we still want to add Word2Vec?

psyyz10:
@linbojin This PR's priority is low, and the network is shallow so not suitable for Synchronize SGD with all-reduce. Spark MLlib's word2vec can be used as well with the same function. Do you have any requirement on this PR?

Issue #340, opened 2017-01-04
MacOS support


yiheng:
The native code of BigDL is only compiled on Linux-x64 platform. This is not convenient for MacOSdeveloper to develop or test code.

Pre-build native package should be supported on MacOS platform.

wydwww:
.jar for macOS is done.

yiheng:
Instead of provide different jar, we can package linux so file and macOS so file into one jar and load different so based on the current OS. Somewhere might be changed: https://github.com/intel-analytics/BigDL/blob/master/native/jni/src/main/java/com/intel/analytics/bigdl/mkl/MKL.java#L20

jason-dai:
The downside of this approach is that the resulting jar can be too big though

wydwww:
In macOS it's actually libiomp5.dylib instead of libiomp5.so @yiheng

jason-dai:
I don't think we need windows support at this moment, yes?

PR #349, opened 2017-01-06
Add macOS support

wydwww:
Tests passed on macOS(local and spark local) and ubuntu(local and spark local & cluster).
On ubuntu: same with before.
On macOS: need a full build, first install Intel Parallel Studio XE, then `bash make-dist.sh -P full-build -P mac`.
#340

### DIFF
--- a/dl/pom.xml
+++ b/dl/pom.xml
@@ -38,7 +38,7 @@
         </dependency>
         <dependency>
             <groupId>com.intel.analytics.bigdl.native</groupId>
-            <artifactId>mkl-java</artifactId>
+            <artifactId>${mkl-java-os-version}</artifactId>
             <version>${project.version}</version>
             <exclusions>
                 <!-- We already copy the dynamic lib files of this project to mkl-java, so we
--- a/native/jni/pom.xml
+++ b/native/jni/pom.xml
@@ -10,7 +10,7 @@
     <modelVersion>4.0.0</modelVersion>
 
     <groupId>com.intel.analytics.bigdl.native</groupId>
-    <artifactId>mkl-java</artifactId>
+    <artifactId>${mkl-java-os-version}</artifactId>
     <packaging>jar</packaging>
 
     <dependencies>
@@ -32,7 +32,7 @@
             <resource>
                 <directory>src/main/resources/intel64</directory>
                 <includes>
-                    <include>libiomp5.so</include>
+                    <include>${iompLib}</include>
                 </includes>
                 <targetPath>${project.build.directory}/classes</targetPath>
             </resource>
@@ -58,10 +58,10 @@
                                     <groupId>com.intel.analytics.bigdl.native</groupId>
                                     <artifactId>bigdl-native</artifactId>
                                     <version>${pom.version}</version>
-                                    <type>so</type>
+                                    <type>${dynamicLibType}</type>
                                     <overWrite>false</overWrite>
                                     <outputDirectory>${project.build.directory}/classes</outputDirectory>
-                                    <destFileName>libjmkl.so</destFileName>
+                                    <destFileName>${destFileName}</destFileName>
                                 </artifactItem>
                             </artifactItems>
                         </configuration>
--- a/native/jni/pom.xml
+++ b/native/jni/pom.xml
@@ -23,7 +23,7 @@
             <groupId>com.intel.analytics.bigdl.native</groupId>
             <artifactId>bigdl-native</artifactId>
             <version>${project.version}</version>
-            <type>so</type>
+            <type>${dynamicLibType}</type>
         </dependency>
     </dependencies>
 
--- a/native/jni/src/main/java/com/intel/analytics/bigdl/mkl/MKL.java
+++ b/native/jni/src/main/java/com/intel/analytics/bigdl/mkl/MKL.java
@@ -17,13 +17,20 @@ public class MKL {
 
     static {
         try {
-            tmpFile = extract("libiomp5.so");
+            String iomp5FileName = "libiomp5.so";
+            String jmklFileName = "libjmkl.so";
+            if (System.getProperty("os.name").toLowerCase().contains("mac")) {
+                iomp5FileName = "libiomp5.dylib";
+                jmklFileName = "libjmkl.dylib";
+            }
+            tmpFile = extract(iomp5FileName);
             System.load(tmpFile.getAbsolutePath());
             tmpFile.delete(); // delete so temp file after loaded
-            tmpFile = extract("libjmkl.so");
+            tmpFile = extract(jmklFileName);
             System.load(tmpFile.getAbsolutePath());
             tmpFile.delete(); // delete so temp file after loaded
             isLoaded = true;
+
         } catch (Exception e) {
             isLoaded = false;
             e.printStackTrace();
--- a/native/jni/src/main/resources/intel64/libiomp5.dylib
+++ b/native/jni/src/main/resources/intel64/libiomp5.dylib
@@ -1,0 +1,1 @@
+/opt/intel/lib/libiomp5.dylib
--- a/native/pom.xml
+++ b/native/pom.xml
@@ -16,6 +16,23 @@
         <module>mkl</module>
         <module>jni</module>
     </modules>
-
-
+    <properties>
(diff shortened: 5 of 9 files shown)


linbojin:
There is free Intel Parallel Studio XE for macOS?

wydwww:
I'm not sure... You can try the trial version for now. Maybe we'll release a partial-pre-build version to avoid full build on mac later. @linbojin

linbojin:
Thx, @wydwww

jason-dai:
We plan to release jars for native lib code for both Linux and macOS afterwards, so that users will just need to do a normal build for macOS.

native/jni/src/main/java/com/intel/analytics/bigdl/mkl/MKL.java
```diff
@@ -17,13 +17,24 @@
...
     static {
         try {
-            tmpFile = extract("libiomp5.so");
-            System.load(tmpFile.getAbsolutePath());
-            tmpFile.delete(); // delete so temp file after loaded
-            tmpFile = extract("libjmkl.so");
-            System.load(tmpFile.getAbsolutePath());
-            tmpFile.delete(); // delete so temp file after loaded
-            isLoaded = true;
+            if (System.getProperty("os.name").toLowerCase().contains("mac")) {
```
yiheng: too much duplicate code here?  I think you just need set the file name based on os.name

native/pom.xml
```diff
@@ -16,6 +16,31 @@
...
+    <profiles>
+        <profile>
+            <id>default</id>
+            <activation>
+                <activeByDefault>true</activeByDefault>
+            </activation>
+            <properties>
+                <compilerStartOptionDirectory>-I ${JAVA_HOME}/include/linux/</compilerStartOptionDirectory>
+                <linkerStartOptionDirectory>-I ${JAVA_HOME}/include/linux/</linkerStartOptionDirectory>
+                <packagingDirectory>so</packagingDirectory>
```
yiheng: packagingDirectory -> dynamicLibType?

native/pom.xml
```diff
@@ -16,6 +16,31 @@
...
+                <includeDirectory>libiomp5.so</includeDirectory>
```
yiheng: includeDirectory -> iompLib?

native/pom.xml
```diff
@@ -16,6 +16,31 @@
...
+                <typeDirectory>so</typeDirectory>
```
yiheng: reuse dynamicLibType?

native/pom.xml
```diff
@@ -16,6 +16,31 @@
...
+                <destFileNameDirectory>libjmkl.so</destFileNameDirectory>
```
yiheng: destFileNameDirectory -> destFileName?

BTW, why you always have a Directory suffix...

native/jni/src/main/resources/intel64/libiomp5.dylib
```diff
@@ -0,0 +1 @@
+/opt/intel/lib/libiomp5.dylib
```
yiheng: what's this for?

yiheng:
I think in the script, you should change the artifactId of jni module instead of native module? We'll upload two jni jar to the public repo. jni(default for linux x64) and jni-mac. So user on their mac only need to type mvn package -P mac, they will download the jni-mac jar instead of jni jar. So they needn't install Parallel studio or compile native code.

yiheng:
In this case, I think we need a script to change the artifiactId but a just another module jni-mac. And we can active the mac profile base on os. No need user input -P mac. See http://maven.apache.org/guides/introduction/introduction-to-profiles.html for how to do it.

jason-dai:
I don't think using script for this type of tricks is needed; just add a parameter and then the user can manage it explicitly.

native/jni/src/main/resources/intel64/libiomp5.dylib
wydwww: Soft link to libiomp5.dylib of intel parallel studio, just like libiomp5.so

yiheng: I think it's not a good way to handle this, as the path is hard coded..

native/jni/src/main/java/com/intel/analytics/bigdl/mkl/MKL.java
```diff
@@ -17,24 +17,20 @@
 
     static {
         try {
+            String fileName1 = "libiomp5.so";
```
yiheng: don't use fileName1 and fileName2 as variable name. It's too confusing

native/pom.xml
```diff
@@ -25,21 +25,19 @@
             <properties>
                 <compilerStartOptionDirectory>-I ${JAVA_HOME}/include/linux/</compilerStartOptionDirectory>
                 <linkerStartOptionDirectory>-I ${JAVA_HOME}/include/linux/</linkerStartOptionDirectory>
-                <packagingDirectory>so</packagingDirectory>
-                <includeDirectory>libiomp5.so</includeDirectory>
-                <typeDirectory>so</typeDirectory>
-                <destFileNameDirectory>libjmkl.so</destFileNameDirectory>
+                <dynamicLibType>so</dynamicLibType>
```
yiheng: I don't think the default profile is necessary.

dl/pom.xml
```diff
@@ -244,6 +244,7 @@
             <properties>
                 <spark-version.version>1.5-plus</spark-version.version>
                 <spark.version>1.5.1</spark.version>
+                <mkl-java-os-version>mkl-java</mkl-java-os-version>
```
yiheng: move this property to bigdl-parent pom.xml

dl/pom.xml
```diff
@@ -259,5 +260,11 @@
                 <spark-scope>compile</spark-scope>
             </properties>
         </profile>
+        <profile>
```
yiheng: move this profile to bigdl-parent pom.xml

scripts/change-os.sh
```diff
@@ -58,6 +58,6 @@ sed_i() {
 
 export -f sed_i
 
-BASEDIR=$(dirname $0)/../native/mkl
+BASEDIR=$(dirname $0)/../native/jni
```
yiheng: we don't need this script

native/jni/pom.xml
```diff
@@ -10,7 +10,7 @@
     <modelVersion>4.0.0</modelVersion>
 
     <groupId>com.intel.analytics.bigdl.native</groupId>
-    <artifactId>mkl-java</artifactId>
+    <artifactId>mkl-java-mac</artifactId>
```
yiheng: you should not change it.

scripts/change-os.sh
yiheng: create a new sub-module project

native/pom.xml
wydwww: You mean choose the properties automatically by OS?

wydwww: You mean choose the properties automatically by OS? @yiheng

yiheng: no. You needn't a default profile to make the property as default

wydwww: Because of the `activeByDefault` tag? So I just need to change `default` to `linux`? @yiheng

yiheng: No. You needn't a profile to make default value of property.

linbojin:
I think you may need to rebase your long commits...

pom.xml
```diff
@@ -514,9 +516,21 @@
             <id>full-build</id>
             <modules>
                 <module>native</module>
+                <module>dl</module>
```
yiheng: don't add this back

native/pom.xml
```diff
@@ -26,6 +26,11 @@
     <profiles>
         <profile>
             <id>mac</id>
+            <activation>
```
jason-dai: I don't think we want to do this automatically; instead, have the user to explicitly enable this when needed.

wydwww: Fixed. Without `-P mac` it will build the linux jar on macOS. @jason-dai

wydwww:
Successfully build on shiqing's mac. @yiheng

yiheng:
test looks good. merged

yiheng:
The pre-build package has been uploaded to sonatype

PR #353, opened 2017-01-09
Distribute partition evenly on each node

yiheng:
In the training process, BigDL will issue one spark task on each node. Currently, we coalesce train data RDD with shuffle to node number to achieve this. But this is not safe, as the task scheduler can still schedule different partitions into a same node.  A solution is coalesce train data RDD to node number * core number with shuffle, so the data will distributed evenly on each node. Then coalesce without shuffle to node number. This is what the PR do.

### DIFF
--- a/dl/src/main/scala/com/intel/analytics/bigdl/dataset/DataSet.scala
+++ b/dl/src/main/scala/com/intel/analytics/bigdl/dataset/DataSet.scala
@@ -285,13 +285,14 @@ object DataSet {
    * @return
    */
   def array[T: ClassTag](localData: Array[T], sc: SparkContext): DistributedDataSet[T] = {
-    val partitionNum = Engine.nodeNumber()
+    val nodeNumber = Engine.nodeNumber()
       .getOrElse(throw new RuntimeException("can't get node number? Have you initialized?"))
+    val coreNumber = Engine.coreNumber()
 
     new CachedDistriDataSet[T](
-      sc.parallelize(localData, partitionNum)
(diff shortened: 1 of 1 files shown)


linbojin:
IMO, spark will schedule tasks based on partitions (one task for one partition) and for performance purpose, number of partitions should be 2 * cores for each node. `BigDL will issue one spark task on each node` is this designed by bigdl?

yiheng:
@linbojin We do this only in training process to get a better performance. 

In the training process, because model need to sync parameter for mini-batch SGD and to reduce the thread context switch, bigdl will place one partition of train dataset on one node(which means bigdl will re-partition the training dataset to node number). So there's one computing task on one node.

Please note that it's only for training process. You can use bigdl model just as normal spark transformer in prediction. See this example: 
https://github.com/intel-analytics/BigDL/tree/master/dl/src/main/scala/com/intel/analytics/bigdl/example/imageclassification

linbojin:
Make sense, thx for your explanation.

yiheng:
Just clarify the root cause of the partition not distributed evenly among the cluster is the partition number is too small. The real fix is this line...

https://github.com/intel-analytics/BigDL/pull/353/files#diff-628659303f629f06eb48d9af1d343db7R293

The coalesce changes has been reverted in https://github.com/intel-analytics/BigDL/pull/368/commits/3f17840f814b7eb3b29bb486b7ce8fbd7ad22e15

Sorry for the confusion.

PR #362, opened 2017-01-09
modify minibatch to support both Table and Tensor input

JerryYanWan:


dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -126,18 +128,36 @@ object DistriOptimizer {
...
+            }
+            val stackSize = tensorBatchType match {
+              case true => batch.data.toTensor[T].size(1) / _subModelNumber
+              case false => batch.data.toTable.length / _subModelNumber
+            }
             while (b < _subModelNumber) {
-              tensorBuffer(b) = (batch.data.narrow(1, b * stackSize + 1, stackSize),
-                batch.labels.narrow(1, b * stackSize + 1, stackSize))
+              batchBuffer(b) = tensorBatchType match {
+                case true => (batch.data.toTensor[T].narrow (1, b * stackSize + 1, stackSize),
```
psyyz10: toTensor[T] can be replaced by 'toTensor', because the 'T' can be implicitly inferred

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -126,18 +128,36 @@ object DistriOptimizer {
...
+                  batch.labels.toTensor[T].narrow (1, b * stackSize + 1, stackSize))
+                case false =>
```
psyyz10: As above

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -126,18 +128,36 @@ object DistriOptimizer {
...
+                  (NarrowTable(b * stackSize + 1, stackSize).updateOutput(batch.data.toTable),
```
psyyz10: It is better to create a NarrowTable out of the while loop. In that case, the narrowTable will not be created each time.
Like that:
define `val narrowTable = NarrowTable(b * stackSize + 1, stackSize)` in line 153, 
and call `narrowTable.updateOutput(batch.data.toTable)` here

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -429,17 +452,34 @@ object DistriOptimizer {
...
-              val input = batch.data.narrow(1, offset + 1, length)
-              val target = batch.labels.narrow(1, offset + 1, length)
+              val input = tensorBatchType match {
+                case true => batch.data.toTensor[T].narrow(1, offset + 1, length)
+                case false => NarrowTable[T](offset + 1, length).updateOutput(batch.data.toTable)
+              }
+              val target = tensorBatchType match {
+                case true => batch.labels.toTensor[T].narrow(1, offset + 1, length)
+                case false => NarrowTable(offset + 1, length).updateOutput(batch.labels.toTable)
+              }
```
psyyz10: It can also be defined as something like:
val (input, target) = tensorBatchType match {
                case true => (batch.data.toTensor.narrow(1, offset + 1, length),
                                       batch.labels.toTensor.narrow(1, offset + 1, length)
                case false => (narrowTable.updateOutput(batch.data.toTable),
                                       narrowTable.updateOutput(batch.labels.toTable))
              }

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
JerryYanWan: fixed

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
JerryYanWan: fixed

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
JerryYanWan: b is updated within the while loop, so it cannot be defined outside

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
JerryYanWan: revised

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -174,7 +194,10 @@ object DistriOptimizer {
           driverMetrics.add("computing time for each node", computingTime)
 
           val finishedThreads = trainingThreads.filter(!_.isCancelled).map(_.get())
-          recordsNum += finishedThreads.size * tensorBuffer.head._2.size(1)
+          recordsNum += finishedThreads.size * (batchBuffer.head._2 match {
```
yiheng: same

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -429,17 +452,32 @@ object DistriOptimizer {
       val workingModels = modelIter.next().localModels
       workingModels.foreach(_.evaluate())
       dataIter.map(batch => {
-        require(batch.data.size(1) == batch.labels.size(1))
-        val stackSize = batch.data.size(1) / _subModelNumber
-        val extraSize = batch.data.size(1) % _subModelNumber
+        val tensorBatchType = batch.data match {
```
yiheng: same

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -429,17 +452,32 @@ object DistriOptimizer {
...
+          case tensor: Tensor[T] =>
+            require(batch.data.toTensor[T].size(1) == batch.labels.toTensor[T].size(1))
+            true
+          case table: Table =>
+            require(batch.data.toTable.length == batch.labels.toTable.length)
+            false
+        }
+        val (stackSize, extraSize) = tensorBatchType match {
```
yiheng: same

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -429,17 +452,32 @@ object DistriOptimizer {
...
+        }
         val parallelism = if (stackSize == 0) extraSize else _subModelNumber
         Engine.default.invokeAndWait(
           (0 until parallelism).map(b =>
             () => {
               val offset = b * stackSize + math.min(b, extraSize)
               val length = stackSize + (if (b < extraSize) 1 else 0)
-              val input = batch.data.narrow(1, offset + 1, length)
-              val target = batch.labels.narrow(1, offset + 1, length)
+              val (input, target) = tensorBatchType match {
```
yiheng: same

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriValidator.scala
```diff
@@ -51,17 +55,34 @@ class DistriValidator[T] private[optim](
       val workingModels = (1 to _subModelNumber)
         .map(_ => localModel.cloneModule().evaluate()).toArray
       dataIter.map(batch => {
-        require(batch.data.size(1) == batch.labels.size(1))
-        val stackSize = batch.data.size(1) / _subModelNumber
-        val extraSize = batch.data.size(1) % _subModelNumber
+        val tensorBatchType = batch.data match {
```
yiheng: same

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriValidator.scala
```diff
@@ -51,17 +55,34 @@ class DistriValidator[T] private[optim](
...
+          case tensor: Tensor[T] =>
+            require(batch.data.toTensor[T].size(1) == batch.labels.toTensor[T].size(1))
+            true
+          case table: Table =>
+            require(batch.data.toTable.length == batch.labels.toTable.length)
+            false
+        }
+        val (stackSize, extraSize) = tensorBatchType match {
```
yiheng: same

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriValidator.scala
```diff
@@ -51,17 +55,34 @@ class DistriValidator[T] private[optim](
...
+        }
         val parallelism = if (stackSize == 0) extraSize else _subModelNumber
         Engine.default.invokeAndWait(
           (0 until parallelism).map(b =>
             () => {
               val offset = b * stackSize + math.min(b, extraSize)
               val length = stackSize + (if (b < extraSize) 1 else 0)
-              val input = batch.data.narrow(1, offset + 1, length)
-              val target = batch.labels.narrow(1, offset + 1, length)
+              val input = tensorBatchType match {
```
yiheng: same

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriValidator.scala
```diff
@@ -51,17 +55,34 @@ class DistriValidator[T] private[optim](
...
+                case true => batch.data.toTensor[T].narrow(1, offset + 1, length)
+                case false => NarrowTable[T](offset + 1, length).updateOutput(batch.data.toTable)
+              }
+              val target = tensorBatchType match {
```
yiheng: same

dl/src/main/scala/com/intel/analytics/bigdl/optim/LocalOptimizer.scala
```diff
@@ -91,16 +93,31 @@ class LocalOptimizer[T: ClassTag] private[optim](
       // Fetch data and prepare tensors
       val batch = iter.next()
       var b = 0
-      require(batch.data.size(1) == batch.labels.size(1))
-      val stackSize = batch.data.size(1) / subModelNumber
-      val extraSize = batch.data.size(1) % subModelNumber
+      val tensorBatchType = batch.data match {
```
yiheng: same

dl/src/main/scala/com/intel/analytics/bigdl/optim/LocalOptimizer.scala
```diff
@@ -91,16 +93,31 @@ class LocalOptimizer[T: ClassTag] private[optim](
...
+        case tensor: Tensor[T] =>
+          require(batch.data.toTensor[T].size(1) == batch.labels.toTensor[T].size(1))
+          true
+        case table: Table =>
```
yiheng: same

dl/src/main/scala/com/intel/analytics/bigdl/optim/LocalOptimizer.scala
```diff
@@ -91,16 +93,31 @@ class LocalOptimizer[T: ClassTag] private[optim](
...
+      }
       val parallelism = if (stackSize == 0) extraSize else subModelNumber
-      val tensorBuffer = new Array[(Tensor[T], Tensor[T])](parallelism)
+      val batchBuffer = new Array[(Activity, Activity)](parallelism)
       while (b < parallelism) {
         val offset = b * stackSize + math.min(b, extraSize)
         val length = stackSize + (if (b < extraSize) 1 else 0)
-        tensorBuffer(b) = (batch.data.narrow(1, offset + 1, length),
-          batch.labels.narrow(1, offset + 1, length))
+        batchBuffer(b) = tensorBatchType match {
```
yiheng: same

dl/src/main/scala/com/intel/analytics/bigdl/optim/LocalOptimizer.scala
```diff
@@ -146,14 +163,21 @@ class LocalOptimizer[T: ClassTag] private[optim](
       optimMethod.optimize(_ => (ev.fromType(loss), grad), weight, state)
       val end = System.nanoTime()
       wallClockTime += end - start
-      count += batch.data.size(1)
+      count += (tensorBatchType match {
```
yiheng: same

dl/src/main/scala/com/intel/analytics/bigdl/optim/LocalOptimizer.scala
```diff
@@ -146,14 +163,21 @@ class LocalOptimizer[T: ClassTag] private[optim](
...
+        case false => batch.data.toTable.length
+      })
       val head =
         header(state[Int]("epoch"), count, dataset.size(), state[Int]("neval"), wallClockTime)
       logger.info(s"$head " +
         s"loss is $loss, iteration time is ${(end - start) / 1e9}s " +
         s"data fetch time is ${(dataFetchTime - start) / 1e9}s, " +
         s"train time ${(end - dataFetchTime) / 1e9}s. " +
-        s"Throughput is ${batch.data.size(1).toDouble / (end - start) * 1e9} record / second")
+        s"Throughput is " +
```
yiheng: same

dl/src/main/scala/com/intel/analytics/bigdl/optim/LocalOptimizer.scala
```diff
@@ -199,18 +223,35 @@ class LocalOptimizer[T: ClassTag] private[optim](
 
     var count = 0
     dataIter.map(batch => {
-      require(batch.data.size(1) == batch.labels.size(1))
-      val stackSize = batch.data.size(1) / subModelNumber
-      val extraSize = batch.data.size(1) % subModelNumber
+      val tensorBatchType = batch.data match {
```
yiheng: same

dl/src/main/scala/com/intel/analytics/bigdl/optim/LocalOptimizer.scala
```diff
@@ -199,18 +223,35 @@ class LocalOptimizer[T: ClassTag] private[optim](
...
       val parallelism = if (stackSize == 0) extraSize else subModelNumber
       val start = System.nanoTime()
       val result = Engine.default.invokeAndWait(
         (0 until parallelism).map(b =>
           () => {
             val offset = b * stackSize + math.min(b, extraSize)
             val length = stackSize + (if (b < extraSize) 1 else 0)
-            val input = batch.data.narrow(1, offset + 1, length)
-            val target = batch.labels.narrow(1, offset + 1, length)
+            val input = tensorBatchType match {
```
yiheng: same

dl/src/main/scala/com/intel/analytics/bigdl/optim/LocalOptimizer.scala
```diff
@@ -222,9 +263,15 @@ class LocalOptimizer[T: ClassTag] private[optim](
           l + r
         }
       })
-      count += batch.data.size(1)
+      count += (tensorBatchType match {
```
yiheng: same

dl/src/main/scala/com/intel/analytics/bigdl/optim/LocalOptimizer.scala
```diff
@@ -222,9 +263,15 @@ class LocalOptimizer[T: ClassTag] private[optim](
...
+        case true => batch.data.toTensor[T].size(1)
+        case false => batch.data.toTable.length
+      })
       logger.info(s"[Validation] $count/${validationDataSet.get.size()} Throughput is ${
-        batch.data.size(1) / ((System.nanoTime() - start) / 1e9)
+        (tensorBatchType match {
```
yiheng: same

dl/src/main/scala/com/intel/analytics/bigdl/optim/LocalValidator.scala
```diff
@@ -19,15 +19,21 @@ package com.intel.analytics.bigdl.optim
 
 import com.intel.analytics.bigdl.dataset.{LocalDataSet, MiniBatch}
 import com.intel.analytics.bigdl._
+import com.intel.analytics.bigdl.nn.NarrowTable
```
yiheng: all same with above

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -119,16 +123,37 @@ class SampleToBatch[T: ClassTag]
             sample.copyToLabel(labelData, i*labelLength, labelLength)
             sample.copyToFeature(featureData, i*featureLength, featureLength)
 
+            featureTable(i + 1) = Tensor()
```
yiheng: The table backend is a hashmap. Is it effecient? should we implement a table with an array backend?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Types.scala
```diff
@@ -89,6 +90,7 @@ class Sample[T: ClassTag] (
 
   def copyToFeature(storage: Array[T], offset: Int, length: Int): Unit = {
     require(offset + length <= storage.length, "index out of boundary")
+    require(length <= featureTensor.storage().array().length, "length too long for feature")
```
JerryYanWan: copyToFeature will copy the class private variable feature out. It assumes the offset of featureTensor.storage always starts from 1, which is initialized by the copy() method.

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Types.scala
```diff
@@ -102,6 +104,7 @@ class Sample[T: ClassTag] (
 
   def copyToLabel(storage: Array[T], offset: Int, length: Int): Unit = {
```
JerryYanWan: copyToLabel  and copyToFeature will not change the featureTensor and labelTensor. They just copy the featureTensor or labelTensor out.

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Types.scala
```diff
@@ -114,8 +117,8 @@ class Sample[T: ClassTag] (
   }
 
   def copy(other: Sample[T]): Sample[T] = {
-    featureTensor.copy(other.featureTensor)
-    labelTensor.copy(other.labelTensor)
+    this.featureTensor.resizeAs(other.featureTensor).copy(other.featureTensor)
+    this.labelTensor.resizeAs(other.labelTensor).copy(other.labelTensor)
     this
   }
```
JerryYanWan: renamed

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Types.scala
yiheng: ok. Please add a comment about this.

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Types.scala
yiheng: so it's not a copyTo? is the name misleading

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
JerryYanWan: batchBuffer.head will return an (Activity, Acitivity) format, which is not the MiniBatch type.

yiheng: batchBuffer should be MiniBatch, right? At least you name it as **batch**Buffer

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Types.scala
JerryYanWan: how about using the same copy() name, but overload all the methods?

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
JerryYanWan: The previous name was tensorBuffer. I changed it to batchBuffer because the variable type for MiniBatch has been changed to support Activity. How about dataBuffer?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Types.scala
yiheng: I think it's more confusing. extractLabel or extractFeature should be what it means here?

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
yiheng: I'm talking about the return value. As you name it as batchBuffer, MiniBatch should be a more proper return type, right?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -72,20 +73,23 @@ class Identity[A] extends Transformer[A, A] {
 
 object SampleToBatch {
   def apply[T: ClassTag]
-  (batchSize: Int)
+  (batchSize: Int,
+   tableOfBatch: Boolean = false)
```
yiheng: tableOfBatch -> isTable?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -72,20 +73,23 @@ class Identity[A] extends Transformer[A, A] {
...
-  (totalBatch: Int)
+  (totalBatch: Int, tableOfBatch: Boolean = false)
   (implicit ev: TensorNumeric[T])
   extends Transformer[Sample[T], MiniBatch[T]] {
 
   override def apply(prev: Iterator[Sample[T]]): Iterator[MiniBatch[T]] = {
     new Iterator[MiniBatch[T]] {
       private val featureTensor: Tensor[T] = Tensor[T]()
       private val labelTensor: Tensor[T] = Tensor[T]()
+      private val featureTable: Table = T()
```
yiheng: featureTensor and featureTable should not occur in the same time, right?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -116,19 +120,40 @@ class SampleToBatch[T: ClassTag]
             if (labelData == null || labelData.length < batchSize * labelLength) {
               labelData = new Array[T](batchSize * labelLength)
             }
-            sample.copyToLabel(labelData, i*labelLength, labelLength)
-            sample.copyToFeature(featureData, i*featureLength, featureLength)
+            sample.copyLabel(labelData, i*labelLength, labelLength)
+            sample.copyFeature(featureData, i*featureLength, featureLength)
 
+            featureTable(i + 1) = Tensor()
```
yiheng: can we reuse the tensor instead create each time?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -116,19 +120,40 @@ class SampleToBatch[T: ClassTag]
             if (labelData == null || labelData.length < batchSize * labelLength) {
               labelData = new Array[T](batchSize * labelLength)
             }
-            sample.copyToLabel(labelData, i*labelLength, labelLength)
-            sample.copyToFeature(featureData, i*featureLength, featureLength)
+            sample.copyLabel(labelData, i*labelLength, labelLength)
```
yiheng: extractLabel?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -116,19 +120,40 @@ class SampleToBatch[T: ClassTag]
...
+            sample.copyFeature(featureData, i*featureLength, featureLength)
```
yiheng: extractFeature?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Types.scala
```diff
@@ -87,50 +90,61 @@ class Sample[T: ClassTag] (
...
 
-  def copyToFeature(storage: Array[T], offset: Int, length: Int): Unit = {
+  def copyFeature(storage: Array[T], offset: Int, length: Int): Unit = {
```
yiheng: extractFeature?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Types.scala
```diff
@@ -87,50 +90,61 @@ class Sample[T: ClassTag] (
...
-        .asInstanceOf[Array[Float]], 0, storage
+        .asInstanceOf[Array[Float]],
+        featureTensor.storageOffset() - 1, storage
         .asInstanceOf[Array[Float]], offset, length)
       case _ => throw new UnsupportedOperationException(s"Only Float/Double supported")
     }
   }
 
-  def copyToLabel(storage: Array[T], offset: Int, length: Int): Unit = {
+  def copyLabel(storage: Array[T], offset: Int, length: Int): Unit = {
```
yiheng: extractLabel?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Types.scala
```diff
@@ -89,6 +90,7 @@ class Sample[T: ClassTag] (
```
yiheng: Have you changes against the above comments?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Types.scala
```diff
@@ -152,7 +166,53 @@ trait Label[T] {
  * @param labels
  * @tparam T
  */
-case class MiniBatch[T](data: Tensor[T], labels: Tensor[T])
+case class MiniBatch[T: ClassTag](data: Activity, labels: Activity)
```
yiheng: The class is long. should we move this class to another file?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Types.scala
yiheng: The class is long. should we move this class to another file?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Types.scala
```diff
@@ -152,7 +166,53 @@ trait Label[T] {
...
+  private def labelLength(): Int = {
+    labels match {
+      case tensor: Tensor[T] => tensor.size(1)
+      case table: Table => table.length
+      case _ => throw new IllegalArgumentException("MiniBatch only supports Table or Tensor")
+    }
+  }
+
+  def size(): Int = {
+    dataLength()
```
yiheng: should we get the dataLength when we construct it insteadOf calculate it every time?

dl/src/main/scala/com/intel/analytics/bigdl/models/rnn/Test.scala
```diff
@@ -65,7 +65,7 @@ object Test {
 
         val dataIter = validationSet.data(train = false)
         val predict = dataIter.map(batch => {
-          require(batch.data.size(1) == 1, "predict sentence one by one")
+          require(batch.data.toTensor[Float].size(1) == 1, "predict sentence one by one")
```
yiheng: why we need toTensor here? why not use batch length?

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -160,7 +160,7 @@ object DistriOptimizer {
               val localModel = cached.localModels(i)
               localModel.training()
               val localCriterion = cached.localCriterions(i)
-              val (input, target) = tensorBuffer(i)
+              val (input, target) = MiniBatch.unapply(batchBuffer(i)).get
```
yiheng: This looks strange. unapply should use with match. If you don't use match, you needn't unapply but a custom function return a Tuple2.

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Types.scala
```diff
@@ -152,7 +166,53 @@ trait Label[T] {
...
+  }
+
+  def narrow(dim: Int, index: Int, size: Int): (Activity, Activity) = {
```
yiheng: why return a (Activity, Activity) instead of MiniBatch?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
JerryYanWan: renamed

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
JerryYanWan: use if-else to branch the case

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
JerryYanWan: Yes, copy Label from Sample and write it to the labelData[Array]

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
JerryYanWan: Yes, copy feature from Sample and write it to the featureData[Array]

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
JerryYanWan: Shall we create an array of Tensor, the size of array set to batchSize, to reuse?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Types.scala
JerryYanWan: yes

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Types.scala
JerryYanWan: yes

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Types.scala
JerryYanWan: under the dataset directory?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Types.scala
JerryYanWan: fixed

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Types.scala
JerryYanWan: fixed

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
JerryYanWan: fixed

dl/src/main/scala/com/intel/analytics/bigdl/models/rnn/Test.scala
JerryYanWan: fixed

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Types.scala
JerryYanWan: java doc not included yet. Currently, haven't changed tensor.set ..

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Types.scala
yiheng: Yes

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Types.scala
yiheng: I mean rename it to extractLabel?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
yiheng: but the data size may not be batchSize

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
yiheng: I mean rename it to extractFeature?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
yiheng: I mean rename it to extractLabel?

JerryYanWan: renamed

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
JerryYanWan: renamed

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
JerryYanWan: that's the problem. If we reuse the same Tensor, then the tensors in the featureTable will be the same at last?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -95,18 +99,26 @@ class SampleToBatch[T: ClassTag]
...
+      }
+
       override def next(): MiniBatch[T] = {
         if (prev.hasNext) {
           var i = 0
           while (i < batchSize && prev.hasNext) {
             val sample = prev.next()
-            val featureLength = sample.getFeature().nElement()
-            val labelLength = sample.getLabel().nElement()
+            val featureLength = sample.feature().nElement()
```
yiheng: As we discussed, there's a bug here. Please fix and add unit test

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Types.scala
```diff
@@ -74,7 +77,8 @@ case class LocalSeqFilePath(val path: Path)
 class Sample[T: ClassTag] (
     protected var featureTensor: Tensor[T],
     protected var labelTensor: Tensor[T])
-    (implicit ev: TensorNumeric[T]) {
+    (implicit ev: TensorNumeric[T])
+ extends Serializable {
 
   def this()(implicit ev: TensorNumeric[T]) = this(Tensor[T](), Tensor[T]())
```
yiheng: change the method name to set, as it actually do a set instead of copy

yiheng: And add some java doc

dl/src/main/scala/com/intel/analytics/bigdl/optim/LocalOptimizer.scala
```diff
@@ -199,18 +198,19 @@ class LocalOptimizer[T: ClassTag] private[optim](
...
       val parallelism = if (stackSize == 0) extraSize else subModelNumber
       val start = System.nanoTime()
       val result = Engine.default.invokeAndWait(
         (0 until parallelism).map(b =>
           () => {
             val offset = b * stackSize + math.min(b, extraSize)
             val length = stackSize + (if (b < extraSize) 1 else 0)
-            val input = batch.data.narrow(1, offset + 1, length)
-            val target = batch.labels.narrow(1, offset + 1, length)
+            val (input, target) = batch.narrow(1, offset + 1, length) match {
```
yiheng: please refactor this and all the above

dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -0,0 +1,95 @@
...
+import com.intel.analytics.bigdl.nn.NarrowTable
+import com.intel.analytics.bigdl.nn.abstractnn.Activity
+import com.intel.analytics.bigdl.tensor.Tensor
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+import com.intel.analytics.bigdl.utils.Table
+
+import scala.reflect.ClassTag
+
+/**
+ * A batch of data feed into the model. The first size is batchsize
```
jason-dai: What do you mean by "The first size"?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * @param data
+ * @param labels
+ * @tparam T
+ */
+case class MiniBatch[T: ClassTag](data: Activity, labels: Activity)
+  (implicit ev: TensorNumeric[T]) {
+
+  val _dataLength = dataLength()
+  val _labelLength = labelLength()
```
jason-dai: This adds two Int for each batch, which can lead to large overheads. Why do we need these two variables?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -0,0 +1,95 @@
...
+import scala.reflect.ClassTag
+
+/**
+ * A batch of data feed into the model. The first size is batchsize
+ *
+ * @param data
+ * @param labels
+ * @tparam T
+ */
+case class MiniBatch[T: ClassTag](data: Activity, labels: Activity)
```
jason-dai: Does this need to be Serializable?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -0,0 +1,95 @@
...
+
+  def unapply[@specialized(Float, Double) T: ClassTag]
+  (batch: MiniBatch[T])
+  (implicit ev: TensorNumeric[T]): Option[(Activity, Activity)] =
+    Some((batch.data, batch.labels))
+
+  private def dataLength(): Int = {
+    data match {
+      case tensor: Tensor[T] => tensor.size(1)
+      case table: Table => table.length
```
jason-dai: table.lengh can be costly?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -0,0 +1,95 @@
...
+    data match {
+      case tensor: Tensor[T] =>
+        MiniBatch[T](tensor.narrow(1, index, size), labels.toTensor[T].narrow(dim, index, size))
+      case table: Table =>
+        MiniBatch[T](NarrowTable[T](index, size).updateOutput(table),
+          NarrowTable[T](index, size).updateOutput(labels.toTable))
+      case _ => throw new IllegalArgumentException("MiniBatch only supports Table or Tensor")
+    }
+  }
+}
```
jason-dai: In general, I think it is a bad idea to use pattern match in the performance critical path. Maybe we should define a base MiniBatch class, and then a TensorBatch and a TableBatch class instead.

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -72,63 +73,109 @@ class Identity[A] extends Transformer[A, A] {
...
+              labelArrayOfTensor(i).resizeAs(sample.label()).copy(sample.label())
+              featureTable(i + 1) = featureArrayOfTensor(i)
+              labelTable(i + 1) = labelArrayOfTensor(i)
+              i += 1
             }
-            if (featureData == null || featureData.length < batchSize * featureLength) {
-              featureData = new Array[T](batchSize * featureLength)
+            var j = i + 1
+            while (j <= featureTable.length) {
+              featureTable.remove(j)
```
jason-dai: Table.remove() is very costly; we should introduce something like reset or create a new Table

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -72,63 +73,109 @@ class Identity[A] extends Transformer[A, A] {
...
+              j += 1
             }
-            if (labelData == null || labelData.length < batchSize * labelLength) {
-              labelData = new Array[T](batchSize * labelLength)
+            j = i + 1
+            while (j <= labelTable.length) {
+              labelTable.remove(j)
```
jason-dai: Table.remove() is very costly; we should introduce something like reset or create a new Table

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -72,63 +73,109 @@ class Identity[A] extends Transformer[A, A] {
 
 object SampleToBatch {
   def apply[T: ClassTag]
-  (batchSize: Int)
+  (batchSize: Int,
+   isTable: Boolean = false)
```
jason-dai: Change to toTable

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -72,63 +73,109 @@ class Identity[A] extends Transformer[A, A] {
...
   (implicit ev: TensorNumeric[T]): SampleToBatch[T]
-  = new SampleToBatch[T](batchSize)
+  = new SampleToBatch[T](batchSize, isTable)
 }
 
 class SampleToBatch[T: ClassTag]
-  (totalBatch: Int)
+  (totalBatch: Int, isTable: Boolean = false)
```
jason-dai: Change to toTable

dl/src/main/scala/com/intel/analytics/bigdl/models/autoencoder/Train.scala
```diff
@@ -24,19 +24,23 @@ import com.intel.analytics.bigdl.dataset.image._
...
   def apply(): toAutoencoderBatch[Float] = new toAutoencoderBatch[Float]()
 }
 
-class toAutoencoderBatch[T] extends Transformer[MiniBatch[T], MiniBatch[T]] {
+class toAutoencoderBatch[T: ClassTag]
+(implicit ev: TensorNumeric[T]) extends Transformer[MiniBatch[T], MiniBatch[T]] {
   override def apply(prev: Iterator[MiniBatch[T]]): Iterator[MiniBatch[T]] = {
     prev.map(batch => {
-      MiniBatch(batch.data, batch.data)
+      MiniBatch[T](batch.data, batch.data)
```
jason-dai: This is bad; Why do we need [T] here?

dl/src/main/scala/com/intel/analytics/bigdl/models/utils/DistriOptimizerPerf.scala
```diff
@@ -117,7 +117,7 @@ object DistriOptimizerPerf {
     val labels = Tensor(param.batchSize).fill(1)
 
     val sc = new SparkContext(conf)
-    val broadcast = sc.broadcast(MiniBatch(input, labels))
+    val broadcast = sc.broadcast(MiniBatch[Float](input, labels))
```
jason-dai: This is bad; Why do we need [T] here?

jason-dai:
Please run performance tests for the changes. And add more scala doc and comments (esp. for public methods).

dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
JerryYanWan: It means the first dimension of the size. I rewrite it to be clearer.

dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
JerryYanWan: We need to check that the dataLength and labelLength are the same. How about calculate labelLength and datalength in the constructor and only save one of them?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
JerryYanWan: Added

dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
JerryYanWan: only calculate once for each batch, using other methods?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
JerryYanWan: If using two classes which extend from an abstract class, the interfaces in the optimizer will be different for tensor and table input.

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
JerryYanWan: changed

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
JerryYanWan: changed

dl/src/main/scala/com/intel/analytics/bigdl/models/autoencoder/Train.scala
JerryYanWan: Inside MiniBatch class, there is a NarrowTable being used to implement a narrow method for table. Thus, the MiniBatch requires a type.

dl/src/main/scala/com/intel/analytics/bigdl/models/utils/DistriOptimizerPerf.scala
JerryYanWan: NarrowTable layer being used.

PR #363, opened 2017-01-10
Mkl Dnn Supports (WorkInProgress)

i8run:


dl/src/main/scala/com/intel/analytics/bigdl/nn/dnn/Affinity.scala
```diff
@@ -0,0 +1,91 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
```
fanshiqing: License

dl/src/main/scala/com/intel/analytics/bigdl/nn/dnn/BatchNormalization.scala
```diff
@@ -0,0 +1,279 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
```
fanshiqing: License ASF => intel corporation

dl/src/main/scala/com/intel/analytics/bigdl/nn/dnn/Linear.scala
```diff
@@ -0,0 +1,391 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
```
fanshiqing: License.
The same as below.

PR #378, opened 2017-01-16
gradient check

zhangxiaoli73:


dl/src/test/scala/com/intel/analytics/bigdl/models/GoogleNet.scala
```diff
@@ -0,0 +1,452 @@
...
+
+import com.intel.analytics.bigdl._
+import com.intel.analytics.bigdl.nn._
+import com.intel.analytics.bigdl.numeric.NumericDouble
+import com.intel.analytics.bigdl.utils.{T, Table}
+
+/**
+ * no Dropout
+ */
+object GoogleNet_v1 {
```
yiheng: can we rename the model names in the files to point out they're models for gradient check test?

dl/src/test/scala/com/intel/analytics/bigdl/models/VggLike.scala
```diff
@@ -0,0 +1,114 @@
...
+ * limitations under the License.
+ */
+
+package com.intel.analytics.bigdl.models
+
+import com.intel.analytics.bigdl._
+import com.intel.analytics.bigdl.nn.{LogSoftMax, _}
+import com.intel.analytics.bigdl.numeric.NumericDouble
+
+object VggLike {
```
yiheng: Can we rename the models in this file to point out they're only for gradient check

yiheng:
I can see there're 3 commits but from the final changed files there're only one. Do you remove the changes in the other two commits? If so, you should do a rebase to merge your commits.

dl/src/main/scala/com/intel/analytics/bigdl/nn/BatchNormalization.scala
```diff
@@ -51,7 +49,9 @@ class BatchNormalization[@specialized(Float, Double) T: ClassTag](
   val gradBias: Tensor[T] = if (affine) Tensor[T](nOutput) else null
 
   @transient
-  private var results: Array[Future[_]] = null
+  private var results : Array[Future[_]] = null
+  @transient
+  private var needInit: Boolean = false
```
yiheng: please add a comment to mention why we need this

dl/src/test/scala/com/intel/analytics/bigdl/nn/GradientChecker.scala
```diff
@@ -22,38 +22,128 @@ import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
...
 
-class GradientChecker(stepSize: Double, threshold: Double) {
+class GradientChecker(stepSize: Double, threshold: Double = 1e-2) {
+
+  private val maxPrNum = 50
+  private var checkModel = false
+
+  // add status for checking model, default or 0 means check maxPrNum points,
+  // 1 means check all points, others mean the checked points number
+  private val status = if ((null != System.getProperty("modelCheck")) &&
```
yiheng: case class is better than number

yiheng:
Another question is in our jenkins job, as we don't set the modelCheck property, we always check 50 sample points?

dl/src/test/scala/com/intel/analytics/bigdl/models/GoogleNet.scala
zhangxiaoli73: ok, changed

dl/src/test/scala/com/intel/analytics/bigdl/models/VggLike.scala
zhangxiaoli73: ok, changed. add "_test" to the end and add some comments

dl/src/main/scala/com/intel/analytics/bigdl/nn/BatchNormalization.scala
zhangxiaoli73: ok, changed

dl/src/test/scala/com/intel/analytics/bigdl/nn/GradientChecker.scala
zhangxiaoli73: good idea. Changed

zhangxiaoli73:
yes, the default check point number is 50.

yiheng:
Jenkins job pass, merge

PR #381, opened 2017-01-16
fix loadcaffe cannot find method in spark2.0

SeaOfOcean:
by changing the protobuf version from 3.0 to 2.5
#369


yiheng:
jenkins job passed. merge the PR

PR #385, opened 2017-01-17
Add necessary comments for RNG class

fanshiqing:


dl/src/main/scala/com/intel/analytics/bigdl/utils/RandomGenerator.scala
```diff
@@ -183,6 +183,9 @@ class RandomGenerator() {
     this
   }
 
+  /**
```
yiheng: It can't pass the format check. Use java-doc style comment

linbojin: Do you use scalafmt: https://github.com/olafurpg/scalafmt A good format tool.

yiheng: @linbojin We just check the style in the build. To do code format, I personally use IDE reformat function...

yiheng:
Jenkins test pass. Merge the code.

dl/src/main/scala/com/intel/analytics/bigdl/utils/RandomGenerator.scala
linbojin: Yes, same to me, scalafmt is already integrated inside latest intellij idea 😄

Issue #395, opened 2017-01-20
Optimizer.optimize need to update the input model when it's done


jason-dai:
The user provides an input model to Optimizer(model, dataset, criterion), and Optimizer.optimize() will return a trained model after it's done:
 1. The local optimizer actually uses the weight of the input model in training
 2. The distributed optimizer will not update the weight of the input model
 3. None of them update other states of the input model (e.g., those in BatchNormalization)

This is very confusing; Optimizer.optimize() should update the input model after it is done, so that the user can use it for prediction.

PR #404, opened 2017-01-22
issue[#395]Optimizer.optimize update and return the input model

qiuxin2012:
#395 
Add two functions to nn.Utils, copyModule and recursiveCopyState.
Use copyModule at the end of DistriOptimizer.optimize to copy the trained Module's parameters and running state to `model`.
Use recursiveCopyState at the end of LocalOptimizer.optimize to copy the `workingModels`'s running state to the `model`, and parameters is already shared between `workingModels` and `model`.

### DIFF
--- a/dl/src/main/scala/com/intel/analytics/bigdl/nn/Utils.scala
+++ b/dl/src/main/scala/com/intel/analytics/bigdl/nn/Utils.scala
@@ -218,4 +218,21 @@ object Utils {
     getModules(model)
     namedModules
   }
+
+  /**
+   * copy src's parameters and running status to dst
+   * @param src source model
+   * @param dst destination model
+   */
+  def copyModule[T](src: Module[T], dst: Module[T]): Module[T] = {
+    // copy parameters
+    val srcParameters = src.getParameters()._1
+    val dstParameters = dst.getParameters()._1
+    require(srcParameters.size(1) == dstParameters.size(1),
+      s"$src and $dst is not the same type.")
+    dstParameters.copy(srcParameters)
+    // copy running status
+    dst.copyStatus(src)
+    dst
+  }
 }
--- a/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
+++ b/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
@@ -468,10 +468,9 @@ object DistriOptimizer {
     })
   }
 
-
-  private def getModel[T: ClassTag](models: RDD[Cache[T]],
-      parameters: AllReduceParameter[T])
-  : Module[T] = {
+  private def getModel[T: ClassTag](
+      models: RDD[Cache[T]],
+      parameters: AllReduceParameter[T]): Module[T] = {
     val partitionNum = models.partitions.length
     val trainedModel = models.map(_.localModels.head.clearState()).first()
     val weights = models.mapPartitions(iter => {
@@ -546,7 +545,10 @@ class DistriOptimizer[T: ClassTag] private[optim](
       isOverWrite
     )
 
-    DistriOptimizer.getModel(models, parameters)
+    val trainedModel = DistriOptimizer.getModel(models, parameters)
+
+    nn.Utils.copyModule(trainedModel, model)
+    model
   }
 }
 
--- a/dl/src/main/scala/com/intel/analytics/bigdl/nn/BatchNormalization.scala
+++ b/dl/src/main/scala/com/intel/analytics/bigdl/nn/BatchNormalization.scala
@@ -17,6 +17,7 @@
 
 package com.intel.analytics.bigdl.nn
 
+import com.intel.analytics.bigdl.Module
 import com.intel.analytics.bigdl.nn.abstractnn.TensorModule
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
 import com.intel.analytics.bigdl.tensor.{DoubleType, FloatType, Tensor}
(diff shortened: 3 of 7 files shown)


dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -461,10 +461,10 @@ object DistriOptimizer {
...
   }
 
-
-  private def getModel[T: ClassTag](models: RDD[Cache[T]],
-      parameters: AllReduceParameter[T])
-  : Module[T] = {
+  private def getModel[T: ClassTag](
+      models: RDD[Cache[T]],
+      parameters: AllReduceParameter[T],
+      result: Module[T] = null): Module[T] = {
```
jason-dai: There is no need to change the interface; you can unconditionally copy the model you get in this method to the user provided model (Optimizer.model), either at the end of this method or at the end of the optimize() method.

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -485,7 +485,17 @@ object DistriOptimizer {
...
+      // Copy trainedModel to result
+      val resultParameter = result.getParameters()._1
+      require(resultParameter.nElement() == parameter.nElement(),
+        "result's parameter should have the same size with train model's parameter")
+      resultParameter.copy(parameter)
+      nn.Utils.recursiveCopyState(trainedModel, result)
+      result
+    } else {
+      trainedModel
+    }
```
jason-dai: You can just return trainedModel unconditionally here

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -485,7 +485,17 @@ object DistriOptimizer {
       parameter.narrow(1, start + 1, length).copy(weights(pid))
     })
 
-    trainedModel
+    if (null != result) {
+      // Copy trainedModel to result
+      val resultParameter = result.getParameters()._1
+      require(resultParameter.nElement() == parameter.nElement(),
+        "result's parameter should have the same size with train model's parameter")
+      resultParameter.copy(parameter)
```
jason-dai: resultParameter.set(parameter)

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -485,7 +485,17 @@ object DistriOptimizer {
...
+      nn.Utils.recursiveCopyState(trainedModel, result)
```
jason-dai: I think you can define a recursiveCopyModule method, which in turn takes care of both the parameter setting and state copy.

dl/src/main/scala/com/intel/analytics/bigdl/nn/Utils.scala
```diff
@@ -218,4 +218,32 @@ object Utils {
...
+        val c2 = result.asInstanceOf[Container[Activity, Activity, T]]
+        require(c1.modules.length == c2.modules.length)
+        var i = 0
+        while (i < c1.modules.length) {
+          recursiveCopyState(c1.modules(i), c2.modules(i))
+          i += 1
+        }
+      }
+    }
+  }
```
jason-dai: Why not
```scala
if (model.isInstanceOf[Container[Activity, Activity, T]]) {
   //report error if result is not a container
   ...
}
else if (model.isInstanceOf[BatchNormalization[T]]) {
   //report error if result is not a BatchNormalization
   ...
}
```

jason-dai:
You should copy state for the local optimizer too

dl/src/main/scala/com/intel/analytics/bigdl/nn/Utils.scala
```diff
@@ -218,4 +218,51 @@ object Utils {
...
+      val b1 = src.asInstanceOf[BatchNormalization[T]]
+      val b2 = dst.asInstanceOf[BatchNormalization[T]]
+      b2.runningMean.copy(b1.runningMean)
+      b2.runningVar.copy(b1.runningVar)
+    } else if (src.isInstanceOf[Container[Activity, Activity, T]]) {
+      require(dst.isInstanceOf[Container[Activity, Activity, T]],
+        s"$dst must be the same type with $src")
+      val c1 = src.asInstanceOf[Container[Activity, Activity, T]]
+      val c2 = dst.asInstanceOf[Container[Activity, Activity, T]]
+      require(c1.modules.length == c2.modules.length)
```
yiheng: require message?

dl/src/main/scala/com/intel/analytics/bigdl/nn/Utils.scala
```diff
@@ -218,4 +218,51 @@ object Utils {
...
+   */
+  def copyModule[T](src: Module[T], dst: Module[T]): Module[T] = {
+    // copy parameters
+    val srcParameters = src.getParameters()._1
+    val dstParameters = dst.getParameters()._1
+    require(srcParameters.size(1) == dstParameters.size(1),
+      s"$src and $dst is not the same type.")
+    dstParameters.copy(srcParameters)
+    // copy state
+    if (src.isInstanceOf[Container[Activity, Activity, T]] ||
```
yiheng: do we still need the if here?

dl/src/main/scala/com/intel/analytics/bigdl/nn/Utils.scala
```diff
@@ -218,4 +218,51 @@ object Utils {
...
+    }
+    dst
+  }
+
+  /**
+   * copy runningMean and runningVar of BatchNormalization into result recursively
+   * @param src
+   * @param dst
+   */
+  def recursiveCopyState[T](src: Module[T], dst: Module[T]): Unit = {
```
yiheng: Can we move this into module memeber?

yiheng:
LGTM

qiuxin2012:
Some update.
Add a module's method `copyStatus()` to copy module's status instead of `recursiveCopyState()`.

PR #406, opened 2017-01-22
fix #396: support save/load model from hdfs

### LINKED ISSUE #396: Support save/load model from HDFS.

fanshiqing:


dl/src/test/scala/com/intel/analytics/bigdl/optim/OptimizerSpec.scala
```diff
@@ -122,7 +123,28 @@ class OptimizerSpec extends FlatSpec with Matchers {
     loadedModel should be(model)
   }
 
-  it should "save model and state to given path with postfix" in {
+  it should "save model to given hdfs path without postfix" in {
+    val conf: Configuration = new Configuration()
+    val dfs = conf.get("fs.defaultFS")
+    if (dfs != null && dfs.startsWith(File.hdfsPrefix)) {
```
fanshiqing: if `conf.get("fs.defaultFS")` doesn't return expected `hdfs` prefix, this UT will not actually make sense.
Is there a more friendly way to design this UT?

dl/src/main/scala/com/intel/analytics/bigdl/utils/File.scala
```diff
@@ -49,28 +57,84 @@ object File {
...
-      throw new RuntimeException("file exists!")
+  def save(obj: Serializable, outputPath: String, isOverwrite: Boolean = false): Unit = {
+    var localFilePath: String = outputPath
+    if (outputPath.startsWith(File.localfsPrefix)) {
+      localFilePath = outputPath.substring(File.localfsPrefix.length)
+      if (Files.exists(Paths.get(outputPath)) && !isOverwrite) {
+        throw new RuntimeException(s"file ${outputPath} already exists!")
+      }
+    } else {
+      localFilePath = System.getProperty("user.dir") + "/output-model.tmp"
```
fanshiqing: `System.getProperty("user.dir")` gets the directory where `java` was run from, so this will always works with any env with java installed. Right?

dl/src/main/scala/com/intel/analytics/bigdl/optim/Optimizer.scala
```diff
@@ -61,8 +62,15 @@ abstract class Optimizer[T: ClassTag, D](
   }
 
   def setCheckpoint(path: String, trigger: Trigger): this.type = {
-    require(Files.isDirectory(Paths.get(path)), s"$path is not a folder")
-    this.checkpointPath = Some(path)
+    if (path.startsWith("/") || path.startsWith("file:")) {
```
qiuxin2012: We need to support three kinds of path:
For the path `apath` --> [hdfs://]user/username/apath
For the path `/apath` --> [hdfs://]apath
For the path `file:///apath` --> 'file:///apath'

dl/src/main/scala/com/intel/analytics/bigdl/utils/File.scala
```diff
@@ -49,28 +57,84 @@ object File {
...
+  }
+
+  /**
+   * Move file from local to HDFS.
+   *
+   * @param localFilePath
+   * @param hdfsPath
+   * @param overwrite
+   */
+  def moveFile2HDFS(localFilePath: String, hdfsPath: String, overwrite: Boolean): Unit = {
```
qiuxin2012: Can you try to save/load file to/from HDFS directly?

PR #408, opened 2017-01-23
[Issue394]Add better documents for Sum, Mean, JoinTable and Padding

psyyz10:


dl/src/main/scala/com/intel/analytics/bigdl/nn/JoinTable.scala
```diff
@@ -27,6 +27,11 @@ import scala.reflect.ClassTag
  * It is a table module which takes a table of Tensors as input and
  * outputs a Tensor by joining them together along the dimension `dimension`.
  *
+ * Note: the input tensor can have indefinite number of dimensions,
+ * so modules can not automatically inference whether the user use the batch or not
+ * and what the actual dimensions of input excluding the batch dimension.
+ * Hence `nInputDims` represents the actual dimensions of given input excluding the batch.
+ *
```
jason-dai: This is somewhat confusing. Just say somehting like "the input to this layer is expected to a tensor, or a batch of tensors; when using mini-batch, a batch of tensors will be passed to the layer and the user need to specify the number of dimensions of each input tensor in the batch using nInputDims"

dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -228,6 +228,19 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
...
+   * otherwise operation will base on no-batch assumption (with pure online stochastic
+   * gradient descent)
+   */
+  protected var batch: Boolean = false
+
+  def isBatchMode: Boolean = batch
+
+  def setBatchMode(useBatch: Boolean): Unit =
+    batch = useBatch
+
```
jason-dai: I don't think we need this for now.

psyyz10: Removed.

dl/src/main/scala/com/intel/analytics/bigdl/nn/JoinTable.scala
psyyz10: Yes, I agree. Have changed, please check

jason-dai: You may want to change the scala doc for the `nInputDims` parameter accordingly too.

yiheng:
Please merge these commits into one before you push to the repo

Issue #397, opened 2017-01-20
Loss (subclass of ValidationMethod) should have the user to specify which Criterion to use


jason-dai:
Currently it use a hard-coded CrossEntropyCriterion

PR #411, opened 2017-01-24
[#397]add criterion for Loss

zhangxiaoli73:


dl/src/main/scala/com/intel/analytics/bigdl/models/rnn/Train.scala
```diff
@@ -85,17 +85,18 @@ object Train {
       }
 
       Engine.init(1, param.coreNumber, false)
+      val criterion = new CrossEntropyCriterion[Float](squeezeFlag = true)
```
yiheng: Are you sure the criterion can be shared in optimization and validation?

JerryYanWan: Pass this criterion to Loss validation method. Inside the validation method, it will clone it when the apply is called.

jason-dai: I don't think we should change the Loss interface at this time. Ideally the optimizer should provide the workingCriterion(i) for the validation method to use; we can leave that to future improvement.

dl/src/main/scala/com/intel/analytics/bigdl/nn/ClassNLLCriterion.scala
```diff
@@ -100,12 +109,17 @@ class ClassNLLCriterion[T: ClassTag](weights: Tensor[T] = null, sizeAverage: Boo
       if (resultsBackward == null || resultsBackward.length != batchSize) {
         resultsBackward = new Array[Future[_]](batchSize)
       }
+      if (target.dim() > 1) {
```
qiuxin2012: Delete this

dl/src/main/scala/com/intel/analytics/bigdl/nn/ClassNLLCriterion.scala
```diff
@@ -53,12 +55,19 @@ class ClassNLLCriterion[T: ClassTag](weights: Tensor[T] = null, sizeAverage: Boo
       if (results == null || results.length != batchSize) {
         results = new Array[Future[(T, T)]](batchSize)
       }
+      if (null == targetBuffer) targetBuffer = Tensor[T]()
```
qiuxin2012: save the size of target, resize before return the output

jason-dai:
Looks good to me. @zhangxiaoli73 @JerryYanWan please merge the code if testing results (including examples from user group and the RNN model) are OK.

PR #412, opened 2017-01-24
add clear() method to Table

JerryYanWan:


dl/src/main/scala/com/intel/analytics/bigdl/utils/Table.scala
```diff
@@ -42,6 +42,11 @@ class Table private[bigdl](
     }
   }
 
+  def clear(): Unit = {
```
qiuxin2012: return itself

dl/src/main/scala/com/intel/analytics/bigdl/utils/Table.scala
```diff
@@ -42,6 +42,12 @@ class Table private[bigdl](
     }
   }
 
+  def clear(): this.type = {
```
yiheng: please add java doc for the new added public method

yiheng:
Please merge commits into one commit

JerryYanWan:
added

PR #416, opened 2017-01-25
print callstack of exception in forward+backward

dding3:


zhichao-li:
looks good. other alternatives may be usingThread.UncaughtExceptionHandler

PR #422, opened 2017-01-27
support user disable singletoncheck

dding3:


spark-version/1.5-plus/src/main/scala/ZippedPartitionsWithLocalityRDD.scala
```diff
@@ -62,7 +62,7 @@ class ZippedPartitionsWithLocalityRDD[A: ClassTag, B: ClassTag, V: ClassTag](
       while (p < candidateLocs.length) {
         locs = candidateLocs(p)._2.intersect(curPrefs)
         if (!locs.isEmpty) {
-          matchPartition = candidateLocs.remove(p)
+          matchPartition = candidateLocs.apply(p)
```
jason-dai: You cannot do this; the logic should be:
1) If partition0 (from rdd(0)) has the same location as partition1 (from rdd(1)), we'll zip them together and remove partition1 from the candidate list
2) If no partition in rdd(1) has the same location as the partition0 from rdd(0), we should randomly select remove one partition from the candidate list and zip it with partition0

jason-dai:
I think we should do something like:
1) Provide a config or flag (e.g., "bigdl.check.singleton"), and init DistriOptimizer.checkSingleton based on that value
2) In DistriOptimizer.initThreadModels, always call Engine.checkSingleton(); however, if DistriOptimizer.checkSingleton is false, only report a warning (instead of an error)
3) In ZippedPartitionsWithLocalityRDD, try to zip partitions with the common locations first; for remaining partitions that don't have common locations, zip them in a random fashion.

PR #423, opened 2017-01-30
bug fix

dding3:
What changes were proposed in this pull request?
1. Support disable singleton check. User can disable singleton check by setting "bigdl.check.singleton" as "false".
2. Zip matched partitions of rdds in ZippartitionWithLocality, then zip left unmatched partitions in order
3. Print callstack of exception from syncPool in Allreduceparameter

How was this patch tested?
UT and manual test

### DIFF
--- a/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
+++ b/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
@@ -362,11 +362,13 @@ object DistriOptimizer {
     val comupteThresholdbatchSize = state.get[Int]("comupteThresholdbatchSize").get
     val models = dataset.originRDD().mapPartitions(_ => {
       val (broadcastModel, broadcastCriterion, broadcastState) = broadcast.value
-      if (checkSingleton) {
-        require(Engine.checkSingleton(), "Detect multi-task run on one Executor/Container. " +
-          "Currently not support this")
-      } else {
-        logger.info("Skip singleton check")
+      if (!Engine.checkSingleton()) {
+        if (checkSingleton) {
+          require(Engine.checkSingleton(), "Detect multi-task run on one Executor/Container. " +
+            "Please disable singleton check or repartition data")
+        } else {
+          logger.warn("Detect multi-task run on one Executor/Container.")
+        }
       }
       val cached = (0 until _subModelNumber).map { _ =>
         val localModel = broadcastModel.cloneModule()
@@ -502,6 +504,9 @@ class DistriOptimizer[T: ClassTag] private[optim](
   val metrics = new Metrics
 
   private var models: RDD[DistriOptimizer.Cache[T]] = null
+
+  private val checkSingleton = System.getProperty("bigdl.check.singleton",
+    true.toString).toBoolean
 
   override def optimize(): Module[T] = {
     this.assertEngineInited()
--- a/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
+++ b/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
@@ -505,9 +505,6 @@ class DistriOptimizer[T: ClassTag] private[optim](
 
   private var models: RDD[DistriOptimizer.Cache[T]] = null
 
-  private val checkSingleton = System.getProperty("bigdl.check.singleton",
-    true.toString).toBoolean
-
   override def optimize(): Module[T] = {
     this.assertEngineInited()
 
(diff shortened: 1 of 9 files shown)


dl/src/main/scala/com/intel/analytics/bigdl/optim/Optimizer.scala
```diff
@@ -49,13 +49,6 @@ abstract class Optimizer[T: ClassTag, D](
   protected var comupteThresholdbatchSize: Int = 100
   protected var warmupIterationNum: Int = 200
 
-  def disableCheckSingleton(): this.type = {
-    this.checkSingleton = false
-    this
-  }
-
-  protected var checkSingleton = true
-
```
jason-dai: Instead of removing `disableCheckSingleton`, maybe we should mark it as deprecated, and report a warning to ask users to use the "bigdl.check.singleton" config instead.

linbojin:
You may give more detailed pr description as this is an open source project. Some one may refer to your pr in the future. 😄

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -362,11 +362,13 @@ object DistriOptimizer {
...
     val models = dataset.originRDD().mapPartitions(_ => {
       val (broadcastModel, broadcastCriterion, broadcastState) = broadcast.value
-      if (checkSingleton) {
-        require(Engine.checkSingleton(), "Detect multi-task run on one Executor/Container. " +
-          "Currently not support this")
-      } else {
-        logger.info("Skip singleton check")
+      if (!Engine.checkSingleton()) {
+        if (checkSingleton) {
+          require(Engine.checkSingleton(), "Detect multi-task run on one Executor/Container. " +
```
jason-dai: Don't checkSingleton twice - the second check is guaranteed to fail

dl/src/main/scala/com/intel/analytics/bigdl/parameters/AllReduceParameter.scala
```diff
@@ -136,7 +136,9 @@ class AllReduceParameter[T: ClassTag](id: Long, partitionNum: Int,
         override def call(): Int = {
           val blockId = getWeightBlockId(pid)
           val localBuffer = BlockManagerWrapper.byteBufferConvert(
-            bm.getLocalBytes(blockId).getOrElse(bm.getRemoteBytes(blockId).get))
+            bm.getLocalBytes(blockId).getOrElse(bm.getRemoteBytes(blockId)
+                .getOrElse(throw new IllegalArgumentException(s"Can't get the block(${blockId})." +
+                s"Please check if there is executor lost"))))
```
jason-dai: Why IllegalArgumentException? And shouldn't we try-catch for all the operations in call(), just as we did for ThreadPool.invokeAndWait?

spark-version/1.5-plus/src/main/scala/ZippedPartitionsWithLocalityRDD.scala
```diff
@@ -62,15 +62,17 @@ class ZippedPartitionsWithLocalityRDD[A: ClassTag, B: ClassTag, V: ClassTag](
...
       }
-      require(matchPartition != null, s"can't find locality partition for partition $i " +
-        s"Partition locations are (${curPrefs}) Candidate partition locations are\n" +
-        s"${candidateLocs.mkString("\n")}. Are you using more executors than the node number? " +
-        s"If yes, try to change your executor number.")
+      if (matchPartition == null) {
+        println(s"can't find locality partition for partition $i " +
+          s"Partition locations are (${curPrefs}) Candidate partition locations are\n" +
+          s"${candidateLocs.mkString("\n")}.")
+        matchPartition = candidateLocs.remove(util.Random.nextInt(candidateLocs.length))
```
jason-dai: While this would work, it may randomly pick a partition (from candidateLocs) that has common location with a (yet to be processed) partition from rdd(0); a better solution is:
1) First zip partitions with common locations as much as we can;
2) After we have zipped all the partitions with common locations, we can then process the remaining partitions (which have no common locations with others); we can simply zip them together in order (just as ZippedPartitionsBaseRDD does).

dl/src/main/scala/com/intel/analytics/bigdl/optim/Optimizer.scala
dding3: OK

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
dding3: If there are more than one task run in an executor and checkSingleton is enable, we need throw exception here, the second checkSingleton is guaranteed to fail to trigger the exception

dl/src/main/scala/com/intel/analytics/bigdl/parameters/AllReduceParameter.scala
dding3: OK.

PR #427, opened 2017-02-03
modify batchsize require to warning

zhangxiaoli73:
requiring totalBatch >= nodeNumber * coreNumber * 2 is too strict, so modify to warning

### DIFF
--- a/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Utils.scala
+++ b/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Utils.scala
@@ -18,8 +18,10 @@
 package com.intel.analytics.bigdl.dataset
 
 import com.intel.analytics.bigdl.utils.Engine
+import org.apache.log4j.Logger
 
 object Utils {
+  private val logger = Logger.getLogger(getClass)
 
   def getBatchSize(totalBatch : Int): Int = {
     if (Engine.nodeNumber().isDefined) {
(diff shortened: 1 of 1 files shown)


dl/src/main/scala/com/intel/analytics/bigdl/dataset/Utils.scala
```diff
@@ -28,9 +30,12 @@ object Utils {
       require(totalBatch % (nodeNumber * coreNumber) == 0
         , s"total batch size($totalBatch) can't be divided by node number($nodeNumber) * " +
           s"core number($coreNumber), please change your batch size")
-      require(totalBatch >= nodeNumber * coreNumber * 2
-        , s"total batch size($totalBatch) should be at least two times of node number" +
-          s"($nodeNumber) * core number($coreNumber), please change your batch size")
+
+      if (totalBatch < nodeNumber * coreNumber * 2) {
+        logger.info(s"Warning: for better property, " +
```
yiheng: logger.warning

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Utils.scala
```diff
@@ -28,9 +30,12 @@ object Utils {
...
+          s"total batch size($totalBatch) should be at least two times of node number" +
+          s"($nodeNumber) * core number($coreNumber), please optimize your batch size")
+      }
```
jason-dai: Change to "... for better training speed, ... is recommended to be at least ... please tune your batch size accordingly"

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Utils.scala
zhangxiaoli73: ok, thanks. actually logger.warn

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Utils.scala
zhangxiaoli73: ok, changed

yiheng:
Looks good to me. Please merge it if jenkins job pass.

jason-dai:
LGTM

PR #429, opened 2017-02-04
Accept sample rdd as the data input and do the batching atomatically

zhichao-li:
In consideration of the performance, SampleToBatch should be limited to only accept contiguous tensor.

### DIFF
--- a/dl/src/main/scala/com/intel/analytics/bigdl/optim/Optimizer.scala
+++ b/dl/src/main/scala/com/intel/analytics/bigdl/optim/Optimizer.scala
@@ -22,10 +22,12 @@ import java.nio.file.{Files, Paths}
 import com.intel.analytics.bigdl._
 import com.intel.analytics.bigdl.utils.{Engine, T, Table}
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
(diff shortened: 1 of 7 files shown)


dl/src/main/scala/com/intel/analytics/bigdl/optim/Optimizer.scala
```diff
@@ -71,6 +73,18 @@ abstract class Optimizer[T: ClassTag, D](
     this
   }
 
+  def setValidation(trigger: Trigger, sampleRDD: RDD[Sample[T]],
+      vMethods : Array[ValidationMethod[T]], batchSize: Int)
+  : this.type = {
```
jason-dai: Change to:

val dataSet = DataSet.rdd(sampleRDD) -> SampleToBatch(batchSize)
setValidation(trigger, dataSet, vMethods)

jason-dai: And what if this is called for LocalOptimizer? I think we need to override the setValidation methods for both LocalOptimizer and DistriOptimizer, which need to check if the validation dataset is LocalDataSet or DistriDataSet respectively.

PR #431, opened 2017-02-04
fix next null exception for unit test

zhangxiaoli73:
when do parallel for an iterator, iterator.next() may be null, so add some judgment.


zhangxiaoli73:
jenkins pass

PR #435, opened 2017-02-06
recurrent supports batchSize greater than one

JerryYanWan:


dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -26,115 +27,94 @@ import scala.reflect.ClassTag
...
-      val curInput = T(input(i), hidden(i).asInstanceOf[Tensor[T]])
+    while (i <= times) {
+      val curInput = T(input.select(2, i), hidden.select(2, i))
       val currentOutput = module.updateOutput(curInput)
       transform.updateOutput(currentOutput)
-      output.update(i, transform.output.asInstanceOf[Tensor[T]])
-      hidden(i + 1) = Tensor[T]()
-        .resizeAs(transform.output.asInstanceOf[Tensor[T]])
-        .copy(transform.output.asInstanceOf[Tensor[T]])
+      output.select(2, i).copy(transform.output.asInstanceOf[Tensor[T]])
```
psyyz10: Maybe `.toTensor` insteadof `.asInstanceOf[Tensor[T]])`

jason-dai:
LGTM

JerryYanWan:
#435 Now the Recurrent layer can accept batch input and do the recurrence. The input size would be 3D: [batch, times, nDim]. The layer will recurrent along the times dimension. The performance of this layer is still needed to be optimized later because the input and hidden Tensor would be discrete for RNNCell by using Tensor.select method. @jason-dai @yiheng

jason-dai:
Please merge the fix.

Issue #271, opened 2016-12-28
counting of epoch & trained size in resume training does not match original counts


shane-huang:
By using checkpoint, the models/states are saved into disk periodically, and user can resume training from certain iteration (checkpoint).
 
However, in current logs, the same iteration maps to different epoch/trained size before and after resuming from certain checkpoint. This is confusing and could bring problems for people who need to combine and analyze the logs together. 

Problem illustrated below. Resumed training from Iteration no. 17149. 
This is the log before pause. The epoch batch counter for iteration 17149 is **[Epoch 13 0/1279867]**
```
INFO  DistriOptimizer$:230 - [Epoch 13 0/1279867][Iteration 17149][Wall Clock xxxs] Train xxx in xxx seconds. Throughput is xxx records/second. Loss is xxx.
```
This is the log after pause. The epoch batch counter for iteration 17149 is **[Epoch 12 896/1279867]**
```
INFO  DistriOptimizer$:230 - [Epoch 12 896/1279867][Iteration 17149][Wall Clock xxxs] Train xxx in xxx seconds. Throughput is xxx records/second. Loss is xxx.
```

PR #445, opened 2017-02-08
Fix Issue 271: [counting of epoch & trained size in resume training does not mat…

psyyz10:

Fix [counting of epoch & trained size in resume training does not match original counts]

## How was this patch tested?

Train a model and save checkpoint by each epoch. Reload a snapshot, see the epoch and iteration are consistent with the saved checkpoint.

fixed https://github.com/intel-analytics/BigDL/issues/271

### DIFF
--- a/dl/src/test/scala/com/intel/analytics/bigdl/optim/DistriOptimizerSpec.scala
+++ b/dl/src/test/scala/com/intel/analytics/bigdl/optim/DistriOptimizerSpec.scala
@@ -17,11 +17,14 @@
 
 package com.intel.analytics.bigdl.optim
 
-import com.intel.analytics.bigdl.dataset.{MiniBatch, DistributedDataSet}
+import java.io.File
+
+import com.intel.analytics.bigdl.dataset.{DistributedDataSet, MiniBatch}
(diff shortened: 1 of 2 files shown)


yiheng:
can you add an unit test for this fix?

psyyz10:
Done with unit tests

yiheng:
LGTM

dl/src/test/scala/com/intel/analytics/bigdl/optim/DistriOptimizerSpec.scala
```diff
@@ -243,6 +246,30 @@ class DistriOptimizerSpec extends FlatSpec with Matchers with BeforeAndAfter {
     batchNormalization.runningVar.storage().array() should be (
       Array(1188.2811870277535, 1188.2811870277535)
     )
+  }
+
+  "DistriOptimizer checkpoint" should "work correclty" in {
+    val path = "/tmp/model/"
```
yiheng: please do not hard code the path

psyyz10: Done

PR #467, opened 2017-02-15
[faster-rcnn] speedup cadd and cmul for pvanet by using continuous storage

SeaOfOcean:

speedup cadd and cmul for pvanet by using continuous storage

## How was this patch tested?

unit test

### DIFF
--- a/dl/src/main/scala/com/intel/analytics/bigdl/nn/CAdd.scala
+++ b/dl/src/main/scala/com/intel/analytics/bigdl/nn/CAdd.scala
@@ -50,21 +50,74 @@ class CAdd[@specialized(Float, Double) T: ClassTag](
       } else {
         bias.view(Array(1) ++ bias.size())
       }
-      if (bias.size().count(x => x != 1) == 1) {
-        val biasDim = expand.size().zip(Stream.from(1)).filter(x => x._1 != 1)(0)._2
-        var k = 0
-        val biasData = bias.storage().array()
-        while (k < bias.nElement()) {
-          val c = output.select(biasDim, k + 1)
-          c.add(biasData(k))
-          k += 1
-        }
+      val pivotDim = getOnlyDimGtOne(bias.size())
+      if (pivotDim > 1) {
+        addOneDimBias(pivotDim, expand, output)
       } else {
         expand.expandAs(output)
         output.add(expand)
       }
     }
     output
+  }
+
+  /**
+   * if there is only one dim of size > 1, return this dim(count from 1)
+   * else return -1
+   * e.g. (1, 2, 1, 1) returns 1, (1, 2, 3, 1) returns -1, and (1, 1, 1, 1) returns -1
+   * @param size size of tensor
+   * @return (the only dim whose value > 1) else (-1)
+   */
+  private def getOnlyDimGtOne(size: Array[Int]): Int = {
+    var i = 0
+    var count = 0
+    var pivot = 0
+    while (i < size.length) {
+      if (size(i) > 1) {
+        count += 1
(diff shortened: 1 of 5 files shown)


SeaOfOcean:
pr validate pass

dl/src/main/scala/com/intel/analytics/bigdl/nn/CAdd.scala
```diff
@@ -50,8 +50,19 @@ class CAdd[@specialized(Float, Double) T: ClassTag](
       } else {
         bias.view(Array(1) ++ bias.size())
       }
-      expand.expandAs(output)
-      output.add(expand)
+      if (bias.size().count(x => x != 1) == 1) {
+        val biasDim = expand.size().zip(Stream.from(1)).filter(x => x._1 != 1)(0)._2
```
yiheng: this looks very confusing... do we really need to write in this style?

dl/src/main/scala/com/intel/analytics/bigdl/nn/CAdd.scala
```diff
@@ -50,8 +50,19 @@ class CAdd[@specialized(Float, Double) T: ClassTag](
       } else {
         bias.view(Array(1) ++ bias.size())
       }
-      expand.expandAs(output)
-      output.add(expand)
+      if (bias.size().count(x => x != 1) == 1) {
```
yiheng: please add some comments to explain why do this optimization?

dl/src/main/scala/com/intel/analytics/bigdl/nn/CMul.scala
```diff
@@ -50,9 +50,19 @@ class CMul[@specialized(Float, Double) T: ClassTag](
       } else {
         weight.view(Array(1) ++ weight.size())
       }
-
-      expand.expandAs(output)
-      output.cmul(expand)
+      if (weight.size().count(x => x != 1) == 1) {
```
yiheng: please add some comments to explain why do this optimization?

dl/src/main/scala/com/intel/analytics/bigdl/nn/CMul.scala
```diff
@@ -50,9 +50,19 @@ class CMul[@specialized(Float, Double) T: ClassTag](
...
+        val weightDim = expand.size().zip(Stream.from(1)).filter(x => x._1 != 1)(0)._2
```
yiheng: this looks very confusing... do we really need to write in this style?

yiheng:
Can you add unit test for the optimized path?

SeaOfOcean:
pr validate pass

dl/src/main/scala/com/intel/analytics/bigdl/nn/CMul.scala
```diff
@@ -53,13 +53,75 @@ class CMul[@specialized(Float, Double) T: ClassTag](
...
   }
 
+  /**
+   * if there is only one dim of size > 1, return this dim(count from 1)
+   * else return -1
+   * e.g. (1, 2, 1, 1) returns 1, (1, 2, 3, 1) returns -1, and (1, 1, 1, 1) returns -1
+   * @param size size of tensor
+   * @return (the only dim whose value > 1) else (-1)
+   */
+  private def getOnlyDimGtOne(size: Array[Int]): Int = {
```
yiheng: is it same with CAdd? if so, can we put them into a utils class instead of two copies of code?

SeaOfOcean:
pr validate pass

yiheng:
LGTM

Issue #460, opened 2017-02-13
toTensor and toTable as virtual functions


jason-dai:
Instead of using pattern match based on types (in essence instanceof operator in JVM), implement them as virtual functions in Tensor or Table.

PR #471, opened 2017-02-16
[Issue #460] Implemented `toTable` and `toTensor` as virtual functions in Tensor or Table

psyyz10:

Instead of using pattern match based on types (in essence instanceof operator in JVM), implement them as virtual functions in Tensor or Table.

Add corresponding tests

## How was this patch tested?

`toTable` and `toTensor` are tested in `DensenTensorSpec` and `TableSpec`

fixed https://github.com/intel-analytics/BigDL/issues/460

### DIFF
--- a/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
+++ b/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
@@ -1881,9 +1881,9 @@ private[tensor] class DenseTensor[@specialized(Float, Double) T: ClassTag](
     if (ev.getType() == env.getType()) {
       this.asInstanceOf[Tensor[D]]
     } else {
-      throw new IllegalArgumentException(s"The type ${env.getType()}" +
-        s" in toTensor[${env.getType()}] is not same" +
-        s"as the numeric type ${ev.getType()} of the " +
+      throw new IllegalArgumentException(s"The type ${env.getType().getClass}" +
+        s" in toTensor[${env.getType().getClass}] is not same" +
+        s"as the numeric type ${ev.getType().getClass} of the " +
         "corresponding module, please keep them same.")
     }
   }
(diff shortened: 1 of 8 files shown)


dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
```diff
@@ -1872,6 +1872,15 @@ private[tensor] class DenseTensor[@specialized(Float, Double) T: ClassTag](
     }
     this
   }
+
+  override def toTensor[D](implicit env: TensorNumeric[D]): Tensor[D] = {
+    if (ev == env) {
```
jason-dai: use ev.getType() = env.getType()

dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
```diff
@@ -1872,6 +1872,15 @@ private[tensor] class DenseTensor[@specialized(Float, Double) T: ClassTag](
...
+      this.asInstanceOf[Tensor[D]]
+    } else {
+      throw new IllegalArgumentException("The type D in toTensor[D] is not same" +
+        "as the numeric type of the corresponding module, please keep them same.")
```
jason-dai: Show env.getType() here

dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
psyyz10: Done

dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
psyyz10: Done

jason-dai:
LGTM

yiheng:
LGTM

PR #475, opened 2017-02-16
[faster-rcnn] add InferReshape

SeaOfOcean:

add InferReshape

## How was this patch tested?
unit test

### DIFF
--- a/dl/src/main/scala/com/intel/analytics/bigdl/nn/InferReshape.scala
+++ b/dl/src/main/scala/com/intel/analytics/bigdl/nn/InferReshape.scala
@@ -1,0 +1,137 @@
+/*
+ * Licensed to Intel Corporation under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * Intel Corporation licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.intel.analytics.bigdl.nn
+
+import breeze.linalg.sum
+import com.intel.analytics.bigdl.nn.abstractnn.TensorModule
+import com.intel.analytics.bigdl.tensor.Tensor
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+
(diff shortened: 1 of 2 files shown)


SeaOfOcean:
pr validate pass

dl/src/main/scala/com/intel/analytics/bigdl/nn/InferReshape.scala
```diff
@@ -0,0 +1,137 @@
...
+ * @tparam T type
+ */
+class InferReshape[@specialized(Float, Double) T: ClassTag](
+  size: Array[Int], var batchMode: Boolean = false)(
+  implicit ev: TensorNumeric[T]) extends TensorModule[T] {
+  val inferedSizes = if (batchMode) new Array[Int](size.length + 1) else new Array[Int](size.length)
+
+  override def updateOutput(input: Tensor[T]): Tensor[T] = {
+    if (size.contains(-1)) {
+      assert(sum(size.map(x => if (x == ev.fromType(-1)) 1 else 0)) == 1,
```
yiheng: Please use require instead of assert here. And you needn't check the size in every forward. I think it should only do once when the class is inited

dl/src/main/scala/com/intel/analytics/bigdl/nn/InferReshape.scala
```diff
@@ -0,0 +1,137 @@
...
+        "at most a single (1) value of -1 may be specified")
+    }
+    var inferIndex = -1
+    var count = 1
+    val startIndex = if (batchMode) 1 else 0
+    var i = 0
+    while (i < size.length) {
+      if (size(i) == 0) {
+        inferedSizes(i + startIndex) = input.size(i + startIndex + 1 - startIndex)
```
yiheng: input.size(i + startIndex + 1 - startIndex) equals input.size(i + 1)?

yiheng: what's size(i) == 0 for? Can you mention in the comments?

yiheng: I can see only the first if branch need input. Can we move the other two if branch to the init of the class?

dl/src/main/scala/com/intel/analytics/bigdl/nn/InferReshape.scala
```diff
@@ -0,0 +1,137 @@
...
+import com.intel.analytics.bigdl.tensor.Tensor
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+
+import scala.reflect.ClassTag
+
+/**
+ * Reshape with the support of infered size, the infered dim is indicated as -1
+ * note that at most one infered dim is supported,
+ * the element number is not changed after reshape
+ * e.g. reshape from (3, 2) to (2, -1) is reshape to (2, 3)
```
yiheng: It's not easy to understand in the example. Please add more explanation to explain where is the 3 come from in the size after reshape.

yiheng:
LGTM

PR #478, opened 2017-02-16
[Issue #477] Make linear support no bias for GRU and LSTM support

psyyz10:

Make `Linear` layer support no bias, for GRU and LSTM support.

## How was this patch tested?

Add tests to LinearSpec

fixed https://github.com/intel-analytics/BigDL/issues/477

### DIFF
--- a/dl/src/main/scala/com/intel/analytics/bigdl/nn/Linear.scala
+++ b/dl/src/main/scala/com/intel/analytics/bigdl/nn/Linear.scala
@@ -199,7 +199,7 @@ object Linear {
       inputSize: Int,
       outputSize: Int,
       initMethod: InitializationMethod = Default,
-      withBias: Boolean = false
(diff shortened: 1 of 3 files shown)


dl/src/main/scala/com/intel/analytics/bigdl/nn/Linear.scala
```diff
@@ -44,14 +44,15 @@ import com.intel.analytics.bigdl.nn.abstractnn.TensorModule
 class Linear[T: ClassTag](
   inputSize: Int,
   outputSize: Int,
-  private var initMethod: InitializationMethod = Default
+  private var initMethod: InitializationMethod = Default,
+  withBias: Boolean = true
```
yiheng: add another private var?

yiheng:
LGTM

PR #483, opened 2017-02-17
Batch padding

zhangxiaoli73:
Add SampleToBatch and SampleToBatchLM transformer, support:
(1) padding all batches to same length with padding value
(2) padding  all features to same length only in one batch with padding value
(3) just makeup batch, no padding

## How was this patch tested?
add unit test.

1. Use little data to produce local dataset, and than compare batch padding result with ground true.

2. Compare with previous transformer.
- when no padding, should be same to SampleToBatch
- when padding to same length for all batch, should be same to LabeledSentenceToSample(dictionaryLength,Some(trainMaxLength), Some(trainMaxLength) + SampleToBatch


dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/SampleToBatchPadding.scala
```diff
@@ -0,0 +1,139 @@
...
+          var i = 0
+          var maxDataSize: Array[Int] = null
+          var maxElement = 0
+          if (sampleData == null) sampleData = new Array[Sample[T]](batchSize)
+          while (i < batchSize && prev.hasNext) {
+            val sample = prev.next()
+            require(sample.feature().isContiguous() && sample.label().isContiguous(),
+              "SampleToBatchPadding: Only support contiguous tensor, pls use " +
+                "tensor.contiguous() before batching")
+            if (sampleData(i) == null) sampleData(i) = Sample()
```
JerryYanWan: What if the next batchSize is larger, then the sampleData will throw indexOutOfBoundary error?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/SampleToBatchPadding.scala
```diff
@@ -0,0 +1,139 @@
...
+            sampleData(i).copy(sample)
+            if (sample.feature().nElement() > maxElement) {
+              maxElement = sample.feature().nElement()
+              maxDataSize = sample.feature().size()
+            }
+            if (!padding) require(util.Arrays.equals(sample.feature().size(), maxDataSize),
```
JerryYanWan: If the sample.feature.nElement is ascending, will the requirement always be satisfied?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/SampleToBatchPadding.scala
```diff
@@ -0,0 +1,139 @@
...
+              "SampleToBatchPadding: all feature size should be same when no padding")
+            i += 1
+          }
+          val batchLength = i
+
+          // init
+          val padLength = paddingLength.getOrElse(maxDataSize(0))
```
JerryYanWan: How about defining a variable batchDim = 0; use batchDim instead of 0?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/SampleToBatchPadding.scala
```diff
@@ -0,0 +1,139 @@
...
+          require(padLength >= maxDataSize(0), "SampleToBatchPadding: paddingLength should not " +
+            "be less than first dimension of feature")
+          oneFeatureLength = padLength * maxDataSize(1)
```
JerryYanWan: define a variable to indicate the meaning of 1?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/SampleToBatchPadding.scala
```diff
@@ -0,0 +1,139 @@
...
+          oneLabelLength = sampleData(0).label().nElement()
+          featureSize = Array(batchLength, padLength, maxDataSize(1))
+          labelSize = Array(batchLength, oneLabelLength)
```
JerryYanWan: the labelSize dimension is constrained to be 2?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/SampleToBatchPadding.scala
zhangxiaoli73: yeah, changed

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/SampleToBatchPadding.scala
zhangxiaoli73: thanks, changed

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/SampleToBatchPadding.scala
zhangxiaoli73: thanks, changed

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/SampleToBatchPadding.scala
zhangxiaoli73: perhaps not, changed

jason-dai:
We need to:
1) Combine SampleToBatch, SampleToBatchPadding and SampleToBatchPaddingLM into one class (instead of 3)
2) Don't treat startToken or stopToken as special case; they are just padding values for feature or label

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/SampleToBatchPadding.scala
```diff
@@ -0,0 +1,167 @@
...
+              labelSize = sample.label().size()
+            }
+            i += 1
+          }
+          val batchLength = i
+
+          // init
+          var padLength = if (padFeature) fixLength.getOrElse(featureFirstDim) else featureFirstDim
+          require(padLength >= featureFirstDim, "SampleToBatchPadding: fixLength " +
+            "should not be less than first dimension of feature when padding")
```
JerryYanWan: please add ${padLength} ${featureFirstDim} to the requirement

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/SampleToBatchPadding.scala
```diff
@@ -0,0 +1,167 @@
...
+          oneFeatureLength = padLength * featureSize.product / featureFirstDim
+          featureSize = Array(batchLength, padLength) ++ featureSize.slice(1, featureSize.length)
+
+          padLength = if (padLabel) fixLength.getOrElse(labelFirstDim) else labelFirstDim
+          require(padLength >= labelFirstDim, "SampleToBatchPadding: fixLength " +
```
JerryYanWan: the same

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/SampleToBatchPadding.scala
```diff
@@ -0,0 +1,167 @@
...
+            i += 1
+          }
+          val batchLength = i
+
+          // init
+          var padLength = if (padFeature) fixLength.getOrElse(featureFirstDim) else featureFirstDim
+          require(padLength >= featureFirstDim, "SampleToBatchPadding: fixLength " +
+            "should not be less than first dimension of feature when padding")
+          oneFeatureLength = padLength * featureSize.product / featureFirstDim
+          featureSize = Array(batchLength, padLength) ++ featureSize.slice(1, featureSize.length)
```
JerryYanWan: reorder these two lines?
featureSize(padDim) = padLength
oneFeatureLength = featureSize.product

The meaning of featureSize might be confused since it indicates the size of one feature at first and points to a batch of input feature later.

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/SampleToBatchPadding.scala
```diff
@@ -0,0 +1,167 @@
...
+              require(util.Arrays.equals(sample.feature().size(), featureSize),
+                "SampleToBatchPadding: all feature size should be same when not padding feature")
+            }
+            if ((labelSize != null) && (!padLabel)) {
+              require(util.Arrays.equals(sample.label().size(), labelSize),
+                "SampleToBatchPadding: all label size should be same when not padding label")
+            }
+            sampleData(i).copy(sample)
+            if ((featureSize == null) || (sample.feature().nElement() > featureSize.product)) {
+              featureFirstDim = sample.feature().size(1)
```
JerryYanWan: remove this line?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/SampleToBatchPadding.scala
```diff
@@ -0,0 +1,167 @@
...
+ * @param fixLength padding fix length, default the max first dimension length of feature
+ *                  or label in a batch
+ */
+class SampleToBatchPadding[T: ClassTag]
+(totalBatch : Int,
+ padFeature : Boolean,
+ padLabel : Boolean,
+ padFeatureValue : Option[T] = None,
+ padLabelValue : Option[T] = None,
+ fixLength: Option[Int] = None)
```
JerryYanWan: add padDim: Option[Int] = Some(1)

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/SampleToBatchPadding.scala
JerryYanWan: featureFirstDim = sample.feature().size(padDim)

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/SampleToBatchPadding.scala
```diff
@@ -0,0 +1,167 @@
...
+                "SampleToBatchPadding: all label size should be same when not padding label")
+            }
+            sampleData(i).copy(sample)
+            if ((featureSize == null) || (sample.feature().nElement() > featureSize.product)) {
+              featureFirstDim = sample.feature().size(1)
+              featureSize = sample.feature().size()
+
+            }
+            if ((labelSize == null) || (sample.label().nElement() > labelSize.product)) {
+              labelFirstDim = sample.label().size(1)
```
JerryYanWan: rename to labelPadLength = sample.label().size(padDim)

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/SampleToBatchPadding.scala
```diff
@@ -0,0 +1,167 @@
...
+          if (padLabel) padding(labelData, padLabelValue.getOrElse(ev.zero))
+
+          i = 0
+          while (i < batchLength) {
+            val sample = sampleData(i)
+            sample.copyFromFeature(featureData, i * oneFeatureLength, sample.feature().nElement())
+            sample.copyFromLabel(labelData, i * oneLabelLength, sample.label().nElement())
+            i += 1
+          }
+          featureTensor.set(Storage[T](featureData), storageOffset = 1, sizes = featureSize)
```
JerryYanWan: sizes = Array(batchLength) ++ featureSize

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/SampleToBatchPadding.scala
```diff
@@ -0,0 +1,184 @@
...
+      case DoubleType =>
+        util.Arrays.fill(data.asInstanceOf[Array[Double]], 0, data.length,
+          ev.toType[Double](padValue))
+      case FloatType =>
+        util.Arrays.fill(data.asInstanceOf[Array[Float]], 0, data.length,
+          ev.toType[Float](padValue))
+      case _ => throw new UnsupportedOperationException(
+        "SampleToBatchPadding: Only Float/Double supported")
+    }
+  }
```
JerryYanWan: How about providing the (fromIndex and toIndex) parameters for Arrays.fill to avoid filling the whole array each time?

def padding(data: Array[T], fromIndex: Int, toIndex: Int, padValue: T)

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/SampleToBatchPadding.scala
```diff
@@ -0,0 +1,184 @@
...
+            }
+            sampleData(i).copy(sample)
+            if (sample.feature().nElement() > maxfeatureElement) {
+              maxfeatureElement = sample.feature().nElement()
+              featureIndex = i
+            }
+            if (sample.label().nElement() > maxlabelElement) {
+              maxlabelElement = sample.label().nElement()
+              labelIndex = i
+            }
```
JerryYanWan: Since it is the same operation for feature and label, how about wrapping the operation to a private function and call that function both for feature and label?

private def less(int a, int b, int i, int j) = if (a < b) (b, j) else (a, i)
maxfeatureElement = less(sample.feature().nElement, maxfeatureElement, i, featureIndex)
maxlabelElement = less(sample.label().nElement, maxlabelElement, i, labelIndex)

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/SampleToBatchPadding.scala
```diff
@@ -0,0 +1,184 @@
...
+          // init
+          padFeatureLength = sampleData(featureIndex).feature().size(padFeatureDim)
+          padLength = if (padFeature) fixLength.getOrElse(padFeatureLength) else padFeatureLength
+          require(padLength >= padFeatureLength, "SampleToBatchPadding: fixLength " +
+            "should not be less than first dimension of feature when padding")
+          featureSize(padFeatureDim) = padLength
+          featureSize(0) = 1
+          oneFeatureLength = featureSize.product
+          featureSize(0) = batchLength
+
```
JerryYanWan: write a private function both for feature and label?

private def getSize(size: Array[Int], length: Int, fixLength: Option[Int], isPad: Boolean): Unit = {
}

PR #488, opened 2017-02-20
[faster-rcnn] add scale module

SeaOfOcean:

add scale module

## How was this patch tested?

unit test


yiheng:
LGTM

PR #490, opened 2017-02-20
sort dataset

zhangxiaoli73:
1. local dataset shuffle index, and use random offset when training
2. get original rdd dataset indexes when no shuffle
3. shuffle index range: form 0 to data.length - groupSize (default 1)
3. no shuffle if users want.

## How was this patch tested?
add some unit test for sorted buffer
1. local dataset should be good with default groupSize = 1
2. local dataset should be good with  groupSize > 1
3. rdd dataset should be good for multi partitions
4. rdd dataset should be same to the original data with one partition

### DIFF
--- a/dl/src/main/scala/com/intel/analytics/bigdl/dataset/DataSet.scala
+++ b/dl/src/main/scala/com/intel/analytics/bigdl/dataset/DataSet.scala
@@ -22,7 +22,6 @@ import java.util.concurrent.atomic.AtomicInteger
 
 import com.intel.analytics.bigdl.DataSet
 import com.intel.analytics.bigdl.dataset.image.{LabeledBGRImage, _}
-import com.intel.analytics.bigdl.dataset.text.LabeledSentence
 import com.intel.analytics.bigdl.utils.{Engine, RandomGenerator}
 import org.apache.hadoop.io.Text
 import org.apache.log4j.Logger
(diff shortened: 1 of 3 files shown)


dl/src/main/scala/com/intel/analytics/bigdl/dataset/DataSet.scala
```diff
@@ -300,20 +318,48 @@ object DataSet {
...
+   * @return
+   */
+  def sortData[T: ClassTag](data: Array[T], isSort: Boolean): Array[T] = {
+    if (isSort) {
+      implicit val ord = Ordering.fromLessThan[Int]((e1, e2) => (e1 > e2))
+      if (classTag[T] == classTag[Sample[Float]]) {
+        data.sortBy(_.asInstanceOf[Sample[Float]].feature().nElement())
+      } else if (classTag[T] == classTag[Sample[Double]]) {
+        data.sortBy(_.asInstanceOf[Sample[Double]].feature().nElement())
+      } else if (classTag[T] == classTag[LabeledSentence[Float]]) {
```
zhangxiaoli73: only sample

dl/src/main/scala/com/intel/analytics/bigdl/dataset/DataSet.scala
```diff
@@ -125,14 +126,25 @@ trait LocalDataSet[T] extends AbstractDataSet[T, Iterator[T]] {
  * Wrap an array as a DataSet.
  * @tparam T
  */
-class LocalArrayDataSet[T] private[dataset](buffer: Array[T]) extends LocalDataSet[T] {
+class LocalArrayDataSet[T] private[dataset](
+  buffer: Array[T],
+  isSort: Boolean = false,
+  groupSize: Int = 1) extends LocalDataSet[T] {
```
zhangxiaoli73: no sort , no groupSize

Issue #463, opened 2017-02-14
BigDL on Apache Spark1.6.2 validation failed


kaiseu:
BigDL on Apache Spark1.6.2 validation failed: 

17/02/14 16:41:09 INFO optim.DistriOptimizer$: [Wall Clock 889.551336265s] Validate model...
Exception in thread "dispatcher-event-loop-6" java.lang.NoSuchMethodError: org.apache.spark.util.Utils$.tryOrIOException(Lscala/Function0;)V
        at org.apache.spark.rdd.ZippedPartitionsLocalityPartition.writeObject(ZippedPartitionsWithLocalityRDD.scala:106)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:497)
        at java.io.ObjectStreamClass.invokeWriteObject(ObjectStreamClass.java:1028)
        at java.io.ObjectOutputStream.writeSerialData(ObjectOutputStream.java:1496)
        at java.io.ObjectOutputStream.writeOrdinaryObject(ObjectOutputStream.java:1432)
        at java.io.ObjectOutputStream.writeObject0(ObjectOutputStream.java:1178)
        at java.io.ObjectOutputStream.defaultWriteFields(ObjectOutputStream.java:1548)
        at java.io.ObjectOutputStream.writeSerialData(ObjectOutputStream.java:1509)
        at java.io.ObjectOutputStream.writeOrdinaryObject(ObjectOutputStream.java:1432)
        at java.io.ObjectOutputStream.writeObject0(ObjectOutputStream.java:1178)
        at java.io.ObjectOutputStream.writeObject(ObjectOutputStream.java:348)
        at org.apache.spark.serializer.JavaSerializationStream.writeObject(JavaSerializer.scala:44)
        at org.apache.spark.serializer.JavaSerializerInstance.serialize(JavaSerializer.scala:101)
        at org.apache.spark.scheduler.Task$.serializeWithDependencies(Task.scala:211)
        at org.apache.spark.scheduler.TaskSetManager.resourceOffer(TaskSetManager.scala:460)
[...]

PR #496, opened 2017-02-21
[Fix Issue 463] Fix the tryOrIOException on Spark 1.6.2

yiheng:

do not use tryOrIOException method as it has different interface in sparrk 1.5 and 1.6. This change doesn't affect spark 2.0 related code

## How was this patch tested?

Unit test

fixed https://github.com/intel-analytics/BigDL/issues/463


yiheng:
test passed. merge the PR

Issue #497, opened 2017-02-22
Seperate `updateGradInput` and `accGradParameters` in `Sequential` 


psyyz10:
`Sequential` Implementation (not same as Torch) has an issue, maybe `updateGradInput` and `accGradParameters` should be seperate rather than updating them together, otherwise it may not work correctly in some case, for example in `Recurrent`.

PR #498, opened 2017-02-22
Add accGradParameters to Sequential and test it

psyyz10:

`Sequential` Implementation (not same as Torch) has an issue, maybe `updateGradInput` and `accGradParameters` should be seperate rather than updating them together, otherwise it may not work correctly in some case, for example in `Recurrent`.

Hence added accGradParameters to Sequential and test it

## How was this patch tested?

Unit test in SequentialSpec

fixed https://github.com/intel-analytics/BigDL/issues/497

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Sequential.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Sequential.scala
@@ -74,6 +74,20 @@ class Sequential[T: ClassTag]
     currentModule.accGradParameters(input, currentGradOutput, scale)
   }
 
+  override def backward(input: Activity, nextError: Activity): Activity = {
+    var i = modules.length - 1
+    var error = nextError.asInstanceOf[Activity]
+    while (i > 0) {
+      val input = modules(i - 1).output
+      error = modules(i).backward(input, error)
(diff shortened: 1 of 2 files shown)


zhangxiaoli73:
as you mentioned, it may not work correctly in some case, for example in Recurrent. So please add a test to cover the case.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Sequential.scala
```diff
@@ -47,6 +47,38 @@ class Sequential[T: ClassTag]
...
+      currentModule.accGradParameters(previousModule.output, currentGradOutput, scale)
+      currentGradOutput = currentModule.gradInput
+      currentModule = previousModule
+      i -= 1
+    }
+
+    currentModule.accGradParameters(input, currentGradOutput, scale)
+  }
+
+  override def backward(input: Activity, nextError: Activity): Activity = {
```
zhangxiaoli73: perhaps you should add more docs to explain the reason to add it.

psyyz10: The origin torch version has this method

PR #502, opened 2017-02-22
[faster-rcnn] generate Anchor

SeaOfOcean:

add anchor

## How was this patch tested?

unit test


dl/src/main/scala/com/intel/analytics/bigdl/models/fasterrcnn/utils/Anchor.scala
```diff
@@ -0,0 +1,224 @@
...
+    allAnchors
+  }
+
+  /**
+   * Generate anchor (reference) windows by enumerating aspect ratios(M) X scales(N)
+   * wrt a reference (0, 0, 15, 15) window.
+   * 1. generate anchors for different ratios (4, M)
+   * 2. for each anchors generated in 1, scale them to get scaled anchors (4, M*N)
+   */
+  def generateBasicAnchors(ratios: Tensor[Float], scales: Tensor[Float],
```
yiheng: private

PR #505, opened 2017-02-23
sort dataset

zhangxiaoli73:

    get original rdd dataset indexes when no shuffle
    index offset range: form 0 to data.length - groupSize (default 1)

## How was this patch tested?

add some unit test for sorted buffer
    rdd dataset should be good for multi partitions
    rdd dataset should be same to the original data with one partition

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/DataSet.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/DataSet.scala
@@ -122,6 +122,7 @@ trait LocalDataSet[T] extends AbstractDataSet[T, Iterator[T]] {
 
 /**
  * Wrap an array as a DataSet.
+ * @param buffer
  * @tparam T
  */
 class LocalArrayDataSet[T] private[dataset](buffer: Array[T]) extends LocalDataSet[T] {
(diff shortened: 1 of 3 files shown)


dl/src/main/scala/com/intel/analytics/bigdl/dataset/DataSet.scala
```diff
@@ -211,16 +216,18 @@ class CachedDistriDataSet[T: ClassTag] private[dataset] (buffer: RDD[Array[T]])
   }).reduce(_ + _)
 
   protected var indexes: RDD[Array[Int]] = buffer.mapPartitions(iter => {
-    Iterator.single(RandomGenerator.shuffle((0 until iter.next().length).toArray))
```
yiheng: I think the old behavior should not be changed.

if(isInOrder) {
   new code
} else {
   old code
}

dl/src/main/scala/com/intel/analytics/bigdl/dataset/DataSet.scala
```diff
@@ -314,6 +323,48 @@ object DataSet {
...
+   * @return
+   */
+  def sortData[T: ClassTag](data: Array[T], isInOrder: Boolean): Array[T] = {
+    if (isInOrder) {
+      if (classTag[T] == classTag[Sample[Float]]) {
+        data.sortBy(_.asInstanceOf[Sample[Float]].feature().nElement())
+      } else if (classTag[T] == classTag[Sample[Double]]) {
+        data.sortBy(_.asInstanceOf[Sample[Double]].feature().nElement())
+      } else {
+        throw new IllegalArgumentException("DataSet.sortData: Only support sort for sample input")
```
yiheng: Sample[Int] is also a sample input

dl/src/main/scala/com/intel/analytics/bigdl/dataset/DataSet.scala
zhangxiaoli73: In optimizer, there is dataset.shuffle() before training, so old behavior will not be changed.

zhangxiaoli73: In optimizer, there is dataset.shuffle() before training, so old behavior will not be changed.

dl/src/main/scala/com/intel/analytics/bigdl/dataset/DataSet.scala
zhangxiaoli73: ok, use Sample[_] instead of Sample[Double], etc.

yiheng:
LGTM

zhangxiaoli73:
jenkins pass

PR #508, opened 2017-02-25
Table class should not publish internal mutable state to the outside world

yangw1234:

The `Table` class has a `getState()` which used to return the internal mutable state to caller. It is a little dangerous to do that since the `state` could be modified somewhere else, which make it hard for Table to ensure its own invariants. 

This pr fixes this issue by 
1. add `foreach` and `keySet` methods for the caller to get easy readonly access to the elements of the internal `state`
2. modify `getState()` to return a defensive copy of the original state

## How was this patch tested?

existing tests

### DIFF
--- a/dl/src/main/scala/com/intel/analytics/bigdl/nn/DotProduct.scala
+++ b/dl/src/main/scala/com/intel/analytics/bigdl/nn/DotProduct.scala
@@ -56,7 +56,7 @@ class DotProduct[T: ClassTag] (implicit ev: TensorNumeric[T])
     var input2: Tensor[T] = input(2)
     var notBatch = false
 
-    if (gradInput.getState().size != 2) {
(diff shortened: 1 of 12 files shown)


yiheng:
LGTM

jason-dai:
Shall we just remove getState()? Or make it private?

yangw1234:
@jason-dai I thought of removing getState() at first, but I was not sure whether we were allowed to change a API like this, so I modify it to return a defensive copy. If it is OK to remove it, I'd love to do that.

yiheng:
I think we can remove it.

yangw1234:
I removed the getState() method

yiheng:
jenkins test job pass. Merge the PR. @yangw1234 thanks for your contribution.

PR #510, opened 2017-02-26
Implement and test SplitTable

psyyz10:

Implement and test SplitTable for GRU support

## How was this patch tested?

Unit tests in SplitTableSpec.scala

fixed https://github.com/intel-analytics/BigDL/issues/509


yiheng:
LGTM

dl/src/main/scala/com/intel/analytics/bigdl/nn/SplitTable.scala
```diff
@@ -0,0 +1,116 @@
...
+@SerialVersionUID(- 4318640284973082779L)
+class SplitTable[T: ClassTag](
+  var dimension: Int,
+  var nInputDims: Int = -1)
+  (implicit ev: TensorNumeric[T]) extends AbstractModule[Tensor[T], Table, T]{
+
+  private def getPositiveDimension(input: Tensor[T]): Int = {
+    if (dimension < 0) {
+      input.dim() + dimension + 1
+    } else if (dimension != Int.MinValue && input.dim() == nInputDims + 1) {
```
zhangxiaoli73: why add dimension != Int.MinValue

dl/src/main/scala/com/intel/analytics/bigdl/nn/SplitTable.scala
```diff
@@ -0,0 +1,116 @@
...
+    } else {
+      dimension
+    }
+  }
+
+  override def updateOutput(input: Tensor[T]): Table = {
+    val dim = getPositiveDimension(input)
+    val slices = input.size(dim)
+
+    val currentOutput = T()
```
zhangxiaoli73: why not use output directly instead of new a table

dl/src/main/scala/com/intel/analytics/bigdl/nn/SplitTable.scala
psyyz10: changed to dimension != Int.MaxValue

zhangxiaoli73:
LGTM

PR #512, opened 2017-02-27
Batch Padding

zhangxiaoli73:
Modify SampleToBatch transformer to support:
(1) padding all features to the same length in one batch with tensor value
(2) padding all labels to the same length in one batch with T value
(3) padding features or labels to the fix length
(4) just makeup batch, no padding

## How was this patch tested?
add unit test.
Use little data to produce local dataset, and than compare batch padding result with ground true.
Compare with previous transformer.
    when no padding, should be same to previous SampleToBatch
    when padding to same length for all batch, should be same to LabeledSentenceToSample(dictionaryLength,Some(trainMaxLength), Some(trainMaxLength) +  previous SampleToBatch


dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/SampleToBatch.scala
```diff
@@ -0,0 +1,190 @@
...
+ fixLength: Option[Int] = None)
+(implicit ev: TensorNumeric[T])
+  extends Transformer[Sample[T], MiniBatch[T]] {
+
+  private def paddingTensor(data: Array[T], padValue: Tensor[T], start: Int, end: Int): Unit = {
+    var offset = start
+    val padArr = padValue.storage().array()
+    while (offset < end) {
+      val length = math.min(end - offset, padArr.length)
+      System.arraycopy(padArr, 0, data, offset, length)
```
JerryYanWan: The padArr should start from its offset instead of 0 ?

zhangxiaoli73: yeah, changed

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -70,66 +72,162 @@ class Identity[A] extends Transformer[A, A] {
...
 }
 
+
 object SampleToBatch {
   def apply[T: ClassTag]
-  (batchSize: Int)
+  (batchSize : Int,
+   padFeatureValue : Option[Tensor[T]] = None,
+   padLabelValue : Option[T] = None,
+   fixLength: Option[Int] = None)
```
jason-dai: please add comments and scala doc; change the arguments to featurePadding, lablePadding, and fixedLength.

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -70,66 +72,162 @@ class Identity[A] extends Transformer[A, A] {
...
+ * @param fixLength padding fix length, default the max first dimension length of feature
+ *                  or label in a batch
+ */
 class SampleToBatch[T: ClassTag]
-  (totalBatch: Int)
-  (implicit ev: TensorNumeric[T])
+(totalBatch : Int,
+ padFeatureValue : Option[Tensor[T]] = None,
+ padLabelValue : Option[T] = None,
+ fixLength: Option[Int] = None)
```
jason-dai: change the arguments to featurePadding, lablePadding, and fixedLength

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -70,66 +72,162 @@ class Identity[A] extends Transformer[A, A] {
...
 
+  private def paddingTensor(data: Array[T], padValue: Tensor[T], start: Int, end: Int): Unit = {
+    var offset = start
+    val padArr = padValue.storage().array()
+    val padOffset = padValue.storageOffset() - 1
+    while (offset < end) {
+      val length = math.min(end - offset, padArr.length)
+      System.arraycopy(padArr, padOffset, data, offset, length)
+      offset += length
+    }
```
jason-dai: This is costly; we should only copy the padding data to where they are needed, instead of to the entire tensor.

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -70,66 +72,162 @@ class Identity[A] extends Transformer[A, A] {
...
+  }
+
+  private def paddingValue(data: Array[T], padValue: T, start: Int, end: Int): Unit = {
+    ev.getType() match {
+      case DoubleType =>
+        util.Arrays.fill(data.asInstanceOf[Array[Double]], start, end, ev.toType[Double](padValue))
+      case FloatType =>
+        util.Arrays.fill(data.asInstanceOf[Array[Float]], start, end, ev.toType[Float](padValue))
```
jason-dai: This is costly; we should only copy the padding data to where they are needed, instead of to the entire tensor.

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -70,66 +72,162 @@ class Identity[A] extends Transformer[A, A] {
...
-              labelSize = Array(1) ++ sample.label().size
-              featureData = new Array[T](batchSize * oneFeatureLength)
-              labelData = new Array[T](batchSize * oneLabelLength)
+            if ((maxlabelElement != 0) && (!padLabel)) {
+              require(sample.label().nElement() == maxlabelElement,
+                "SampleToBatchPadding: all label size should be same when not padding label")
             }
-            sample.copyFromLabel(labelData, i*oneLabelLength, oneLabelLength)
-            sample.copyFromFeature(featureData, i*oneFeatureLength, oneFeatureLength)
+            sampleData(i).copy(sample)
```
jason-dai: Why do we need to copy here? Why not sampleData(i) = sample?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -70,66 +72,162 @@ class Identity[A] extends Transformer[A, A] {
...
             val sample = prev.next()
-            if(!sample.feature().isContiguous() || !sample.label().isContiguous()) {
-              throw new IllegalArgumentException(
-                "Only support contiguous tensor, pls use tensor.contiguous() before batching")
+            require(sample.feature().isContiguous() && sample.label().isContiguous(),
+              "SampleToBatchPadding: Only support contiguous tensor, pls use " +
+                "tensor.contiguous() before batching")
+            if (sampleData(i) == null) sampleData(i) = Sample()
+            if ((maxfeatureElement != 0) && (!padFeature)) {
+              require(sample.feature().nElement() == maxfeatureElement,
```
jason-dai: The check is not correct. Ideally we need to check if all samples in the batch have the same size (instead of same element#).

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -70,66 +72,162 @@ class Identity[A] extends Transformer[A, A] {
...
-               *
-               * the featureSize of a fixed sentence input is: [10, 1000]
-               * the labelSize of a fixed sentence input is: [10]
-               */
-              featureSize = Array(1) ++ sample.feature().size
-              labelSize = Array(1) ++ sample.label().size
-              featureData = new Array[T](batchSize * oneFeatureLength)
-              labelData = new Array[T](batchSize * oneLabelLength)
+            if ((maxlabelElement != 0) && (!padLabel)) {
+              require(sample.label().nElement() == maxlabelElement,
```
jason-dai: The check is not correct. Ideally we need to check if all labels in the batch have the same size (instead of same element#).

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -70,66 +72,162 @@ class Identity[A] extends Transformer[A, A] {
...
+                "SampleToBatchPadding: all label size should be same when not padding label")
             }
-            sample.copyFromLabel(labelData, i*oneLabelLength, oneLabelLength)
-            sample.copyFromFeature(featureData, i*oneFeatureLength, oneFeatureLength)
+            sampleData(i).copy(sample)
+            if (sample.feature().nElement() > maxfeatureElement) {
+              maxfeatureElement = sample.feature().nElement()
```
jason-dai: instead of maxfeatureElement, we actually need to compute max value of sample.feature().size(0)

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -70,66 +72,162 @@ class Identity[A] extends Transformer[A, A] {
...
+              featureIndex = i
+            }
+            if (sample.label().nElement() > maxlabelElement) {
+              maxlabelElement = sample.label().nElement()
```
jason-dai: instead of maxlabelElement , we actually need to compute max value of sample.label().size(0)

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -70,66 +72,162 @@ class Identity[A] extends Transformer[A, A] {
...
+      private var sampleData: Array[Sample[T]] = null
+
       private val batchSize = Utils.getBatchSize(totalBatch)
       private var featureSize: Array[Int] = null
       private var labelSize: Array[Int] = null
       private var oneFeatureLength: Int = 0
       private var oneLabelLength: Int = 0
+      private var padFeatureLength: Int = 0
+      private var padLabelLength: Int = 0
+      private val padFeature: Boolean = if (padFeatureValue.isEmpty) false else true
```
jason-dai: padFeature = padFeatureValue.isEmpty

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -70,66 +72,162 @@ class Identity[A] extends Transformer[A, A] {
...
+      private val padLabel: Boolean = if (padLabelValue.isEmpty) false else true
```
jason-dai: padLabel = !padLabelValue.isEmpty

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -70,66 +72,162 @@ class Identity[A] extends Transformer[A, A] {
...
+
   override def apply(prev: Iterator[Sample[T]]): Iterator[MiniBatch[T]] = {
     new Iterator[MiniBatch[T]] {
-      private var featureTensor: Tensor[T] = Tensor[T]()
-      private var labelTensor: Tensor[T] = Tensor[T]()
+      private val featureTensor: Tensor[T] = Tensor[T]()
+      private val labelTensor: Tensor[T] = Tensor[T]()
       private var featureData: Array[T] = null
       private var labelData: Array[T] = null
+      private var sampleData: Array[Sample[T]] = null
```
jason-dai: change to sampleArray?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -70,66 +72,162 @@ class Identity[A] extends Transformer[A, A] {
...
+            }
+            if (sample.label().nElement() > maxlabelElement) {
+              maxlabelElement = sample.label().nElement()
+              labelIndex = i
+            }
+            i += 1
+          }
+          val batchLength = i
+          if (featureSize == null) {
+            featureSize = Array(1) ++ sampleData(featureIndex).feature().size()
```
jason-dai: This is actually wrong: even if featureSize is not null, it can be different from the feature size in the current batch

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
jason-dai: and I think we should just preallocate for sampleData, something like: sampleData = Array.tabulate(batchSize)(_=>Sample())

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -70,66 +72,162 @@ class Identity[A] extends Transformer[A, A] {
...
+      private val labelTensor: Tensor[T] = Tensor[T]()
       private var featureData: Array[T] = null
       private var labelData: Array[T] = null
+      private var sampleData: Array[Sample[T]] = null
+
       private val batchSize = Utils.getBatchSize(totalBatch)
       private var featureSize: Array[Int] = null
       private var labelSize: Array[Int] = null
       private var oneFeatureLength: Int = 0
       private var oneLabelLength: Int = 0
```
jason-dai: Change to something like oneFeatureElement and oneLabelElement?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -70,66 +72,162 @@ class Identity[A] extends Transformer[A, A] {
...
+      private var padFeatureLength: Int = 0
+      private var padLabelLength: Int = 0
```
jason-dai: Why do we need to define these two here instead of inside next()?

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -70,66 +72,162 @@ class Identity[A] extends Transformer[A, A] {
...
+      case DoubleType =>
+        util.Arrays.fill(data.asInstanceOf[Array[Double]], start, end, ev.toType[Double](padValue))
+      case FloatType =>
+        util.Arrays.fill(data.asInstanceOf[Array[Float]], start, end, ev.toType[Float](padValue))
+      case _ => throw new UnsupportedOperationException(
+        "SampleToBatchPadding: Only Float/Double supported")
+    }
+  }
+
+  private def getSize(padLength: Int, batchLength: Int, tensorSize: Array[Int]): Int = {
```
jason-dai: The method name is confusing - it returns element# for the tensor, and then update tensorSize with batch size. Chane it to something meaningful, and add more comments.

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -70,66 +72,162 @@ class Identity[A] extends Transformer[A, A] {
...
+          if (padFeature) {
+            require(padFeatureValue.get.isContiguous(), "SampleToBatch: padFeatureValue " +
+              "should be contiguous")
+            require((padFeatureValue.get.dim() + 1) == sampleData(featureIndex).feature().dim(),
+              "SampleToBatch: padFeatureValue dim should be Feature dim - 1")
+            paddingTensor(featureData, padFeatureValue.get, 0, featureData.length)
+          }
+          if (padLabel) {
+            paddingValue(labelData, padLabelValue.get, 0, labelData.length)
+          }
```
jason-dai: As mentioned before, only copy the padding values to where they are needed.

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
zhangxiaoli73: ok

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
zhangxiaoli73: ok

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
zhangxiaoli73: ok

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
zhangxiaoli73: ok

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
zhangxiaoli73: ok

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
zhangxiaoli73: ok

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
zhangxiaoli73: ok

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
zhangxiaoli73: ok

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -70,66 +71,164 @@ class Identity[A] extends Transformer[A, A] {
   }
 }
 
+/**
+ * convert sample to MiniBatch, and you can choose whether padding feature or label
+ * to same length in a batch
+ */
```
jason-dai: Convert a sequence of Sample to a sequence of MiniBatch, optionally padding all the features (or labels) in the mini-batch to the same length

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -70,66 +71,164 @@ class Identity[A] extends Transformer[A, A] {
...
 object SampleToBatch {
   def apply[T: ClassTag]
-  (batchSize: Int)
+  (batchSize : Int,
+   featurePadding : Option[Tensor[T]] = None,
+   labelPadding : Option[T] = None,
+   fixedLength: Option[Int] = None)
```
jason-dai: featurePadding  feature padding value (by default None, meaning no feature padding)

labelPadding label padding value (by default None, meaning no label padding)

fixedLength if padding, it specifies the length of feature/label after padding (by default None, meaning the length after padding is set to the max length of feature/label in a mini-batch)

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -70,66 +71,164 @@ class Identity[A] extends Transformer[A, A] {
...
 
+/**
+ * convert sample to MiniBatch, and you can choose whether padding feature or label
+ * to same length in a batch
+ * @param totalBatch batchsize
+ * @param featurePadding  padding feature value if do padding
+ * @param labelPadding padding label value if do padding
+ * @param fixedLength padding fixed length, default the max first dimension length of feature
+ *                  or label in a batch
+ */
```
jason-dai: See above comments.

jason-dai:
Please update the comments and scala doc; otherwise LGTM.

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
zhangxiaoli73: ok

dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
zhangxiaoli73: ok

zhangxiaoli73:
comments changed, and jenkins pass.

PR #519, opened 2017-02-28
insert layer to the constructor of the TimeDistributed layer

JerryYanWan:
TimeDistributed, insert layer to the constructor instead of using add

## How was this patch tested?
Unit Test,
Test the correctness by comparing the results from TimeDistributed and added layer.

### DIFF
--- a/dl/src/main/scala/com/intel/analytics/bigdl/nn/TimeDistributed.scala
+++ b/dl/src/main/scala/com/intel/analytics/bigdl/nn/TimeDistributed.scala
@@ -119,7 +119,6 @@ class TimeDistributed[T : ClassTag] (layer: TensorModule[T])
    * to these parameters. Otherwise, it does nothing.
    */
   override def zeroGradParameters(): Unit = {
-    super.zeroGradParameters()
     layer.zeroGradParameters()
   }
 
@@ -177,7 +176,6 @@ class TimeDistributed[T : ClassTag] (layer: TensorModule[T])
    * @return this
    */
   override def copyStatus(src: Module[T]): TimeDistributed.this.type = {
-    super.copyStatus(src)
     layer.copyStatus(src)
     this
   }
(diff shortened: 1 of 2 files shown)


dl/src/main/scala/com/intel/analytics/bigdl/nn/TimeDistributed.scala
```diff
@@ -34,17 +35,23 @@ import scala.reflect.ClassTag
  * @tparam T
  */
 
-class TimeDistributed[T : ClassTag] ()
+class TimeDistributed[T : ClassTag] (layer: TensorModule[T])
 (implicit ev: TensorNumeric[T]) extends Container[Tensor[T], Tensor[T], T] {
```
jason-dai: Why do we need to make it a subclass of container then?

JerryYanWan: Otherwise, most of the methods should be overrided, such as parameters(), training(), evaluate(), getTimes(), etc. Might make it hard for other people to revise?

dl/src/main/scala/com/intel/analytics/bigdl/nn/TimeDistributed.scala
```diff
@@ -119,8 +114,77 @@ class TimeDistributed[T : ClassTag] ()
     layer.accGradParameters(fInput, fGradOutput)
   }
 
+  /**
+   * If the module has parameters, this will zero the accumulation of the gradients with respect
+   * to these parameters. Otherwise, it does nothing.
+   */
+  override def zeroGradParameters(): Unit = {
+    super.zeroGradParameters()
```
qiuxin2012: delete this line.

dl/src/main/scala/com/intel/analytics/bigdl/nn/TimeDistributed.scala
```diff
@@ -119,8 +114,77 @@ class TimeDistributed[T : ClassTag] ()
...
+   * Copy the useful running status from src to this.
+   *
+   * The subclass should override this method if it has some parameters besides weight and bias.
+   * Such as runningMean and runningVar of BatchNormalization.
+   *
+   * @param src source Module
+   * @return this
+   */
+  override def copyStatus(src: Module[T]): TimeDistributed.this.type = {
+    super.copyStatus(src)
```
qiuxin2012: delete this line

PR #520, opened 2017-02-28
set spark.shuffle.reduceLocality.enabled to false

yiheng:

set spark.shuffle.reduceLocality.enabled to false, which avoid many Detect multi-task run on one Executor/Container error after Spark 1.6 when data size is small

## How was this patch tested?

unit test


yiheng:
jenkins test job pass. merge it

PR #536, opened 2017-03-03
add getParametersTable for module

qiuxin2012:

This pr provides a method to access parameters in a model by moduleName and parameterName.

## How was this patch tested?

unit tests

### DIFF
--- a/dl/src/main/scala/com/intel/analytics/bigdl/nn/BatchNormalization.scala
+++ b/dl/src/main/scala/com/intel/analytics/bigdl/nn/BatchNormalization.scala
@@ -21,7 +21,7 @@ import com.intel.analytics.bigdl.Module
 import com.intel.analytics.bigdl.nn.abstractnn.TensorModule
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
 import com.intel.analytics.bigdl.tensor.{DoubleType, FloatType, Tensor}
-import com.intel.analytics.bigdl.utils.Engine
+import com.intel.analytics.bigdl.utils.{Engine, T, Table}
 import com.intel.analytics.bigdl.utils.RandomGenerator._
 
 import scala.concurrent.Future
@@ -615,6 +615,16 @@ class BatchNormalization[@specialized(Float, Double) T: ClassTag](
 
   override def parameters(): (Array[Tensor[T]], Array[Tensor[T]]) = {
     (Array(this.weight, this.bias), Array(this.gradWeight, this.gradBias))
+  }
+
+  override def getParametersTable(): Table = {
+    if (affine) {
+      T(getName() -> T("weight" -> weight, "bias" -> bias,
+        "gradWeight" -> gradWeight, "gradBias" -> gradBias,
+        "runningMean" -> runningMean, "runningVar" -> runningVar))
+    } else {
+      T(getName() -> T("runningMean" -> runningMean, "runningVar" -> runningVar))
+    }
   }
 
   override def toString(): String = {
--- a/dl/src/main/scala/com/intel/analytics/bigdl/nn/Cosine.scala
+++ b/dl/src/main/scala/com/intel/analytics/bigdl/nn/Cosine.scala
@@ -20,6 +20,7 @@ import com.intel.analytics.bigdl.nn.abstractnn.TensorModule
 import com.intel.analytics.bigdl.tensor.Tensor
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
 import com.intel.analytics.bigdl.utils.RandomGenerator._
+import com.intel.analytics.bigdl.utils.{T, Table}
 
 import scala.reflect.ClassTag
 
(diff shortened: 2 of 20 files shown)


dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -280,7 +299,7 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
   override def hashCode(): Int = {
     def getHashCode(a: Object): Int = if (a == null) 0 else a.hashCode()
     val state = Seq(output, gradInput, this.getClass)
-    state.map(getHashCode).foldLeft(0)((a, b) => 31 * a + b)
+    state.map(getHashCode).foldLeft(0)((a, b) => 31 * a + b) + super.hashCode()
```
yiheng: The breaks the define of hashCode.

dl/src/main/scala/com/intel/analytics/bigdl/nn/BatchNormalization.scala
```diff
@@ -617,6 +617,16 @@ class BatchNormalization[@specialized(Float, Double) T: ClassTag](
     (Array(this.weight, this.bias), Array(this.gradWeight, this.gradBias))
   }
 
+  override def getParametersTable(): Table = {
+    if (affine) {
+      T(getName() -> T("weight" -> weight, "bias" -> bias,
+        "gradWeight" -> gradWeight, "gradBias" -> gradBias,
```
zhichao-li: gradWeight -> weightGrad, gradBias ->biasGrad

dl/src/main/scala/com/intel/analytics/bigdl/nn/Cosine.scala
```diff
@@ -182,6 +183,10 @@ class Cosine[T: ClassTag](val inputSize : Int, val outputSize : Int)(
     (Array(this.weight), Array(this.gradWeight))
   }
 
+  override def getParametersTable(): Table = {
+    T(getName() -> T("weight" -> weight, "gradWeight" -> gradWeight))
```
zhichao-li: Move this as a default implementation to AbstractModule and only override it if needed.

dl/src/main/scala/com/intel/analytics/bigdl/nn/Linear.scala
```diff
@@ -158,6 +158,15 @@ class Linear[T: ClassTag](
     (Array(this.weight, this.bias), Array(this.gradWeight, this.gradBias))
   }
 
+  override def getParametersTable(): Table = {
```
zhichao-li: can it be the general default implementation?

dl/src/main/scala/com/intel/analytics/bigdl/nn/Cosine.scala
qiuxin2012: 80% of the modules don't have weight and bias

dl/src/test/scala/com/intel/analytics/bigdl/nn/ModuleSpec.scala
```diff
@@ -126,4 +127,40 @@ class ModuleSpec extends FlatSpec with Matchers {
     module2.eq(module) should be(false)
   }
 
+  "getName for different module" should "return different name" in {
+    val module1 = new Linear[Double](2, 3)
+    val module2 = new Linear[Double](4, 5)
+
+    module1.getName() should not be (module2.getName())
```
zhichao-li: add another checking: if the two models is equal than they share the same hashcode as we've just changed the hashcode method?

dl/src/main/scala/com/intel/analytics/bigdl/nn/Cosine.scala
zhichao-li: Or if there are no bias and no weight then we return None. Eliminate duplicate code if possible.

dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -42,6 +42,7 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
 @specialized(Float, Double) T: ClassTag](
   implicit ev: TensorNumeric[T]) extends Serializable {
 
+  val UUID = java.util.UUID.randomUUID()
```
yiheng: private?

dl/src/main/scala/com/intel/analytics/bigdl/nn/Cosine.scala
zhichao-li: @yiheng @qiuxin2012  maybe we can some how move weight or bias to an abstract class, then eliminate those duplicate *parameters() code in the following PR. 
What it want to get is simple, if bias is not null, add bias and if weight is not null add weight, but currently they are spread across many layers.

dl/src/test/scala/com/intel/analytics/bigdl/nn/ModuleSpec.scala
qiuxin2012: Added

PR #537, opened 2017-03-03
update caffeloader with the change of parameters

SeaOfOcean:

update caffeloader with the change of parameters

## How was this patch tested?

manual check


dl/src/main/scala/com/intel/analytics/bigdl/utils/CaffeLoader.scala
```diff
@@ -135,8 +136,8 @@ class CaffeLoader[T: ClassTag](prototxtPath: String, modelPath: String,
         logger.info(s"$name uses initialized parameters")
         return
       }
-      val (weight, _) = loadParameters(name, mod.parameters()._1)
```
qiuxin2012: You can try getParametersTable, to set get the parameter.

qiuxin2012: And throw an exception when the parameters don't match with caffe's model.

dl/src/main/scala/com/intel/analytics/bigdl/utils/CaffeLoader.scala
```diff
@@ -75,79 +74,82 @@ class CaffeLoader[T: ClassTag](prototxtPath: String, modelPath: String,
...
-        s"data shape in caffe is ${ caffeWeight.get.getShape() }," +
-        s" while data shape in bigdl is ${ destPara(0).size().mkString(",") }")
-    require(destPara(0).isContiguous())
-    val weightData = destPara(0).storage().array()
-    for (i <- 0 until weightList.size()) {
-      weightData(i) = ev.fromType[Float](weightList.get(i))
+    if (caffeWeight.isDefined) {
+      require(params.contains("weight"), s"$name should contain weight")
+      val caffeWeightData = caffeWeight.get.getDataList
+      val weight = params[Tensor[T]]("weight")
```
qiuxin2012: how about params don't contain weight?

dl/src/main/scala/com/intel/analytics/bigdl/utils/CaffeLoader.scala
```diff
@@ -75,79 +74,82 @@ class CaffeLoader[T: ClassTag](prototxtPath: String, modelPath: String,
...
-    if (destPara.length > 1 && destPara(1) != null) {
-      val caffeBias = getBlob(name, 1)
-      if (caffeBias.isEmpty) return (destPara(1), null)
-      val biasList = caffeBias.get.getDataList
-      require(destPara(1).nElement() == biasList.size(),
+    val caffeBias = getBlob(name, 1)
+    if (caffeBias.isDefined) {
+      require(params.contains("bias"), s"$name should contain bias")
+      val caffeBiasList = caffeBias.get.getDataList
+      val bias = params[Tensor[T]]("bias")
```
qiuxin2012: how about params don't contain bias?

dl/src/main/scala/com/intel/analytics/bigdl/utils/CaffeLoader.scala
SeaOfOcean: if caffe layer contains weight, then weight is required in bigdl module, as shown in line 90

dl/src/main/scala/com/intel/analytics/bigdl/utils/CaffeLoader.scala
SeaOfOcean: if caffe layer contains bias, then bias is required in bigdl module, as shown in line 108

PR #539, opened 2017-03-04
Structured stream udf 2.0

jenniew:

Add structured stream with udf example, using file(parquet) as data source for streaming.

## How was this patch tested?

Run this example according to README


dl/src/main/scala/com/intel/analytics/bigdl/utils/DLClassifier.scala
```diff
@@ -52,7 +52,7 @@ class DLClassifier[@specialized(Float, Double) T: ClassTag]
    */
   def setOutputCol(outputColName: String): this.type = set(outputCol, outputColName)
 
-  override def validateParams(): Unit = {
+  def validateParameters(): Unit = {
```
jason-dai: why the change?

jenniew: In spark mllib 2.1, trait Params does not have validateParams() any more. Or we may keep the DLClassifier code, and create Params wrapper in spark-version?

dl/src/main/scala/com/intel/analytics/bigdl/example/structuredStreamUdf/TextProducerParquet.scala
```diff
@@ -0,0 +1,84 @@
...
+              .save(param.destFolder)
+            count = 0
+            send_count += 1
+            Thread.sleep(interval*1000)
+          }
+
+        } catch {
+          case e: Exception => println(e)
+        }
+      }
```
yangw1234: Maybe we could use the `array.grouped` and `foreach` to simplify the while loop a little bit? Like
```
val batches = data.grouped(batchsize)
batches.foreach{ batch => ...}
```

jenniew:
This PR is merged to PR #660

PR #540, opened 2017-03-04
Fix optimizer can only run once issue

yiheng:

Optimizer will set a flag in executor but not clean. The cause there's at most one Optimizer.optimize job in the life cycle.

In the PR, the optimizer will clean the flag when an optimization is done. We also provide a public method of DistriOptimizer to allow user to clean it manually if some exception happened.

## How was this patch tested?
all old and unit tests and a new unit test

### DIFF
--- a/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
+++ b/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
@@ -511,9 +511,12 @@ class DistriOptimizer[T: ClassTag] (
   private var models: RDD[DistriOptimizer.Cache[T]] = null
 
   /**
-   * Clean some flags left on executor.
+   * Clean some internal states, so this or other optimizers can run optimize again
+   *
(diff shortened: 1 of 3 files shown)


yiheng:
@shane-huang This PR should fix the notebook optimizer can only run once issue

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -510,6 +510,17 @@ class DistriOptimizer[T: ClassTag] (
 
   private var models: RDD[DistriOptimizer.Cache[T]] = null
 
+  /**
+   * Clean some flags left on executor.
+   */
+  def cleanExecutorFlags() : Unit = {
```
jason-dai: Change to reset() or clearState()?

yiheng: changed to clearState

PR #541, opened 2017-03-04
Refine ClassNLLCriterion

yiheng:

Refine ClassNLLCriterion

## How was this patch tested?
old unit test and add new unit test

### DIFF
--- a/dl/src/main/scala/com/intel/analytics/bigdl/nn/ClassNLLCriterion.scala
+++ b/dl/src/main/scala/com/intel/analytics/bigdl/nn/ClassNLLCriterion.scala
@@ -25,10 +25,33 @@ import scala.concurrent.{Await, Future}
 import scala.reflect.ClassTag
 import com.intel.analytics.bigdl.utils.Engine
 
+/**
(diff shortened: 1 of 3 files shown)


dl/src/main/scala/com/intel/analytics/bigdl/nn/ClassNLLCriterion.scala
```diff
@@ -90,21 +147,15 @@ class ClassNLLCriterion[T: ClassTag](weights: Tensor[T] = null, sizeAverage: Boo
   override def updateGradInput(input: Tensor[T], target: Tensor[T]): Tensor[T] = {
     require(input.dim() == 1 || input.dim() == 2,
       "ClassNLLCriterion: " + ErrorInfo.constrainInputAsVectorOrBatch)
-    assert(ev.toType[Double](total_weight) > 0.0, "total weight must larger than 0")
+    require(ev.toType[Double](total_weight) > 0.0, "total weight must larger than 0")
```
jason-dai: require(..., "ClassNLLCriterion: total weight must be larger than 0")

zhangxiaoli73:
LGTM, and the user rnn example passed.

PR #542, opened 2017-03-05
Put classes in right dir for package; fix misnamed Model.scala classes ignored by gitignore

srowen:

gitignore has a rule to ignore `model.*` that was actually causing real `Model.scala` classes to be ignored. These files had no `Model` class anyway so I suggest renaming to match the primary class in the file. (In the inception case, looks like this had to be broken out into two files at least to be meaningful.)

Two classes under `org.apache.spark` namespace should also be in an `org/apache/spark` dir.

## How was this patch tested?

Existing tests; should only be a change to packaging without any functional effect.


yiheng:
Thanks for pointing this out. Should we also modify the gitignore file?
```
model.* -> model*.[0-9]*
state.* -> state*.[0-9]*
```

srowen:
@yiheng sure, looks like that shell glob syntax is supported. Are these files generated somewhere like a build or target dir? if so it may be easier to just exclude those dirs if not already. Or if they aren't generated in a build-like dir, maybe they can be.

yiheng:
They're checkpoint files in the training process. The format is model.#iteration_number for model and state.#iteration_number for training state.

Sometimes we train some models for test purpose after a build in the project folder. They're put here for prevent accident checking the checkpoint files...

yiheng:
jenkins test job pass. Merge the PR

PR #548, opened 2017-03-07
add unit test for epoch step

qiuxin2012:

read epoch form `state` table instead of `config` table, and add some unit test.

## How was this patch tested?

unit tests, and manual tests


yiheng:
should we merge the conf and state?

yiheng:
LGTM

PR #558, opened 2017-03-08
Add object to SelectTable and add more tests

psyyz10:
fix scala style

Add object to SelectTable and add more tests

## How was this patch tested?

Unit tests


zhangxiaoli73:
LGTM

Issue #559, opened 2017-03-08
It is better to define constant type in AddConstant and MultConstant layer as Double rather than T


psyyz10:
In order to be more user-friendly, it is better to define constant type in `AddConstant` and `MultConstant` layer as `Double` rather than `T`.

Because when user define model which including `AddConstant` or `MultConstant`, they may just `add(AddConstant[T](5.5))`, something like that, but not based on the tensor numeric type `T` like `add(AddConstant[T](ev.FromType[Double](5.5)))`

PR #560, opened 2017-03-08
Replace scalar type T with Double

psyyz10:

In order to be more user-friendly, it is better to define constant type in `AddConstant` and `MultConstant` layer as `Double` rather than `T`.

Because when user define model which including `AddConstant` or `MultConstant`, they may just `add(AddConstant[T](5.5))`, something like that, but not based on the tensor numeric type `T` like `add(AddConstant[T](ev.FromType[Double](5.5)))`

## How was this patch tested?

Unit tests

fixed https://github.com/intel-analytics/BigDL/issues/559


jason-dai:
Why not float instead of double?

psyyz10:
Because inside the layer, the constant will be transformed to T, no matter Double or Float. Double can be transformed to Float, but float transformed to double will loss some precision.

Additionally, if defined with float, user have to add another `f` like `add(AddConstant[T](5.5f))`.

zhangxiaoli73:
LGTM

PR #562, opened 2017-03-08
remove scala lib from dependency jar

dding3:
Remove scala library from bigdl-0.1.0-SNAPSHOT-jar-with-dependencies.jar

## How was this patch tested?
unit test and manual test

### DIFF
--- a/dl/pom.xml
+++ b/dl/pom.xml
@@ -94,7 +94,7 @@
         <dependency>
             <groupId>org.scalatest</groupId>
             <artifactId>scalatest_${scala.major.version}</artifactId>
-            <scope>${spark-scope}</scope>
+            <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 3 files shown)


dl/pom.xml
```diff
@@ -94,7 +94,7 @@
         <dependency>
             <groupId>org.scalatest</groupId>
             <artifactId>scalatest_${scala.major.version}</artifactId>
-            <scope>test</scope>
+            <scope>${spark-scope}</scope>
```
yiheng: why change it?

srowen: Agree, scalatest shouldn't be anything but test scope.
There is a particular reason it's _not_ in one case in Spark itself, but, it's specific to Spark.

dding3: OK. Change it back.

PR #563, opened 2017-03-08
add python notebook example for lenet and textclassfier

shane-huang:
add 2 python notebook examples, lenet and text classifier

### DIFF
--- a/dl/src/main/python/models/lenet/lenet5.ipynb
+++ b/dl/src/main/python/models/lenet/lenet5.ipynb
@@ -1,0 +1,449 @@
+{
+ "cells": [
+  {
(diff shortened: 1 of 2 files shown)


shane-huang:
links to rendered notebook for preview. 

- Text Classification Demo

https://nbviewer.jupyter.org/github/shane-huang/BigDL/blob/notebookdemo/dl/src/main/python/models/textclassifier/textclassifier.ipynb

- Mnist Classification using LeNet Example

https://nbviewer.jupyter.org/github/shane-huang/BigDL/blob/notebookdemo/dl/src/main/python/models/lenet/lenet5.ipynb

PR #566, opened 2017-03-09
explict call sc.stop in vgg example

dding3:
Explicit to call sc.stop in vgg. 
Similar change for other examples will be submitted in another PR

## How was this patch tested?
manual test

### DIFF
--- a/dl/src/main/scala/com/intel/analytics/bigdl/models/vgg/Train.scala
+++ b/dl/src/main/scala/com/intel/analytics/bigdl/models/vgg/Train.scala
@@ -94,6 +94,10 @@ object Train {
         .setState(state)
         .setEndWhen(Trigger.maxEpoch(param.maxEpoch))
         .optimize()
+
+      if (sc.isDefined) {
+        sc.get.stop()
+      }
     })
   }
 }


dl/src/main/scala/com/intel/analytics/bigdl/models/vgg/Train.scala
```diff
@@ -94,6 +94,10 @@ object Train {
         .setState(state)
         .setEndWhen(Trigger.maxEpoch(param.maxEpoch))
         .optimize()
+
+      if (sc.isDefined) {
```
yiheng: The program won't close the sc when it exit?

dding3: Usually it will, but I met a issue the application hung after failed to save model. I think it's a good idea to explicitly call sc.stop() to avoid application hung, just as spark example code do.

yiheng: do you test the fix? Will it really fix the issue?

dding3: Yes, I tested and it fixed the issue.

yiheng: Why not add this to all other example code?

dding3: Looks like there are lots of example and we need a quick fix for user to have a try(see google user group). I have created a issue to track add similar code for other examples(See issue 565)

PR #567, opened 2017-03-09
TimeDistributedCriterion

shenyurun:
Take each time slice of input and target, and add up all outputs of slices
For example
input.size = [B, T, D] => inputSize = [B * T, D]
target.size = [B, T] => targetSize = [B * T]
input.size = [B, T, D] => fInput.size = [B, D]
target.size = [B, T] => fTarget.size = [B]

 If sizeAverage is true, the output is averaged through time dimension

Q: I wonder if timeDim can be a parameter of TimeDistributedCriterion, or we can just confirm timeDim = 2 ?

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TimeDistributedCriterion.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TimeDistributedCriterion.scala
@@ -19,13 +19,17 @@ package com.intel.analytics.bigdl.nn
 import com.intel.analytics.bigdl.nn.abstractnn.TensorCriterion
 import com.intel.analytics.bigdl.tensor.Tensor
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+import com.intel.analytics.bigdl.utils.Engine
 
+import scala.collection.mutable.ArrayBuffer
+import scala.concurrent.Future
 import scala.reflect.ClassTag
 
 /**
  * This class is intended to support inputs with 3 or more dimensions.
  * Apply Any Provided Criterion to every temporal slice of an input.
  * @param critrn
+ * @param sizeAverage
  */
 
 class TimeDistributedCriterion[T : ClassTag](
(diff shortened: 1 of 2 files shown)


dl/src/main/scala/com/intel/analytics/bigdl/nn/TimeDistributedCriterion.scala
```diff
@@ -29,77 +29,66 @@ import scala.reflect.ClassTag
...
+     * target.size = [B, T] => fTarget.size = [B]
+     * If sizeAverage is true, the output is averaged through time dimension
      */
-    if (inputSize == null) {
-      inputSize = new Array[Int](input.size.length - 1)
+    val nstep = input.size(2)
+    var i = 1
+    while (i <= nstep) {
+      fInput = input.select(2, i)
+      fTarget = target.select(2, i)
```
JerryYanWan: define a variable timeDim = 2

dl/src/main/scala/com/intel/analytics/bigdl/nn/TimeDistributedCriterion.scala
```diff
@@ -29,77 +29,66 @@ import scala.reflect.ClassTag
...
 
   override def updateGradInput(input: Tensor[T], target: Tensor[T]): Tensor[T] = {
-    val _gradInput = critrn.updateGradInput(fInput, fTarget).toTensor[T]
-    gradInput = _gradInput.resize(input.size)
+    /**
+      * Take each time slice of input and target, and calculate gradInput of each slice
+      * If sizeAverage is true, the gradInput is also averaged through time dimension
+      */
+    gradInput.resizeAs(input)
+    gradInput.zero()
```
JerryYanWan: gradInput.resizeAs(input).zero

dl/src/main/scala/com/intel/analytics/bigdl/nn/TimeDistributedCriterion.scala
```diff
@@ -29,77 +29,66 @@ import scala.reflect.ClassTag
...
+    var _gradInput = Tensor[T]()
```
JerryYanWan: define _gradInput as global variable like fInput and fTarget. Use val instead of var

dl/src/main/scala/com/intel/analytics/bigdl/nn/TimeDistributedCriterion.scala
```diff
@@ -29,77 +29,66 @@ import scala.reflect.ClassTag
...
+    val nstep = input.size(2)
```
JerryYanWan: use timeDim instead of 2

dl/src/main/scala/com/intel/analytics/bigdl/nn/TimeDistributedCriterion.scala
```diff
@@ -29,77 +29,66 @@ import scala.reflect.ClassTag
...
   override def updateOutput(input: Tensor[T], target: Tensor[T]): T = {
     /**
+     * Take each time slice of input and target, and add up all outputs of slices
      * For example
-     * input.size = [B, T, D] => inputSize = [B * T, D]
-     * target.size = [B, T] => targetSize = [B * T]
+     * input.size = [B, T, D] => fInput.size = [B, D]
+     * target.size = [B, T] => fTarget.size = [B]
+     * If sizeAverage is true, the output is averaged through time dimension
      */
```
JerryYanWan: add requirement that input should be a 3D tensor

dl/src/main/scala/com/intel/analytics/bigdl/nn/TimeDistributedCriterion.scala
```diff
@@ -29,77 +29,66 @@ import scala.reflect.ClassTag
...
-    if (inputSize == null) {
-      inputSize = new Array[Int](input.size.length - 1)
+    val nstep = input.size(2)
+    var i = 1
+    while (i <= nstep) {
```
JerryYanWan: can this while be a parallel loop?

dl/src/main/scala/com/intel/analytics/bigdl/nn/TimeDistributedCriterion.scala
```diff
@@ -29,77 +29,66 @@ import scala.reflect.ClassTag
  * @param critrn
  */
 
-class TimeDistributedCriterion[T : ClassTag](critrn : TensorCriterion[T])
+class TimeDistributedCriterion[T : ClassTag](critrn : TensorCriterion[T],
+                                             sizeAverage: Boolean = true)
```
psyyz10: Default sizeAverage is false?

shenyurun: The default value of sizeAverage in ClassNLLCriterion (or other criterions) is true, so I kept it as true. Is there any reason to change it as false?

dl/src/main/scala/com/intel/analytics/bigdl/nn/TimeDistributedCriterion.scala
shenyurun: Input dimension is not necessarily 3D since it depends on which criterion is used. But requirement such as input.dim() >= 2 can be added anyway. Is it necessary?

dl/src/main/scala/com/intel/analytics/bigdl/nn/TimeDistributedCriterion.scala
psyyz10: Yes. For recurrent nn, it is better to user no sizeAverage criterion training for quick converge. If the default is false, the convergence is slow. Please refer to the SequencerCriterion in torch rnn library, which set default value to false.

psyyz10:
I have tested the new version of TimeDistributedCriterion, unfortunately which not works correctly

psyyz10:
I simply modified a worked version on [pull request 576](https://github.com/intel-analytics/BigDL/pull/576) for urgent request.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TimeDistributedCriterion.scala
```diff
@@ -19,88 +20,123 @@ package com.intel.analytics.bigdl.nn
...
+   * @param times
+   */
+  private def extend(times: Int): Unit = {
+    var t = cells.length
+    if (t < times) {
+      while (t < times) {
+        cells += critrn.cloneCriterion()
+          .asInstanceOf[TensorCriterion[T]]
+        t += 1
+      }
```
JerryYanWan: no need if

Issue #415, opened 2017-01-24
Missing Implementation for LSTM


whitebread:
Is there any support for LSTM networks?

jason-dai:
We have been working on that, and please check back later.

whitebread:
@jason-dai  Jason, do you think we have a broad timeframe for adding this feature?

jason-dai:
@whitebread it's currently WIP and you may take a look at #481; let us know if you have any feedback.

PR #571, opened 2017-03-09
LSTM Suport

psyyz10:

Add LSTM Suport

## How was this patch tested?
Unit tests

fixed https://github.com/intel-analytics/BigDL/issues/415


dl/src/main/scala/com/intel/analytics/bigdl/nn/LSTM.scala
```diff
@@ -0,0 +1,192 @@
...
+ * D. https://github.com/wojzaremba/lstm
+ *
+ * @param inputSize the size of each input vector
+ * @param hiddenSize Hidden unit size in the LSTM
+ * @param p p is the threshold for Dropout. If p equals 0, create
+ *          a LSTM without Dropout
+ */
+class LSTM[T : ClassTag] (
+  inputSize: Int = 4,
+  hiddenSize: Int = 3,
```
jason-dai: What's the point of giving inputSize and hiddenSize default values?

jason-dai:
We need reset, clearState, equals, hashCode, toString, etc., yes?

dl/src/main/scala/com/intel/analytics/bigdl/nn/LSTM.scala
```diff
@@ -0,0 +1,192 @@
...
+          .add(Dropout(p))
+          .add(Dropout(p)))
+        .add(ParallelTable()
+          .add(Linear(inputSize, hiddenSize))
+          .add(Linear(inputSize, hiddenSize))
+          .add(Linear(inputSize, hiddenSize))
+          .add(Linear(inputSize, hiddenSize)))
+        .add(JoinTable(2, 1))
+
+      val o2g = Sequential()
```
jason-dai: where is o2g used?

psyyz10:
Done

jason-dai:
LGTM

PR #580, opened 2017-03-13
textToLabeledSentence transformer

JerryYanWan:
Add a new Transformer,

TextToLabeledSentence: transform the text to the integer index.
The integer index of a word is got from a Dictionary.

## How was this patch tested?
Unit Test:

input a sample sentence: [x1 x2 x3]
the expected output labeledSentence should be:
input = [a1 a2 a3] targets = [a2 a3 a4]; a_i is the integer

### DIFF
--- a/dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/TextToLabeledSentence.scala
+++ b/dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/TextToLabeledSentence.scala
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2016 The BigDL Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.intel.analytics.bigdl.dataset.text
+
+import com.intel.analytics.bigdl.dataset.Transformer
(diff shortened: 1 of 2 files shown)


dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/TextToLabeledSentence.scala
```diff
@@ -0,0 +1,48 @@
...
+
+package com.intel.analytics.bigdl.dataset.text
+
+import com.intel.analytics.bigdl.dataset.Transformer
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+
+import scala.collection.Iterator
+import scala.reflect.ClassTag
+
+object TextToLabeledSentence {
```
zhangxiaoli73: add some doc for it

zhangxiaoli73: add some doc for it, especially what is the component of Array[String]

dl/src/test/scala/com/intel/analytics/bigdl/dataset/TextToLabeledSentenceSpec.scala
```diff
@@ -0,0 +1,117 @@
...
+    val sentence3 = "I think I hear them. Stand ho! Who is there?"
+    val sentence4 = "The Dr. lives in a blue-painted box."
+
+    val sentences = Array(sentence1, sentence2, sentence3, sentence4)
+
+    new PrintWriter(tmpFile) {
+      write(sentences.mkString("\n")); close
+    }
+
+    sc = new SparkContext("local[1]", "TextToLabeledSentence")
```
zhangxiaoli73: perhaps you can define sc in "before"

zhangxiaoli73:
oh, don't forget to rebase commits before checking in.

dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/TextToLabeledSentence.scala
JerryYanWan: added

PR #584, opened 2017-03-14
Fail recovery

dding3:
Add failure recovery logistic. If there is excpetion, application will rerun from last model/state snapshort.

## How was this patch tested?
manual test

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
@@ -16,14 +16,17 @@
 
 package com.intel.analytics.bigdl.optim
 
-
 import com.intel.analytics.bigdl.{Module, _}
 import com.intel.analytics.bigdl.dataset.{DistributedDataSet, MiniBatch}
-import java.io.{File, FileFilter, FilenameFilter}
+import com.intel.analytics.bigdl.nn.Module
 import com.intel.analytics.bigdl.parameters.AllReduceParameter
 import com.intel.analytics.bigdl.tensor.Tensor
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
 import com.intel.analytics.bigdl.utils._
+import java.io.{File, FileFilter, FilenameFilter}
+import java.text.SimpleDateFormat
+import java.util.Calendar
+
 import org.apache.commons.lang.exception.ExceptionUtils
 import org.apache.log4j.Logger
 import org.apache.spark.TaskContext
@@ -609,7 +612,13 @@ class DistriOptimizer[T: ClassTag] (
     models = DistriOptimizer.initThreadModels(
       model, dataset, criterion, state, nodeNumber, coresPerNode, checkSingleton, parameters)
 
-    
+    val path = if (checkpointPath.isDefined) {
+      val file = checkpointPath.get + "/" +
+        new SimpleDateFormat("yyyyMMdd_HHmmss").format(Calendar.getInstance().getTime())
+      new File(file).mkdir()
+      Some(file)
+    } else None
+
     var retryNum = 0
     val maxRetry = System.getProperty("bigdl.failure.retryTimes", "5").toInt
     var lastFailureTimestamp = System.nanoTime()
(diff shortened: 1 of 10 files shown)


dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -544,23 +548,62 @@ class DistriOptimizer[T: ClassTag] (
...
+          validationMethods,
+          checkpointTrigger,
+          checkpointPath,
+          isOverWrite
+        )
+      } catch {
+        case t: Throwable =>
+          DistriOptimizer.logger.error("Error: " + ExceptionUtils.getStackTrace(t))
+          if (checkpointPath.isDefined) {
+            if (System.nanoTime() - lastFailureTimestamp < maxRetry * 2 * 60 * 1e9) {
```
yiheng: why?

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -571,6 +614,25 @@ class DistriOptimizer[T: ClassTag] (
...
+
+  private def getLatestFile(fileName: String): String = {
+    val fl = new java.io.File(checkpointPath.get)
+    val files = fl.listFiles(new FilenameFilter {
+      override def accept(dir: File, name: String): Boolean = {
+        name.startsWith(fileName)
+      }
+    })
+
+    var lastMod = Long.MinValue
```
yiheng: what if there're some model files from other optimization with same name?

dding3: I think we don't need worry about this if they are not running in the same time as we can distinguish model files by time. If user are running more than one model at the same time and set the same checkpoint directory, what about if we create a sub directory with UUID to save the model files.

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
dding3: It's used to make sure in a long time training, each exception is able to rerun maxRetry times.  Will add the explaination later in the comments.

dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
dding3: @yiheng Any thoughts about create a sub directory with UUID in name to avoid model files with same name? If there is no concern, I will start work on it.

yiheng: what about start training time? It's should not be conflict and more meaningful than UUID

dding3: OK. Sounds good to me.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -606,24 +612,75 @@ class DistriOptimizer[T: ClassTag] (
...
+          trainSummary,
+          validationSummary,
+          isOverWrite
+        )
+      } catch {
+        case t: Throwable =>
+          DistriOptimizer.logger.error("Error: " + ExceptionUtils.getStackTrace(t))
+          if (checkpointPath.isDefined) {
+            /* To avoid retry number is used up by first few exceptions, we count time here.
+             * If exception exceeds maxRetry times in maxRetry*2 minutes, we will give up retry.
```
yiheng: can the 2 minutes be configured?

yiheng:
Can we add some unit tests for the failover logic?

spark/dl/src/test/scala/com/intel/analytics/bigdl/optim/DistriOptimizerSpec.scala
```diff
@@ -358,4 +370,23 @@ class DistriOptimizerSpec extends FlatSpec with Matchers with BeforeAndAfter {
     trainSummary.readScalar("Loss").last._2 should be (0.0f +- 1e-3f)
     trainSummary.close()
   }
+
+  "Train with MSE and SGD" should "be trained with good result with fail recovery" in {
```
yiheng: add a failure time exceed than max retry case?

yiheng: also a test case exceed maxInterval?

PR #591, opened 2017-03-14
Resolve build warnings and related code/build issues

srowen:

I hope this is useful. This PR resolves all the build warnings that can be, mostly concerning deprecated API usage. It also fixes up some related issues while there, like files in wrong dir according to its package, and some foreach/map mixup.

No worries if some or all of this isn't viewed as useful to adjust.

## How was this patch tested?

Manually built and tested with the Spark 1.6  / Scala 2.10 profile. I only get some test failures due to running out of memory, which I believe are not related.

### DIFF
--- a/dl/src/main/scala/com/intel/analytics/bigdl/models/inception/Test.scala
+++ b/dl/src/main/scala/com/intel/analytics/bigdl/models/inception/Test.scala
@@ -33,7 +33,7 @@ object Test {
   val imageSize = 224
 
   def main(args: Array[String]) {
-    testParser.parse(args, new TestParams()).map(param => {
+    testParser.parse(args, new TestParams()).foreach { param =>
       val batchSize = param.batchSize.getOrElse(128)
       val sc = Engine.init(param.nodeNumber, param.coreNumber, param.env == "spark")
         .map(conf => {
@@ -54,9 +54,7 @@ object Test {
       val model = Module.load[Float](param.model)
       val validator = Validator(model, valSet)
       val result = validator.test(Array(new Top1Accuracy[Float], new Top5Accuracy[Float]))
-      result.foreach(r => {
-        println(s"${r._2} is ${r._1}")
-      })
-    })
+      result.foreach(r => println(s"${r._2} is ${r._1}"))
+    }
   }
 }
--- a/dl/src/main/scala/com/intel/analytics/bigdl/nn/Utils.scala
+++ b/dl/src/main/scala/com/intel/analytics/bigdl/nn/Utils.scala
@@ -40,7 +40,7 @@ object Utils {
       if (v.isInstanceOf[Table]) {
         t1.update(k, zeroTableCopy(if (t1.contains(k)) t1(k) else T(), t2(k)))
       } else {
-        require(v.isInstanceOf[Tensor[T]], "Input can only consist of Tensor or Table")
+        require(v.isInstanceOf[Tensor[_]], "Input can only consist of Tensor or Table")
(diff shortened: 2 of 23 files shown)


dl/src/main/scala/com/intel/analytics/bigdl/models/inception/Test.scala
```diff
@@ -34,7 +34,7 @@ object Test {
   val imageSize = 224
 
   def main(args: Array[String]) {
-    testParser.parse(args, new TestParams()).map(param => {
+    testParser.parse(args, new TestParams()).foreach { param =>
```
srowen: Generally the issue here is that map produces a new collection (and, in some cases is only lazily evaluated) while foreach will always execute and doesn't bother to create a pointless collection of Unit

dl/src/main/scala/com/intel/analytics/bigdl/nn/Utils.scala
```diff
@@ -41,7 +41,7 @@ object Utils {
       if (v.isInstanceOf[Table]) {
         t1.update(k, zeroTableCopy(if (t1.contains(k)) t1(k) else T(), t2(k)))
       } else {
-        require(v.isInstanceOf[Tensor[T]], "Input can only consist of Tensor or Table")
+        require(v.isInstanceOf[Tensor[_]], "Input can only consist of Tensor or Table")
```
srowen: Warns because the generic type is runtime-only and can't be used with isInstanceOf

dl/src/main/scala/com/intel/analytics/bigdl/utils/Table.scala
```diff
@@ -224,7 +224,7 @@ class Table private[bigdl](
     val newState = mutable.Map[Any, Any]()
 
     while (i <= state.size) {
-      state.get(i).get match {
+      state(i) match {
```
srowen: Not strictly a warning but avoids creating an Option only to unpack it.

dl/src/test/scala/com/intel/analytics/bigdl/dataset/DataSetSpec.scala
```diff
@@ -35,7 +35,8 @@ class DataSetSpec extends FlatSpec with Matchers with BeforeAndAfter {
     val nodeNumber = 1
     val coreNumber = 1
     Engine.init(nodeNumber, coreNumber, true)
-    sc = new SparkContext("local[1]", "DataSetSpec")
+    val conf = new SparkConf().setMaster("local[1]").setAppName("DataSetSpec")
```
srowen: Several instances of using this deprecated constructor

spark-version/1.5-plus/src/main/scala/org/apache/spark/rdd/ZippedPartitionsWithLocalityRDD.scala
```diff
@@ -95,8 +94,8 @@ class ZippedPartitionsWithLocalityRDD[A: ClassTag, B: ClassTag, V: ClassTag](
 
 private[spark] class ZippedPartitionsLocalityPartition(
   idx: Int,
-  @transient indexes: Seq[Int],
-  @transient rdds: Seq[RDD[_]],
+  @transient val indexes: Seq[Int],
```
srowen: Obscure scala annotation warning, but this needs to be val

yiheng:
This looks good to me. Jenkins test pass

yiheng:
My colleague just simplify the license header of all codes. Can you rebase the code?

srowen:
Thanks @yiheng , done.

PR #598, opened 2017-03-16
Fix from_path API 

zhichao-li:
Fix from_path which can be used to load bigdl pre-trained model.

## How was this patch tested?

covered by unittest


yiheng:
I don't think we should check in binary files or train dataset files. Or people will spend too much time to down load the project if the network is not good.

zhichao-li:
@yiheng  Testing data has been reduce to an acceptable size which is around 100K

Issue #602, opened 2017-03-17
Created customer criterion throws exception 'java.lang.NoSuchMethodError: my.cool.packg.MyMSECriterion.gradInput()Lcom/intel/analytics/bigdl/nn/abstractnn/Activity;'


psyyz10:
Created customer criterion throws exception 'java.lang.NoSuchMethodError: my.cool.packg.MyMSECriterion.gradInput()Lcom/intel/analytics/bigdl/nn/abstractnn/Activity;'

PR #603, opened 2017-03-17
[fix issue 602] Fixed error of created customer criterion throwing exception 'java.lang.NoSuchMethodError

psyyz10:

Fixed error of created customer criterion throwing exception 'java.lang.NoSuchMethodError

## How was this patch tested?
Create a new separate project and import bigdl. Then create a custom criterion, putting it into a mlp.
Training works fine.

fixed https://github.com/intel-analytics/BigDL/issues/602


yiheng:
LGTM

PR #607, opened 2017-03-17
Add more nn layer

plrectco:

+ add more layers to layer.py and PythonBigDL.scala

## How was this patch tested?

- python doc test
- mvn verify

### DIFF
--- a/dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDL.scala
+++ b/dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDL.scala
@@ -243,6 +243,576 @@ class PythonBigDL[T: ClassTag](implicit ev: TensorNumeric[T]) extends Serializab
     View[T](sizes.asScala.toArray).setNumInputDims(num_input_dims)
   }
 
+  def createAbs()
+  : Abs[T] = {
+    Abs[T]()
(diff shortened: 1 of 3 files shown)


dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDL.scala
```diff
@@ -242,6 +242,664 @@ class PythonBigDL[T: ClassTag](implicit ev: TensorNumeric[T]) extends Serializab
...
+
+  def createSoftSign()
+  : SoftSign[T] = {
+    SoftSign[T]()
+  }
+
+  def createSoftmaxWithCriterion(ignoreLabel: Option[Int] = None,
+                                 normalizeMode: String = "VALID")
+  : SoftmaxWithCriterion[T] = {
+    val normM = normalizeMode match {
```
zhichao-li: throw exception if match nothing.

zhichao-li: A user friendly message, ie:  Invalid normalizeMode : $normalizeMode , we only support FULL, BATCH_SIZE and NONE for now.

zhichao-li:
Squash those commits before merge.

PR #613, opened 2017-03-19
Loadcaffe

jenniew:
Add python load caffe feature, load torch/bigdl model, also some fundamental layers.

## How was this patch tested?
add unit test.


dl/src/main/python/nn/layer.py
```diff
@@ -337,6 +392,14 @@ def __init__(self,
                                    num_input_dims)
 
 
+def build_model():
```
jason-dai: why do we have this method here?

dl/src/main/python/nn/layer.py
```diff
@@ -82,9 +98,14 @@ def test(self, val_rdd, batch_size, val_methods):
...
-    def from_path(path, bigdl_type="float"):
-        jmodel = callBigDlFunc(bigdl_type, "modelFromPath", path)
-        return Model.of(jmodel, bigdl_type)
+    def from_path(path, model_type="bigdl", bigdl_type="float"):
+        jmodel = callBigDlFunc(bigdl_type, "modelFromPath", path, model_type)
+        return Model.of(jmodel)
+
+    @staticmethod
+    def load_caffe(model, defPath, modelPath, match_all=True, bigdl_type="float"):
+        jmodel = callBigDlFunc(bigdl_type, "loadCaffe", model, defPath, modelPath, match_all)
```
jason-dai: model.value?

dl/src/main/python/nn/layer.py
jenniew: This is a utility function for loadcaffe doctest, as we already have separate loadcaffe test, I remove it.

dl/src/main/python/nn/layer.py
```diff
@@ -82,9 +82,14 @@ def test(self, val_rdd, batch_size, val_methods):
                              val_rdd, batch_size, val_methods)
 
     @staticmethod
-    def from_path(path, bigdl_type="float"):
-        jmodel = callBigDlFunc(bigdl_type, "modelFromPath", path)
-        return Model.of(jmodel, bigdl_type)
+    def from_path(path, model_type="bigdl", bigdl_type="float"):
```
jason-dai: change to:
1)load
2)load_torch
3)local_caffe
@zhichao-li

jenniew: updated.

dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDL.scala
```diff
@@ -970,10 +970,14 @@ class PythonBigDL[T: ClassTag](implicit ev: TensorNumeric[T]) extends Serializab
     testResultArray.toList.asJava
   }
 
-  def modelFromPath(path: String): AbstractModule[Activity, Activity, T] = {
+  def loadBigdl(path: String): AbstractModule[Activity, Activity, T] = {
```
jason-dai: loadBigDL

dl/src/main/python/nn/layer.py
```diff
@@ -415,11 +424,9 @@ class View(Model):
 
     def __init__(self,
                  sizes,
-                 num_input_dims=0,
```
zhichao-li: let's keep this default parameters.

dl/src/main/python/nn/layer.py
```diff
@@ -415,11 +424,9 @@ class View(Model):
...
                  bigdl_type="float"):
         super(View, self).__init__(None, bigdl_type,
-                                   sizes,
-                                   num_input_dims)
```
zhichao-li: keep this.

zhichao-li:
@jenniew let's rebase and merge it after pass the jenkins.

zhichao-li:
Squash those commits into one before merging it.

PR #633, opened 2017-03-21
add doctest for criterion.py

plrectco:

## How was this patch tested?
doctest

### DIFF
--- a/pyspark/bigdl/nn/criterion.py
+++ b/pyspark/bigdl/nn/criterion.py
@@ -29,6 +29,7 @@ if sys.version >= '3':
 
 
 class Criterion(JavaValue):
+
     def __init__(self, jvalue, bigdl_type, *args):
         self.value = jvalue if jvalue else callBigDlFunc(
             bigdl_type, JavaValue.jvm_class_constructor(self), *args)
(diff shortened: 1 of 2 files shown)


dl/src/main/python/nn/criterion.py
```diff
@@ -251,7 +241,6 @@ def __init__(self,
 
 
 class SoftmaxWithCriterion(Criterion):
-
```
zhichao-li: why delete blank line here?

dl/src/main/python/nn/criterion.py
```diff
@@ -47,21 +47,23 @@ class ClassNLLCriterion(Criterion):
     >>> classNLLCriterion = ClassNLLCriterion()
     creating: createClassNLLCriterion
     '''
-    def __init__(self, bigdl_type="float"):
-        JavaValue.__init__(self, None, bigdl_type)
+
+    def __init__(self,
```
zhichao-li: add another parameters:　sizeAverage: Boolean = true

dl/src/main/python/nn/criterion.py
plrectco: for consistency purpose

zhichao-li:
Squash and then merge it after pass the jenkins

Issue #641, opened 2017-03-22
clone() method throw exception


zhichao-li:
```
   val tensor: Tensor[Float] = Tensor.ones[Float](10)
   val cloneTensor = tensor.clone()
```

java.lang.ArrayStoreException was thrown.
java.lang.ArrayStoreException
	at java.lang.System.arraycopy(Native Method)
	at com.intel.analytics.bigdl.tensor.DenseTensor$.copy(DenseTensor.scala:2229)
	at com.intel.analytics.bigdl.tensor.DenseTensor$.newClone(DenseTensor.scala:2032)
	at com.intel.analytics.bigdl.tensor.DenseTensor.clone(DenseTensor.scala:365)

cc @zhangxiaoli73 @yiheng

yiheng:
Can't reproduce this.

PR #643, opened 2017-03-22
Add a Tenor wrapper 

zhichao-li:
1. Rename bigdl to dl and move dev folder under the test folder
2. Add a tensor wrapper: JTensor, which can easy our work when need to pass or return Tensor to/from Scala.
This class should be use Internally to save some boilerplate code as the user should operate on ndarray directly.
3.Add CrossEntropyCriterion
## How was this patch tested?
doctest


spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDL.scala
```diff
@@ -99,6 +101,23 @@ class PythonBigDL[T: ClassTag](implicit ev: TensorNumeric[T]) extends Serializab
...
   }
 
+  def toTensor(jTensor: JTensor): Tensor[T] = {
+    Tensor(jTensor.storage.asScala.toArray.asInstanceOf[Array[T]],
+      jTensor.shape.asScala.toArray)
+  }
+
+  def toJTensor(tensor: Tensor[T]): JTensor = {
+    // TODO: we should clone here, in case the underlying storage large than the tensor size.
+    val cloneTensor = tensor
```
jason-dai: Need to do a copy

zhichao-li: clone method is break at the moment. Would add it back after fixing that issue.

zhichao-li: https://github.com/intel-analytics/BigDL/issues/641

jason-dai: Don't use clone - you need to make sure it's continuous. use resizeAs() and copy() instead.

zhichao-li: clone would copy and create a new tensor for us, what it really does is exactly combining resizeAs() and copy().

PR #647, opened 2017-03-22
RNN Dropout

psyyz10:

Make RNNs support dropout

## How was this patch tested?

Unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Cell.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Cell.scala
@@ -36,6 +36,8 @@ abstract class Cell[T : ClassTag](val hiddensShape: Array[Int])
   (implicit ev: TensorNumeric[T])
   extends AbstractModule[Table, Table, T] {
 
+  var cell: AbstractModule[Activity, Activity, T]
+
   /**
    * resize the hidden parameters wrt the batch size, hiddens shapes.
    *
@@ -76,5 +78,36 @@ abstract class Cell[T : ClassTag](val hiddensShape: Array[Int])
         hidden
       }
     }
+  }
+
+  override def updateOutput(input: Table): Table = {
+    output = cell.updateOutput(input).toTable
+    output
+  }
+
+  override def updateGradInput(input: Table, gradOutput: Table): Table = {
+    gradInput = cell.updateGradInput(input, gradOutput).toTable
+    gradInput
+  }
+
+  override def accGradParameters(input: Table, gradOutput: Table,
+    scale: Double = 1.0): Unit = {
+    cell.accGradParameters(input, gradOutput, scale)
+  }
+
+  override def updateParameters(learningRate: T): Unit = {
+    cell.updateParameters(learningRate)
+  }
+
+  override def zeroGradParameters(): Unit = {
+    cell.zeroGradParameters()
+  }
+
+  override def parameters(): (Array[Tensor[T]], Array[Tensor[T]]) = {
+    cell.parameters()
+  }
+
+  override def getParametersTable(): Table = {
+    cell.getParametersTable()
   }
 }
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Cell.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Cell.scala
@@ -24,18 +24,33 @@ import com.intel.analytics.bigdl.utils.{T, Table}
 import scala.reflect.ClassTag
 
 /**
- * hidden sizes in the Cell, whose length is the number of hiddens.
- * The elements correspond to the hidden sizes of returned hiddens
+ * The Cell class is a super class of any recurrent kernels, such as
+ * [[RnnCell]], [[LSTM]] and [[GRU]]. All the kernels in a recurrent
+ * network should extend the [[Cell]] abstract class.
  *
- * E.g. For RnnCell, it should be Array(hiddenSize)
- *      For LSTM, it should be Array(hiddenSize, hiddenSize)
- *     (because each time step a LSTM return two hiddens `h` and `c` in order,
- *     which have the same size.)
+ * @param hiddensShape represents the shape of hiddens which would be
+ *                     transferred to the next recurrent time step.
+ *                     E.g. For RnnCell, it should be Array(hiddenSize)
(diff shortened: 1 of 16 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Dropout.scala
```diff
@@ -26,10 +26,14 @@ import scala.reflect.ClassTag
...
-  val initP: Double = 0.5, val inplace: Boolean = false, var scale: Boolean = true)(
+  val initP: Double = 0.5,
+  val inplace: Boolean = false,
+  var scale: Boolean = true,
+  val isLazy: Boolean = false)(
   implicit ev: TensorNumeric[T]) extends TensorModule[T] {
   private var p = initP
-  val noise = Tensor[T]()
+  var noise = Tensor[T]()
+  var flag = Array(true) // used by lazy noise
```
jason-dai: why array instead of a boolean?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Dropout.scala
```diff
@@ -49,45 +53,48 @@ class Dropout[T: ClassTag](
...
-            }
-          })
-          i += 1
+          Engine.model.sync(results)
+          flag(0) = false
         }
-
-        Engine.model.sync(results)
         this.output
       } else {
```
jason-dai: this does not seem to be correct. it should be sth like:
```scala
if (train) {
  if (!isLazy || !flag(0)) {
     noise.resizeAs(input)
     if (input.isContiguous()) {
       ...
       Engine.model.sync(results)
       this.output
     }
     else {
        ...
        this.output.cmul(noise)
     }
  }
  else {
    output
  }
} else if (!scale) {
  this.output.mul(ev.fromType[Double](1 - p))
} else {
  output
}
```

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Dropout.scala
```diff
@@ -26,10 +26,14 @@ import scala.reflect.ClassTag
 
 @SerialVersionUID(- 4636332259181125718L)
 class Dropout[T: ClassTag](
-  val initP: Double = 0.5, val inplace: Boolean = false, var scale: Boolean = true)(
+  val initP: Double = 0.5,
+  val inplace: Boolean = false,
+  var scale: Boolean = true,
+  val isLazy: Boolean = false)(
```
jason-dai: What do you mean by isLazy? Please add comments and scaladoc.

psyyz10: Documentation has been added.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Dropout.scala
psyyz10: Because, the flag need to be shared by dropouts in each time step.
A boolean is a primitive value not being able to be shared. So I used a array to make dropouts point to the same reference.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Dropout.scala
psyyz10: Yes, that was the old version not passed the dropout test with a logic problem. I uploaded the new version code just now and fixed the problem.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -108,9 +110,36 @@ class Recurrent[T : ClassTag]()
...
+
+        dropouts.append(findDropouts(c))
       }
-      c
     })
+
+    val stepLength = dropouts.length
+    for (i <- dropouts.head.indices) {
+      val head = dropouts.head.apply(i)
+      val noise = head.noise.resize(1)
```
jason-dai: why resize?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -108,9 +110,36 @@ class Recurrent[T : ClassTag]()
...
         }
+
+        dropouts.append(findDropouts(c))
       }
-      c
     })
+
+    val stepLength = dropouts.length
+    for (i <- dropouts.head.indices) {
+      val head = dropouts.head.apply(i)
```
jason-dai: head = dropouts.head(i)

psyyz10: Done

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
psyyz10: Because at start I made the noise share the same storage by `current.noise.set(noise)`. The size of noise is empty is empty, and `NullPointerException` was thrown, so I give the tensor a initial size. But I thought is it better to set the noises point to the same tensor by `current.noise = noise`. Now the resize is not necessary and I have removed it.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Cell.scala
```diff
@@ -36,6 +36,8 @@ abstract class Cell[T : ClassTag](val hiddensShape: Array[Int])
   (implicit ev: TensorNumeric[T])
   extends AbstractModule[Table, Table, T] {
 
+  var cell: AbstractModule[Activity, Activity, T]
```
jason-dai: AbstractModule[Table, Table, T]?

psyyz10: Usually cell is a Sequential. The `Sequential` extends` Container[Activity, Activity, T]`, which is not a subclass of `AbstractModule[Table, Table, T]`. So if I set cell to AbstractModule[Table, Table, T], there will be a compile error.

psyyz10: I added more scala docs for the `Cell` class and the `cell` variable to guide the user how to use it.

Issue #642, opened 2017-03-22
Refactor SoftmaxWithCriterion


zhichao-li:
None should be removed or we can pass in a List here and then map the list ->Option[Int] on java side
Check and change such behavior if there's still any. 
```
class SoftmaxWithCriterion(Criterion):
    def __init__(self,
                 normalize_mode="VALID",
                 bigdl_type="float"):
        super(SoftmaxWithCriterion, self).__init__(None, bigdl_type,
                                                  ``` None,```
                                                   normalize_mode)
```
cc @plrectco

PR #652, opened 2017-03-23
Softmax

plrectco:
delete duplicate view
update softmaxwithcriterion

## How was this patch tested?
doc test

fixed https://github.com/intel-analytics/BigDL/issues/642

### DIFF
--- a/pyspark/dl/nn/criterion.py
+++ b/pyspark/dl/nn/criterion.py
@@ -260,15 +260,18 @@ class SmoothL1CriterionWithWeights(Criterion):
 class SoftmaxWithCriterion(Criterion):
 
     '''
-    >>> softmaxWithCriterion = SoftmaxWithCriterion("VALID")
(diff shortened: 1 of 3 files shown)


pyspark/dl/nn/layer.py
```diff
@@ -235,6 +235,7 @@ def __init__(self, kw,
 
 
 class Select(Model):
+
```
zhichao-li: why adding new line here?

pyspark/dl/nn/criterion.py
```diff
@@ -260,15 +260,17 @@ def __init__(self,
 class SoftmaxWithCriterion(Criterion):
 
     '''
-    >>> softmaxWithCriterion = SoftmaxWithCriterion("VALID")
+    >>> softmaxWithCriterion = SoftmaxWithCriterion(1)
```
zhichao-li: Let's pass normalize_mode into this test as well.

PR #663, opened 2017-03-24
get back progress bar on console.

i8run:

After redirecting spark log, the progress bar will disappear on Console. This patch will get back it.

## How was this patch tested?

Manual tested on Inception V1.


yiheng:
LGTM

PR #668, opened 2017-03-27
add some scala doc and python doc

zhangxiaoli73:

add scala doc and python doc

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/image/LocalImgReader.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/image/LocalImgReader.scala
@@ -38,9 +38,9 @@ object LocalImgReader {
 }
 
 /**
- * Read BGR images from given paths. After read the image, it will resize the shorted edge to the
- * given scale to value and resize the other edge properly. It will also divide the pixel value
- * by the given normalize value.
+ * Read BGR images from local given paths. After read the image, it will resize the shorted
+ * edge to the given scale to value and resize the other edge properly. It will also divide
+ * the pixel value by the given normalize value.
  * @param scaleTo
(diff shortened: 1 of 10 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Tanh.scala
```diff
@@ -23,6 +23,10 @@ import com.intel.analytics.bigdl.tensor._
 
 import scala.reflect.ClassTag
 
+/**
+ * Applies the Tanh function element-wise to the input Tensor,
```
qiuxin2012: Please specify the definition of Tanh

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/image/LocalImgReader.scala
```diff
@@ -59,6 +59,13 @@ class LocalScaleImgReader private[dataset](scaleTo: Int, normalize: Float)
   }
 }
 
+/**
+ * Read BGR images from given paths. After read the image, it will resize the images to the given
```
qiuxin2012: I think it's a local reader.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Tanh.scala
zhangxiaoli73: ok, changed

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/image/LocalImgReader.scala
zhangxiaoli73: ok, changed

PR #674, opened 2017-03-28
add some docs

zhangxiaoli73:

add some docs for python and scala


spark/spark-version/2.0/src/main/scala/org/apache/spark/rdd/ZippedPartitionsWithLocalityRDD.scala
```diff
@@ -34,6 +34,15 @@ object ZippedPartitionsWithLocalityRDD {
   }
 }
 
+/**
+ * Prefer to zip partitions of two rdd in the same location.
+ * Remaining partitions not in same location do zip by order.
```
qiuxin2012: What do you mean here?

zhangxiaoli73: yeah, I add an example to explain it more clear

Issue #648, opened 2017-03-22
support per layer weight initializer


shane-huang:
We need to support using different weight initializers for different layesr. Initializers include
1) random number generators

yangw1234:
pr #677

shane-huang:
i moved the pretrained layer setting into another issue #724

PR #677, opened 2017-03-28
[Issue #648] add custom initializing method

yangw1234:

Add a CustomInitializer to initialize certain layer to user specified weights and bias.

## How was this patch tested?
unit tests added


jason-dai:
@shane-huang does this work for you?

zhichao-li:
How about we simplify the process a little bit:
```initMethod``` is really just a method: ```def initMethod(params: Any*): (Tensor, Tensor)```
User only need to pass in a method if he want to use their own version.
Maybe we can pick up this if have time :)

**Comment by jason-dai** [comment]
@zhichao-li How do we make it work for pyhton then?

**Comment by zhichao-li** [comment]
Maybe not a good idea:( .
but if we take a step back, seems like what this ```CustomInitializer``` can achieve is to setParameters.
An API setParameters(weights: Tensor, bias: Tensor) might be more straight forward.
and we can always inherit the InitializationMethod if want to have a customize version without the new adding.
```
case class CustomInitializer[T](weights: Tensor[T], bias: Tensor[T])
 +  extends InitializationMethod {
 +
 +  def initialize(weights: Tensor[T], bias: Tensor[T])
```

**Comment by yangw1234** [comment]
@zhichao-li @jason-dai Inheriting InitializationMethod by user does not work for the current implementaition, because the `reset()` method will do a pattern match on the InitializationMethod and adding a new one requires adding a case branch.

Anyway, I think both `setParameters` and my implementation are ok. I would go either way.

**Comment by zhichao-li** [comment]
Ideally, ```InitializationMethod``` should have a ```init()``` method, then what ```reset``` need to do is simply invoke it instead of pattern matching. @yiheng @yangw1234 we can discuss this offline though.

**Comment by yangw1234** [comment]
pr validation pass

**Comment by yangw1234** [comment]
@zhichao-li @yiheng please take a look

File: spark/dl/src/main/scala/com/intel/analytics/bigdl/models/resnet/ResNet.scala [hunk 1]
```diff
@@ -42,7 +42,7 @@ object Convolution {
      padH: Int = 0,
      nGroup: Int = 1,
      propagateBack: Boolean = true,
-     initMethod: InitializationMethod = Default,
+     initMethod: Initializer = RandomUniform,
```
zhichao-li: How about ``` initMethod``` accept a method like this``` (shape: Array[Int], dimOrder: String) => Tensor[T] ``` or  ```(shape: Array[Int]) => Tensor[T]``` as dim_ordering should be fix within BigDL.
and then ```reset() ``` can set weight base on that value.
For bias, it always be filled with zero, so here its more about how to set weight only. cc @yiheng @jason-dai

spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDL.scala
```diff
@@ -55,12 +55,13 @@ object PythonBigDL {
 
   def ofDouble(): PythonBigDL[Double] = doubleInstance
 
-  def getInitMethod(initMethod: String): InitializationMethod = {
+  def getInitMethod(initMethod: String): Initializer = {
```
zhichao-li: We can have a object Initializer{} with multiple pre-defined init-methods, and can match by name for now.

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/resnet/ResNet.scala
yangw1234: For method signature, do you mean allocating a new tensor and assign it back to the original weight instead of just setting the original weight in place. 
As for bias, filling with zero is not exactly the same as current implementation, so we need to change the reset() behavior?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Initializer.scala
```diff
@@ -0,0 +1,125 @@
...
+/**
+ * Initialization method to initialize bias and weight.
+ * The init method will be called in Module.reset()
+ */
+
+trait Initializer {
+
+  type Shape = Array[Int]
+
+  def init[T](weight: Tensor[T], bias: Option[Tensor[T]], fmt: String = "output_first")
```
yiheng: For all public method, please add javadoc

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Initializer.scala
```diff
@@ -0,0 +1,125 @@
...
+          (implicit ev: TensorNumeric[T]): Unit
+
+  protected def getFans(shape: Shape, dataFormat: String = "output_first"): (Int, Int) = {
```
yiheng: same questions for the dataFormat.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Initializer.scala
```diff
@@ -0,0 +1,125 @@
...
+}
+
+case object RandomUniform extends Initializer {
+
+  def init[T](weight: Tensor[T], bias: Option[Tensor[T]], fmt: String = "output_first")
+          (implicit ev: TensorNumeric[T]): Unit = {
+    val shape = weight.size()
+    val (fanIn, _) = getFans(shape, fmt)
+    val stdv = 1.0 / math.sqrt(fanIn)
+    weight.apply1(_ => ev.fromType[Double](RNG.uniform(-stdv, stdv)))
```
yiheng: Can we wrap this to a tensor method?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Initializer.scala
```diff
@@ -0,0 +1,125 @@
...
+    bias.foreach(_.apply1(_ => ev.fromType[Double](RNG.uniform(-stdv, stdv))))
```
yiheng: same with above

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Initializer.scala
```diff
@@ -0,0 +1,125 @@
...
+ *  [Understanding the difficulty of training deep feedforward neural networks]
+ *  (http://jmlr.org/proceedings/papers/v9/glorot10a/glorot10a.pdf)
+ */
+case object Xavier extends Initializer {
+  def init[T](weight: Tensor[T], bias: Option[Tensor[T]], fmt: String = "output_first")
+          (implicit ev: TensorNumeric[T]): Unit = {
+    val shape = weight.size()
+    val (fanIn, fanOut) = getFans(shape)
+    val stdv = math.sqrt(6.0 / (fanIn + fanOut))
+    weight.apply1(_ => ev.fromType[Double](RNG.uniform(-stdv, stdv)))
```
yiheng: same with above

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Initializer.scala
```diff
@@ -0,0 +1,125 @@
...
+    bias.foreach(_.fill(ev.fromType(0)))
```
yiheng: ev.zero. And can we wrap it with a tensor method?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Initializer.scala
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+}
+case object BilinearFiller extends Initializer {
```
yiheng: Please add a javadoc to explain the init algoirthm. Including its basic idea and paper(if possible)

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Initializer.scala
```diff
@@ -0,0 +1,125 @@
...
+    } else if (dataFormat == "output_first") {
+      (second, first)
+    } else {
+      throw new IllegalArgumentException(s"Invalid inputFormat: $dataFormat")
+    }
+    (Math.max(1, fanIn), Math.max(1, fanOut))
+  }
+}
+
+case object RandomUniform extends Initializer {
```
yiheng: Please add a javadoc to explain the algorithm

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Initializer.scala
```diff
@@ -0,0 +1,125 @@
...
+    weight.apply1(_ => ev.fromType[Double](RNG.uniform(-stdv, stdv)))
+    bias.foreach(_.fill(ev.fromType(0)))
+  }
+
+}
+case object BilinearFiller extends Initializer {
+  def init[T](weight: Tensor[T], bias: Option[Tensor[T]], fmt: String = "output_first")
+             (implicit ev: TensorNumeric[T]): Unit = {
+    val shape = weight.size()
+    require(shape.length == 5, s"SpatialFullConvolution: weight must be 5 dim, " +
```
yiheng: why SpatialFullConvolution?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Initializer.scala
```diff
@@ -0,0 +1,125 @@
...
+      s"but got ${shape.length}")
+    val kH = shape(3)
+    val kW = shape(4)
+    require(kH == kW, s"SpatialFullConvolution: Kernel $kH * $kW must be square")
```
yiheng: why SpatialFullConvolution?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Initializer.scala
```diff
@@ -0,0 +1,125 @@
...
+    val f = Math.ceil(kW / 2.0).toInt
+    val c = (2 * f - 1 - f % 2) / (2.0f * f)
+    val weightArray = weight.storage().array()
```
yiheng: If you do this, the tensor must be continuous. Why not use apply1?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Linear.scala
```diff
@@ -54,26 +51,13 @@ class Linear[T: ClassTag](
   val gradBias: Tensor[T] = if (withBias) Tensor[T](outputSize) else null
   reset()
 
-  def setInitMethod(initMethod: InitializationMethod): this.type = {
+  def setInitMethod(initMethod: Initializer): this.type = {
```
yiheng: reset after setInitMethod

yiheng: change the name to setInitializer?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Linear.scala
```diff
@@ -54,26 +51,13 @@ class Linear[T: ClassTag](
...
     this.initMethod = initMethod
     this
```
yiheng: call reset at the end of this method

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Linear.scala
```diff
@@ -34,16 +33,14 @@ import com.intel.analytics.bigdl.nn.abstractnn.TensorModule
...
- *                   [[InitializationMethod]]
+ * @param initMethod the method to initialize weight and bias, please refer to
+ *                   [[Initializer]]
  */
 @SerialVersionUID( 359656776803598943L)
 class Linear[T: ClassTag](
   inputSize: Int,
   outputSize: Int,
-  private var initMethod: InitializationMethod = Default,
+  private var initMethod: Initializer = RandomUniform,
```
yiheng: why not change the name to initializer?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Linear.scala
```diff
@@ -34,16 +33,14 @@ import com.intel.analytics.bigdl.nn.abstractnn.TensorModule
...
   withBias: Boolean = true
```
yiheng: should we move initializer to the abstract model?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RNN.scala
```diff
@@ -28,13 +27,13 @@ class RnnCell[T : ClassTag] (
   inputSize: Int = 4,
   hiddenSize: Int = 3,
   activation: TensorModule[T],
-  private var initMethod: InitializationMethod = Default)
+  private var initMethod: Initializer = Xavier)
```
yiheng: same with above

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialConvolution.scala
```diff
@@ -38,7 +38,7 @@ class SpatialConvolution[T: ClassTag](
   val padH: Int = 0, // The additional zeros added per height to the input planes.
   val nGroup: Int = 1, // Kernel group number
   val propagateBack: Boolean = true, // propagate gradient back
-  private var initMethod: InitializationMethod = Default
+  private var initMethod: Initializer = RandomUniform
```
yiheng: same with above

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialDilatedConvolution.scala
```diff
@@ -61,7 +61,7 @@ class SpatialDilatedConvolution[T: ClassTag](
   val padH: Int = 0,
   val dilationW: Int = 1,
   val dilationH: Int = 1,
-  private var initMethod: InitializationMethod = Default
+  private var initMethod: Initializer = RandomUniform
```
yiheng: same with above

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialFullConvolution.scala
```diff
@@ -75,7 +75,7 @@ class SpatialFullConvolution[A <: Activity : ClassTag, T: ClassTag](
   var adjH: Int = 0,
   val nGroup: Int = 1,
   val noBias: Boolean = false,
-  private var initMethod: InitializationMethod = Default
+  private var initMethod: Initializer = RandomUniform
```
yiheng: same with above

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialShareConvolution.scala
```diff
@@ -35,7 +35,7 @@ class SpatialShareConvolution[T: ClassTag](
   padH: Int = 0, // The additional zeros added per height to the input planes.
   nGroup: Int = 1, // Kernel group number
   propagateBack: Boolean = true, // propagate gradient back
-  private var initMethod: InitializationMethod = Default)
+  private var initMethod: Initializer = RandomUniform)
```
yiheng: same with above

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/InitializationMethod.scala
```diff
@@ -16,12 +16,86 @@
 
 package com.intel.analytics.bigdl.nn
 
+import com.intel.analytics.bigdl.tensor.Tensor
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+import com.intel.analytics.bigdl.utils.RandomGenerator.RNG
+
+sealed trait DataFormat
+
+case object OutputFirst extends DataFormat
```
yiheng: Please add some doc to examplain what is OutputFirst

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/InitializationMethod.scala
```diff
@@ -16,12 +16,86 @@
...
+
+case object InputFirst extends DataFormat
```
yiheng: Please add some doc to explain what is InputFirst

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/InitializationMethod.scala
```diff
@@ -16,12 +16,116 @@
...
+      case OutputFirst => (second, first)
+      case _ =>
+        throw new IllegalArgumentException(s"Invalid inputFormat: $dataFormat")
+    }
+    (Math.max(1, fanIn), Math.max(1, fanOut))
+  }
+
+  protected def initWithUniform[T](tensor: Tensor[T], stdv: Double)
+                                  (implicit ev: TensorNumeric[T]): Unit = {
+    tensor.apply1(_ => ev.fromType[Double](RNG.uniform(-stdv, stdv)))
```
yiheng: I notice this is used in a lot of place. Add a tensor.rand(a, b) API for it?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/InitializationMethod.scala
```diff
@@ -16,12 +16,116 @@
...
+  /**
+    * Initialize the given weight and bias.
+    *
+    * @param weight    the weight to initialize
+    * @param bias      the bias to initialize, do nothing if bias is None
+    * @param dataFormat       the data format of weight indicating the dimension order of
+    *                  the weight. "output_first" means output is in the lower dimension
+    *                  "input_first" means input is in the lower dimension.
+    */
+  def init[T](weight: Tensor[T], bias: Option[Tensor[T]], dataFormat: DataFormat = OutputFirst)
```
yiheng: what if there's only bias in the layer?

yiheng: change it to init(Tensor, Shape, DataFormat)?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Linear.scala
```diff
@@ -34,16 +33,14 @@ import com.intel.analytics.bigdl.nn.abstractnn.TensorModule
...
- *                   detailed information about `initMethod`, please refer to
+ * @param initMethod the method to initialize weight and bias, please refer to
  *                   [[InitializationMethod]]
  */
 @SerialVersionUID( 359656776803598943L)
 class Linear[T: ClassTag](
   inputSize: Int,
   outputSize: Int,
-  private var initMethod: InitializationMethod = Default,
+  private var initMethod: InitializationMethod = RandomUniform,
```
yiheng: move this out of constructor list?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RNN.scala
```diff
@@ -28,13 +27,13 @@ class RnnCell[T : ClassTag] (
   inputSize: Int = 4,
   hiddenSize: Int = 3,
   activation: TensorModule[T],
-  private var initMethod: InitializationMethod = Default)
+  private var initMethod: InitializationMethod = Xavier)
```
yiheng: why change the default init method?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/InitializationMethod.scala
```diff
@@ -16,12 +16,116 @@
...
+    *     are input and output. The 3th and 4th dimensions are the kernel width and kernel height.
+    *  3. If shape has 5 dimensions (convolutional weight), the first dimension is
+    *  group size, the 2nd and 3th dimensions are input and output, and the last two dimensions
+    *  are kernel width and kernel height.
+    *
+    * @param shape the shape of which to get the fans
+    * @param dataFormat the dimension order of the shape
+    * @return the (fanIn, fanOut) tuple
+    */
+  protected def getFans(shape: Shape, dataFormat: DataFormat = OutputFirst): (Int, Int) = {
```
yiheng: Add some unit test?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/InitializationMethod.scala
```diff
@@ -16,12 +16,116 @@
...
+
+
+/**
+ * Initializer that generates tensors with a uniform distribution.
+ *
+ * It draws samples form a uniform distribution within [-limit, limit]
+ * where "limit" is "1/sqrt(fan_in)"
+ *
+ */
+case object RandomUniform extends InitializationMethod {
```
yiheng: add unit test?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/InitializationMethod.scala
```diff
@@ -42,6 +146,44 @@ case object Default extends InitializationMethod
...
-case object BilinearFiller extends InitializationMethod
+}
+
+/**
+ * Initialize the weight with coefficients for bilinear interpolation.
+ *
+ * A common use case is with the DeconvolutionLayer acting as upsampling.
+ *
+ */
+case object BilinearFiller extends InitializationMethod {
```
yiheng: add unit test?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/InitializationMethod.scala
```diff
@@ -42,6 +146,44 @@ case object Default extends InitializationMethod
  *  [Understanding the difficulty of training deep feedforward neural networks]
  *  (http://jmlr.org/proceedings/papers/v9/glorot10a/glorot10a.pdf)
  */
-case object Xavier extends InitializationMethod
+case object Xavier extends InitializationMethod {
```
yiheng: add unit test?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RNN.scala
yangw1234: Because the original "default" behavior is Xavier, not random uniform

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/InitializationMethod.scala
yangw1234: In this way, each layer will have to accept two initializers, one for weight and one for bias. Actually, in keras, they have two initializers for each layer. So should we do this? @yiheng

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RNN.scala
yiheng: do you mean the original code is wrong?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/InitializationMethod.scala
yiheng: somes layer with weight and bias should accept two initMethod. some layers with weight should accept one initMethod and with bias only should accept one initMethod

yiheng: why not do that?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RNN.scala
yangw1234: I do not know what do you mean by "wrong". But this is only a change of name not a change of behavior. You should checkout the original code. In the reset method of RNN, if it matches the "Default" initMethod, it does a Xavier initialization. In this sense, Xavier is the default initialization method of RNN.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/InitializationMethod.scala
yangw1234: ok, great

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RNN.scala
yiheng: Oh. I misunderstanding the original code.

yangw1234:
too many conflicts and comments. I will close this and open another one.

Issue #634, opened 2017-03-21
support L1 regularization


shane-huang:
L2 is supported thru weight decay.  L1 is not supported right now.

shane-huang:
seems Zhang, Yao's work of #648 covers this issue.

yangw1234:
Zhang, Yao's work is to support L1/L2 regularizer **per layer**, but to do **global** regularization we still have to (or at least is eazier to) use the "weightDecay" in SGD/Adagrad, (which does not support L1), right? @shane-huang @psyyz10

yiheng:
I think it's not conflict.

PR #680, opened 2017-03-29
Experimental  Implemation of L1 regularization support on SGD and Adagrad

yangw1234:

support L1 regularization in SGD and Adagrad

## How was this patch tested?

add unit test

fixed https://github.com/intel-analytics/BigDL/issues/634

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Adagrad.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Adagrad.scala
@@ -40,6 +40,8 @@ class Adagrad[@specialized(Float, Double) T: ClassTag](implicit ev: TensorNumeri
    * @param config    a table with configuration parameters for the optimizer
    *                  config("learningRate") : learning rate
    *                  config("learningRateDecay") : learning rate decay
+   *                  config("weightDecay") : weight decay rate
+   *                  config("l1Regularizer") : l1 regluarization rate
    * @param state     a table describing the state of the optimizer; after each call the state
    *                  is modified
    *                  state("paramVariance") : vector of temporal variances of parameters
@@ -55,19 +57,22 @@ class Adagrad[@specialized(Float, Double) T: ClassTag](implicit ev: TensorNumeri
     val lrd = _config.get[Double]("learningRateDecay").getOrElse(0.0)
     val nevals = _state.get[Int]("evalCounter").getOrElse(0)
     val wd = config.get[Double]("weightDecay").getOrElse(0.0)
-    val l1wd = config.get[Double]("l1WeightDecay").getOrElse(0.0)
-    val l2wd = config.get[Double]("l2WeightDecay").getOrElse(wd)
-    val weightsSignBuffer = _state.get[Tensor[T]]("weightSigns").getOrElse(Tensor(parameter.size()))
-    _state.update("weightSigns", weightsSignBuffer)
+    val l1wd = config.get[Double]("l1Regularizer").getOrElse(0.0)
+
 
     val (fx, dfdx) = feval(parameter)
 
-    val l1Regularizer = weightsSignBuffer.copy(parameter).sign()
-    val l2Regularizer = parameter
+    if (wd != 0) {
+      dfdx.add(ev.fromType[Double](wd), parameter)
+    }
 
-    if (l1wd != 0 || l2wd != 0) {
-      dfdx.add(ev.fromType[Double](l2wd), l2Regularizer)
-        .add(ev.fromType[Double](l1wd), l1Regularizer)
+    if (l1wd != 0) {
+      val weightsSignBuffer = _state.get[Tensor[T]]("weightSigns").getOrElse{
+        val buffer = Tensor(parameter.size())
+        _state.update("weightSigns", buffer)
+        buffer
+      }
+      dfdx.add(ev.fromType[Double](l1wd), weightsSignBuffer.copy(parameter).sign())
     }
 
     val clr = lr / (1 + nevals * lrd)
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Adagrad.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Adagrad.scala
@@ -55,11 +55,19 @@ class Adagrad[@specialized(Float, Double) T: ClassTag](implicit ev: TensorNumeri
     val lrd = _config.get[Double]("learningRateDecay").getOrElse(0.0)
     val nevals = _state.get[Int]("evalCounter").getOrElse(0)
     val wd = config.get[Double]("weightDecay").getOrElse(0.0)
+    val l1wd = config.get[Double]("l1WeightDecay").getOrElse(0.0)
+    val l2wd = config.get[Double]("l2WeightDecay").getOrElse(wd)
+    val weightsSignBuffer = _state.get[Tensor[T]]("weightSigns").getOrElse(Tensor(parameter.size()))
+    _state.update("weightSigns", weightsSignBuffer)
 
     val (fx, dfdx) = feval(parameter)
 
-    if (wd != 0) {
-      dfdx.add(ev.fromType[Double](wd), parameter)
+    val l1Regularizer = weightsSignBuffer.copy(parameter).sign()
+    val l2Regularizer = parameter
+
+    if (l1wd != 0 || l2wd != 0) {
+      dfdx.add(ev.fromType[Double](l2wd), l2Regularizer)
+        .add(ev.fromType[Double](l1wd), l1Regularizer)
     }
 
     val clr = lr / (1 + nevals * lrd)
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/SGD.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/SGD.scala
@@ -57,27 +57,47 @@ class SGD[@specialized(Float, Double) T: ClassTag](implicit ev: TensorNumeric[T]
     lrSchedule.updateHyperParameter(config, _state)
 
     val wd = config.get[Double]("weightDecay").getOrElse(0.0)
+    val l1wd = config.get[Double]("l1WeightDecay").getOrElse(0.0)
+    val l2wd = config.get[Double]("l2WeightDecay").getOrElse(wd)
     val mom = config.get[Double]("momentum").getOrElse(0.0)
     val damp = config.get[Double]("dampening").getOrElse(mom)
     val nesterov = config.get[Boolean]("nesterov").getOrElse(false)
-    val lrs = config.get[Tensor[T]]("learningRates").getOrElse(null)
-    val wds = config.get[Tensor[T]]("weightDecays").getOrElse(null)
+    val lrs = config.get[Tensor[T]]("learningRates")
+    val wds = config.get[Tensor[T]]("l2weightDecays")
+    val l2wds = if (wds.isDefined) wds else config.get[Tensor[T]]("weightDecays")
+    val l1wds = config.get[Tensor[T]]("l1WeightDecays")
+    val weightsSignBuffer = _state.get[Tensor[T]]("weightSigns").getOrElse({
+      val buffer = Tensor[T](x.size())
+      _state("weightSigns") = buffer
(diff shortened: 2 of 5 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/OptimMethod.scala
```diff
@@ -63,6 +63,17 @@ trait OptimMethod[@specialized(Float, Double) T] extends Serializable {
    * @param config a table contains the hyper parameter.
    */
   def getHyperParameter(config: Table): String = ""
+
+  protected def getRegularizer(config: Table, weights: Tensor[T], weightsSign: Tensor[T]) = {
+    val rType = config.get[String]("RegularizationType").getOrElse("L2")
+    if (rType == "L1") {
+      weightsSign.copy(weights).sign()
```
psyyz10: why copy weigths to weightsSign rather than return weights.sign() ?

yangw1234: Current DenseTensor `.sign()` implementation alter the caller in place instead of  return a new one.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/OptimMethod.scala
```diff
@@ -63,6 +63,17 @@ trait OptimMethod[@specialized(Float, Double) T] extends Serializable {
    * @param config a table contains the hyper parameter.
    */
   def getHyperParameter(config: Table): String = ""
+
+  protected def getRegularizer(config: Table, weights: Tensor[T], weightsSign: Tensor[T]) = {
+    val rType = config.get[String]("RegularizationType").getOrElse("L2")
```
yangw1234: That's actually a good idea. Treat "weightDecay" as an alias of "l2WeightDecay" to avoid the API break?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/OptimMethod.scala
yangw1234: also avoid allocating a new tensor each time

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Adagrad.scala
```diff
@@ -55,11 +55,19 @@ class Adagrad[@specialized(Float, Double) T: ClassTag](implicit ev: TensorNumeri
     val lrd = _config.get[Double]("learningRateDecay").getOrElse(0.0)
     val nevals = _state.get[Int]("evalCounter").getOrElse(0)
     val wd = config.get[Double]("weightDecay").getOrElse(0.0)
+    val l1wd = config.get[Double]("l1WeightDecay").getOrElse(0.0)
+    val l2wd = config.get[Double]("l2WeightDecay").getOrElse(wd)
```
yiheng: This is confusing. What if user specify different value for weightDecay and l2WeightDecay?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Adagrad.scala
```diff
@@ -55,11 +55,19 @@ class Adagrad[@specialized(Float, Double) T: ClassTag](implicit ev: TensorNumeri
...
+    val weightsSignBuffer = _state.get[Tensor[T]]("weightSigns").getOrElse(Tensor(parameter.size()))
```
yiheng: What's the weightsSignBuffer for? Do we always need it? It is bad for perf if the parameter is very large.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Adagrad.scala
```diff
@@ -55,11 +55,19 @@ class Adagrad[@specialized(Float, Double) T: ClassTag](implicit ev: TensorNumeri
...
 
     val (fx, dfdx) = feval(parameter)
 
-    if (wd != 0) {
-      dfdx.add(ev.fromType[Double](wd), parameter)
+    val l1Regularizer = weightsSignBuffer.copy(parameter).sign()
+    val l2Regularizer = parameter
+
+    if (l1wd != 0 || l2wd != 0) {
+      dfdx.add(ev.fromType[Double](l2wd), l2Regularizer)
```
yiheng: You should follow the old way. I think it's bad for memory and cache. Do you do any perf test compare between the old way and the vector add way?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Adagrad.scala
```diff
@@ -55,11 +55,19 @@ class Adagrad[@specialized(Float, Double) T: ClassTag](implicit ev: TensorNumeri
...
+    _state.update("weightSigns", weightsSignBuffer)
 
     val (fx, dfdx) = feval(parameter)
 
-    if (wd != 0) {
-      dfdx.add(ev.fromType[Double](wd), parameter)
+    val l1Regularizer = weightsSignBuffer.copy(parameter).sign()
+    val l2Regularizer = parameter
+
+    if (l1wd != 0 || l2wd != 0) {
```
yiheng: if l1wd/l2wd is zero. You can skip the l1/l2 weigth decay. But here you always do it.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Adagrad.scala
yiheng: Is it right that user can do l1 and l2 regular at the same time?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Adagrad.scala
```diff
@@ -55,11 +55,19 @@ class Adagrad[@specialized(Float, Double) T: ClassTag](implicit ev: TensorNumeri
     val lrd = _config.get[Double]("learningRateDecay").getOrElse(0.0)
     val nevals = _state.get[Int]("evalCounter").getOrElse(0)
     val wd = config.get[Double]("weightDecay").getOrElse(0.0)
+    val l1wd = config.get[Double]("l1WeightDecay").getOrElse(0.0)
+    val l2wd = config.get[Double]("l2WeightDecay").getOrElse(wd)
+    val weightsSignBuffer = _state.get[Tensor[T]]("weightSigns").getOrElse(Tensor(parameter.size()))
+    _state.update("weightSigns", weightsSignBuffer)
```
yiheng: change the clear history?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/SGD.scala
```diff
@@ -57,27 +57,47 @@ class SGD[@specialized(Float, Double) T: ClassTag](implicit ev: TensorNumeric[T]
...
 
     val wd = config.get[Double]("weightDecay").getOrElse(0.0)
+    val l1wd = config.get[Double]("l1WeightDecay").getOrElse(0.0)
+    val l2wd = config.get[Double]("l2WeightDecay").getOrElse(wd)
     val mom = config.get[Double]("momentum").getOrElse(0.0)
     val damp = config.get[Double]("dampening").getOrElse(mom)
     val nesterov = config.get[Boolean]("nesterov").getOrElse(false)
-    val lrs = config.get[Tensor[T]]("learningRates").getOrElse(null)
-    val wds = config.get[Tensor[T]]("weightDecays").getOrElse(null)
+    val lrs = config.get[Tensor[T]]("learningRates")
```
yiheng: why remove the getOrElse?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/SGD.scala
```diff
@@ -57,27 +57,47 @@ class SGD[@specialized(Float, Double) T: ClassTag](implicit ev: TensorNumeric[T]
...
+    val wds = config.get[Tensor[T]]("l2weightDecays")
```
yiheng: why remove the getOrElse?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/SGD.scala
```diff
@@ -57,27 +57,47 @@ class SGD[@specialized(Float, Double) T: ClassTag](implicit ev: TensorNumeric[T]
...
+    val l2wds = if (wds.isDefined) wds else config.get[Tensor[T]]("weightDecays")
+    val l1wds = config.get[Tensor[T]]("l1WeightDecays")
+    val weightsSignBuffer = _state.get[Tensor[T]]("weightSigns").getOrElse({
```
yiheng: same with adagrad question

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/SGD.scala
```diff
@@ -57,27 +57,47 @@ class SGD[@specialized(Float, Double) T: ClassTag](implicit ev: TensorNumeric[T]
...
+      val buffer = Tensor[T](x.size())
+      _state("weightSigns") = buffer
+      buffer
+    })
+    val decayParameters = _state.get[Tensor[T]]("decayParameters").getOrElse({
```
yiheng: same with adagrad question

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/SGD.scala
```diff
@@ -129,8 +149,8 @@ class SGD[@specialized(Float, Double) T: ClassTag](implicit ev: TensorNumeric[T]
     val mom = config.get[Double]("momentum").getOrElse(0.0)
     val damp = config.get[Double]("dampening").getOrElse(mom)
     val nesterov = config.get[Boolean]("nesterov").getOrElse(false)
```
yiheng: you should add your new hp, right?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/SGD.scala
```diff
@@ -57,27 +57,47 @@ class SGD[@specialized(Float, Double) T: ClassTag](implicit ev: TensorNumeric[T]
...
-        val DP = Tensor[T]().resizeAs(dfdx)
-        _state("decayParameters") = DP
-        DP
-      })
-      decayParameters.copy(wds).cmul(x)
-      dfdx.add(decayParameters)
+    val l1Regularizer = weightsSignBuffer.copy(x).sign()
+    val l2Regularizer = x
+
+    if (l1wd != 0 || l2wd != 0) {
```
yiheng: same question with adagrad

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Adagrad.scala
```diff
@@ -55,11 +55,19 @@ class Adagrad[@specialized(Float, Double) T: ClassTag](implicit ev: TensorNumeri
     val lrd = _config.get[Double]("learningRateDecay").getOrElse(0.0)
     val nevals = _state.get[Int]("evalCounter").getOrElse(0)
     val wd = config.get[Double]("weightDecay").getOrElse(0.0)
+    val l1wd = config.get[Double]("l1WeightDecay").getOrElse(0.0)
```
yiheng: And add the new hp to the javadoc?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/SGD.scala
```diff
@@ -57,27 +57,47 @@ class SGD[@specialized(Float, Double) T: ClassTag](implicit ev: TensorNumeric[T]
     lrSchedule.updateHyperParameter(config, _state)
 
     val wd = config.get[Double]("weightDecay").getOrElse(0.0)
+    val l1wd = config.get[Double]("l1WeightDecay").getOrElse(0.0)
```
yiheng: And add the new hp to the javadoc?

yangw1234:
Test performance on lenet and vgg net on desktop. No performance degradation found.

yiheng:
It's duplicated with #714. Should we do this on model, so we needn't add the regularization for each of the optim method.

PR #681, opened 2017-03-29
add some scala Doc

qiuxin2012:

add some scala doc

## How was this patch tested?

unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
@@ -65,6 +65,26 @@ object DistriOptimizer {
     localMethods: Array[Option[Array[ValidationMethod[T]]]]
   )
 
+  /**
+   * Train the model.
+   *
+   * @param dataset train dataset
+   * @param coresPerNode cores per node
+   * @param state state table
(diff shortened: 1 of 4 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -400,6 +439,20 @@ object DistriOptimizer {
       }
   }
 
+  /**
+   * Init engine and cache model on worker nodes.
```
zhangxiaoli73: not just cache model

spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDL.scala
```diff
@@ -35,6 +35,14 @@ import scala.collection.JavaConverters._
 import scala.language.existentials
 import scala.reflect.ClassTag
 
+/**
+ * [[com.intel.analytics.bigdl.dataset.Sample]] for python.
+ * @param features
+ * @param label
+ * @param featuresShape
+ * @param labelShape
+ * @param bigdlType
```
zhangxiaoli73: please explain the bigdlType. It doesn't exist in scala Sample.

zhangxiaoli73:
LGTM

PR #686, opened 2017-03-29
add python doc

SeaOfOcean:

add some python doc


pyspark/dl/nn/layer.py
```diff
@@ -1322,6 +1371,12 @@ def __init__(self,
 class Power(Model):
 
     '''
+    Apply an element-wise power operation with scale and shift.
+    f(x) = (shift + scale    x)^power^
```
psyyz10: f(x) = (shift + scale * x)^power^

PR #688, opened 2017-03-29
Use Iterator.single(elem) instead of Iterator(elem)

yangw1234:

Use Iterator.single(elem), which is equivalent but more efficient than Iterator(elem).
(According to scala doc [https://www.scala-lang.org/api/current/scala/collection/Iterator$.html](https://www.scala-lang.org/api/current/scala/collection/Iterator$.html) )

## How was this patch tested?

existing tests


yiheng:
LGTM

@dding3 Any comments?

dding3:
LGTM

Issue #659, opened 2017-03-24
spark.executor.extraJavaOptions only support one option when submitting spark job with bigdl.sh


SeaOfOcean:
when I use 
```
dist/bin/bigdl.sh -- spark-submit \
--conf 'spark.executor.extraJavaOptions=-XX:+PrintGCTimeStamps -XX:+PrintGCDetails'
```

only the first jvm option is recognized

current work around is to first source bigdl.sh, then submit with the following command
```
source bigdl.sh
spark-submit \
--conf 'spark.executor.extraJavaOptions=-XX:+PrintGCTimeStamps -XX:+PrintGCDetails'
```

PR #695, opened 2017-03-30
[Fix 659] Add quto when passed argument token contains space in bigdl.sh

yiheng:

Add quto when passed argument token contains space in bigdl.sh

## How was this patch tested?

manual test, unit test

fixed https://github.com/intel-analytics/BigDL/issues/659


SeaOfOcean:
manual test is ok, LGTM

PR #696, opened 2017-03-30
Add python Docs, scala docs and fix a example

shane-huang:
Add python Docs and scala


shane-huang:
PR validation passed

qiuxin2012:
LGTM

Issue #690, opened 2017-03-29
Looks like Reshape has different constructor parameters in scala and python


yiheng:
@zhichao-li 

Python
```python
class Reshape(Model):
    '''
    >>> reshape = Reshape([1, 28, 28])
    creating: createReshape
    :param size: the reshape size
    '''

    def __init__(self, size, bigdl_type="float"):
        super(Reshape, self).__init__(None, bigdl_type, size)
```

scala:
```scala
class Reshape[@specialized(Float, Double) T: ClassTag](
  size: Array[Int], var batchMode: Option[Boolean] = None)(
```

PR #697, opened 2017-03-30
update Reshape

plrectco:

update Reshape API

## How was this patch tested?
unit test

fixed https://github.com/intel-analytics/BigDL/issues/690


yiheng:
No need to change Reshape.apply in the scala side? what about the jenkins test result?

plrectco:
The Reshape.apply is fine, and the Jenkins has passed

PR #703, opened 2017-03-30
Add some docs

shenyurun:
Add some docs for the following classes:

class LabeledSentenceToSample[T: ClassTag](vocabLength: Int,
class Lighting extends Transformer[LabeledBGRImage, LabeledBGRImage] 
class LocalResizeImgReader private[dataset](resizeW: Int, resizeH: Int, normalize: Float)
class LocalResizeImgWithName private[dataset](resizeW: Int, resizeH: Int, normalize: Float)
class LocalValidator[T] private[optim](model: Module[T], dataSet: LocalDataSet[MiniBatch[T]])
class LogSoftMax[T: ClassTag](
class Log[T: ClassTag] (implicit ev: TensorNumeric[T])
class LossResult(private var loss: Float, private var count: Int)
class Loss[@specialized(Float, Double)T: ClassTag]()
class Metrics extends Serializable
class MSECriterion[@specialized(Float, Double) T: ClassTag]
class PythonBigDL[T: ClassTag](implicit ev: TensorNumeric[T]) extends Serializable
class RandomGenerator() {
class Recurrent[T : ClassTag]()
class ReLU6[T: ClassTag](inplace: Boolean = false)
class ReLU[T: ClassTag](ip: Boolean = false)(
class RnnCell[T : ClassTag] (

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/image/Lighting.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/image/Lighting.scala
@@ -34,10 +34,6 @@ object Lighting {
  * Krizhevsky et al. proposed fancy PCA when training the famous Alex-Net in 2012
  * Fancy PCA alters the intensities of the RGB channels in training images
  * For each training image, add the corresponding quantity to each RGB image pixel
- *
- * alphastd: the standard deviation of alpha
- * eigval: eigenvalues of the 3x3 covariance matrix of RGB pixel values
- * eigvec: eigenvectors of the 3x3 covariance matrix of RGB pixel values
  */
 class Lighting extends Transformer[LabeledBGRImage, LabeledBGRImage] {
   val alphastd = 0.1f
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/image/Lighting.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/image/Lighting.scala
@@ -28,6 +28,17 @@ object Lighting {
   }
 }
 
+/**
+ * Lighting noise for data augmentation
+ *
+ * Krizhevsky et al. proposed fancy PCA when training the famous Alex-Net in 2012
+ * Fancy PCA alters the intensities of the RGB channels in training images
+ * For each training image, add the corresponding quantity to each RGB image pixel
+ *
+ * alphastd: the standard deviation of alpha
+ * eigval: eigenvalues of the 3x3 covariance matrix of RGB pixel values
+ * eigvec: eigenvectors of the 3x3 covariance matrix of RGB pixel values
+ */
 class Lighting extends Transformer[LabeledBGRImage, LabeledBGRImage] {
   val alphastd = 0.1f
   val eigval = Tensor[Float](Storage(Array(0.2175f, 0.0188f, 0.0045f)), 1, Array(3))
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Metrics.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Metrics.scala
@@ -22,7 +22,11 @@ import org.apache.spark.{Accumulable, Accumulator, SparkContext}
 import scala.collection.mutable.{ArrayBuffer, Map}
 
 /**
- * Set metrics of local models on each executor
+ * Performance metrics for the training process.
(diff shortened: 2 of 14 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/image/Lighting.scala
```diff
@@ -28,6 +28,17 @@ object Lighting {
...
 }
 
+/**
+ * Lighting noise for data augmentation
+ *
+ * Krizhevsky et al. proposed fancy PCA when training the famous Alex-Net in 2012
+ * Fancy PCA alters the intensities of the RGB channels in training images
+ * For each training image, add the corresponding quantity to each RGB image pixel
+ *
+ * alphastd: the standard deviation of alpha
```
yiheng: please follow the Javadoc format for the parameter

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/image/Lighting.scala
```diff
@@ -28,6 +28,17 @@ object Lighting {
...
+ * eigval: eigenvalues of the 3x3 covariance matrix of RGB pixel values
```
yiheng: please follow the Javadoc format for the parameter

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/image/Lighting.scala
```diff
@@ -28,6 +28,17 @@ object Lighting {
...
+ * eigvec: eigenvectors of the 3x3 covariance matrix of RGB pixel values
```
yiheng: please follow the Javadoc format for the parameter

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Metrics.scala
```diff
@@ -21,6 +21,9 @@ import org.apache.spark.{Accumulable, Accumulator, SparkContext}
 
 import scala.collection.mutable.{ArrayBuffer, Map}
 
+/**
+ * Set metrics of local models on each executor
```
yiheng: Performance metrics for the training process. Beyond collect local metrics(e.g. throughput) in driver node, it can also be used to collect distributed metrics(e.g. time of some steps among the workers). The is useful for performance analysis.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/ValidationMethod.scala
```diff
@@ -172,6 +172,11 @@ class Top5Accuracy[T] extends ValidationMethod[T] {
   override def format(): String = "Top5Accuracy"
 }
 
+/**
+ * Return loss result as a subclass of [[ValidationResult]] module
```
yiheng: Use loss as a validation result.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/ValidationMethod.scala
```diff
@@ -214,6 +219,14 @@ class LossResult(private var loss: Float, private var count: Int)
   }
 }
 
+/**
+ * Calculate loss of output with respect to target
```
yiheng: This evaluation method is calculate loss of output with respect to target.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/RandomGenerator.scala
```diff
@@ -46,6 +46,10 @@ object RandomGenerator {
   }
 }
 
+/**
+ * A non-global random generator that carries its own state
```
yiheng: A mersenne twister based fake random number generator. Please refer https://en.wikipedia.org/wiki/Mersenne_Twister. Note that it has its own state so it is not thread safe. So you should use RandomGenerator.RNG to get a thread local instance to use. That's thread-safe.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/RandomGenerator.scala
```diff
@@ -46,6 +46,10 @@ object RandomGenerator {
...
+ * It can be used to generate random numbers for shuffling the data
+ */
 class RandomGenerator() {
```
yiheng: by the way, can you make it constructor private?

yiheng:
LGTM.

Issue #704, opened 2017-03-30
ModelBroadcast does not broadcast correct weight 


yangw1234:
When the model is loaded from a pre-trained model file, ModelBroadcast.broadcast will accidently reset all the weight storage, thus resulting in broadcasting the wrong weight.

PR #705, opened 2017-03-30
[Issue #704] fix the bug of modelbroadcast when all weights and biases share the one storage

SeaOfOcean:

fix the bug of modelbroadcast when all weights and biases share the one storage

## How was this patch tested?

unit test

fixed https://github.com/intel-analytics/BigDL/issues/704

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
@@ -76,8 +76,10 @@ class ModelBroadcast[T: ClassTag](implicit ev: TensorNumeric[T]) extends Seriali
     }
     i = 0
     while (i < parameters._1.length) {
-      if (parameters._2(i) != null) {
+      if (parameters._1(i) != null) {
         parameters._1(i).set()
+      }
+      if (parameters._2(i) != null) {
         parameters._2(i).set()
       }
       i += 1
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
@@ -74,10 +74,10 @@ class ModelBroadcast[T: ClassTag](implicit ev: TensorNumeric[T]) extends Seriali
       }
       i += 1
     }
+    i = 0
     while (i < parameters._1.length) {
-      if (parameters._1(i) != null) {
-        val wb = parameters._1(i)
-        wb.set()
+      if (parameters._2(i) != null) {
+        parameters._1(i).set()
         parameters._2(i).set()
       }
       i += 1
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
@@ -71,6 +71,12 @@ class ModelBroadcast[T: ClassTag](implicit ev: TensorNumeric[T]) extends Seriali
         val wb = parameters._1(i)
         weightsBias(i) = Tensor[T](Storage(wb.storage().array()),
           wb.storageOffset(), wb.size(), wb.stride())
+      }
+      i += 1
+    }
+    while (i < parameters._1.length) {
+      if (parameters._1(i) != null) {
+        val wb = parameters._1(i)
         wb.set()
         parameters._2(i).set()
       }
--- a/spark/dl/src/test/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcastSpec.scala
+++ b/spark/dl/src/test/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcastSpec.scala
@@ -25,11 +25,27 @@ class ModelBroadcastSpec extends FlatSpec with Matchers {
   Logger.getLogger("org").setLevel(Level.WARN)
   Logger.getLogger("akka").setLevel(Level.WARN)
 
-  val sc = new SparkContext(new SparkConf().setMaster("local[1]").setAppName("ModelBroadcast"))
-  val model = LeNet5(10)
+  "model broadcast" should "work properly" in {
+    val sc = new SparkContext(new SparkConf().setMaster("local[1]").setAppName("ModelBroadcast"))
 
-  val modelBroadCast = ModelBroadcast[Float].broadcast(sc, model)
-  modelBroadCast.value().toString should be(model.toString)
-  modelBroadCast.value().parameters()._1 should be(model.parameters()._1)
-  sc.stop()
+    val model = LeNet5(10)
+
+    val modelBroadCast = ModelBroadcast[Float].broadcast(sc, model)
+    modelBroadCast.value().toString should be(model.toString)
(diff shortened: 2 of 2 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
```diff
@@ -71,6 +71,12 @@ class ModelBroadcast[T: ClassTag](implicit ev: TensorNumeric[T]) extends Seriali
         val wb = parameters._1(i)
         weightsBias(i) = Tensor[T](Storage(wb.storage().array()),
           wb.storageOffset(), wb.size(), wb.stride())
+      }
+      i += 1
+    }
+    while (i < parameters._1.length) {
```
zhangxiaoli73: i need init again.

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
```diff
@@ -71,6 +71,12 @@ class ModelBroadcast[T: ClassTag](implicit ev: TensorNumeric[T]) extends Seriali
...
+      if (parameters._1(i) != null) {
+        val wb = parameters._1(i)
```
zhangxiaoli73: perhaps parameters._1(i).set() is ok, wb is not necessary

spark/dl/src/test/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcastSpec.scala
```diff
@@ -25,11 +25,27 @@ class ModelBroadcastSpec extends FlatSpec with Matchers {
...
-  val model = LeNet5(10)
+  "model broadcast" should "work properly" in {
+    val sc = new SparkContext(new SparkConf().setMaster("local[1]").setAppName("ModelBroadcast"))
+
+    val model = LeNet5(10)
+
+    val modelBroadCast = ModelBroadcast[Float].broadcast(sc, model)
+    modelBroadCast.value().toString should be(model.toString)
+    modelBroadCast.value().parameters()._1 should be(model.parameters()._1)
+    sc.stop()
```
yiheng: move the sc.stop to
after {
}

to make sure it won't affect other test result when it fail

spark/dl/src/test/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcastSpec.scala
```diff
@@ -25,11 +25,27 @@ class ModelBroadcastSpec extends FlatSpec with Matchers {
...
+  "model broadcast with getParameters" should "work properly" in {
+    val sc = new SparkContext(new SparkConf().setMaster("local[1]").setAppName("ModelBroadcast"))
+
+    val model = LeNet5(10)
+    model.getParameters()
+
+    val modelBroadCast = ModelBroadcast[Float].broadcast(sc, model)
+    modelBroadCast.value().toString should be(model.toString)
+    modelBroadCast.value().parameters()._1 should be(model.parameters()._1)
+    sc.stop()
```
yiheng: move the sc.stop to
after {
}

to make sure it won't affect other test result when it fail

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
```diff
@@ -71,6 +71,12 @@ class ModelBroadcast[T: ClassTag](implicit ev: TensorNumeric[T]) extends Seriali
         val wb = parameters._1(i)
         weightsBias(i) = Tensor[T](Storage(wb.storage().array()),
           wb.storageOffset(), wb.size(), wb.stride())
+      }
+      i += 1
+    }
+    while (i < parameters._1.length) {
+      if (parameters._1(i) != null) {
```
zhangxiaoli73: parameters._2(i) != null

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
```diff
@@ -71,7 +71,13 @@ class ModelBroadcast[T: ClassTag](implicit ev: TensorNumeric[T]) extends Seriali
         val wb = parameters._1(i)
         weightsBias(i) = Tensor[T](Storage(wb.storage().array()),
           wb.storageOffset(), wb.size(), wb.stride())
-        wb.set()
+      }
+      i += 1
+    }
+    i = 0
+    while (i < parameters._1.length) {
+      if (parameters._2(i) != null) {
```
zhangxiaoli73: should be parameters._2(i) != null && parameters._1(i) != null, right?

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
```diff
@@ -71,7 +71,13 @@ class ModelBroadcast[T: ClassTag](implicit ev: TensorNumeric[T]) extends Seriali
...
+        parameters._1(i).set()
```
yiheng: I don't think use the parameters() is a good choice, why not use getParameters?

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
```diff
@@ -71,7 +71,13 @@ class ModelBroadcast[T: ClassTag](implicit ev: TensorNumeric[T]) extends Seriali
         val wb = parameters._1(i)
         weightsBias(i) = Tensor[T](Storage(wb.storage().array()),
           wb.storageOffset(), wb.size(), wb.stride())
-        wb.set()
```
jason-dai: I don't understand - why resetting a tensor will impact another tensor? If so, this seems to be a bug of tensor

yiheng: reset will reset the storage, while all the weight/bias tensor sharing the same storage(with different offset)

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
SeaOfOcean: getParameters will copy all the parameters, which may not be necessary if only the inference is needed

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
```diff
@@ -71,7 +71,15 @@ class ModelBroadcast[T: ClassTag](implicit ev: TensorNumeric[T]) extends Seriali
         val wb = parameters._1(i)
         weightsBias(i) = Tensor[T](Storage(wb.storage().array()),
```
zhichao-li: Why we need this? can we change to wb.clone()?

yiheng: The weights are shared among models when prediction.

yiheng: @SeaOfOcean I think when you add the optimization, you should add comments to explain it. Or people will get confused. Can you add some comments?

PR #709, opened 2017-03-31
add python doc

qiuxin2012:

Add python doc

## How was this patch tested?

unit tests


yiheng:
LGTM

PR #713, opened 2017-03-31
make the error message of singleton check more clearly

yiheng:
make the error message of singleton check more clearly

## How was this patch tested?
unit test


spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -484,10 +484,14 @@ object DistriOptimizer {
       val (broadcastModel, broadcastCriterion, broadcastState, broadcastMethod) = broadcast.value
       if (!Engine.checkSingleton()) {
         if (checkSingleton) {
-          require(Engine.checkSingleton(), "Detect multi-task run on one Executor/Container. " +
-            "Please disable singleton check or repartition data")
+          require(Engine.checkSingleton(), "Partitions of the training data are not evenly" +
+            "distributed across the executors in the Spark cluster; are there sufficient training" +
+            "data to be distributed? Set property \"bigdl.check.singleton\" to false to skip " +
+            "this error")
```
jason-dai: "skip this check"

PR #718, opened 2017-04-01
clear the weights then clone the module to save memory

JerryYanWan:

In recurrent layer, clear the weights of a cell then clone this cell to save memory.

## How was this patch tested?
Test the correctness on the BiRecurrentSpec.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
@@ -87,6 +87,7 @@ class Recurrent[T : ClassTag]()
     if (t < times) {
       val cloneCell = cells.head.cloneModule()
       cloneCell.parameters()._1.foreach(x => null)
+      cloneCell.parameters()._2.foreach(x => null)
       while (t < times) {
         cells += cloneCell.cloneModule()
           .asInstanceOf[Cell[T]]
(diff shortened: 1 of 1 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -85,8 +85,11 @@ class Recurrent[T : ClassTag]()
     }
     var t = cells.length
     if (t < times) {
+      val cloneCell = cells.head.cloneModule()
+      cloneCell.parameters()._1.foreach(x => null)
```
yiheng: why not use Tensor.set() here? should not set tensor to null as it will loss the storage share information.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -85,8 +85,11 @@ class Recurrent[T : ClassTag]()
...
+      cloneCell.parameters()._2.foreach(x => null)
```
yiheng: why not use Tensor.set() here? should not set tensor to null as it will loss the storage share information.

yiheng:
https://github.com/intel-analytics/BigDL/blob/898de2353f85d51ecc6ea3fdde0687df2ffab2cc/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala#L66

PR #720, opened 2017-04-01
fix the bug of Engine init with single executor

SeaOfOcean:

fix the bug of Engine init with single executor

## How was this patch tested?

unit test


yiheng:
LGTM

SeaOfOcean:
pr validate pass

PR #722, opened 2017-04-01
add a profile for redhat5

yiheng:

add a profile for redhat5
## How was this patch tested?
unit test


luchy0120:
in mkl2017-xeon-blas ,we only use it to name the jni jar

PR #726, opened 2017-04-01
Make README simpler

122689305:

Make README simpler

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/ValidationMethod.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/ValidationMethod.scala
@@ -33,10 +33,13 @@ import scala.reflect.ClassTag
 trait ValidationMethod[T] extends Serializable {
   def apply(output: Activity, target: Activity): ValidationResult
 
+  // return the name of this method
   protected def format(): String
 
+  // return the name of this method
   override def toString(): String = format()
 
+  // deep clone the object
   override def clone(): ValidationMethod[T] = SerializationUtils.clone(this)
 }
 
@@ -47,15 +50,16 @@ trait ValidationMethod[T] extends Serializable {
  */
 trait ValidationResult extends Serializable {
 
+  // return the calculation results over all the samples in the batch
   def result(): (Float, Int) // (Result, TotalNum)
 
   // scalastyle:off methodName
   def +(other: ValidationResult): ValidationResult
 
-  // scalastyle:on methodName
-
+  // return the name of this trait
   protected def format(): String
 
+  // return the name of this trait
   override def toString(): String = format()
 }
 


spark/dl/src/main/scala/com/intel/analytics/bigdl/visualization/README.md
```diff
@@ -8,113 +8,39 @@
 >You can find the wheel repository at https://pypi.python.org/pypi/tensorboard
```
jason-dai: This wheel distribution is provided by https://github.com/dmlc/tensorboard

spark/dl/src/main/scala/com/intel/analytics/bigdl/visualization/README.md
```diff
@@ -8,113 +8,39 @@
...
 
 
-Simply, with one pip command you can install it in Ubuntu. But it is very likely that you will run into problems while installing. If you meet any error, please check the instructions below.
+Simply, with one pip command you can install it in Ununtu. But it is very likely that you will run into problems while installing. If you meet any error, please check the instructions below.
 
-If you install it with Mac OS, similar in ubuntu, it is easy to install with pip. Notice that the wheels for mac do not require CPython to support wide unicode. If you meet any error, read the instructions below carefully. You may find your own solutions from these information. 
+If you install it with Mac OS, similar in ubuntu, it is easy to install with pip. Notice that the wheels for mac do not require CPython to support wide unicode. If you meet any error, read the instructions below carefully. You may find your own solutions from these information.
+
```
jason-dai: Delete the above to paragraph; simply say something like:

```
Please follow the instructions below to install TensorBoard; it has been tested on both Ubuntu and Mac OS. Please refer to the "Know Issues" section for possible errors (please note that CPython is not require to support wide unicode on Mac OS)
```

Two questions though:
1) Have we tested it on Redhat or CentOS?
2) Why do we call out CPython is not required on mac? Does it hurt to have CPython installed on mac?

122689305: 1. Not tested yet.
2. It is that CPython which support wide unicode is not required on mac. Because the wheel for mac does not have this requiremennt according to the wheel tag: tensorboard-1.0.0a6-cp27-cp27m-macosx_10_11_x86_64.whl

jason-dai: 1) We need to test it on CentOS or Redhat
2) What if CPython is installed? Does it hurt?

122689305: 1. Qiuxin told me we have no time to test before 0.1.0-release.
2. Actually, on mac, python has to be compiled with CPython.  I think by default the python on mac is also launched by CPython.  I think the sentence about CPython on mac could be deleted. I wrote it just for a notice, because on many-linux platform it requires python2.7 to be launched by CPython which support wide unicode

PR #731, opened 2017-04-01
use variable in script to specify BigDL version

yiheng:

use variable in script to specify BigDL version
## How was this patch tested?
unit test


yiheng:
jenkins pass

PR #746, opened 2017-04-06
[WIP Issue 450] add trainable to module in order to support "freeze layers" in finetune

SeaOfOcean:

add a trainable parameter to AbstractModule, default = true

When doing backward, if the module is not trainable, accGradParameters is ignored

When updating the parameters in optim, getParameters(filterTrainable=true) (filterTrainable is false by default) is called, and only trainable parameters are collected and updated.

## How was this patch tested?

unit test

fixed https://github.com/intel-analytics/BigDL/issues/450

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/BatchNormalization.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/BatchNormalization.scala
@@ -658,6 +658,14 @@ class BatchNormalization[@specialized(Float, Double) T: ClassTag](
     }
   }
 
+  override def trainables(): Array[Boolean] = {
+    if (affine) {
+      Array(trainable, trainable)
+    } else {
+      null
+    }
+  }
+
   override def getParametersTable(): Table = {
     if (affine) {
       T(getName() -> T("weight" -> weight, "bias" -> bias,
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/BatchNormalization.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/BatchNormalization.scala
@@ -658,14 +658,6 @@ class BatchNormalization[@specialized(Float, Double) T: ClassTag](
     }
   }
 
-  override def trainables(): Array[Boolean] = {
-    if (affine) {
-      Array(trainable, trainable)
-    } else {
-      null
-    }
-  }
-
   override def getParametersTable(): Table = {
     if (affine) {
       T(getName() -> T("weight" -> weight, "bias" -> bias,
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Concat.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Concat.scala
@@ -189,13 +189,10 @@ class Concat[T: ClassTag](val dimension: Int)(
     offset = 1
     while (i < this.modules.length) {
       val currentOutput = this.modules(i).output.asInstanceOf[Tensor[T]]
+      if (!trainable) this.modules(i).setTrainable(trainable)
       val currentGradInput = this.modules(i)
-        .updateGradInput(input.asInstanceOf[Activity], gradouts(i).asInstanceOf[Activity])
+        .backward(input.asInstanceOf[Activity], gradouts(i).asInstanceOf[Activity])
         .asInstanceOf[Tensor[T]]
-      if (trainable) {
-        this.modules(i).accGradParameters(input.asInstanceOf[Activity],
-          gradouts(i).asInstanceOf[Activity])
-      }
 
       before = System.nanoTime()
       if (currentGradInput != null) {
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Concat.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Concat.scala
@@ -140,10 +140,12 @@ class Concat[T: ClassTag](val dimension: Int)(
     var i = 0
     while (i < this.modules.length) {
       val currentOutput = this.modules(i).output.asInstanceOf[Tensor[T]]
-      this.modules(i).accGradParameters(
-        input.asInstanceOf[Activity],
-        gradOutput.narrow(dimension, offset, currentOutput.size(dimension))
-          .asInstanceOf[Activity], scale)
+      if (this.modules(i).isTrainable()) {
+        this.modules(i).accGradParameters(
+          input.asInstanceOf[Activity],
+          gradOutput.narrow(dimension, offset, currentOutput.size(dimension))
+            .asInstanceOf[Activity], scale)
+      }
 
       i += 1
       offset += currentOutput.size(dimension)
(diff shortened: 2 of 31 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -236,6 +287,16 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
...
 
   /**
+   * This function returns the array of trainable states. each element corresponds to
+   * the element of weights parameters
+   * If module only has weight or only has bias, then the trainalbles.length == 1
+   * If module contains both weight and bias, then the trainalbles.length == 2, and they should
+   * be both true or false
+   * @return trainables
+   */
+  def trainables(): Array[Boolean] = null
```
qiuxin2012: why not use something like `if (trainables) parameters()._1 else null`

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/BatchNormalization.scala
```diff
@@ -658,6 +658,14 @@ class BatchNormalization[@specialized(Float, Double) T: ClassTag](
     }
   }
 
+  override def trainables(): Array[Boolean] = {
+    if (affine) {
+      Array(trainable, trainable)
+    } else {
+      null
```
yiheng: Is empty array better here?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -130,6 +131,26 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
     }
   }
 
+  protected var trainable = true
+
+  /**
+   * set trainable
+   * if trainable is true, train this module
```
yiheng: please explain what do you mean of train this module?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Concat.scala
```diff
@@ -140,10 +140,12 @@ class Concat[T: ClassTag](val dimension: Int)(
     var i = 0
     while (i < this.modules.length) {
       val currentOutput = this.modules(i).output.asInstanceOf[Tensor[T]]
-      this.modules(i).accGradParameters(
-        input.asInstanceOf[Activity],
-        gradOutput.narrow(dimension, offset, currentOutput.size(dimension))
-          .asInstanceOf[Activity], scale)
+      if (this.modules(i).isTrainable()) {
```
yiheng: also change the behavior in backward()?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
SeaOfOcean: do normal backward in the training process

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialConvolution.scala
```diff
@@ -90,6 +90,8 @@ class SpatialConvolution[T: ClassTag](
   @transient
   protected var results: Array[Future[Unit]] = null
 
+  setTrainable(propagateBack)
```
yiheng: do we still need propagateBack?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TimeDistributed.scala
```diff
@@ -162,7 +162,8 @@ class TimeDistributed[T : ClassTag] (layer: TensorModule[T])
    *
    * @return
    */
-  override def getParameters(): (Tensor[T], Tensor[T]) = layer.getParameters()
+  override def getParameters(filterTrainable: Boolean = false)
+  : (Tensor[T], Tensor[T]) = layer.getParameters()
```
yiheng: why not layer.getParameters(filterTrainable)?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/BatchNormalization.scala
yiheng: Is empty array better here?

yiheng: Is empty array better here?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
yiheng: what do you mean normal???

yiheng: Do you such general words in the docs.

yiheng: You should update the doc to make it more clear.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialConvolution.scala
SeaOfOcean: in SpatialConvolution, if propagateBack = false, the gradInput is not computed.

if trainable = false, accGradParameters is not called.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TimeDistributed.scala
SeaOfOcean: should have

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Concat.scala
SeaOfOcean: done

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -236,6 +263,18 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
   def parameters(): (Array[Tensor[T]], Array[Tensor[T]]) = null
 
   /**
+   * This function returns the array of trainable parameters.
+   * @return trainable parameters
+   */
+  def trainableParameters(): (Array[Tensor[T]], Array[Tensor[T]]) = {
```
yiheng: why not add an optional flag to the parameter API?

yiheng:
what if some people write a new container/module but doesn't use trainable variable to control its behavior? any way to give a compile/runtime error?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
SeaOfOcean: parameters() is override in those layers with weight or bias, if I use optional flag, I need to modify many files in order to filter trainable. Instead, trainableParameters is only needed in AbstractModule.scala and Container.scala

SeaOfOcean:
only newly added container needs to control trainable behavior

yiheng:
This should be duplicated with PR #969

PR #759, opened 2017-04-07
[fix] multi task issue on Mesos

i8run:

Multi task issue on Mesos. Because when spark driver starts to schedule, some of executors may be not started yet. So we should wait for all executors started.

## How was this patch tested?

Currently, only manual tests have done.


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -74,6 +75,29 @@ object Engine {
     _conf
   }
 
+  def waitAllExecutorsStartedInMesos: Unit = {
+    if (System.getProperty("spark.master").toLowerCase.startsWith("mesos")) {
+      val sc = SparkContext.getOrCreate(new SparkConf()
```
yiheng: Do you test it? I don't think it work because there will be two spark context in the JVM

PR #764, opened 2017-04-10
add rng.uniform for python

zhangxiaoli73:

add rng.uniform for python

## How was this patch tested?

unit test

### DIFF
--- a/pyspark/test/simple_integration_test.py
+++ b/pyspark/test/simple_integration_test.py
@@ -122,5 +122,22 @@ class TestWorkFlow(unittest.TestCase):
         for i in range(0, total_length):
             self.assertTrue(np.allclose(p[i], ground_label[i], atol=1e-6, rtol=0))
 
+    def test_rng(self):
+        rng = RNG()
+        rng.set_seed(100)
+        result = rng.uniform(0.1, 0.2, [2, 3])
+        ground_label = np.array([[0.15434049, 0.16711557, 0.12783694],
+                                 [0.14120464, 0.14245176, 0.15263824]])
+        self.assertTrue(result.shape == [2, 3])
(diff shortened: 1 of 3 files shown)


pyspark/test/simple_integration_test.py
```diff
@@ -122,5 +122,16 @@ def test_predict(self):
         for i in range(0, total_length):
             self.assertTrue(np.allclose(p[i], ground_label[i], atol=1e-6, rtol=0))
 
+    def test_rng(self):
+        rng = RNG()
+        rng.set_seed(100)
+        result = rng.uniform(0.1, 0.2, [2, 3])
```
zhichao-li: Check if the result still the same if ``` uniform``` being invoked multiple times with the same seed.

spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDL.scala
```diff
@@ -1329,6 +1329,13 @@ class PythonBigDL[T: ClassTag](implicit ev: TensorNumeric[T]) extends Serializab
   def initEngine(): Unit = {
     Engine.init
   }
+
+  def uniform(a: Double, b: Double, size: JList[Int]): JTensor = {
+    val result = Tensor[T]().resize(size.asScala.toArray)
+    result.apply1(i => ev.fromType(RandomGenerator.RNG.uniform(a, b)))
+    JTensor(result.storage().toList.map(_.asInstanceOf[Any]).asJava,
+      result.size().toList.asJava, typeName)
```
zhichao-li: use toJTensor(tensor) instead.

zhangxiaoli73:
jenkins pass

PR #765, opened 2017-04-10
modify colorjitter

JerryYanWan:
Fix bugs in colorjitter

## How was this patch tested?
Unit test
compare the results from Torch


spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/image/ColorJitter.scala
```diff
@@ -46,43 +49,56 @@ class ColorJitter extends Transformer[LabeledBGRImage, LabeledBGRImage] {
...
+    while (i < img1.length) {
+      img1(i) = img1(i) * alpha + (1 - alpha) * img2(i)
+      i += 1
+    }
+    img1
+  }
 
   private def saturation(variance: Float)(input: Array[Float]): Array[Float] = {
-    val gs = new Array[Float](input.length)
+    if (gs == null) gs = new Array[Float](input.length)
```
SeaOfOcean: what if input.length > gs.length

PR #768, opened 2017-04-11
use maven package to make a zip file for make dist 

glorysdj:
This change is [Reviewable](https://reviewable.io/reviews/intel-analytics/bigdl/768)

### DIFF
--- a/make-dist.sh
+++ b/make-dist.sh
@@ -49,10 +49,6 @@ mvn clean package -DskipTests $*
 
 BASEDIR=$(dirname "$0")
 DIST_DIR=$BASEDIR/dist
-BIN_DIR=$DIST_DIR/bin
-LIB_DIR=$DIST_DIR/lib
-CONF_DIR=$DIST_DIR/conf
-VERSION=0.2.0-SNAPSHOT
 
 if [ ! -d "$DIST_DIR" ]
 then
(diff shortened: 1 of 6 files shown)


glorysdj:
pass the validation at
http://172.168.2.101:8080/view/PR-Validation/job/BigDL-PR-Validation/588/

dist.xml
```diff
@@ -0,0 +1,46 @@
...
+          <include>bigdl.sh</include>
+          <include>classes.lst</include>
+          <include>img_class.lst</include>
+      </includes>
+    </fileSet>
+    <fileSet>
+      <outputDirectory>/lib</outputDirectory>
+      <directory>spark/dl/target</directory>
+      <includes>
+          <include>bigdl-*-jar-with-dependencies.jar</include>
```
yiheng: please also add the dependencies-and-spark jar. I notice in some example, some data preprocess job is still standalone java program. We still need that jar

yiheng:
can you also update the make-dist.sh script?

glorysdj:
both changes are done and pass the pr-validation:http://10.239.47.211:18888/view/PR-Validation/job/BigDL-PR-Validation/624

yiheng:
I don't think we should delete the make-dist.sh
1. it will check java/maven version for user
2. it appears a lot in our document. delete it need all theses document need to be updated
3. we just need to remove the copy logic in the scripts and copy the zip to the root folder?

make-dist.sh
```diff
@@ -47,30 +47,3 @@ fi
 
 mvn clean package -DskipTests $*
```
yiheng: please unzip to dist folder to keep consisted with the original build doc

yiheng:
please squash these commits

Issue #621, opened 2017-03-20
refactor the optimizer initialization process in python api


shane-huang:
1.  add *set_model()* to optimizer 
Currently the model could only be set to optimizer during creation. That means we can't reuse an optimizer for different models.  Creating optimizer may be time consuming sometimes. 
1. add *prepare_input()* to optimizer
Currently optimizer takes a long time to create and initialize.  The reason is loading input and caching takes time. We could warp this part of work into one function  and name it something like prepare_input/load_input, and allow user to train different models on the same set of input data. and it won't be confusing why initialization takes so long.

PR #774, opened 2017-04-13
[Issue 621]Support set model in optimizer

dding3:
1. add set model in optimizer
2. currently optimizer takes a long time to create and initialize, to avoid confusion, explicitly tell user it's loading data

## How was this patch tested?
unit test and manual test

fixed https://github.com/intel-analytics/BigDL/issues/621

### DIFF
--- a/pyspark/dl/optim/optimizer.py
+++ b/pyspark/dl/optim/optimizer.py
@@ -175,7 +175,7 @@ class Optimizer(JavaValue):
                            training_rdd, criterion, optim_method,
                            state, end_trigger, batch_size)
 
-    def setvalidation(self, batch_size, val_rdd, trigger, val_method=["Top1Accuracy"]):
+    def set_validation(self, batch_size, val_rdd, trigger, val_method=["Top1Accuracy"]):
         """
         Configure validation settings. 
 
@@ -188,7 +188,7 @@ class Optimizer(JavaValue):
         callBigDlFunc(self.bigdl_type, "setValidation", self.value, batch_size,
                       trigger, val_rdd, val_method)
 
-    def setmodel(self, model):
+    def set_model(self, model):
         """
         Set model. 
 
@@ -196,7 +196,7 @@ class Optimizer(JavaValue):
         """
         self.value.setModel(model.value)
 
-    def setcheckpoint(self, checkpoint_trigger,
+    def set_checkpoint(self, checkpoint_trigger,
                       checkpoint_path, isOverWrite=True):
         """
         Configure checkpoint settings. 
@@ -246,6 +246,14 @@ class Optimizer(JavaValue):
         callBigDlFunc(self.bigdl_type, "setValSummary", self.value,
                       summary)
         return self
+
+    def prepare_input(self):
+        """
+        Load input. Notebook user can call this method to seprate load data and
+        create optimizer time 
+        """
+        print("Loading input ...")
+        self.value.prepareInput()
 
 
 class TrainSummary(JavaValue, ):
--- a/pyspark/dl/optim/optimizer.py
+++ b/pyspark/dl/optim/optimizer.py
@@ -188,6 +188,14 @@ class Optimizer(JavaValue):
         callBigDlFunc(self.bigdl_type, "setValidation", self.value, batch_size,
                       trigger, val_rdd, val_method)
 
+    def setmodel(self, model):
+        """
+        Set model. 
+
+        :param model: new model
+        """
+        self.value.setModel(model.value)
+
     def setcheckpoint(self, checkpoint_trigger,
                       checkpoint_path, isOverWrite=True):
         """
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
@@ -677,6 +677,10 @@ class DistriOptimizer[T: ClassTag] (
 
   private var models: RDD[DistriOptimizer.Cache[T]] = null
 
+  import DistriOptimizer._
+  logger.info("Loading and caching data ...")
+  dataset.originRDD().cache().count()
+
   /**
    * Clean some internal states, so this or other optimizers can run optimize again
    *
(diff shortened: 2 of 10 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -677,6 +677,12 @@ class DistriOptimizer[T: ClassTag] (
 
   private var models: RDD[DistriOptimizer.Cache[T]] = null
 
+  import DistriOptimizer._
+  if (dataset.originRDD != null) {
+    logger.info("Loading and caching data ...")
+    dataset.originRDD().cache().count()
```
yiheng: The rdd has been already cached, why cache it again?

yiheng:
From the code, looks like after use this method to set model, the original model is not changed. I think it's really confused.

pyspark/dl/optim/optimizer.py
```diff
@@ -188,6 +188,14 @@ def setvalidation(self, batch_size, val_rdd, trigger, val_method=["Top1Accuracy"
         callBigDlFunc(self.bigdl_type, "setValidation", self.value, batch_size,
                       trigger, val_rdd, val_method)
 
+    def setmodel(self, model):
```
yiheng: From the issue, I think you should add another prepare_input method. And in python, the naming style should be set_model instead of setmodel

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -677,6 +677,12 @@ class DistriOptimizer[T: ClassTag] (
 
   private var models: RDD[DistriOptimizer.Cache[T]] = null
 
+  import DistriOptimizer._
+  if (dataset.originRDD != null) {
```
yiheng: From the issue, wrap this code in a method?

pyspark/dl/optim/optimizer.py
dding3: I checked other methods in optimizer.py, looks like the naming style is Java styple(eg, setvalidation, setcheckpoint). I use the same style to make sure the consistent style.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
dding3: My origin thought is to avoid user created his own rdd and didn't cache it. Since the optimizer use dataset.originRDD to training, what about we cache dataset.originRDD here and remove cache operation in dataset.scala.

dding3:
Hmm, I think the original model and _model in optimize reference the same object, since _model is updated, original model is updated as well. Please correct me if I understand wrong.

pyspark/dl/optim/optimizer.py
dding3: I didn't add a separate method perpare_input as I think the concern here is when create optimizer, it silently load/caching data which makes user confuse why create optimizer takes so long. I think it should be fine if we explicit told user we are loading data. Besides I think prepare_input is used to tell user we are loading data. I am not very likely to add the method and ask user call it. But if most of you think it's necessary to add the method, I will do it.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
yiheng: I don't agree optimizer do anything related with manipulate data, which should be delegate to dataset. The name of dataset is CachedDataset, so it cache the data for performance. If we have a new dataset named NoCacheDataSet to handle extra large data, will you cache it again?

pyspark/dl/optim/optimizer.py
yiheng: but you can see there are set_train_summary, set_val_summary in the same class. And most python classes use the under_score style. Why don't you follow the majority? Please use the underscore style, and change the other two:)

@zhichao-li looks like code styled in python is not consisted.

yiheng: Again, I think optimizer should not cache the data. It's not a job of the optimizer.

I'm agreed with your opinion. A prepare_input is misleading. I think the problem mentioned in the issue only a problem in some interactive programming(e.g. notebook), user may feel it takes long time when execute create optimizer. It's not a problem in a batch execution.

A prepare_input allow notebook user to seperate the data loading time and training time. If you don't have it, user may think the training take so long time, in which most is spent on loading data. I think maybe we provide this method in optimizer, even user didn't call that method, optimizer will do it at the start of training. And don't cache the data directly in optimizer. A dataset should also provide a method prepareData to hide the details.

yiheng:
It's wrong. In setModel, you point _model to another object, the model still point to the old object.

dding3:
Synced offline and learned the concern here is there are origin model and _model in optimizer and it's confused. I tried to remove _model and update origin model from val to var in optimizer, but find after reset model, subclass DistriOptimizer still point to the old module due to our current implementation. and we cannot update model in DistriOptimizer as var no matter model in base class is var or val. I keep current implementation and give some explanation in the comments.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Optimizer.scala
```diff
@@ -32,17 +32,17 @@ import scala.reflect.ClassTag
...
+ * @param _model the model to be trained
  * @param dataset the data set used to train a model
  * @param criterion the criterion used to evaluate the loss of the model given an input
  * @tparam T numeric type, which can be [[Float]] or [[Double]]
  * @tparam D the type of elements in DataSet, such as [[MiniBatch]]
  */
 // TODO: remove D to be MiniBatch[T]
 abstract class Optimizer[T: ClassTag, D](
-    protected val model: Module[T],
+  protected var _model: Module[T],
```
yiheng: why rename the variable? Can we only change val to var?

yiheng:
can you backport this to branch-0.1?

PR #777, opened 2017-04-13
Support graph container and add functional model defination

yiheng:
Support graph container and add functional model definition. So users can easier define more complex graph structure model.

## How was this patch tested?
add more unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
@@ -1,0 +1,335 @@
+/*
+ * Copyright 2016 The BigDL Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.intel.analytics.bigdl.nn
+
+import com.intel.analytics.bigdl.nn.Graph.ModuleNode
+import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, Activity, TensorModule}
+import com.intel.analytics.bigdl.tensor.Tensor
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+import com.intel.analytics.bigdl.utils.{Node, T, Table}
+
+import scala.reflect.ClassTag
+
+/**
+ * A graph container. Each node can have multiple inputs. The output of the node should be a tensor.
+ * The output tensor can be connected to multiple nodes. So the module in each node can have a
+ * tensor or table input, and should have a tensor output.
+ *
+ * The graph container can have multiple inputs and multiple outputs. If there's one input, the
+ * input data fed to the graph module should be a tensor. If there're multiple inputs, the input
+ * data fed to the graph module should be a table, which is actually an sequence of tensor. The
+ * order of the input tensors should be same with the order of the input nodes. This is also
+ * applied to the gradient from the module in the back propagation.
+ *
+ * All of the input modules must accept a tensor input. If your input module accept multiple
+ * tensors as input, you should add some Input module before it as input nodes and connect the
+ * output of the Input modules to that module.
+ *
+ * If there's one output, the module output is a tensor. If there're multiple outputs, the module
+ * output is a table, which is actually an sequence of tensor. The order of the output tensors is
+ * same with the order of the output modules. This is also applied to the gradient passed to the
+ * module in the back propagation.
+ *
+ * All inputs should be able to connect to outputs through some paths in the graph. It is
+ * allowed that some successors of the inputs node are not connect to outputs. If so, these nodes
+ * will be excluded in the computation.
+ *
+ * @param inputs input nodes
+ * @param outputs output nodes
+ * @tparam T Numeric type. Only support float/double now
+ */
+@SerialVersionUID(- 2896121321564992779L)
+class Graph[T: ClassTag](inputs : Seq[ModuleNode[T]],
+  outputs : Seq[ModuleNode[T]])(implicit ev: TensorNumeric[T])
+  extends Container[Activity, Activity, T]{
+
+  override def updateOutput(input: Activity): Activity = {
+    var i = 0
+    if (inputs.length == 1) {
+      // A tensor is expected here
+      inputsBP(i) = input.toTensor[T]
+      inputs(i).element.updateOutput(inputsBP(i))
+      i += 1
+    } else {
+      val inputTable = input.toTable
+      require(inputTable.length() == inputs.length,
+        s"Graph.updateOutput: Input tensor number(${inputTable.length()}) is not equal to " +
+          s"inputs node number(${inputs.length})")
+      while(i < inputTable.length()) {
+        // Each element of the table must be a tensor
+        inputsBP(i) = inputTable[Tensor[T]](i + 1)
+        inputs(i).element.updateOutput(inputsBP(i))
+        i += 1
+      }
+    }
+
+    while(i < executions.length) {
+      val node = executions(i)
+      inputsBP(i) = if (node.prevNodes.length == 1) {
+        node.prevNodes.head.element.output.toTensor[T]
+      } else {
+        seqToTable(node.prevNodes.map(_.element.output))
+      }
+      node.element.updateOutput(inputsBP(i))
+      i += 1
+    }
+
+    output = if (outputs.length == 1) {
+      outputs(0).element.output
+    } else {
+      seqToTable(outputs.map(_.element.output))
+    }
+    output
+  }
+
+  override def updateGradInput(input: Activity, gradOutput: Activity): Activity = {
+    var j = outputs.length - 1
+    var i = executions.length - 1
+    if (outputs.length == 1) {
+      // Expect a tensor here
+      outputs(j).element.updateGradInput(inputsBP(i), gradOutput.toTensor)
+      j -= 1
+      i -= 1
+    } else {
+      val gradOutputTable = gradOutput.toTable
+      require(gradOutputTable.length() == outputs.length,
+        s"Graph.updateGradInput:gradient tensor number(${gradOutputTable.length()}) is not " +
+          s"equal to outputs node number(${outputs.length})")
+      while(j >= 0) {
+        outputs(j).element.updateGradInput(inputsBP(i), gradOutputTable[Tensor[T]](j + 1))
+        j -= 1
+        i -= 1
+      }
+    }
+
+    while(i >= 0) {
+      val curNode = executions(i)
+      var curGradOutput : Tensor[T] = null
+      curNode.nextNodes.foreach(n => {
+        val nextGradOutput = if (n.prevNodes.length == 1) {
+          n.element.gradInput.toTensor
+        } else {
+          val nextGradOutputTable = n.element.gradInput.toTable
+          nextGradOutputTable[Tensor[T]](n.prevNodes.indexOf(curNode) + 1)
+        }
+
+        if (curGradOutput == null) {
+          curGradOutput = nextGradOutput
+        } else {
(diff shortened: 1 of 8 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -0,0 +1,283 @@
...
+  // the outputs will be excluded
+  private val dummyOutput = new ModuleNode[T](null)
+  outputs.foreach(_ -> dummyOutput)
+  private val backGraph = dummyOutput.graph(reverse = true)
+
+  // Build execution plan
+  private val executions = backGraph.topologySort.filter(_.element != null).reverse
+  modules.appendAll(executions.map(_.element.asInstanceOf[AbstractModule[Activity, Activity, T]]))
+
+  // Make input output nodes in executions be same order with how them are passed in
```
SeaOfOcean: them -> they

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -0,0 +1,283 @@
...
+ * will be excluded in the computation.
+ *
+ * @param inputs input nodes
+ * @param outputs output nodes
+ * @tparam T Numeric type. Only support float/double now
+ */
+class Graph[T: ClassTag](inputs : Seq[ModuleNode[T]],
+  outputs : Seq[ModuleNode[T]])(implicit ev: TensorNumeric[T])
+  extends Container[Activity, Activity, T]{
+
```
SeaOfOcean: can input same as output? (when the graph depth is 1?)

yiheng: it works. Add an unit test for this.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -0,0 +1,283 @@
...
+ * All inputs should be able to connect to outputs through some paths in the graph. It is
+ * allowed that some successors of the inputs node are not connect to outputs. If so, these nodes
+ * will be excluded in the computation.
+ *
+ * @param inputs input nodes
+ * @param outputs output nodes
+ * @tparam T Numeric type. Only support float/double now
+ */
+class Graph[T: ClassTag](inputs : Seq[ModuleNode[T]],
+  outputs : Seq[ModuleNode[T]])(implicit ev: TensorNumeric[T])
```
jason-dai: inputs can only accept Tensor? if so, shall we check that?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -0,0 +1,283 @@
...
+  extends Container[Activity, Activity, T]{
```
jason-dai: this is a container - what if someone calls add()?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -0,0 +1,283 @@
...
+  }
+
+  // Add a dummy output node, to get an one end graph. So the nodes that are not dependent by
+  // the outputs will be excluded
+  private val dummyOutput = new ModuleNode[T](null)
+  outputs.foreach(_ -> dummyOutput)
+  private val backGraph = dummyOutput.graph(reverse = true)
+
+  // Build execution plan
+  private val executions = backGraph.topologySort.filter(_.element != null).reverse
```
jason-dai: no need to filer - dummy node is the first node return by topologySort(), yes?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -0,0 +1,283 @@
...
+  modules.appendAll(executions.map(_.element.asInstanceOf[AbstractModule[Activity, Activity, T]]))
+
+  // Make input output nodes in executions be same order with how they are passed in
+  {
+    var i = 0
+    while(i < inputs.length) {
+      executions(i) = inputs(i)
```
jason-dai: what if some input is not connected to outputs?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -0,0 +1,283 @@
...
+      i += 1
+    }
+
+    val offset = executions.length - outputs.length
+    i = 0
+    while(i < outputs.length) {
+      executions(i + offset) = outputs(i)
```
jason-dai: there is no guarantee that the last several nodes are outputs; and what if some output is not connected to inputs?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -0,0 +1,283 @@
...
+      i += 1
+    }
+  }
+
+  private val inputsBP = new Array[Activity](executions.length)
+
+  // Check all inputs of the graph should be passed in
+  checkRoots
```
jason-dai: Instead of checkRoots, we need to check the graph more thoroughly:
1) All outputs are connected to inputs, and all inputs are connected to outputs
2) All outputs have no nextNodes, and all inputs have no prevNodes
3) There are no cycles in the graph

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -0,0 +1,283 @@
...
+
+object Graph {
+  /**
+   * Node for graph container. The module should have a tensor/table input while a tensor output
+   * @tparam T
+   */
+  type ModuleNode[T] = Node[AbstractModule[Activity, Tensor[T], T]]
+
+  /**
+   * Build a single input, single output graph container.
```
jason-dai: wrong comment

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -0,0 +1,283 @@
...
+   * @param output output node
+   * @return a graph container
+   */
+  def apply[T: ClassTag](input : Array[ModuleNode[T]], output : ModuleNode[T])
+    (implicit ev: TensorNumeric[T]) : Graph[T] = {
+    new Graph[T](input, Array(output))
+  }
+
+  /**
+   * Build a multiple inputs, multiple outputs graph container
```
jason-dai: wrong comment

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -0,0 +1,283 @@
...
+    var i = 0
+    if (inputs.length == 1) {
+      // A tensor is expected here
+      inputsBP(i) = input.toTensor[T]
+      inputs(i).element.updateOutput(inputsBP(i))
+      i += 1
+    } else {
+      val inputTable = input.toTable
+      require(inputTable.length() == inputs.length,
+        "Input tensor number is not equal to inputs node number")
```
jason-dai: Graph.updateOutput:

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -0,0 +1,283 @@
...
+    var j = outputs.length - 1
+    var i = executions.length - 1
+    if (outputs.length == 1) {
+      // Expect a tensor here
+      outputs(j).element.updateGradInput(inputsBP(i), gradOutput.toTensor)
+      j -= 1
+      i -= 1
+    } else {
+      val gradOutputTable = gradOutput.toTable
+      require(gradOutputTable.length() == outputs.length,
```
jason-dai: Graph.updateGradInput:

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -0,0 +1,283 @@
...
+    var j = outputs.length - 1
+    var i = executions.length - 1
+    if (outputs.length == 1) {
+      // Expect a tensor here
+      outputs(j).element.accGradParameters(inputsBP(i), gradOutput.toTensor)
+      j -= 1
+      i -= 1
+    } else {
+      val gradOutputTable = gradOutput.toTable
+      require(gradOutputTable.length() == outputs.length,
```
jason-dai: Graph.accGradParameters:

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DirectedGraph.scala
```diff
@@ -0,0 +1,170 @@
...
+ * @param reverse use the original direction or the reversed direction
+ * @tparam T Node element type
+ */
+class DirectedGraph[T](source : Node[T], reverse : Boolean = false) {
+
+  /**
+   * Topology sort.
+   * @return A sequence of sorted graph nodes
+   */
+  def topologySort : Array[Node[T]] = {
```
jason-dai: topologySort  can be implemented using DFS

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
yiheng: If there are multiple inputs, the input data fed to the graph module should be a table, which is actually an sequence of tensor. The order of the input tensors should be same with the order of the input nodes. 

We have code check that. Have added two unit test to cover this.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
yiheng: Good questions... Add makes no sense in the graph container. I can override it and throw exceptions to hint user.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
yiheng: changed. But I don't quite understand this as you can see where is the error from in the stack, why need to add Graph.updateOutput?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
yiheng: Yes. But it's easy to write some following code if we exclude it. e.g. if you don't filter here, you have to filter in the next line: modules.appendAll. It's one time work, don't need to worry the perf

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
yiheng: We needn't guarantee these nodes are first of the total graph. User can specify a part of graph to Graph container for computing. It's common use in tf and keras.

> If some input is not connected to outpus, there' will be error in graph init. And we have covered this in the doc and unit test.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
yiheng: We needn't guarantee these nodes are first of the total graph. User can specify a part of graph to Graph container for computing. It's common use in tf and keras.

> If some input is not connected to outpus, there' will be error in graph init. And we have covered this in the doc and unit test.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
yiheng: for 1,
>  All inputs should be able to connect to outputs through some paths in the graph. It is allowed that some successors of the inputs node are not connect to outputs. If so, these nodes will be excluded in the computation.

for 2,
I don't think we need do such limitation. User can specify a part of graph to Graph container for computing. It's common use in tf and keras.

for 3
It is already check in the topology sort.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DirectedGraph.scala
yiheng: I check the output of the DFS based topology sort. Notice that it may loss the parallel property. If we use the indegree based topology sort, we can know which nodes can be run in parallel. This property is useless now as we run a single model per thread but I think it's a useful property to have.

This remind me I should build a separated execution plan for forward and backward.

PR #779, opened 2017-04-13
Fix a few build warnings; move test files in wrong dir; up surefire memory

srowen:

This PR fixes a few build warnings I noted while getting tests up and running. More comments inline below.

## How was this patch tested?

Existing tests


spark/dl/pom.xml
```diff
@@ -142,6 +142,7 @@
                     <systemProperties>
                         <bigdl.disableCheckSysEnv>true</bigdl.disableCheckSysEnv>
                     </systemProperties>
+                    <argLine>-Xmx6g</argLine>
```
srowen: This is high, but I couldn't get all the tests to finish without running out of memory without this, like `GoogleNet_v2 model in batch mode`. 5GB might barely work.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -30,12 +30,14 @@ case object MklBlas extends EngineType
 
 
 object Engine {
-  @deprecated
+  @deprecated(
```
srowen: Build warning is generated without message and since

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -102,15 +104,8 @@ object Engine {
...
-  private val singletonCounter: AtomicInteger = new AtomicInteger(0)
-  private[bigdl] var localMode: Boolean = {
-    val env = System.getenv("BIGDL_LOCAL_MODE")
-    if(env == null) {
-      false
-    } else {
-      true
-    }
-  }
+  private val singletonCounter = new AtomicBoolean()
```
srowen: Some of this was a bit of simplification I indulged in while also touching up the docs and getting to know this code. Is it reasonable?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -367,7 +356,7 @@ object Engine {
       val patternLocalStar = "local\\[\\*\\]".r
       master match {
         case patternLocalN(n) => Some(1, n.toInt)
-        case patternLocalStar => Some(1, getCoreNumberFromEnv)
+        case patternLocalStar(_*) => Some(1, getCoreNumberFromEnv)
```
srowen: This was actually not matching as a regex otherwise

spark/dl/src/test/scala/com/intel/analytics/bigdl/nn/ReverseSpec.scala
```diff
@@ -15,10 +15,9 @@
  */
 package com.intel.analytics.bigdl.nn
 
-import com.intel.analytics.bigdl.nn.JoinTable
```
srowen: This actually caused a warning due to shadowing

spark/dl/src/test/scala/com/intel/analytics/bigdl/python/api/PythonSpec.scala
```diff
@@ -127,12 +127,13 @@ class PythonSpec extends FlatSpec with Matchers with BeforeAndAfter {
       valRdd = data.toJavaRDD(),
       vMethods = util.Arrays.asList("Top1Accuracy", "Loss"))
 
-    val trainSummary = TrainSummary(sc.appName, "lenet")
+    val logdir = com.google.common.io.Files.createTempDir()
```
srowen: Not a warning, but without this it generates its test output in the source tree and git complains it's an untracked file

spark/dl/pom.xml
qiuxin2012: It seems don't work. I get an OOM Error while I'm running `maven clean package`
```
Exception in thread "dag-scheduler-event-loop" Exception in thread "SparkListenerBus" 
Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "dag-scheduler-event-loop"

Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "SparkListenerBus"

Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "sparkDriver-akka.remote.default-remote-dispatcher-6"

```

spark/dl/src/test/scala/com/intel/analytics/bigdl/python/api/PythonSpec.scala
qiuxin2012: Looks Good.

spark/dl/pom.xml
srowen: I hesitated to increase it further, but could. Really it's quite a big test! @qiuxin2012 do you know how much memory worked for you? 8g?

qiuxin2012: @srowen Aha, I know what's happening. It's not caused by the total size of heap. 6GB is enough for me. But my maven is running with JAVA7, so I will get an OOM Error without "-XX:MaxPermSize=1G".
Could you help to add this configuration to your PR?

srowen: Sure, I added that. It'll generate a warning in Java 8, but that's not a big deal.

qiuxin2012:
Great work, test passed. Thanks very much.

PR #784, opened 2017-04-17
fix: at the end of make-dist.sh, copy wrong file.

i8run:

copy right jar and zip, which has been changed with SNAPSHOT.

## How was this patch tested?

manual.

### DIFF
--- a/make-dist.sh
+++ b/make-dist.sh
@@ -71,6 +71,6 @@ cp $BASEDIR/scripts/classes.lst $BIN_DIR/
 cp $BASEDIR/scripts/img_class.lst $BIN_DIR/
 cp $BASEDIR/spark/dl/src/main/resources/spark-bigdl.conf $CONF_DIR/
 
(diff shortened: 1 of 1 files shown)


make-dist.sh
```diff
@@ -71,6 +71,6 @@ cp $BASEDIR/scripts/classes.lst $BIN_DIR/
 cp $BASEDIR/scripts/img_class.lst $BIN_DIR/
 cp $BASEDIR/spark/dl/src/main/resources/spark-bigdl.conf $CONF_DIR/
 
-cp $BASEDIR/spark/dl/target/bigdl-$VERSION-jar-with-dependencies.jar $LIB_DIR/
```
srowen: It generally won't be a SNAPSHOT build, right?

yiheng: On master, it's a SNAPSHOT build. On release branch(e.g. branch-0.1), it's not SNAPSHOT build.

yiheng:
I think we needn't this if PR#768 is merge.

yiheng:
PR#768 is merge.

PR #791, opened 2017-04-18
fix: sometimes model spec will abort unexpectedly.

i8run:

Make mode spec parallel to serial.


yiheng:
LGTM

PR #794, opened 2017-04-18
remove unnecessary clone in FlattenTable

SeaOfOcean:

remove unnecessary clone in FlattenTable
as input.flatten() will not change input

## How was this patch tested?

unit test


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/FlattenTable.scala
```diff
@@ -39,8 +39,7 @@ class FlattenTable[T: ClassTag] (implicit ev: TensorNumeric[T])
   }
 
   override def updateOutput(input: Table): Table = {
-    createInputMap()
-    inputMap = input.clone
+    inputMap = input
```
qiuxin2012: I think we doesn't need `inputMap` and `createInputMap`

psyyz10:
LGTM

PR #796, opened 2017-04-18
support DiceCoefficient criterion

JerryYanWan:
add a new method DiceCoefficient criterion

## How was this patch tested?

Unit Test;
Test the output and gradInput of this layer by comparing them to those from Theano results.

Some similar implementations for Dice-Coefficient criterion
https://github.com/jocicmarko/ultrasound-nerve-segmentation/blob/master/train.py
https://groups.google.com/forum/#!topic/keras-users/QyuYygRB91M
https://github.com/fchollet/keras/issues/2994

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/DiceCoefficientCriterion.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/DiceCoefficientCriterion.scala
@@ -44,6 +44,8 @@ class DiceCoefficientCriterion[@specialized(Float, Double) T: ClassTag]
   @transient
   private val buffer2: Tensor[T] = Tensor[T]()
   @transient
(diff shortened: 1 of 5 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/DiceCoefficientCriterion.scala
```diff
@@ -0,0 +1,154 @@
...
+    if (null == buffer1) buffer1 = Tensor[T]()
+    if (null == w1) w1 = Tensor[T]()
+    if (null == w2) w2 = Tensor[T]()
+    if (null == _outputs) _outputs = Tensor[T]()
+
+    _input = input
+    _target = target
+
+    if (_input.dim() == 1) {
+      _input = _input.view(1, _input.nElement())
```
qiuxin2012: merge with  `_input = input  _target = target`

PR #798, opened 2017-04-18
use maven package instead of script to copy dist artifact together

yiheng:

use maven package instead of script to copy dist artifact together

## How was this patch tested?
unit test


yiheng:
duplicate with #768

PR #799, opened 2017-04-19
Change some vars to vals

plrectco:

Change some vars to vals, as a pre-step for sharing memory feature 

## How was this patch tested?
style check


plrectco:
Jenkins passed

PR #806, opened 2017-04-21
[WIP] remove exclude files from test profile

yangw1234:

remove exclude files from test profile

## How was this patch tested?

existing tests


yangw1234:
does not work this way, close for now

Issue #792, opened 2017-04-18
DistriOptimizer.optimize() failed silently when batchSize is not correctly set


yangw1234:
When batchSize is not correct, calling DistriOptimizer.optimize() will fail silently. It seems optimizer only catches and logs the optimizer but did not throw that exception to caller. [https://github.com/intel-analytics/BigDL/blob/master/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala#L744](https://github.com/intel-analytics/BigDL/blob/master/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala#L744)

yiheng:
@dding3 I think for IllegualArgumentException, we should not retry but just throw it.

dding3:
Sure, make sense.

PR #811, opened 2017-04-23
[Issue 792]throw exceptions after retry exceeds settings

dding3:
Optimizer throw exceptions if retry numbers exceed max retry numbers.

## How was this patch tested?
unit test

fixed https://github.com/intel-analytics/BigDL/issues/792


yiheng:
LGTM

PR #813, opened 2017-04-24
[WIP]Add synchronous easgd

fanshiqing:

Add synchronous Elastic Averaging SGD (EASGD) algorithm to BigDL.

## How was this patch tested?

WIP:
-- UT
-- Distri test on dataset

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
@@ -133,6 +133,7 @@ object DistriOptimizer {
     val driverSubModelNum = partitionNum * _subModelNumber
     var dropModelNumBatch = 0
     var lossArray = new Array[Double](_subModelNumber)
+    val isEASGD = optimMethod.isInstanceOf[EAMSGD[T]]
 
     var epochStart = System.nanoTime()
     var dataRDD = dataset.data(train = true)
@@ -144,9 +145,7 @@ object DistriOptimizer {
       metrics.set("computing time for each node", mutable.ArrayBuffer[Double](), sc)
       metrics.set("computing time average", 0.0, sc, partitionNum)
       metrics.set("aggregate gradient time", 0.0, sc, partitionNum)
-      if (optimMethod.isInstanceOf[EAMSGD[T]]) {
-        metrics.set("computing elastic difference time", 0.0, sc, partitionNum)
-      }
+      if (isEASGD) metrics.set("computing elastic difference time", 0.0, sc, partitionNum)
       metrics.set("get weights average", 0.0, sc, partitionNum)
       metrics.set("get weights for each node", mutable.ArrayBuffer[Double](), sc)
 
@@ -158,10 +157,10 @@ object DistriOptimizer {
         (data, modelIter) => {
           val cached = modelIter.next()
           val syWStart = System.nanoTime()
-          val localParamsBuffer = if (optimMethod.isInstanceOf[EAMSGD[T]]) {
-            cached.elasticDifference
+          val (localParamsBuffer, putParamsBuffer) = if (isEASGD) {
+            (cached.elasticDifference, cached.elasticDifference)
           } else {
-            cached.modelWeights.head
+            (cached.modelWeights.head, cached.gradient)
           }
           val weightsResult = parameters.getWeights(localParamsBuffer)
           val tensorBuffer = new Array[(Tensor[T], Tensor[T])](_subModelNumber)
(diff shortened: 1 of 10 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -260,11 +291,24 @@ object DistriOptimizer {
         models.mapPartitions(modelIter => {
           val modelCache = modelIter.next()
           parameters.aggregrateGradientPartition()
-          parameters.gradientPartition.div(ev.fromType(finishedModelNum))
+          if (!optimMethod.isInstanceOf[EAMSGD[T]]) {
```
dding3: I think we can put this logic in the following if(optimMethod.isInstanceOf[EAMSGD[T]]). It will make code cleaner.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -522,6 +566,7 @@ object DistriOptimizer {
         cached.map(_._4), // criterions
         cached.map(_._5), // states
         cached.head._2.clone(), // a tensor buffer
+        cached.head._2.clone(), // a tensor buffer for elastic difference
```
dding3: Looks like elastic difference in a node is the same, I think we can just keep one copy in node to save memory just as what we did for weight.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/EAMSGD.scala
```diff
@@ -0,0 +1,137 @@
...
+   *                config("commPeriod"): sync update (communication period) \atu
+   * @param state   a table describing the state of the optimizer; after each call the state
+   *                is modified
+   *                state("vt"): vector of movement at time t
+   *                state("suw"): vector of weight for communication with the center variable
+   *                state("sug"): vector of gradient for communication with the center variable
+   * @return the new x 1D vector and the function list, evaluated before the update
+   */
+  override def optimize(feval: (Tensor[T]) => (T, Tensor[T]), x: Tensor[T],
+    config: Table, state: Table = null): (Tensor[T], Array[T]) = {
```
dding3: Is this method called in somewhere? And i think we need add some ut with EAMSGD.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -244,9 +253,31 @@ object DistriOptimizer {
...
+                  .add(cached.modelWeights.head.narrow(1, offset + 1, length))
+                  .mul(ev.fromType(cached.localStates.head.getOrElse[Double]("movingRate", 0.0)))
+              }))
+              driverMetrics.add("computing elastic difference time", System.nanoTime() - time)
+            }
           }
 
-          parameters.putGradients(cached.gradient)
+          if (!optimMethod.isInstanceOf[EAMSGD[T]]) {
+            parameters.putGradients(cached.gradient)
```
dding3: move this logic to previous "if (optimMethod.isInstanceOf[EAMSGD[T]])" ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/EAMSGD.scala
fanshiqing: No.
Indeed currently we simply implement a **Synchronous** EASGD and the update logic for Synchronous one is simple and was moved to `DistOptimizer`. 
This `EAMSGD.optimzie` method is planned for **Asynchronous** EASGD which is not implemented now.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
fanshiqing: For EASGD `parameters.gradientPartition` actually holds `elasticDifferencepartition` which do not need to do the average operation.

dding3: Yes, i understand. I mean we can merge this two if into one, i think there are too many "if (optimMethod.isInstanceOf[EAMSGD[T]])" which makes code looks not clean

dding3:
1. Looks like allreduceparameter now stores elasticdifference instead of weight, which will break getmodel and checkpoint function. We also need update getmodel and checkpoint as well.
2. In this PR, model rdd is updated inplace(weight here) and looks like there is not a way to recover it.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
fanshiqing: Good idea, thanks!

fanshiqing:
@dding3 

- For 1st point: AllreduceParameter keeps unchanged which global distributed `parameters` still stores weight. The detailed synchronous EASGD update rule is as follows:
image

- For 2nd point: Each executor update local full-size model and update elastic difference to AllReduceParameter synchronous, so each executor's model differs from each other as the training goes. This is what EASGD algorithm differs from other optim algorithms. So in my mind there is no need to recover local model if some executor goes wrong.

PR #816, opened 2017-04-25
assign updateOutput return value to output

SeaOfOcean:
assign updateOutput return value to output

## How was this patch tested?

unit test


qiuxin2012:
LGTM

PR #820, opened 2017-04-25
throw exceptions after retry exceeds settings

dding3:
Optimizer will throw exception if retry times exceed max retry times

## How was this patch tested?
unit test

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
@@ -741,6 +741,8 @@ class DistriOptimizer[T: ClassTag] (
         )
         retryNum = Int.MaxValue
       } catch {
+        case e: IllegalArgumentException =>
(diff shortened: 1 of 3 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -750,6 +752,9 @@ class DistriOptimizer[T: ClassTag] (
              */
             if (System.nanoTime() - lastFailureTimestamp < maxRetry * retryTimeInterval * 1e9) {
               retryNum += 1
+              if (retryNum == maxRetry) {
```
zhichao-li: What if ```  (checkpointPath.isDefined)  ``` is false?  exception would be swallow silently.

dding3: Thank you for point this out. See this PR https://github.com/intel-analytics/BigDL/pull/889/files

PR #822, opened 2017-04-26
refactor squeeze layer

yiheng:

Squeeze layer can squeeze multiple singleton dimensions. Fix a bug when batch size is 1.

## How was this patch tested?
unit test


qiuxin2012:
LGTM

PR #826, opened 2017-04-27
add py3.5 support (lenet5 tested)

122689305:

**Modify the codes to support Python 3.5**

1. lambda function.
Python3.5 doesn't support something like 'lamda (x,y): f(x,y)'. It only supports something like 'lambda x:f(x[0], x[1])'.

2. print function.
Python3.5 use 'print' as a function

3. bytes and str.
Python3.5 provides a new data type called 'bytes'. 'bytes' object need to be decoded to 'str' type, so that it is compatible.

4. iteritems
Python3.5 removes 'dict.iteritems' method. It only supports 'dict.items'

**Modify the LeNet5 README.md**

- --num-executors need to be specified

**Modify dataset/mnist.py**
change 'import base' to 'from . import base'
Otherwise in yarn-client mode, 'base' cannot be found.

## How was this patch tested?

passed with spark 1.6 or 2.0

### DIFF
--- a/pyspark/dl/models/lenet/README.md
+++ b/pyspark/dl/models/lenet/README.md
@@ -36,7 +36,6 @@ We would train a LeNet model in spark local mode with the following commands and
         --driver-cores 2  \
         --driver-memory 2g  \
         --total-executor-cores 2  \
-        --num-executors 4 \
         --executor-cores 2  \
         --executor-memory 4g \
         --conf spark.akka.frameSize=64 \
--- a/pyspark/dl/models/lenet/README.md
+++ b/pyspark/dl/models/lenet/README.md
@@ -34,11 +34,12 @@ We would train a LeNet model in spark local mode with the following commands and
     ${SPARK_HOME}/bin/spark-submit \
         --master ${MASTER} \
         --driver-cores 2  \
-       --driver-memory 2g  \
-       --total-executor-cores 2  \
-       --executor-cores 2  \
-       --executor-memory 4g \
-       --conf spark.akka.frameSize=64 \
+        --driver-memory 2g  \
+        --total-executor-cores 2  \
+        --num-executors 4 \
+        --executor-cores 2  \
+        --executor-memory 4g \
+        --conf spark.akka.frameSize=64 \
         --py-files ${PYTHON_API_ZIP_PATH},${BigDL_HOME}/pyspark/dl/models/lenet/lenet5.py  \
         --properties-file ${BigDL_HOME}/dist/conf/spark-bigdl.conf \
         --jars ${BigDL_JAR_PATH} \
--- a/pyspark/dl/models/textclassifier/README.md
+++ b/pyspark/dl/models/textclassifier/README.md
@@ -17,8 +17,8 @@ or [20 Newsgroup dataset](http://www.cs.cmu.edu/afs/cs.cmu.edu/project/theo-20/w
 ```{r, engine='sh'}
 $ [/tmp/news20]$ tree . -L 1
   .
-  ├── 20_newsgroup
-  └── glove.6B
+  ├── 20news-19997.tar.gz
+  └── glove.6B.zip
 ```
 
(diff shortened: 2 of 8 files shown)


pyspark/dl/models/textclassifier/textclassifier.py
```diff
@@ -35,10 +35,10 @@ def text_to_words(review_text):
 
 
 def analyze_texts(data_rdd):
-    return data_rdd.flatMap(lambda (text, label): text_to_words(text)) \
+    return data_rdd.flatMap(lambda args: text_to_words(args[0])) \
```
zhichao-li: change to : 
```
lambda text_label: text_to_words(text_label[0])
```

pyspark/dl/models/lenet/README.md
```diff
@@ -34,11 +34,12 @@ We would train a LeNet model in spark local mode with the following commands and
...
         --master ${MASTER} \
         --driver-cores 2  \
-       --driver-memory 2g  \
-       --total-executor-cores 2  \
-       --executor-cores 2  \
-       --executor-memory 4g \
-       --conf spark.akka.frameSize=64 \
+        --driver-memory 2g  \
+        --total-executor-cores 2  \
+        --num-executors 4 \
```
zhichao-li: remove   --num-executors 4 , as it's a parameter on yarn mode.

pyspark/dl/models/textclassifier/textclassifier.py
```diff
@@ -35,10 +35,10 @@ def text_to_words(review_text):
...
         .map(lambda word: (word, 1)).reduceByKey(lambda a, b: a + b) \
-        .sortBy(lambda (w, c): - c).zipWithIndex() \
-        .map(lambda ((w, c), i): (w, (i + 1, c))).collect()
+        .sortBy(lambda args: - args[1]).zipWithIndex() \
```
zhichao-li:
```
lambda (x, y): x + y
translated to:

lambda x_y: x_y[0] + x_y[1]
```

pyspark/dl/models/textclassifier/textclassifier.py
```diff
@@ -35,10 +35,10 @@ def text_to_words(review_text):
 
 
 def analyze_texts(data_rdd):
-    return data_rdd.flatMap(lambda (text, label): text_to_words(text)) \
+    return data_rdd.flatMap(lambda text_label: text_to_words(text_label[0])) \
         .map(lambda word: (word, 1)).reduceByKey(lambda a, b: a + b) \
-        .sortBy(lambda (w, c): - c).zipWithIndex() \
-        .map(lambda ((w, c), i): (w, (i + 1, c))).collect()
+        .sortBy(lambda w_c: - w_c[1]).zipWithIndex() \
+        .map(lambda w_c__i: (w_c__i[0][0], (w_c__i[1] + 1, w_c__i[0][1]))).collect()
```
zhichao-li: it's not easy to understand the intention via lambda. how about add a function like this:

```

def index(w_c_i):
    ((w, c), i) = w_c_i
    return (w, (i + 1, c))
```

pyspark/dl/models/textclassifier/README.md
```diff
@@ -17,15 +17,16 @@ or [20 Newsgroup dataset](http://www.cs.cmu.edu/afs/cs.cmu.edu/project/theo-20/w
...
+  ├── 20news-19997.tar.gz
+  └── glove.6B.zip
 ```
 
 then running the flowing script would automatically download the data during the first run.
 
 bigdl.sh would setup the essential environment for you and it would accept a spark-submit command as an input parameter.
 
 ```{r, engine='sh'}
+        PYTHONHASHSEED=...
```
zhichao-li: @122689305 , This has been fixed in Spark 2.2. we might need to remove this later. 
```
self.environment["PYTHONHASHSEED"] = os.environ.get("PYTHONHASHSEED", "0")
```

Issue #788, opened 2017-04-17
Retire env vars, System properties?


srowen:
I am working on fixing calls to the deprecated `Engine.init` from tests. The only thing making it somewhat difficult is that configuration is passed around via system properties, not a config object (like the Spark config), which makes testing hard in the new code path.

More generally, there are some instances where BigDL takes input from environment vars, and some where it uses System properties, and some where it uses Spark config. I think it is generally bad practice to use env vars or System properties; just makes things difficult to test and encapsulate. At least, seems unnecessary to use env vars and properties separately.

Is any non-Spark usage actually supported? could config actually just use the Spark config object alone? that is can "local" mode go away?

It seems early enough in the project that changes like this are easy to make so wanted to raise it.

jason-dai:
There are three sets of env variables used in BigDL:
1. MKL related, including KMP_BLOCKTIME, OMP_NUM_THREADS, etc. We need to set them properly using env variables (for both driver and executors), as these are expected by MKL.

2. "Local" mode related including "BIGDL_LOCAL_MODE" and "DL_CORE_NUMBER". These are supposed to be used for internal testing only, and shouldn't be used by the users. (Currently "DL_CORE_NUMBER" is also used when Spark running with local[*]; we can eliminate that and just use all the available cores on the local node).

3. "DL_ENGINE_TYPE". I think we should change this to a Java property.

srowen:
Thanks @jason-dai -- yes the MKL env variables must remain of course. Internal-only env variables aren't a big deal. I was wondering whether those env variables can even go away, but that's secondary.

The issue is really using Java System properties instead of a config system. I think this causes problems eventually. Right now, it makes it hard to rewrite the tests that all call the deprecated `Engine.init` method. Updating them to use the version that takes an existing SparkConf doesn't work because the code expects certain Spark props to be set as global system properties, and won't read them from the conf. This happens to work when the app is run with spark-submit but not necessarily otherwise.

That's really the motivation for wanting to reduce or eliminate them. Maybe I can work on a prototype that narrowly addresses that problem.

srowen:
@jason-dai as a particular example of envs though, why is `DL_ENGINE_TYPE` an env variable? seems like simple user-land config.

jason-dai:
Yes, as mentioned above, we should change "DL_ENGINE_TYPE" to a config.

srowen:
[...]

PR #814, opened 2017-04-24
[WIP] Remove calls to deprecated Engine.init, by removing use of System properties for Spark config

srowen:
See https://github.com/intel-analytics/BigDL/issues/788
It's a big change and am interested in feedback on whether this is good cleanup or a bit too far.

## Changes

- Remove (unused?) "local" mode in `Engine`, along with `BIG_DL_LOCAL_MODE` and script arg `--local`
- Remove use of System property for Spark config, instead using SparkConf to get/set them
- As a result, remove calls to deprecated `Engine.init` method
- Standardize how `Engine.createContext`, `Engine.init` and creation of `SparkConf` and `SparkContext` are handled
- Use `SparkContext.getOrCreate()` instead of `new SparkContext` consistently
- Introduce `SparkContextSpec` to uniformly handled creating and cleaning up a SparkContext for tests that need it.
- Remove duplicated definitions of `processPath`
- (A few minor changes along the way: fix build warning, remove dead imports)

Note that docs will need to be updated to match the new style of invoking `Engine.init` if this is accepted.

## Test failures

I am seeing test failures when I run with these changes, although each one I believe is not due to the change.

I am seeing the failures mentioned at https://github.com/intel-analytics/BigDL/issues/810

One type is memory-related, and I'm not sure how to give `luajit` more memory. I wonder if the test is too big?

```
- should generate correct output and grad
InceptionSpec:
Inception+bn
...
============== lua code end ================

/var/folders/gn/2df1bw014_d91xr49gpdm3b00000gn/T/UnitTest3826245994945462751lua
/Users/srowen/Documents/torch/install/bin/luajit: not enough memory
- should generate correct output *** FAILED ***
  java.lang.RuntimeException: Nonzero exit value: 1
  at scala.sys.package$.error(package.scala:27)
  at scala.sys.process.ProcessBuilderImpl$AbstractBuilder.slurp(ProcessBuilderImpl.scala:132)
  at scala.sys.process.ProcessBuilderImpl$AbstractBuilder.$bang$bang(ProcessBuilderImpl.scala:102)
  at com.intel.analytics.bigdl.torch.TH$.runNM(TH.scala:121)
...
```

This one looks like the test is just a little overly sensitive; should I adjust it here?

```
============== lua code end ================

/var/folders/gn/2df1bw014_d91xr49gpdm3b00000gn/T/UnitTest2144912581200729840lua
luaTime:0.0021841526031494
- should generate correct output and grad *** FAILED ***
[...]

PR #827, opened 2017-04-27
Replace DL_ENGINE_TYPE env variable with property bigdl.engineType for now, and also make it default to mklblas

srowen:

Replace `DL_ENGINE_TYPE` env variable with property `bigdl.engineType` for now, and also make it default to `mklblas`. This makes it work with the only value that's supported out of the box.

## How was this patch tested?

Existing tests.

This is a small subset of https://github.com/intel-analytics/BigDL/issues/788


srowen:
@jason-dai I think this much you were supportive of.

I am actually wondering if `DL_CORE_NUMBER` is relevant now either; it's only used to select the number of cores in Spark `local[*]` mode but that is implicitly the number of cores on the machine already.

Likewise for `BIGDL_LOCAL_MODE` -- is there a mode that doesn't even use Spark? it seems like this mode is covered by Spark local modes already, or could be.

jason-dai:
@srowen Yes, I think this is good change. For `DL_CORE_NUMBER` - I agree that we shouldn't use it even in Spark `local[*]` (just use all the cores available in the local machine).

`DL_CORE_NUMBER` is needed for `BIGDL_LOCAL_MODE`, which does not use Spark and is only used for internal testing.

srowen:
Done, thanks @jason-dai . I guess this is a question then -- what is local mode if it's not Spark local mode? it looks like everything is always using Spark in all events, but "local mode" just causes it to not try to read core count, etc from the Spark master. But using master `local[n]` will achieve the same thing.

That is, I'm sure the purpose is important, just wondering if this is even necessary to achieve it.

The reason I'm asking about it more than a little is that it complicates Engine and tests a little bit with the extra global variable, and makes it a little harder to eliminate the calls to deprecated Engine.init.

jason-dai:
@srowen in "local mode" the program runs as a local Java program (not using Spark local), which is easier for some testing (e.g., requiring fewer resources than Spark local), and makes low latency prediction possible (e.g., embedding in an application server).

On the other hand, even in "local model", BigDL still use the Spark jar, and I think maybe we can make all these config as SparkConf, which will make life easier as you suggested.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -124,16 +126,13 @@ object Engine {
...
   private var engineType: EngineType = {
-    val dlEngineType = System.getenv("DL_ENGINE_TYPE")
-
-    if (dlEngineType == null || dlEngineType.toLowerCase == "mklblas") {
-      MklBlas
-    } else {
-      throw new Error(s"Unknown DL_ENGINE_TYPE. $ENV_VAR_ERROR")
+    System.getProperty("bigdl.engineType", "mklblas").toLowerCase(Locale.ROOT) match {
+      case "mklblas" => MklBlas
+      case _ => throw new IllegalArgumentException(s"Unknown engine type. $ENV_VAR_ERROR")
```
jason-dai: Maybe we shouldn't use `ENV_VAR_ERROR` here?

srowen: Done. Should I try to remove the local mode flag or is that best left for another time, to consider separately?

jason-dai: We can leave the "local mode" flag for another time.

jason-dai:
LGTM. @yiheng please take a look.

yiheng:
This looks good to me. Let me run a jenkins job.

yiheng:
There's one style check error:
error file=/var/jenkins_home/workspace/BigDL-PR-Validation-StyleCheck/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala message=Whitespace at end of line line=153 column=0

and some python errors:
To reproduce them, please
1. run `mvn package -DskipTests`(or `./make-dist.sh`)
2. run `source scripts/bigdl.sh`
3. go to **pyspark/test/dev**
4. run `./run-all.sh`

srowen:
Got it, fixed, squashed and pushed. I'll check out the test script, from now on for vetting not just tests but style.

srowen:
Pardon @yiheng would this be ready to go now? the style error was fixed and the test script passes locally.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -356,7 +359,7 @@ object Engine {
       val patternLocalStar = "local\\[\\*\\]".r
       master match {
         case patternLocalN(n) => Some(1, n.toInt)
-        case patternLocalStar(_*) => Some(1, getCoreNumberFromEnv)
+        case patternLocalStar(_*) => Some(1, getNumMachineCores)
```
yiheng: This changes the original logic. In the original code, local[*] will use DL_CORE_NUMBER cores if it is set, then all the cores on the machine. Now, it will ignore the DL_CORE_NUMBER setting.

I think the change is reasonable. But it breaks some unit tests in python if the machine cores number is not 4 or 8(e.g. on some server). 

After some dig, I find that the root cause is even we specify the master to local[4] in python test file pyspark/test/simple_integration_test.py line:30, the master config value read in scala is local[*]...

yiheng:
@srowen Just realize that the script passed in your local but it fails on our server. Please see my comment. I think the change is okay but need to fix the broken unit tests.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
jason-dai: I think it is still a good idea to eliminate the env as much as we can; the error above seems to be an bug (in Spark or BigDL) and we should fix that.

srowen:
Here, I pulled in changes from https://github.com/intel-analytics/BigDL/pull/898 to avoid a system property for spark.master. This basically requires also removing use of system properties for other spark.* properties.

This then led to changes in EngineSpec to correctly configure itself through a SparkConf. One side effect is that it's not quite possible to test YARN/Mesos/standalone integration because setting such a master will fail without a cluster. These are disabled for the moment.

However, this all passes now.

Some possible next steps that I noted in https://github.com/intel-analytics/BigDL/pull/814 but have not implemented here, to keep it simpler:

- Passing SparkConf to Engine.init explicitly would let the tests hack around the cluster master issue, but don't know if it's worth that breaking change
- All tests that use SparkContext could be refactored to handle it consistently and correctly; this was an element of https://github.com/intel-analytics/BigDL/pull/814
- Code should use SparkContext.getOrCreate consistently, ideally
- Fix duplication of TestUtils.processPath

What do you think @yiheng @zhichao-li?

srowen:
This now restores testing of YARN, Mesos, standalone env config

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -334,41 +317,40 @@ object Engine {
...
   }
 
-  private def getSparkMaster : String = {
-    System.getProperty("spark.master")
+  /**
+   * Extract spark executor number and executor cores from environment.
+   * @return (nExecutor, executorCore)
+   */
+  private[utils] def sparkExecutorAndCore(): Option[(Int, Int)] = {
+    parseExecutorAndCore(SparkContext.getOrCreate().getConf)
```
zhichao-li: Looks good to me. Not sure why we didn't do this at the first try any other concerns? @yiheng

zhichao-li:
LGTM. I would opt for using sparkconf directly instead of java properties.

zhichao-li:
Jenkins pass.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -88,15 +90,12 @@ object Engine {
    */
   def init: Unit = this.synchronized {
     if (localMode) {
-      require(getSparkMaster == null, "Detect BIGDL_LOCAL_MODE and spark.master are both set." +
-        " They're conflict. Please reset BIGDL_LOCAL_MODE if you're use Spark or remove " +
-        "spark.master if you're run without spark")
       logger.info("Detect BIGDL_LOCAL_MODE is set. Run workload without spark")
       // The physical core number should have been initialized by env variable in bigdl.sh
-      setNodeAndCore(1, getCoreNumberFromEnv)
+      setNodeAndCore(1, getNumMachineCores)
```
yiheng: Here we need a way to control how many cores are used in local mode(run as a standalone java program). We used the env var DL_CORE_NUMBER before. Now should we use a java property to control it?

jason-dai: I think we can continue to use DL_CORE_NUMBER for that purpose - it is currently used for internal testing together with DL_LOCAL_MODE.

srowen: Can you simply set the spark master to `local[n]` to control the number of cores? that's the most natural way to do it. Do you need this to be plumbed back to the script that parses the --cores argument? @jason-dai @yiheng

yiheng: Here the local mode is not spark local. It's pure local program(normal JVM program) run without spark. So there's no way to get the 'local[n]'.

Some of our internal perf test jobs run under this mode.

srowen: @yiheng I'm still not so sure how the local mode differs from Spark local mode, since this uses Spark in all cases and Spark local is in one JVM, but this is a separate topic, maybe. See https://github.com/intel-analytics/BigDL/pull/827#issuecomment-298637426 for example. It seems like you can still set a Spark master in this case to control the configure Engine cores, even if some code will not actually use Spark.

Is the ask here to put back the `DL_CORE_NUMBER` property? that's fine by me, whatever would make this easier to digest. More changes could happen later.

yiheng: @srowen I think it's a little weird if the user uses Spark master control the core number while the program will not actually use Spark...

One of the differences between spark local and the local mode is that spark local will still use spark component(e.g. RDD, BlockManager) on a single node, while local mode will directly work on arrays. See [LocalOptimizer](https://github.com/intel-analytics/BigDL/blob/master/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/LocalOptimizer.scala). This will bring some performance difference when training some **small** models. For example, when train LeNet, the local mode can get a 4000+ throughput on my desktop, but spark local can only get 2000+. When you train some bigger model(spend more time on model forward/backward), the perf reduce become less as the overhead of spark become relatively smaller.

We still recommend the user to use spark local when development on the single node. But the local mode is still kept and used in some of our internal tests.

Ideally, I think we can use a JVM property to replace the DL_CORE_NUMBER. But keep it temperately is also fine to me.

yiheng:
By the way, looks like there's some compile error with the latest master code.

[ERROR] /var/jenkins_home/workspace/BigDL-PR-Validation-2.1-python/spark/dl/src/test/scala/com/intel/analytics/bigdl/optim/DLEstimatorSpec.scala:49: value sparkLocalEnv is not a member of object com.intel.analytics.bigdl.utils.TestUtils
[ERROR]     TestUtils.sparkLocalEnv(core = 1) {

srowen:
Thanks @yiheng , yeah, a new file added yesterday needed to change as well to work with this change. I fixed that, and fixed up an unrelated test that was missing a `torchCheck()` to get it passing.

I just restored `DL_CORE_NUMBER` for now.

yiheng:
Jenkins pass

PR #829, opened 2017-04-28
add VolumetricMaxPooling

SeaOfOcean:

add VolumetricMaxPooling

## How was this patch tested?

unit test


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/NNPrimitive.scala
```diff
@@ -495,6 +495,216 @@ object NNPrimitive {
     }
   }
 
+  def volumetricMaxPoolingForwardDouble(input: Array[Double], inputOffset: Int,
```
qiuxin2012: this is not a public nn primitive, so move it to VolumetricMaxPooling

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/VolumetricMaxPooling.scala
```diff
@@ -0,0 +1,296 @@
...
+      if ((otime - 1) * dT >= itime + padT) otime -= 1
+      if ((oheight - 1) * dH >= iheight + padH) oheight -= 1
+      if ((owidth - 1) * dW >= iwidth + padW) owidth -= 1
+    }
+    require(otime >= 1 && owidth >= 1 && oheight >= 1,
+      s"Given input size: (${ nslices }x${ itime }x${ iheight }x${ iwidth })." +
+        s" Calculated output size:" +
+        s" (${ nslices }x${ otime }x${ oheight }x${ owidth }). Output size is too small")
+
+    // non-batch mode
```
qiuxin2012: Right place?

PR #830, opened 2017-04-28
Add backward and forward for module and criterion

zhichao-li:
Add backward and forward python api for module and criterion. 
These API would not as efficient as the scala version as it introduce some data conversion overhead, so you may want to use ``` optimizer.optimize()``` in production.

## How was this patch tested?
unittest


yiheng:
LGTM

PR #835, opened 2017-05-02
[WIP]Caffe model Support

wzhongyuan:

Support Caffe pre-trained models in BigDL
## How was this patch tested?
Unit testing by comparing the result of dynamic loaded model and pre-defined model ,also testing perf


jenniew:
Can you also add python API?

PR #837, opened 2017-05-03
fix LookupTable python api

yangw1234:

fix LookupTable python api, otherwise python will interpret the `padding_value` as a integer and a MethodNotFoundException will be thrown.


yangw1234:
pr validation pass

PR #841, opened 2017-05-04
revise sentenceSplitter and sentenceTokenizer to load files from hdfs

JerryYanWan:
load en-sent.bin and en-token.bin from hdfs.
Revise SentenceTokenizer and SentenceSplitter transformers

## How was this patch tested?
Run the RNN train model script.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/SentenceSplitter.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/SentenceSplitter.scala
@@ -32,19 +32,19 @@ import scala.collection.Iterator
  * If sentFile is None, the default sentence delimiter is period.
  * @param sentFile A trained model by OpenNLP
  */
-class SentenceSplitter(sentFile: Option[String] = None, host: Option[String] = None)
+class SentenceSplitter(sentFile: Option[String] = None)
   extends Transformer[String, Array[String]] {
 
   var modelIn: FileInputStream = _
   var model: SentenceModel = _
   var sentenceDetector: SentenceDetector = _
 
-  def this(sentFileURL: URL, host: Option[String]) {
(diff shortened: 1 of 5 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/SentenceSplitter.scala
```diff
@@ -30,19 +32,19 @@ import scala.collection.Iterator
  * If sentFile is None, the default sentence delimiter is period.
  * @param sentFile A trained model by OpenNLP
  */
-class SentenceSplitter(sentFile: Option[String] = None)
+class SentenceSplitter(sentFile: Option[String] = None, host: Option[String] = None)
```
zhangxiaoli73: try input whole file path, such as: hdfs://127.0.0.1:9000/tmp/xxx

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/rnn/Utils.scala
```diff
@@ -170,11 +175,12 @@ object SequencePreprocess {
...
+    host: Option[String])
   : DataSet[Array[String]] = {
 
-    val sentenceSplitter = SentenceSplitter(sentBin)
-    val sentenceTokenizer = SentenceTokenizer(tokenBin)
+    val sentenceSplitter = SentenceSplitter(sentBin, host)
+    val sentenceTokenizer = SentenceTokenizer(tokenBin, host)
     val lines = load(fileName)
     val tokens = DataSet.array(lines, sc)
       .transform(sentenceSplitter).toDistributed().data(false)
```
zhangxiaoli73: don't collect, you can just operate on rdd.

zhangxiaoli73:
Code LGTM, please check example property and accuracy

PR #846, opened 2017-05-04
modify py files to support Sphinx pydoc gen

122689305:

You may view http://172.168.2.168:5555 to see the change

The change were made according to the instructions http://docutils.sourceforge.net/docs/user/rst/quickref.html#substitution-references-and-definitions and https://pythonhosted.org/an_example_pypi_project/sphinx.html#substitutions

I use 'literal blocks' to quote the comments. So the whitespaces and indentations will not be changed. But the problem it doesn't support auto line-wrapped. Another way is to use 'Line Blocks', but continuous whitespaces will be reduced to one.


122689305:
I also upload the auto-modify script into another commit. It's a simple finite state machine

122689305:
According to the python doc of spark, I embedded the FSA codes to be as an extension to the Sphinx. Also I modified this FSA so it can auto transform ':param xxx:' with no proceeding blank line, ':param xxx yyy', ':param xx' and '@param xxx:' to the right syntax.

I wrote a doc-gen.sh and a doc-web.sh so people can easily generate the doc and launch a web server to see the result.

I also wrote a README.md to introduce it.

pyspark/docs/conf.py
```diff
@@ -0,0 +1,343 @@
...
+# serve to show the default.
+
+import sys
+import os
+
+# If extensions (or modules to document with autodoc) are in another directory,
+# add these directories to sys.path here. If the directory is relative to the
+# documentation root, use os.path.abspath to make it absolute, like shown here.
+sys.path.insert(0, os.path.abspath('.'))
+sys.path.insert(1, u'/home/yuchao/BigDL/pyspark/dl')
```
zhichao-li: yuchao?

pyspark/docs/index.rst
```diff
@@ -0,0 +1,25 @@
+.. dl documentation master file, created by
```
zhichao-li: I'm thinking this generated code should not be included in BigDL.

PR #848, opened 2017-05-05
fix: preload libjmkl.so

### LINKED ISSUE #821: Shall we need pre-load MKL?
MKL library will be loaded after one of methods in `MKL.java` is called.
Currently, tensor computing (add, mul) will take some time to pre-load the libjmkl.so at first.

i8run:

Preload libjmkl.so by hand.

## How was this patch tested?

Manual test.

fixed #821

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
@@ -35,10 +35,6 @@ private[tensor] class DenseTensor[@specialized(Float, Double) T: ClassTag](
   var nDimension: Int)(implicit ev: TensorNumeric[T])
   extends Tensor[T] {
 
-  // pre-load MKL library. If we do not do it here,
-  // libjmkl.so will be loaded when one of the methods of in MKL is called.
-  MKL.isMKLLoaded
-
   override def storage(): Storage[T] = _storage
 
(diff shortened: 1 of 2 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
```diff
@@ -35,6 +35,10 @@ private[tensor] class DenseTensor[@specialized(Float, Double) T: ClassTag](
   var nDimension: Int)(implicit ev: TensorNumeric[T])
   extends Tensor[T] {
 
+  // pre-load MKL library. If we do not do it here,
```
yiheng: why not put it in an object?

i8run: If it's in `object DenseTensor`, `new DenseTensor` will not load library automatically.

yiheng: If you put here, each time new DenseTensor, the MKL.isloaded will be run.

If you put in an Tensor object, the MKL.isloaded will be run only when JVM load the class

jason-dai: Yes, we should just put it in the initialization code of a Tensor object

PR #849, opened 2017-05-05
Remove *.txt from gitignore as it matched versioned files; rearrange entries and standardize entries

srowen:

The only real change here is to not match `*.txt*` in `.gitignore` because that matches `requirements.txt`, which is versioned, and generates a warning. The rest of the changes are just rearranging and rationalizing the entries a bit.

## How was this patch tested?

Existing tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,29 +1,33 @@
+# Compile products
 *.pyc
 *.class
-*.log
-dl/src/main/python/.idea/
+target/
+dist/
 
-# sbt specific
+# SBT, Maven specific
 .cache
 .history
 .lib/
-dist/*
-target/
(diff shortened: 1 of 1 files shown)


yiheng:
LGTM

.gitignore
```diff
@@ -1,29 +1,33 @@
+# Compile products
 *.pyc
 *.class
-*.log
-dl/src/main/python/.idea/
+target/
+dist/
```
srowen: Ack @yiheng I made a mistake here. This still actually matches the `spark/dist` directory. I should have changed it to `/dist/` to only match the root dist directory. I'm sorry about that, I'll open another PR to correct it.

PR #855, opened 2017-05-08
share memory in CAddTable for LSTM, GRU, SimpleRNN

JerryYanWan:

share memory in CAddTable for LSTM, GRU, SimpleRNN;
share memory for Reverse layer.

## How was this patch tested?

Unit Tests


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/GRU.scala
```diff
@@ -135,9 +135,8 @@ class GRU[T : ClassTag] (
             bRegularizer = bRegularizer)))
         .add(Sequential()
           .add(Dropout(p))
-          .add(Linear(outputSize, outputSize, withBias = false,
-            wRegularizer = uRegularizer))))
-      .add(CAddTable())
+          .add(Linear(outputSize, outputSize, withBias = false))))
```
psyyz10: why remove remove regularizer?

PR #860, opened 2017-05-09
close stream in File.scala after use

### LINKED ISSUE #564: Stream needed to be closed after dumping object to file
I noticed in File.scala, in save method, we didn't close stream after using it dumping object to file. I think it's necessary to close it to release related resource.

qiuxin2012:
close stream in File.scala after use

## How was this patch tested?

unit tests

fixed #564

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/File.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/File.scala
@@ -24,7 +24,7 @@ import org.apache.hadoop.fs.Path
 import org.apache.hadoop.io.IOUtils
 
 object File {
-  private[bigdl]  val hdfsPrefix: String = "hdfs:"
+  private[bigdl] val hdfsPrefix: String = "hdfs:"
 
   /**
    * Load torch object from a torch binary file
@@ -66,8 +66,11 @@ object File {
       if (Files.exists(Paths.get(fileName)) && !isOverwrite) {
         throw new RuntimeException("file exists!")
       }
-      val objFile = new ObjectOutputStream(new FileOutputStream(fileName))
+      val os = new FileOutputStream(fileName)
+      val objFile = new ObjectOutputStream(os)
       objFile.writeObject(obj)
+      objFile.close()
+      os.close()
     }
   }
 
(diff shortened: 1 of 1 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/File.scala
```diff
@@ -66,8 +66,11 @@ object File {
       if (Files.exists(Paths.get(fileName)) && !isOverwrite) {
         throw new RuntimeException("file exists!")
       }
-      val objFile = new ObjectOutputStream(new FileOutputStream(fileName))
+      val os = new FileOutputStream(fileName)
+      val objFile = new ObjectOutputStream(os)
       objFile.writeObject(obj)
+      objFile.close()
```
srowen: `finally` block?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/File.scala
```diff
@@ -93,7 +96,12 @@ object File {
     val byteArrayOut = new ByteArrayOutputStream()
     val objFile = new ObjectOutputStream(byteArrayOut)
     objFile.writeObject(obj)
-    IOUtils.copyBytes(new ByteArrayInputStream(byteArrayOut.toByteArray), out, 1024, true)
+    val byteArrayIn = new ByteArrayInputStream(byteArrayOut.toByteArray)
+    IOUtils.copyBytes(byteArrayIn, out, 1024, true)
+    objFile.close()
+    byteArrayIn.close()
```
srowen: Closing byte array in/out streams is a no-op. It could still be good form.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/File.scala
yiheng: +1

dding3:
I am thinking do we need add e.printStackTrace in catch block so it can print stack trace if an error occurs.

qiuxin2012:
@dding3 I think we don't need to catch the exception here, let application to deal with the exception.

PR #867, opened 2017-05-10
Touch up gitignore to only exclude root /dist dir

srowen:

I made an error in https://github.com/intel-analytics/BigDL/pull/849 and this still doesn't only gitignore the root /dist directory.

## How was this patch tested?

N/A

https://github.com/intel-analytics/BigDL/pull/849

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 *.pyc
 *.class
 target/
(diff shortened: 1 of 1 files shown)


srowen:
Sorry about this @yiheng

PR #872, opened 2017-05-11
fix bug of CMul and CAdd when getParameters is called

SeaOfOcean:

fix bug of CMul and CAdd when getParameters is called

## How was this patch tested?

Unit test


qiuxin2012:
LGTM.

PR #889, opened 2017-05-15
throw exception if retry exceeds max retryNum without setting checkpoint dir

dding3:
throw exception if retry exceeds max retryNum without setting checkpoint dir

## How was this patch tested?
unit test


spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -788,9 +788,9 @@ class DistriOptimizer[T: ClassTag] (
             models = DistriOptimizer.initThreadModels(newModel, dataset, criterion, state,
               nodeNumber, coresPerNode, checkSingleton, parameters, validationMethods)
           } else {
-            retryNum = Int.MaxValue
             DistriOptimizer.logger.info("Failed to recover since no model snapshot" +
               "checkpoint path is not set")
```
jason-dai: remove either no or not from the log

Issue #883, opened 2017-05-15
java.lang.NoSuchMethodError when training Inception v1 and VGG on spark 2.1.1


ZeweiChen11:
BigDL train InceptionV1  with apache Spark2.1.1, spark is download from official website directly, training failed with below error. Build spark manually:
`./dev/make-distribution.sh  --tgz -Pyarn  -Phadoop-2.7  -Dhadoop.version=2.7.3  -DskipTests`
But we encountered the following issue when training both Inception v1 and VGG.
We trained with bigdl latest master branch using spark yarn-client mode.
BigDL building commands:
 `bash make-dist.sh -P spark_2.1`
Running command to train vgg: 
`./dist/bin/bigdl.sh -- spark-submit --class com.intel.analytics.bigdl.models.vgg.Train --master yarn --deploy-mode client --executor-cores=44 --num-executors=8 --executor-memory=300g --conf "spark.yarn.executor.memoryOverhead=60g" /opt/BigDL/dist/lib/bigdl-0.2.0-SNAPSHOT-jar-with-dependencies.jar --batchSize 1408 -f /root/workspace/dataset/cifar-10/ --checkpoint /mnt/disk2/vgg_cifar10_Spark2.1.1_yarn_client/`

Running command to train Inception v1: 
`./dist/bin/bigdl.sh -- spark-submit --class com.intel.analytics.bigdl.models.inception.TrainInceptionV1 --master yarn --deploy-mode client --executor-cores=44 --num-executors=8 --executor-memory=300g --conf "spark.yarn.executor.memoryOverhead=60g" /opt/BigDL/dist/lib/bigdl-0.2.0-SNAPSHOT-jar-with-dependencies.jar --batchSize 1408 -f hdfs:///user/root/sequence/ --checkpoint /mnt/disk2/inceptionV1_Spark2.1.1_yarn_client/  --learningRate 0.0896
`

17/05/12 15:15:21 INFO utils.ThreadPool$: Set mkl threads to 1 on thread 1
17/05/12 15:15:33 INFO utils.Engine$: Auto detect executor number and executor cores number
17/05/12 15:15:33 INFO utils.Engine$: Executor number is 8 and executor cores number is 44
17/05/12 15:15:33 INFO utils.Engine$: Find existing spark context. Checking the spark conf...
17/05/12 15:15:34 INFO optim.DistriOptimizer$: caching training rdd ...
[...]

PR #895, opened 2017-05-16
add profile for spark2.1.1

yiheng:

In Spark 2.1.1, the Blockmanager API putBytes is changed. An optional parameter is added. This won't break the code which invoke it, but will break the compiled package. So if you want your application run on Spark 2.1.1, you need to recompile against spark-core 2.1.1 version.

To compatible with spark 2.1.1, this profile is added.

## How was this patch tested?

manual test and existing unit test

fix https://github.com/intel-analytics/BigDL/issues/883

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -556,6 +556,16 @@
             </properties>
         </profile>
         <profile>
+            <id>spark_2.1.1</id>
+            <properties>
+                <spark-version.project>2.0</spark-version.project>
+                <spark.version>2.1.1</spark.version>
+                <scala.major.version>2.11</scala.major.version>
+                <scala.version>2.11.8</scala.version>
+                <scala.macros.version>2.1.0</scala.macros.version>
+            </properties>
+        </profile>
+        <profile>
             <id>scala_2.11</id>
             <properties>
                 <scala.major.version>2.11</scala.major.version>


srowen:
This is getting hairy. 2.2 was going to need a new profile too, but it may end up being the same change you're trying to account for here.

What about using reflection to patch over these differences? having so many profiles is going to run into problems.

jason-dai:
Yes, we should look into patch in future.

yiheng:
Create an issue for the reflection refactor. https://github.com/intel-analytics/BigDL/issues/900

pom.xml
```diff
@@ -556,6 +556,16 @@
             </properties>
         </profile>
         <profile>
+            <id>spark_2.1.1</id>
+            <properties>
+                <spark-version.project>2.0</spark-version.project>
```
srowen: By the way, this profile isn't really needed. The profile only really has to exist to select the 2.0 vs 1.5-plus module. from there callers can easily modify `-Dspark.version=...` as needed. I get that it's a convenience, but also a maintenance burden. For example, you don't want a cloudera-2.1 profile; you just need to use `-Pcloudera` to activate the repo and then select `-Dspark.version=` accordingly.

Of course this doesn't solve the bigger issue of how many different versions of the library to publish, which is a real problem. I'm slugging through that now; it's going to be tricky, but might be able to make one Spark 2 binary work for everything.

PR #898, opened 2017-05-16
Get master from spark context instead of java property

zhichao-li:

Get master from spark context instead of java property

## How was this patch tested?
unit-test


yiheng:
Does this change pass all existing unit test? Does it solve the problem in https://github.com/intel-analytics/BigDL/pull/827?

srowen:
@yiheng @zhichao-li yeah, this is related to what https://github.com/intel-analytics/BigDL/pull/827 is trying to do, but not quite the same. But, yes, I am more generally hoping we can make exactly this kind of change.

Let me go update that PR with the further changes I alluded to, in that Engine.init needs to know about the SparkConf. Once that's in place, it's also trivial to use it for the master config.

srowen:
I've made further changes in https://github.com/intel-analytics/BigDL/pull/827 which incorporate these changes. Does this change pass tests? I am not sure, for instance, that EngineSpec will pass this way, or if it does, it's because it's still trying to test YARN by setting the sys property spark.master, but it's no longer used. I think its "YARN" test would actually just test a local master's behavior.

I will post more thoughts on the other PR about what I believe we can merge now, and changes that can follow later.

yiheng:
LGTM

zhichao-li:
The current test is trying to simulate different cluster environments to verify the ``` Engin.init``` method. @srowen I agree that some of the parsing methods should be refactor and be tested directly. Maybe we can do that in a following PR.

zhichao-li:
Jenkins pass. I'm going to merge this if no objections.

srowen:
I'd only ask you consider my existing PR first. I think it is simpler and goes farther.  Is there a hurry?

zhichao-li:
@srowen That's exactly what the pr use for to correct some test failure which would be trigger by #827  as some of the python tests can not get the correct spark master url.

jason-dai:
@srowen #827 is currently blocked due to some unit test failures; this PR will fix these failures, so that it is possible to merge #827

srowen:
Yes, but https://github.com/intel-analytics/BigDL/pull/827 also fixes this. It should pass too

zhichao-li:
Let me trigger the jenkins test for #827.  The previous try it fail on the batch size was not divisible by the ```physical cores``` although we've already create a SparkContext with local[4]. You can also verify this locally by running ```BigDL/pyspark/test/dev/run-all.sh```

srowen:
Yeah it passed for me yesterday locally, including the pyspark tests. Let me run again with the latest change to verify.

PR #899, opened 2017-05-16
Pyspark test

122689305:

This test is used in Jenkins. It can test comments style in python files. All blocks wrapped by ''' will be parsed. Literal blocks wrapped by ``` will be skipped. The rules for comments are:

line strating with '>>>' means a prompt line.
line starting with :?@?\s*(?:param|return) means a parameter line.

It will check whether there is a blank line before the first prompt line/ parameter line or not, check whether a parametet line is a single line or multiple line, check the format of ':param xx:' and ':return:' is right or not.

## How was this patch tested?

Exception: line 156, in /home/yuchao/BigDL/pyspark/dl/util/common.py:
         :return (storage, shape)
A colon is required to be attached just after ':param xx', and no blank gap is allowed. For example, ':param xx:', not ':param xx :' or ':param xx'. For 'return', a colon is also required, like ':return:'.

Exception: line 110, in /home/yuchao/BigDL/pyspark/dl/optim/optimizer.py:
            every "n" iterations
Add a blank line before it to make it as a new paragraph. Or append it to the previous parameter line. You have to write the parameter definition in one line.

Exception: line 1413, in /home/yuchao/BigDL/pyspark/dl/nn/layer.py:
     :param min_value: minValue in f(x), default is -1.
A blank line preceding the first parameter line is required. Try adding a blank line before it.

### DIFF
--- a/pyspark/test/dev/comment.py
+++ b/pyspark/test/dev/comment.py
@@ -24,11 +24,13 @@ def _process_docstring(filepath, lines):
     param_re = re.compile(r'^\s*:?@?\s*(?:param\s+|return:?).*$')
     blank_re = re.compile(r'\s*$')
     prmpt_re = re.compile(r'\s*>>>.*$')
+    liter_re = re.compile(r'\s*```\s*$')
 
     quote_flag = False
     param_flag = False
     blank_flag = False
     prmpt_flag = False
+    liter_flag = False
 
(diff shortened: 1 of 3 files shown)


122689305:
Also the logic fault in dev/test/diff.py is fixed too

122689305:
@zhichao-li

Issue #782, opened 2017-04-16
ResNet on Cifar will throw java.lang.IllegalArgumentException: requirement failed


i8run:
Error message is below,

```
2017-04-16 16:32:34 ERROR ThreadPool$:98 - Error: java.lang.IllegalArgumentException: requirement failed
	at scala.Predef$.require(Predef.scala:221)
	at com.intel.analytics.bigdl.nn.Concat.updateGradInput(Concat.scala:124)
	at com.intel.analytics.bigdl.nn.Concat.updateGradInput(Concat.scala:42)
	at com.intel.analytics.bigdl.nn.Sequential.updateGradInput(Sequential.scala:50)
	at com.intel.analytics.bigdl.nn.ConcatTable.updateGradInput(ConcatTable.scala:136)
	at com.intel.analytics.bigdl.nn.ConcatTable.updateGradInput(ConcatTable.scala:34)
	at com.intel.analytics.bigdl.nn.abstractnn.AbstractModule.backward(AbstractModule.scala:173)
	at com.intel.analytics.bigdl.nn.Sequential.backward(Sequential.scala:85)
	at com.intel.analytics.bigdl.nn.Sequential.backward(Sequential.scala:85)
	at com.intel.analytics.bigdl.nn.Sequential.backward(Sequential.scala:82)
	at com.intel.analytics.bigdl.optim.LocalOptimizer$$anonfun$5$$anonfun$apply$1.apply$mcD$sp(LocalOptimizer.scala:116)
	at com.intel.analytics.bigdl.optim.LocalOptimizer$$anonfun$5$$anonfun$apply$1.apply(LocalOptimizer.scala:107)
	at com.intel.analytics.bigdl.optim.LocalOptimizer$$anonfun$5$$anonfun$apply$1.apply(LocalOptimizer.scala:107)
	at com.intel.analytics.bigdl.utils.ThreadPool$$anonfun$invokeAndWait$1$$anonfun$apply$2.apply(ThreadPool.scala:95)
	at scala.concurrent.impl.Future$PromiseCompletingRunnable.liftedTree1$1(Future.scala:24)
	at scala.concurrent.impl.Future$PromiseCompletingRunnable.run(Future.scala:24)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[...]

PR #902, opened 2017-05-17
[Issue #782 ResNet Exception]Fix bug in Concat layer and Add backward to ConcatTable

psyyz10:

Fix bug in the `updateGradinput` method of Concat layer and add override `backward` to ConcatTable

## How was this patch tested?
Unit tests

fixed https://github.com/intel-analytics/BigDL/issues/782


yiheng:
LGTM

Issue #900, opened 2017-05-17
Use reflection to interact with BlockManger


yiheng:
BlockManger putBytes API looks like not so stable. Directly invoke it will result in multiple compiled versions for different spark platforms. Use reflection can reduce the compiled version number.

srowen:
(BlockManager isn't an API by the way - that's the problem.)
I think this might even be higher priority, as the current builds won't likely work with Spark 2.2 which will be out soon, and 0.2.0 will want to support that.

srowen:
I looked at this in depth today, and my first conclusion is that it would be quite hard to abstract over Spark 1.x vs 2.x given the use of private APIs. They have changed a lot and so it's probably going to be necessary to keep the two `1.5-plus` and `2.0` modules.

However, I don't see any variation in the versions of Spark 2.x _except_ for the `putBytes` method. It changed, then changed back between 2.1.0 - 2.1.1 - 2.2.0. So this one we really do need to use reflection to support, and I think that's easy.

This would let us get rid of all of the profiles except `spark_2.0`, because all 1.x versions should (?) be compatible with the base build, and all 2.x versions should then be compatible with the 2.0 build. I will try it now.

PR #907, opened 2017-05-17
Use reflection to call BlockManager.putBytes in Spark 2.x to handle different method signatures

srowen:

Use reflection to call `BlockManager.putBytes` in Spark 2.x to handle different method signatures.
This also removes the profiles, which only set versions now, which could be done on the command line. The profiles don't have to be removed, strictly, but won't be necessary to produce published Maven builds.

## How was this patch tested?

Existing tests pass with `-Pspark_2.0`. The same build works on Cloudera's distribution of Spark 2.1.0, which has a differing signature for `BlockManager.putBytes`

fixed https://github.com/intel-analytics/BigDL/issues/900

### DIFF
--- a/spark/spark-version/2.0/src/main/scala/org/apache/spark/storage/BlockManagerWrapper.scala
+++ b/spark/spark-version/2.0/src/main/scala/org/apache/spark/storage/BlockManagerWrapper.scala
@@ -32,7 +32,7 @@ object BlockManagerWrapper {
     require(bytes != null, "Bytes is null")
     val blockManager = SparkEnv.get.blockManager
     blockManager.removeBlock(blockId)
-    doPutBytes(blockId, new ChunkedByteBuffer(bytes), level)
+    putBytesFn(blockId, new ChunkedByteBuffer(bytes), level)
   }
 
   def getLocal(blockId: BlockId): Option[BlockResult] = {
@@ -61,7 +61,7 @@ object BlockManagerWrapper {
     }
   }
 
-  private val putBytesMethod = {
+  private val putBytesFn: (BlockId, ChunkedByteBuffer, StorageLevel) => Unit = {
     val bmClass = classOf[BlockManager]
     // Spark 2.0.0 - 2.1.0, and 2.2.0+ (as of this writing), declare the method:
     // def putBytes[T: ClassTag](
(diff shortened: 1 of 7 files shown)


yiheng:
The changes looks good to me. The putBytes is a quite intense method used in training. Let me run some perf test.

yiheng:
@dding3 Please take a look at this change.

srowen:
Agree, to make it as fast as possible I pulled out the logic that discovers the right method to call. The reflective call itself shouldn't be expensive and right now that's all that happens in the live path.

dding3:
LGTM. I did some performance test on this PR with spark2.0 and the result looks good. There is no increase in blockmanager related operations and no downgrade in the throughput.

BTW, I am thinking do we need update spark2.0 in spark-version to spark2.0-plus and make related changes in pom? I think it would be more clear.

srowen:
Ideally there are only ever modules and profiles for Spark 1 vs Spark 2. So maybe the ideal names are like spark-1.x and spark-2.x, but, anything similar works. Although we don't know whether 2.0-plus will work with Spark 3, maybe it's fine for symmetry with 1.5-plus, which is implicitly 1.5+ (but < 2.0).

I think the modules might shuffle around a bit anyway if we change packaging such that there are two publishes modules for Spark 1 vs 2, rather than 1 module that selects 1 of 2 dependencies.

srowen:
@dding3 if you'd like a specific change to the module names, let me know. I'm leaving it as-is for the moment.

dding3:
I think it's a good idea to rename it as spark-2.x. It will make less confusion if user want to use spark 2.1 but have to build bigdl with "-P spark_2.0"

srowen:
Done @dding3 . I know this would require wiki updates too. I can't open a PR for wiki pages even though there's a git repo (?) but the change is simple enough that I just attach a patch here:

[Update_build_page_to_reflect_spark_2_x_profile.patch.txt](https://github.com/intel-analytics/BigDL/files/1016105/Update_build_page_to_reflect_spark_2_x_profile.patch.txt)

dding3:
LGTM. Thanks for your contribution.

spark/spark-version/2.0/src/main/scala/org/apache/spark/storage/BlockManagerWrapper.scala
```diff
@@ -57,4 +60,48 @@ object BlockManagerWrapper {
...
+          classOf[Boolean], classOf[Boolean], classOf[ClassTag[_]])
+    }
+  }
+  private val putBytesArgCount = putBytesMethod.getParameterTypes.length
+
+  private def doPutBytes(
+      blockId: BlockId,
+      bytes: ChunkedByteBuffer,
+      level: StorageLevel): Unit = {
+    putBytesArgCount match {
```
yiheng: can we use a function object here instead of using match in every method call?

Something like this
```scala
  private val putBytes: (BlockId, ChunkedByteBuffer, StorageLevel) => Unit
      = putBytesMethod.getParameterTypes.length match {
    case 5 =>
      (blockId: BlockId, bytes: ChunkedByteBuffer, level: StorageLevel) =>
        putBytesMethod.invoke(SparkEnv.get.blockManager,
          blockId, bytes, level, JBoolean.TRUE, null)
    case 6 =>
      (blockId: BlockId, bytes: ChunkedByteBuffer, level: StorageLevel) =>
        putBytesMethod.invoke(SparkEnv.get.blockManager,
          blockId, bytes, level, JBoolean.TRUE, JBoolean.FALSE, null)
    case _ =>
      throw new IllegalArgumentException(s"invalid putBytesArgCount " +
        s"${putBytesMethod.getParameterTypes.length}")
  }

  private def doPutBytes(
      blockId: BlockId,
      bytes: ChunkedByteBuffer,
      level: StorageLevel): Unit = {
    putBytes(blockId, bytes, level)
  }
```

I write a trival program to test the performance difference, match spend a little more time than function object:
Sum 2000000, Match total time cost is 5.243864ms
Sum 2000000, Funtion object total time cost is 4.684357ms

srowen: Ah yeah good idea @yiheng , done.

yiheng:
jenkins pass and perf test looks good

srowen:
Thanks @yiheng . I think the wiki also needs to be updated, and I wasn't sure how to make a PR, because it's not forkable in github. Here's the change, attached.

yiheng:
@srowen Thanks. Has updated it.

Issue #874, opened 2017-05-11
Add maven artifacts for other spark platforms and use dependency only module to manage them


yiheng:
BigDL support running on Spark 1.5, 1.6, 2.0 and 2.1, but only artifacts for 1.6 and 2.0 are released to maven. The artifacts for other Spark platforms are expected.

**How to distinguish artifacts for different spark platforms?**
Currently, the artifacts are distinguished by names, e.g. bigdl is for Spark 1.6 and bigdl_Spark2.0 is for Spark 2.0. We can keep using the name conversions but a better way is to use classifier.

yiheng:
As discussed in the [mail list](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/bigdl-user-group/X6D9R3TA-1E), it's better to use a bunch of dependency only module to manage the published artifacts

srowen:
I will take a shot at making a change like this, and will open a PR to show a proposed change when it seems ready to look at.

srowen:
https://github.com/intel-analytics/BigDL/issues/900 _might_ make this obsolete, but probably not entirely. If we could remove the per-version dependencies we wouldn't have a need for multiple modules. Anyway, it should come before.

PR #908, opened 2017-05-18
Add a DependencyReducedPom back to get a fine pom to install/deploy

yiheng:

In the snapshot published pom, the properties in the pom dependency are not replaced by the value we specified in the building. This will cause problem when user use the published artifacts in maven. But the published 0.1.0 artifacts doesn't have the problem.

The root cause is we disable DependencyReducedPom when shading the bigdl artifact later in the master branch. When generating DependencyReducedPom, the property will be replaced in the generated pom. That's why 0.1.0 published pom is fine but master published pom has problem.

This PR will fix this issue by bringing back the DependencyReducedPom. And it will also bring the shade to the parent project and add a dummy project dependency in the parent project. So all the project will be shaded and all the pom will be processed to replace the property with its value.

This is how spark deal with this problem. See https://github.com/apache/spark/blob/master/pom.xml#L246

## How was this patch tested?

Unit test

This PR is related with #874

### DIFF
--- a/make-dist.sh
+++ b/make-dist.sh
@@ -30,11 +30,16 @@ else
     exit 1
 fi
 
+export MAVEN_OPTS="-Xmx2g -XX:ReservedCodeCacheSize=512m"
+
 if [[ "$_java" ]]; then
     version=$("$_java" -version 2>&1 | awk -F '"' '/version/ {print $2}')
     if [[ "$version" < "1.7" ]]; then
         echo Require a java version not lower than 1.7
         exit 1
+    fi
+    if [[ "$version" < "1.8" ]]; then
+        export MAVEN_OPTS="$MAVEN_OPTS -XX:MaxPermSize=1G"
     fi
 fi
 
--- a/make-dist.sh
+++ b/make-dist.sh
@@ -30,7 +30,7 @@ else
     exit 1
 fi
 
-export MAVEN_OPTS="-Xmx2g -XX:ReservedCodeCacheSize=512m"
+MVN_OPTS_LIST="-Xmx2g -XX:ReservedCodeCacheSize=512m"
 
 if [[ "$_java" ]]; then
     version=$("$_java" -version 2>&1 | awk -F '"' '/version/ {print $2}')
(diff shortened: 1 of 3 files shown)


yiheng:
Jenkins test pass

spark/pom.xml
```diff
@@ -18,4 +18,60 @@
...
+        <!--
+          This is a dummy dependency that is used to trigger the maven-shade plugin so that BigDL's
+          published POMs are flattened and do not contain variables. Without this dependency, some
+          subprojects' published POMs would contain variables like ${spark.version} that will
+          be substituted according to the default properties instead of the ones determined by the
+          profiles that were active during publishing. By ensuring that maven-shade runs for all
+          subprojects, we eliminate this problem because the substitutions are baked into the final POM.
+        -->
+        <dependency>
+            <groupId>org.spark-project.spark</groupId>
```
srowen: This shouldn't say spark-project, right?

spark/pom.xml
```diff
@@ -18,4 +18,60 @@
...
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
+                <version>3.0.0</version>
+                <configuration>
+                    <artifactSet>
+                        <includes>
+                            <include>org.spark-project.spark:unused</include>
+                            <include>com.google.protobuf</include>
```
srowen: Does this project actually include protobuf at all?

make-dist.sh
```diff
@@ -30,12 +30,17 @@ else
     exit 1
 fi
 
+export MAVEN_OPTS="-Xmx2g -XX:ReservedCodeCacheSize=512m"
```
srowen: I think you can write `export MAVEN_OPTS=${MAVEN_OPTS:-"..."}` to not overwrite the value if it's set already. Maybe you want to overwrite it every time though.

srowen: I think you can write `export MAVEN_OPTS=${MAVEN_OPTS:-"..."}` to not overwrite the value if it's set already. Maybe you want to overwrite it every time though.

spark/pom.xml
yiheng: one of the submodule includes protobuf dependency. It won't affects submodules without protobuf dependency.

make-dist.sh
yiheng: changed

yiheng: changed

spark/pom.xml
yiheng: Just 'borrow' the spark dummy project. Or we need to publish one to maven central. I think it's not so necessary. I will add more comments to explain this.

yiheng:
jenkins test pass

Issue #893, opened 2017-05-16
Python model print


SeaOfOcean:
print python model structure

better to use short name than full class name

PR #912, opened 2017-05-18
Refine model name

zhichao-li:
Previously the name of module is hardcode and cannot reflect the user input.
After the changes: 
```
        >>> conv2 = SpatialConvolution(6, 12, 5, 5).set_name("conv2")
        >>> print(conv2)
        conv2(6 -> 12, 5 x 5, 1, 1, 0, 0)
```

and use simple class name for the default value: ie. ``` com.intel.analytics.bigdl.nn.SoftMax@4df4dc0``` -> ``` SoftMax@4df4dc0```

## How was this patch tested?
manual

https://github.com/intel-analytics/BigDL/issues/893

### DIFF
--- a/pyspark/dl/nn/criterion.py
+++ b/pyspark/dl/nn/criterion.py
@@ -38,6 +38,9 @@ class Criterion(JavaValue):
             bigdl_type, JavaValue.jvm_class_constructor(self), *args)
         self.bigdl_type = bigdl_type
 
+    def __str__(self):
+        return self.value.toString()
+
     def forward(self, input, target):
         """
         NB: It's for debug only, please use optimizer.optimize() in production.
--- a/pyspark/dl/nn/layer.py
+++ b/pyspark/dl/nn/layer.py
@@ -44,6 +44,15 @@ class Model(JavaValue):
             bigdl_type, JavaValue.jvm_class_constructor(self), *args)
         self.bigdl_type = bigdl_type
 
+    def __str__(self):
+        """
+        >>> conv2 = SpatialConvolution(6, 12, 5, 5).set_name("conv2")
+        creating: createSpatialConvolution
+        >>> print(conv2)
+        conv2(6 -> 12, 5 x 5, 1, 1, 0, 0)
+        """
+        return self.value.toString()
+
     @classmethod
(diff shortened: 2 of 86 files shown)


zhichao-li:
cc @shane-huang  @SeaOfOcean  @qiuxin2012  any comments?

SeaOfOcean:
For SoftMax, getName will return com.intel.analytics.bigdl.nn.SoftMax@4df4dc0,
we should simplify it if no setName

zhichao-li:
@SeaOfOcean  this has been addressed as well, let me update the description.

yiheng:
I think it's different when you print the name and print the class.

If print the class, the class type, name and parameter; if print the name, should just return the name.

print class
```python
conv2 = SpatialConvolution(6, 12, 5, 5).set_name("conv2")
print(conv2)
com.intel.xxx.SpaitialConv[conv](6 -> 12, 5 x 5, 1, 1, 0, 0)
```

print name
```python
conv2 = SpatialConvolution(6, 12, 5, 5).set_name("conv2")
print(conv2.name)
conv2
```

zhichao-li:
What this pr going to address is ``` print class ```, I'm going to change like this:
```
conv2 = SpatialConvolution(6, 12, 5, 5).set_name("conv2")
print(conv2)
```
```
com.intel.xxx.SpaitialConv(6 -> 12, 5 x 5, 1, 1, 0, 0) 
->
SpaitialConv[conv2](6 -> 12, 5 x 5, 1, 1, 0, 0) 
```

zhichao-li:
cc @yiheng Just add a printName method which would responsible for the class printing. I'm going to merge this if no other comments.

yiheng:
printName should only print name, right? __toString__ should print class

zhichao-li:
No. getName would focus on the ``` name ``` itself (i.e. user can use this name to pinpoint a module), and ```printName``` is used to construct a printing format for a module.

yiheng:
If you want to print the module, why not override this method in python? 

```python
def __str__(self):
```

zhichao-li:
already did that.

yiheng:
then why need the printName method? User can
```python
print(layer.getName)
```
right?

PR #915, opened 2017-05-19
[branch-0.1]fix data may distribute unevenly in vgg example

qiuxin2012:

fix data may distribute unevenly in vgg example when the cores and executors is bigger than 20 * 64.

## How was this patch tested?

unit tests, manual tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/DataSet.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/DataSet.scala
@@ -326,9 +326,8 @@ object DataSet {
    */
   def array[T: ClassTag](localData: Array[T], sc: SparkContext): DistributedDataSet[T] = {
     val nodeNumber = Engine.nodeNumber()
-    val coreNumber = Engine.coreNumber()
     new CachedDistriDataSet[T](
-      sc.parallelize(localData, nodeNumber * coreNumber)
+      sc.parallelize(localData, nodeNumber)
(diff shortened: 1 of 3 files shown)


jason-dai:
Please test the examples on several large clusters too

qiuxin2012:
@jason-dai Just passed testing VGG and Lenet examples on a 20cores*64nodes cluster with a 5120 batchSize.

jason-dai:
Can you test other cluster configures - larger and/or smaller sizes?

qiuxin2012:
@jason-dai Also works for 28core*64node, 10core*5node.
1core*1node, 4code*1node will cause another error: 
`Exception in thread "main" org.apache.spark.SparkException: Job aborted due to stage failure: Serialized task 0:0 was 155763217 bytes, which exceeds max allowed: spark.rpc.message.maxSize (134217728 bytes). Consider increasing spark.rpc.message.maxSize or using broadcast variables for large values.`
adding an spark conf spark.rpc.message.maxSize=200 could solve this exception. (Current branch-0.1 also have this problem)

jason-dai:
We may need to add this config to the examples? @yiheng

yiheng:
agree to add the config to the example.

jason-dai:
How about ResNet? Other examples using Cifar-10 data?

qiuxin2012:
@jason-dai Yeah, Resnet also need this configuration. After checking all usages of the `array()`, I find lenet(mnist) rnn(tinyshakespeare) autoencoder(mnist) don't need this configuration.

PR #925, opened 2017-05-22
Some new Sample/MiniBatch implements, and SampleToMiniBatch transformer

### LINKED ISSUE #392: It is better to support Table format as feature and labels for a Sample.
In some cases, we need to use data in Table format as a Sample.

qiuxin2012:
Use an uniform Sample, ArraySample,  to store all Sample Data.

Use an ArrayTensorMiniBatch to host single/multi input data.
Add an SampleToMiniBatch transformer to transform Sample to miniBatch.  

## How was this patch tested?
unit tests

Fixed #392

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
@@ -18,9 +18,8 @@ package com.intel.analytics.bigdl.dataset
 
 import com.intel.analytics.bigdl.nn.abstractnn.Activity
 import com.intel.analytics.bigdl.tensor.Tensor
-import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+import com.intel.analytics.bigdl.utils.{T, Table}
 
-import scala.reflect.ClassTag
 
 /**
  * A interface for MiniBatch.
@@ -100,9 +99,102 @@ class TensorMiniBatch[T](
   }
 }
 
+/**
+ * A MiniBatch with [[com.intel.analytics.bigdl.utils.Table]] input and [[Tensor]] target.
+ * The size of first dimension in input and target should be the mini-batch size.
+ *b
+ * @param input input Table
+ * @param target target Tensor
+ * @tparam T Numeric type
+ */
+class ArrayTensorMiniBatch[T](
+      val input: Table,
+      val target: Tensor[T]) extends MiniBatch[T]{
+
+  override def size(): Int = {
+    input[Tensor[T]](1).size(1)
+  }
+
+  override def slice(offset: Int, length: Int): MiniBatch[T] = {
+    val t = T()
+    var b = 1
+    while(b <= input.length()) {
+      t(b) = input[Tensor[T]](b).narrow(1, offset, length)
+      b += 1
+    }
+    MiniBatch(t, target.narrow(1, offset, length))
+  }
+
+  override def getInput(): Activity = {
+    input
+  }
+
+  override def getTarget(): Activity = {
+    target
+  }
+}
+
+class UnlabeledTensorMiniBatch[T](
+      val input: Tensor[T]) extends MiniBatch[T]{
+
+  override def size(): Int = {
+    input.size(1)
+  }
+
+  override def slice(offset: Int, length: Int): MiniBatch[T] = {
+    MiniBatch(input.narrow(1, offset, length))
+  }
+
+  override def getInput(): Activity = {
+    input
+  }
+
+  override def getTarget(): Activity = {
+    throw new UnsupportedOperationException("UnlabeledTensorMiniBatch: no target in this MiniBatch")
+  }
+}
+
+class UnlabeledArrayTensorMiniBatch[T](
+      val input: Table) extends MiniBatch[T]{
+
+  override def size(): Int = {
+    input[Tensor[T]](1).size(1)
+  }
+
+  override def slice(offset: Int, length: Int): MiniBatch[T] = {
+    val t = T()
+    var b = 1
+    while(b <= input.length()) {
+      t(b) = input[Tensor[T]](b).narrow(1, offset, length)
+      b += 1
+    }
+    MiniBatch(t)
+  }
+
+  override def getInput(): Activity = {
+    input
+  }
+
+  override def getTarget(): Activity = {
+    throw new UnsupportedOperationException("No target in this UnlabeledArrayTensorMiniBatch")
+  }
+}
+
 object MiniBatch {
   def apply[T](input: Tensor[T], target: Tensor[T]): MiniBatch[T] = {
     new TensorMiniBatch[T](input, target)
+  }
+
+  def apply[T](input: Table, target: Tensor[T]): MiniBatch[T] = {
+    new ArrayTensorMiniBatch[T](input, target)
+  }
+
+  def apply[T](input: Tensor[T]): MiniBatch[T] = {
+    new UnlabeledTensorMiniBatch[T](input)
+  }
+
+  def apply[T](input: Table): MiniBatch[T] = {
+    new UnlabeledArrayTensorMiniBatch[T](input)
   }
 }
 
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
@@ -18,6 +18,9 @@ package com.intel.analytics.bigdl.dataset
 
 import com.intel.analytics.bigdl.nn.abstractnn.Activity
 import com.intel.analytics.bigdl.tensor.Tensor
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+
+import scala.reflect.ClassTag
 
 /**
  * A interface for MiniBatch.
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
@@ -101,8 +101,8 @@ class TensorMiniBatch[T](
 
 /**
  * A MiniBatch with [[com.intel.analytics.bigdl.utils.Table]] input and [[Tensor]] target.
- * The size of first dimension in input and target should be the mini-batch size.
- *b
+ * The size of first dimension in input's first tensor and target is the mini-batch size.
+ *
  * @param input input Table
  * @param target target Tensor
  * @tparam T Numeric type
@@ -134,6 +134,11 @@ class ArrayTensorMiniBatch[T](
   }
 }
 
+/**
+ * TensorMiniBatch without target.
+ * @param input input tensor of this MiniBatch
+ * @tparam T Numeric type
+ */
 class UnlabeledTensorMiniBatch[T](
       val input: Tensor[T]) extends MiniBatch[T]{
 
@@ -154,6 +159,11 @@ class UnlabeledTensorMiniBatch[T](
   }
 }
 
+/**
+ * ArrayTensorMiniBatch without target.
+ * @param input input table of this MiniBatch
+ * @tparam T Numeric type
+ */
 class UnlabeledArrayTensorMiniBatch[T](
       val input: Table) extends MiniBatch[T]{
 
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
@@ -17,8 +17,11 @@
 package com.intel.analytics.bigdl.dataset
 
 import com.intel.analytics.bigdl.nn.abstractnn.Activity
-import com.intel.analytics.bigdl.tensor.Tensor
+import com.intel.analytics.bigdl.tensor.{DoubleType, FloatType, Tensor}
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
 import com.intel.analytics.bigdl.utils.{T, Table}
+
+import scala.reflect.ClassTag
 
 
 /**
@@ -67,6 +70,8 @@ trait MiniBatch[T] extends Serializable{
       " Only support TensorMiniBatch.")
     this.asInstanceOf[TensorMiniBatch[T]].input
   }
+
+  def setValue(samples: Array[Sample[T]])(implicit ev: TensorNumeric[T]): this.type
 }
 
 /**
@@ -77,7 +82,7 @@ trait MiniBatch[T] extends Serializable{
  * @param target target Tensor
  * @tparam T Numeric type
  */
-class TensorMiniBatch[T](
+class TensorMiniBatch[T: ClassTag](
       val input: Tensor[T],
       val target: Tensor[T]) extends MiniBatch[T]{
   require(input.size(1) == target.size(1))
@@ -97,32 +102,66 @@ class TensorMiniBatch[T](
   override def getTarget(): Activity = {
     target
   }
+
+  override def setValue(samples: Array[Sample[T]])(implicit ev: TensorNumeric[T]): this.type = {
+    // todo
+    throw new UnsupportedOperationException("unimplemented")
+  }
+
 }
 
 /**
  * A MiniBatch with [[com.intel.analytics.bigdl.utils.Table]] input and [[Tensor]] target.
  * The size of first dimension in input's first tensor and target is the mini-batch size.
  *
- * @param input input Table
- * @param target target Tensor
+ * @param inputData input Table
+ * @param targetData target Tensor
  * @tparam T Numeric type
+ * @since 0.2.0
  */
-class ArrayTensorMiniBatch[T](
-      val input: Table,
-      val target: Tensor[T]) extends MiniBatch[T]{
+class ArrayTensorMiniBatch[T: ClassTag](
+    val inputData: Array[Tensor[T]],
+    val targetData: Array[Tensor[T]],
+    featurePadding: Option[Array[Tensor[T]]] = None,
+    featureFixedLength: Option[Array[Int]] = None,
+    featureIncrement: Option[Array[Int]] = None,
+    labelPadding: Option[Array[T]] = None,
+    labelFixedLength: Option[Array[Int]] = None,
+    labelIncrement: Option[Array[Int]] = None) extends MiniBatch[T]{
+  require(inputData.length > 0, "Input data in MiniBatch is empty.")
+  val input: Activity = if (inputData.length == 1) {
+    inputData(0)
+  } else {
+    T.array(inputData.map(_.asInstanceOf[Any]))
+  }
+
+  val target: Activity = if (targetData.length == 0) {
+    null
+  } else if (targetData.length == 1) {
+    targetData(0)
+  } else {
+    T.array(targetData.map(_.asInstanceOf[Any]))
+  }
 
   override def size(): Int = {
-    input[Tensor[T]](1).size(1)
+    inputData(0).size(1)
   }
 
   override def slice(offset: Int, length: Int): MiniBatch[T] = {
-    val t = T()
+    val is = new Array[Tensor[T]](inputData.length)
+    val ts = new Array[Tensor[T]](targetData.length)
     var b = 1
-    while(b <= input.length()) {
-      t(b) = input[Tensor[T]](b).narrow(1, offset, length)
+    while(b <= inputData.size) {
+      is(b) = inputData(b).narrow(1, offset, length)
       b += 1
     }
-    MiniBatch(t, target.narrow(1, offset, length))
+    b = 1
+    while(b <= targetData.size) {
+      ts(b) = inputData(b).narrow(1, offset, length)
+      b += 1
+    }
+
+    MiniBatch(is, ts)
   }
 
   override def getInput(): Activity = {
@@ -132,79 +171,276 @@ class ArrayTensorMiniBatch[T](
   override def getTarget(): Activity = {
     target
   }
-}
 
-/**
- * TensorMiniBatch without target.
- * @param input input tensor of this MiniBatch
- * @tparam T Numeric type
- */
-class UnlabeledTensorMiniBatch[T](
-      val input: Tensor[T]) extends MiniBatch[T]{
+  def setValue(samples: Array[Sample[T]])(implicit ev: TensorNumeric[T]): this.type = {
+    require(samples.length > 0)
+    require(samples(0).isInstanceOf[ArraySample[T]],
+      "ArrayTensorMiniBatch: Only support ArraySample")
+    val s = samples.map(_.asInstanceOf[ArraySample[T]])
 
-  override def size(): Int = {
-    input.size(1)
-  }
+    val longestFeature = MiniBatch.findLongestFeatures(samples)
+    val longestLabel = MiniBatch.findLongestLabels(samples)
 
-  override def slice(offset: Int, length: Int): MiniBatch[T] = {
-    MiniBatch(input.narrow(1, offset, length))
-  }
-
-  override def getInput(): Activity = {
-    input
-  }
-
-  override def getTarget(): Activity = {
-    throw new UnsupportedOperationException("UnlabeledTensorMiniBatch: no target in this MiniBatch")
-  }
-}
-
-/**
- * ArrayTensorMiniBatch without target.
- * @param input input table of this MiniBatch
- * @tparam T Numeric type
- */
-class UnlabeledArrayTensorMiniBatch[T](
-      val input: Table) extends MiniBatch[T]{
-
-  override def size(): Int = {
-    input[Tensor[T]](1).size(1)
-  }
-
-  override def slice(offset: Int, length: Int): MiniBatch[T] = {
-    val t = T()
-    var b = 1
-    while(b <= input.length()) {
-      t(b) = input[Tensor[T]](b).narrow(1, offset, length)
-      b += 1
-    }
-    MiniBatch(t)
-  }
-
-  override def getInput(): Activity = {
-    input
-  }
-
-  override def getTarget(): Activity = {
-    throw new UnsupportedOperationException("No target in this UnlabeledArrayTensorMiniBatch")
+    MiniBatch.arraySampleToMiniBatch(s, this, longestFeature, longestLabel, featurePadding,
+      featureFixedLength, featureIncrement, labelPadding, labelFixedLength, labelIncrement)
+    this
   }
 }
 
 object MiniBatch {
-  def apply[T](input: Tensor[T], target: Tensor[T]): MiniBatch[T] = {
-    new TensorMiniBatch[T](input, target)
+  def apply[T: ClassTag](
+    nInputs: Int,
+    nTargets: Int,
+    featurePadding: Option[Array[Tensor[T]]],
+    featureFixedLength: Option[Array[Int]],
+    featureIncrement: Option[Array[Int]],
+    labelPadding: Option[Array[T]],
+    labelFixedLength: Option[Array[Int]],
+    labelIncrement: Option[Array[Int]])(
+    implicit ev: TensorNumeric[T]): MiniBatch[T] = {
+
+    new ArrayTensorMiniBatch[T](Array.tabulate(nInputs)(_ => Tensor[T]()),
+      Array.tabulate(nTargets)(_ => Tensor[T]()),
+      featurePadding, featureFixedLength, featureIncrement,
+      labelPadding, labelFixedLength, labelIncrement)
   }
 
-  def apply[T](input: Table, target: Tensor[T]): MiniBatch[T] = {
+  def apply[T: ClassTag](input: Tensor[T], target: Tensor[T]): MiniBatch[T] = {
+    MiniBatch[T](Array(input), Array(target))
+  }
+
+  def apply[T: ClassTag](input: Array[Tensor[T]], target: Tensor[T]): MiniBatch[T] = {
+    MiniBatch[T](input, Array(target))
+  }
+
+  def apply[T: ClassTag](input: Array[Tensor[T]], target: Array[Tensor[T]]): MiniBatch[T] = {
     new ArrayTensorMiniBatch[T](input, target)
   }
 
-  def apply[T](input: Tensor[T]): MiniBatch[T] = {
-    new UnlabeledTensorMiniBatch[T](input)
+  def apply[T: ClassTag](input: Tensor[T]): MiniBatch[T] = {
+    MiniBatch[T](Array(input), new Array[Tensor[T]](0))
   }
 
-  def apply[T](input: Table): MiniBatch[T] = {
-    new UnlabeledArrayTensorMiniBatch[T](input)
+  def apply[T: ClassTag](input: Array[Tensor[T]]): MiniBatch[T] = {
+    MiniBatch[T](input, new Array[Tensor[T]](0))
+  }
+
+  private def resizeData[T: ClassTag](
+        data: Array[Tensor[T]],
+        sampleSize: Array[Array[Array[Int]]],
+        longestData: Array[Int],
+        dataFixedLength: Option[Array[Int]] = None,
+        dataIncrement: Option[Array[Int]] = None): Unit = {
+    val inputSize = new Array[Array[Int]](longestData.length)
+    var ii = 0
+    while (ii < inputSize.length) {
+      inputSize(ii) = Array(sampleSize.length) ++ sampleSize(longestData(ii))(ii)
+      ii += 1
+    }
+
+    var i = 0
+    while (i < inputSize.length) {
+      if (dataFixedLength.isDefined) {
+        val fixedLength = dataFixedLength.get(i)
+        require(fixedLength >= inputSize(i)(1),
+          s"${i}th FixedLength=${fixedLength} is smaller than its FeatureLength=${inputSize(i)(1)}")
+        inputSize(i)(1) = fixedLength
+      } else if (dataIncrement.isDefined) {
+        val increment = dataIncrement.get(i)
+        inputSize(i)(1) += increment
+      }
+      data(i).resize(inputSize(i))
+      i += 1
+    }
+
+  }
+
+  def arraySampleToMiniBatch[T: ClassTag](
+      samples: Array[ArraySample[T]],
+      miniBatch: ArrayTensorMiniBatch[T],
+      longestFeature: Array[Int],
+      longestLabel: Array[Int],
+      featurePadding: Option[Array[Tensor[T]]] = None,
+      featureFixedLength: Option[Array[Int]] = None,
+      featureIncrement: Option[Array[Int]] = None,
+      labelPadding: Option[Array[T]] = None,
+      labelFixedLength: Option[Array[Int]] = None,
+      labelIncrement: Option[Array[Int]] = None
+      )(implicit ev: TensorNumeric[T]): MiniBatch[T] = {
+    val inputs = miniBatch.inputData
+    val target = miniBatch.targetData
+
+    val featureSizes = samples.map(_.getFeatureSize())
+    val labelSizes = samples.map(_.getLabelSize())
+    val unlabeled = if (labelSizes.flatMap(_.flatMap(_.toIterator)).product == 0) true else false
+    resizeData(inputs, featureSizes,
+      longestFeature, featureFixedLength, featureIncrement)
+    if (!unlabeled) {
+      resizeData(target, labelSizes,
+        longestLabel, labelFixedLength, labelIncrement)
+    }
+
+    if (featurePadding.isDefined) {
+      // check if featurePadding is right.
+      var i = 0
+      while (i < inputs.length) {
+        require(featurePadding.get.length == inputs.length, s"Number of tensor padding should " +
+          s"equals to Number of feature tensor in Sample. Excepted ${inputs.length}," +
+          s" but got ${featurePadding.get.length}")
+        if (inputs(i).dim() == 2) {
+          require(featurePadding.get(i).nElement() == 1, s"${i}thFeature is 1D, featurePadding " +
+            s"should have only one element, but got ${featurePadding.get(i)}")
+        } else {
+          require(featurePadding.get(i).dim() == inputs(i).dim() - 2,
+            s"${i}thFeature's featurePadding should have the " +
+            s"same dimension with the feature in sample. Excepted: ${inputs(i).dim() - 2}, " +
+            s"but got ${featurePadding.get(i).dim()}")
+        }
+        require(featurePadding.get(i).isContiguous(), "featurePadding should be contiguous")
+        i += 1
+      }
+    }
+
+    // init input and target
+    if (labelPadding.isDefined) {
+      // fill target with labelPadding first.
+      var l = 0
+      while(l < target.length) {
+        target(l).fill(labelPadding.get(l))
+        l += 1
+      }
+    } else {
+      target.foreach(_.zero())
+    }
+    if (!featurePadding.isDefined) {
+      inputs.foreach(_.zero())
+    }
+
+    // Copy sample data to miniBatch
+    var s = 0
+    while (s < samples.length) {
+      var f = 0
+      var offset = 0
+      val featureSize = featureSizes(s)
+      while (f < inputs.length) {
+        val length = featureSize(f).product
+        if (featurePadding.isDefined) {
+          // copy data
+          copy(samples(s).getData(), offset,
+            length, inputs(f)(s + 1), featurePadding.get(f))
+        } else {
+          // copy data without padding.
+          copy(samples(s).getData(), offset,
+            length, inputs(f)(s + 1))
+        }
+        f += 1
+        offset += length
+      }
+
+      if (!unlabeled) {
+        var l = 0
+        val labelSize = labelSizes(s)
+        while (l < target.length) {
+          val length = labelSize(l).product
+          copy(samples(s).getData(), offset,
+            length, target(l)(s + 1))
+          l += 1
+          offset += length
+        }
+      }
+
+      s += 1
+    }
+
+    miniBatch
+  }
+
+  /**
+   * Find Sample in Array[Sample] who has the biggest featureLength
+   */
+  def findLongestFeatures[T: ClassTag](
+        samples: Array[Sample[T]])(implicit ev: TensorNumeric[T]): Array[Int] = {
+    val featureIndices =
+      new Array[Int](samples(0).featureLength().length).map(_ => 0)
+    var i = 1
+    while (i < samples.length) {
+      var j = 0
+      while (j < featureIndices.length) {
+        if (samples(i).featureLength()(j) > samples(featureIndices(j)).featureLength()(j)) {
+          featureIndices(j) = i
+        }
+        j += 1
+      }
+      i += 1
+    }
+    featureIndices
+  }
+
+  /**
+   * Find Sample in Array[Sample] who has the biggest labelLength
+   */
+  def findLongestLabels[T: ClassTag](
+        samples: Array[Sample[T]])(implicit ev: TensorNumeric[T]): Array[Int] = {
+    val labelIndices =
+      new Array[Int](samples(0).labelLength().length).map(_ => 0)
+    var i = 1
+    while (i < samples.length) {
+      var j = 0
+      while (j < labelIndices.length) {
+        if (samples(i).labelLength()(j) > samples(labelIndices(j)).labelLength()(j)) {
+          labelIndices(j) = i
+        }
+        j += 1
+      }
+      i += 1
+    }
+    labelIndices
+  }
+
+  /**
+   * Copy tensor src to tensor dest with a padding tensor.
+   */
+  private def copy[T: ClassTag](
+      src: Array[T],
+      offset: Int,
+      length: Int,
+      dest: Tensor[T],
+      paddingTensor: Tensor[T] = null)(implicit ev: TensorNumeric[T]): Unit = {
+    arrayCopy(src,
+      offset,
+      dest.storage().array(),
+      dest.storageOffset() - 1,
+      length)
+    if (null != paddingTensor) {
+      var j = length
+      while (j < dest.nElement()) {
+        arrayCopy(paddingTensor.storage().array(), paddingTensor.storageOffset() - 1,
+          dest.storage().array(), dest.storageOffset() - 1 + j, paddingTensor.nElement())
+        j += paddingTensor.nElement()
+      }
+    }
+  }
+
+  /**
+   * A wrapper for System.arraycopy
+   */
+  private def arrayCopy[T: ClassTag](
+      src: AnyRef,
+      srcPos: Int,
+      dest: AnyRef,
+      destPos: Int,
+      length: Int)(implicit ev: TensorNumeric[T]): Unit = {
+    ev.getType() match {
+      case DoubleType => Array.copy(src
+        .asInstanceOf[Array[Double]],
+        srcPos, dest
+          .asInstanceOf[Array[Double]], destPos, length)
+      case FloatType => System.arraycopy(src
+        .asInstanceOf[Array[Float]],
+        srcPos, dest
+          .asInstanceOf[Array[Float]], destPos, length)
+      case _ => throw new UnsupportedOperationException(s"Only Float/Double supported")
+    }
   }
 }
 
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
@@ -71,6 +71,11 @@ trait MiniBatch[T] extends Serializable{
     this.asInstanceOf[TensorMiniBatch[T]].input
   }
 
+  /**
+   * Set values to this miniBatch with a set of Sample.
+   * @param samples a set of Sample
+   * @return self
+   */
   def setValue(samples: Array[Sample[T]])(implicit ev: TensorNumeric[T]): this.type
 }
 
@@ -119,7 +124,7 @@ class TensorMiniBatch[T: ClassTag](
  * @tparam T Numeric type
  * @since 0.2.0
  */
-class ArrayTensorMiniBatch[T: ClassTag](
+private[bigdl] class ArrayTensorMiniBatch[T: ClassTag](
     val inputData: Array[Tensor[T]],
     val targetData: Array[Tensor[T]],
     featurePadding: Option[Array[Tensor[T]]] = None,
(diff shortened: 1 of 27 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -97,9 +99,112 @@ class TensorMiniBatch[T](
...
+  }
+}
+
+/**
+ * TensorMiniBatch without target.
+ * @param input input tensor of this MiniBatch
+ * @tparam T Numeric type
+ */
+class UnlabeledTensorMiniBatch[T](
+      val input: Tensor[T]) extends MiniBatch[T]{
```
yiheng: remove val?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -97,9 +99,112 @@ class TensorMiniBatch[T](
...
+/**
+ * A MiniBatch with [[com.intel.analytics.bigdl.utils.Table]] input and [[Tensor]] target.
+ * The size of first dimension in input's first tensor and target is the mini-batch size.
+ *
+ * @param input input Table
+ * @param target target Tensor
+ * @tparam T Numeric type
+ */
+class ArrayTensorMiniBatch[T](
+      val input: Table,
```
yiheng: remove val?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -97,9 +99,112 @@ class TensorMiniBatch[T](
...
+      val target: Tensor[T]) extends MiniBatch[T]{
```
yiheng: remove val?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -97,9 +99,112 @@ class TensorMiniBatch[T](
...
+  }
+}
+
+/**
+ * ArrayTensorMiniBatch without target.
+ * @param input input table of this MiniBatch
+ * @tparam T Numeric type
+ */
+class UnlabeledArrayTensorMiniBatch[T](
+      val input: Table) extends MiniBatch[T]{
```
yiheng: remove val?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -97,9 +99,112 @@ class TensorMiniBatch[T](
...
+    throw new UnsupportedOperationException("No target in this UnlabeledArrayTensorMiniBatch")
+  }
+}
+
 object MiniBatch {
   def apply[T](input: Tensor[T], target: Tensor[T]): MiniBatch[T] = {
     new TensorMiniBatch[T](input, target)
   }
+
+  def apply[T](input: Table, target: Tensor[T]): MiniBatch[T] = {
```
yiheng: input: Array[Tensor[T]]. I think Table is our internal data structure. We should not expose this to a public API, right?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -97,9 +99,112 @@ class TensorMiniBatch[T](
...
+    new ArrayTensorMiniBatch[T](input, target)
+  }
+
+  def apply[T](input: Tensor[T]): MiniBatch[T] = {
+    new UnlabeledTensorMiniBatch[T](input)
+  }
+
+  def apply[T](input: Table): MiniBatch[T] = {
```
yiheng: input: Array[Tensor[T]]. I think Table is our internal data structure. We should not expose this to a public API, right?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -97,9 +99,112 @@ class TensorMiniBatch[T](
...
 
+/**
+ * A MiniBatch with [[com.intel.analytics.bigdl.utils.Table]] input and [[Tensor]] target.
+ * The size of first dimension in input's first tensor and target is the mini-batch size.
+ *
+ * @param input input Table
+ * @param target target Tensor
+ * @tparam T Numeric type
+ */
+class ArrayTensorMiniBatch[T](
```
yiheng: make these constructor of the MiniBatch private in bigdl package

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
```diff
@@ -132,8 +143,187 @@ class TensorSample[T: ClassTag](
...
 
+/**
+ * A kind of sample. Feature is a tensor, and label is a double/float.
+ * @param featureTensor feature tensor
+ * @param labelValue label
+ * @tparam T numeric type
+ */
+class TensorTSample[T: ClassTag](
+    val featureTensor: Tensor[T],
+    var labelValue: T) extends Sample[T]{
```
yiheng: why var?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
```diff
@@ -35,6 +35,13 @@ abstract class Sample[T: ClassTag] extends Serializable {
   def featureLength(): Int
 
   /**
+   * Length of the feature.
```
yiheng: Make these constructor private in bigdl package

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
```diff
@@ -35,6 +35,13 @@ abstract class Sample[T: ClassTag] extends Serializable {
...
+   * This function could be used to find the longest label.
```
yiheng: should we provide one method to get feature and one method to get label?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -296,3 +299,601 @@ class SampleToBatch[T: ClassTag]
     }
   }
 }
+
+/**
+ * Convert a sequence of [[Sample]] to a sequence of [[MiniBatch]] through function toMiniBatch.
+ *
+ * @param totalBatch total batch size
+ * @param partitionNum partition number of dataset, default means partitionNum
```
yiheng: default means partitionNum equals executor number

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
qiuxin2012: For copy Sample.

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
yiheng: val input: Seq[Tensor]?

yiheng: Merge multiple miniBatch into one?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
yiheng: make it private?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
```diff
@@ -132,8 +143,187 @@ class TensorSample[T: ClassTag](
...
+  }
+
+  override def labelLength(): Int = {
+    labels.size(1)
+  }
+
+  override def copy(other: Sample[T]): this.type = {
+    require(other.isInstanceOf[ArrayTensorSample[T]], "Sample.copy: sample type not match.")
+    val s = other.asInstanceOf[ArrayTensorSample[T]]
+    require(s.features.length == features.length, "Sample.copy: sample type not match.")
```
yiheng: sample length not match?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -97,9 +99,112 @@ class TensorMiniBatch[T](
...
+  override def getTarget(): Activity = {
+    throw new UnsupportedOperationException("No target in this UnlabeledArrayTensorMiniBatch")
+  }
+}
+
 object MiniBatch {
   def apply[T](input: Tensor[T], target: Tensor[T]): MiniBatch[T] = {
     new TensorMiniBatch[T](input, target)
   }
+
```
yiheng: we don't need these if there's one MiniBatch, right? Just a 
def apply[T](inputSize: Int, outputSize: Int)

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -296,3 +299,601 @@ class SampleToBatch[T: ClassTag]
...
+  /**
+   * Apply an SampleToMiniBatch transformer.
+   *
+   * @param batchSize total batch size
+   * @param toMiniBatch array[Sample] to MiniBatch function
+   * @return
+   */
+  def apply[T: ClassTag](
+        batchSize : Int,
+        toMiniBatch : (Array[Sample[T]], Array[Tensor[T]], Array[Tensor[T]]) => MiniBatch[T]
```
yiheng: change the last two Array[Tensor] to minibatch?

yiheng: what about move toMiniBatch into MiniBatch

And pass in a MiniBatch to the SampleToMiniBatch

We can have a default one.

MiniBatch.copy(samples: Seq[Sample[T]])

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -296,3 +299,601 @@ class SampleToBatch[T: ClassTag]
...
+        featurePadding : Option[Tensor[T]] = None,
+        featureIndex: Int = 0,
+        labelPadding : Option[T] = None,
+        labelIndex: Int = 0,
+        fixedLength: Option[Int] = None)(implicit ev: TensorNumeric[T]): MiniBatch[T] = {
+    val input = buf1(0)
+    val target = buf2(0)
+
+    // Compute the input's Size and target's Size
+    val inputSize = Array(samples.length) ++ samples(featureIndex).featureTensor.size()
```
yiheng: create a lot of temp object here

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
yiheng: seal the Sample class

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -296,3 +299,601 @@ class SampleToBatch[T: ClassTag]
...
+   * @param labelPadding label padding value (by default None, meaning no label padding)
+   * @param fixedLength if padding, it specifies the second dimension of feature/label
+   *                    after padding. If has multi feature tensor, only pad the first one.
+   *                    (by default None, meaning the length after padding is set to the max
+   *                    length of feature/label in a mini-batch)
+   * @param ev numeric operator
+   * @tparam T numeric type
+   * @return MiniBatch
+   */
+  def samplesToMiniBatch[T: ClassTag](
```
yiheng: Move this to the MiniBatch

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
```diff
@@ -132,8 +143,187 @@ class TensorSample[T: ClassTag](
   }
 }
 
+/**
```
yiheng: I don't think we need multiple type of sample. For storage save purpose, we can have one single Array to store all the information. Array need less storage than (Tensor, T). Another benifit is you need not write so many sampleToBatch.

@jason-dai

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
```diff
@@ -132,8 +143,187 @@ class TensorSample[T: ClassTag](
...
+  override def labelLength(): Int = 0
+
+  override def copy(other: Sample[T]): this.type = {
+    require(other.isInstanceOf[UnlabeledArrayTensorSample[T]],
+      "Sample.copy: sample type not match.")
+    val s = other.asInstanceOf[UnlabeledArrayTensorSample[T]]
+    require(s.features.length == features.length, "Sample.copy: sample type not match.")
+    var i = 0
+    while (i < features.length) {
+      features(i).resizeAs(s.features(i)).copy(s.features(i))
```
yiheng: don't rely on tensor copy method here.

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
qiuxin2012: In my testing, use a data array[T] and a size array[Int], we can save about 60 bytes for each sample.

qiuxin2012:
@yiheng please review.

qiuxin2012:
jenkins passed

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -54,38 +59,75 @@ trait MiniBatch[T] extends Serializable{
 
   @deprecated("Old interface", "0.2.0")
   def data(): Tensor[T] = {
-    require(this.isInstanceOf[TensorMiniBatch[T]], "Deprecated method," +
-      " Only support TensorMiniBatch.")
-    this.asInstanceOf[TensorMiniBatch[T]].input
+    throw new UnsupportedOperationException("MiniBatch.data(): unimplemented deprecated method")
```
yiheng: please add comments to explain this.

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -54,38 +59,75 @@ trait MiniBatch[T] extends Serializable{
...
   }
 
   @deprecated("Old interface", "0.2.0")
   def labels(): Tensor[T] = {
-    require(this.isInstanceOf[TensorMiniBatch[T]], "Deprecated method," +
-      " Only support TensorMiniBatch.")
-    this.asInstanceOf[TensorMiniBatch[T]].input
+    throw new UnsupportedOperationException("MiniBatch.labels(): unimplemented deprecated method")
```
yiheng: please add comment to explain this.

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -54,38 +59,75 @@ trait MiniBatch[T] extends Serializable{
...
   }
+
+  /**
+   * Set values to this miniBatch with a set of Sample.
```
yiheng: -> Replace the original content of the MiniBatch with a set of Samples

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -54,38 +59,75 @@ trait MiniBatch[T] extends Serializable{
...
-class TensorMiniBatch[T](
-      val input: Tensor[T],
-      val target: Tensor[T]) extends MiniBatch[T]{
-  require(input.size(1) == target.size(1))
+private[bigdl] class ArrayTensorMiniBatch[T: ClassTag](
+    val inputData: Array[Tensor[T]],
+    val targetData: Array[Tensor[T]],
+    paddingParam: Option[PaddingParam[T]] = None) extends MiniBatch[T]{
+  require(inputData.length > 0, "Input data in MiniBatch is empty.")
+  private lazy val input: Activity = if (inputData.length == 1) {
```
yiheng: don't use lazy. It may have the performance impact on Spark.

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -54,38 +59,75 @@ trait MiniBatch[T] extends Serializable{
...
+    inputData(0)
+  } else {
+    T.array(inputData.map(_.asInstanceOf[Any]))
+  }
+
+  private lazy val target: Activity = if (targetData.length == 0) {
```
yiheng: don't use lazy

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -54,38 +59,75 @@ trait MiniBatch[T] extends Serializable{
...
+  }
 
   override def size(): Int = {
-    input.size(1)
+    inputData(0).size(1)
   }
 
   override def slice(offset: Int, length: Int): MiniBatch[T] = {
-    MiniBatch(input.narrow(1, offset, length), target.narrow(1, offset, length))
+    val is = new Array[Tensor[T]](inputData.length)
```
yiheng: what is the is/ts stand for? DO NOT use name like this

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -95,11 +137,299 @@ class TensorMiniBatch[T](
...
+
+    val longestFeature = MiniBatch.findLongestFeatures(samples)
+    val longestLabel = MiniBatch.findLongestLabels(samples)
+
+    if (paddingParam.isDefined) {
+      val padding = paddingParam.get
+      MiniBatch.arraySampleToMiniBatch(s, this, longestFeature, longestLabel,
+        padding.featurePadding, padding.featureFixedLength, padding.featureIncrement,
+        padding.labelPadding, padding.labelFixedLength, padding.labelIncrement)
+    }
```
yiheng: this is a bug

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -95,11 +137,299 @@ class TensorMiniBatch[T](
...
     target
   }
+
+  override def setValue(samples: Array[Sample[T]])(implicit ev: TensorNumeric[T]): this.type = {
+    require(samples.length > 0)
+    samples.foreach(s => require(s.isInstanceOf[ArraySample[T]],
+      "ArrayTensorMiniBatch: Only support ArraySample"))
+    val s = samples.map(_.asInstanceOf[ArraySample[T]])
+
+    val longestFeature = MiniBatch.findLongestFeatures(samples)
```
yiheng: move into the if?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -95,11 +137,299 @@ class TensorMiniBatch[T](
   override def getTarget(): Activity = {
     target
   }
+
+  override def setValue(samples: Array[Sample[T]])(implicit ev: TensorNumeric[T]): this.type = {
+    require(samples.length > 0)
+    samples.foreach(s => require(s.isInstanceOf[ArraySample[T]],
+      "ArrayTensorMiniBatch: Only support ArraySample"))
+    val s = samples.map(_.asInstanceOf[ArraySample[T]])
```
yiheng: Why ArraySample

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -95,11 +137,299 @@ class TensorMiniBatch[T](
...
+
+  private def resizeData[T: ClassTag](
+        data: Array[Tensor[T]],
+        sampleSize: Array[Array[Array[Int]]],
+        longestData: Array[Int],
+        dataFixedLength: Option[Array[Int]] = None,
+        dataIncrement: Option[Array[Int]] = None): Unit = {
+    val inputSize = new Array[Array[Int]](longestData.length)
+    var ii = 0
+    while (ii < inputSize.length) {
```
yiheng: This is hard to understand. Can you add some comments?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -95,11 +137,299 @@ class TensorMiniBatch[T](
...
+      inputSize(ii) = Array(sampleSize.length) ++ sampleSize(longestData(ii))(ii)
+      ii += 1
+    }
+
+    var i = 0
+    while (i < inputSize.length) {
```
yiheng: This is hard to understand. Can you add some comments?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
```diff
@@ -16,132 +16,343 @@
...
 abstract class Sample[T: ClassTag] extends Serializable {
   /**
-   * Length of the feature.
+   * Length of the features.
    * This function could be used to sort samples in [[DataSet]].
+   *
    * @return
    */
-  def featureLength(): Int
+  def featureLength(): Array[Int]
```
yiheng: add a new method
featureLength(i : Int): Int

and try to use it as possible, to avoid creat array every time

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
```diff
@@ -16,132 +16,343 @@
...
+
+  /**
+   * Length of the labels.
+   * This function could be used to find the longest label.
+   *
+   * @return
+   */
+  def labelLength(): Array[Int]
```
yiheng: add a new method
featureLength(i : Int): Int

and try to use it as possible, to avoid creat array every time

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
```diff
@@ -16,132 +16,343 @@
...
+  /**
+   * Get label tensor, for one label Sample only.
+   * @return label tensor
+   */
   @deprecated("Old interface", "0.2.0")
-  def label(): Tensor[T] = {
-    require(this.isInstanceOf[TensorSample[T]], "Deprecated method, Only support TensorSample.")
-    this.asInstanceOf[TensorSample[T]].labelTensor
+  def label()(implicit ev: TensorNumeric[T]): Tensor[T] = {
+    throw new UnsupportedOperationException("Sample.label(): unimplemented deprecated method")
```
yiheng: add a comment to explain this

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
```diff
@@ -16,132 +16,343 @@
...
+  /**
+   * Get feature tensor, for one feature Sample only.
+   * @return feature tensor
+   */
   @deprecated("Old interface", "0.2.0")
-  def feature(): Tensor[T] = {
-    require(this.isInstanceOf[TensorSample[T]], "Deprecated method, Only support TensorSample.")
-    this.asInstanceOf[TensorSample[T]].featureTensor
+  def feature()(implicit ev: TensorNumeric[T]): Tensor[T] = {
+    throw new UnsupportedOperationException("Sample.feature(): unimplemented deprecated method")
```
yiheng: add a comment to explain the weird implementation

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
```diff
@@ -16,132 +16,343 @@
...
 
 /**
  * Class that represents the features and labels of a data sample.
+ *
  * @tparam T numeric type
  */
 abstract class Sample[T: ClassTag] extends Serializable {
   /**
-   * Length of the feature.
+   * Length of the features.
```
yiheng: we only need on Sample now, right? User should not have the reason to develop their owns

yiheng: remove the abstract

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
```diff
@@ -16,132 +16,343 @@
...
+  def getData(): Array[T] = data
+
+  override def featureLength(): Array[Int] = {
+    require(null != featureSize, "featureSize is empty")
+    featureSize.map(_.product)
+  }
+
+  override def labelLength(): Array[Int] = {
+    if (null == labelSize) {
+      Array(0)
```
yiheng: don't create Array(0) every time, move it to the object

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
```diff
@@ -16,132 +16,343 @@
...
+      Array(Array(0))
+    }
   }
 
   override def copy(other: Sample[T]): this.type = {
-    require(other.isInstanceOf[TensorSample[T]], "Sample.copy: sample type not match.")
-    val s = other.asInstanceOf[TensorSample[T]]
-    featureTensor.resizeAs(s.featureTensor).copy(s.featureTensor)
-    labelTensor.resizeAs(s.labelTensor).copy(s.labelTensor)
+    require(other.isInstanceOf[ArraySample[T]], "Sample.copy: sample type does not match.")
```
yiheng: no need =

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
```diff
@@ -16,132 +16,343 @@
...
-  def numFeature(): Int = 1
+  @deprecated("Old interface", "0.2.0")
+  override def feature()(implicit ev: TensorNumeric[T]): Tensor[T] = {
+    require(featureSize.length == 1)
+    Tensor[T](Storage(data), 1, getFeatureSize()(0))
+  }
+
+  @deprecated("Old interface", "0.2.0")
+  override def label()(implicit ev: TensorNumeric[T]): Tensor[T] = {
+    require(labelSize.length == 1)
```
yiheng: add error message

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
```diff
@@ -16,132 +16,343 @@
...
+      i += 1
+    }
+
     this
   }
 
-  def numFeature(): Int = 1
+  @deprecated("Old interface", "0.2.0")
+  override def feature()(implicit ev: TensorNumeric[T]): Tensor[T] = {
+    require(featureSize.length == 1)
```
yiheng: add error message

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
```diff
@@ -16,132 +16,343 @@
...
   }
 
   override def copy(other: Sample[T]): this.type = {
-    require(other.isInstanceOf[TensorSample[T]], "Sample.copy: sample type not match.")
-    val s = other.asInstanceOf[TensorSample[T]]
-    featureTensor.resizeAs(s.featureTensor).copy(s.featureTensor)
-    labelTensor.resizeAs(s.labelTensor).copy(s.labelTensor)
+    require(other.isInstanceOf[ArraySample[T]], "Sample.copy: sample type does not match.")
+    val that = other.asInstanceOf[ArraySample[T]]
+    val thatDataLength = that.featureSize.map(_.product).sum +
```
yiheng: data.length?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -95,11 +137,299 @@ class TensorMiniBatch[T](
   override def getTarget(): Activity = {
     target
   }
+
+  override def setValue(samples: Array[Sample[T]])(implicit ev: TensorNumeric[T]): this.type = {
+    require(samples.length > 0)
```
yiheng: check the size of the array[Sample]

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -95,11 +137,299 @@ class TensorMiniBatch[T](
...
+ *                         (by default None, meaning no increment)
+ * @param labelPadding feature paddings for each target (by default None, meaning zero padding)
+ * @param labelFixedLength it specifies the length of each target after padding (by default
+ *                         None, meaning using the longest target length in the mini Batch)
+ * @param labelIncrement it specifies adding an constant length to the longest length
+ *                       of each target in each MiniBatch.
+ *                       (by default None, meaning no increment)
+ * @tparam T numeric type
+ */
+case class PaddingParam[T: ClassTag](
```
yiheng: no need to have both feature and label

yiheng: merge featureFixedLength and featureIncrement into one field

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/image/BGRImgToSample.scala
```diff
@@ -32,18 +32,19 @@ object BGRImgToSample {
...
-      if (featureBuffer.nElement() != 3 * img.height() * img.width()) {
-        featureBuffer.resize(3, img.height(), img.width())
+      labelBuffer(0).storage.array()(0) = img.label()
+      if (featureBuffer(0).nElement() != 3 * img.height() * img.width()) {
+        featureBuffer(0).resize(3, img.height(), img.width())
       }
 
-      img.copyTo(featureBuffer.storage().array(), 0, toRGB)
+      img.copyTo(featureBuffer(0).storage().array(), 0, toRGB)
+      buffer.set(featureBuffer, labelBuffer)
```
yiheng: why not set storage?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/image/GreyImgToSample.scala
```diff
@@ -32,19 +32,19 @@ object GreyImgToSample {
...
     prev.map(img => {
-      labelBuffer.storage.array()(0) = img.label()
+      labelBuffer(0).storage.array()(0) = img.label()
       featureSize(0) = img.height()
       featureSize(1) = img.width()
-      featureBuffer.set(Storage(img.content), sizes = featureSize)
+      featureBuffer(0).set(Storage(img.content), sizes = featureSize)
 
-      buffer
+      buffer.set(featureBuffer, labelBuffer)
```
yiheng: why not set storage?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/text/LabeledSentenceToSample.scala
```diff
@@ -116,7 +115,7 @@ class LabeledSentenceToSample[T: ClassTag](vocabLength: Int,
         i += 1
       }
 
-      buffer
+      Sample[T](feature, label)
```
yiheng: why remove the buffer?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Optimizer.scala
```diff
@@ -300,26 +300,93 @@ object Optimizer {
...
       _model = model,
-      dataset = (DataSet.rdd(sampleRDD) -> SampleToBatch(batchSize))
+      dataset = (DataSet.rdd(sampleRDD) -> SampleToMiniBatch(batchSize))
         .asInstanceOf[DistributedDataSet[MiniBatch[T]]],
       criterion = criterion
     ).asInstanceOf[Optimizer[T, MiniBatch[T]]]
   }
 
+  /**
+   * A apply an Optimizer who could apply padding to the Samples.
```
yiheng: typo?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/image/GreyImgToSample.scala
yiheng: don't use buffer, or if user want to cache RDD, they will get a incorrect result

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/image/BGRImgToSample.scala
yiheng: don't use buffer, or if user want to cache RDD, they will get a incorrect result

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
```diff
@@ -16,132 +16,343 @@
...
+
+  override def numLabel(): Int = {
+    if (null == labelSize) {
+      0
+    } else {
+      labelSize.length
+    }
+  }
+
+  override def set(
```
yiheng: remove it to make Sample immutable?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
yiheng: remove it to make Sample immutable?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
```diff
@@ -296,3 +302,122 @@ class SampleToBatch[T: ClassTag]
...
+
+  private val batchPerPartition = Utils.getBatchSize(totalBatch)
+  var miniBatchBuffer = miniBatch.orNull
+
+  override def apply(prev: Iterator[Sample[T]]): Iterator[MiniBatch[T]] = {
+    val batchSizePerPartition = batchPerPartition
+    new Iterator[MiniBatch[T]] {
+      private val batchSize = batchSizePerPartition
+
+      private val sampleData = new Array[Sample[T]](batchSize)
```
yiheng: no need this as if the sample is immutable

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Optimizer.scala
```diff
@@ -300,26 +300,93 @@ object Optimizer {
...
+  }
+
+  /**
+   * Apply an optimizer with User-Defined MiniBatch.
+   *
+   * @param model model will be optimizied
+   * @param sampleRDD training Samples
+   * @param criterion loss function
+   * @param batchSize mini batch size
+   * @param miniBatch An User-Defined MiniBatch to construct a mini batch.
```
yiheng: toMiniBatch?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Optimizer.scala
```diff
@@ -300,26 +300,93 @@ object Optimizer {
...
+   * @return an Optimizer
+   */
+  def apply[T: ClassTag](
+          model: Module[T],
+          sampleRDD: RDD[Sample[T]],
+          criterion: Criterion[T],
+          batchSize: Int,
+          miniBatch: MiniBatch[T]
```
yiheng: toMiniBatch?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -95,11 +137,299 @@ class TensorMiniBatch[T](
...
+    MiniBatch[T](Array(input), new Array[Tensor[T]](0))
+  }
+
+  def apply[T: ClassTag](input: Array[Tensor[T]]): MiniBatch[T] = {
+    MiniBatch[T](input, new Array[Tensor[T]](0))
+  }
+
+  private def resizeData[T: ClassTag](
+        data: Array[Tensor[T]],
+        sampleSize: Array[Array[Array[Int]]],
```
yiheng: can you add a comment for the Array[Array[Array[...]]]?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -95,11 +137,299 @@ class TensorMiniBatch[T](
...
+        longestData: Array[Int],
+        dataFixedLength: Option[Array[Int]] = None,
+        dataIncrement: Option[Array[Int]] = None): Unit = {
+    val inputSize = new Array[Array[Int]](longestData.length)
+    var ii = 0
```
yiheng: ii -> i

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -95,11 +137,299 @@ class TensorMiniBatch[T](
...
+        ev.arraycopy(paddingTensor.storage().array(), paddingTensor.storageOffset() - 1,
+          dest.storage().array(), dest.storageOffset() - 1 + j, paddingTensor.nElement())
+        j += paddingTensor.nElement()
+      }
+    }
   }
 }
 
+/**
+ * Padding param for ArrayTensorMiniBatch.
```
yiheng: you should tell user in the doc that it will append tensor to the first dimension

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
qiuxin2012: A buffer to read Sample from Iterator.

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
```diff
@@ -16,132 +16,306 @@
...
- * A kind of sample. Feature is a tensor, and label is a tensor too.
- * @param featureTensor feature tensor
- * @param labelTensor label tensor
- * @tparam T numeric type
+ * A kind of sample who use only one array
  */
-class TensorSample[T: ClassTag](
-      val featureTensor: Tensor[T],
-      val labelTensor: Tensor[T]) extends Sample[T] {
+private[bigdl] class ArraySample[T: ClassTag](
```
yiheng: should we merge ArraySample to Sample?

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/rnn/Train.scala
```diff
@@ -71,22 +71,23 @@ object Train {
...
       val trainSet = DataSet.rdd(tokens)
         .transform(TextToLabeledSentence[Float](dictionary))
         .transform(LabeledSentenceToSample[Float](totalVocabLength))
-        .transform(SampleToBatch[Float](batchSize = param.batchSize,
-          featurePadding = Some(padFeature),
-          labelPadding = Some(padLabel),
-          fixedLength = Some(maxTrainLength)))
+        .transform(SampleToMiniBatch[Float](param.batchSize,
+          Some(featurePadding), None))
+//          featurePadding = Some(padFeature),
```
yiheng: remove these comments

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Optimizer.scala
```diff
@@ -300,26 +300,64 @@ object Optimizer {
...
-    isInOrder: Boolean,
-    featurePadding : Option[Tensor[T]] = None,
-    labelPadding : Option[T] = None,
-    fixedLength: Option[Int] = None
-  )(implicit ev: TensorNumeric[T]): Optimizer[T, MiniBatch[T]] = {
+                          model: Module[T],
+                          sampleRDD: RDD[Sample[T]],
+                          criterion: Criterion[T],
+                          batchSize: Int,
+                          featurePaddingParam: Option[FeaturePaddingParam[T]],
```
yiheng: use function overload
apply(..., featurePadding, labelPadding)

apply(...)

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
```diff
@@ -16,132 +16,306 @@
...
+  override def getFeatureSize(): Array[Array[Int]] = {
+    featureSize
   }
 
-  def numFeature(): Int = 1
+  override def getLabelSize(): Array[Array[Int]] = {
+    if (null != labelSize) {
+      labelSize
+    } else {
+      Array(Array(0))
```
yiheng: move it to the object?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -27,6 +32,13 @@ import com.intel.analytics.bigdl.tensor.Tensor
  */
 trait MiniBatch[T] extends Serializable{
   /**
+   * Set a batchSize of this miniBatch.
+   * @param batchSize batch Size
+   * @return this
+   */
+  def setBatchSize(batchSize: Int): this.type
```
yiheng: remove this method

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -52,40 +64,115 @@ trait MiniBatch[T] extends Serializable{
...
-    this.asInstanceOf[TensorMiniBatch[T]].input
+    throw new UnsupportedOperationException("MiniBatch.labels(): unimplemented deprecated method")
   }
+
+  /**
+   * Replace the original content of the miniBatch with a set of Sample.
+   * @param samples a set of Sample
+   * @return self
+   */
+  def setValue(samples: Seq[Sample[T]])(implicit ev: TensorNumeric[T]): this.type
```
yiheng: setValue -> set?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -95,11 +182,407 @@ class TensorMiniBatch[T](
...
+
+    if (featurePaddingParam.isDefined || labelPaddingParam.isDefined) {
+      val longestFeature = MiniBatch.findLongestFeatures(samples)
+      val longestLabel = MiniBatch.findLongestLabels(samples)
+
+      MiniBatch.arraySampleToMiniBatch[T](samples, this,
+        Some(longestFeature), Some(longestLabel),
+        featurePadding, featurePaddingStrategy, labelPadding, labelPaddingStrategy)
+    } else {
+      MiniBatch.arraySampleToMiniBatch[T](samples, this)
```
yiheng: val longestFeature = if(featurePaddingParam.isDefined) {
    Some(MiniBatch.findLongestFeatures(samples))
} else {
    None
}

val longestLabel = if(featurePaddingParam.isDefined) {
    Some(MiniBatch.findLongestLabels(samples))
} else {
    None
}

MiniBatch.arraySampleToMiniBatch[T](samples, this, longestFeature, longestLabel, featurePadding, featurePaddingStrategy, labelPadding, labelPaddingStrategy)

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -95,11 +182,407 @@ class TensorMiniBatch[T](
...
+      longestLabel: Option[Array[Int]] = None,
+      featurePadding: Option[Array[Tensor[T]]] = None,
+      featurePaddingStrategy: Option[PaddingStrategy] = None,
+      labelPadding: Option[Array[T]] = None,
+      labelPaddingStrategy: Option[PaddingStrategy] = None
+      )(implicit ev: TensorNumeric[T]): MiniBatch[T] = {
+    val inputs = miniBatch.inputData
+    val targets = miniBatch.targetData
+
+    val featureSizes = samples.map(_.getFeatureSize())
```
yiheng: don't do the resize every time when there's no padding

yiheng: check size here

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -95,11 +182,407 @@ class TensorMiniBatch[T](
...
+ *  -1, -2, -3}
+ *
+ * @param paddingTensor paddings tensor for the first dimension(by default None,
+ *                       meaning zero padding).
+ * @param paddingStrategy See [[PaddingLongest]], [[FixedLength]]
+ * @tparam T numeric type
+ */
+case class FeaturePaddingParam[T: ClassTag](
+      paddingTensor: Option[Array[Tensor[T]]] = None,
+      paddingStrategy: Option[PaddingStrategy] = None)
```
yiheng: Add a defaultStrategy

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -95,11 +182,407 @@ class TensorMiniBatch[T](
...
+ * And the paddingValue is `0`, fixedLength is 4, the MiniBatch will be
+ * a tensor of 3*4:
+ * {1, 0, 0, 0
+ *  2, 3, 0, 0
+ *  4, 5, 6, 0}
+ *
+ * @param paddingValue padding value
+ * @param paddingStrategy padding strategy
+ */
+case class LabelPaddingParam[T: ClassTag](
```
yiheng: remove this as it is special case of the FeaturePaddingParam?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -95,11 +182,407 @@ class TensorMiniBatch[T](
...
+  private def resizeData[T: ClassTag](
+        data: Array[Tensor[T]],
+        // 1st Seq is batchSize, 2nd Array is number of features, 3th Array is feature size
+        sampleSize: Seq[Array[Array[Int]]],
+        longestData: Option[Array[Int]],
+        paddingStrategy: Option[PaddingStrategy] = None): Unit = {
+    // Size of input data. 1st Array is number of input, 2nd Array is input size.
+    val finalSizes = if (longestData.isDefined) {
+      val longest = longestData.get
+      val sizes = new Array[Array[Int]](longest.length)
```
yiheng: val sizes = new Array[Array[Int]](sampleSize(0).length)?
move size to outside?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -95,11 +182,407 @@ class TensorMiniBatch[T](
...
     target
   }
+
+  override def setValue(samples: Seq[Sample[T]])(implicit ev: TensorNumeric[T]): this.type = {
+    require(samples.length > 0, "samples is empty")
+    require(batchSize == 0 || samples.length <= batchSize, "setValue: samples's size doesn't " +
+      s"match mini batch size, excepted ${size()} got ${samples.length}")
+
+    if (featurePaddingParam.isDefined || labelPaddingParam.isDefined) {
+      val longestFeature = MiniBatch.findLongestFeatures(samples)
```
yiheng: Do we need findLongest if the strategy is fixlength?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
qiuxin2012: I set labelSize to null to save memory when no label in Sample. And getLabelSize will be called only once each iteration, I think it doesn't matter.

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
qiuxin2012: Yes, we should.   featurePaddingParam.isDefined means samples's size is different.

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
qiuxin2012: we need to resize when `batchSize != samples.length || featurePaddingParam.isDefined ||
      labelPaddingParam.isDefined`, am I right?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
yiheng: in the following code

```scala
sampleData(i) = sample.clone()

sampleData(i).copy(sample)
```

why we need them? as sample is immutable, why can it be copied?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
yiheng: You Should NEVER assume how a public method is called

yiheng: I mean move Array(Array(0)) to object, and treat it as a const value.

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
yiheng: yes

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala
qiuxin2012: I have changed it to `sampleData(i) = sample` last week. No copy any more. Now, `sampleData` is just an array to cache the samples' reference.

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
qiuxin2012: But Array(Array(0)) is not immutable, 
User may change the content:
```
scala> val a = Array(Array(0))
a: Array[Array[Int]] = Array(Array(0))

scala> a(0)
res0: Array[Int] = Array(0)

scala> a(0) = Array(2)

scala> a
res2: Array[Array[Int]] = Array(Array(2))
```
Return an new array is much safe.

qiuxin2012: I think I can add another function `hasLabel()`, and throw an exception here if the sample is unLabeled?

Issue #793, opened 2017-04-18
pyspark data read from mnist.read_data_sets method is not consistent with mnist.TRAIN_MEAN etc.


yangw1234:
Data read from pyspark from mnist.read_data_sets method is in the 0\~255 scale, but mnist.TRAIN_MEAN is in the 0\~1 scale.

PR #935, opened 2017-05-24
[issue #793] Fix mnist dataset

yangw1234:

`mnist.TRAIN_MEAN` .etc is calculated using mnist data of scale of 0\~1, however `mnist.read_data_sets` return data of  scale 0\~255. This pr fix this issue.

fix issue #793


yangw1234:
pr validation pass

PR #937, opened 2017-05-25
change version to 0.1.1-snapshot

yiheng:

Change version to 0.1.1-snapshot so customers can get a stable snapshot version to evaluate.

## How was this patch tested?
unit test


yiheng:
jenkins pass

PR #946, opened 2017-05-26
checkin more tensorflow protobuf classes

yiheng:

Check in more Tensorflow protobuf classes. So we can provide more features like fine-tune/inference Tensorflow models on Spark and save BigDL model in format which can be loaded by Tensorflow(e.g. inference on mobile).

**Why not user Tensorflow java API**
You cannot just import Tensorflow java API jar from maven and use it. You need to install Tensorflow on your machine. This brings unnecessary dependency.

Tensorflow Java API is not sufficient when we what to parse the model graph.

**Which version of Tensorflow are these codes generate from?
Tensorflow 1.0.0

**Why check in Java class files instead of protobuf files?
Don't want compile errors in IDE

**What if the class changed?
Small change should be back compatible. If necessary, we will check in new version Java classes.

## How was this patch tested?
unit test


yiheng:
jenkins job pass

PR #954, opened 2017-06-01
Make top 1 Accuracy support binary input (labels)

psyyz10:
Make top 1 Accuracy support binary input (labels), eg. the labels are 0, 1, 0, 0, 1, ...

## How was this patch tested?
Unit test

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/ValidationMethod.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/ValidationMethod.scala
@@ -113,7 +113,9 @@ class AccuracyResult(private var correct: Int, private var count: Int)
 /**
  * Caculate the percentage that output's max probability index equals target
  */
-class Top1Accuracy[T] extends ValidationMethod[T] {
+class Top1Accuracy[T](
+  implicit ev: TensorNumeric[T])
+  extends ValidationMethod[T] {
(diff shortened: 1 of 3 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/ValidationMethod.scala
```diff
@@ -122,7 +124,11 @@ class Top1Accuracy[T] extends ValidationMethod[T] {
     val _output = output.asInstanceOf[Tensor[T]]
     val _target = target.asInstanceOf[Tensor[T]]
```
JerryYanWan: require _output.dim <= 2 ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/ValidationMethod.scala
```diff
@@ -122,7 +124,11 @@ class Top1Accuracy[T] extends ValidationMethod[T] {
...
     if (_output.dim() == 2) {
-      _output.max(2)._2.squeeze().map(_target, (a, b) => {
+      (if (_output.size(2) == 1) {
+        _output.apply1(x => if (ev.isGreater(ev.fromType(0.5), x)) ev.zero else ev.one)
```
JerryYanWan: greater than 0.5, the output should be one?

PR #957, opened 2017-06-02
some new layer and existing layer refactor

yiheng:
Add new layers and existing layer refactor.

New layers:

- Const
A layer produce a const output and zero backward gradient.
- Fill
Input a 1D shape tensor and output a tensor filled with given value
- Pack
Stacks a list of n-dimensional tensors into one (n+1)-dimensional tensor
- Shape
Given input, return the shape of this input as a 1-D tensor
- SplitAndSelect
Split the given input tensor on specified dimension into specified sub-tensors then return the sub-tensor on the given index
- StrideSlice
Extracts a strided slice from the input tensor

Layer refactor:

- BatchNormalization
Allow pass in pre-created for weight/bias/weightGrad/biasGrad
- Graph
Allow pass in tensors serve as the parameters
- Linear
Allow pass in pre-created for weight/bias/weightGrad/biasGrad
- Mean
Allow keeping original dimension number after mean
- SpatialConvolution
Allow pass in pre-created for weight/bias/weightGrad/biasGrad
- Sum
Allow keeping original dimension number after Sum

## How was this patch tested?
Existing unit test and add new unit test


yangw1234:
LGTM. However, the changes in reset may conflict with the initMethod pr #960, which should we merge first?

pyspark/bigdl/nn/layer.py
```diff
@@ -2983,6 +2983,56 @@ def __init__(self,
                                                               n_input_plane,
                                                               JTensor.from_ndarray(kernel))
 
+class Const(Model):
+    '''
+    Return a constant tensor defined by value when forward
+    '''
```
zhichao-li: Unit test?

zhichao-li: https://github.com/intel-analytics/BigDL/issues/1040

Issue #25, opened 2016-09-27
Make Module:backward() final


jason-dai:
We should make `Module:backward` final, and overload the `updategradInput` and `accGradParameters` methods in its subclasses instead.

zhichao-li:
A similar one need to be final: https://github.com/intel-analytics/BigDL/issues/966

yiheng:
**Forward and backward should be final for all non-container layers**. The non-container layer should only override updateOutput(), updateGradInput() and accGrad(). Some common logic(e.g. bypass the layer/freeze the parameter) should be put into the forward/backward method and not be changed by the non-container layers.

Forward/Backward should be overridden in container layers. In container forward/backward method, the submodules should always call forward/backward method of the submodule.

PR #969, opened 2017-06-06
[Issue 450] add PropagateBackSwitch to disable some layers back propagation

SeaOfOcean:

add PropagateBackSwitch to disable some layers back propagation

## How was this patch tested?

unit test
https://github.com/intel-analytics/BigDL/issues/450


zhichao-li:
I was thinking if only want to disable backward, could we just simply add the the flag in ```AbstractModule.backward``` to control that?

SeaOfOcean:
That's one choice, but the problem is that the backward may be override, it is not easy to control the behavior in this case

zhichao-li:
For my understanding, ```backward``` is kind of like prototype method, we might not want to override it. In theory, customized logic can always go to  ```updateGradInput``` or ```accGradParameters``` if required.
Anyway, I'm ok to add another layer for this but its just too much at the first glance.

yiheng:
@zhichao-li There's an issue for make backward final #25

yiheng:
where is your test files

SeaOfOcean:
BackwardSwitchSpec

PR #973, opened 2017-06-06
Add bigdl as the top module

zhichao-li:

It's confuse and easy to conflict with other modules if without the bigdl prefix. 
We need to update the importing statement after this change:
i.e
```from nn.layer import Linear```
->
```from bigdl.nn.layer import Linear```

## How was this patch tested?
unit test


yiheng:
Update the documents?

yiheng:
Why checkin the dist folder? Any code review for this PR?

jason-dai:
BTW, should we backport the package change to 0.1.1?

PR #974, opened 2017-06-06
set default for updateHyperParameter

zhangxiaoli73:

set default for updateHyperParameter

## How was this patch tested?
unit test


zhangxiaoli73:
jenkins pass

PR #985, opened 2017-06-07
Support some layers working only in evaluate mode

SeaOfOcean:
Support some layers working only in evaluate mode

A->B->C->D

If I only want B C to work in evaluate mode(when do model validation),
then we can wrap B->C to a sequential and setEvaluateOnly(true)

## How was this patch tested?

unit test

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Sequential.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Sequential.scala
@@ -61,9 +61,7 @@ class Sequential[T: ClassTag]
   }
 
   override def updateGradInput(input: Activity, nextError: Activity): Activity = {
-    if (evaluateOnly && isTraining()) {
-      gradInput = nextError
-    } else {
+    if (!evaluateOnly || !isTraining()) {
       var i = modules.length - 1
       var error = nextError.asInstanceOf[Activity]
       while (i > 0) {
@@ -74,6 +72,8 @@ class Sequential[T: ClassTag]
       error = modules(0).updateGradInput(input, error)
 
       this.gradInput = error
+    } else {
+      gradInput = nextError
     }
 
     gradInput
@@ -100,9 +100,7 @@ class Sequential[T: ClassTag]
   }
 
   override def backward(input: Activity, nextError: Activity): Activity = {
-    if (evaluateOnly && isTraining()) {
-      gradInput = nextError
-    } else {
+    if (!evaluateOnly || !isTraining()) {
       var i = modules.length - 1
       var error = nextError.asInstanceOf[Activity]
       while (i > 0) {
@@ -113,6 +111,8 @@ class Sequential[T: ClassTag]
       error = modules(0).backward(input, error)
 
       this.gradInput = error
+    } else {
+      gradInput = nextError
     }
 
     gradInput
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Sequential.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Sequential.scala
@@ -33,8 +33,10 @@ class Sequential[T: ClassTag]
   private var evaluateOnly = false
 
   /**
-   * If evaluateOnly = true, then this sequential only works in evaluate mode, and not work in
-   * training mode
+   * If evaluateOnly = true,
+   * in training mode, the submodules will be skipped and input directly be passed to output;
+   * in evaluation mode, the modules will process the input tensors.
+   * If evaluateOnly = false, the modules will process the data in both mode.
    * @param status contral evaluateOnly
    * @return this
    */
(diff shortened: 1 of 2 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Sequential.scala
```diff
@@ -30,61 +30,91 @@ import scala.reflect.ClassTag
...
+  }
+
   override def updateOutput(input: Activity): Activity = {
-    var i = 0
-    var result = input.asInstanceOf[Activity]
-    while (i < modules.length) {
-      result = modules(i).forward(result)
-      i += 1
-    }
+    if (!evaluateOnly || !isTraining()) {
```
yiheng: If evaluateOnly = true, then this sequential only works in evaluate mode

if(!evaluateOnly || isTraining)???

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Sequential.scala
```diff
@@ -30,61 +30,91 @@ import scala.reflect.ClassTag
...
+      while (i < modules.length) {
+        result = modules(i).forward(result)
+        i += 1
+      }
 
-    this.output = result
+      this.output = result
+    } else {
+      // evaluateOnly == true and is in training mode
+      output = input
```
yiheng: If evaluateOnly = true, then this sequential only works in evaluate mode

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Sequential.scala
```diff
@@ -30,61 +30,91 @@ import scala.reflect.ClassTag
...
   }
 
   override def updateGradInput(input: Activity, nextError: Activity): Activity = {
-    var i = modules.length - 1
-    var error = nextError.asInstanceOf[Activity]
-    while (i > 0) {
-      val input = modules(i - 1).output
-      error = modules(i).updateGradInput(input, error)
-      i -= 1
+    if (evaluateOnly && isTraining()) {
```
yiheng: make if {} else {} consisitent across the methods

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Sequential.scala
SeaOfOcean: yes

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Sequential.scala
SeaOfOcean: If in training mode, we need to send the input to the layer after evaluteonly, e.g. A->evaluteOnly(B->C)->D, the input of B should give to D

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Sequential.scala
```diff
@@ -30,61 +30,91 @@ import scala.reflect.ClassTag
 class Sequential[T: ClassTag]
 (implicit ev: TensorNumeric[T]) extends Container[Activity, Activity, T] {
 
+  private var evaluateOnly = false
+
+  /**
+   * If evaluateOnly = true, then this sequential only works in evaluate mode, and not work in
```
yiheng: what do you mean not work?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Sequential.scala
yiheng: I think the document here is not clear. What about

If evaluateOnly = true, in training mode, the submodules will be skipped and input directly be passed to output; in evaluation mode, the modules will process the input tensors. If evaluateOnly = false, the modules will process the data in both mode.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Sequential.scala
SeaOfOcean: If evaluateOnly = true, when in training, it will not do layer forward (but do output = input, for layers after that), and will not do layer backward (but do gradInput = gradOutput, for layers before that)

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Sequential.scala
yiheng: Frankly, I think this should be a part of the layers. Add a container is complex.

SeaOfOcean: Yes, but modifying layers may not be general, since user may want to control the behavior of any layers. Or we can use a separate container to control this behavior

yiheng: Another solution is we provide some utility to modify the model easily.

Say in training, you want
LayerA -> LayerD

in evaluating, you want
LayerA -> LayerB -> LayerC -> LayerD

we can
```scala
val trainModel = Sequential().add(LayerA).add(LayerD)
// This is just an example
// relicate means only clone container, but the layers will be shared between both model
val testModel = trainModel.replicate().add(2, LayerB).add(3, LayerC)
```

SeaOfOcean: in this way, training model and test model are two different models, may not work in current optimizer when you want to validate the training model

PR #988, opened 2017-06-08
[FIX943] bash environment variables to java properties

### LINKED ISSUE #943: Remove all MKL ENV vars
MKL read environment variables to get configuration. These variables include MKL_DISABLE_FAST_MM, KMP_BLOCKTIME, OMP_WAIT_POLICY, OMP_NUM_THREADS. See https://github.com/intel-analytics/BigDL/blob/master/spark/dl/src/main/resources/spark-bigdl.conf.

As after JVM started, it seems no way to change the environment variable of the JVM process. So currently we need to set it before the JVM start.

However, MKL provides some methods to change the configurations. See:
https://software.intel.com/en-us/mkl-developer-reference-c-mkl-disable-fast-mm
omp.h
omp_set_num_threads, kmp_set_blocktime

Looks like there's no way to configure OMP_WAIT_POLICY.

Maybe another solution is when MKL class is loaded, use native method to set the process env vars?

i8run:

1. Convert environment variables to java properties.

+ `mkl_disable_fast_mm()` api to disable `MKL_DISABLE_FAST_MM`. 
+ `mkl_set_num_threads(1)` is `OMP_NUM_THREADS`
+ `kmp_set_blocktime()` is `KMP_BLOCKTIME`
+ `kmp_set_library_throughput()` is `OMP_WAIT_POLICY`

2. The relative java properties is blow,

| Property Name          | Default                                          | Meaning                                                                                              |
| ---------------------- | ------------------------------------------------ | ---------------------------------------------------------------------------------------------------- |
| bigdl.localMode        | false                                            | Runs without spark.                                                                                  |
| bigdl.coreNumber       | `Runtime.getRuntime().availableProcessors() / 2` | Runs with the number of cores user specified.                                                        |
| bigdl.mklBlockTime     | 0                                                | Uses by mkl library. Sets the time (milliseconds) that a mkl thread should wait before sleeping.     |
| bigdl.mklNumThreads    | 1                                                | Ues by mkl library. How many parallel threads omp will use. Do not suggest to change.                |
| bigdl.mklDisableFastMM | true                                             | Uses by mkl library. Turns off the MKL Memory Allocator if setted to true. Do not suggest to change. |
| bigdl.mklWaitPolicy    | passive                                          | Uses by mkl library. Sets omp wait policy to passive. Do not suggest to change.                      |

3. Deprecate `bigdl.sh`.

4. Because we have modify the BigDL-core, which contains the new JNI     methods called in MKL.java. So we should modify the bigdl.native to 0.2.0-SNAPSHOT in spark/dl/pom.xml.

## How was this patch tested?

It's test on jenkins of unit tests and perf tests.

fixed #943


yiheng:
Please also update the wiki page

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -66,7 +68,7 @@ object Engine {
    * make the Spark conf values changes transparent to you. However, if you use spark-shell or
    * Jupyter notebook, as the Spark context is created before your code, you have to
    * set them directly (through command line options or properties file)
-   *
+   *false
```
yiheng: remove the false here?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -137,10 +142,15 @@ object Engine {
...
+  @volatile private var _model: ThreadPool = new ThreadPool(1).setMKLThread(Env.mklNumThreads)
 
-  private def getCoreNumberFromEnv : Int = {
-    Option(System.getenv("DL_CORE_NUMBER")).map(_.toInt).getOrElse(getNumMachineCores)
+  /**
+   * return core number user defined.
+   */
+  private def getCoreNumberFromProperty() = {
+    val coreNumber = System.getProperty("bigdl.coreNumber", getNumMachineCores.toString).toInt
+    if (coreNumber > getNumMachineCores) getNumMachineCores
```
yiheng: if () {
} else {
}

BTW, I think we need to leave a comment here to explain why we do this. A log warning message may be better.

There's an assumption that user enable their HT, right? We may need to mention it in the comment.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -400,24 +409,46 @@ object Engine {
...
   }
+}
 
-  private def checkSysEnv() : Unit = {
-    // bigdl.disableCheckSysEnv is only for test purpose
-    if (System.getProperty("bigdl.disableCheckSysEnv") != null) {
-      return
+object Env {
+  val waitTimeMilliSec: Int = {
+    System.getProperty("bigdl.mklBlockTime", "0").toLowerCase(Locale.ROOT) match {
```
yiheng: leave an comment to explain why mklBlockTime = 0?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -400,24 +409,46 @@ object Engine {
...
     }
   }
+}
 
-  private def checkSysEnv() : Unit = {
-    // bigdl.disableCheckSysEnv is only for test purpose
-    if (System.getProperty("bigdl.disableCheckSysEnv") != null) {
-      return
+object Env {
+  val waitTimeMilliSec: Int = {
```
yiheng: all public method must have java doc

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -400,24 +409,46 @@ object Engine {
...
-      .filter(_._1.startsWith("spark.executorEnv."))
-      .map(c => (c._1.substring(18), c._2))
-      .foreach(env => {
-        require(System.getenv(env._1) != null,
-          s"Engine.init: Cannot find ${env._1} in environment variables. $ENV_VAR_ERROR")
-        require(System.getenv(env._1) == env._2,
-          s"Engine.init: Environment variable ${env._1} is ${System.getenv(env._1)}. " +
-          s"But it should be ${env._2}. $ENV_VAR_ERROR")
-    })
+  val mklNumThreads: Int = {
```
yiheng: all public method must have java doc

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -400,24 +409,46 @@ object Engine {
...
+    System.getProperty("bigdl.mklNumThreads", "1").toLowerCase(Locale.ROOT) match {
+      case x if x.toInt > 0 => x.toInt
+      case option =>
+        throw new IllegalArgumentException(s"Unknown option of $option for bigdl.mklNumThreads")
+    }
+  }
+
+  def mklDisableFastMM(): Unit = {
```
yiheng: all public method must have java doc

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -400,24 +409,46 @@ object Engine {
...
+    System.getProperty("bigdl.mklDisableFastMM", "true").toLowerCase(Locale.ROOT) match {
+      case "true" => MKL.disableFastMM()
+      case "fasle" =>
+      case option =>
+        throw new IllegalArgumentException(s"Unknown option of $option for bigdl.mklDisableFastMM")
+    }
   }
 
-  checkSysEnv()
+  def waitPolicyPassive(): Unit = {
```
yiheng: all public method must have java doc

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -400,24 +409,46 @@ object Engine {
...
+    System.getProperty("bigdl.mklWaitPolicy", "passive").toLowerCase(Locale.ROOT) match {
+      case "passive" => MKL.waitPolicyPassive()
+      case option =>
+        throw new IllegalArgumentException(s"Unknown option of $option for bigdl.waitPolicy")
+    }
+  }
+
+  def setMKLEnv(): Unit = {
```
yiheng: all public method must have java doc

yiheng: You call setMKLEnv in Engine.init, which will not be called in the executors. So all executor won't set the MKL env. This is wrong.

Remove the setMKLEnv(). Call this piece of code When the class is loaded to JVM.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
yiheng: private?

yiheng: Please put some comments here to explain why need to set the waitPolicyPassive

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
yiheng: pirvate?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
yiheng: The bigdl.mklWaitPolicy will never take affect in the executor side.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
yiheng: The change of bigdl.mklDisableFastMM will never take affect in the executor side.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
yiheng: The change of bigdl.mklNumThreads will never take affect in the executor side.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
yiheng: The change of bigdl. mklBlockTime will never take affect in the executor side.

yiheng:
The change of these jvm property won't take affect on executor side

yiheng:
also leave some comments in the bigdl-core PR

yiheng:
@qiuxin2012  please take a look at this PR

.gitignore
```diff
@@ -31,3 +31,4 @@ state*.[0-9]*
 # other
 nohup.out
 *.log
+*.un~
```
srowen: (BTW what is this for?)

pyspark/test/resources/conf/test_spark-bigdl.conf
```diff
@@ -27,16 +27,6 @@
 # https://github.com/intel-analytics/BigDL/wiki/Programming-Guide#engine
 #
 
-spark.executorEnv.DL_ENGINE_TYPE                    mklblas
```
srowen: It would be great to be able to remove this, but doesn't the MKL .so need to read these env variables? if there's another way to do it that's great

spark/dl/pom.xml
```diff
@@ -51,7 +51,7 @@
         <dependency>
             <groupId>com.intel.analytics.bigdl.native</groupId>
             <artifactId>${mkl-java-os-version}</artifactId>
-            <version>0.1.0</version>
+            <version>0.2.0-SNAPSHOT</version>
```
srowen: Was this unintentional? I recall that this was fixed to 0.1.0 on purpose

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
qiuxin2012: Just for discussion, I think setMklEnv should be called after loading MKL. So should this object be moved to BigDL-Core and called in MKL.java? @yiheng @i8run

spark/dl/pom.xml
i8run: Yes. I forgot changing it back.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
i8run: @qiuxin2012 Do you mean calling these code after loading dynamic library in MKL.java? So does this will also be called on executor side?

qiuxin2012: @i8run Yes, you are right.

.gitignore
i8run: Oh, it's the temporary file of some editors.

pyspark/test/resources/conf/test_spark-bigdl.conf
i8run: MKL and OMP support some apis for relative environment variables. So we can set it after loading mkl library.

spark/dl/pom.xml
i8run: Currently, we should modify the bigdl.native to 0.2.0-SNAPSHOT in spark/dl/pom.xml because of changed BigDL-core.

yiheng:
LGTM

qiuxin2012:
LGTM

PR #994, opened 2017-06-09
Add local integration test

zhichao-li:

Go through examples/notebook locally to ensure pr changes would not break the existing examples/notebook.
The scope of this local integration is just to make sure examples can be run successfully(i.e compatible between python2 and python3) and the API accuracy verification should be covered unit test.

## How was this patch tested?
jenkins test

### DIFF
--- a/pyspark/example/functional_api_forward_backward.ipynb
+++ b/pyspark/example/functional_api_forward_backward.ipynb
@@ -1,0 +1,89 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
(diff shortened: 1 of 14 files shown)


zhichao-li:
Would enable this once Jenkins environment ready

pyspark/example/functional_api_forward_backward.ipynb
```diff
@@ -0,0 +1,89 @@
+{
```
jason-dai: Why add the example here? Move it to bigdl-tutorials.

PR #997, opened 2017-06-12
integration-test change lenet dir to bigdl

luchy0120:

change dl to bigdl

## How was this patch tested?

integration test


yiheng:
LGTM

PR #998, opened 2017-06-12
add try/catch to invoke and invoke2

qiuxin2012:

add try/catch to invoke and invoke2

## How was this patch tested?

unit tests, manual test


yiheng:
LGTM

PR #1000, opened 2017-06-12
update pom.xml, separate shade-plugin to dl and spark-version

qiuxin2012:

update pom.xml, separate shade-plugin to dl and spark-version

## How was this patch tested?

unit tests, manual tests

### DIFF
--- a/spark/dl/pom.xml
+++ b/spark/dl/pom.xml
@@ -151,6 +151,43 @@
             </plugin>
 
             <plugin>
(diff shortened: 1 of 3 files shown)


yiheng:
LGTM

qiuxin2012:
@yiheng  I think we need a backport for this.

yiheng:
@qiuxin2012 I think so

PR #1010, opened 2017-06-14
Remove hard code spark version as 2.0 in pom

dding3:
Remove hard code spark version as 2.0 in pom

## How was this patch tested?
manual run lenet with spark 2.1.1 and spark 2.1.0


yiheng:
LGTM

PR #1011, opened 2017-06-14
[Issue 1007] load caffe model from hdfs

SeaOfOcean:

load caffe model from hdfs

## How was this patch tested?

unit test, integration test

fixed https://github.com/intel-analytics/BigDL/issues/1007

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/CaffeLoader.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/CaffeLoader.scala
@@ -60,7 +60,7 @@ class CaffeLoader[T: ClassTag](prototxtPath: String, modelPath: String,
     }
   }
 
-  private def loadFromHdfs(fileName: String): FSDataInputStream = {
+  private def createHdfsInputStream(fileName: String): FSDataInputStream = {
     val src = new Path(fileName)
     val fs = src.getFileSystem(new Configuration())
     fs.open(src)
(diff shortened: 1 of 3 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/CaffeLoader.scala
```diff
@@ -16,7 +16,7 @@
 
 package com.intel.analytics.bigdl.utils
 
-import java.io.{FileInputStream, InputStreamReader}
+import java.io._
```
wzhongyuan: One of these 2 lines can be removed ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/CaffeLoader.scala
```diff
@@ -58,18 +60,41 @@ class CaffeLoader[T: ClassTag](prototxtPath: String, modelPath: String,
     }
   }
 
+  private def loadFromHdfs(fileName: String): FSDataInputStream = {
```
wzhongyuan: rename to another meaningful method name ?  this method just create the inputstream

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/CaffeLoader.scala
SeaOfOcean: import java.io.{FileInputStream, InputStreamReader} is replaced by import java.io._

SeaOfOcean:
pr validate and integration test pass

wzhongyuan:
@SeaOfOcean  LGTM, please let me know when it's merged, then I will resolve the conflict and send the update

PR #1015, opened 2017-06-14
fix concat return incorrect result when output is 1D(no batch)

### LINKED ISSUE #1014: Concat will return wrong result when output's dimension=1
```scala
val model = Concat[Float](1)
model.add(Identity[Float]())
model.add(Identity[Float]())
print(model.forward(Tensor[Float].range(1, 3, 1)))
```
Get result 
```
0.0
0.0
0.0
0.0
0.0
0.0
[com.intel.analytics.bigdl.tensor.DenseTensor of size 6]
```

Correct one is 
```
1.0
2.0
3.0
1.0
2.0
3.0
[com.intel.analytics.bigdl.tensor.DenseTensor of size 6]
```

qiuxin2012:

fix concat return incorrect result when output is 1D

## How was this patch tested?

unit tests

fixed #1014


yiheng:
LGTM

PR #1017, opened 2017-06-15
Add default spark config

zhichao-li:
Previously we use ``` SparkConf()``` as the default value, but actually the correct one should be ``` create_spark_conf()``` which would combining bigdl config into the default spark conf


zhichao-li:
Jenkins passed

PR #1028, opened 2017-06-16
using set instead of copy for output and gradInput

JerryYanWan:
Optimize Recurrent layer; set storage of cell.output to recurrent.output

## How was this patch tested?
Unit test

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
@@ -208,7 +208,7 @@ class Recurrent[T : ClassTag]()
 
   override def accGradParameters(input: Tensor[T], gradOutput: Tensor[T],
                                  scale: Double = 1.0): Unit = {
-    cellAppendStartIdx = times
+    cellAppendStartIdx = cells.length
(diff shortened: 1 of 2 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -221,10 +247,15 @@ class Recurrent[T : ClassTag]()
         else hidden
       _input(inputDim) = input.select(timeDim, i)
       cells(i - 1).updateGradInput(_input, currentGradOutput)
-      gradInput.select(timeDim, i).copy(cells(i - 1).gradInput.toTable(inputDim))
       currentGradOutput(hidDim) = cells(i - 1).gradInput.toTable(hidDim)
       i -= 1
     }
+    if (cellAppendStartIdx == 0 || cellAppendStartIdx < times) {
+      set(cells.slice(cellAppendStartIdx, times)
+        .map(x => x.gradInput.toTable(inputDim).asInstanceOf[Tensor[T]]),
```
qiuxin2012: x.gradInput.toTable()[Tensor[T]](inputDim)

qiuxin2012:
LGTM

PR #1041, opened 2017-06-20
Convert 1-dimensional list into Table instead of Tensor

zhichao-li:
Before:
[ndarray.ones(5)] -> Tensor(5)
After:
[ndarray.ones(5)] -> Table(1-> Tensor(5))

Ideally, we should unify the input of forward/backward to accept List[Tensor] instead of Tensor/Table.

## How was this patch tested?
unittest


zhichao-li:
@yiheng  @wzhongyuan

zhichao-li:
Jenkins passed

PR #1045, opened 2017-06-20
remove tf layers from python api

yangw1234:

Const, Fill, SplitAndSelect, Shape

The above layers are created to match the tf counterparts and should not be used by user directly. Remove them from python api. 

## How was this patch tested?
add unit tests, modify existing tests


yangw1234:
pr validation pass

zhichao-li:
LGTM

PR #1046, opened 2017-06-20
fix: localMode should be false in HDFSSpec

i8run:

localMode should be false in HDFSSpec

## How was this patch tested?

jenkins pr validation and integrating tests.


zhangxiaoli73:
LGTM

PR #1069, opened 2017-06-22
rename apply to inputs when define connection in Graph API

yiheng:
In functional model definition, rename the apply to inputs. In the beginning, the method to define an edge
is named as apply, we'd like write code like
```scala
val y = Linear(3, 4)(x)
```
However, there's an implicit parameter in the Linear.apply function. So that code can not compile.

Now we change it to
```scala
val y = Linear(3, 4).inputs(x)
```

This change won't affect the python functional method
## How was this patch tested?
Unit test


qiuxin2012:
LGTM

PR #1080, opened 2017-06-23
add docs

yangw1234:

add docs
**Loss Functions**

- ClassSimplexCriterion.md
- MSECriterion.md
- SoftMarginCriterion.md

**Non-linear Activations**

- SoftMax.md

**Normalization Layers**

- BatchNormalization.md

**Recurrent Layers**

- GRU.md

- RNN.md

**Simple Layers**

- Log.md

- Min.md

- MulConstant.md

- Reshape.md

**optim**

- LBFGS.md

### DIFF
--- a/docs/docs/APIdocs/layers/nn/Loss Functions/ClassSimplexCriterion.md
+++ b/docs/docs/APIdocs/layers/nn/Loss Functions/ClassSimplexCriterion.md
@@ -1,0 +1,60 @@
+## ClassSimplexCriterion ##
+
+**Scala:**
+```scala
+val criterion = ClassSimplexCriterion(nClasses)
+```
+**Python:**
+```python
+criterion = ClassSimplexCriterion(nClasses)
+```
+
+ClassSimplexCriterion implements a criterion for classification.
+It learns an embedding per class, where each class' embedding is a
+point on an (N-1)-dimensional simplex, where N is the number of classes.
+
+**Parameters:**
+* **nClasses** - An integer, the number of classes.
+
+**Scala example:**
+```scala
+import com.intel.analytics.bigdl.nn._
+import com.intel.analytics.bigdl.utils.T
+import com.intel.analytics.bigdl.tensor.Tensor
+val criterion = ClassSimplexCriterion[Float](5)
+val input = Tensor[Float](T(
+ T(1.0f, 2.0f, 3.0f, 4.0f, 5.0f),
+ T(4.0f, 5.0f, 6.0f, 7.0f, 8.0f)
+))
+val target = Tensor[Float](2)
+target(Array(1)) = 2.0f
+target(Array(2)) = 1.0f
+val output = criterion.forward(input, target)
+val gradient = criterion.backward(input, target)
+-> print(output)
+23.562702
+-> print(gradient)
+0.25    0.20635083      0.6     0.8     1.0     
+0.6     1.0     1.2     1.4     1.6     
+[com.intel.analytics.bigdl.tensor.DenseTensor of size 2x5]
+```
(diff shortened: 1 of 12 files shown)


docs/docs/APIdocs/layers/optim/LBFGS.md
```diff
@@ -0,0 +1,28 @@
+**Scala:**
+```scala
+val optimMethod = new LBFGS(maxIter)
+```
+
+**Python:**
+```python
+val optim_method = LBFGS(maxIter, maxEval, tolFun, tolX, learningRate, lineSearch, lineSearchOptions)
```
zhichao-li: add an example here on how to integrated with optimizer?

docs/docs/APIdocs/layers/nn/Simple Layers/Reshape.md
```diff
@@ -0,0 +1,50 @@
...
+**Python:**
+```python
+reshape = Reshape(size, batch_mode)
+```
+
+The `forward(input)` reshape the input tensor into `size(0) * size(1) * ...` tensor,
+taking the elements row-wise.
+
+**parameters**
+* **size** - the size after reshape
```
zhichao-li: Not sure if we need to add the parameters details here as user can always get those from API doc. My concern is if we put those parameter descriptions here, it would be hard to sync with the java doc or python doc, and also the naming convention is diff between scala and python. @shane-huang

docs/docs/APIdocs/layers/nn/Simple Layers/Min.md
```diff
@@ -0,0 +1,67 @@
...
+min = Min(2)
+input = np.array([
+  [1.0, 2.0],
+  [3.0, 4.0]
+])
+
+target = np.array([1.0, 1.0])
+output = min.forward(input)
+gradient = min.backward(input, target)
+-> print output
```
zhichao-li: the other guys would print the result at the end of the example code instead of mixed between them. Anyway, I think it's acceptable written in this style. @shane-huang

docs/docs/APIdocs/layers/nn/Loss Functions/ClassSimplexCriterion.md
```diff
@@ -0,0 +1,62 @@
...
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric.NumericFloat
+
+val criterion = ClassSimplexCriterion(5)
+val input = Tensor(T(
+ T(1.0f, 2.0f, 3.0f, 4.0f, 5.0f),
+ T(4.0f, 5.0f, 6.0f, 7.0f, 8.0f)
+))
+val target = Tensor(2)
+target(Array(1)) = 2.0f
+target(Array(2)) = 1.0f
```
JerryYanWan: `target(1) = 2.0f target(2) = 1.0f`

docs/docs/APIdocs/layers/nn/Recurrent Layers/GRU.md
```diff
@@ -0,0 +1,105 @@
...
+  T(2.0f, 3.0f),
+  T(4.0f, 5.0f)
+)).resize(Array(1, seqLength, inputSize))
+val rec = Recurrent()
+
+val model = Sequential()
+    .add(rec.add(GRU(inputSize, hiddenSize)))
+    .add(TimeDistributed(Linear(hiddenSize, outputSize)))
+val output = model.forward(input)
+val gradient = model.backward(input, target)
```
JerryYanWan: No criterion? Should the target be defined as gradOutput ?

PR #964, opened 2017-06-05
Caffe Model Support - Load Caffe Dynamically from Prototxt and Binary

wzhongyuan:
## What changes were proposed in this pull request?
1) Load Caffe Dynamically from prototxt and Binary files
2) Support both V1 and V2 caffe layer types
3) Support customized layer
4) Load Criterion layers/functions

## How was this patch tested?
Unit testing and common models

PR #1095, opened 2017-06-26
Caffe persister

wzhongyuan:

This is a follwing PR for https://github.com/intel-analytics/BigDL/pull/964
1) Save BigDL module to Caffe compatible model dynamically
2) Save both prototxt and binary weights
3) Support both V1 and V2 layer versions

## How was this patch tested?
Unit test

https://github.com/intel-analytics/BigDL/pull/964


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/caffe/CaffePersister.scala
```diff
@@ -0,0 +1,268 @@
...
+import com.intel.analytics.bigdl.nn.Graph.ModuleNode
+import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, Activity}
+import com.intel.analytics.bigdl.tensor.Tensor
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.Path
+import org.apache.hadoop.io.IOUtils
+import org.apache.log4j.Logger
+/**
+ * An utility to convert BigDL model into caffe format and persist into local/hdfs file system
+ *
```
SeaOfOcean: A utility

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/caffe/CaffePersister.scala
```diff
@@ -0,0 +1,268 @@
...
+ * @param prototxtPath  path to store model definition file
+ * @param modelPath  path to store model weights and bias
+ * @param module BigDL module to be converted
+ * @param useV2  convert to V2 layer or not
```
SeaOfOcean: Do we need to convert to V1? I think this is deprecated in the future? maybe convert to V2 is enough?

spark/dl/src/test/scala/com/intel/analytics/bigdl/utils/CaffePersisterSpec.scala
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  "Persist V1 module" should "works properly" in {
+    val module = new ModuleNode(Linear[Double](100, 10).
+      setName("simple linear").asInstanceOf[AbstractModule[Activity, Tensor[Double], Double]])
+    val graph = Graph(module, module)
+    CaffePersister.persist("/tmp/v1.prototxt", "/tmp/v1.caffemodel",
+      graph, useV2 = false, overwrite = true)
+  }
+
```
SeaOfOcean: add test for v2 save/load, and you also need hdfs test (like this https://github.com/intel-analytics/BigDL/blob/0c8b8437d8d353dc56e1559afe63de7a09e02a5d/spark/dl/src/test/scala/com/intel/analytics/bigdl/integration/HdfsSpec.scala)

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DirectedGraph.scala
```diff
@@ -170,6 +170,17 @@ class Node[T](val element: T) extends Serializable {
...
+   *  @param node another node
+   *  @return current node
+   */
+  def delete(node: Node[T]): Node[T] = {
+    if (node.prevs.contains(this)) node.prevs.-=(this)
+    if (this.nexts.contains(node)) this.nexts.-=(node)
+    this
+  }
+
+  /**
```
SeaOfOcean: will this delete affect original model definition?

SeaOfOcean:
add corresponding python API

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DirectedGraph.scala
wzhongyuan: it won't. as per https://github.com/intel-analytics/BigDL/pull/964, we need to add a view before linear if inputsize is not equals to outputsize, so we need to remove it vice versa when persisting.

spark/dl/src/test/scala/com/intel/analytics/bigdl/utils/CaffePersisterSpec.scala
wzhongyuan: Sounds great, Added new test cases and tested locally as well

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/caffe/CaffePersister.scala
wzhongyuan: I think we can support both for now in case users have legacy versions model.

wzhongyuan:
@SeaOfOcean Refined as per your comments. For Python API, we may add in future but not in this PR

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Module.scala
```diff
@@ -44,7 +44,7 @@ object Module {
     CaffeLoader.load[T](model, defPath, modelPath, matchAll)
   }
 
-  def loadCaffeDynamic[T: ClassTag](defPath: String, modelPath: String, matchAll: Boolean = true)(
+  def loadCaffeModel[T: ClassTag](defPath: String, modelPath: String, matchAll: Boolean = true)(
     implicit ev: TensorNumeric[T]): AbstractModule[Activity, Activity, T] = {
```
SeaOfOcean: document for public method

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/caffe/CaffePersister.scala
```diff
@@ -0,0 +1,268 @@
...
+ * @param overwrite whether to overwirte existing caffe files
+ */
+class CaffePersister[T: ClassTag](val prototxtPath: String,
+      val modelPath: String, val module : Container[Activity, Activity, T],
+      useV2 : Boolean = true, overwrite : Boolean = false)(implicit ev: TensorNumeric[T]) {
+
+  private val logger = Logger.getLogger(getClass)
+
+  private val hdfsPrefix: String = "hdfs:"
+
```
SeaOfOcean: s3 support https://github.com/intel-analytics/BigDL/pull/1136

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/caffe/CaffePersister.scala
```diff
@@ -0,0 +1,268 @@
...
+  private val v1Converter = new V1LayerConverter[T]()
+  private val v2Converter = new LayerConverter[T]()
+
+  private var netparam: NetParameter.Builder = NetParameter.newBuilder
+
+  def saveAsCaffe() : Unit = {
```
SeaOfOcean: private

SeaOfOcean:
Do we need set initMethod for caffe layers (filter) ? layer-wise learning rate? And layer-wise regularizer

wzhongyuan:
@SeaOfOcean, thanks. I've made refinement based on your comments. For the initialization, it's more for how to initialize the model, but we are more for copy trained model, Personally I don't think it's a must but we can consider to add in future. I tend to push all the functionalities first.
For s3 support, I will merge @qiuxin2012 's change after it's put into master and add support for persister.

Thanks

wzhongyuan:
PR Validation passed 
http://10.239.47.211:18888/job/BigDL-PR-Validation/1257/

SeaOfOcean:
LGTM

wzhongyuan:
@SeaOfOcean  Thanks for the review, I will raise a separate one for s3 support shortly

PR #1103, opened 2017-06-27
add Exponential learning rate decay

SeaOfOcean:

add Exponential learning rate decay

## How was this patch tested?

unit test

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/SGD.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/SGD.scala
@@ -453,6 +453,29 @@ object SGD {
   }
 
   /**
+   * [[Exponential]] is a learning rate schedule, which rescale the learning rate by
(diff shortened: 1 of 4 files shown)


SeaOfOcean:
pr validate pass

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/SGD.scala
```diff
@@ -453,6 +453,29 @@ object SGD {
   }
 
   /**
+   * [[Exponential]] is a learning rate schedule, which rescale the learning rate by
```
zhangxiaoli73: Why [[...]]

qiuxin2012:
LGTM

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/SGD.scala
SeaOfOcean: It can help recognize Exponential as class in IDE

zhichao-li:
LGTM

PR #1117, opened 2017-06-29
add preTopology to boost performance of rnn

JerryYanWan:

Revise Cell.
Add pre-topology for RNN to boost the performance.
For example, i2h for input can be pre-calculated before recurrence.

## How was this patch tested?
Unit Test.
The fast mode cannot influence the correctness of the model.


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -66,11 +75,12 @@ class Recurrent[T : ClassTag]()
    */
   private def extend(times: Int, batchSize: Int, hiddenSize: Int): Unit = {
     if (hidden == null) {
-      require(modules != null && modules.length == 1,
-        "Recurrent extend: should contain only one cell")
+      require(topology != null && modules != null && modules.length <= 2,
```
qiuxin2012: add if preTopology == null

PR #1120, opened 2017-06-29
record forward time in Graph correctly

frankfzw:

Bug fix: use `model.forward` instead of `model.updateOutput` in `Graph` to record model forward time.
Add unit test to test the forward/backward time recording.

## How was this patch tested?

unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
@@ -72,7 +72,7 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
       } else {
         seqToTable(node.prevNodes.map(_.element.output))
       }
-      node.element.updateOutput(inputsBP(i))
+      node.element.forward(inputsBP(i))
       i += 1
     }
 
--- a/spark/dl/src/test/scala/com/intel/analytics/bigdl/nn/GraphSpec.scala
+++ b/spark/dl/src/test/scala/com/intel/analytics/bigdl/nn/GraphSpec.scala
@@ -335,6 +335,24 @@ class GraphSpec extends FlatSpec with Matchers {
     gradient should be(T(Tensor(T(-1.0f, 1.0f)), Tensor(T(-1.0f, 1.0f))))
   }
 
+  "Graph forward/backward time" should "be recorded" in {
+    val input1 = Input()
+    val input2 = Input()
+    val add = CAddTable().inputs(input1, input2)
+    val add2 = AddConstant(2.0f).inputs(add)
+    val relu = ReLU().inputs(add2)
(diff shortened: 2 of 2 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -72,7 +72,7 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
       } else {
         seqToTable(node.prevNodes.map(_.element.output))
       }
-      node.element.updateOutput(inputsBP(i))
+      node.element.forward(inputsBP(i))
```
yiheng: don't do this. Please override the forward method of the graph

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -62,6 +62,14 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
     extends Container[Activity, Activity, T]{
 
   override def updateOutput(input: Activity): Activity = {
+    val output = doForward(input) { (element, forwardInput) =>
```
yiheng: why there's an output here?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -62,6 +62,14 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
...
+      element.updateOutput(forwardInput)
+    }
+    output
+  }
+
+  private def doForward(input: Activity)
```
yiheng: private def forwardExecution(action: (AbstractModule[Activity, Tensor[T], T], Activity) => Unit)(input: Activity): Activity

so you can def

private def doForward = forwardExecution((layer, input) => {layer.forward(input))
private def doUpdateOutput = forwardExecution((layer, input) => {layer.updateOutput(input)})

...
doFoward(node.element, ...)
doUpdateOutput(node.element, ...)

yiheng:
Remove the final in the forward method in AbstractModule. Also what about all layers inherit from a Layer class and the forward/backward in the Layer class are final?

yiheng:
LGTM

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -62,6 +62,19 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
     extends Container[Activity, Activity, T]{
 
   override def updateOutput(input: Activity): Activity = {
+    forwardExecution(input) { (element, forwardInput) =>
```
yiheng: should we use scala macro to do this?

PR #1137, opened 2017-07-03
fix set bugs for fInput and fGradOutput in TimeDistributed layer

JerryYanWan:
fInput should not be set to the same storage as input
fGradOutput should not be set to the same storage as gradOutput

## How was this patch tested?
Unit Test


jason-dai:
Please add a unit test for the fix.

PR #1142, opened 2017-07-05
Add plateau learning rate schedule, add trigger for loss and score

SeaOfOcean:

Add plateau learning rate schedule, add trigger for loss and score
Add two float variable loss and score to driverState and optimState
If there are multiple validation method, only the first score will be recorded.

## How was this patch tested?

unit test

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
@@ -109,7 +109,9 @@ object DistriOptimizer {
     var wallClockTime = 0L
     var lastEpochTime = 0L
     val driverState = T("epoch" -> optimMethod.state.get[Int]("epoch").getOrElse(1),
-      "neval" -> optimMethod.state.get[Int]("neval").getOrElse(1))
+      "neval" -> optimMethod.state.get[Int]("neval").getOrElse(1),
+      "Loss" -> optimMethod.state.get[Float]("Loss").getOrElse(Float.PositiveInfinity),
+      "score" -> optimMethod.state.get[Float]("score").getOrElse(0f))
     val _subModelNumber = Engine.getEngineType match {
       case MklBlas => coresPerNode
       case _ => throw new IllegalArgumentException()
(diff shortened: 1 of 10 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Trigger.scala
```diff
@@ -99,5 +99,29 @@ object Trigger {
...
+   * A trigger that triggers an action when training loss less than "min" loss
+   * @param min min loss
+   */
+  def minLoss(min: Float): Trigger = {
+    new Trigger() {
+      override def apply(state: Table): Boolean = {
+        state[Float]("Loss") < min
+      }
+    }
+  }
```
shane-huang: May not be addressed in this pr, but we could give it a thought of below cases: 
- we are at some point, where the loss is below "min" and will actually continue to drop if we train longer
- we are at some point, where the loss is just a little bit above the "min" but never goes below "min". It trains forever.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -262,6 +264,10 @@ object DistriOptimizer {
           parameters.gradientPartition.div(ev.fromType(finishedModelNum))
           modelCache.optimMethod.state.update("epoch", driverState[Int]("epoch"))
           modelCache.optimMethod.state.update("neval", driverState[Int]("neval"))
+          modelCache.optimMethod.state.update("Loss", driverState[Float]("Loss"))
```
zhangxiaoli73: If you don't use Loss and score in your learningRate schedule, I think you don't have to pass then to modelCache.optimMethod

SeaOfOcean: To support Loss trigger and score Trigger, we need record them

SeaOfOcean:
pr validate pass

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
zhangxiaoli73: ok

zhichao-li:
LGTM

Issue #1119, opened 2017-06-29
Different criterion used in train and validation


yunhuiguo:
For a binary classification task, the label is 0 and 1.  And when use the same data for train and validation, and set the criterion in Optimizer() to BCECriterion(),  I have the following error message:

Caused by: java.lang.AssertionError: assertion failed: curTarget 0 is out of range 1 to 1
        at scala.Predef$.assert(Predef.scala:170)
        at com.intel.analytics.bigdl.nn.ClassNLLCriterion$$anonfun$updateOutput$4.apply(ClassNLLCriterion.scala:96)
        at com.intel.analytics.bigdl.nn.ClassNLLCriterion$$anonfun$updateOutput$4.apply(ClassNLLCriterion.scala:94)
        at com.intel.analytics.bigdl.utils.ThreadPool$$anonfun$invoke$2.apply(ThreadPool.scala:162)
        at scala.concurrent.impl.Future$PromiseCompletingRunnable.liftedTree1$1(Future.scala:24)
        at scala.concurrent.impl.Future$PromiseCompletingRunnable.run(Future.scala:24)

PR #1149, opened 2017-07-05
Change type of ValidationMethod from string to ValidationMethod instance

zhichao-li:

Change  ValidationMethod from string to ValidationMethod Object

## How was this patch tested?
unittest

fixed https://github.com/intel-analytics/BigDL/issues/1119

### DIFF
--- a/pyspark/bigdl/optim/optimizer.py
+++ b/pyspark/bigdl/optim/optimizer.py
@@ -647,12 +647,13 @@ class L2Regularizer(JavaValue):
 
 def _test():
     import doctest
+    from pyspark import SparkContext
     from bigdl.optim import optimizer
     from bigdl.util.common import init_engine
-    from bigdl.util.common import get_spark_context
-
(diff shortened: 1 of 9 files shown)


pyspark/bigdl/optim/optimizer.py
```diff
@@ -400,8 +444,10 @@ def set_validation(self, batch_size, val_rdd, trigger, val_method=["Top1Accuracy
         :param trigger: validation interval
         :param val_method: the ValidationMethod to use,e.g. "Top1Accuracy", "Top5Accuracy", "Loss"
         """
+        if val_method is None:
+            val_method = [Top1Accuracy()]
```
dding3: Looks like if we set val_method's default value in the parameter list, code is more clean. Is there a reason why we set none in the parameter list and update it in the implementation.

zhichao-li:
jenkins passed

PR #1150, opened 2017-07-05
optimize GRU by pre-calculating three linear operations

JerryYanWan:
define preTopology in GRU

## How was this patch tested?
Unit Test.

Since the structure of GRU in BigDL and Torch are different, I revised the Unit Test by rearrange the parameters in copy operation.

The speed:

I set batchSize = 12, inputSize = 1000, hiddenSize = 1000, and time = 200.
Running on my Desktop.
```
elapsed time | before    | after
------------------------------------------
1                | 3.22405 | 2.91554
2                | 3.26716 | 2.94787
3                | 3.17547 | 2.96847
4                | 3.15196 | 2.99957
5                | 3.16941 | 2.93977
------------------------------------------
average        | 3.19761 | 2.95424
------------------------------------------
throughput    | 3.7528  | 4.06195 
```
Thus, the revised version will give a ~8.24% throughput improvement.
With larger batchSize, inputSize and hiddenSize, the improvement will be more salient.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/GRU.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/GRU.scala
@@ -65,10 +65,8 @@ class GRU[T : ClassTag] (
   var i2g: AbstractModule[_, _, T] = _
   var h2g: AbstractModule[_, _, T] = _
   var gates: AbstractModule[_, _, T] = _
-  val timeDim = 2
+  val featDim = 2
   override var cell: AbstractModule[Activity, Activity, T] = buildGRU()
-
-  var linear: AbstractModule[Activity, Activity, T] = null
 
   override def preTopology: AbstractModule[Activity, Activity, T] =
     if (p != 0) {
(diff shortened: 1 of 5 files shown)


jason-dai:
Shall we do something similar for LSTM?

JerryYanWan:
The LSTM has already been optimized for input2hidden calculation.
I'm working on LSTMPeephole with similar strategy. Will propose PR soon.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/GRU.scala
```diff
@@ -57,30 +57,44 @@ class GRU[T : ClassTag] (
...
     regularizers = Array(wRegularizer, uRegularizer, bRegularizer)
   ) {
   var i2g: AbstractModule[_, _, T] = _
   var h2g: AbstractModule[_, _, T] = _
   var gates: AbstractModule[_, _, T] = _
+  val timeDim = 2
   override var cell: AbstractModule[Activity, Activity, T] = buildGRU()
 
-  def buildGates(): AbstractModule[Activity, Activity, T] = {
+  var linear: AbstractModule[Activity, Activity, T] = null
```
qiuxin2012: remove

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -271,6 +271,9 @@ class Recurrent[T : ClassTag]()
   override def updateGradInput(input: Tensor[T], gradOutput: Tensor[T]): Tensor[T] = {
 
     gradInput = if (preTopology != null) {
+      if (preTopology.gradInput == null) {
```
qiuxin2012: Add comment

PR #1158, opened 2017-07-05
Optimize LSTMPeephole by pre-calculating the input2hidden linear

JerryYanWan:
add preTopology in LSTMPeephole to boost the performance

## How was this patch tested?
Unit Test.

The structure of LSTMPeephole in BigDL and Torch are different. So I rearrange the weights in copy operation.

The speed:

I set batchSize = 12, inputSize = 300, hiddenSize = 300, and time = 200.
Running on my Desktop.
```
elapsed time | before    | after
------------------------------------------
1                | 1.25213 | 1.11945
2                | 1.25913 | 1.13069
3                | 1.27919 | 1.22337
4                | 1.27735 | 1.18582
5                | 1.23374 | 1.16419
------------------------------------------
average        | 1.26031 | 1.16470
------------------------------------------
throughput    | 9.5214   | 10.303 
```
Thus, the revised version will give a ~8.2% throughput improvement.
With larger batchSize, inputSize and hiddenSize, the improvement will be more salient.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/LSTMPeephole.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/LSTMPeephole.scala
@@ -65,7 +65,7 @@ class LSTMPeephole[T : ClassTag] (
   var outputGate: Sequential[T] = _
   var hiddenLayer: Sequential[T] = _
   var cellLayer: Sequential[T] = _
-  val timeDim = 2
+  val featDim = 2
   override var cell: AbstractModule[Activity, Activity, T] = buildLSTM()
 
   override def preTopology: AbstractModule[Activity, Activity, T] =
@@ -93,17 +93,17 @@ class LSTMPeephole[T : ClassTag] (
   }
 
   def buildInputGate(): Sequential[T] = {
-    inputGate = buildGate(timeDim, 1, hiddenSize)
+    inputGate = buildGate(featDim, 1, hiddenSize)
     inputGate
   }
 
   def buildForgetGate(): Sequential[T] = {
-    forgetGate = buildGate(timeDim, 1 + hiddenSize, hiddenSize)
+    forgetGate = buildGate(featDim, 1 + hiddenSize, hiddenSize)
     forgetGate
   }
 
   def buildOutputGate(): Sequential[T] = {
-    outputGate = buildGate(timeDim, 1 + 3 * hiddenSize, hiddenSize)
+    outputGate = buildGate(featDim, 1 + 3 * hiddenSize, hiddenSize)
     outputGate
   }
 
@@ -111,7 +111,7 @@ class LSTMPeephole[T : ClassTag] (
     val hidden = Sequential()
       .add(NarrowTable(1, 2))
 
-    val i2h = Narrow(timeDim, 1 + 2 * hiddenSize, hiddenSize)
+    val i2h = Narrow(featDim, 1 + 2 * hiddenSize, hiddenSize)
 
     val h2h = Sequential()
       .add(Dropout(p))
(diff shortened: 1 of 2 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/LSTMPeephole.scala
```diff
@@ -67,13 +67,16 @@ class LSTMPeephole[T : ClassTag] (
...
+      bRegularizer = bRegularizer)))
+
+  def buildGate(dimension: Int, offset: Int, length: Int): Sequential[T] = {
     val gate = Sequential()
 
-    val i2g = Sequential()
-      .add(Dropout(p))
-      .add(Linear(inputSize, hiddenSize, wRegularizer = wRegularizer,
-        bRegularizer = bRegularizer))
+    val i2g = Narrow(dimension, offset, length)
```
dding3: I think "Narrow" will do a deep copy which will affect the performance. We need either update narrow to get rid of deep copy or find another way here.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/LSTMPeephole.scala
```diff
@@ -89,28 +92,26 @@ class LSTMPeephole[T : ClassTag] (
   }
 
   def buildInputGate(): Sequential[T] = {
-    inputGate = buildGate()
+    inputGate = buildGate(2, 1, hiddenSize)
```
dding3: I think it would be better if we define a constant similar variable, set it as 2 and use it here

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/LSTMPeephole.scala
JerryYanWan: We use Narrow to select the corresponding input instead of calculating the Linear layer.
For example, the input size is [3, 12]. Then we need to select the sub-tensor from the second dimension: [3, 4], [3, 4], [3, 4].
Do you have any suggestions?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/LSTMPeephole.scala
JerryYanWan: It is hard coded. Set a class variable like val time = 2?

dding3: Name variable as timeDim?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/LSTMPeephole.scala
dding3: Add an option in Narrow layer which allow user avoid deep copy, like what AddConstant did?

JerryYanWan: The narrow layer will copy the subtensor out. It will be beneficial for the next layer such as CAddTable. Thus, keep it?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/LSTMPeephole.scala
JerryYanWan: revised

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/LSTMPeephole.scala
qiuxin2012: Narrow is like Contiguous

qiuxin2012:
LGTM

Issue #769, opened 2017-04-11
Support SpatialConvolution noBias


SeaOfOcean:
SpatialConvolution noBias
https://github.com/torch/nn/blob/master/SpatialConvolution.lua

PR #1162, opened 2017-07-05
support spatial convolution without bias

dding3:
Support spatial convolution without bias

## How was this patch tested?
new ut

fixed https://github.com/intel-analytics/BigDL/issues/769

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialConvolution.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialConvolution.scala
@@ -68,7 +68,7 @@ class SpatialConvolution[T: ClassTag](
     Tensor[T](nGroup, nOutputPlane / nGroup, nInputPlane / nGroup, kernelH, kernelW)
   }
 
-  val bias: Tensor[T] = if (!withBias) Tensor[T]()
+  val bias: Tensor[T] = if (!withBias) null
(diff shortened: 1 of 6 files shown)


dding3:
pr validation pass

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialConvolution.scala
```diff
@@ -160,11 +164,14 @@ class SpatialConvolution[T: ClassTag](
       }
       var g = 0
       while (g < nGroup) {
+        val biasUse = if (withBias) {
+          bias.narrow(1, g * nOutputPlane / nGroup + 1, nOutputPlane / nGroup)
+        } else bias
```
yiheng: what's the bias here? is it a null?

dding3: It depends, if withBias is false, it's an empty tensor.

yiheng: The logic just looks a little weird. what about

```scala
val biasUse = if (withBias) {
   ...
} else {
   null
}
```
So people will know it's a null instead of guessing what the bias is.

yiheng:
LGTM

PR #1169, opened 2017-07-06
feat: add new property: bigdl.utils.LoggerFilter.enableSparkLog

i8run:

Add a new property: `bigdl.utils.LoggerFilter.enableSparkLog`. It will not output spark logs to bigdl.log(default) when it's disabled.

## How was this patch tested?

1. Jenkins.
2. LeNet5 model manually.


spark/todo.list
```diff
@@ -0,0 +1,20 @@
+visualization/SummarySpec.scala
```
yiheng: what's this?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/LoggerFilter.scala
```diff
@@ -18,11 +18,18 @@ package com.intel.analytics.bigdl.utils
 
 import org.apache.log4j._
 import java.nio.file.{Paths, Files}
-import scala.collection.JavaConverters._
+
+ // scalastyle:off
```
yiheng: should we move this to the javadoc of LoggerFilter?

i8run: I tried it, but style check failed.

Issue #1024, opened 2017-06-16
MV forward twice will add the two output.


qiuxin2012:
```
scala> val module = MV()
module: com.intel.analytics.bigdl.nn.MV[Float] = MV()

scala> 

scala> module.forward(T(Tensor.range(1, 12, 1).resize(2, 2, 3), Tensor.range(1, 6, 1).resize(2, 3)))
res8: com.intel.analytics.bigdl.tensor.Tensor[Float] =
14.0	32.0
122.0	167.0
[com.intel.analytics.bigdl.tensor.DenseTensor of size 2x2]

scala> module.forward(T(Tensor.range(1, 12, 1).resize(2, 2, 3), Tensor.range(1, 6, 1).resize(2, 3)))
res9: com.intel.analytics.bigdl.tensor.Tensor[Float] =
28.0	64.0
244.0	334.0
[com.intel.analytics.bigdl.tensor.DenseTensor of size 2x2]
```

PR #1171, opened 2017-07-07
Fix a bug in the MV layer

psyyz10:
Fix a bug in the MV layer

## How was this patch tested?
Unit tests

fixed https://github.com/intel-analytics/BigDL/issues/1024


yiheng:
LGTM

PR #1172, opened 2017-07-07
Fix reference error and add script for packing pypi package

zhichao-li:

Fix reference error and add script for packing pypi package

## How was this patch tested?
manual test.


zhichao-li:
jenkins passed

PR #1176, opened 2017-07-07
Recurrent supports BatchNormalization

JerryYanWan:
add a method that can support BatchNormalization

Usage:

```
Recurrent[Double](BatchNormParams())
        .add(RnnCell[Double](inputSize, hiddenSize, Tanh[Double]()))
```

## How was this patch tested?

Unit Test
The programme can run added BatchNormalization correctly.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
@@ -356,9 +356,7 @@ class Recurrent[T : ClassTag]()
     copy(cells.map(x => x.gradInput.toTable[Tensor[T]](inputDim)),
         gradInputCell, 0)
     if (preTopology != null) {
-      preTopology.updateGradInput(input, gradInputCell)
-      gradInput = preTopology.gradInput.toTensor[T]
-      // gradInput = preTopology.updateGradInput(input, gradInputCell).toTensor[T]
+      gradInput = preTopology.updateGradInput(input, gradInputCell).toTensor[T]
     }
     gradInput
   }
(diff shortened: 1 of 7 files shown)


jason-dai:
@dding3 @qiuxin2012 Please take a look.

dding3:
I think we also need update python recurrent and related document ?

spark/dl/src/test/scala/com/intel/analytics/bigdl/nn/RecurrentSpec.scala
```diff
@@ -27,6 +27,106 @@ import scala.collection.mutable.ArrayBuffer
...
+
+    val input = Tensor[Double](batchSize, time, inputSize)
+    val gradOutput = Tensor[Double](batchSize, time, outputSize)
+
+    val output = model.forward(input)
+    val gradInput = model.backward(input, gradOutput)
+
+    println("add normalization")
+  }
+
```
dding3: Could you also add a case for convlstm, since it currently doesn't have pretopology and has a different code path.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -65,15 +67,58 @@ class Recurrent[T : ClassTag]()
...
+    preTopology = preTopology match {
+      case sequential: Sequential[T] =>
+        if (layer != null) {
+          sequential.add(layer)
+        } else if (sequential.modules.length == 0) {
+          null
+        } else {
+          sequential
+        }
+      case module: AbstractModule[Activity, Activity, T] =>
```
jason-dai: if (layer != null)
   Sequential[T]().add(module).add(layer)
else
   module

PR #1180, opened 2017-07-07
Add 1D Convolution (TemporalConvolution)

psyyz10:

Add 1D Convolution (TemporalConvolution)

## How was this patch tested?
Unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TemporalConvolution.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TemporalConvolution.scala
@@ -64,6 +64,10 @@ class TemporalConvolution[T: ClassTag](
   val bias: Tensor[T] = Tensor[T](outputFrameSize)
   val gradWeight: Tensor[T] = Tensor[T](outputFrameSize, inputFrameSize * kernelW)
   val gradBias: Tensor[T] = Tensor[T](outputFrameSize)
+  @transient val inputWindow = Tensor[T]()
+  @transient var outputWindow = Tensor[T]()
+  @transient val gradInputWindow = Tensor[T]()
+  @transient val gradOutputWindow = Tensor[T]()
 
   {
     val stdv = 1.0 / math.sqrt(kernelW * inputFrameSize)
@@ -77,7 +81,7 @@ class TemporalConvolution[T: ClassTag](
 
   override def reset(): Unit = {
     if (initWeight == null) {
-      weightInitMethod.init(weight, VariableFormat.GP_OUT_IN_KW_KH)
+      weightInitMethod.init(weight, VariableFormat.OUT_IN)
     }
     if (initBias == null) {
       biasInitMethod.init(bias, VariableFormat.ONE_D)
@@ -105,8 +109,6 @@ class TemporalConvolution[T: ClassTag](
 
     val nInputFrame = input.size(dimSeq)
     var nOutputFrame = (nInputFrame - kernelW) / strideW + 1
-    val inputWindow = Tensor[T]()
-    var outputWindow = Tensor[T]()
 
     // Shape check on input with inputFrameSize and kernelW
     require(input.size(dimFeat) == inputFrameSize, "Invalid input frame size. Got: " +
@@ -117,7 +119,7 @@ class TemporalConvolution[T: ClassTag](
     val weightT = weight.transpose(1, 2)
 
     if (input.dim() == 2) {
-      output.resize(Array(nOutputFrame, outputFrameSize))
+      output.resize(nOutputFrame, outputFrameSize)
       // Add bias first
       var j = 1
       while (j <= nOutputFrame) {
@@ -146,7 +148,7 @@ class TemporalConvolution[T: ClassTag](
       }
     } else {
       val batchSize = input.size(1)
-      output.resize(Array(batchSize, nOutputFrame, outputFrameSize))
+      output.resize(batchSize, nOutputFrame, outputFrameSize)
       if (results == null || results.length != batchSize) {
         results = new Array[Future[Unit]](batchSize)
       }
@@ -205,9 +207,6 @@ class TemporalConvolution[T: ClassTag](
     val dimFeat = if (input.dim() == 2) 2 else 3
     val nInputFrame = input.size(dimSeq)
     var nOutputFrame = (nInputFrame - kernelW) / strideW + 1
-
-    val gradInputWindow = Tensor[T]()
-    val gradOutputWindow = Tensor[T]()
 
     // Shape check on input with inputFrameSize and kernelW
     require(input.size(dimFeat) == inputFrameSize, "Invalid input frame size. Got: " +
@@ -285,13 +284,10 @@ class TemporalConvolution[T: ClassTag](
     val nInputFrame = input.size(dimSeq)
     var nOutputFrame = (nInputFrame - kernelW) / strideW + 1
 
-    var gradOutputWindow = Tensor[T]()
-    val inputWindow = Tensor[T]()
-
     if (input.nDimension() == 2) {
       var j = 0
       while (j < nOutputFrame) {
-        gradOutputWindow = gradOutput.select(1, j + 1)
+        gradOutputWindow.set(gradOutput.select(1, j + 1))
         gradBias.add(gradBias, ev.fromType[Double](scaleB), gradOutputWindow)
         j += 1
       }
@@ -326,7 +322,7 @@ class TemporalConvolution[T: ClassTag](
           var nOutputSampleFrame = nOutputFrame
           var j = 0
           while (j < nOutputFrame) {
-            gradOutputWindow = gradOutputSample.select(1, j + 1)
+            gradOutputWindow.set(gradOutputSample.select(1, j + 1))
             gradBias.add(gradBias, ev.fromType[Double](scaleB), gradOutputWindow)
             j += 1
           }
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TemporalConvolution.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TemporalConvolution.scala
@@ -25,6 +25,27 @@ import com.intel.analytics.bigdl.utils.Engine
 import scala.concurrent.Future
 import scala.reflect.ClassTag
 
+/**
+ * Applies a 1D convolution over an input sequence composed of nInputFrame frames..
+ *   The input tensor in `forward(input)` is expected to be a 2D tensor
+ *   (`nInputFrame` x `inputFrameSize`) or a 3D tensor
+ *   (`nBatchFrame` x `nInputFrame` x `inputFrameSize`).
+ *
+ * @param inputFrameSize The input frame size expected in sequences given into `forward()`.
+ * @param outputFrameSize The output frame size the convolution layer will produce.
+ * @param kernelW The kernel width of the convolution
+ * @param strideW The step of the convolution in the width dimension.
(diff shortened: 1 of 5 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TemporalConvolution.scala
```diff
@@ -0,0 +1,447 @@
...
+
+    if (input.dim() == 3) {
+      dimSeq = 2
+      dimFeat = 3
+    }
+
+    val nInputFrame = input.size(dimSeq)
+    var nOutputFrame = (nInputFrame - kernelW) / strideW + 1
+    val inputWindow = Tensor[T]()
+    var outputWindow = Tensor[T]()
```
JerryYanWan: It will new Tensor for each input, revise it to a class variable?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TemporalConvolution.scala
```diff
@@ -0,0 +1,447 @@
...
+
+    // Shape check on input with inputFrameSize and kernelW
+    require(input.size(dimFeat) == inputFrameSize, "Invalid input frame size. Got: " +
+      s"${input.size(dimFeat)}, Expected: $inputFrameSize")
+    require(nOutputFrame >= 1, "Input sequence smaller than kernel size. Got: " +
+      s"$nInputFrame, Expected: $kernelW")
+
+    val weightT = weight.transpose(1, 2)
```
JerryYanWan: Revise it to class variable?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TemporalConvolution.scala
```diff
@@ -0,0 +1,447 @@
...
+
+    if (input.dim() == 2) {
+      output.resize(Array(nOutputFrame, outputFrameSize))
```
JerryYanWan: output.resize(nOutputFrame, outputFrameSize)

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TemporalConvolution.scala
```diff
@@ -0,0 +1,447 @@
...
+          Array(nFrame, output.size(dimFeat)),
+          Array(outputFrameStride * output.size(dimFeat), 1))
+
+        outputWindow.addmm(ev.fromType[Int](1), outputWindow,
+          ev.fromType[Int](1), inputWindow, weightT)
+        j += 1
+      }
+    } else {
+      val batchSize = input.size(1)
+      output.resize(Array(batchSize, nOutputFrame, outputFrameSize))
```
JerryYanWan: output.resize(batchSize, nOutputFrame, outputFrameSize)

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TemporalConvolution.scala
```diff
@@ -0,0 +1,447 @@
...
+      if (results == null || results.length != batchSize) {
+        results = new Array[Future[Unit]](batchSize)
+      }
```
JerryYanWan: remove this part?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TemporalConvolution.scala
```diff
@@ -0,0 +1,447 @@
...
+      var i = 0
+      while (i < batchSize) {
+        results(i) = Engine.model.invoke(() => {
```
JerryYanWan: use while only since the Engine.model.threadpool = 1

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TemporalConvolution.scala
```diff
@@ -0,0 +1,447 @@
...
+    // Require input to be contiguous
+    require(input.isContiguous())
+
+    val dimSeq = if (input.dim() == 2) 1 else 2
+    val dimFeat = if (input.dim() == 2) 2 else 3
+    val nInputFrame = input.size(dimSeq)
+    var nOutputFrame = (nInputFrame - kernelW) / strideW + 1
+
+    val gradInputWindow = Tensor[T]()
+    val gradOutputWindow = Tensor[T]()
```
JerryYanWan: move it to a class variable?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TemporalConvolution.scala
```diff
@@ -0,0 +1,447 @@
...
+    // Require input to be contiguous
+    require(gradOutput.isContiguous())
+
+    val dimSeq = if (input.dim() == 2) 1 else 2
+    val dimFeat = if (input.dim() == 2) 2 else 3
+    val nInputFrame = input.size(dimSeq)
+    var nOutputFrame = (nInputFrame - kernelW) / strideW + 1
+
+    var gradOutputWindow = Tensor[T]()
+    val inputWindow = Tensor[T]()
```
JerryYanWan: move it to be a class variable?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TemporalConvolution.scala
```diff
@@ -0,0 +1,447 @@
...
+          gradOutputWindowT, inputWindow)
+        j += 1
+      }
+    } else {
+      val batchSize = input.size(1)
+      var gradOutputSample = Tensor[T]()
+      var inputSample = Tensor[T]()
+      var i = 0
+      while (i < batchSize) {
+        results(i) = Engine.model.invoke(() => {
```
JerryYanWan: use while only? do not use Engine.model.invoke,

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TemporalConvolution.scala
psyyz10: What the reason is?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TemporalConvolution.scala
psyyz10: But why SpatialConvolution's code  like that?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TemporalConvolution.scala
psyyz10: Any differences?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TemporalConvolution.scala
psyyz10: Yes, thanks

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TemporalConvolution.scala
psyyz10: Yes, thanks

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TemporalConvolution.scala
psyyz10: But why SpatialConvolution's code like that?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TemporalConvolution.scala
```diff
@@ -0,0 +1,443 @@
...
+  val propagateBack: Boolean = true,
+  var wRegularizer: Regularizer[T] = null,
+  var bRegularizer: Regularizer[T] = null,
+  val initWeight: Tensor[T] = null,
+  val initBias: Tensor[T] = null,
+  val initGradWeight: Tensor[T] = null,
+  val initGradBias: Tensor[T] = null
+)(implicit ev: TensorNumeric[T]) extends TensorModule[T]  with Initializable {
+
+  val weight: Tensor[T] = Tensor[T](outputFrameSize, inputFrameSize * kernelW)
```
yangw1234: The main reason of add `initWeight`, `initBias`, `initGradWeight`, `initGradBias` is to allow two or more layers to share the same weight tensor in `Graph`. So the `weight` variable should be
```scala
val weight = if (initWeight == null) Tensor[T](xxx) else iniWeight
```
Add the same to the other three.

spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDL.scala
```diff
@@ -863,6 +863,35 @@ class PythonBigDL[T: ClassTag](implicit ev: TensorNumeric[T]) extends Serializab
...
+    outputFrameSize: Int,
+    kernelW: Int,
+    strideW: Int = 1,
+    propagateBack: Boolean = true,
+    wRegularizer: Regularizer[T] = null,
+    bRegularizer: Regularizer[T] = null,
+    initWeight: Tensor[T] = null,
+    initBias: Tensor[T] = null,
+    initGradWeight: Tensor[T] = null,
+    initGradBias: Tensor[T] = null
```
yangw1234: This actually does not work. `JTensor.from_ndarray` returns a `JTensor` not a `Tensor`. So the parameter should be instance of `JTensor` and use `toTensor()` method to convert it to a `Tensor`

spark/dl/src/test/scala/com/intel/analytics/bigdl/torch/TemporalConvolutionSpec.scala
```diff
@@ -0,0 +1,95 @@
...
+    val weight = layer.weight
+    val bias = layer.bias
+
+    weight should be equals luaWeight
+    bias should be equals luaBias
+    output should be equals luaOutput
+    gradInput should be equals luaGradInput
+  }
+
+  "A SpatialConvolution" should "be good in gradient check for input" in {
```
yangw1234: Typo here? Should be "TemporalConvolution". Also the other two case.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TemporalConvolution.scala
JerryYanWan: Otherwise, each forward will transpose the weight. Is it enough to transpose weight once in the construction?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TemporalConvolution.scala
JerryYanWan: because, it is too complicated that no one has revised it....

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TemporalConvolution.scala
psyyz10: Transpose would create a new tensor, weight will update each iteration, so we need to get a new transpose each iteration, is it right?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TemporalConvolution.scala
psyyz10: I think it can be revised with other  convolutions together

jason-dai:
Do we have related doc?

Issue #1033, opened 2017-06-19
Stop Gradient in model


SeaOfOcean:
There is an example in keras

[stop_gradient](https://keras.io/backend/)

```
stop_gradient(variables)
```
Returns variables but with zero gradient w.r.t. every other variable

PR #1181, opened 2017-07-07
[Issue #1033 #450] Disable backpropagation for general layers

SeaOfOcean:

User can freeze some layers in the training process, or unfreeze afterwards.

```
val graph = Graph(Array(input1, input2), Array(output1, output2))
// set an array of layers to be trainable, the rest of layers are freezed
graph.setTrainable(Array("fc1_1", "conv1"))
// set an array of layers to be freezed, the rest of layers are trainable
graph.setFreeze(Array("fc2_1"))
// unfreeze and set all layers to be trainable
graph.unFreeze()
// or freeze certain layer
layer.setTrainable(false)
```

## How was this patch tested?

unit test

fixed https://github.com/intel-analytics/BigDL/issues/1033
https://github.com/intel-analytics/BigDL/issues/450

### DIFF
--- a/docs/docs/APIGuide/Module.md
+++ b/docs/docs/APIGuide/Module.md
@@ -151,13 +151,15 @@ val preductResult = model.predict(predictSet)
 To "freeze" a module means to exclude some layers of model from training.
 
 ```scala
-layer.setTrainable(false)
+layer.freeze()
+layer.unFreeze()
 model.freeze(Array("layer1", "layer2"))
 model.unFreeze()
 model.stopGradient(Array("layer1"))
 ```
-* A single layer can be "freezed" by calling ```setTrainable(false)```. If a layer is freezed,
+* A single layer can be "freezed" by calling ```freeze()```. If a layer is freezed,
 its parameters(weight/bias, if exists) are not changed in training process
+* A single layer can be "unFreezed" by calling ```unFreeze()```.
 * User can set freeze of list of layers in model by calling ```freeze```
 * User can unfreeze all layers by calling ```unFreeze```
 * stop the input gradient of layers that match the given names. Their input gradient are not computed.
@@ -165,7 +167,8 @@ And they will not contributed to the input gradient computation of layers that d
 
 **Python**
 ```python
-layer.set_trainable(false)
+layer.freeze()
+layer.unfreeze()
 model.freeze(["layer1", "layer2"])
 model.unfreeze()
 model.stop_gradient(["layer1"])
--- a/docs/docs/APIGuide/Module.md
+++ b/docs/docs/APIGuide/Module.md
@@ -152,13 +152,13 @@ To "freeze" a module means to exclude some layers of model from training.
 
 ```scala
 layer.setTrainable(false)
-model.setFreeze(Array("layer1", "layer2"))
+model.freeze(Array("layer1", "layer2"))
 model.unFreeze()
-model.setStopGradient(Array("layer1"))
+model.stopGradient(Array("layer1"))
 ```
 * A single layer can be "freezed" by calling ```setTrainable(false)```. If a layer is freezed,
 its parameters(weight/bias, if exists) are not changed in training process
-* User can set freeze of list of layers in model by calling ```setFreeze```
+* User can set freeze of list of layers in model by calling ```freeze```
 * User can unfreeze all layers by calling ```unFreeze```
 * stop the input gradient of layers that match the given names. Their input gradient are not computed.
 And they will not contributed to the input gradient computation of layers that depend on them.
@@ -166,9 +166,9 @@ And they will not contributed to the input gradient computation of layers that d
 **Python**
 ```python
 layer.set_trainable(false)
-model.set_freeze(["layer1", "layer2"])
+model.freeze(["layer1", "layer2"])
 model.unfreeze()
-model.set_stop_gradient(["layer1"])
+model.stop_gradient(["layer1"])
 ```
 
 **Scala**
@@ -216,7 +216,7 @@ println("fc2 weight \n", fc2.element.parameters()._1(0))
 fc1.element.getParameters()._1.apply1(_ => 1.0f)
 fc2.element.getParameters()._1.apply1(_ => 2.0f)
 model.zeroGradParameters()
-model.setFreeze(Array("fc2"))
+model.freeze(Array("fc2"))
 println("output2: \n", model.forward(input))
 model.backward(input, gradOutput)
 model.updateParameters(1)
@@ -269,7 +269,7 @@ println("fc2 weight \n", fc2.element.parameters()._1(0))
 ```scala
 fc1.element.getParameters()._1.apply1(_ => 1.0f)
 fc2.element.getParameters()._1.apply1(_ => 2.0f)
-model.setStopGradient(Array("cadd"))
+model.stopGradient(Array("cadd"))
 model.zeroGradParameters()
 println("output4: \n", model.forward(input))
 model.backward(input, gradOutput)
@@ -344,7 +344,7 @@ print "fc2 weight \n", fc2.element().parameters()['fc2']['weight']
 ```
 fc1.element().set_weights([np.array([[1,1,1,1],[1,1,1,1]]), np.array([1,1])])
 fc2.element().set_weights([np.array([[2,2,2,2],[2,2,2,2]]), np.array([2,2])])
-m3 = model.set_freeze(["fc2"])
+m3 = model.freeze(["fc2"])
 model.zero_grad_parameters()
 output = model.forward(input)
 print "output2 ", output
@@ -384,7 +384,7 @@ print "fc2 weight \n", fc2.element().parameters()['fc2']['weight']
 ```
 
 ```
-m3 = model.set_stop_gradient(["cadd"])
+m3 = model.stop_gradient(["cadd"])
 model.zero_grad_parameters()
 output = model.forward(input)
 print "output4 ", output
--- a/docs/docs/APIGuide/Module.md
+++ b/docs/docs/APIGuide/Module.md
@@ -147,3 +147,209 @@ val preductResult = model.predict(predictSet)
  preductResult = model.predict(predictSet)
 ```
 
+## Module Freeze
+To "freeze" a module means to exclude some part of model from training.
+Note that it only takes effect in BigDL graph model.
+
+```scala
+layer.setTrainable(false)
+model.setFreeze(Array("layer1", "layer2"))
+model.setTrainable(Array("layer3", "layer4"))
+model.unFreeze()
+```
+* A single layer can be "freezed" by calling ```setTrainable```
+* User can set freeze of list of layers in model by calling ```setFreeze```
+* User can set a list of trainable layers and keep the remaining layers freezed by calling ```setTrainable```
+* User can unfreeze all layers by calling ```unFreeze```
+**Python**
+```python
+layer.set_trainable(false)
+model.set_freeze(["layer1", "layer2"])
+model.set_trainable(["layer1", "layer2"])
+model.unfreeze()
+```
+
+**Scala**
+```scala
+val reshape = Reshape(Array(4)).inputs()
+val fc1_1 = Linear(4, 2).setName("fc1_1").inputs()
+val fc2_1 = Linear(4, 2).setName("fc2_1").setTrainable(false).inputs(reshape)
+val cadd_1 = CAddTable().inputs(fc1_1, fc2_1)
+val output1_1 = ReLU().inputs(cadd_1)
+val output2_1 = Threshold(10.0).inputs(cadd_1)
+
+val model = Graph(Array(reshape, fc1_1), Array(output1_1, output2_1))
+
+fc1_1.element.getParameters()._1.apply1(_ => 1.0f)
+fc2_1.element.getParameters()._1.apply1(_ => 2.0f)
+
+val input = T(Tensor(T(0.1f, 0.2f, -0.3f, -0.4f)),
+  Tensor(T(0.5f, 0.4f, -0.2f, -0.1f)))
+val gradOutput = T(Tensor(T(1.0f, 2.0f)), Tensor(T(3.0f, 4.0f)))
+
+println("output1: \n", model.forward(input))
+println("gradInput1: \n", model.backward(input, gradOutput))
+
+model.unFreeze()
+println("output2: \n", model.forward(input))
+println("gradInput2: \n", model.backward(input, gradOutput))
+
+model.setFreeze(Array("fc2_1"))
+println("output3: \n", model.forward(input))
+println("gradInput3: \n", model.backward(input, gradOutput))
+
+model.unFreeze()
+model.setTrainable(Array("fc1_1"))
+println("output4: \n", model.forward(input))
+println("gradInput4: \n", model.backward(input, gradOutput))
+```
+
+```
+(output1: 
+, {
+	2: 0.0
+	   0.0
+	   [com.intel.analytics.bigdl.tensor.DenseTensor of size 2]
+	1: 2.8
+	   2.8
+	   [com.intel.analytics.bigdl.tensor.DenseTensor of size 2]
+ })
+(gradInput1: 
+, {
+	2: 3.0
+	   3.0
+	   3.0
+	   3.0
+	   [com.intel.analytics.bigdl.tensor.DenseTensor of size 4]
+	1: [com.intel.analytics.bigdl.tensor.DenseTensor with no dimension]
+ })
+(output2: 
+, {
+	2: 0.0
+	   0.0
+	   [com.intel.analytics.bigdl.tensor.DenseTensor of size 2]
+	1: 3.6
+	   3.6
+	   [com.intel.analytics.bigdl.tensor.DenseTensor of size 2]
+ })
+(gradInput2: 
+, {
+	2: 3.0
+	   3.0
+	   3.0
+	   3.0
+	   [com.intel.analytics.bigdl.tensor.DenseTensor of size 4]
+	1: 6.0
+	   6.0
+	   6.0
+	   6.0
+	   [com.intel.analytics.bigdl.tensor.DenseTensor of size 4]
+ })
+(output3: 
+, {
+	2: 0.0
+	   0.0
+	   [com.intel.analytics.bigdl.tensor.DenseTensor of size 2]
+	1: 3.6
+	   3.6
+	   [com.intel.analytics.bigdl.tensor.DenseTensor of size 2]
+ })
+(gradInput3: 
+, {
+	2: 3.0
+	   3.0
+	   3.0
+	   3.0
+	   [com.intel.analytics.bigdl.tensor.DenseTensor of size 4]
+	1: [com.intel.analytics.bigdl.tensor.DenseTensor with no dimension]
+ })
+(output4: 
+, {
+	2: 0.0
+	   0.0
+	   [com.intel.analytics.bigdl.tensor.DenseTensor of size 2]
+	1: 3.6
+	   3.6
+	   [com.intel.analytics.bigdl.tensor.DenseTensor of size 2]
+ })
+(gradInput4: 
+, {
+	2: 3.0
+	   3.0
+	   3.0
+	   3.0
+	   [com.intel.analytics.bigdl.tensor.DenseTensor of size 4]
+	1: [com.intel.analytics.bigdl.tensor.DenseTensor with no dimension]
+ })
+```
+
+
+
+**Python**
+```python
+from bigdl.nn.layer import *
+import numpy as np
+
+reshape = Reshape([4])()
+fc1_1 = Linear(4, 2).set_name("fc1_1")()
+# set fc2_1 to non-trainable
+fc2_1 = Linear(4, 2).set_name("fc2_1").set_trainable(False)(reshape)
+cadd_1 = CAddTable()([fc1_1, fc2_1])
+output1_1 = ReLU()(cadd_1)
+output2_1 = Threshold(10.0)(cadd_1)
+model = Model([reshape, fc1_1], [output1_1, output2_1])
+
+input = [
+    np.array([0.1, 0.2, -0.3, -0.4]),
+    np.array([0.5, 0.4, -0.2, -0.1])]
+gradOutput = [
+    np.array([1.0, 2.0]), np.array([3.0, 4.0])]
+
+output = model.forward(input)
+print "output1: \n", output
+gradInput = model.backward(input, gradOutput)
+print "gradInput1 \n", gradInput
+
+# unfreeze all layers in model
+model.unfreeze()
+output = model.forward(input)
+print "output2 \n", output
+gradInput = model.backward(input, gradOutput)
+print "gradInput2 \n", gradInput
+
+# set fc2_1 to non-trainable in model API
+m3 = model.set_freeze(["fc2_1"])
+output = model.forward(input)
+print "output3 \n", output
+gradInput = model.backward(input, gradOutput)
+print "gradInput3 \n", gradInput
+
+model.unfreeze()
+# set only fc1_1 to trainable in model API
+m3 = model.set_trainable(["fc1_1"])
+output = model.forward(input)
+print "output4 \n", output
+gradInput = model.backward(input, gradOutput)
+print "gradInput4 \n", gradInput
+```
+
+```
+output1: 
+[array([ 0.        ,  0.58186555], dtype=float32), array([ 0.,  0.], dtype=float32)]
+gradInput1 
+[array([], dtype=float32), array([ 0.510382  ,  0.85159016, -0.55450386, -0.54459006], dtype=float32)]
+output2 
+[array([ 0.        ,  0.58186555], dtype=float32), array([ 0.,  0.], dtype=float32)]
+gradInput2 
+[array([ 0.04848308, -0.04261303,  0.04275261, -0.94517899], dtype=float32), array([ 0.510382  ,  0.85159016, -0.55450386, -0.54459006], dtype=float32)]
+output3 
+[array([ 0.        ,  0.58186555], dtype=float32), array([ 0.,  0.], dtype=float32)]
+gradInput3 
+[array([], dtype=float32), array([ 0.510382  ,  0.85159016, -0.55450386, -0.54459006], dtype=float32)]
+output4 
+[array([ 0.        ,  0.58186555], dtype=float32), array([ 0.,  0.], dtype=float32)]
+gradInput4 
+[array([], dtype=float32), array([ 0.510382  ,  0.85159016, -0.55450386, -0.54459006], dtype=float32)]
+```
+
+
--- a/docs/docs/APIGuide/Module.md
+++ b/docs/docs/APIGuide/Module.md
@@ -148,64 +148,55 @@ val preductResult = model.predict(predictSet)
 ```
 
 ## Module Freeze
-To "freeze" a module means to exclude some part of model from training.
-Note that it only takes effect in BigDL graph model.
+To "freeze" a module means to exclude some layers of model from training.
 
 ```scala
 layer.setTrainable(false)
 model.setFreeze(Array("layer1", "layer2"))
-model.setTrainable(Array("layer3", "layer4"))
 model.unFreeze()
+model.setStopGradient(Array("layer1"))
 ```
-* A single layer can be "freezed" by calling ```setTrainable```
+* A single layer can be "freezed" by calling ```setTrainable(false)```. If a layer is freezed,
+its parameters(weight/bias, if exists) are not changed in training process
 * User can set freeze of list of layers in model by calling ```setFreeze```
-* User can set a list of trainable layers and keep the remaining layers freezed by calling ```setTrainable```
 * User can unfreeze all layers by calling ```unFreeze```
+* stop the input gradient of layers that match the given names. Their input gradient are not computed.
+And they will not contributed to the input gradient computation of layers that depend on them.
+
 **Python**
 ```python
 layer.set_trainable(false)
 model.set_freeze(["layer1", "layer2"])
-model.set_trainable(["layer1", "layer2"])
 model.unfreeze()
+model.set_stop_gradient(["layer1"])
 ```
 
 **Scala**
+Original model without "freeze" or "stop gradient"
 ```scala
 val reshape = Reshape(Array(4)).inputs()
-val fc1_1 = Linear(4, 2).setName("fc1_1").inputs()
-val fc2_1 = Linear(4, 2).setName("fc2_1").setTrainable(false).inputs(reshape)
-val cadd_1 = CAddTable().inputs(fc1_1, fc2_1)
+val fc1 = Linear(4, 2).setName("fc1").inputs()
+val fc2 = Linear(4, 2).setName("fc2").inputs(reshape)
+val cadd_1 = CAddTable().setName("cadd").inputs(fc1, fc2)
 val output1_1 = ReLU().inputs(cadd_1)
 val output2_1 = Threshold(10.0).inputs(cadd_1)
 
-val model = Graph(Array(reshape, fc1_1), Array(output1_1, output2_1))
-
-fc1_1.element.getParameters()._1.apply1(_ => 1.0f)
-fc2_1.element.getParameters()._1.apply1(_ => 2.0f)
+val model = Graph(Array(reshape, fc1), Array(output1_1, output2_1))
 
 val input = T(Tensor(T(0.1f, 0.2f, -0.3f, -0.4f)),
   Tensor(T(0.5f, 0.4f, -0.2f, -0.1f)))
 val gradOutput = T(Tensor(T(1.0f, 2.0f)), Tensor(T(3.0f, 4.0f)))
 
+fc1.element.getParameters()._1.apply1(_ => 1.0f)
+fc2.element.getParameters()._1.apply1(_ => 2.0f)
+model.zeroGradParameters()
 println("output1: \n", model.forward(input))
-println("gradInput1: \n", model.backward(input, gradOutput))
-
-model.unFreeze()
-println("output2: \n", model.forward(input))
-println("gradInput2: \n", model.backward(input, gradOutput))
-
-model.setFreeze(Array("fc2_1"))
-println("output3: \n", model.forward(input))
-println("gradInput3: \n", model.backward(input, gradOutput))
-
-model.unFreeze()
-model.setTrainable(Array("fc1_1"))
-println("output4: \n", model.forward(input))
-println("gradInput4: \n", model.backward(input, gradOutput))
+model.backward(input, gradOutput)
+model.updateParameters(1)
+println("fc2 weight \n", fc2.element.parameters()._1(0))
 ```
-
 ```
-(output1: 
+(output1:
 , {
 	2: 0.0
 	   0.0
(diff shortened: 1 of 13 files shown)


yiheng:
I think this API is not so straight forward. Should we let user set which layer with parameter should be skipped instead of defining how the backward explicitly?

SeaOfOcean:
Now users can skip layer by calling setPropagateBack(false), the above graphs are the examples to show how gradients are processed.

yiheng:
say if the user loads a very complex model, only want to train/skip parameters of some layers, is it better just let the user tell you which layer to skip/train? If I were the user, I don't want to study the complex topology of the model if not nesscary.

SeaOfOcean:
For complex models, maybe we can allow users pass a list of TrainableScope or FreezeScope

For example
1. user can give a list of trainable layers' names
```
model.setTrainableScope(Array("conv3", "conv4"...))
``` 
the the layers with these names will be trained, and the left layers are unchanged.

2. user can give a list of freeze layers' names
```
model.setFreezeScope(Array("conv1", "conv2"...))
``` 
the layers with these names will be freezed, and the left layers are trained.

yiheng:
setTrainableScope -> setTrainable
setFreezeScope -> setFreeze

The random name in BigDL is not easy to use

SeaOfOcean:
pr validate pass, ready for review

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -65,14 +67,14 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
     var i = 0
     while(i < executions.length) {
       val node = executions(i)
-      inputsBP(i) = if (node.prevNodes.isEmpty && !node.element.isInstanceOf[WithoutInput]) {
+      val nodeInput = if (node.prevNodes.isEmpty && !node.element.isInstanceOf[WithoutInput]) {
```
yiheng: why remove the inputs cache? you can use a hashmap to replace the array.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -85,11 +87,15 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
...
-    while(i >= 0) {
-      val curNode = executions(i)
+    dummyOutputGrad.element.gradInput = gradOutput
+
+    var i = 0
+    while (i < executionsGrad.length) {
+      val curNode = executionsGrad(i)
       var curGradOutput : Tensor[T] = null
+
+      val backInput = getInput(input, curNode, backInputs)
```
yiheng: don't calculate it every time. use hashmap as cache

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -120,11 +126,15 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
...
```
yiheng: don't calculate it every time. use hashmap as cache

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -155,11 +165,12 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
...
   override def accGradParameters(input: Activity, gradOutput: Activity): Unit = {
-    var i = executions.length - 1
-    while(i >= 0) {
-      val curNode = executions(i)
-      curNode.element.accGradParameters(inputsBP(i), gradOutputBP(i))
-      i -= 1
+    var i = 0
+    while (i < executionsGrad.length) {
+      val curNode = executionsGrad(i)
+      val backInput = getInput(input, curNode, backInputs)
```
yiheng: don't calculate it every time. use hashmap as cache

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -175,26 +186,110 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
       "A graph container should not be changed after it is constructed")
   }
 
+  // input node for backward graph
+  val backInputs = new Array[ModuleNode[T]](inputs.length)
```
yiheng: why public?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -175,26 +186,110 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
...
+
   // Add a dummy output node, to get an one end graph. So the nodes that are not dependent by
   // the outputs will be excluded
   private val dummyOutput = new ModuleNode[T](new Dummy[T]())
+  // dummy output of backward graph
```
yiheng: don't understand the what the comment means. Please give more explaination

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -175,26 +186,110 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
...
+  private var dummyOutputGrad: ModuleNode[T] = null
   outputs.foreach(_ -> dummyOutput)
 
   /**
    * Computing backgraph
    */
   val backGraph = dummyOutput.graph(reverse = true)
+  var gradGraph: DirectedGraph[AbstractModule[Activity, Tensor[T], T]] = null
```
yiheng: why create the gradGraph? why not use the backGraph?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -175,26 +186,110 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
...
 
   /**
    * Execution plan
    */
   val executions = backGraph.topologySort.filter(!_.element.isInstanceOf[Dummy[T]]).reverse
+  var executionsGrad: Array[Node[AbstractModule[Activity, Tensor[T], T]]] = null
```
yiheng: backwardExecutions?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -175,26 +186,110 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
...
+
   modules.appendAll(executions.map(_.element.asInstanceOf[AbstractModule[Activity, Activity, T]]))
 
+
+
+  private def setTrainable(): Unit = {
```
yiheng: what's it for?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -175,26 +186,110 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
...
+      if (!node.element.trainable) {
+        node.nextNodes.foreach(next => {
+          if (!needBackward(next)) next.element.trainable = false
+        })
+      }
+      i += 1
+    }
+  }
+
+  private def needBackward(node: ModuleNode[T]): Boolean = {
```
yiheng: what's it for?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -257,6 +352,80 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
     }
   }
 
+  private def getInput(input: Activity, node: Node[AbstractModule[Activity, Tensor[T], T]],
```
yiheng: remove it

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -257,6 +352,80 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
...
+    if (node.prevNodes.isEmpty && !node.element.isInstanceOf[WithoutInput]) {
+      backInputData(node, input, inputs)
+    } else if (node.prevNodes.length == 1) {
+      node.prevNodes.head.element.output.toTensor[T]
+    } else {
+      seqToTable(node.prevNodes.map(_.element.output))
+    }
+  }
+
+  private def backInputData(
```
yiheng: remove it

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -175,26 +186,110 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
...
+    if (!node.element.trainable) return false
+    if (node.element.parameters() != null) return true
+    node.prevNodes.foreach(n => {
+      if (needBackward(n)) return true
+    })
+    false
+  }
+
+  def build(): this.type = {
+    setTrainable()
```
yiheng: what's it for?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -257,6 +352,80 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
...
+      input.toTable[Tensor[T]](i + 1)
+    }
+  }
+
+  /**
+   * set an array of layers to be trainable, the rest of layers are freezed
+   * @param names an array of layer names
+   */
+  def setTrainable(names: Array[String]): this.type = {
+    val namedModules = Utils.getNamedModules(this)
```
yiheng: module should provide a method to get submodules by name. You should use that.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -257,6 +352,80 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
...
+    this
+  }
+
+  /**
+   * set an array of layers to be freezed, the rest of layers are trainable
+   * @param names an array of layer names
+   * @return
+   */
+  def setFreeze(names: Array[String]): this.type = {
+    val namedModules = Utils.getNamedModules(this)
```
yiheng: module should provide a method to get submodules by name. You should use that.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -200,6 +200,20 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
   }
 
   /**
+   * whether this layer is trainable
+   */
+  var trainable: Boolean = true
```
yiheng: only layers with parameters are trainable, right? the default value should be false

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -200,6 +200,20 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
...
+
+  /**
+   * set trainable
+   * @param trainable whether this layer is trainable
+   */
+  def setTrainable(trainable: Boolean): this.type = {
```
yiheng: user should not set non-parameter layer trainable to true, right?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -175,26 +186,110 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
...
+  }
+
+  def build(): this.type = {
+    setTrainable()
+    dummyOutputGrad = new ModuleNode[T](new Dummy[T]())
+    val vertexMap = new util.HashMap[ModuleNode[T], ModuleNode[T]]()
+    val q = new util.LinkedList[ModuleNode[T]]()
+    q.add(dummyOutput)
+    vertexMap.put(dummyOutput, new ModuleNode(dummyOutput.element))
+    while (!q.isEmpty) {
```
yiheng: I'm confused by these logic. Shouldn't we point to all trainable nodes from the dummyOutput?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
SeaOfOcean: The default value of trainable in layer should be true, since they need perform updateGradInput to support the whole back propagation link

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
SeaOfOcean: backGraph is used for forward, gradGraph is used for backward. Since the backward graph may be different from forward graph, we need a new graph to get backward executions

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
SeaOfOcean: mark the non-trainable nodes from input node to output node
 For example, Reshape() -> Identity() -> Conv() -> SoftMax()
Here the Reshape and Identity can be non-trainable, since they do not need to propagate the
gradients to any layers with parameters

jason-dai:
Several comments:
1) Need to add python support
2) Need to add md docs
3) Can we do a comparison with how this is done in, say, Keras?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
SeaOfOcean: dummyOutput is for forward graph, we need a new dummy node for backward graph

SeaOfOcean:
Keras set single layer 

```
x = Input(shape=(32,))
layer = Dense(32)
layer.trainable = False
y = layer(x)

frozen_model = Model(x, y)
# in the model below, the weights of `layer` will not be updated during training
frozen_model.compile(optimizer='rmsprop', loss='mse')

layer.trainable = True
trainable_model = Model(x, y)
# with this model the weights of the layer will be updated during training
# (which will also affect the above model since it uses the same layer instance)
trainable_model.compile(optimizer='rmsprop', loss='mse')

frozen_model.fit(data, labels)  # this does NOT update the weights of `layer`
trainable_model.fit(data, labels)  # this updates the weights of `layer`
```

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
yiheng: the name is confusing. can you rename the backGraph to forwardGraph and the gradGraph to backwardGraph?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -175,26 +182,74 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
...
+  // input node for backward graph
+  private val backInputs = new Array[ModuleNode[T]](inputs.length)
+
   // Add a dummy output node, to get an one end graph. So the nodes that are not dependent by
   // the outputs will be excluded
   private val dummyOutput = new ModuleNode[T](new Dummy[T]())
+  // Add a dummy output node for backward graph,
+  // dummyOutputGrad has the same function as dummyOutput
+  // used to construct a backward graph
+  private var dummyOutputGrad: ModuleNode[T] = null
```
yiheng: private var dummyOutputGrad: ModuleNode[T] = _

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
yiheng: var gradGraph: DirectedGraph[AbstractModule[Activity, Tensor[T], T]] = _

should it be private?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -175,26 +182,74 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
...
    * Computing backgraph
    */
   val backGraph = dummyOutput.graph(reverse = true)
+  var gradGraph: DirectedGraph[AbstractModule[Activity, Tensor[T], T]] = null
 
   /**
    * Execution plan
    */
   val executions = backGraph.topologySort.filter(!_.element.isInstanceOf[Dummy[T]]).reverse
+  var backwardExecutions: Array[Node[AbstractModule[Activity, Tensor[T], T]]] = null
```
yiheng: private

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -175,26 +182,74 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
...
   /**
    * Computing backgraph
    */
   val backGraph = dummyOutput.graph(reverse = true)
+  var gradGraph: DirectedGraph[AbstractModule[Activity, Tensor[T], T]] = null
 
   /**
    * Execution plan
    */
   val executions = backGraph.topologySort.filter(!_.element.isInstanceOf[Dummy[T]]).reverse
```
yiheng: executions -> forwardExecutions

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -175,26 +182,74 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
...
+  var backwardExecutions: Array[Node[AbstractModule[Activity, Tensor[T], T]]] = null
+
   modules.appendAll(executions.map(_.element.asInstanceOf[AbstractModule[Activity, Activity, T]]))
 
+
+  def build(): this.type = {
```
yiheng: private? Or add java doc. It should be private

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -204,6 +205,27 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
   }
 
   /**
+   * whether this layer is trainable
+   */
+  private var trainable: Boolean = true
```
yiheng: default should be false

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -580,5 +602,23 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
...
+  def getSubModules(): Map[String, Module[T]] = {
+    if (nameToSubModules == null) nameToSubModules = Utils.getNamedModules(this)
+    nameToSubModules
+  }
+
+  /**
+   * get submodule by module name
+   */
+  def getSubModule(name: String): Module[T] = {
+    getSubModules()
```
yiheng: DO NOT write such code. The side effect here is totally unnecessary!!!

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -580,5 +602,23 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
               ): Array[(ValidationResult, ValidationMethod[T])] = {
     Validator(this, dataSet).test(vMethods)
   }
+
+  private var nameToSubModules: Map[String, Module[T]] = _
+  /**
+   * get all submodules map, use the module name as key
+   */
+  def getSubModules(): Map[String, Module[T]] = {
+    if (nameToSubModules == null) nameToSubModules = Utils.getNamedModules(this)
```
yiheng: this makes no sense at all. Why a layer should have a method named getSubModules? A layer don't have sub modules.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -580,5 +602,23 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
...
+    nameToSubModules
+  }
+
+  /**
+   * get submodule by module name
+   */
+  def getSubModule(name: String): Module[T] = {
+    getSubModules()
+    require(nameToSubModules.contains(name), s"$name does not exists")
+    nameToSubModules(name)
```
yiheng: Why don't you use the apply method but write a new one?

model("layer1") will give you the corrrectLayer

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
yiheng: BTW, why the return type is Module[T]. Shouldn't it be Option[Module[T]]?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -257,6 +312,46 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
     }
   }
 
+
+  /**
+   * set an array of layers to be trainable, the rest of layers are freezed
```
yiheng: why are the rest layers frozen? Why you change the state of other layers? The method should only care what passed in. It should NEVER touch the others. DO NOT design API like this.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -257,6 +312,46 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
...
+   * @param names an array of layer names
+   */
+  def setTrainable(names: Array[String]): this.type = {
```
yiheng: override this in Container.

yiheng: You're changing the Module class to a giant class.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -257,6 +312,46 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
...
+    val namedModules = getSubModules()
+    names.foreach(name => require(namedModules.contains(name),
+      s"there is no $name in the module, please check the layer name"))
+    namedModules.foreach(namedModule => {
+      if (!names.contains(namedModule._1) && namedModule._2.parameters() != null) {
+        namedModule._2.setTrainable(false)
```
yiheng: Only set layer to trainable. Donot touch other layers

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -257,6 +312,46 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
...
+      }
+    })
+    build()
+    this
+  }
+
+  /**
+   * set an array of layers to be freezed, the rest of layers are trainable
```
yiheng: You only set layers to freeze. The comments of the API is incorrect

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
yiheng: Still don't get the logic here. Shouldn't it be
1. list all trainable modules in the graph
2. point all trainable modules to the dummyOutputGrad
3. build a graph from dummyOutputGrad
4. top sort

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DirectedGraph.scala
```diff
@@ -193,6 +193,20 @@ class Node[T](val element: T) extends Serializable {
 
   private val nexts = new ArrayBuffer[Node[T]]()
   private val prevs = new ArrayBuffer[Node[T]]()
+
+  private var isInput: Boolean = false
```
yiheng: why do we need this?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
yiheng: point to all trainable nodes from the dummyOutputGrad? and generate the graph. What's the so much code for?

SeaOfOcean:
pr validate pass, please have a further review @yiheng thanks

pyspark/bigdl/nn/layer.py
```diff
@@ -376,6 +376,27 @@ def setBRegularizer(self, bRegularizer):
...
+    def set_trainable(self, trainable):
+        '''
+        set trainable
+        :param trainable: whether this layer is trainable
+        '''
+        callBigDlFunc(self.bigdl_type,
+                        "setLayerTrainable", self.value, trainable)
+        return self
+
+    def set_stop_gradient(self, stop_gradient):
```
yiheng: set_stop_gradient -> stop_gradient?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Container.scala
```diff
@@ -202,4 +203,22 @@ abstract class Container[A <: Activity : ClassTag,
...
+   */
+  def getSubModules(): Map[String, Module[T]] = {
+    val nameToSubModules = Utils.getNamedModules(this)
+    nameToSubModules
+  }
+
+  /**
+   * get submodule by module name
+   */
+  def getSubModule(name: String): Option[Module[T]] = {
```
yiheng: why not use the apply method in the AbstractModule?
model("layer_name")

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Container.scala
```diff
@@ -202,4 +203,22 @@ abstract class Container[A <: Activity : ClassTag,
       }
     }
   }
+
+  /**
+   * get all submodules map, use the module name as key
+   */
+  def getSubModules(): Map[String, Module[T]] = {
```
yiheng: why add this method?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Reshape.scala
```diff
@@ -48,6 +48,8 @@ class Reshape[@specialized(Float, Double) T: ClassTag](
     nElement *= size(i - 1)
   }
 
+  private var inplace: Boolean = true
```
yiheng: move it to constructor

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Reshape.scala
```diff
@@ -125,6 +131,15 @@ class Reshape[@specialized(Float, Double) T: ClassTag](
   override def toString(): String = {
     s"${getPrintName}(${size.mkString("x")})"
   }
+
+  override def clearState(): this.type = {
+    if (!inplace) {
```
yiheng: why need this? set twice should be ok, right?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -204,6 +205,41 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
   }
 
   /**
+   * set trainable. If trainable is false, its parameters(weight/bias, if exists)
+   * are not changed in training process
+   * @param trainable whether this layer is trainable
+   */
+  def setTrainable(trainable: Boolean): this.type = {
```
yiheng: this is weird API.  should it be split into two APIs?
unFreeze
freeze

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -204,6 +205,41 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
...
+    if (!trainable) {
+      scaleW = 0
+      scaleB = 0
+    }
+    this
+  }
+
+  private var _stopGradient: Boolean = false
```
yiheng: Do name any vairable in Java/Scala like this _stopGradient

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Reshape.scala
yiheng: can you add a comment to explain this?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DirectedGraph.scala
```diff
@@ -123,6 +122,29 @@ class DirectedGraph[T](val source : Node[T], val reverse : Boolean = false) exte
       }
     }
   }
+
+  /**
+   * Clone the graph structure, will not change the node element
```
yiheng: will not clone the ode element

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DirectedGraph.scala
```diff
@@ -123,6 +122,29 @@ class DirectedGraph[T](val source : Node[T], val reverse : Boolean = false) exte
...
+   * @return new graph
+   */
+  def cloneGraph(): DirectedGraph[T] = {
+    val oldToNew = new util.HashMap[Node[T], Node[T]]()
+    val oldNodes = new util.LinkedList[Node[T]]()
+    oldNodes.add(source)
+    oldToNew.put(source, source.cloneNode())
+    while (!oldNodes.isEmpty) {
```
yiheng: why not use the exist BFS or DFS method?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DirectedGraph.scala
```diff
@@ -193,8 +225,12 @@ class Node[T](val element: T) extends Serializable {
 
   private val nexts = new ArrayBuffer[Node[T]]()
   private val prevs = new ArrayBuffer[Node[T]]()
+
+  def cloneNode(): Node[T] = {
```
yiheng: This is not cloneNode. You didn't copy the edgs

yiheng: remove this method

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
yiheng: after user build a graph, he/she change the layer/trainable or stopGradient. Is this right?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -283,8 +323,65 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
...
 
-  def getExecutions : Array[Node[AbstractModule[Activity, Tensor[T], T]]] = {
-    return executions
+  /**
+   * set an array of layers that match the given ```names``` to be "freezed",
+   * i.e. their parameters(weight/bias, if exists) are not changed in training process
+   * @param names an array of layer names
+   * @return current graph model
+   */
+  def setFreeze(names: Array[String]): this.type = {
```
yiheng: setFreeze -> freeze

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -283,8 +323,65 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
...
+
+  /**
+   * stop the input gradient of layers that match the given ```names```
+   * their input gradient are not computed.
+   * And they will not contributed to the input gradient computation of
+   * layers that depend on them.
+   * @param names an array of layer names
+   * @return current graph model
+   */
+  def setStopGradient(names: Array[String]): this.type = {
```
yiheng: can we change the stopGradient back?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -167,8 +179,11 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
       })
 
       gradOutputBP(i) = curGradOutput
-      curNode.element.updateGradInput(inputsBP(i), curGradOutput)
-      i -= 1
+
+      if (!curNode.element.isStopGradient()) {
```
yiheng: why we need this? I think we already filter this layer in the backwardExecution generation

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -204,6 +205,41 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
...
+  }
+
+  /**
+   * set stop gradient. If isStopGradient is true, its input gradient
+   * is not computed. And it will not contributed to the input gradient computation of
+   * layers that depend on this layer.
+   * @param isStopGradient default is true.
+   * @return
+   */
+  def setStopGradient(isStopGradient: Boolean = true): this.type = {
```
yiheng: after user build a graph, he/she change the layer/trainable or stopGradient. Is this right?

docs/docs/APIGuide/Module.md
```diff
@@ -147,3 +147,262 @@ val preductResult = model.predict(predictSet)
  preductResult = model.predict(predictSet)
 ```
 
+## Module Freeze
+To "freeze" a module means to exclude some layers of model from training.
```
yiheng: means the parameter(weight, bias) of the module won't be changed during the training.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -205,6 +206,19 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
   }
 
   /**
+   * set trainable. If trainable is false, its parameters(weight/bias, if exists)
+   * are not changed in training process
+   * @param trainable whether this layer is trainable
+   */
+  def setTrainable(trainable: Boolean): this.type = {
```
yiheng: Looks like setTrainable(true) won't take any affect?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Container.scala
```diff
@@ -16,6 +16,7 @@
 
 package com.intel.analytics.bigdl.nn
 
+import com.intel.analytics.bigdl._
```
yiheng: please revert this change.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
SeaOfOcean: it will, if setTrainable is false, then scaleW=0 scaleB=0

yiheng: but if setTrainable is true, the user may expect that the layer become trainable again, but the method doesn't, right?

SeaOfOcean: maybe need to change name to freeze()

SeaOfOcean:
pr validate pass

Issue #1020, opened 2017-06-15
Can BigDL run on the windows local?


lawlietAi:
I'm using IDEA to develop the spark program **on windows local**
so for convenience i also hope to run bigDL program on windows.
but there are some methods missed when i attempt to run the model **Aotuencoder**.
i have added the jars of **spark-2.10-2.11** and download the bigDL's jars from maven.
what had i downloaded the bigDL dependencies from maven is 
**com.intel.analytics.bigdl:bigdl-SPARK_2.1:0.1.1-SNAPSHOT2**
should i download another jars?  
plz help me thank you

qiuxin2012:
Sorry, windows is unsupported right now, but will be supported in the future.
Please use linux or mac, thanks.

lawlietAi:
i see,thanks.await the windows version!

yiheng:
@lawlietAi I have uploaded mkl_win64.jar to maven. You can use the new added profile win64 to use it. Please give a try and let me know if there's any issue.

```bash
mvn clean package -DskipTests -P win64
```

or select the win64 profile in the IDE.

PR #1182, opened 2017-07-07
support windows platform

yiheng:
Support Windows platform(win64). 

This PR depends on the PR in BigDL-core
https://github.com/intel-analytics/BigDL-core/pull/13

Actually the core change is in that PR. In this PR
* Add a new profile to enable import the libmkl.jar built on windows plat
* Check and cancel some unit test which doesn't work on windows

## How was this patch tested?
Existing unit tests. PR validation pass

Have run mvn test on windows10 platform. and regression test on linux/mac.

Jenkin validation test.

#1020

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
@@ -150,12 +150,7 @@ object Engine {
    * Runtime.getRuntime().availableProcessors() / 2
    */
   private def getCoreNumberFromProperty() = {
-    val coreNumber = System.getProperty("bigdl.coreNumber", getNumMachineCores.toString).toInt
-    if (coreNumber > getNumMachineCores) {
-      getNumMachineCores
-    } else {
-      coreNumber
-    }
+    System.getProperty("bigdl.coreNumber", getNumMachineCores.toString).toInt
   }
 
   private def getNumMachineCores: Int = {
--- a/spark/dl/src/test/scala/com/intel/analytics/bigdl/dataset/TransformersSpec.scala
+++ b/spark/dl/src/test/scala/com/intel/analytics/bigdl/dataset/TransformersSpec.scala
@@ -24,7 +24,7 @@ import com.intel.analytics.bigdl.dataset.image._
 import com.intel.analytics.bigdl.dataset.text.{LabeledSentence, LabeledSentenceToSample}
 import com.intel.analytics.bigdl.tensor.{Storage, Tensor}
 import com.intel.analytics.bigdl.utils.RandomGenerator.RNG
-import com.intel.analytics.bigdl.utils.{Engine, RandomGenerator}
+import com.intel.analytics.bigdl.utils.{Engine, RandomGenerator, TestUtils}
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.Path
 import org.apache.hadoop.io.{SequenceFile, Text}
@@ -464,6 +464,7 @@ class TransformersSpec extends FlatSpec with Matchers with BeforeAndAfter {
   }
 
   "RGBImage To SeqFile without file name" should "be good" in {
+    TestUtils.cancelOnWindows()
     val resource = getClass().getClassLoader().getResource("imagenet")
     val pathToImage = LocalImgReaderWithName(BGRImage.NO_SCALE)
     val dataSet = DataSet.ImageFolder.paths(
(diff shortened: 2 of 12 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -150,12 +150,7 @@ object Engine {
    * Runtime.getRuntime().availableProcessors() / 2
    */
   private def getCoreNumberFromProperty() = {
-    val coreNumber = System.getProperty("bigdl.coreNumber", getNumMachineCores.toString).toInt
-    if (coreNumber > getNumMachineCores) {
-      getNumMachineCores
-    } else {
-      coreNumber
-    }
+    System.getProperty("bigdl.coreNumber", getNumMachineCores.toString).toInt
```
yiheng: I think here should at least throw a warning...

I remove it because I think sometimes user want to use more thread to run the program.

spark/dl/src/test/scala/com/intel/analytics/bigdl/dataset/TransformersSpec.scala
```diff
@@ -458,6 +458,7 @@ class TransformersSpec extends FlatSpec with Matchers with BeforeAndAfter {
   }
 
   "RGBImage To SeqFile without file name" should "be good" in {
+    TestUtils.cancelOnWindows()
```
yiheng: hadoop will try to call chmod underhood and it will cause a failure. So cancel it on windows

spark/dl/src/test/scala/com/intel/analytics/bigdl/optim/DistriOptimizerSpec.scala
```diff
@@ -328,6 +326,7 @@ class DistriOptimizerSpec extends FlatSpec with Matchers with BeforeAndAfter {
   }
 
   "TrainSummary with MSE and LBFGS" should "work correctly" in {
+    TestUtils.cancelOnWindows()
```
yiheng: hadoop will try to call chmod underhood and it will cause a failure. So cancel it on windows

spark/dl/src/test/scala/com/intel/analytics/bigdl/python/api/PythonSpec.scala
```diff
@@ -141,6 +141,7 @@ class PythonSpec extends FlatSpec with Matchers with BeforeAndAfter {
   }
 
   "Double prototype" should "be test" in {
+    TestUtils.cancelOnWindows()
```
yiheng: hadoop will try to call chmod underhood and it will cause a failure. So cancel it on windows

spark/dl/src/test/scala/com/intel/analytics/bigdl/torch/ColorJitterSpec.scala
```diff
@@ -30,13 +30,10 @@ import com.intel.analytics.bigdl.utils.RandomGenerator
 
 @com.intel.analytics.bigdl.tags.Serial
 class ColorJitterSpec extends FlatSpec with BeforeAndAfter with Matchers {
-  before {
+  "A ColorJitter" should "blend image correctly" in {
     if (!TH.hasTorch()) {
```
yiheng: This should be move into the test case from before. Or it will cause test suite failure

spark/dl/src/test/scala/com/intel/analytics/bigdl/utils/LoggerFilterSpec.scala
```diff
@@ -56,6 +56,7 @@ class LoggerFilterSpec extends FlatSpec with BeforeAndAfter with Matchers {
   }
 
   "A LoggerFilter" should "output correct info on console and bigdl.log" in {
+    TestUtils.cancelOnWindows()
```
yiheng: In this test suite, a lot of linux local folder path is used. Should we change to temp foler? @i8run

spark/dl/src/test/scala/com/intel/analytics/bigdl/visualization/SummarySpec.scala
```diff
@@ -91,6 +91,7 @@ class SummarySpec extends FlatSpec with Matchers with BeforeAndAfter {
   }
 
   "read/write event file" should "work properly" in {
+    TestUtils.cancelOnWindows()
```
yiheng: hadoop will try to call chmod underhood and it will cause a failure. So cancel it on windows

spark/dl/src/test/scala/com/intel/analytics/bigdl/utils/LoggerFilterSpec.scala
i8run: No problem.

PR #1189, opened 2017-07-10
add toGraph

zhangxiaoli73:

add toGraph, and Bigdl model can transfer to graph by call toGraph()

## How was this patch tested?

unit test (change inception, resnet, alexnet model to graph, and compare forward/backward results with original models')

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
@@ -279,6 +279,8 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
       "A graph container should not be changed after it is constructed")
   }
 
+  override def toGraph(): Module[T] = this
+
   // Add a dummy output node, to get an one end graph. So the nodes that are not dependent by
   // the outputs will be excluded
   private val dummyOutput = new ModuleNode[T](new Identity[T]())
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
@@ -279,7 +279,8 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
       "A graph container should not be changed after it is constructed")
   }
 
-  override def toGraph(): Module[T] = this
+  // todo: expand the graph
+  override def toGraph(startNodes: Seq[ModuleNode[T]]): Graph[T] = this
 
   // Add a dummy output node, to get an one end graph. So the nodes that are not dependent by
   // the outputs will be excluded
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
@@ -280,7 +280,7 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
   }
 
   // todo: expand the graph
-  override def toGraph(startNodes: Seq[ModuleNode[T]]): Graph[T] = this
+  override def toGraph(startNodes: ModuleNode[T]*): Graph[T] = this
 
   // Add a dummy output node, to get an one end graph. So the nodes that are not dependent by
   // the outputs will be excluded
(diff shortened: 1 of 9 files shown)


jason-dai:
Why do we need these?

zhangxiaoli73:
We can just use toGraph() to change bigdl models to graph models.The original code is from intern xinyue, I pulled his code, did some code change and committed a pr.

zhangxiaoli73:
Your request to do graph implementations should not do like this? @jason-dai

jason-dai:
It will be helpful if we can add corresponding functional model definition (in addition to the current sequential ones) for all the models/examples in BigDL, instead of having an automatic converter.

zhangxiaoli73:
OK, got it. I will close this pr, and add some corresponding functional model definition in other branch.

yiheng:
@jason-dai this auto-converter may be helpful when we want to do some graph-based optimization, e.g. memory sharing.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -473,5 +473,30 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
    batchSize: Option[Int] = None): Array[(ValidationResult, ValidationMethod[T])] = {
     Evaluator(this).test(dataset, vMethods, batchSize)
   }
+
+  /**
+   * Connect the current module to the current Graph with start nodes and end nodes
+   * @param startEnd: current start nodes and end nodes, they are both empty or both not empty
+   * @return
+   */
+  def toGraphNodes(startEnd: (Array[ModuleNode[T]], Array[ModuleNode[T]])):
```
yiheng: rename it to 'toGraph'?

Why use a tuple as input instead of two parameters?

def toGraph(starts: Array[ModuleNode[T]]): Array[ModuleNode[T]]

and 

def toGraph(): Array[ModuleNode[T]]

These methods take inputs node and output the end nodes. The second one creates some Identity nodes then invoke the first one.

zhangxiaoli73: ok, got it

yiheng: I don't think your implementation is align with my suggestion

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -668,5 +667,27 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
...
+   * Generate end nodes of current module with start nodes
+   * @param startNodes: current start nodes
+   * @return current end nodes
+   */
+  def getEndNodes(startNodes: Array[ModuleNode[T]]): Array[ModuleNode[T]] = {
+    val endNodes = Array(this.inputs(startNodes: _*))
+    endNodes
+  }
+
+  def toGraph(): Module[T] = {
```
yiheng: remove this method

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -668,5 +667,27 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
...
+    val startNodes = Array(Input[T]())
+    toGraph(startNodes)
+  }
+
+  def toGraph(startNodes: Array[ModuleNode[T]]): Module[T] = {
```
yiheng: Add Java doc. Change return type from Module[T] to Graph[T] and change Array to Seq

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -668,5 +667,27 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
...
     Validator(this, dataSet).test(vMethods)
   }
+
+
+  /**
+   * Generate end nodes of current module with start nodes
+   * @param startNodes: current start nodes
+   * @return current end nodes
+   */
+  def getEndNodes(startNodes: Array[ModuleNode[T]]): Array[ModuleNode[T]] = {
```
yiheng: protect method

yiheng:
More container getEndNodes

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -279,6 +279,8 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
       "A graph container should not be changed after it is constructed")
   }
 
+  override def toGraph(): Module[T] = this
```
yiheng: add a ToDo: expand the graph

yiheng:
Should you remove the change in core? Other looks good to me

zhangxiaoli73:
remove core, and jenkins pass

PR #1191, opened 2017-07-10
add a property to control the blockmanger behavior

yiheng:
For spark 1.5, we observe nio block manager has better performance than netty block manager. So we will force set block manager to nio. But some user want to control this behavior.

In this PR a new property ```bigdl.nio.force``` is added . The default value is true. User can set it to force to customize the blockmanager choice.

## How was this patch tested?
Add new unit test


jason-dai:
change to bigdl.network.nio

dding3:
I noticed "spark.shuffle.blockTransferService" is set in spark-bigdl.conf, can we ask user update this config file instead of adding another setting?

yiheng:
@dding3 this file is in the jar file and read by bigdl in runtime. User should modify the jar file to update the conf file.

yiheng:
@jason-dai changed

yiheng:
PR validation pass. merge the PR

PR #1192, opened 2017-07-10
fix convLSTMpeephole

JerryYanWan:
ConvLSTMPeephole's output will be set to Recurrent's output.
Thus, the output of ConvLSTMPeephole will be incontiguous which will 
throw an error in SpatialConvolution since it requires a contiguous input.

Add Contiguous layer inside ConvLSTMPeephole network.

## How was this patch tested?
Unit Test


dding3:
LGTM

PR #1195, opened 2017-07-10
add CosineDistanceCriterion

JerryYanWan:
Support CosineDistanceCriterion
Loss = 1 - cos(x, y)
Accept input and target as tensor.
Similar function compared to CosineEmbeddingCriterion.

## How was this patch tested?
Unit Test; compared to Torch CosineEmbeddingCriterion

https://github.com/intel-analytics/BigDL/issues/1186

### DIFF
--- a/docs/docs/APIdocs/Losses/CosineDistanceCriterion.md
+++ b/docs/docs/APIdocs/Losses/CosineDistanceCriterion.md
@@ -1,0 +1,74 @@
+## CosineDistanceCriterion ##
+
+**Scala:**
+```scala
+val criterion = CosineDistanceCriterion()
+```
+**Python:**
+```python
+criterion = CosineDistanceCriterion()
+```
+
+ This loss function measures the Cosine Distance between the target and the output
+``` 
+ loss(o, t) = 1 - cos(o, t)
+```
+ By default, the losses are averaged for each mini-batch over observations as well as over
+ dimensions. However, if the field sizeAverage is set to false, the losses are instead summed.
+
+**Scala example:**
+```scala
+import com.intel.analytics.bigdl.numeric.NumericFloat
+
+val criterion = CosineDistanceCriterion()
+val input = Tensor(1, 5).rand
+val target = Tensor(1, 5).rand
+val loss = criterion.forward(input, target)
+
+> println(target)
+target: com.intel.analytics.bigdl.tensor.Tensor[Float] =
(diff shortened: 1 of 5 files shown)


jason-dai:
Please add docs

JerryYanWan:
added

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/CosineDistanceCriterion.scala
```diff
@@ -0,0 +1,140 @@
...
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+import com.intel.analytics.bigdl.utils.Table
+
+import scala.reflect.ClassTag
+
+/**
+ * Creates a criterion that measures the loss given an input tensor and target tensor.
+ * loss(x, y) = 1 - cos(x, y)
+ */
+
```
shane-huang: need specifications about parameters

pyspark/bigdl/nn/criterion.py
```diff
@@ -190,6 +190,27 @@ def __init__(self,
                                                     n_classes)
 
 
+class CosineDistanceCriterion(Criterion):
+
+    """
+    Creates a criterion that measures the loss given an input and target,
+    Loss = 1 - cos(x, y)
+
+
```
shane-huang: explain parameters in API doc

docs/docs/APIdocs/Losses/CosineDistanceCriterion.md
```diff
@@ -0,0 +1,74 @@
+## CosineDistanceCriterion ##
+
+**Scala:**
+```scala
+val criterion = CosineDistanceCriterion()
```
shane-huang: add parameter `size_average=True`

docs/docs/APIdocs/Losses/CosineDistanceCriterion.md
```diff
@@ -0,0 +1,74 @@
...
+```
+**Python:**
+```python
+criterion = CosineDistanceCriterion()
```
shane-huang: add parameter `size_average=True`

pyspark/bigdl/nn/criterion.py
JerryYanWan: No other parameters in this layer. Only size_average.

The x and y are the forward(x, y).

PR #1196, opened 2017-07-10
Optim convlstm

dding3:
1. Add prepotopy for convlstm
2. Add convlstm doc
3. Some minor change

## How was this patch tested?
ut


JerryYanWan:
LGTM, please test on jenkins :)

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ConvLSTMPeephole.scala
```diff
@@ -43,35 +43,37 @@ import scala.reflect.ClassTag
...
-    val i2g = Sequential()
-      .add(Contiguous())
-      .add(SpatialConvolution(inputSize, hiddenSize, kernelI, kernelI,
-      stride, stride, kernelI/2, kernelI/2, wRegularizer = wRegularizer,
-        bRegularizer = bRegularizer))
+  override def preTopology: AbstractModule[Activity, Activity, T] =
+    Sequential()
+        .add(TimeDistributed(SpatialConvolution(inputSize, outputSize*4, kernelI, kernelI,
+          stride, stride, kernelI/2, kernelI/2, wRegularizer = wRegularizer,
+          bRegularizer = bRegularizer)))
```
JerryYanWan: I add a   `override def getPreTopologyNOutput: Int` in Cell. It is used to insert a BatchNormalization layer between preTopology and topology. If preTopology returns a Sequential, please all insert this line:

`  override def getPreTopologyNOutput: Int = 4 * outputSize`

JerryYanWan: A model definitin can be:

To insert a BatchNormalization layer:
```
val model = Sequential[Double]()
      .add(Recurrent[Double](BatchNormParams())
        .add(LSTMPeephole[Double](inputSize, hiddenSize)))
```

Or Not:
```

val model = Sequential[Double]()
      .add(Recurrent[Double]()
        .add(LSTMPeephole[Double](inputSize, hiddenSize)))
```

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ConvLSTMPeephole.scala
```diff
@@ -43,35 +43,37 @@ import scala.reflect.ClassTag
...
```
dding3: In convlstm, there is no dropout.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ConvLSTMPeephole.scala
dding3: For spatialconvolution, batchnormalization is a little different. Let me do it in another pr.

PR #1199, opened 2017-07-12
Remove example folder

zhichao-li:

We would put examples in this project in the future: https://github.com/intel-analytics/BigDL-Tutorials

## How was this patch tested?
unittest


zhichao-li:
cc @yiheng

PR #1215, opened 2017-07-12
Make ClassNLL can process -1 label for padding, not update the model.

psyyz10:
 If the target is `-1`, the training process will skip this sample.
 In other words, the forward process will return zero output and the backward process
 will also return zero `gradInput`.

## How was this patch tested?
Unit tests


JerryYanWan:
LGTM

PR #1233, opened 2017-07-14
Add Optim method markdown

zhangxiaoli73:

Add optim method makedown, mainly describe how to use  optim methods when training

### DIFF
--- a/docs/docs/APIdocs/Optimizers/OptimMethod.md
+++ b/docs/docs/APIdocs/Optimizers/OptimMethod.md
@@ -3,7 +3,6 @@
 OptimMethod is used to update model gradient parameters.We have defined SGD method, Adagrad method, etc.
 Details about those optim methods, you can refer to [Optim-Methods](https://github.com/intel-analytics/BigDL/tree/master/docs/docs/APIdocs/Optimizers/Optim-Methods).
 Now, method construct parameters(e.g."learningRate") and internal training parameters(e.g."epoch") store in optim method instead of state(since version 0.2.0)
-
 Here is mainly to describe how to use those methods when training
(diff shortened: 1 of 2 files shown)


docs/docs/APIdocs/Optimizers/OptimMethod.md
```diff
@@ -0,0 +1,90 @@
...
+
+optimizer.setOptimMethod(method) // set optim method
+
+optimizer.setCheckpoint(param.checkpoint.get, checkpointTrigger) // set checkpoint to save model and optim method
+
+val trainedModel = optimizer.optimize()
+```
+
+### Python example ###
+Here is an example to train LeNet5 model with a loading method.
```
shane-huang: So Python don's support loading an optimMethod snapshot? If Python don't support it.  Maybe we need to add it into description.

zhangxiaoli73: OK, I will add more description. But I think python should support loading optimMethod snapshot, I will add an issue.

docs/docs/APIdocs/Optimizers/OptimMethod.md
```diff
@@ -0,0 +1,90 @@
...
+  end_trigger=MaxEpoch(15),
+  batch_size=batch_size,
+  optim_method=SGD(learningrate=0.05) # set optim method
+)
+
+optimizer.set_validation(
+    batch_size=2048,
+    val_rdd=test_data,
+    trigger=EveryEpoch(),
+    val_method=["Top1Accuracy"]
```
zhangxiaoli73: yeah, ok, changed

Issue #1203, opened 2017-07-12
Missing md docs for Module


shane-huang:
Describe below functionalities. 
- load
- save
- predict
- evalute
- etc.

Update docs/docs/APIdocs/Model/ModuleAPI.md

shane-huang:
Python is Layer instead of Module in Scala. Do we need to make them sync?

wzhongyuan:
@shane-huang  sure, I will add a MD doc for this one. Currently I am working on the new serialization method, I will create one based on current master and refine it in future if needed once my change is in.

PR #1236, opened 2017-07-14
Module api

wzhongyuan:

Add module API  MD doc

## How was this patch tested?

Tested locally for all the example code

https://github.com/intel-analytics/BigDL/issues/1203

### DIFF
--- a/docs/docs/APIdocs/Model/ModuleAPI.md
+++ b/docs/docs/APIdocs/Model/ModuleAPI.md
@@ -1,0 +1,144 @@
+BigDL provides model API to faciliate user's requirement to save model to specific path, load model from given path, evaluate model and predict with model, etc.
+
+Here we introduce how to leverage this APIs
+
+
+Suppose we have a sequential model defined as below : 
+
+```
+Linear -> Sigmoid -> Softmax
+```
+
+**Scala:**
+
+###### Import necessary dependencies
+
+```scala
+import com.intel.analytics.bigdl.dataset.Sample
+import com.intel.analytics.bigdl.nn._
+import com.intel.analytics.bigdl.numeric.NumericFloat
+import com.intel.analytics.bigdl.optim.Top1Accuracy
+import com.intel.analytics.bigdl.tensor.Tensor
+```
+
+###### Define the model
+
+```scala
+val model = Sequential().add(Linear(10, 5)).add(Sigmoid()).add(SoftMax())
+```
+
+###### Save model to specific path
+
+BigDL supports different file systems like Linux file system, HDFS and AWS S3, Assume we want to save to local file system
+```scala
+model.save("/tmp/model.bigdl", true)
+```
+
+In the above code , the first parameter is the path where we want to save our model, the second paramter is to specify if we need to overwrite the file if it already exists, it's set to false by default
+
+###### Load model from given path
+
+We just need to specify the model path where we previously saved the model to load it to memory for resume training or prediction purpose
+
+```scala
+val loadedModel = Module.load("/tmp/model.bigdl")
+```
+`Module` above is a utilily  to manipulate module APIs
+
+###### Evaluate model with test set
+
+Evaluation needs real test data set, here we just simulate some dummy data for function introduction purpose
+
+Create dummy test data set
+
+```scala
+val feature = Tensor(10).rand()
+
+val label = Tensor(1).randn()
+
+val testSample = Sample(feature, label)
+    
+val testSet = sc.parallelize(Seq(testSample))
+```
+In above code, `sc` is the SparkContxt instance
+
+Use the test dataset to evaluate the model
+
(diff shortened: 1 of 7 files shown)


docs/docs/APIdocs/Model/ModuleAPI.md
```diff
@@ -0,0 +1,127 @@
...
+```
+###### Save model to specific path
+
+```python
+model.save("/tmp/model.bigdl", True)
+```
+###### Load model from given path
+
+```python
+model = Model.load("/tmp/model.bigdl")
```
shane-huang: can we add the import libraries here so that user will know which lib `Model` belongs to?

docs/docs/APIdocs/Model/ModuleAPI.md
```diff
@@ -0,0 +1,127 @@
...
+
+Create dummy test data set 
+
+```python
+samples=[Sample.from_ndarray(np.array([1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0, 10.0]), np.array([2.0]))]
+
+testSet = sc.parallelize(samples)
+
+```
+`np` above is numpy package
```
shane-huang: you can just use `import numpy as np` instead of describe what `np` is.

docs/docs/APIdocs/Model/ModuleAPI.md
```diff
@@ -0,0 +1,127 @@
...
+Predict with the data set
+```scala
+val preductResult = model.predict(predictSet)
+```
+
+**Python:**
+
+###### Define the model
+```python
+model = Sequential().add(Linear(10, 5)).add(Sigmoid()).add(SoftMax())
```
shane-huang: better add some `import...`

docs/docs/APIdocs/Model/ModuleAPI.md
```diff
@@ -0,0 +1,127 @@
...
+
+Use the test dataset to evaluate the model
+
+```scala
+val evaluateResult = model.evaluate(testSet, Array(new Top1Accuracy))
+```
+
+###### Predict with model
+Similar to above, we create input data first
+```scala
```
shane-huang: can we add some `import` sentences so that user will know which libs to import to use these API's?

docs/docs/APIdocs/Model/ModuleAPI.md
```diff
@@ -0,0 +1,127 @@
...
+Create dummy test data set
+
+```scala
+val feature = Tensor(10).apply1(e => Random.nextFloat())
+
+val label = Tensor(1).apply1(e => Random.nextFloat())
+
+val testSample = Sample(feature, label)
+    
+val testSet : RDD[Sample[Float]] = sc.parallelize(Seq(testSample))
```
shane-huang: remove [Float]?

docs/docs/APIdocs/Model/ModuleAPI.md
```diff
@@ -0,0 +1,127 @@
...
+```
+
+In the above code , the first parameter is the path where we want to save our model, the second paramter is to specify if we need to overwrite the file if it already exists, it's set to false by default
+
+###### Load model from given path
+
+We just need to specify the model path where we previously saved the model to load it to memory for resume training or prediction purpose
+
+```scala
+val loadedModel = Module.load("/tmp/model.bigdl")
```
shane-huang: Maybe some `import ...`?

wzhongyuan:
@shane-huang  I refined the file per your comments, please help check. Thanks

docs/docs/APIdocs/Model/ModuleAPI.md
```diff
@@ -0,0 +1,144 @@
...
+`Module` above is a utilily  to manipulate module APIs
+
+###### Evaluate model with test set
+
+Evaluation needs real test data set, here we just simulate some dummy data for function introduction purpose
+
+Create dummy test data set
+
+```scala
+val feature = Tensor(10).apply1(e => Random.nextFloat())
```
yiheng: Tensor(10).rand ?

docs/docs/APIdocs/Model/ModuleAPI.md
```diff
@@ -0,0 +1,144 @@
...
+
+val label = Tensor(1).apply1(e => Random.nextFloat())
```
yiheng: Tensor(1).random()?

wzhongyuan:
@yiheng  Modified.

yiheng:
Looks like the PR cannot be rebased...

wzhongyuan:
@yiheng  there should be no problem as it's a newly added file.

wzhongyuan:
there is some issues with my local files, I will create another one and merge directly after it's resolved @yiheng

PR #1220, opened 2017-07-13
fix inception and xavier

qiuxin2012:
## What changes were proposed in this pull request?

Inception's parameter init incorrect.

## How was this patch tested?

unit tests, manual tests

## Related links or issues (optional)
fixed #1219

PR #1244, opened 2017-07-17
fix more Xavier misusage

yangw1234:

fix more Xavier misusage, a follow up of pr #1220

## How was this patch tested?

existing tests


yiheng:
can we add some test cases to cover this?

yangw1234:
added unit tests and pr validation passed

PR #1247, opened 2017-07-17
modify the font of header

helenlly:

sync the style of  header, such as font, size, bold...

use the style  ## **header** for every page's header

### DIFF
--- a/docs/docs/UserGuide/install-build-src.md
+++ b/docs/docs/UserGuide/install-build-src.md
@@ -2,7 +2,7 @@
 
 BigDL source code is available at [GitHub](https://github.com/intel-analytics/BigDL)
 
-```sbt
+```bash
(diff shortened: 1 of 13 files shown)


docs/docs/UserGuide/install-build-src.md
```diff
@@ -54,47 +56,48 @@ $ bash make-dist.sh -P spark_2.x
...
     $ source PATH_TO_MKL/bin/mklvars.sh intel64
-    ```
+```
     where the `PATH_TO_MKL` is the installation directory of the MKL.
     
  3. Full build
    
    Clone BigDL as follows:
-   ```{r, engine='sh'}
+   ```sbt
```
shane-huang: we rarely use `sbt` now. Maybe we should use bash.

PR #1250, opened 2017-07-17
fix Sample's hashcode and equals

qiuxin2012:

Fix issue Sample without label's hashcode() will throw NullPointerException.

## How was this patch tested?

unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
@@ -130,8 +130,8 @@ private[bigdl] class ArraySample[T: ClassTag](
       private val data: Array[T],
       private val featureSize: Array[Array[Int]],
       private val labelSize: Array[Array[Int]]) extends Sample[T] {
-  require(data != null, "Sample's Data couldn't be empty")
-  require(featureSize != null, "Sample's Feature couldn't be empty")
+  require(data != null, "Sample: Data couldn't be empty")
+  require(featureSize != null, "Sample: Feature couldn't be empty")
 
   override def getData(): Array[T] = data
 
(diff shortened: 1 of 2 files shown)


yiheng:
LGTM

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
```diff
@@ -204,21 +205,20 @@ private[bigdl] class ArraySample[T: ClassTag](
...
-        !(labelSize.deep == that.labelSize.deep) ||
+        !(data.deep == that.data.deep) ||
         !(featureSize.deep == that.featureSize.deep)) {
         return false
       }
-      var i = labelSize.map(_.product).sum + featureSize.map(_.product).sum - 1
-      while (i >= 0) {
-        if (data(i) != that.data(i)) return false
-        i -= 1
+      if (null != featureSize && null != that.featureSize) {
```
zhangxiaoli73: If featureSize.deep != that.featureSize.deep, it will return directly. Why do you judge them again here?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
```diff
@@ -130,7 +130,8 @@ private[bigdl] class ArraySample[T: ClassTag](
       private val data: Array[T],
       private val featureSize: Array[Array[Int]],
       private val labelSize: Array[Array[Int]]) extends Sample[T] {
-  require(featureSize != null, "Feature couldn't be empty")
+  require(data != null, "Sample's Data couldn't be empty")
```
zhangxiaoli73: please add class name in error message

qiuxin2012:
jenkins passed

PR #1271, opened 2017-07-18
refactor file and add s3 support for CaffePersister

qiuxin2012:

refactor file and add s3 support for CaffePersister

## How was this patch tested?

unit tests, integration tests, manual tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/File.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/File.scala
@@ -22,64 +22,9 @@ import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.{FSDataInputStream, FSDataOutputStream, FileSystem, Path}
 import org.apache.hadoop.io.IOUtils
 
-import scala.collection.mutable.ArrayBuffer
-
-/**
- * A uniform approach to get Input/Output stream in BigDL.
- * @param fileName
- */
-private[bigdl] class File(fileName: String) {
-  private val inputStreams: ArrayBuffer[InputStream] = new ArrayBuffer[InputStream]()
-  private var outputStream: OutputStream = null
-  private val conf = File.getConfiguration(fileName)
-  private val path = new Path(fileName)
-  private val fs: FileSystem = path.getFileSystem(conf)
-
-  /**
-   * get an InputStream
-   * @return
-   */
-  def open(): InputStream = {
-    require(fs.exists(path), s"$fileName is empty!")
-    val inputStream = fs.open(path)
-    inputStreams.append(inputStream)
-    inputStream
-  }
-
-  /**
-   * get an OutputStream
-   * @param overwrite if overwrite
-   * @return
-   */
-  def create(overwrite: Boolean = false): OutputStream = {
-    require(outputStream == null, s"File $fileName has been opened already.")
-    if (!overwrite) {
-      require(!fs.exists(path), s"$fileName already exists!")
-    }
-    outputStream = fs.create(path, overwrite)
-    outputStream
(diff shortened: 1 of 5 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/File.scala
```diff
@@ -17,16 +17,70 @@
...
+/**
+ * A uniform approach to get Input/Output stream in BigDL.
+ * @param fileName
+ */
+private[bigdl] class File(fileName: String) {
+  private val inputStreams: ArrayBuffer[InputStream] = new ArrayBuffer[InputStream]()
+  private var outputStream: OutputStream = null
+  private val conf = File.getConfiguration(fileName)
+  private val path = new Path(fileName)
+  private val fs: FileSystem = path.getFileSystem(conf)
```
wzhongyuan: seems this piece of code specifically for HDFS ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/File.scala
```diff
@@ -17,16 +17,70 @@
...
 
+import scala.collection.mutable.ArrayBuffer
+
+/**
+ * A uniform approach to get Input/Output stream in BigDL.
+ * @param fileName
+ */
+private[bigdl] class File(fileName: String) {
+  private val inputStreams: ArrayBuffer[InputStream] = new ArrayBuffer[InputStream]()
+  private var outputStream: OutputStream = null
```
wzhongyuan: why do we need a declared variable ? I didn't see any usage inside the class  (just assigin a created output stream to it and close it), the create method below just return the outoutstream to external reference and we could close it there as you do it now

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/File.scala
```diff
@@ -17,16 +17,70 @@
...
 import org.apache.hadoop.io.IOUtils
 
+import scala.collection.mutable.ArrayBuffer
+
+/**
+ * A uniform approach to get Input/Output stream in BigDL.
+ * @param fileName
+ */
+private[bigdl] class File(fileName: String) {
+  private val inputStreams: ArrayBuffer[InputStream] = new ArrayBuffer[InputStream]()
```
wzhongyuan: do we have the scenario we need to maintain multiple input stream for the same file ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/File.scala
qiuxin2012: also works for local file system and s3.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/caffe/CaffePersister.scala
```diff
@@ -211,9 +211,9 @@ class CaffePersister[T: ClassTag](val prototxtPath: String,
       })
     }
 
-    var prototxtFile: File = null
+    var prototxtFile: FileWriter = null
     try {
-      prototxtFile = File(prototxtPath)
+      prototxtFile = FileWriter(prototxtPath)
```
wzhongyuan: Generally change the prototxtFile to prototxtFileWriter would make more sense ?

wzhongyuan:
@qiuxin2012   this generally looks good to me, only thing that should we modify a little bit on the naming like from ***file to ***fileWriter(Reader) as we separate reading and writing to different objects
thanks

wzhongyuan:
LGTM

qiuxin2012:
jenkins passed

PR #1277, opened 2017-07-18
change os-flag when win64 is set

yiheng:
change os-flag when win64 is set and update build docs

## How was this patch tested?
PR validation

https://github.com/intel-analytics/BigDL/issues/1020


yiheng:
PR validation pass

PR #1283, opened 2017-07-19
fix rnn example

JerryYanWan:
RNN LM example should perform well using SampleToMinibatch

## How was this patch tested?
Run the example and observe the convergence.


spark/dl/src/main/scala/com/intel/analytics/bigdl/models/rnn/Train.scala
```diff
@@ -70,21 +70,25 @@ object Train {
       val endIdx = dictionary.getIndex(SentenceToken.end)
       val padFeature = Tensor[Float]().resize(totalVocabLength)
       padFeature.setValue(endIdx + 1, 1.0f)
-      val padLabel = startIdx
+      val padLabel = Tensor[Float](Storage(Array[Float](startIdx)), 1, Array(1))
```
qiuxin2012: Tensor[Float](T(startIdx.toFloat))

zhangxiaoli73:
Code LGTM

PR #1295, opened 2017-07-20
Recurrent layer: fix batchSize non-consistency

JerryYanWan:
When batchsize changes, the recurrent layer will perform wrongly.

## How was this patch tested?
Unit Test


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -275,6 +286,15 @@ class Recurrent[T : ClassTag]()
   }
 
   override def updateGradInput(input: Tensor[T], gradOutput: Tensor[T]): Tensor[T] = {
+    val batchSizeOfCell = if (!cells.isEmpty) {
+      cells.head.output.toTable[Tensor[T]](inputDim).size(batchDim)
```
qiuxin2012: gradInput？

qiuxin2012:
LGTM

PR #1302, opened 2017-07-20
Fix caffe parameter exception

SeaOfOcean:

Fix caffe parameter exception

## How was this patch tested?

manual local test


wzhongyuan:
LGTM

Issue #1294, opened 2017-07-20
Link error in python doc Example link


hhbyyh:
The examples in the python document should not point to the master branch(should use the release branch) since the master branch will keep evolving.

https://bigdl-project.github.io/master/#PythonSupport/python-examples/

This tutorial describes the textclassifier example written...

The link for textclassifier is wrong, it should be https://github.com/intel-analytics/BigDL/tree/master/pyspark/bigdl/models/textclassifier while it's using the old link (https://github.com/intel-analytics/BigDL/tree/master/pyspark/bigdl/models/textclassifier)

jenniew:
On https://bigdl-project.github.io/master/#PythonSupport/python-resources/, the links for lenet, textclassifier and Jupyter notebook tutorial are wrong too. They are pointing to old "dl" package not "bigdl" package.

Issue #1280, opened 2017-07-18
Python install doc should be improved.


hhbyyh:
https://bigdl-project.github.io/latest/#PythonSupport/python-install/ is the first page of the python doc, yet it's missing important information for users. Some suggestions for the improvement:
1. add a general introduction segment, which version of python is supported. Which distribution has been tested. Install on Master or all the worker nodes? Which version of Spark is supported.
2. If the pip install is the preferred way to install BigDL for python,  "Use Python without pip Install" should be moved to the bottom.
2. The title and the first graph for "Use Python without pip Install" should be changed. Does it mean "Install BigDL without pip install"? And I think all the "Use python" should be replaced by "Installing BigDL".

hhbyyh:
Also 
tar -zxvf spark-1.6.3-bin-hadoop2.6.tgz
export SPARK_HOME=path to spark-1.6.3-bin-hadoop2.6.tgz

the export command should set the path after decompression.

Issue #1259, opened 2017-07-17
Update python document


jason-dai:
Install (https://github.com/intel-analytics/BigDL/blob/master/docs/docs/PythonSupport/python-install.md): need to specify how to get Spark and how to set the ENV variables

Run (https://github.com/intel-analytics/BigDL/blob/master/docs/docs/PythonSupport/python-run.md): please update (e.g., no bigdl.sh)

shane-huang:
Need also specify the limitations of pip install:
- pip only supports certain version
- pip install can not be on cluster deployment

shane-huang:
Python examples also needs to update to use 2.0 API. 
`optim_method="Adagrad"` should became `optim_method=Adagrad ()`
```
optimizer = Optimizer(
    model=build_model(news20.CLASS_NUM),
    training_rdd=train_rdd,
    criterion=ClassNLLCriterion(),
    end_trigger=MaxEpoch(max_epoch),
    batch_size=batch_size,
    optim_method="Adagrad",
    state=state)
...
train_model = optimizer.optimize()
```

shane-huang:
Maybe also revise the navigation layout

jason-dai:
And update how to pack python dependency for YARN

JerryYanWan:
https://bigdl-project.github.io/latest/#PythonSupport/python-install/
In "Install BigDL-0.2.0-snapshot", command typo:
should be
`wget https://d3kbcqa49mib13.cloudfront.net/spark-1.6.3-bin-hadoop2.6.tgz`

PR #1304, opened 2017-07-20
Update python doc

zhichao-li:
https://github.com/intel-analytics/BigDL/issues/1259
https://github.com/intel-analytics/BigDL/issues/1280
https://github.com/intel-analytics/BigDL/issues/1294


zhichao-li:
Merging this for now then we can have a fresh doc site via the nightly build. we can have a follow up PR for the doc if there's anything left.

PR #1315, opened 2017-07-20
update download link for nightly build

yiheng:
update download link for nightly build

## How was this patch tested?
doc change, manul check the links


SeaOfOcean:
LGTM

PR #1317, opened 2017-07-20
fix: change the order of layers

### LINKED ISSUE #1311: change the order of layers

i8run:

fix: change the order of layers

fixed #1311


shane-huang:
Why this commit?
fix: delete docs/extra.css

Issue #1292, opened 2017-07-19
Compile error with Spark 2.2.0 / BlockManager.getLocalBytes changed in Spark 2.2.0


srowen:
```
mvn -Pspark_2.x -Dspark.version=2.2.0 clean compile
...
[error] /Users/srowen/Documents/Cloudera/BigDL/spark/dl/src/main/scala/com/intel/analytics/bigdl/parameters/AllReduceParameter.scala:194: type mismatch;
[error]  found   : Object
[error]  required: org.apache.spark.util.io.ChunkedByteBuffer
[error]               val localBuffer = BlockManagerWrapper.byteBufferConvert(weightBuffer)
[error]                                                                       ^
[error] /Users/srowen/Documents/Cloudera/BigDL/spark/dl/src/main/scala/com/intel/analytics/bigdl/parameters/AllReduceParameter.scala:229: type mismatch;
[error]  found   : Object
[error]  required: org.apache.spark.util.io.ChunkedByteBuffer
[error]               .getOrElse(bm.getRemoteBytes(blockId).get))
[error]         
```

The signature of `BlockManager.getLocalBytes` changed from Spark 2.1.x to Spark 2.2.0 in https://github.com/apache/spark/commit/b56ad2b1ec19fd60fa9d4926d12244fd3f56aca4 . 

I know that's not how BigDL is built, but it ends up causing problems at runtime. For example, @nishamuktewar noted this happens when you run on CDH 5.12 + Spark 2.2:

```
ERROR optim.DistriOptimizer$: Error: org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 11.0 failed 4 times, most recent failure: Lost task 0.3 in stage 11.0 (TID 40, bottou02.sjc.cloudera.com, executor 2): java.util.concurrent.ExecutionException: java.lang.ClassCastException: org.apache.spark.storage.ByteBufferBlockData cannot be cast to org.apache.spark.util.io.ChunkedByteBuffer
  at java.util.concurrent.FutureTask.report(FutureTask.java:122)
  at java.util.concurrent.FutureTask.get(FutureTask.java:192)
  at com.intel.analytics.bigdl.parameters.FutureResult$$anonfun$waitResult$1.apply(AllReduceParameter.scala:244)
  at com.intel.analytics.bigdl.parameters.FutureResult$$anonfun$waitResult$1.apply(AllReduceParameter.scala:244)
[...]

PR #1321, opened 2017-07-20
Move references to BlockManager from AllReduceParameter into BlockMan…

srowen:

Move references to BlockManager from AllReduceParameter into BlockManagerWrapper. Offer new getLocalOrRemoteBytes method instead. Use reflection internally to be compatible with Spark 2.2 but also earlier Spark 2.

## How was this patch tested?

Built vs Spark 1.5, 1.6, 2.0, 2.1, 2.2.
I believe full tests need to be run to verify the change.

fixed https://github.com/intel-analytics/BigDL/issues/1292

### DIFF
--- a/spark/spark-version/2.0/src/main/scala/org/apache/spark/storage/BlockManagerWrapper.scala
+++ b/spark/spark-version/2.0/src/main/scala/org/apache/spark/storage/BlockManagerWrapper.scala
@@ -50,14 +50,48 @@ object BlockManagerWrapper {
     SparkEnv.get.blockManager.removeBlock(blockId)
   }
 
-  def byteBufferConvert(chunkedByteBuffer: ChunkedByteBuffer): ByteBuffer = {
-    ByteBuffer.wrap(chunkedByteBuffer.toArray)
+  def getLocalOrRemoteBytes(blockId: BlockId): Option[ByteBuffer] = {
+    val maybeLocalBytes = getLocalBytesFn(blockId)
+    if (maybeLocalBytes.isDefined) {
+      maybeLocalBytes
+    } else {
+      SparkEnv.get.blockManager.getRemoteBytes(blockId).map(_.toByteBuffer)
+    }
   }
 
   def unlock(blockId : BlockId): Unit = {
     val blockInfoManager = SparkEnv.get.blockManager.blockInfoManager
     if(blockInfoManager.get(blockId).isDefined) {
       blockInfoManager.unlock(blockId)
+    }
+  }
+
+  private val getLocalBytesFn: (BlockId) => Option[ByteBuffer] = {
+    val bmClass = classOf[BlockManager]
+    val getLocalBytesMethod = bmClass.getMethod("getLocalBytes", classOf[BlockId])
+
+    // Spark versions before 2.2.0 declare:
+    // def getLocalBytes(blockId: BlockId): Option[ChunkedByteBuffer]
+    // Spark 2.2.0+ declares:
+    // def getLocalBytes(blockId: BlockId): Option[BlockData]
+    // Because the latter change happened in the commit that introduced BlockData,
+    // and because you can't discover the generic type of the return type by reflection,
+    // distinguish the cases by seeing if BlockData exists.
+    try {
+      val blockDataClass = Class.forName("org.apache.spark.storage.BlockData")
+      // newer method, apply reflection to transform BlockData after invoking
+      val toByteBufferMethod = blockDataClass.getMethod("toByteBuffer")
+      (blockId: BlockId) =>
+        getLocalBytesMethod.invoke(SparkEnv.get.blockManager, blockId)
+          .asInstanceOf[Option[_]]
+          .map(blockData => toByteBufferMethod.invoke(blockData).asInstanceOf[ByteBuffer])
+    } catch {
+      case _: ClassNotFoundException =>
+        // older method, can be invoked directly
+        (blockId: BlockId) =>
+          getLocalBytesMethod.invoke(SparkEnv.get.blockManager, blockId)
+            .asInstanceOf[Option[ChunkedByteBuffer]]
+            .map(_.toByteBuffer)
     }
   }
 
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/parameters/AllReduceParameter.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/parameters/AllReduceParameter.scala
@@ -25,7 +25,7 @@ import org.apache.commons.lang.exception.ExceptionUtils
 import org.apache.log4j.Logger
 import org.apache.spark.sparkExtension.SparkExtension
 import org.apache.spark.storage.{BlockId, BlockManagerWrapper, StorageLevel}
-import org.apache.spark.{SparkEnv, TaskContext}
+import org.apache.spark.TaskContext
 
 import scala.collection.JavaConverters._
 import scala.reflect._
@@ -178,20 +178,16 @@ class AllReduceParameter[T: ClassTag](id: Long, partitionNum: Int, size: Int) ex
    * @return A [[FutureResult]] which contains a [[Future]] for each thread.
    */
   def getWeights(localParameter: Tensor[T]): FutureResult[Int] = {
-    val bm = SparkEnv.get.blockManager
     val tasks = (0 until partitionNum).map { pid =>
       syncPool.submit {
         new Callable[Int] {
           override def call(): Int = {
             try {
               val blockId = getWeightBlockId(pid)
-              val weightBuffer = bm.getLocalBytes(blockId).getOrElse {
-                bm.getRemoteBytes(blockId).getOrElse {
-                  throw new RuntimeException(s"Didn't find weight block $blockId in the block " +
-                    s"manager. Did you initialize this AllReduceParameter on every executor?")
-                }
+              val localBuffer = BlockManagerWrapper.getLocalOrRemoteBytes(blockId).getOrElse {
+                throw new RuntimeException(s"Didn't find weight block $blockId in the block " +
+                  s"manager. Did you initialize this AllReduceParameter on every executor?")
               }
-              val localBuffer = BlockManagerWrapper.byteBufferConvert(weightBuffer)
               val start = pid * taskSize + math.min(pid, extraSize)
               val length = taskSize + (if (pid < extraSize) 1 else 0)
               require(localBuffer.array().length == length * 2)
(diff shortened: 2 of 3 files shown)


srowen:
@nishamuktewar maybe you can build my branch and test vs CDH + Spark 2.2? I know you are set up for it so maybe it's easy to try

spark/spark-version/2.0/src/main/scala/org/apache/spark/storage/BlockManagerWrapper.scala
```diff
@@ -61,6 +66,34 @@ object BlockManagerWrapper {
...
   }
 
+  private val getLocalBytesFn: (BlockId) => Option[ByteBuffer] = {
+    val bmClass = classOf[BlockManager]
+    // Spark versions before 2.2.0 declare:
+    // def getLocalBytes(blockId: BlockId): Option[ChunkedByteBuffer]
+    // Spark 2.2.0+ declares:
+    // def getLocalBytes(blockId: BlockId): Option[BlockData]
+    val getLocalBytesMethod = bmClass.getMethod("getLocalBytes", classOf[BlockId])
+    if (getLocalBytesMethod.getReturnType == classOf[ChunkedByteBuffer]) {
```
srowen: Ah, this doesn't work. The type is `Option[_]`. Working on it.

nishamuktewar:
@srowen built using your branch and tested it, another exception this time:

Exception in thread "main" java.io.InvalidClassException: com.intel.analytics.bigdl.nn.abstractnn.AbstractModule; local class incompatible: stream classdesc serialVersionUID = -4389769600335923191, local class serialVersionUID = 6252017636270165777
	at java.io.ObjectStreamClass.initNonProxy(ObjectStreamClass.java:621)
	at java.io.ObjectInputStream.readNonProxyDesc(ObjectInputStream.java:1623)
	at java.io.ObjectInputStream.readClassDesc(ObjectInputStream.java:1518)
	at java.io.ObjectInputStream.readNonProxyDesc(ObjectInputStream.java:1623)
	at java.io.ObjectInputStream.readClassDesc(ObjectInputStream.java:1518)
	at java.io.ObjectInputStream.readNonProxyDesc(ObjectInputStream.java:1623)
	at java.io.ObjectInputStream.readClassDesc(ObjectInputStream.java:1518)
	at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:1774)
	at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1351)
	at java.io.ObjectInputStream.readObject(ObjectInputStream.java:371)
	at com.intel.analytics.bigdl.utils.File$.load(File.scala:172)
	at com.intel.analytics.bigdl.nn.Module$.load(Module.scala:42)
	at com.cloudera.datascience.dl.bigdl.FineTuneTrain$$anonfun$main$1.apply(FineTuneTrain.scala:44)
	at com.cloudera.datascience.dl.bigdl.FineTuneTrain$$anonfun$main$1.apply(FineTuneTrain.scala:21)
	at scala.Option.map(Option.scala:146)
	at com.cloudera.datascience.dl.bigdl.FineTuneTrain$.main(FineTuneTrain.scala:21)
	at com.cloudera.datascience.dl.bigdl.FineTuneTrain.main(FineTuneTrain.scala)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.apache.spark.deploy.SparkSubmit$.org$apache$spark$deploy$SparkSubmit$$runMain(SparkSubmit.scala:755)
	at org.apache.spark.deploy.SparkSubmit$.doRunMain$1(SparkSubmit.scala:180)
	at org.apache.spark.deploy.SparkSubmit$.submit(SparkSubmit.scala:205)
	at org.apache.spark.deploy.SparkSubmit$.main(SparkSubmit.scala:119)
	at org.apache.spark.deploy.SparkSubmit.main(SparkSubmit.scala)

srowen:
That's a different issue, yeah. It suggests that the serialized form of a class has changed in the new build. Are you loading an existing model? I'd remove that or put it aside and train a new one.

srowen:
@yiheng what do you think of this approach? I believe it resolves the compatibility issue in the same way as before and restores Spark 2.2 compatibility. Would be great to have that work in an upcoming 0.2.0 release.

yiheng:
@srowen The change looks good to me. I'm running some test.

@dding3 Can you take a look at this PR?

yiheng:
Some errors were reported when I run validation jenkins job. I found that in **DistriOptimizerSpec** class, if you run a single test case, they all pass; if you run them all together, only the first one pass. The error message is

```
2017-07-26 17:40:28 ERROR AllReduceParameter$:199 - Error: java.lang.AssertionError: assertion failed: Block test_1weightBytes3 is not locked for reading
	at scala.Predef$.assert(Predef.scala:170)
	at org.apache.spark.storage.BlockInfoManager.unlock(BlockInfoManager.scala:294)
	at org.apache.spark.storage.BlockManagerWrapper$.unlock(BlockManagerWrapper.scala:65)
	at com.intel.analytics.bigdl.parameters.AllReduceParameter$$anonfun$1$$anon$2.call(AllReduceParameter.scala:195)
	at com.intel.analytics.bigdl.parameters.AllReduceParameter$$anonfun$1$$anon$2.call(AllReduceParameter.scala:183)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

It doesn't take place in the master code.

srowen:
Yeah I noticed that too and wasn't sure whether it was just that my local test env wasn't able to run tests in the same way, or a real problem. Or maybe some other problem uncovered by this change.

The call to `unlock` didn't change. I can only guess that it happens when getLocalBytes doesn't return a block (so it's not locked for reading) but getRemoteBytes does, and the block then exists, so later the call to `unlock` thinks it needs to be unlocked, but it can't be.

However I don't see why that wasn't already possible. The logic is essentially identical as far as I can see: getLocalBytes, then try getRemoteBytes. I'm puzzled.

It seems like `unlock` needs to check whether `info.writerTask != BlockInfo.NO_WRITER`? That's what Spark is checking. That feels kind of fragile.

Or maybe `getLocalBytes` needs to call unlock after it gets the data, if it succeeds? because it will have copied the result into a `ByteBuffer`? and then the BigDL code shouldn't try to manually unlock.

yiheng:
Looks like change the **getLocalBytesFn** method like this, the issue goes.

I moved the ``val bm = SparkEnv.get.blockManager`` into the function definiation.

I guess when invoke getLocalBytesFn, the ``SparkEnv.get.blockManager`` return a different instance from the one when create getLocalBytesFn
 
```scala
private val getLocalBytesFn: (BlockId) => Option[ByteBuffer] = {
    val bmClass = classOf[BlockManager]
    val getLocalBytesMethod = bmClass.getMethod("getLocalBytes", classOf[BlockId])

    // Spark versions before 2.2.0 declare:
    // def getLocalBytes(blockId: BlockId): Option[ChunkedByteBuffer]
    // Spark 2.2.0+ declares:
    // def getLocalBytes(blockId: BlockId): Option[BlockData]
    // Because the latter change happened in the commit that introduced BlockData,
    // and because you can't discover the generic type of the return type by reflection,
    // distinguish the cases by seeing if BlockData exists.
    try {
      val blockDataClass = Class.forName("org.apache.spark.storage.BlockData")
      // newer method, apply reflection to transform BlockData after invoking
      val toByteBufferMethod = blockDataClass.getMethod("toByteBuffer")
      (blockId: BlockId) => {
        val bm = SparkEnv.get.blockManager
        getLocalBytesMethod.invoke(bm, blockId).asInstanceOf[Option[_]]
          .map(blockData => toByteBufferMethod.invoke(blockData).asInstanceOf[ByteBuffer])
      }
    } catch {
      case _: ClassNotFoundException =>
        // older method, can be invoked directly
        (blockId: BlockId) => {
          val bm = SparkEnv.get.blockManager
          getLocalBytesMethod.invoke(bm, blockId).asInstanceOf[Option[ChunkedByteBuffer]]
            .map(_.toByteBuffer)
        }
    }
  }
```

yiheng:
raise a PR for fixing the issue.
https://github.com/srowen/BigDL/pull/1

srowen:
Ah nice catch @yiheng . I updated it and rebased here.

spark/spark-version/2.0/src/main/scala/org/apache/spark/storage/BlockManagerWrapper.scala
```diff
@@ -61,6 +66,35 @@ object BlockManagerWrapper {
...
+    } catch {
+      case _: ClassNotFoundException =>
+        // older method, can be invoked directly
+        (blockId: BlockId) =>
+          getLocalBytesMethod.invoke(SparkEnv.get.blockManager, blockId)
+            .asInstanceOf[Option[ChunkedByteBuffer]]
+            .map(_.toByteBuffer)
+    }
+  }
+
```
dding3: getLocalBytesFn will be heavily called during training. In current implementation with spark 2.0, spark 2.1, code will frequently go to "catch exception" block and it will affect training performance. Can we add a flag which is used to mark if BlockData exists and set it for the first time getLocalBytesFn called. After that we can simply check this flag to decide which getLocalBytesMethod should be invoked. In this way, we don't need go to catch execption every time getLocalBytesFn is called which can avoid performance downgrade.

yiheng:
jenkins job passed

spark/spark-version/2.0/src/main/scala/org/apache/spark/storage/BlockManagerWrapper.scala
yiheng: I think here the catchException only execute once(when create the getLocalBytesFn function object)

dding3: OK. I see. I didn't notice it's a object.

dding3:
LGTM

zhangxiaoli73:
Test inception and vgg model with this code, results:
inception: 1 epoch, throughput: 385.63
vgg: 15 epoch, throughput: 597.408

zhangxiaoli73:
inception another test: Average throughput =  392.11
checkout master and test results as follows:
Average throughput =  392.386  iters: 818
Average throughput =  402.211  iters: 818

zhangxiaoli73:
inception another test: Average throughput = 392.11
checkout master and test results as follows:
inception:
Average throughput = 392.386 iters: 818
Average throughput = 402.211 iters: 818

PR #1328, opened 2017-07-20
update Simple-Layer.md and Activations.md to modify some display format

helenlly:

modify some display format:

1) add a line between each section
2) use `` to highlight maths formats and parameters
3) change the word" the output is..." to " gives the output"
4) remove bold font for some words 
5)  correct the image link of ELU.png


helenlly:
shane, pls help to review the format; if it is ok, I will modify other MD files as well.

PR #1329, opened 2017-07-20
TemporalConvolutionSpec should use TorchSpec to avoid failing the build on non-Torch machines

srowen:

TemporalConvolutionSpec should use TorchSpec to avoid failing the build on non-Torch machines. Right now, `mvn package` will fail instead of just marking this suite as cancelled.

## How was this patch tested?

Existing tests


yiheng:
LGTM. PR validation pass. merged.

PR #1357, opened 2017-07-24
remove unnecessary contiguous check in volumetricconvolution

dding3:
Looks like input is not used for calculation in updateGradInput and accGradParameters, I think we don't require it's contiguous. And if input.dim==5, in updateGradInput, we require sliced gradOutputT is contiguous but looks like we don't need gradOutput is contiguous. 

## How was this patch tested?
ut

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/VolumetricConvolution.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/VolumetricConvolution.scala
@@ -265,7 +265,7 @@ class VolumetricConvolution[T: ClassTag](
         val gradInputT = gradInput.select(1, t)
         val gradOutputT = gradOutput.select(1, t)
         val fGradInputT = fGradInput.select(1, t)
-        require(gradOutputT.isContiguous())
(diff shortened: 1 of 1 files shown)


SeaOfOcean:
input need to be contiguous in forward

dding3:
Yes, i understand and i keep input contigus check in forward. I think we don't require input is contigus in backward as it looks like not used in calculation, no?

SeaOfOcean:
Yes, but since the input for forward and backward is the same tensor, so I also check it in backward

SeaOfOcean:
we can remove the check of input here

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/VolumetricConvolution.scala
```diff
@@ -266,6 +265,7 @@ class VolumetricConvolution[T: ClassTag](
         val gradInputT = gradInput.select(1, t)
         val gradOutputT = gradOutput.select(1, t)
         val fGradInputT = fGradInput.select(1, t)
+        require(gradOutputT.isContiguous())
```
SeaOfOcean: add error message in require

SeaOfOcean:
LGTM

PR #1361, opened 2017-07-25
Convlstm3d

dding3:
support 3d image processing for convlstm

## How was this patch tested?
ut


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Cell.scala
```diff
@@ -81,39 +81,39 @@ abstract class Cell[T : ClassTag](
...
+        if (null == imageSize) {
           while (i <= hidden.toTable.length()) {
-            hidden.toTable[Tensor[T]](i).resize(size, hiddensShape(i - 1))
+            hidden.toTable[Tensor[T]](i).resize(batchSize, hiddensShape(i - 1))
             i += 1
           }
         } else {
           while (i <= hidden.toTable.length()) {
-            hidden.toTable[Tensor[T]](i).resize(size, hiddensShape(i - 1), rows, columns)
+            hidden.toTable[Tensor[T]](i).resize(Array(batchSize, hiddensShape(i - 1)) ++ imageSize)
```
JerryYanWan: hidResize will be called for each input. Thus, ++ operation will new three arrays and trigger minor gc.

Can imageSize be a fixed dimension array?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -81,12 +81,13 @@ class Recurrent[T : ClassTag]()
...
-   * @param hiddenSize
+   * @param sizes, the first element is batchSize, the second is times, the third is hiddensize
+    *             the left is size of images
    */
-  private def extend(times: Int, batchSize: Int, hiddenSize: Int,
-    rows: Int = 1, columns: Int = 1): Unit = {
+  private def extend(sizes: Array[Int]): Unit = {
+    val times = sizes(1)
+    val batchSize = sizes(0)
+    val imageSize = sizes.drop(3)
```
JerryYanWan: drop will call slice method, which will return a new array.. same issue with minor gc..

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -229,15 +230,11 @@ class Recurrent[T : ClassTag]()
...
-    if (input.dim() == 3) {
-      output.resize(batchSize, times, hiddenSize)
-      // Clone N modules along the sequence dimension.
-      extend(times, batchSize, hiddenSize)
-    } else if (input.dim() == 5) {
-      output.resize(batchSize, times, hiddenSize, input.size(4), input.size(5))
-      // Clone N modules along the sequence dimension.
-      extend(times, batchSize, hiddenSize, input.size(4), input.size(5))
-    }
+    val outputSize = input.size().clone()
```
JerryYanWan: input.size() will return a new array. It does not need to call clone again.
`val outputSize = input.size()`

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Cell.scala
dding3: not really, its size is different with different kind of images.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
dding3: Actually array used in the change is quite small, at most 24 bytes, I doubted if it will cause perf issue with minor gc as we can see lots of small arrays are created with tensor operations.

JerryYanWan:
LGTM

PR #1364, opened 2017-07-25
add SpatialWithinChannelLRN, update the mapping of caffe lrn

SeaOfOcean:

add SpatialWithinChannelLRN, update the mapping of caffe lrn

## How was this patch tested?

unit test


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialWithinChannelLRN.scala
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * The local response normalization layer performs a kind of “lateral inhibition”
+ * by normalizing over local input regions. the local regions extend spatially,
+ * in separate channels (i.e., they have shape 1 x local_size x local_size).
+ * @param size  the side length of the square region to sum over
+ * @param alpha the scaling parameter
+ * @param beta the exponent
+ * @tparam T The numeric type in the criterion, usually which are [[Float]] or [[Double]]
+ */
+class SpatialWithinChannelLRN[@specialized(Float, Double) T: ClassTag]
```
wzhongyuan: Remove specialized to avoid creating multiple class objects, since we use the factory method (where we have 'specialzied') to create instance, there is no type inference cost. Also this is required in module serialization :)

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/caffe/Converter.scala
```diff
@@ -274,6 +282,7 @@ abstract class Converter[T: ClassTag](implicit ev: TensorNumeric[T]) {
       case convolution : SpatialConvolution[_] => toCaffeConvolution(moduleNode, bottoms, nextSize)
       case relu : ReLU[_] => toCaffeRelu(moduleNode, bottoms, nextSize)
       case lrn : SpatialCrossMapLRN[_] => toCaffeLRN(moduleNode, bottoms, nextSize)
+      case lrn : SpatialWithinChannelLRN[_] => toCaffeLRN(moduleNode, bottoms, nextSize)
```
zhangxiaoli73: two case "lrn"

SeaOfOcean: with different lrn type

wzhongyuan:
LGTM

PR #1369, opened 2017-07-25
support get final state and cell

dding3:
support get final state and cell for convlstm

## How was this patch tested?
unit test


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -264,6 +265,16 @@ class Recurrent[T : ClassTag]()
     output
   }
 
+  def getFinalstateAndCell(): mutable.HashMap[String, Tensor[T]] = {
+    require(cells != null && cells(times - 1).output != null,
+      "getFinalstateAndCell need to be called after updateOutput")
+    val map = new mutable.HashMap[String, Tensor[T]]()
+    map("cell") = cells(times - 1).output.toTable(hidDim).asInstanceOf[Table]
```
JerryYanWan: For SimpleRNN, the cells(times-1).output.toTable(hidDim) is a Tensor. Thus, it will throw an error.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -264,6 +265,16 @@ class Recurrent[T : ClassTag]()
     output
   }
 
+  def getFinalstateAndCell(): mutable.HashMap[String, Tensor[T]] = {
```
JerryYanWan: `getFinalStateAndCell`

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -261,6 +262,17 @@ class Recurrent[T : ClassTag]()
     output
   }
 
+  def getFinalStateAndCellStatus(): (Tensor[T], Tensor[T]) = {
+    require(cells != null && cells(times - 1).output != null,
+      "getFinalStateAndCell need to be called after updateOutput")
+    val cell = cells(times - 1).output.toTable(hidDim).asInstanceOf[Activity]
+    val cellState = if (cell.isInstanceOf[Table]) cell.asInstanceOf[Table]
```
JerryYanWan: val cellState = if (cell.isTable) cell.toTable

PR #1372, opened 2017-07-26
Caffe prototxt only

wzhongyuan:

Sometimes Caffe model binary has different structure from prototxt and user want to create net based on prototxt, this change is to create net based on prototxt only and binary file is used to copy weight and bias

## How was this patch tested?

Unit test, tested locally with AlexNet and Inception


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/caffe/CaffeLoader.scala
```diff
@@ -115,6 +116,62 @@ class CaffeLoader[T: ClassTag](prototxtPath: String, modelPath: String,
...
+        } else {
+          weightLayer.asInstanceOf[LayerParameter].getBlobsList.asScala.foreach(
+            blob => v1LayerBuilder.addBlobs(blob)
+          )
+        }
+      }
+      builder.addLayers(v1LayerBuilder.build)
+    })
+
+    v2Layers.foreach(v2Layer => {
```
zhichao-li: could we somehow extract this as a method instead of duplicate the codes.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/caffe/CaffeLoader.scala
```diff
@@ -115,6 +116,62 @@ class CaffeLoader[T: ClassTag](prototxtPath: String, modelPath: String,
...
 
+  private def mergeNetWithParams(net : NetParameter, weights : NetParameter): NetParameter = {
+
+    val builder = NetParameter.newBuilder(net)
+    val layers = new mutable.HashMap[String, GeneratedMessage]
+    val v1Layers = new ArrayBuffer[V1LayerParameter]
+    val v2Layers = new ArrayBuffer[LayerParameter]
+
+    net.getLayersList.asScala.foreach(v1Layer => v1Layers.append(v1Layer))
+    net.getLayerList.asScala.foreach(v2Layer => v2Layers.append(v2Layer))
```
zhichao-li: how to differentiate v1 and v2?

zhichao-li: oh,got it. getLayersList vs getLayerList

wzhongyuan:
@zhichao-li   refinement as per your comment , please help check

SeaOfOcean:
LGTM

wzhongyuan:
pr passed
http://10.239.47.211:18888/job/BigDL-PR-Validation/1418/

PR #1378, opened 2017-07-26
fix caffe dilated, group convolution mapping, update customized layer

SeaOfOcean:

fix caffe dilated, group convolution mapping, update customized layer

## How was this patch tested?

unit test

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/caffe/CaffeLoader.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/caffe/CaffeLoader.scala
@@ -17,7 +17,6 @@ package com.intel.analytics.bigdl.utils.caffe
 
 import java.io._
 
-import scala.collection.JavaConverters._
 import caffe.Caffe
 import caffe.Caffe._
 import com.google.protobuf.{CodedInputStream, GeneratedMessage, TextFormat}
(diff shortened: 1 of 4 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/caffe/CaffeLoader.scala
```diff
@@ -26,14 +25,25 @@ import com.intel.analytics.bigdl.nn.Graph.ModuleNode
...
+import scala.collection.JavaConverters._
 import scala.collection.mutable
 import scala.collection.mutable.ArrayBuffer
 import scala.reflect.ClassTag
 
+abstract class Customizable[T: ClassTag](implicit ev: TensorNumeric[T]) {
+  var contexts: Map[String, Any] = _
+  def convertor(layer: GeneratedMessage): Seq[ModuleNode[T]]
+
+  def registorContext(name: String, context: Any): Unit = {
```
wzhongyuan: register

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/caffe/CaffeLoader.scala
```diff
@@ -45,44 +55,38 @@ import scala.reflect.ClassTag
...
-  private def registerCustomizedConverter() : Unit = {
+  private def registerCustomizedConverter(): Unit = {
     if (customizedConverters != null) {
       customizedConverters.foreach(entry => {
-        layerConverter.registerCutomizedConverter(entry._1, entry._2)
-        v1layerConverter.registerCutomizedConverter(entry._1, entry._2)
+        layerConverter.registerCutomizedConverter(entry._1, entry._2.convertor)
+        v1layerConverter.registerCutomizedConverter(entry._1, entry._2.convertor)
+        entry._2.registorContext("name2LayerV1", name2LayerV1)
+        entry._2.registorContext("name2LayerV2", name2LayerV2)
```
wzhongyuan: Not sure if needed, we can also add NetParameter as well as a generic case

wzhongyuan:
@SeaOfOcean  why there is no conflict with master after I merged my change in ?

wzhongyuan:
LGTM

Issue #589, opened 2017-03-14
support resume training in python


shane-huang:
Currently optimizer does not support resume training, because state cannot be initialized from snapshot file. Need to support this.

zhichao-li:
we should wrap the bellowing code and provide a resuming api:
val m = model.load
val o = optmMethod.load
new optimizer(m, o)
.optimize()

zhichao-li:
@zhangxiaoli73 should we close this?

PR #1387, opened 2017-07-27
add load optimmethod for python

zhangxiaoli73:

fix issue https://github.com/intel-analytics/BigDL/issues/589
add load optimmethod for python


pyspark/bigdl/optim/optimizer.py
```diff
@@ -313,6 +313,26 @@ def __init__(self,
         JavaValue.__init__(self, None, bigdl_type, monitor, factor, patience, mode, epsilon,
                            cooldown, min_lr)
 
+class OptimMethod(JavaValue):
```
zhichao-li: This should be the base class of SGD and anyother concrete OptimMethod.

pyspark/bigdl/optim/optimizer.py
```diff
@@ -313,6 +313,26 @@ def __init__(self,
...
+    def __init__(self, bigdl_type="float"):
+        self.bigdl_type = bigdl_type
+
+    def load(self, path):
```
zhichao-li: add static annotation to turn this be a static method.

pyspark/bigdl/optim/optimizer.py
```diff
@@ -313,6 +313,26 @@ def __init__(self,
...
+        """
+        load optim method
+        :param path: file path
+        """
+        return callBigDlFunc(self.bigdl_type, "createLoadOptimMethod", path)
+
+    def save(self, method, path, overWrite):
```
zhichao-li: remove parameter method: `method`, self should be the method.

pyspark/bigdl/optim/optimizer.py
```diff
@@ -313,6 +313,26 @@ def __init__(self,
...
+        """
+        save OptimMethod
+        :param method    method to be saved
+        :param path      path
+        :param overWrite whether to overwrite
+        """
+        return callBigDlFunc(self.bigdl_type, "createSaveOptimMethod", method, path, overWrite)
```
zhichao-li: Rename `createSaveOptimMethod` to `saveOptimMethod`

pyspark/bigdl/optim/optimizer.py
```diff
@@ -313,6 +313,26 @@ def __init__(self,
...
+class OptimMethod(JavaValue):
+    def __init__(self, bigdl_type="float"):
+        self.bigdl_type = bigdl_type
+
+    def load(self, path):
+        """
+        load optim method
+        :param path: file path
+        """
+        return callBigDlFunc(self.bigdl_type, "createLoadOptimMethod", path)
```
zhichao-li: `createLoadOptimMethod` to `loadOptimMethod`

pyspark/test/simple_integration_test.py
```diff
@@ -72,6 +72,39 @@ def test_load_model(self):
...
+        model.add(l1)
+
+        sgd = SGD(learningrate=0.01, learningrate_decay=0.0002, weightdecay=0.0,
+                  momentum=0.0, dampening=0.0, nesterov=False,
+                  leaningrate_schedule=Poly(0.5, int((data_len/batch_size)*epoch_num)))
+
+        tmp_path = tempfile.mktemp()
+        optim = OptimMethod()
+        optim.save(sgd, tmp_path, True)
+        optim_method = optim.load(tmp_path)
```
zhichao-li: should we check part of the re-loaded fields is same as the original one or not?

zhichao-li:
Seems like there's no `resume` api in Scala, maybe we need to add one for scala as well. @yiheng

zhangxiaoli73:
jekins pass

zhangxiaoli73:
@shane-huang do you think we need to add another resume api in bigdl ? Now, user can load model and optim-method to resume training.

zhichao-li:
LGTM

shane-huang:
I don't think we need a resume API.  Keras doesn't have one either.  It is good we allow OptimMethod and Model to be loaded seperately. Sometimes we only need to load the model without the OptimMethod.

shane-huang:
Please change the docs accordingly. https://bigdl-project.github.io/master/#APIdocs/Optimizers/OptimMethod/ 

Also in https://bigdl-project.github.io/master/#ProgrammingGuide/optimization/#resume-training
change ``` (Scala only, Python does not support this yet) ```
> Use Module.load (Scala) or Model.load(Python) to load the model snapshot into an model object, and OptimMethod.load (Scala only, Python does not support this yet) to load optimization method into an OptimMethod object.

zhangxiaoli73:
update md

Issue #1375, opened 2017-07-26
document optimizer usage difference between 0.1 and 0.2


shane-huang:
- Optimizer - Usage of setOptimMethod changed in python
- LBFGS: master constructor add a lot of parameters while branch-0.1 does not have any

Issue #1377, opened 2017-07-26
revise ml pipeline related docs in 0.1 


shane-huang:
- remove DLEstimator related docs. 
- change DLClassifier api, etc.

PR #1390, opened 2017-07-27
update optimmethod and metrics md for branch-0.1

zhangxiaoli73:
fix issue https://github.com/intel-analytics/BigDL/issues/1375
issue https://github.com/intel-analytics/BigDL/issues/1377
update optimmethod and metrics md for branch-0.1

### DIFF
--- a/docs/docs/APIdocs/Optimizers/Optim-Methods.md
+++ b/docs/docs/APIdocs/Optimizers/Optim-Methods.md
@@ -1,27 +1,11 @@
+Here, we just describe some optim methods. Method parameters(e.g."learningRate") and internal training parameters(e.g."epoch") store in Table state.
+If you want to set and save methods when training, you can refer to [OptimMethod](OptimMethod.md) for Details.
 ## Adam ##
 
 **Scala:**
 ```scala
-val optim = new Adam(learningRate=1e-3, learningRateDecay=0.0, beta1=0.9, beta2=0.999, Epsilon=1e-8)
+val optim = new Adam()
 ```
-**Python:**
-```python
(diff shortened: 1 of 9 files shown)


docs/docs/APIdocs/Optimizers/Optim-Methods.md
```diff
@@ -68,102 +50,66 @@ def rosenBrock(x: Tensor[Float]): (Float, Tensor[Float]) = {
...
 each iteration.
  
 **Scala example:**
 ```scala
-val optimMethod = new SGD[Float](learningRate= 1e-3,learningRateDecay=0.0,
-                               weightDecay=0.0,momentum=0.0,dampening=Double.MaxValue,
-                               nesterov=false,learningRateSchedule=Default(),
-                               learningRates=null,weightDecays=null)
-optimizer.setOptimMethod(optimMethod)
+optimizer.setOptimMethod(new SGD[Float]())
```
shane-huang: maybe we add a line about set state (learning rate, etc.)?

shane-huang:
please also change ProgrammingGuide/optimization.md #resume-training
```
After training stops, you can resume from any saved point. Choose one of the model snapshots and the corresponding optim-method snapshot to resume (saved in checkpoint path, details see Checkpointing). Use Module.load (Scala) or Model.load(Python) to load the model snapshot into an model object, and OptimMethod.load (Scala only, Python does not support this yet) to load optimization method into an OptimMethod object. Then create a new Optimizer with the loaded model and optim method. Call Optimizer.optimize, and you will resume from the point where the snapshot is taken. Refer to OptimMethod Load and Model Load for details.

You can also resume training without loading the optim method, if you intend to change the learning rate schedule or even the optimization algorithm. Just create an Optimizer with loaded model and a new instance of OptimMethod (both Scala and Python).
```

PR #1394, opened 2017-07-27
[backport0.2] Add fix to create network based on prototxt only

wzhongyuan:

Sometimes Caffe model binary has different structure from prototxt and user want to create net based on prototxt, this change is to create net based on prototxt only and binary file is used to copy weight and bias

## How was this patch tested?

Unit tests, test locally with common models


wzhongyuan:
PR passed http://10.239.47.211:18888/job/BigDL-PR-Validation/1434/

PR #1413, opened 2017-08-01
fix load caffe v1layer bug, load caffe layers only once in dynamic mode

SeaOfOcean:

fix load caffe v1layer bug, load caffe layers only once in dynamic mode

## How was this patch tested?

unit test

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/caffe/CaffeLoader.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/caffe/CaffeLoader.scala
@@ -81,17 +81,15 @@ class CaffeLoader[T: ClassTag](prototxtPath: String, modelPath: String,
     }
   }
 
-  private var isCaffeLoaded = false
-
   private def loadCaffe(prototxtPath: String, modelPath: String): Unit = {
-    if (isCaffeLoaded) return
-    netparam = loadBinary(prototxtPath, modelPath)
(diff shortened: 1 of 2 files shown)


wzhongyuan:
I am a little confused here, if you can loadCaffeModel() from the companion object, each time a new instance of CaffeLoader will be created. is it going to be a problem ? what's the scenario that duplicated loading will happen ?

SeaOfOcean:
createCaffeModel will loadCaffe, it will also call copyParameters (loadCaffe again)

wzhongyuan:
I see, Not sure why but I just checked previous check-ins, there is a checking to avoid double loading

  private def loadCaffe(prototxtPath: String, modelPath: String): Unit = {
    if (name2LayerV2 == null) {
      netparam = loadBinary(prototxtPath, modelPath)
      name2LayerV2 = Map[String, LayerParameter]()
      name2LayerV1 = Map[String, V1LayerParameter]()
      import scala.collection.JavaConverters._
      // V1LayerParameter
      netparam.getLayersList.asScala.foreach(layer => name2LayerV1 += (layer.getName -> layer))
      // V2LayerParameter
      netparam.getLayerList.asScala.foreach(layer => name2LayerV2 += (layer.getName -> layer))

    }
  }

Can we do similarly to avoid declaring a new variable ? Maybe wen can check netparam to see if it's not null ?

SeaOfOcean:
pr validate pass

wzhongyuan:
LGTM

Issue #413, opened 2017-01-24
DistriOptimizier's counting is wrong.


qiuxin2012:
```
2017-01-24 14:13:05 INFO  DistriOptimizer$:241 - [Epoch 10 810000/1000000][Iteration 118][Wall Clock 54.905153972s] Train 90000 in 0.463609974seconds. Throughput is 194128.69663584934 records/second. Loss is 0.25111403044428254. 
2017-01-24 14:13:06 INFO  DistriOptimizer$:241 - [Epoch 10 900000/1000000][Iteration 119][Wall Clock 55.368763946s] Train 90000 in 0.457162282seconds. Throughput is 196866.6347675638 records/second. Loss is 0.2511395179269162. 
2017-01-24 14:13:06 INFO  DistriOptimizer$:241 - [Epoch 10 990000/1000000][Iteration 120][Wall Clock 55.825926228s] Train 90000 in 0.393377465seconds. Throughput is 228787.8895147184 records/second. Loss is 0.25113855703782745. 
2017-01-24 14:13:06 INFO  DistriOptimizer$:285 - [Epoch 10 990000/1000000][Iteration 120][Wall Clock 55.825926228s] Epoch finished. Wall clock time is 56232.849532ms
```
The last line should be 
`2017-01-24 14:13:06 INFO  DistriOptimizer$:285 - [Epoch 10 1080000/1000000][Iteration 120][Wall Clock 55.825926228s] Epoch finished. Wall clock time is 56232.849532ms`

Issue #1354, opened 2017-07-21
Epoch stopping criterion is non-sensical


sethah:
In the distributed optimizer, an epoch ends based on the following line:

````scala
if (accumulateCount >= dataset.size()) {
  ...
}
````

This does not make sense to me. `accumulateCount` is a var that is reset every epoch, and is updated via `accumulateCount += recordsNum.value`. Here, `recordsNum` is an accumulator which keeps track of the number of **training samples** that were used to train in each _iteration_. So, the stopping criterion above is comparing something with units of samples, to something with units of batches. 

Take a case where you have a batch size of 256, and your distributed dataset has 1000 minibatches in total. In this case, each epoch will consist of training on only four minibatches out of 1000, since `256 * 4 > 1000`. However, most of the time an epoch is defined as making a single pass through all of the mini-batches. 

One problem is that there doesn't seem to be any documentation on what the intended behavior is, but I'm fairly confident this is not it.

sethah:
BTW, I have a pr pretty much ready for this. I'm happy to submit it, but wanted to raise the issue to make sure of what the intended behavior really is.

dding3:
It's correct that accumulateCount is used to keep track of the number of trained samples, however dataset.size() is actually the total number of samples, so the stopping criterion is comparing units of trained samples with total samples. It should be correct to mark an epoch as finished if accumulateCount >= dataset.size as all the samples has been made a pass.

sethah:
In the `DistriOptimizer`, the dataset is type ` DistributedDataSet[MiniBatch[T]]`. `dataset.size()` calls `dataset.count`, defined as follows

````scala
buffer.mapPartitions(iter => {
    require(iter.hasNext)
    val array = iter.next()
    require(!iter.hasNext)
    Iterator.single(array.length)
  }).reduce(_ + _)
````

`buffer` here, is of type `RDD[Array[Minibatch[T]]]`, so `iter.next()` will produce `Array[Minibatch[T]]`. The length of that array will be the number of minibatches in that partition, and the sum of that across partitions will be the total number of minibatches in the dataset, not the total number of samples contained in all those minibatches.

Please also see `DistriOptimizerSpec` for further evidence, where the dataset used for tests defines the size method to be `256 * nodeNumber` which is the number of _minibatches_ in the dataset. There are 1024 minibatches in that dataset, but if you run the tests you will see that it only processes 32 minibatches per epoch, instead of 1024. It prints 

````
[...]

PR #1417, opened 2017-08-01
Correction to Epoch stopping criterion in DistriOptimizer + cleanups

sethah:

This patch relates to #1354. Basically the optimizer was comparing the number of records processed in an epoch to `dataset.size()` to determine if an epoch was finished. This is only valid in certain cases, as `dataset.size()` has inconsistent semantics. Instead, this patch proposes to compute the number of samples in the dataset as 

````scala
val numSamples = dataset.data(train = false).map(_.size()).reduce(_ + _)
````
Now, the epoch stopping criterion always compares variables with units of "samples."

## How was this patch tested?

Unit tests were added to verify that datasets created by constructing minibatches manually as well as datasets created using the `Transformer` API both have the expected number of iterations per epoch.

## Other notes

The other unit tests in the `DistriOptimizerSpec` had specified number of epochs to be > 1 before this patch, but there were far fewer iterations in each epoch due to this error. Because of that, I adjusted most of them down to a single epoch so that the tests don't take an unreasonable amount of time. 

#1354

### CODE REVIEW

**Comment by sethah** [comment]
cc @qiuxin2012 @dding3

File: spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala [hunk 1]
```diff
@@ -277,31 +287,33 @@ object DistriOptimizer {
...
         optimMethod.state.update("Loss", driverState[Float]("Loss"))
         if (validationMethods.isDefined) {
           optimMethod.state.update("score", driverState[Float]("score"))
         }
         optimMethod.updateHyperParameter()
         driverState("Throughput") = recordsNum.value.toFloat / ((end - start) / 1e9f)
         driverState("LearningRate") = -optimMethod.getLearningRate().toFloat
-        logger.info(s"${_header} Train ${recordsNum.value} in ${(end - start) / 1e9}seconds. " +
-          s"Throughput is ${driverState("Throughput")} records/second. Loss is ${
+        val _header = header(driverState[Int]("epoch"), recordsProcessedThisEpoch, numSamples,
```
sethah: I moved this down because of what I see as an error in the logging before. When an epoch was finished you might see the logger print out something like:

````
2017-08-01 14:58:03 INFO  DistriOptimizer$:307 - [Epoch 1 8160/8160][Iteration 256][Wall Clock 6.607695635s] Trained 32 records in 0.029281884 seconds. Throughput is 1092.8258 records/second. Loss is 1.6109945E-4.
2017-08-01 14:58:03 INFO  DistriOptimizer$:351 - [Epoch 1 8192/8192][Iteration 256][Wall Clock 6.607695635s] Epoch finished. Wall clock time is 6691.791083 ms
````

The first number was, incorrectly IMO, the number of samples processed _before_ this iteration, not the number of samples that had been currently processed. That also has the consequence of the final message, which says "Epoch finished" indicating that it did not process all of the records e.g. only 8160 out of 8192. After moving this here, it will read correctly.

**Comment by qiuxin2012** [comment]
Great, thanks for your PR.

File: spark/dl/src/test/scala/com/intel/analytics/bigdl/optim/DistriOptimizerSpec.scala [hunk 2]
```diff
@@ -136,6 +133,52 @@ class DistriOptimizerSpec extends FlatSpec with Matchers with BeforeAndAfter {
     }
   }
 
+  "DistriOptimizer" should "train all minibatches per epoch" in {
+    val numSamples = 64
+    val numClasses = 3
+    val height = 32
+    val width = 32
+    val images = Array.tabulate(64) { i =>
```
qiuxin2012: I think this Unit Test won't cause the bug described in #1417 . You should create a `val miniBatches: Array[MiniBatch]` here, and create DataSet using `val dataset = DataSet.rdd(miniBatches, sc)`.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
qiuxin2012: Seems also fixed #413 .

spark/dl/src/test/scala/com/intel/analytics/bigdl/optim/DistriOptimizerSpec.scala
```diff
@@ -157,7 +200,7 @@ class DistriOptimizerSpec extends FlatSpec with Matchers with BeforeAndAfter {
     mm.getParameters()._1.fill(0.125)
     val optimizer = new DistriOptimizer[Double](mm, dataSet, new MSECriterion[Double]())
       .setState(T("learningRate" -> 20.0))
-      .setEndWhen(Trigger.maxEpoch(5))
```
qiuxin2012: @yiheng I'm not sure there. @sethah changed `maxEpoch` from `5` to `1`. I think this `maxEpoch` should be 2 at least. But why you set a `5`?

qiuxin2012:
Get an failure from jenkins, please fix this issue:
```
com.intel.analytics.bigdl.utils.LoggerFilterSpec.A LoggerFilter should output correct info on console and bigdl.log

Error Message

requirement failed: output can't matchs the specific output
Stacktrace

      java.lang.IllegalArgumentException: requirement failed: output can't matchs the specific output
      at scala.Predef$.require(Predef.scala:233)
      at com.intel.analytics.bigdl.utils.LoggerFilterSpec$$anonfun$2.apply$mcV$sp(LoggerFilterSpec.scala:130)
      at com.intel.analytics.bigdl.utils.LoggerFilterSpec$$anonfun$2.apply(LoggerFilterSpec.scala:60)
      at com.intel.analytics.bigdl.utils.LoggerFilterSpec$$anonfun$2.apply(LoggerFilterSpec.scala:60)
      at org.scalatest.Transformer$$anonfun$apply$1.apply$mcV$sp(Transformer.scala:22)
      at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
      at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
      at org.scalatest.Transformer.apply(Transformer.scala:22)
      at org.scalatest.Transformer.apply(Transformer.scala:20)
      at org.scalatest.FlatSpecLike$$anon$1.apply(FlatSpecLike.scala:1647)
      at org.scalatest.Suite$class.withFixture(Suite.scala:1122)
      at org.scalatest.FlatSpec.withFixture(FlatSpec.scala:1683)
      at org.scalatest.FlatSpecLike$class.invokeWithFixture$1(FlatSpecLike.scala:1644)
      at org.scalatest.FlatSpecLike$$anonfun$runTest$1.apply(FlatSpecLike.scala:1656)
      at org.scalatest.FlatSpecLike$$anonfun$runTest$1.apply(FlatSpecLike.scala:1656)
      at org.scalatest.SuperEngine.runTestImpl(Engine.scala:306)
      at org.scalatest.FlatSpecLike$class.runTest(FlatSpecLike.scala:1656)
      at com.intel.analytics.bigdl.utils.LoggerFilterSpec.org$scalatest$BeforeAndAfter$$super$runTest(LoggerFilterSpec.scala:37)
      at org.scalatest.BeforeAndAfter$class.runTest(BeforeAndAfter.scala:200)
      at com.intel.analytics.bigdl.utils.LoggerFilterSpec.runTest(LoggerFilterSpec.scala:37)
      at org.scalatest.FlatSpecLike$$anonfun$runTests$1.apply(FlatSpecLike.scala:1714)
      at org.scalatest.FlatSpecLike$$anonfun$runTests$1.apply(FlatSpecLike.scala:1714)
      at org.scalatest.SuperEngine$$anonfun$traverseSubNodes$1$1.apply(Engine.scala:413)
      at org.scalatest.SuperEngine$$anonfun$traverseSubNodes$1$1.apply(Engine.scala:401)
      at scala.collection.immutable.List.foreach(List.scala:318)
      at org.scalatest.SuperEngine.traverseSubNodes$1(Engine.scala:401)
      at org.scalatest.SuperEngine.org$scalatest$SuperEngine$$runTestsInBranch(Engine.scala:390)
      at org.scalatest.SuperEngine$$anonfun$traverseSubNodes$1$1.apply(Engine.scala:427)
      at org.scalatest.SuperEngine$$anonfun$traverseSubNodes$1$1.apply(Engine.scala:401)
      at scala.collection.immutable.List.foreach(List.scala:318)
      at org.scalatest.SuperEngine.traverseSubNodes$1(Engine.scala:401)
      at org.scalatest.SuperEngine.org$scalatest$SuperEngine$$runTestsInBranch(Engine.scala:396)
      at org.scalatest.SuperEngine.runTestsImpl(Engine.scala:483)
      at org.scalatest.FlatSpecLike$class.runTests(FlatSpecLike.scala:1714)
      at org.scalatest.FlatSpec.runTests(FlatSpec.scala:1683)
      at org.scalatest.Suite$class.run(Suite.scala:1424)
      at org.scalatest.FlatSpec.org$scalatest$FlatSpecLike$$super$run(FlatSpec.scala:1683)
      at org.scalatest.FlatSpecLike$$anonfun$run$1.apply(FlatSpecLike.scala:1760)
      at org.scalatest.FlatSpecLike$$anonfun$run$1.apply(FlatSpecLike.scala:1760)
      at org.scalatest.SuperEngine.runImpl(Engine.scala:545)
      at org.scalatest.FlatSpecLike$class.run(FlatSpecLike.scala:1760)
      at com.intel.analytics.bigdl.utils.LoggerFilterSpec.org$scalatest$BeforeAndAfter$$super$run(LoggerFilterSpec.scala:37)
      at org.scalatest.BeforeAndAfter$class.run(BeforeAndAfter.scala:241)
      at com.intel.analytics.bigdl.utils.LoggerFilterSpec.run(LoggerFilterSpec.scala:37)
      at org.scalatest.Suite$class.callExecuteOnSuite$1(Suite.scala:1492)
      at org.scalatest.Suite$$anonfun$runNestedSuites$1.apply(Suite.scala:1528)
      at org.scalatest.Suite$$anonfun$runNestedSuites$1.apply(Suite.scala:1526)
      at scala.collection.IndexedSeqOptimized$class.foreach(IndexedSeqOptimized.scala:33)
      at scala.collection.mutable.ArrayOps$ofRef.foreach(ArrayOps.scala:108)
      at org.scalatest.Suite$class.runNestedSuites(Suite.scala:1526)
      at org.scalatest.tools.DiscoverySuite.runNestedSuites(DiscoverySuite.scala:29)
      at org.scalatest.Suite$class.run(Suite.scala:1421)
      at org.scalatest.tools.DiscoverySuite.run(DiscoverySuite.scala:29)
      at org.scalatest.tools.SuiteRunner.run(SuiteRunner.scala:55)
      at org.scalatest.tools.Runner$$anonfun$doRunRunRunDaDoRunRun$3.apply(Runner.scala:2563)
      at org.scalatest.tools.Runner$$anonfun$doRunRunRunDaDoRunRun$3.apply(Runner.scala:2557)
      at scala.collection.immutable.List.foreach(List.scala:318)
      at org.scalatest.tools.Runner$.doRunRunRunDaDoRunRun(Runner.scala:2557)
      at org.scalatest.tools.Runner$$anonfun$runOptionallyWithPassFailReporter$2.apply(Runner.scala:1044)
      at org.scalatest.tools.Runner$$anonfun$runOptionallyWithPassFailReporter$2.apply(Runner.scala:1043)
      at org.scalatest.tools.Runner$.withClassLoaderAndDispatchReporter(Runner.scala:2722)
      at org.scalatest.tools.Runner$.runOptionallyWithPassFailReporter(Runner.scala:1043)
      at org.scalatest.tools.Runner$.main(Runner.scala:860)
      at org.scalatest.tools.Runner.main(Runner.scala)
```

spark/dl/src/test/scala/com/intel/analytics/bigdl/optim/DistriOptimizerSpec.scala
sethah: It does. You can verify it by changing the line `if (recordsProcessedThisEpoch >= numSamples)` in DistriOptimizer to `if (recordsProcessedThisEpoch >= dataset.size())` as it was before, and the test will fail. That's also why I added the lines

````scala
assert(sampleDataSet.size() == numSamples)
assert(batchDataSet.size() == numBatches)
````

To make it clear that dataset.size() is different for each.

**Comment by sethah** [comment]
@qiuxin2012 I think I've fixed the failing test. Though, I got some sporadic failed tests in the same suite on my local machine, but the failures weren't related to this PR. It may be a fragile test suite. Let's see what Jenkins says.

File: spark/dl/src/test/scala/com/intel/analytics/bigdl/optim/DistriOptimizerSpec.scala (hunk 3 above)
Comment (sethah) [inline spark/dl/src/test/scala/com/intel/analytics/bigdl/optim/DistriOptimizerSpec.scala]: 5 epochs before was still less iterations (160) than a single epoch now (256). So, as long as the test passes, I don't think this is concerning.

File: spark/dl/src/test/scala/com/intel/analytics/bigdl/optim/DistriOptimizerSpec.scala (hunk 2 above)
Comment (qiuxin2012) [inline spark/dl/src/test/scala/com/intel/analytics/bigdl/optim/DistriOptimizerSpec.scala]: Looks good. Thanks.

**Comment by qiuxin2012** [comment]
@sethah Jenkins passed.
Could you also apply this fix to [LocalOptimizer](https://github.com/intel-analytics/BigDL/blob/master/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/LocalOptimizer.scala)?

File: spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala [hunk 4]
```diff
@@ -103,20 +103,23 @@ object DistriOptimizer {
...
-    val _subModelNumber = Engine.getEngineType match {
+      "score" -> optimMethod.state.get[Float]("score").getOrElse(0f)
+    )
+    val _subModelNumber = Engine.getEngineType() match {
       case MklBlas => coresPerNode
-      case _ => throw new IllegalArgumentException()
     }
-    var accumulateCount = 0
+
+    val numSamples = dataset.data(train = false).map(_.size()).reduce(_ + _)
```
qiuxin2012: I just test the time cost for this counting on Imagenet DataSet on 8 E5-haswell nodes.
```
2017-08-03 09:41:03 INFO  DistriOptimizer$:121 - Counting samples...
2017-08-03 09:41:36 INFO  DistriOptimizer$:125 - Total: 1281167. Takes 32.803411577s
```
32s seems not long compared with the looading time and training time.
Could you add a log info for this counting?

sethah:
@qiuxin2012 I added this to the local optimizer, but it seems that the local optimizer does not record a training summary, so the test fails. Do you know of a better way to get the number of iterations from the optimizer?

spark/dl/src/test/scala/com/intel/analytics/bigdl/optim/LocalOptimizerSpec.scala
```diff
@@ -128,6 +128,52 @@ class LocalOptimizerSpec extends FlatSpec with Matchers with BeforeAndAfter{
...
+    Seq(sampleDataSet, batchDataSet).foreach { dataset =>
+      RandomGenerator.RNG.setSeed(10)
+      val maxEpochs = 2
+      val logdir = com.google.common.io.Files.createTempDir()
+      val trainSummary = TrainSummary(logdir.getPath, "minibatch-test")
+      val optimizer = new LocalOptimizer(
+        nn,
+        dataset,
+        ClassNLLCriterion[Float]())
+        .setOptimMethod(new LBFGS)
```
qiuxin2012: You can get the length from `SGD.state`.
```
val sgd = new SGD[Float]()
sgd.state[Option[Int]]("neval").get
```

qiuxin2012:
@sethah I just provide a tricky method for you. No TrainSummary in LocalOptimizer is another issue, would like to propose another PR to fix it?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -103,23 +103,30 @@ object DistriOptimizer {
...
-      case _ => throw new IllegalArgumentException()
     }
-    var accumulateCount = 0
+
+    logger.info("Count dataset")
+    val countBefore = System.nanoTime()
+    val numSamples = dataset.data(train = false).map(_.size()).reduce(_ + _)
+    val countAfter = System.nanoTime()
+    logger.info(s"Count dataset complete. Time elapsed: ${(countAfter - countBefore) / 1e9}s")
+
```
qiuxin2012: As my discussion with @yiheng, we think we should leave a warning here when `numSamples != dataset.size()`. Because
```
If the user is caching RDD[MiniBatch] directly, data in MiniBatches is fixed in each partitions, we just choose one MiniBatch from the RDD[MiniBatch]'s partition randomly. This may leads the model couldn't converge to the best possible model.
If user is caching RDD[Sample], in each iteration, we randomly transform a few sample to a MiniBatch. So data in every MiniBatch is random, this random data is better.
```

spark/dl/src/test/scala/com/intel/analytics/bigdl/optim/DistriOptimizerSpec.scala
yiheng: It looks good to me.

yiheng:
@dding3 Would you like review this PR?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -374,8 +392,10 @@ object DistriOptimizer {
         )
 
       } else {
-        logger.info(s"Warning!!! Ignore this iteration as more than maxDropPercentage " +
-          s"module is dropped!! Finished modules number: ${finishedModelNum}")
+        logger.info(s"Warning! Not enough training samples were successfully processed in this " +
+          s"iteration due to some slow tasks. The gradients computed in this iteration will be " +
+          s"discarded. Only $numFinishedModelUpdates/$driverSubModelNum threads successfully " +
```
dding3: Only {($numFinishedModelUpdates/$driverSubModelNum) * 100}% percentage threads?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -252,20 +258,28 @@ object DistriOptimizer {
...
           Iterator.single(finishedThreads.size)
-        }).reduce(_ + _)
-
-      dropModelNumBatch += (driverSubModelNum - finishedModelNum)
-      if (dropPercentage == 0 || finishedModelNum >= driverSubModelNum * (1-maxDropPercentage)) {
-        val value = lossSum.value / finishedModelNum
-        models.mapPartitions(modelIter => {
+        }
+      }
+      val numFinishedModelUpdates = finishedThreads.reduce(_ + _)
```
dding3: Update code like val numFinishedModel = dataRDD.zipPartitoins(xxxxx).reduce(_ + _), similar to original code? Currently looks like there are two finishedThread variables which have different meaning and different type, which kind of cause code less clear.

yiheng:
ping @sethah Any update of the PR?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
sethah: Personally I think showing both numbers is more informative. I can change it, though, if you feel strongly?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
sethah: Ok, I added a comment to make the meaning of this variable more clear.

sethah:
My apologies all for the delay - got busy with other things. Thanks for the review! I think my last commit addresses the issues raised.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
dding3: OK. I understand. It should be enough with these two numbers.

dding3:
Updated code looks good to me.

qiuxin2012:
@sethah 
We find LocalOptimizerSpec will run into a dead loop at line
```
val numSamples = dataset.data(train = false).map(_.size()).reduce(_ + _)
```
After some debugging, we find it's caused by line 88 in LocalOptimizerSpec
```
override def hasNext: Boolean = true
```
Please change this line to 
```
override def hasNext: Boolean = train || i < totalSize
```
to fix this issue.

qiuxin2012:
Jenkins find another bug in DLEstimatorSpec, I have fixed it for you, just copy from https://github.com/intel-analytics/BigDL/pull/1505/commits/42eb118c0fda8df7f7b4b169f05dd511a01b8ad6
@sethah

qiuxin2012:
Please fix the two issues above, I think we can merge this pr tomorrow. @sethah

qiuxin2012:
Great!! Jenkins passed

PR #1422, opened 2017-08-02
fix stride array index mismatch in Conv2D while loading NCHW tensorflow model

frankfzw:

1. fix stride array index mismatch in Conv2D while loading NCHW tensorflow model
2. add corresponding unit test

## How was this patch tested?

unit test


yiheng:
LGTM

PR #1424, opened 2017-08-02
optimize logsoftmax

JerryYanWan:
Optimize LogSoftmax using MKL.

## How was this patch tested?
Unit Tests

For input = Tensor(20, 10000) and gradOutput = Tensor(20, 10000)

running on local desktop:
average elapsed time:

before = 0.029517 seconds
after = 8.99 * 1e-4 seconds

Boosting *32.8 times.
 

Test rnn model using spark-local with 4 cores
batchSize = 128
throughput is:
before:
174.9 records/s
after:
512.4 records/s

Tests on Broadwell with 44 cores.
one LSTM layer, throughput

outputSize before after
1000 	377.679 408.977
5000	346.314	331.818
10000	316.691	340.478
15000	306.378 338.387
20000	252.381 276.823

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/LogSoftMax.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/LogSoftMax.scala
@@ -82,63 +82,31 @@ class LogSoftMax[T: ClassTag](
     if (buffer2 == null || buffer2.length < in.nElement) {
       buffer2 = new Array[T](in.nElement)
     }
-    ev.getType match {
-      case FloatType =>
 
-        MKL.vsExp(in.nElement,
-          in.storage.asInstanceOf[Storage[Float]].array,
-          in.storageOffset - 1,
-          buffer2.asInstanceOf[Array[Float]],
-          0)
+    ev.vExp(in.nElement,
+      in.storage.array,
+      in.storageOffset - 1,
+      buffer2,
+      0)
 
-        val dot = MKL.vsdot(in.nElement,
-          buffer2.asInstanceOf[Array[Float]],
-          0,
-          1,
-          buffer1.asInstanceOf[Array[Float]],
-          0,
-          1)
-        val sum = ev.negative(
-          ev.log(ev.fromType[Float](dot)))
+    val dot = ev.dot(in.nElement,
+      buffer2,
+      0,
+      1,
+      buffer1,
+      0,
+      1)
 
-        MKL.vsaxpy(in.nElement,
-          ev.toType[Float](sum),
-          buffer1.asInstanceOf[Array[Float]],
-          0,
-          1,
-          out.storage.asInstanceOf[Storage[Float]].array,
-          out.storageOffset - 1,
-          1)
+    val sum = ev.negative(ev.log(dot))
 
-      case DoubleType =>
-
-        MKL.vdExp(in.nElement,
-          in.storage.asInstanceOf[Storage[Double]].array,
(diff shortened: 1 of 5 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/LogSoftMax.scala
```diff
@@ -68,63 +76,158 @@ class LogSoftMax[T: ClassTag](
...
-    Engine.model.sync(results)
-
     gradInput
   }
+
+  private def updateGradInputFrame(out: Tensor[T], gradOut: Tensor[T]): Unit = {
+    ev.getType match {
+       case FloatType =>
+
+         MKL.vsExp(out.nElement,
```
qiuxin2012: why not use ev.vExp

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/LogSoftMax.scala
```diff
@@ -39,11 +42,16 @@ class LogSoftMax[T: ClassTag](
   implicit ev: TensorNumeric[T]) extends TensorModule[T] {
   @transient
   private var results: Array[Future[Unit]] = null
+  @transient
+  private var buffer1: Array[T] = null
```
qiuxin2012: also clear this in clearState()

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/LogSoftMax.scala
```diff
@@ -39,11 +42,16 @@ class LogSoftMax[T: ClassTag](
   implicit ev: TensorNumeric[T]) extends TensorModule[T] {
   @transient
   private var results: Array[Future[Unit]] = null
```
qiuxin2012: As mkl is enabled, you can remove this multi thread.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/LogSoftMax.scala
```diff
@@ -68,63 +76,158 @@ class LogSoftMax[T: ClassTag](
...
+    if (buffer1 == null || buffer1.length < in.nElement) {
+      buffer1 = Array.fill(in.nElement)(ev.fromType[Int](1))
+    }
+    if (buffer2 == null || buffer2.length < in.nElement) {
+      buffer2 = new Array[T](in.nElement)
+    }
+    ev.getType match {
+      case FloatType =>
+
+        MKL.vsExp(in.nElement,
```
qiuxin2012: why not ev.vExp

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/LogSoftMax.scala
```diff
@@ -68,69 +76,107 @@ class LogSoftMax[T: ClassTag](
...
-    in.apply1(v => {
-      logsum = ev.plus(logsum, ev.exp(ev.minus(v, maxInput))); v
-    })
-    logsum = ev.plus(maxInput, ev.log(logsum))
-
-    out.map(in, (outData, inData) => {
-      ev.minus(inData, logsum)
-    })
+    if (buffer1 == null || buffer1.length < in.nElement) {
+      buffer1 = Array.fill(in.nElement)(ev.fromType[Int](1))
```
qiuxin2012: rename `ones`

PR #1431, opened 2017-08-07
SimpleRNN to Graph

JerryYanWan:
SimpleRNN to Graph

## How was this patch tested?
Unit Tests (RecurrentSpec)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RNN.scala
```diff
@@ -53,12 +54,9 @@ class RnnCell[T : ClassTag] (
   (implicit ev: TensorNumeric[T])
   extends Cell[T](Array(hiddenSize)) {
 
-  val parallelTable = ParallelTable[T]()
   val i2h = Identity[T]()
   val h2h = Linear[T](hiddenSize, hiddenSize,
```
psyyz10: Remove them, and move to the buildGraph to keep consistency?

psyyz10:
Do we need to keep both version of with graph and with sequential?

PR #1433, opened 2017-08-07
optimize Adam and Adamax optimizer

JerryYanWan:
Replace element-wise operations with MKL vector operations.
Add buffer for sharing memory

## How was this patch tested?
Unit Tests
JVM, single-threading

Adam:
Build a model with 5 linear layers. (500 * 500)
Total parameters(1million)

Test the average eta time in the Adam:

previous:
7.702 ms / iteration
after:
5.918 ms / iteration

Adamax:
Build a model with 5 linear layers. (500 * 500)
Total parameters(1million)

previous:
17.353 ms / iteration
after:
16.6666 ms / iteration

There is a max operation in Adamax which takes 1/4 of the elapsed time.
Currently, using (x + |x|) / 2 will give a worse performance w.r.t if-else branching method.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Adam.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Adam.scala
@@ -17,7 +17,7 @@
 package com.intel.analytics.bigdl.optim
 
 import breeze.linalg.*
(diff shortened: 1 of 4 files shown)


jason-dai:
is the test single-threading or multi-threading?

JerryYanWan:
single-threading

PR #1439, opened 2017-08-10
Set hiddenstate

dding3:
support set/get state in recurrent
add related ut

## How was this patch tested?
ut

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
@@ -267,7 +267,7 @@ class Recurrent[T : ClassTag]()
 
   def getState(): Activity = {
     require(cells != null && cells(times - 1).output != null,
-      "getHiddenState need to be called after updateOutput")
+      "getState need to be called after updateOutput")
     cells(times - 1).output.toTable(hidDim)
   }
 
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
@@ -244,8 +244,10 @@ class Recurrent[T : ClassTag]()
      * identical elements T(output, output). One of the elements from the cell output is
      * the updated hidden. Thus the currentInput will update its hidden element with this output.
      */
-    currentInput(hidDim) = hidden
     var i = 1
+    // init state
+    currentInput(hidDim) = if (initState != null) initState
+     else hidden
     while (i <= times) {
       currentInput(inputDim) = outputCell.select(timeDim, i)
       cells(i - 1).updateOutput(currentInput)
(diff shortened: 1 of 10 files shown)


jason-dai:
How is this used in real world applications?

dding3:
It's used in sequence to sequence learning. initial states of decoder network with last state of the encoding network.  In which encoder compresses the whole input sequence into a hidden state and decoder unfolds this hidden state to give the prediction. This PR provide basic set/get state api and in another pr will give a new layer to define this behavior.

jason-dai:
But how do you use it inside BigDL training?

dding3:
I planned to define a new layer(eg. seq2seq) which includes an encoder and decoder. get state of encoder by rec.get_state and inital decoder state with rec2.set_state(state) in forward function. Add seq2seq to model, with optimize logic which will continually call model.forward, I think we should be able to get what we want during training.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -244,15 +244,18 @@ class Recurrent[T : ClassTag]()
...
+    currentInput(hidDim) = if (initState != null) initState
+     else hidden
     while (i <= times) {
       currentInput(inputDim) = outputCell.select(timeDim, i)
       cells(i - 1).updateOutput(currentInput)
       currentInput(hidDim) = cells(i - 1).output.toTable(hidDim)
       i += 1
     }
 
+    initState = null
```
JerryYanWan: If we use the following code:

```
model.forward()
model.setState()
model.forward()
model.forward()
```
Then only the second forward will use initState. 
How to handle the multi layer cases?
If there are two LSTM, then the first LSTM layer should use 0 hidden state and the second LSTM layer should use the last hidden output from the previous layer?

spark/dl/src/test/scala/com/intel/analytics/bigdl/nn/ConvLSTMPeepholeSpec.scala
```diff
@@ -1020,17 +1022,56 @@ class ConvLSTMPeepholeSpec extends FlatSpec with BeforeAndAfter with Matchers {
...
+      .add(rec
+        .add(ConvLSTMPeephole[Double](inputSize, hiddenSize, 3, 3, 1, withPeephole = false)))
+
+    RNG.setSeed(890)
+    val rec2 = Recurrent[Double]()
+    val model2 = Sequential[Double]()
+      .add(rec2
+        .add(ConvLSTMPeephole[Double](inputSize, hiddenSize, 3, 3, 1, withPeephole = false)))
+
+    val output = model.forward(input).asInstanceOf[Tensor[Double]]
```
JerryYanWan: If call` model.forward` multiple times? Will it perform differently?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
dding3: As synced offline, It's expected behavior, if want to use state from previous layer, we need explicitly call setState(). I am not very like to propogate state in Recurrent, the code will not easy to maintain and less readable. Leave propogate state in up level layer like other framework do.

JerryYanWan:
LGTM

PR #1442, opened 2017-08-10
fix: TorchFile should use contains for checking the key

i8run:

## How was this patch tested?

jenkins


yiheng:
Have you tested the PR? One test case failed

http://172.168.2.101:8080/job/BigDL-PR-Validation-2.1-serial/1862/

PR #1451, opened 2017-08-14
fix copyStatus method in TimeDistributed layer

JerryYanWan:
Fix bug in copyStatus method in TimeDistributed layer

TimeDistributed is a recurrent layer and including one contained layer as its class variable.
Thus copyStatus should be applied to its contained layer instead of itself.

## How was this patch tested?
Unit Test


zhangxiaoli73:
LGTM

Issue #608, opened 2017-03-17
Use other format to save Model instead of Java serialization


zhichao-li:
Use other format to save Model instead of Java serialization, otherwise the serialVersionUID  would restrict the model loading. 
i.e If there's new function added to the DenseTensor:

: java.io.InvalidClassException: com.intel.analytics.bigdl.tensor.DenseTensor; local class incompatible: stream classdesc serialVersionUID = -5574015350443049520, local class serialVersionUID = 5876322619614900645
        at java.io.ObjectStreamClass.initNonProxy(ObjectStreamClass.java:616)
        at java.io.ObjectInputStream.readNonProxyDesc(ObjectInputStream.java:1630)
        at java.io.ObjectInputStream.readClassDesc(ObjectInputStream.java:1521)
        at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:1781)
        at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1353)
        at java.io.ObjectInputStream.defaultReadFields(ObjectInputStream.java:2018)
        at java.io.ObjectInputStream.readSerialData(ObjectInputStream.java:1942)
        at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:1808)
        at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1353)

zhichao-li:
Protobuf might be an option just like Caffe or Tensorflow. cc @yiheng @jason-dai

yiheng:
I think this problem is not related with Java serialization. If you change the layer, e.g. add a new field without a default value. You can't load old model even it is saved as protobuf.

The advantage of protobuf is its size is smaller. But it's not critical in model persistent scenario.

zhichao-li:
For this specific issue, it's due to there's no fix serialVersionUID for ```DenseTensor```  which would make it vulnerable from classes changes.
We need to check and ensure the related classes appended with a fix serialVersionUID if still stick with the Java serialization mode. 
cc @zhangxiaoli73

SeaOfOcean:
If use different scala version, it will also throw serialVersionUID error

srowen:
I would _avoid_ adding `serialVersionUID` to "fix" class incompatibility. It creates a potentially worse problem, that you update the class and don't change the ID and then are willing to load data that could be written by a different class version. It could actually succeed while having invalid data. It doesn't actually resolve the problem of loading different versions of a serialization automatically.

[...]

PR #1453, opened 2017-08-14
Module Serialization - layer-wise serialization support

wzhongyuan:

Following the default serialization framework, this PR contains
1) Layer-wise serialization support
For most layers, just one declaration needs , e.g,  object Linear extends ModuleSerialization {...}
15 Layers override the serialization traits as they need to store  their own states, they are

BinaryTreeLSTM
BiRecurrent
Graph
MapTable
MaskedSelect
Recurrent
Reshape
Scale
SpatialContrastiveNormalization
SpatialDivisiveNormalization
SpatialFullConvolution
SpatialMaxPooling
SpatialSubtractiveNormalization
Transpose
VolumetricMaxPooling

2) Unit test cases for all layers 
3) another 2 unit tests added to DataConverter
4) Module/AbstractModule to use ModuleLoader/ModulePersister instead of java based serialization

## How was this patch tested?
Unit tests and common models tested locally

https://github.com/intel-analytics/BigDL/issues/608

### DIFF
--- a/scripts/launch-dataproc.sh
+++ b/scripts/launch-dataproc.sh
@@ -1,14 +1,13 @@
 #!/bin/bash
 
 #
-# Licensed to Intel Corporation under one or more
-# contributor license agreements.  See the NOTICE file distributed with
-# this work for additional information regarding copyright ownership.
-# Intel Corporation licenses this file to You under the Apache License, Version 2.0
-# (the "License"); you may not use this file except in compliance with
-# the License.  You may obtain a copy of the License at
+# Copyright 2016 The BigDL Authors.
 #
-#    http://www.apache.org/licenses/LICENSE-2.0
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
 #
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


wzhongyuan:
@yiheng  this PR contains layer-wise support, but most modifications are very simple (just make the object extend from serialization trait) so you could go through very quickly.  Only 15 layers override the traits to store their states.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/BinaryTreeLSTM.scala
```diff
@@ -40,8 +42,8 @@ class BinaryTreeLSTM[T: ClassTag](
   withGraph: Boolean = true
 )(implicit ev: TensorNumeric[T])
   extends TreeLSTM[T](inputSize, hiddenSize) {
-  val composer: Module[T] = createComposer()
-  val leafModule: Module[T] = createLeafModule()
+  var composer: Module[T] = createComposer()
```
yiheng: can we wrap it in a method?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Module.scala
```diff
@@ -38,10 +39,25 @@ object Module {
    * @tparam T numeric type
    * @return model loaded from path
    */
+  @deprecated
```
yiheng: please add parameters to the annotation. Or it will throw an warning

yiheng:
As discussed, we should not override each factory method.

PR #1454, opened 2017-08-14
boost MSE

JerryYanWan:
Boost the performance of MSE with MKL

## How was this patch tested?
Unit Tests

~15x faster
inputsize: 3 * 50 * 50
before:
average eta = 4.503038999999999E-4
after:
average eta = 2.8870380000000003E-5


jason-dai:
LGTM.

PR #1458, opened 2017-08-15
support same padding in maxpooling and avgpooling

yangw1234:

support same padding in maxpooling and avgpooling

## How was this patch tested?

add unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialAveragePooling.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialAveragePooling.scala
@@ -374,6 +374,7 @@ class SpatialAveragePooling[@specialized(Float, Double) T: ClassTag](
 
     val (padTop, padBottom, padLeft, padRight, outputHeight, outputWidth) =
       if (padW == -1 && padH == -1) {
+        // no ceil/floor mode in SAME padding
         Utils.getSAMEOutSizeAndPadding(inputHeight, inputWidth, dH, dW, kH, kW)
       } else {
         require(inputWidth >= kW - padW && inputHeight >= kH - padH,
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialAveragePooling.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialAveragePooling.scala
@@ -98,7 +98,8 @@ class SpatialAveragePooling[@specialized(Float, Double) T: ClassTag](
   private def updateOutputFrameDouble(input: Tensor[Double], output: Tensor[Double],
     nInputPlane: Int, inputHeight: Int, inputWidth: Int,
     outputHeight: Int, outputWidth: Int,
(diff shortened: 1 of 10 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialAveragePooling.scala
```diff
@@ -29,6 +29,18 @@ import com.intel.analytics.bigdl.utils.Engine
...
+ *
+ * padAlongHeight = Math.max(0, (outHeight - 1) * strideH + kernelH - inHeight)
+ * padAlongWidth = Math.max(0, (outWidth - 1) * strideW + kernelW - inWidth)
+ *
+ * padTop = padAlongHeight / 2
+ * padLeft = padAlongWidth / 2
+ *
  * @param kW kernel width
  * @param kH kernel height
  * @param dW step width
```
yiheng: update the java doc param list?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialAveragePooling.scala
```diff
@@ -186,6 +200,24 @@ class SpatialAveragePooling[@specialized(Float, Double) T: ClassTag](
     }
   }
 
+  // return (padTop, padDown, padLeft, padRight)
+  protected def getPadding(inputHeight: Int, inputWidth: Int): (Int, Int, Int, Int) = {
```
yiheng: As discussed, move it to a util function

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialAveragePooling.scala
```diff
@@ -186,6 +200,24 @@ class SpatialAveragePooling[@specialized(Float, Double) T: ClassTag](
...
+    if (padW == -1 && padH == -1) {
+      // deal with SAME padding
+      val oW = Math.ceil(inputWidth.toFloat / dW.toFloat).toInt
```
yiheng: is Math.ceil right when poolingLayer.ceilMode=false?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialAveragePooling.scala
yangw1234: It's just an explanation, no parameters added here

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialAveragePooling.scala
yangw1234: It is correct actually. In the "SAME" padding algorithm, when padding is need, the algorithm will guarantee `inputHeight + padAlongH - kH` is always divisible by `dH` where the notion of `ceiling` and `floor` does not make sense. However, the code is a little confusing, I'll try to refactor that.

yangw1234:
pr validation pass

yiheng:
LGTM

PR #1464, opened 2017-08-15
remove set operation

JerryYanWan:
remove set operation in Recurrent because the memory binding will cause error when the last layer of Cell is activation.

## How was this patch tested?
Unit Tests


zhangxiaoli73:
Code LGTM. Have you run some examples like simplernn and lstm?

JerryYanWan:
yes

PR #1471, opened 2017-08-16
optimize Sigmoid

zhangxiaoli73:

optimize Sigmoid, use mkl tensor operation to increase perf

## How was this patch tested?

unit test


zhangxiaoli73:
jenkins pass

PR #1476, opened 2017-08-16
Support feed prior prediction back into recurrent

dding3:
Add RecurrentDecoder, it is a container of rnn cells which can make
a prediction of the next timestep based on the prediction we made from
the previous timestep.

Input for RecurrentDecoder has to be batch x ???(depends on cell type) without time information. 

During training, input at t(i) is output at t(i-1), input at t(0) is
user input.

Output for RecurrentDecoder has to be batch x outputLen x ???(depends on cell type).

Different types of rnn cells can be added using add() function. 

## How was this patch tested?
ut


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -219,13 +220,45 @@ class Recurrent[T : ClassTag]()
...
+        }
+        newInput.narrow(timeDim, i, 1).copy(inputTmp)
+        require(inputTmp.nElement() == input.select(timeDim, i).nElement(), "outputsize is " +
+          "not the same with input size!! Please update cell settings or turn off feedbackOutput.")
+        currentInput(inputDim) = if (preTopology != null) {
+          val sizes = 1 +: inputTmp.size()
+          inputTmp.resize(sizes)
+          val _input = preTopology.updateOutput(inputTmp).toTensor[T]
+          inputTmp.resize(sizes.takeRight(sizes.length - 1))
+          _input.select(1, 1)
```
JerryYanWan: I think the performance is poor using this setting. How about modify Cell layer to let preTopology be a choice.
Then if feedbackOutput == true, simply define another while loop.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -260,7 +293,13 @@ class Recurrent[T : ClassTag]()
       currentGradOutput(inputDim) = gradOutput.select(timeDim, i)
       _input(hidDim) = if (i > 1) cells(i - 2).output.toTable(hidDim)
         else hidden
-      _input(inputDim) = outputCell.select(timeDim, i)
+
+      _input(inputDim) = if (feedbackOutput && preTopology == null) {
+        newInput.select(timeDim, i)
+      } else {
+        outputCell.select(timeDim, i)
+      }
```
JerryYanWan: Such if-else condition will make code hard to rebase and review.
Split preTopology and feedbackOutput into two different branch?

jason-dai:
It's not clear what changes are proposed in the PR. Please describe, at a high level, what the proposed changes. 

It seems that you are adding an "decoder" or "generator" mode for the Recurrent layer, which generates the output sequence by predicting the next symbol Y(t), which depends on Y(t-1)? If so, we can add an flag for the "decoder" or "generator" mode; and please also describe the expected input of the mode.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
dding3: Actually I am thinking pretopology is transparent to user, i am not very likely to ask them to choose if or if not use it. So your concern here is with feedbackoutput, preTopology is kind of having poorer performance than without it? If this is the case, what about we turn off pretopology in recurrent if feedbackoutput is used.

jason-dai:
It is not clear what `feedbackoutput` means; we can change it to a `decoder` flag.

And please describe what the expect input of the `Recurrent` layer (not individual cell) is when the `decoder` flag is set to true.

And then please describe the expected behavior of the `Recurrent` layer when the `decoder` flag is set to true.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
```diff
@@ -0,0 +1,244 @@
...
+    require(input.dim == 2 || input.dim == 4 || input.dim == 5,
+      "Recurrent: input should be a 2D/4D/5D Tensor, e.g [batch, nDim], " +
+        s"current input.dim = ${input.dim}")
+
+    batchSize = input.size(batchDim)
+
+    val hiddenSize = topology.hiddensShape(0)
+    val outputSize = input.size()
+    outputSize(1) = hiddenSize
+    require(hiddenSize == input.size()(1), "hiddenSize is " +
```
JerryYanWan: `require(hiddenSize == input.size(2)`

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
```diff
@@ -0,0 +1,244 @@
...
+    // Clone N modules along the sequence dimension.
+    extend(featureSizes)
+
+    /**
+     * currentInput forms a T() type. It contains two elements, hidden and input.
+     * Each time it will feed the cell with T(hidden, input) (or T(input, hidden) depends on
+     * your hidDim and inputDim), and the cell will give a table output containing two
+     * identical elements T(output, output). One of the elements from the cell output is
+     * the updated hidden. Thus the currentInput will update its hidden element with this output.
+     */
```
JerryYanWan: please revise this docs

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
```diff
@@ -0,0 +1,244 @@
...
+    var i = 1
+    // init state
+    currentInput(hidDim) = if (initState != null) initState
+     else hidden
+
+    while (i <= times) {
```
JerryYanWan: what is `times` in this scenario?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
```diff
@@ -0,0 +1,244 @@
...
+        _input(inputDim) = input
+      }
+
+      cells(i - 1).accGradParameters(_input, currentGradOutput)
+      currentGradOutput(hidDim) = cells(i - 1).gradInput.toTable(hidDim)
+      i -= 1
+    }
+  }
+
+  override def updateGradInput(input: Tensor[T], gradOutput: Tensor[T]): Tensor[T] = {
```
JerryYanWan: is `GradOutput` a 3D tensor?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
```diff
@@ -0,0 +1,244 @@
...
+ * a prediction of the next timestep based on the prediction we made from
+ * the previous timestep. Input for RecurrentDecoder is dynamically composed
+ * during training. input at t(i) is output at t(i-1), input at t(0) is
+ * user input, and user input has to be batch x ???(depends on cell type)
+ * without time information.
+
+ * Different types of rnn cells can be added using add() function. Currently
+ * only support lstmpeephole, convlstm, convlstm3D cell.
+ */
+class RecurrentDecoder[T : ClassTag](outputLength: Int)
```
JerryYanWan: rename to `seqLen`?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
```diff
@@ -0,0 +1,244 @@
...
+
+    batchSize = input.size(batchDim)
+
+    val hiddenSize = topology.hiddensShape(0)
+    val outputSize = input.size()
+    outputSize(1) = hiddenSize
+    require(hiddenSize == input.size()(1), "hiddenSize is " +
+      "not the same with input size!! Please update cell settings or use Recurrent instead!")
+    val featureSizes = outputSize.drop(1)
+    output.resize(Array(batchSize, times) ++ featureSizes)
```
JerryYanWan: Is `output` a 3D tensor?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
```diff
@@ -0,0 +1,244 @@
...
+
+      if (i > 1) {
+        _input = cells(i - 2).output
+      } else {
+        _input(hidDim) = hidden
+        _input(inputDim) = input
+      }
+
+      cells(i - 1).updateGradInput(_input, currentGradOutput)
+      currentGradOutput(hidDim) = cells(i - 1).gradInput.toTable(hidDim)
```
JerryYanWan: A major problem here is that the gradOutput of cells(i) relies on two factors: one from cells(i+1).gradInput and the other from gradOutput.select(timeDim, i).
Thus, the gradOutput of cells(i) should be the summation of the above two factors, other wise the gradient might be inaccurate.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
```diff
@@ -0,0 +1,244 @@
...
+      if (i > 1) {
+        cells(i - 1).regluarized(false)
+        _input = cells(i - 2).output
+      } else {
+        cells(i - 1).regluarized(true)
+        _input(hidDim) = hidden
+        _input(inputDim) = input
+      }
+      cells(i - 1).backward(_input, currentGradOutput)
+      currentGradOutput(hidDim) = cells(i - 1).gradInput.toTable(hidDim)
```
JerryYanWan: The same, the gradOutput of cells(i) should be the summation of two factors.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
dding3: for recurrentdecoder, input is batch x outputsize x xxx, there is no time dimention. the second dimention is hiddensize.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
dding3: OK

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
dding3: output is the same with recurrent, it's batch x time x xxxx

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
dding3: I think currently logic is currentGradOutput(hidDim) = cells(i - 1).gradInput.toTable(hidDim), currentGradOutput(inputDim) = gradOutput.select(timeDim, i). it's a combination of that two factors.

jason-dai:
I don't think you need to change the behavior of the Cell wrt `pretopology`.

The `pretopology` is a D x W linear for RNN, and a D x 4W linear for LSTM; and previously `Recurrent` just pass the (B*T) x D input to the `pretopology`, and slice the appropriate portion for each timestep to sent to each cell.

`recurrent Decoder` can instead, for each time step, pass the B x H input to the `pretopology`, and then send the result to each cell.

dding3:
pretopology is set as null with recurrent because there is concern about performance downgrade @JerryYanWan ?

We can benefit from pretopology because we get w*input at all time step at one calculation. But with decoder, for one calculation we can only get w*input at one time step. We kind of lost benefits of pretopology. Besides we need do some tricky thing if we want to use pretopology with input which is B X D, like extend input to B X T X D before call pretopology and squeeze output of pretopology as B X D. so currently I turn pretopology off in recurrentdecoder.

jason-dai:
I don't think we should change the behavior of Cell, otherwise it becomes too complex. We can think of the cell in recurrentDecoder becomes a sequential of pretopology and cell

zhangxiaoli73:
Code LGTM, but I think it is better to add getTimes and resetTimes in RecurrentDecoder (like Recurrent) when debuging times of each layer.

dding3:
Thank you for your review. Jekins pass

docs/docs/APIGuide/Layers/Recurrent-Layers.md
```diff
@@ -148,6 +148,108 @@ array([[[ 0.69091094,  0.97150528,  0.9562254 ,  1.14894259],
...
+**Python:**
+```python
+module = RecurrentDecoder(output_length = 5)
+```
+
+RecurrentDecoder module is a container of rnn cells which used to make
+a prediction of the next timestep based on the prediction we made from
+the previous timestep.
+
+Input for RecurrentDecoder has to be batch x ???(depends on cell type) without time information.
```
jason-dai: what do you mean by ???

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Cell.scala
```diff
@@ -123,9 +123,9 @@ abstract class Cell[T : ClassTag](
             i += 1
           }
         } else {
-          val sizes = new Array[Int](imageSize.length + 2)
+          val sizes = new Array[Int](imageSize.length + 1)
           sizes(0) = batchSize
-          Array.copy(imageSize, 0, sizes, 2, imageSize.size)
+          Array.copy(imageSize, 0, sizes, 1, imageSize.size)
```
jason-dai: Need to add scaladoc for imageSize

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -37,23 +36,23 @@ import scala.reflect.ClassTag
...
-  private var topology: Cell[T] = null
-  private val outputBuffer = Tensor[T]()
+  protected var hidden: Activity = null
+  protected var gradHidden: Activity = null
+  protected var hiddenShape: Array[Int] = null
+  protected val currentInput = T()
+  protected val currentGradOutput = T()
+  protected val gradInputCell = Tensor[T]()
+  protected var outputCell = Tensor[T]()
+  protected var _input = T()
```
jason-dai: why change to var?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -111,18 +110,16 @@ class Recurrent[T : ClassTag](var batchNormParams: BatchNormParams[T] = null)
...
    * Clone N models; N depends on the time dimension of the input
    * @param sizes, the first element is batchSize, the second is times, the third is hiddensize
     *             the left is size of images
    */
-  private def extend(sizes: Array[Int]): Unit = {
-    val times = sizes(timeDim - 1)
-    val batchSize = sizes(batchDim - 1)
-    val imageSize = sizes.drop(3)
+  protected def extend(sizes: Array[Int]): Unit = {
+    val imageSize = sizes
```
jason-dai: change to stepShape?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
```diff
@@ -0,0 +1,217 @@
...
+/**
+ * [[RecurrentDecoder]] module is a container of rnn cells that used to make
+ * a prediction of the next timestep based on the prediction we made from
+ * the previous timestep. Input for RecurrentDecoder is dynamically composed
+ * during training. input at t(i) is output at t(i-1), input at t(0) is
+ * user input, and user input has to be batch x ???(depends on cell type)
+ * without time information.
+
+ * Different types of rnn cells can be added using add() function. Currently
+ * only support lstmpeephole, convlstm, convlstm3D cell.
```
jason-dai: Do we check that? And why can't we support RNN or LSTM?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
```diff
@@ -0,0 +1,217 @@
...
+ */
+class RecurrentDecoder[T : ClassTag](seqLength: Int)
```
jason-dai: add scala doc

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -258,7 +256,7 @@ class Recurrent[T : ClassTag](var batchNormParams: BatchNormParams[T] = null)
     cells(times - 1).output.toTable(hidDim)
   }
 
-  private var initState: Activity = null
+  protected var initState: Activity = null
   def setState(state: Activity): Unit = {
```
jason-dai: who will call setState?

docs/docs/APIGuide/Layers/Recurrent-Layers.md
dding3: Updated in another pr

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
dding3: We support RNN/LSTM, It's an out of date comment. Removed it in a new PR.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
dding3: In our current scenario, it will be used in decoder if there is only one cell. Besides getState/setState is a pair. we provide setState in case some user want to use a pretrained init state instead of zeros in encoder.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Cell.scala
dding3: Updated

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
jason-dai: Who will call that even for decoder?

dding3: My plan is giving a seq2seq layer, it will call setState for decoder.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
jason-dai: if hiddenSize == input.size()(1), why do you set outputSize(1) = hiddenSize?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -37,23 +36,23 @@ import scala.reflect.ClassTag
...
-  private var (batchSize, times) = (0, 0)
-  private var topology: Cell[T] = null
-  private val outputBuffer = Tensor[T]()
+  protected var hidden: Activity = null
+  protected var gradHidden: Activity = null
+  protected var hiddenShape: Array[Int] = null
+  protected val currentInput = T()
+  protected val currentGradOutput = T()
+  protected val gradInputCell = Tensor[T]()
+  protected var outputCell = Tensor[T]()
```
jason-dai: We should change outputCell and gradInputCell to more meaningful names. outputCells are actually the input sequence to the cells, and gradInputCell are gradInput for the input sequence to the cells.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -37,23 +36,23 @@ import scala.reflect.ClassTag
...
+  protected var _input = T()
+  protected val batchDim = Recurrent.batchDim
+  protected val timeDim = Recurrent.timeDim
+  protected val inputDim = 1
+  protected val hidDim = 2
+  protected var (batchSize, times) = (0, 0)
+  protected var topology: Cell[T] = null
+  protected val outputBuffer = Tensor[T]()
```
jason-dai: the name is confusing - it should be the buffer for the input of a single time step

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
```diff
@@ -0,0 +1,217 @@
...
+       * Thus, it needs to create a new Tensor.
+       */
+      if (preTopology.gradInput == null) {
+        preTopology.gradInput = Tensor[T]()
+      }
+      preTopology.gradInput.toTensor[T]
+    } else {
+      gradInputCell
+    }
+    gradInputCell.resizeAs(outputCell)
```
jason-dai: I don't think this is correct if preTopology == null: outputCell is B*T*H, and so gradInput is B*T*H, which is different from input (which is B*H)

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
```diff
@@ -0,0 +1,217 @@
...
+      "as it has override backward")
+    gradInput
+  }
+
+  override def backward(input: Tensor[T], gradOutput: Tensor[T]): Tensor[T] = {
+    val st = System.nanoTime
+    currentGradOutput(hidDim) = gradHidden
+    var i = times
+    while (i >= 1) {
+      currentGradOutput(inputDim) = if (i == times) gradOutput.select(timeDim, i)
```
jason-dai: You should use Recurrent.selectCopy instead of gradOutput.select

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
```diff
@@ -0,0 +1,217 @@
...
+      else {
+        gradOutput.select(timeDim, i).clone()
+          .add(cells(i).gradInput.toTable[Tensor[T]](inputDim).clone())
```
jason-dai: Use Recurrent.selectCopy instead of gradOutput.select, and then you can directly add cells(i).gradInput.toTable[Tensor[T]](inputDim); no need to clone the latter

jason-dai: Use Recurrent.selectCopy instead of gradOutput.select, and then you can directly add cells(i).gradInput; no need to clone the latter

jason-dai: I don't think this is correct: you cannot directly pass cells(i).gradInput to cells(i-1); you should propagate cells(i).gradInput through pretopology, and then to cells(i-1)

jason-dai: Use Recurrent.selectCopy instead of gradOutput.select, and then you can directly add cells(i).gradInput; no need to clone the latter

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
```diff
@@ -0,0 +1,217 @@
...
+        inputTmp
+      }
+      cells(i - 1).updateOutput(currentInput)
+      currentInput(hidDim) = cells(i - 1).output.toTable(hidDim)
+      i += 1
+    }
+
+    if (preTopology != null) {
+      // For backward preTopology use
+      outputCell = preTopology.updateOutput(newInput).toTensor[T]
```
jason-dai: I don't think this is correct; see comments below.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
```diff
@@ -0,0 +1,217 @@
...
+
+      currentInput(inputDim) = if (preTopology != null) {
+        newInput.narrow(2, i, 1).copy(inputTmp)
+        val sizes = 1 +: inputTmp.size()
+        inputTmp.resize(sizes)
+        val _input = preTopology.updateOutput(inputTmp).toTensor[T]
+        inputTmp.resize(sizes.takeRight(sizes.length - 1))
+        _input.select(1, 1)
+      } else {
+        outputCell.narrow(2, i, 1).copy(inputTmp)
```
jason-dai: no need to copy; just maintain an array of inputTmp, which can be then used in backward

jason-dai:
I think we can simply the implementation by:
1) Make preTopology single time step only; Reurrent can wrap preTopology using imeDistributed in itself

2) ReurrentDecoder does not need to handle preTopology at all; instead, it should combine its preTopology and topology into a new Cell (ParalleTable(preTopology, Identity) + topology), and then just use the new Cell as the cell for ReurrentDecoder

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
dding3: Removed unnecessary outputSize(1) = hiddenSize

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
dding3: Updated

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
dding3: Updated

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
dding3: Rename outputCell to input2Cell, gradInputCell to gradInput2Cell

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
dding3: Rename to stepInput2CellBuf

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
dding3: This code is a little old and this issue is fixed in a following multirnncell pr https://github.com/intel-analytics/BigDL/pull/1591/files

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
dding3: Fixed in multirnncell pr

PR #1477, opened 2017-08-17
Add Squeeze, Prod, Sum, Reshape, Identity, ReLU

psyyz10:
Add Squeeze, Prod, Sum, Reshape, Identity, ReLU

## How was this patch tested?
Unit tests


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Squeeze.scala
```diff
@@ -98,6 +98,6 @@ object Squeeze {
   def apply[T: ClassTag](
     dims : Array[Int], batchMode: Boolean)(implicit ev: TensorNumeric[T])
   : Squeeze[T] = {
-    new Squeeze[T](dims.sortWith(_>_), batchMode)
+    new Squeeze[T](if (dims != null) dims.sortWith(_>_) else null, batchMode)
```
yiheng: can it be a null?

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensorMath.scala
```diff
@@ -394,6 +394,18 @@ object DenseTensorMath {
     self
   }
 
+  def prodAll[@specialized(Float, Double) T](self: DenseTensor[T])(
+    implicit ev: TensorNumeric[T]): T = {
+    var product = ev.fromType[Int](1)
+    val func = new TensorFunc2[T] {
```
yiheng: Should we consider if it's contiguous?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Squeeze.scala
psyyz10: Yes, the default is null, and squeeze all dimensions.

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensorMath.scala
psyyz10: `prodAll()` is not used in any layer, just keep consistent with sumAll and similar implementation.
For the `Prod` operations, it will call `prod` method, which using `dimApply2`.

PR #1483, opened 2017-08-18
Make tensor support boolean, short, int, long, string, char

psyyz10:
Make tensor support boolean, short, int, long, string, char


yiheng:
LGTM

yiheng:
Add some unit test?

yiheng:
@dding3 what do you think?

psyyz10:
@yiheng Add some new operations and corresponding tests

dding3:
Could you give some explanation about why do we need support boolean and string for tensor?

psyyz10:
@dding3 for supporting tensorflow operations, such as readers for string and logicalOr and logicalAnd for boolean and so on.

Issue #1484, opened 2017-08-18
inplace layer should overriede clearState to avoid set output


SeaOfOcean:
inplace layer should overriede clearState to avoid set output

since it may set the model input if the first layer is an in-place layer

PR #1496, opened 2017-08-22
fix inplace layer clearstate

SeaOfOcean:

fix inplace layer clearstate
If it is the first layer, it will avoid reset the input tensor

## How was this patch tested?

unit test

fix https://github.com/intel-analytics/BigDL/issues/1484


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/AddConstant.scala
```diff
@@ -75,6 +75,14 @@ class AddConstant[T: ClassTag](
     val state = Seq(super.hashCode(), constant_scalar, inplace)
     state.map(getHashCode).foldLeft(0)((a, b) => 31 * a + b)
   }
+
+  override def clearState(): this.type = {
+    if (!inplace) {
+      output.set()
+    }
+    gradInput.set()
```
qiuxin2012: Put this to if block?

SeaOfOcean:
pr validate pass

PR #1498, opened 2017-08-23
Export tf checkpoint

yiheng:
Allow user export tensor values from tensorflow checkpoint file

## How was this patch tested?
existed unit test, new unit test

### DIFF
--- a/pyspark/bigdl/util/tf_utils.py
+++ b/pyspark/bigdl/util/tf_utils.py
@@ -66,6 +66,7 @@ def convert(input_ops, output_ops, byte_order, bigdl_type):
 
     return model
 
+
 def export_checkpoint(checkpoint_path):
     """
     Export variable tensors from the checkpoint files.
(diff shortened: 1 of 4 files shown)


yiheng:
PR validation pass

pyspark/bigdl/util/tf_utils.py
```diff
@@ -63,6 +66,74 @@ def convert(input_ops, output_ops, byte_order, bigdl_type):
...
+    :return: nothing
+    """
+    jtensors = {}
+    for tn in tensors.keys():
+        jtensors[tn] = JTensor.from_ndarray(tensors[tn])
+        
+    callBigDlFunc(bigdl_type, "saveTensorDictionary", jtensors, target_path)
+
+
+def dump_model(path, bigdl_type="float"):
```
yangw1234: Should we consider let user pass in the `Graph` and `Session` object explicitly, so that they can pass in other `Graph` besides the default graph and also pass in a session that restored from a existing pre-trained checkpoint if he/she want to.

yangw1234:
LGTM

PR #1499, opened 2017-08-23
support loading a subset of tf graph and support control dependency in graph

yangw1234:

1. support loading a subset of tf graph
2. Change `Context` key type to String for future development convenience.
3. handle tensorflow control dependency

## How was this patch tested?

add unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/TensorflowLoader.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/TensorflowLoader.scala
@@ -64,7 +64,7 @@ object TensorflowLoader{
   }
 
   /**
-   * Parse a tensorflow model protobuf file, read a list of op nodes from it
+   * Parse a tensorflow model protobuf binary file, read a list of op nodes from it
    * @param graphProtoTxt where is the tf protobuf file
    * @return
    */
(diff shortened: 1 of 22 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/TensorflowLoader.scala
```diff
@@ -105,29 +127,75 @@ object TensorflowLoader{
...
-      name2Node.valuesIterator.filter(_.nextNodes.length == 0).toArray
+      name2Node.valuesIterator.filter(_.nextNodes.isEmpty).toArray
     } else {
       val results = name2Node.valuesIterator.toArray.filter(n =>
         outputNodeNames.contains(n.element.getName))
       require(results.length == outputNodeNames.length, "Invalid outputNode names")
       results
     }
 
+    def connect(nodes: Seq[Node[NodeDef]]): Seq[String] = {
```
yiheng: private

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/TensorflowLoader.scala
```diff
@@ -84,13 +87,32 @@ object TensorflowLoader{
   }
 
   /**
+   * Parse a tensorflow model protobuf file, read a list of op nodes from it
+   * @param graphProtoTxt where is the tf protobuf file
+   * @return
+   */
+  private[bigdl] def parseTxt(graphProtoTxt: String) : List[NodeDef] = {
```
yiheng: what's the difference from parse

yangw1234: parseTxt process model.pbtxt, parse process model.pb

yangw1234:
pr validation pass

PR #1507, opened 2017-08-24
Allow module output multiple tensor in Graph

yiheng:

In this PR, user can put a module with multiple output tensor into a graph. Here is an example:
```scala
val x = Input()
val y = SplitTable.inputs(x)
val z1 = Linear.inputs((y, 1))  // Or val z1 = Linear.inputs(y(1)) 
val z2 = Linear.inputs((y, 2))  // Or val z2 = Linear.inputs(y(2))
val output = CAddTable(z1, z2)  // or val output = CAddTable(x)
val model = Graph(x, output)
```

## How was this patch tested?
unit tests


jason-dai:
Can we do sth like:

`val z1 = Linear.inputs(y[1]))`

yiheng:
@jason-dai Yes. Now we support that format.

yiheng:
PR validation pass

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -586,16 +586,32 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
...
-   * Some other modules point to current module
+   * Build graph: some other modules point to current module
    * @param nodes upstream module nodes
    * @return node containing current module
    */
   def inputs(nodes : ModuleNode[T]*): ModuleNode[T] = {
     require(this.isInstanceOf[AbstractModule[_, Tensor[T], T]],
-      "AbstractModule: Only module with tensor output can be added into a graph node")
-    val curNode = new ModuleNode[T](this.asInstanceOf[AbstractModule[Activity, Tensor[T], T]])
+      s"AbstractModule: ${this.getClass.getName} has multiple outputs, please use (module, index)" +
```
jason-dai: use module[i]?

yiheng: Because of the type erase, this will be always true.. Have removed this, as we support connect a table to a node.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -77,21 +76,26 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
       val node = forwardExecutions(i)
       val nodeInput = if (node.prevNodes.isEmpty && !node.element.isInstanceOf[WithoutInput]) {
         inputData(node, input)
-      } else if (node.prevNodes.length == 1) {
-        node.prevNodes.head.element.output.toTensor[T]
       } else {
-        seqToTable(node.prevNodes.map(_.element.output))
+        var toOffset = 1
+        val prevTensors = node.prevNodesAndEdges.flatMap(n => {
```
yangw1234: It seems Graph outputs one big flat table. Say a Graph has multiple output nodes, and one of the output node outputs unknown number of tensors, then the user of this Graph seems have no way to determine which tensors in the output table belongs to which output node.

yiheng: fixed

yiheng:
PR validation pass

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -102,20 +105,26 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
...
     while (i < backwardExecutions.length) {
       val curNode = backwardExecutions(i)
-      var curGradOutput : Tensor[T] = null
+      var curGradOutput : Activity = null
 
-      curNode.nextNodes.foreach(n => {
-        val nextGradOutput = if (n.prevNodes.length == 1) {
-          n.element.gradInput.toTensor
+      curNode.nextNodesAndEdges.foreach(n => {
+        val index = n._1.prevEdges.indexOf(n._2) + 1
```
jason-dai: val otherActivity = if (n._1.element.gradInput.isTensor || n._1.prevEdges.length == 1)  {
...
} else {
   val index = n._1.prevEdges.indexOf(n._2) + 1
   ...
}

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -167,20 +176,26 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
...
+      var curGradOutput : Activity = null
+      if (curNode.element.output.isTable) {
+        curGradOutput = T()
+      }
 
-      curNode.nextNodes.foreach(n => {
-        val nextGradOutput = if (n.prevNodes.length == 1) {
-          n.element.gradInput.toTensor
+      curNode.nextNodesAndEdges.foreach(n => {
+        val index = n._1.prevEdges.indexOf(n._2) + 1
```
jason-dai:
```scala
val otherActivity = if (n._1.element.gradInput.isTensor 
                                  || n._1.prevEdges.length == 1)  {
...
} else {
   val index = n._1.prevEdges.indexOf(n._2) + 1
   ...
}
```

jason-dai:
Do you have some results on the performance impacts of the new implementation?

yiheng:
@jason-dai 

(warm up 20 iterations, single core)

AlexNet 4x3x227x227 forward+backward before(343.52830755ms), after(334.16659185ms)
Inception 4x3x224x224 forward+backward before(729.88021925ms), after(745.8929125ms)
ResNet20 4x3x224x224 forward before(1054.4564839500001ms), after(1045.0969059ms)
Lenet 32x28x28 forward+backward before(14.650214299999998ms), after(13.906648950000001ms)

PR #1510, opened 2017-08-25
Refactor the serde logic of JTensor and Sample

zhichao-li:

Previously we pass JTensor from python to java via ArrayList which would result in bad performance.
After this changes we would use Array[Byte] directly for those transforming.

## How was this patch tested?
unittest

### DIFF
--- a/pyspark/bigdl/util/common.py
+++ b/pyspark/bigdl/util/common.py
@@ -190,7 +190,7 @@ class Sample(object):
     @classmethod
     def from_ndarray(cls, features, label, bigdl_type="float"):
         """
-        Convert a ndarray to Tensor which would be used in Java side.
+        Convert a ndarray of features and label to Sample, which would be used in Java side.
 
         >>> import numpy as np
         >>> from bigdl.util.common import callBigDlFunc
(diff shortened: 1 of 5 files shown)


zhichao-li:
Jenkins passed. http://172.168.2.101:8080/view/PR-Validation/job/BigDL-PR-Validation/1712/

zhichao-li:
cc @jenniew

zhichao-li:
cc @jason-dai  @yiheng  any comments? I would merge it if no objections.

pyspark/bigdl/util/common.py
```diff
@@ -148,75 +154,69 @@ def from_ndarray(cls, a_ndarray, bigdl_type="float"):
...
+        :param bigdl_type: "double" or "float"
+        """
+        self.features = features
+        self.label = label
         self.bigdl_type = bigdl_type
 
     @classmethod
     def from_ndarray(cls, features, label, bigdl_type="float"):
+        """
+        Convert a ndarray to Tensor which would be used in Java side.
```
jason-dai: to Sample?

PR #1511, opened 2017-08-25
serialization - add custom data conversion support

wzhongyuan:

Advanced feature to support user defined object conversion -

if user implements his/her own layer with customized data type not supported in bigdl, the serialization will fail, in order to still serialize user implemented layer, this PR provides interface to register data converter for customized data types

## How was this patch tested?

unit test


wzhongyuan:
PR validation passed 
http://10.239.47.211:18888/job/BigDL-PR-Validation/1730/

yiheng:
LGTM

PR #1514, opened 2017-08-28
add isSplitInput in BiRecurrent

JerryYanWan:
add a different behavior inside BiRecurrent layer:

isCloneInput = true  => ConcatTable
isCloneInput = false => BifurcateSplitTable

## How was this patch tested?
Unit Tests


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/BiRecurrent.scala
```diff
@@ -35,24 +35,34 @@ import scala.reflect.runtime._
  */
 class BiRecurrent[T : ClassTag] (
   private val merge: AbstractModule[Table, Tensor[T], T] = null,
-  val batchNormParams: BatchNormParams[T] = null)
+  val batchNormParams: BatchNormParams[T] = null,
+  val isCloneInput: Boolean = true)
```
jason-dai: Maybe use a splitInput flag instead of a cloneInput flag?

PR #1517, opened 2017-08-29
Support 0-D Tensor (Scalar)

yangw1234:
### support scalar in tensor

1. a scalar tensor has 0 dimension
2. a scalar tensor's size is an empty int array
3. a scalar tensor only has one element

## How was this patch tested?

add unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
@@ -39,7 +39,6 @@ private[tensor] class DenseTensor[@specialized(Float, Double) T: ClassTag](
 
   override def isScalar: Boolean =
     this.nDimension == 0 &&
-      this._size != null && this.size().isEmpty &&
       this._storage.length() == 1
 
   override def storage(): Storage[T] = _storage
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
@@ -35,6 +35,13 @@ private[tensor] class DenseTensor[@specialized(Float, Double) T: ClassTag](
   var nDimension: Int)(implicit ev: TensorNumeric[T])
   extends Tensor[T] {
 
+  override def isEmpty: Boolean = this.storage() == null || this.storage().length() == 0
+
+  override def isScalar: Boolean =
+    this.nDimension == 0 &&
+      this._size != null && this.size().isEmpty &&
+      this._storage.length() == 1
+
   override def storage(): Storage[T] = _storage
 
   override def storageOffset(): Int = _storageOffset + 1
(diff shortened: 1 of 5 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/Tensor.scala
```diff
@@ -34,6 +34,10 @@ import scala.reflect.ClassTag
  * @tparam T should be Double or Float
  */
 trait Tensor[T] extends Serializable with TensorMath[T] with Activity {
+
+  def isEmpty: Boolean
```
yiheng: please add java doc for public method

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/Tensor.scala
```diff
@@ -34,6 +34,10 @@ import scala.reflect.ClassTag
...
+
+  def isScalar: Boolean
```
yiheng: please add java doc for public method

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/Tensor.scala
```diff
@@ -34,6 +34,18 @@ import scala.reflect.ClassTag
  * @tparam T should be Double or Float
  */
 trait Tensor[T] extends Serializable with TensorMath[T] with Activity {
+
+  /**
+   * @return whether this tensor is an empty tensor. Note that nDimension == 0 is not
+   *         sufficient to determine a tensor is empty, because a scalar tensor's nDimension
+   *         is also 0.
+   */
+  def isEmpty: Boolean
```
yiheng: can nElement == 0 check if a tensor is empty

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
```diff
@@ -35,14 +35,21 @@ private[tensor] class DenseTensor[@specialized(Float, Double) T: ClassTag](
   var nDimension: Int)(implicit ev: TensorNumeric[T])
   extends Tensor[T] {
 
+  override def isEmpty: Boolean = this.storage() == null || this.storage().length() == 0
+
+  override def isScalar: Boolean =
+    this.nDimension == 0 &&
```
yiheng: this.nDimension == 0 && this._storage.length() != 1
does such tensor valid?

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/Tensor.scala
yangw1234: Yes, but it may iterate through all the dimensions if the tensor is not empty.

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
yangw1234: User passes in an empty storage?

yangw1234: nDimension and storage length is sufficient to determine whether a tensor is scalar. I will remove the test on size.

yangw1234:
pr validation pass

PR #1518, opened 2017-08-30
fix: only copy the elements in tensor when serializing

i8run:

## How was this patch tested?

jenkins.


wzhongyuan:
LGTM

PR #1520, opened 2017-08-30
Make layer suppor different input and output numeric type and Add Equal Operation as an example

psyyz10:
Make layer suppor different input and output numeric type and Add Equal Operation as an example

## How was this patch tested?
Unit tests


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/Equal.scala
```diff
@@ -0,0 +1,209 @@
...
+   * Iterate through tensor1, tensor2, and apply func to the elements,
+   * set function result to tensor 3
+   *
+   * @param tensor1 the tensor1
+   * @param tensor2 the tensor2
+   * @param tensor3 the result tensor
+   * @param func    (tensor1Data, tensor1Offset, tensor2Data,
+   *                tensor2Offset, tensor3Data, tensor3Offset)
+   */
+  def apply2[A, B, C](tensor1: Tensor[A], tensor2: Tensor[B], tensor3: Tensor[C],
```
yangw1234: How about move this method to a until object in ops package, so that other Operation can also use it?

PR #1525, opened 2017-08-31
Add Greater and Less operations

psyyz10:
Add Greater and Less operations


yiheng:
Do not merge such code without code review!!!

spark/dl/src/test/scala/com/intel/analytics/bigdl/nn/ops/LessSpec.scala
```diff
@@ -0,0 +1,113 @@
...
+  "Less Double operation" should "works correctly" in {
+    val input =
+      T(
+        Tensor[Double](T(4.0, 3.0, 1.0)),
+        Tensor[Double](T(2.0, 3.0, 2.0))
+      )
+
+    val expectOutput = Tensor[Boolean](T(false, false, true))
+
+    val output = Less[Boolean]().forward(input)
```
yiheng: What do you mean of Less[Boolean]()? The T of Layer[T] is parameter type? So do you mean the layer has a boolean parameter?

PR #1529, opened 2017-09-04
Add tfrecord format reader and tfrecord Example, Features protobuf generated classes

yangw1234:

1. Add reader to read tfrecord format file
2. Add tfrecord Example, Features protobuf generated classes for ParseExample operator.
3. Upgrade protobuf version to 3.4.0 (manually tested on inception module loader and saver)

## How was this patch tested?

add unit tests


jason-dai:
What do you mean by tf Example?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/TFRecordIterator.scala
```diff
@@ -0,0 +1,78 @@
...
+ * Internal use only.
+ *
+ * TF record format:
+ *  uint64 length
+ *  uint32 masked_crc32_of_length
+ *  byte   data[length]
+ *  uint32 masked_crc32_of_data
+ *
+ */
+private[tf] class TFRecordIterator(fileName: File) extends Iterator[Array[Byte]] {
```
jason-dai: this is local file only?

yangw1234:
@jason-dai Example is a protobuf class provided by tensorflow to represent a record in tfrecod file. The tensorflow doc says  "The recommended format for TensorFlow is a TFRecords file containing tf.train.Example protocol buffers (which contain Features as a field)."

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/TFRecordIterator.scala
yangw1234: Yes, for local file only.

jason-dai:
So it's tfrecord example, yes?

yangw1234:
pr validation pass

PR #1533, opened 2017-09-04
Add Floor L2Loss RandomUniform Rank MatMul SoftMax operations

psyyz10:
Add Floor L2Loss RandomUniform Rank MatMul SoftMax operations

## How was this patch tested?
Unit tests


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/Floor.scala
```diff
@@ -0,0 +1,47 @@
...
+ */
+package com.intel.analytics.bigdl.nn.ops
+
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+import com.intel.analytics.bigdl.tensor._
+
+import scala.reflect.ClassTag
+
+class Floor[T: ClassTag]()
+  (implicit ev: TensorNumeric[T]) extends Operation[Tensor[T], T] {
```
yangw1234: Should we change the Operation signature so that the output type is not coupled with type parameter T. CC @yiheng

PR #1536, opened 2017-09-06
add tensor of ByteString in tf package

yangw1234:

remove tensor of String and add tensor of ByteString in tf package

## How was this patch tested?

add unit tests


yangw1234:
pr validation pass

PR #1506, opened 2017-08-24
Remove the code dealing with translating tensorflow NHWC format to BigDL NCHW format

yangw1234:
## What changes were proposed in this pull request?

Refactor and rm the code dealing with translating tensorflow NHWC format to BigDL NCHW format

## How was this patch tested?

existing tests and manually tested

PR #1538, opened 2017-09-06
update integration test for tensorflow model load/save on hdfs/s3

frankfzw:

1. Fix data format mismatch in integration test of tensorflow model load/save on hdfs and s3 
2. Change data format in `TensorflowSaverSpec` to `NHWC` except `BatchNormalization` since BigDL has not support `NHWC` `BatchNormalization` yet.
3. Remove format transformation while saving BigDL model to Tensorflow model. The saved tf model will have same format as the original BigDL model.
4. Add Reshape node in `BatchNorm2DToTF` to support `NCHW` batch normalization. The dimension size 
 of ReShape is set to [1, C, 1, 1] since the `BatchNormalization` in BigDL is `NCHW`. The new topology of saved graph is same as `tf.layers.batch_normalization`.
graph-run

## How was this patch tested?

unit tests, integration tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/BigDLToTensorflow.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/BigDLToTensorflow.scala
@@ -18,7 +18,7 @@ package com.intel.analytics.bigdl.utils.tf
 import java.nio.ByteOrder
 
 import com.intel.analytics.bigdl.nn._
-import com.intel.analytics.bigdl.nn.abstractnn.AbstractModule
+import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, DataFormat}
 import com.intel.analytics.bigdl.tensor.Tensor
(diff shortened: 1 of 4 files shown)


yiheng:
@yangw1234 please take a look at this PR. Looks like your previous change https://github.com/intel-analytics/BigDL/pull/1506  break the integration test.

yangw1234:
@frankfzw Hi, can you also change the tensorflow saver code, so that a bigdl NCHW model saves to a tensorflow NCHW model and so as NHWC. Then we can get rid of the `transpose` code in this test case, which is very confusing. cc @yiheng

frankfzw:
@yangw1234 I can add data format in `TensorflowSaver` and refactor the corresponding test code. cc @yiheng

frankfzw:
@yangw1234 The unit tests are all passed under tensorflow cpu, tensorflow mkl and tensorflow gpu in a local server. cc @yiheng

frankfzw:
pr validation and integration tests passed

yangw1234:
LGTM

PR #1539, opened 2017-09-06
Implement and test Conv2d operation

psyyz10:

Implement and test Conv2d operation

## How was this patch tested?

Unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/Conv2D.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/Conv2D.scala
@@ -1,0 +1,105 @@
+/*
+ * Copyright 2016 The BigDL Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.intel.analytics.bigdl.nn.ops
+
+import com.intel.analytics.bigdl.nn.SpatialConvolution
+import com.intel.analytics.bigdl.nn.abstractnn.DataFormat
+import com.intel.analytics.bigdl.tensor._
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+import com.intel.analytics.bigdl.utils.Table
+
+import scala.reflect.ClassTag
+
+class Conv2D[T: ClassTag](
+  input: Tensor[T],
+  filter: Tensor[T],
+  strides: Array[Int],
+  padding: String,
+  format: DataFormat = DataFormat.NHWC
+)(implicit ev: TensorNumeric[T]) extends Operation[Tensor[T], T] {
+
+  val conv: SpatialConvolution[T] = format match {
+    case DataFormat.NHWC =>
+      if (padding == "SAME") {
+        SpatialConvolution(
+          nInputPlane = input.size(4),
+          nOutputPlane = filter.size(4),
+          kernelH = filter.size(2),
+          kernelW = filter.size(3),
+          strideH = strides(2),
+          strideW = strides(3),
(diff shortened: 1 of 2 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/Conv2D.scala
```diff
@@ -0,0 +1,108 @@
...
+    case DataFormat.NCHW =>
+      if (padding == "SAME") {
+        SpatialConvolution(
+          nInputPlane = input.size(2),
+          nOutputPlane = filter.size(4),
+          kernelH = filter.size(1),
+          kernelW = filter.size(2),
+          strideH = strides(2),
+          strideW = strides(3),
+          padH = (filter.size(3) - 1) / 2,
```
yangw1234: Actually, in "SAME" padding, padLeft/padTop can be different from padRight/padBottom. Maybe you should try using padH = -1 and padW = -1

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/Conv2D.scala
```diff
@@ -0,0 +1,108 @@
...
+import com.intel.analytics.bigdl.tensor._
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+import com.intel.analytics.bigdl.utils.Table
+
+import scala.reflect.ClassTag
+
+class Conv2D[T: ClassTag](
+  input: Tensor[T],
+  filter: Tensor[T],
+  strides: Array[Int],
```
yangw1234: I think input and filter should be Conv2D's input instead of its constructor parameters.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/Conv2D.scala
```diff
@@ -0,0 +1,107 @@
...
+            strideH = strides(2),
+            strideW = strides(3),
+            format = format
+          )
+        } else {
+          throw new RuntimeException("Padding can only support SAME and VALID padding")
+        }
+    }
+
+    conv.bias.zero()
```
yangw1234: SpatialConvolution constructor has an option indicating whether to add bias

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/Conv2D.scala
```diff
@@ -0,0 +1,107 @@
...
+      case DataFormat.NCHW =>
+        if (padding == "SAME") {
+          SpatialConvolution(
+            nInputPlane = input.size(2),
+            nOutputPlane = filter.size(4),
+            kernelH = filter.size(1),
+            kernelW = filter.size(2),
+            strideH = strides(2),
+            strideW = strides(3),
+            padH = (filter.size(3) - 1) / 2,
```
yangw1234: padH = -1, padW = -1

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/Conv2D.scala
```diff
@@ -0,0 +1,107 @@
...
+            format = format
+          )
+        } else {
+          throw new RuntimeException("Padding can only support SAME and VALID padding")
+        }
+    }
+
+    conv.bias.zero()
+
+    conv.weight.resizeAs(filter).copy(filter)
```
yangw1234: Consider using SpatialConvolution constructor's initWeight to set the weight?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/Conv2D.scala
```diff
@@ -0,0 +1,107 @@
...
+  strides: Array[Int],
+  padding: String,
+  format: DataFormat = DataFormat.NHWC
+)(implicit ev: TensorNumeric[T]) extends Operation[Table, Tensor[T], T] {
+
+  override def updateOutput(inputs: Table): Tensor[T] = {
+    val input: Tensor[T] = inputs[Tensor[T]](1)
+    val filter: Tensor[T] = inputs[Tensor[T]](2)
+
+    val conv: SpatialConvolution[T] = format match {
```
yangw1234: Should we consider caching the conv object?

psyyz10: The filer is a input, it can be changed, so the filter size can be changed. They will be different convolution objects.

PR #1544, opened 2017-09-08
Support dynamic control flow in BigDL graph

yiheng:
Instead of using top sort, this PR implement a scheduler which can support dynamic path execution in BigDL graph. Some control flow related ops are also included in this PR(switch, merge, enter loop, exit loop, next iteration, loop condition). Users should not create these nodes directly, but use them through provided API.

Another improvement in this PR is that for const nodes, or node can be inferred by const nodes, the scheduler won't execute them multiple times.

Because of the overhead of the scheduler, comparing to a static graph, the performance is a little decrease in testing, see below(single core, warm up 20, avg 20)

AlexNet 4x3x227x227 forward+backward before(334ms), after(380ms)
Inception 4x3x224x224 forward+backward before(745ms), after(770ms)
ResNet20 4x3x224x224 forward before(1045ms), after(1201ms)
Lenet 32x28x28 forward+backward before(13ms), after(16.79013465ms)

## How was this patch tested?
unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Scheduler.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Scheduler.scala
@@ -1,0 +1,308 @@
+/*
+ * Copyright 2016 The BigDL Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Scheduler.scala
```diff
@@ -0,0 +1,308 @@
...
+            Ready()
+          }
+        } else {
+          val constNodes = node.prevNodes.filter(nodeStatus.isConst(_))
+          if (constNodes.length == node.prevNodes.length) {
+            Const()
+          } else {
+            val iterationNodes = node.prevNodes.filter(nodeStatus.isIteration(_))
+              .filter(!nodeStatus.iteration(_).outOfDate)
+              .filter(!_.element.isInstanceOf[LoopCondition[_]])
```
yangw1234: Is it possible that two different loop variables have some kind interaction, so that the iteration of one loop variable can be falsely passed to the other one by some connection other than the "LoopCondition" node

PR #1548, opened 2017-09-11
fix scalar api

yangw1234:

1. remove the Tensor factory method that is ambiguous with original ones
2. add a convenient valueAt method to return a scalar value  

## How was this patch tested?

add unit tests


yangw1234:
pr validation pass

PR #1550, opened 2017-09-12
[Enhancement] Revise activity allocate

yiheng:
revise activity allocate method

## How was this patch tested?
unit test


yiheng:
PR validation pass

PR #1551, opened 2017-09-12
fix tensorflow loader bug when two nodes are connected by two edges

yangw1234:

fix tensorflow loader bug when two nodes are connected by two edges

## How was this patch tested?

add unit test


yiheng:
LGTM

PR #1555, opened 2017-09-13
Support switch and merge operation

yiheng:
Instead of using top sort, this PR implement a scheduler which can support dynamic path execution in BigDL graph. Some control flow related ops are also included in this PR(switch, merge). Users should not create these nodes directly, but use them through provided API.

Another improvement in this PR is that for const nodes, or node can be inferred by const nodes, the scheduler won't execute them multiple times.

AlexNet 4x3x227x227 forward+backward before(334ms), after(336ms)
Inception 4x3x224x224 forward+backward before(745ms), after(715ms)
ResNet20 4x3x224x224 forward before(1045ms), after(1054ms)
Lenet 32x28x28 forward+backward before(13ms), after(12.3ms)

## How was this patch tested?
unit test


yangw1234:
LGTM

yiheng:
PR validation pass

PR #1556, opened 2017-09-13
optimize copy in recurrent

zhangxiaoli73:

optimize copy in recurrent layer

## How was this patch tested?

unit test

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
@@ -161,65 +161,6 @@ class Recurrent[T : ClassTag](var batchNormParams: BatchNormParams[T] = null)
   }
 
   /**
-   * set the cells' output and gradInput to recurrent's output and gradInput
-   * to decrease the copy expense.
-   * @param src
-   * @param dst
-   */
-  private[bigdl] def copy(src: ArrayBuffer[Tensor[T]], dst: Tensor[T], offset: Int): Unit = {
-    val dstArr = dst.storage().array()
-    val dstOffset = dst.storageOffset() - 1
-    val otherSize = dst.nElement() / (batchSize * times)
-
-    var t = 1
-    val length2 = batchSize * otherSize
-    while ((t + offset) <= times) {
-      val srcArr = src(t - 1).storage().array()
-      val srcOffset = src(t - 1).storageOffset() - 1
-      val length1 = (t - 1) * otherSize + dstOffset
-      var l = 0
-      while (l < length2) {
-        System.arraycopy(srcArr, l + srcOffset, dstArr, times * l + length1, otherSize)
-        l += otherSize
-      }
-      t += 1
-    }
-  }
-
-  /**
-    * Create a new tensor dst which exchanges the 1 and 2 dimensions of src tensor
-    * @param src
-    * @param dst
-    */
-  private[bigdl] def transposeMemory(src: Tensor[T], dst: Tensor[T]): Unit = {
-    val srcSize = src.size()
-    val batchSize = srcSize(0)
-    val timeSize = srcSize(1)
-    val otherSize = src.nElement() / (batchSize * timeSize)
-    val srcArr = src.storage().array()
-    val srcOffset = src.storageOffset() - 1
-
-    srcSize(0) = timeSize
-    srcSize(1) = batchSize
-    dst.resize(srcSize)
-    val dstArr = dst.storage().array()
-    val dstOffset = dst.storageOffset() - 1
-
-    val length3 = timeSize * otherSize
-    var t = 1
-    while (t <= batchSize) {
-      var l = 0
-      val length1 = timeSize * otherSize * (t-1) + srcOffset
-      val length2 = (t-1) * otherSize + dstOffset
-      while (l < length3) {
-        System.arraycopy(srcArr, length1 + l, dstArr, l * batchSize + length2, otherSize)
-        l += otherSize
-      }
-      t += 1
-    }
-  }
-
-  /**
    * Sharing weights, bias, gradWeights across all the cells in time dim
    * @param cells
    */
@@ -283,7 +224,7 @@ class Recurrent[T : ClassTag](var batchNormParams: BatchNormParams[T] = null)
     }
 
     size = preOutput.size()
-    transposeMemory(preOutput, outputCell)
+    Recurrent.transposeMemory(preOutput, outputCell)
 
     val hiddenSize = topology.hiddensShape(0)
     val outputSize = input.size()
@@ -310,7 +251,7 @@ class Recurrent[T : ClassTag](var batchNormParams: BatchNormParams[T] = null)
       i += 1
     }
 
-    copy(cells.map(x => x.output.toTable[Tensor[T]](inputDim)),
+    Recurrent.copy(cells.map(x => x.output.toTable[Tensor[T]](inputDim)),
         output, 0)
     output
   }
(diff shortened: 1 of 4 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -229,12 +276,15 @@ class Recurrent[T : ClassTag](var batchNormParams: BatchNormParams[T] = null)
...
 
-    outputCell = if (preTopology != null) {
+    val preOutput = if (preTopology != null) {
       preTopology.forward(input).toTensor[T]
     } else {
       input
     }
 
+    size = preOutput.size()
+    transposeMemory(preOutput, outputCell)
```
qiuxin2012: copy before compute, and you can reduce the size to preOutput / times

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -43,14 +42,17 @@ class Recurrent[T : ClassTag](var batchNormParams: BatchNormParams[T] = null)
   private val currentInput = T()
   private val currentGradOutput = T()
   private val gradInputCell = Tensor[T]()
-  private var outputCell = Tensor[T]()
+  private val outputCell = Tensor[T]()
+  private var size: Array[Int] = null
   private val _input = T()
   private val batchDim = 1
   private val timeDim = 2
+  private val timeTranspose = 1
```
jason-dai: Change to timeTransposeDim?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -229,12 +217,15 @@ class Recurrent[T : ClassTag](var batchNormParams: BatchNormParams[T] = null)
...
     times = input.size(timeDim)
 
-    outputCell = if (preTopology != null) {
+    val preOutput = if (preTopology != null) {
       preTopology.forward(input).toTensor[T]
     } else {
       input
     }
 
+    size = preOutput.size()
```
jason-dai: Change to preSize or preOutputSize

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -229,12 +217,15 @@ class Recurrent[T : ClassTag](var batchNormParams: BatchNormParams[T] = null)
...
+    Recurrent.transposeMemory(preOutput, outputCell)
```
jason-dai: Change to Recurrent.transpose.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -497,6 +492,70 @@ object Recurrent extends ContainerSerializable {
...
+   * exchanges the 1 and 2 dimensions of src tensor, and copy memory to dst
+   * @param src
+   * @param dst
+   */
+  private[bigdl] def transposeMemory[@specialized(Float, Double) T: ClassTag](
+    src: Tensor[T], dst: Tensor[T]): Unit = {
+    val srcSize = src.size()
+    val batchSize = srcSize(0)
+    val timeSize = srcSize(1)
+    val otherSize = src.nElement() / (batchSize * timeSize)
```
jason-dai: Change to stepSize

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -497,6 +492,70 @@ object Recurrent extends ContainerSerializable {
...
+
+    srcSize(0) = timeSize
+    srcSize(1) = batchSize
+    dst.resize(srcSize)
+    val dstArr = dst.storage().array()
+    val dstOffset = dst.storageOffset() - 1
+
+    val length3 = timeSize * otherSize
+    var t = 1
+    while (t <= batchSize) {
```
jason-dai: It is very complex. Maybe change to sth clearer, like below:
```scala

val recordSize = timeSize * stepSize
val batchStepSize = batchSize * stepSize

var srcOffset = src.storageOffset() – 1
var dstOffset = dst.storageOffset() - 1

var b = 1
while(b <= batchSize) {
  var t = 1
  var l = 0
  var m = 0
  while (t <= timeSize) {
    System.arraycopy(srcArr, srcOffset + l, dstArr, dstOffset + m, stepSize)
    l += stepSize
    m += batchStepSize
    t += 1
  }
  srcOffset  += recordSize 
  dstOffset += stepSize
} 
```
Likewise for the copy method above.

jason-dai: And actually it's possible to use copy method to implement the transpose method, so that we don't need two pieces of similar code, yes?

zhangxiaoli73:
I have updated code to not transpose the whole tensor, but just the selected subset. This method can use less memory.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -497,6 +487,96 @@ object Recurrent extends ContainerSerializable {
...
+    }
+  }
+
+  /**
+   * select srcIndex subset of the 2-th dimension from src, and copy to dst
+   * @param src
+   * @param srcIndex the index of 2-th dimension from src
+   * @param dst
+   */
+  private[bigdl] def selectCopy[@specialized(Float, Double) T: ClassTag](
```
jason-dai: define batchDim and timeDim as a constant, instead of using 1 or 2 directly.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -497,6 +487,96 @@ object Recurrent extends ContainerSerializable {
...
 
+  /**
+   * set the cells' output and gradInput to recurrent's output and gradInput
+   * to decrease the copy expense.
+   * Copy src tensor to dst tensor along timeDime, default timeDime 2, batchDim 1
+   * @param src
+   * @param dst
+   */
+  private[bigdl] def copy[@specialized(Float, Double) T: ClassTag](
+    src: ArrayBuffer[Tensor[T]], dst: Tensor[T], offset: Int): Unit = {
```
jason-dai: why do we need the offset parameter?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
zhangxiaoli73: ok

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
zhangxiaoli73: I just keep the original interface, from the whole code, maybe it can be deleted.

zhangxiaoli73:
jekins pass

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -491,12 +478,106 @@ class Recurrent[T : ClassTag](var batchNormParams: BatchNormParams[T] = null)
 }
 
 object Recurrent extends ContainerSerializable {
+
+  private val batchDim = 1
```
qiuxin2012: Seems duplicated in both object and class.

zhangxiaoli73: changed

PR #1559, opened 2017-09-13
Add assert, image decoders, parseExample operation

yangw1234:

Add assert, image decoders, parseExample operation

## How was this patch tested?

unit tests


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/Assert.scala
```diff
@@ -0,0 +1,38 @@
...
+import com.google.protobuf.ByteString
+import com.intel.analytics.bigdl.nn.abstractnn.Activity
+import com.intel.analytics.bigdl.tensor.Tensor
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+import com.intel.analytics.bigdl.utils.Table
+
+import scala.reflect.ClassTag
+
+
+class Assert[T: ClassTag]()(implicit ev: TensorNumeric[T]) extends Operation[Table, Activity, T] {
```
yiheng: Please add some document for the operation

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/tf/Const.scala
```diff
@@ -29,8 +29,8 @@ private[bigdl] trait WithoutInput
  * @param value the constant tensor to be returned in forward
  */
 @SerialVersionUID(-4008935551091949324L)
-private[bigdl] class Const[T: ClassTag](value: Tensor[T])(implicit ev: TensorNumeric[T])
-  extends AbstractModule[Activity, Tensor[T], T] with WithoutInput {
+private[bigdl] class Const[T: ClassTag](value: Tensor[_])(implicit ev: TensorNumeric[T])
```
yiheng: Is it conflicted with https://github.com/intel-analytics/BigDL/pull/1569/files#diff-e9e9bf4d715a7a08d0391cc28feede35 ?

yangw1234:
pr validation pass

PR #1561, opened 2017-09-14
[WIP] Use MKL method to implement a 'lazy' gradient zero

frankfzw:

1. Remove `tensor.fill(0)` when the `zeroGradParameters` is called. Instead, only a flag named  `zeroGradFlag` is set to be `true` and then this method return instantly.
2. The clearance of gradient parameter is finished in `accGradParameters`. The different MKL math functions are called with different parameters according to `zeroGradFlag`.
3. `zeroGradFlag` is set to `false` after each `accGradParameters`.
3. Add corresponding unit tests or test cases.

## How was this patch tested?

unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Linear.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Linear.scala
@@ -141,20 +141,38 @@ class Linear[T: ClassTag](
 
     if (input.dim() == 1) {
       if (scaleW != 0) {
-        gradWeight.addr(ev.fromType[Double](scaleW), gradOutput, input)
+        if (zeroGradFlag) {
+          gradWeight.addr(ev.fromType[Double](0.0), gradOutput, ev.fromType[Double](scaleW), input)
+        } else {
+          gradWeight.addr(ev.fromType[Double](scaleW), gradOutput, input)
+        }
       }
 
       if (withBias && scaleB != 0) {
-        gradBias.add(ev.fromType[Double](scaleB), gradOutput)
+        if (zeroGradFlag) {
+          gradBias.mul(gradOutput, ev.fromType[Double](scaleB))
+        } else {
+          gradBias.add(ev.fromType[Double](scaleB), gradOutput)
+        }
       }
     }
     else if (input.dim() == 2) {
       if (scaleW != 0) {
-        gradWeight.addmm(ev.fromType[Double](scaleW), gradOutput.t, input)
+        if (zeroGradFlag) {
+          gradWeight.addmm(ev.fromType[Double](0.0),
+            ev.fromType[Double](scaleW), gradOutput.t, input)
+        } else {
+          gradWeight.addmm(ev.fromType[Double](scaleW), gradOutput.t, input)
+        }
       }
 
       if (withBias && scaleB != 0) {
-        gradBias.addmv(ev.fromType[Double](scaleB), gradOutput.t, addBuffer)
+        if (zeroGradFlag) {
+          gradBias.addmv(ev.fromType[Double](0.0),
+            ev.fromType[Double](scaleB), gradOutput.t, addBuffer)
+        } else {
+          gradBias.addmv(ev.fromType[Double](scaleB), gradOutput.t, addBuffer)
+        }
       }
     }
 
@@ -164,6 +182,7 @@ class Linear[T: ClassTag](
     if (null != bRegularizer && scaleB != 0) {
       bRegularizer.accRegularization(bias, gradBias, scaleB)
     }
+    zeroGradFlag = false
   }
 
   override def updateParameters(learningRate: T): Unit = {
@@ -173,11 +192,10 @@ class Linear[T: ClassTag](
 
   override def zeroGradParameters(): Unit = {
     gradWeight.resize(outputSize, inputSize)
-    gradWeight.zero()
     if (withBias) {
       gradBias.resize(outputSize)
-      gradBias.zero()
     }
+    zeroGradFlag = true
   }
 
   override def clearState() : this.type = {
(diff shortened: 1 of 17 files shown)


frankfzw:
@yiheng Please take a look at this.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialConvolution.scala
```diff
@@ -793,13 +813,22 @@ class SpatialConvolution[T: ClassTag](
         val sWFloat = ev.toType[Float](scaleW)
         val sBFloat = ev.toType[Float](scaleB)
         val gradBFloat = gradBias.asInstanceOf[Tensor[Float]]
+        val update = if (zeroGradFlag) {
```
yiheng: do not use function object in computing intensive code

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialConvolution.scala
```diff
@@ -471,9 +471,14 @@ class SpatialConvolution[T: ClassTag](
       val gradView = gradWeightMMInBatch.view(batchSize,
         nOutputPlane * nInputPlane * kernelH * kernelW / nGroup).t
       val grad = gradWeight.view(nOutputPlane * nInputPlane * kernelH * kernelW / nGroup)
-      grad.addmv(ev.fromType(1.0), ev.fromType(1.0), gradView, onesBatch)
+      val beta = if (zeroGradFlag) {
+        ev.fromType(0.0)
```
yiheng: ev.zero

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialConvolution.scala
```diff
@@ -471,9 +471,14 @@ class SpatialConvolution[T: ClassTag](
...
+      } else {
+        ev.fromType(1.0)
```
yiheng: ev.one

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialConvolution.scala
```diff
@@ -471,9 +471,14 @@ class SpatialConvolution[T: ClassTag](
...
+      }
+      grad.addmv(beta, ev.fromType(1.0), gradView, onesBatch)
       if (withBias) {
-        gradBias.addmv(ev.fromType(1.0), ev.fromType(1.0), gradientBiasMT.t, onesBatch)
+        gradBias.addmv(beta, ev.fromType(1.0), gradientBiasMT.t, onesBatch)
```
yiheng: ev.one

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Linear.scala
```diff
@@ -141,20 +141,38 @@ class Linear[T: ClassTag](
...
+            ev.fromType[Double](scaleW), gradOutput.t, input)
+        } else {
+          gradWeight.addmm(ev.fromType[Double](scaleW), gradOutput.t, input)
+        }
       }
 
       if (withBias && scaleB != 0) {
-        gradBias.addmv(ev.fromType[Double](scaleB), gradOutput.t, addBuffer)
+        if (zeroGradFlag) {
+          gradBias.addmv(ev.fromType[Double](0.0),
```
yiheng: ev.zero

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Linear.scala
```diff
@@ -141,20 +141,38 @@ class Linear[T: ClassTag](
...
+        } else {
+          gradBias.add(ev.fromType[Double](scaleB), gradOutput)
+        }
       }
     }
     else if (input.dim() == 2) {
       if (scaleW != 0) {
-        gradWeight.addmm(ev.fromType[Double](scaleW), gradOutput.t, input)
+        if (zeroGradFlag) {
+          gradWeight.addmm(ev.fromType[Double](0.0),
```
yiheng: ev.zero

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Linear.scala
```diff
@@ -141,20 +141,38 @@ class Linear[T: ClassTag](
 
     if (input.dim() == 1) {
       if (scaleW != 0) {
-        gradWeight.addr(ev.fromType[Double](scaleW), gradOutput, input)
+        if (zeroGradFlag) {
+          gradWeight.addr(ev.fromType[Double](0.0), gradOutput, ev.fromType[Double](scaleW), input)
```
yiheng: ev.zero

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Linear.scala
```diff
@@ -141,20 +141,38 @@ class Linear[T: ClassTag](
...
+        } else {
+          gradBias.add(ev.fromType[Double](scaleB), gradOutput)
+        }
       }
     }
     else if (input.dim() == 2) {
       if (scaleW != 0) {
-        gradWeight.addmm(ev.fromType[Double](scaleW), gradOutput.t, input)
+        if (zeroGradFlag) {
+          gradWeight.addmm(ev.zero,
```
frankfzw: It seems this operation slows down the backward of Linear @yiheng

yiheng: why, can you dig deeper? These two methods call the same method with only one parameter difference

yiheng: You may need to run some micro-benchmark on this method

PR #1564, opened 2017-09-15
support multi output in graph follow up

yangw1234:

1. Use new multi output node graph api in tensorflow loader
2. Support the scenario where some output tensor are not connected to any following nodes.
3. Fix match graph bug:
    We cannot connect bigdl module node when we traversing and matching the tf graph, for that way we 
    cannot enforce the same input order. We should connect bigdl module node when all the tf graph 
    nodes are converted. 
 

## How was this patch tested?

existing tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DirectedGraph.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DirectedGraph.scala
@@ -323,6 +323,13 @@ private[bigdl] class Edge private (val fromIndex: Option[Int]) extends Serializa
   override def toString: String = {
     s"Edge(fromIndex: $fromIndex)"
   }
+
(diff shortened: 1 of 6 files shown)


yangw1234:
pr validation pass

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DirectedGraph.scala
```diff
@@ -323,6 +323,13 @@ private[bigdl] class Edge private (val fromIndex: Option[Int]) extends Serializa
   override def toString: String = {
     s"Edge(fromIndex: $fromIndex)"
   }
+
+  def newInstance(): Edge = {
```
yiheng: document

yiheng:
LGTM

yangw1234:
pr validation pass

PR #1572, opened 2017-09-20
Support Tensorflow Preprocessing

yangw1234:

1. Support Tensorflow Preprocessing
2. Change the behavior of selecting on 1-d tensor, return a scalar tensor instead of a 1-d tensor.
Normally selecting on a n-d tensor would return a (n-1)-d tensor.
Originally in BigDL, however, selecting on a 1-d tensor would also return a 1-d tensor, as we change the select operation to a narrow operation.
Since we support scalar now, selecting on a 1-d tensor should return a scalar.

## How was this patch tested?

unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
@@ -297,8 +297,7 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
   /**
    * Execution plan
    */
-  private val forwardNodes = backGraph.DFS
-    .filterNot(_.element.isInstanceOf[ControlDependency[T]]).toArray
+  private val forwardNodes = backGraph.DFS.toArray
   private val forwardScheduler = new Scheduler(
     forwardNodes.filter(_.prevNodes.length == 0),
     Seq(dummyOutput)
@@ -347,7 +346,8 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
     require(forwardNodes.map(_.element.getName()).distinct.length == forwardNodes.length,
       "the name of node in the graph should be unique")
     val roots = forwardNodes.filter(_.prevNodes.size == 0)
-      .filter(node => !node.element.isInstanceOf[WithoutInput])
+      .filter(node => !node.element.isInstanceOf[WithoutInput]
+        && !node.element.isInstanceOf[ControlDependency[_]])
     require(roots.size == inputs.length,
       s"There're ${inputs.length} inputs, but graph has ${roots.size} roots")
     inputs.foreach(n =>
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
@@ -370,7 +370,7 @@ private[tensor] class DenseTensor[@specialized(Float, Double) T: ClassTag](
     val _dimension = dim - 1
     val _sliceIndex = index - 1
 
-    require(this.nDimension > 0, "empty or scalar tensor")
+    require(this.nDimension > 0, "empty or scalar tensor cannot be selected")
     val result = DenseTensor.newWithTensor(this)
     DenseTensor.select(result, null, _dimension, _sliceIndex)
     result
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
@@ -37,9 +37,7 @@ private[tensor] class DenseTensor[@specialized(Float, Double) T: ClassTag](
 
   override def isEmpty: Boolean = this.storage() == null || this.storage().length() == 0
 
-  override def isScalar: Boolean =
-    this.nDimension == 0 &&
-      this._storage.length() == 1
+  override def isScalar: Boolean = !this.isEmpty && this.nDimension == 0
 
   override def storage(): Storage[T] = _storage
 
@@ -372,14 +370,10 @@ private[tensor] class DenseTensor[@specialized(Float, Double) T: ClassTag](
     val _dimension = dim - 1
     val _sliceIndex = index - 1
 
-    if (this.nDimension > 1) {
-      val result = DenseTensor.newWithTensor(this)
-      DenseTensor.select(result, null, _dimension, _sliceIndex)
-      result
-    } else {
-      require(this.nDimension == 1, "empty tensor")
-      this.narrow(1, index, 1)
-    }
+    require(this.nDimension > 0, "empty or scalar tensor")
+    val result = DenseTensor.newWithTensor(this)
+    DenseTensor.select(result, null, _dimension, _sliceIndex)
+    result
   }
 
   override def clone(): Tensor[T] = {
@@ -464,13 +458,9 @@ private[tensor] class DenseTensor[@specialized(Float, Double) T: ClassTag](
     require(_index >= 0 && _index < this._size(0),
       s"out of range, ${_index}: 0 to ${this._size(0)}")
 
-    if (this.nDimension == 1) {
-      this.narrow(1, index, 1)
-    } else {
-      val result = DenseTensor.newWithTensor(this)
-      DenseTensor.select(result, null, 0, _index)
-      result
-    }
+    val result = DenseTensor.newWithTensor(this)
+    DenseTensor.select(result, null, 0, _index)
+    result
   }
 
   override def apply(table: Table): Tensor[T] = {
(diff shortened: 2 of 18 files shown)


yangw1234:
pr validation pass

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -297,8 +297,7 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
   /**
    * Execution plan
    */
-  private val forwardNodes = backGraph.DFS
-    .filterNot(_.element.isInstanceOf[ControlDependency[T]]).toArray
+  private val forwardNodes = backGraph.DFS.toArray
```
yiheng: do we need to update forwardSchduler, check roots and getForwardExecutions? saveGraphTopology?

yiheng: why remove the control dependency filter here?

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
```diff
@@ -372,14 +370,10 @@ private[tensor] class DenseTensor[@specialized(Float, Double) T: ClassTag](
...
 
-    if (this.nDimension > 1) {
-      val result = DenseTensor.newWithTensor(this)
-      DenseTensor.select(result, null, _dimension, _sliceIndex)
-      result
-    } else {
-      require(this.nDimension == 1, "empty tensor")
-      this.narrow(1, index, 1)
-    }
+    require(this.nDimension > 0, "empty or scalar tensor")
```
yiheng: empty or scalar tensor cannot be selected

yangw1234: Yes, select will decrease the nDimension by 1. It would not make sense to select a scalar or empty tensor as their nDimension is already 0.

spark/dl/src/test/scala/com/intel/analytics/bigdl/nn/ops/RankSpec.scala
```diff
@@ -24,7 +24,7 @@ class RankSpec extends FlatSpec with Matchers {
     val input =
         Tensor[Float](T(1f, 2f, 2f))
 
-    val expectOutput = Tensor[Int](T(1))
+    val expectOutput = Tensor[Int](Array(1), Array[Int]())
```
yiheng: should we create some Scalar factory method for user to use. So user can write something like

```scala
val output = Scala(1)
```

spark/dl/src/test/scala/com/intel/analytics/bigdl/tensor/DenseTensorSpec.scala
```diff
@@ -886,22 +899,28 @@ class DenseTensorSpec extends FlatSpec with Matchers {
...
   "Tensor add" should "catch exception when broadcasting size not match 2" in {
     val t1 = Tensor[Double](T(T(1, 2, 3), T(1, 2, 3), T(1, 2, 3)))
     val t2 = Tensor[Double](T(T(2, 5, 3), T(3, 6, 4)))
     intercept[IllegalArgumentException] {
-      t2.add(t1) should be (Tensor[Double](T(T(3, 7, 6), T(4, 8, 7))))
+      t2.add(t1) should be(Tensor[Double](T(T(3, 7, 6), T(4, 8, 7))))
     }
   }
+
+  "Select on a Vector " should "be a scalar" in {
```
yiheng: what if select on a vector with size != 1?

yangw1234: Why the size of this vector matters here?

spark/dl/src/test/scala/com/intel/analytics/bigdl/utils/tf/TensorflowLoaderSpec.scala
```diff
@@ -103,6 +104,19 @@ class TensorflowLoaderSpec extends TensorflowSpecHelper{
     System.setProperty("bigdl.enableNHWC", "false")
   }
 
+  "TensorFlow loader" should "convert tensor protobuf to bigdl tensor" in {
```
yangw1234: This is just for debug. I'll remove this.

spark/dl/src/test/scala/com/intel/analytics/bigdl/nn/ops/RankSpec.scala
yangw1234: Do you mean
```scala
Tensor.scalar(1)
```
Or create a scalar object?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/Session.scala
```diff
@@ -29,6 +29,9 @@ import org.apache.spark.SparkContext
 import org.apache.spark.api.java.JavaRDD
 import org.apache.spark.rdd.RDD
 import org.tensorflow.framework.{GraphDef, NodeDef}
+import com.google.protobuf.ByteString
```
yiheng: val sc = SparkContext.getOrCreate()

Should not init a sc in this way. Use Engine.init

yiheng: val optMethod = new SGD[T]()
    opt.setOptimMethod(optMethod).setEndWhen(endWhen)
      .optimize()

Always use a SGD, user cannot change?

yangw1234: I can submit another pr to fix this problem.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
yangw1234: This is a subtle point. If a ControlDependency node's in-degree is zero, then it's children can never be scheduled.

yiheng: why a controlDependency node in-degree is zero?

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
yiheng: I mean change the error message

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/Session.scala
yiheng: why not fix in this PR?

spark/dl/src/test/scala/com/intel/analytics/bigdl/tensor/DenseTensorSpec.scala
yiheng: size of the select method parameter

spark/dl/src/test/scala/com/intel/analytics/bigdl/utils/tf/TensorflowLoaderSpec.scala
yiheng: I think it's fine for a test case

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/Session.scala
```diff
@@ -52,10 +55,357 @@ class BigDLSessionImpl[T: ClassTag](
...
+
+  private val enqueueOp = Set("QueueEnqueueV2", "QueueEnqueueManyV2")
+
   private val (wholeTFGraph, _, _) = TensorflowLoader.buildTFGraph(graph.asJava, null)
 
   private val name2Node = wholeTFGraph.
-    DFS.filter(n => n.element != null).map(node => (node.element.getName, node)).toMap
+    DFS.filter(_.element != null).map(node => (node.element.getName, node)).toMap
+
+  private def tableToSeq(table: Table): Seq[Tensor[T]] = {
```
yiheng: can we move such utils functions to table?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/Session.scala
```diff
@@ -52,10 +55,357 @@ class BigDLSessionImpl[T: ClassTag](
...
+    for (i <- 0 until table.length()) yield {
+      table(i).asInstanceOf[Tensor[T]]
+    }
+  }
+
+  private def seqToTable(tensors: Seq[Tensor[T]]): Table = {
```
yiheng: can we move such utils function to table?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/Session.scala
```diff
@@ -44,18 +47,364 @@ abstract class Session[T: ClassTag] {
...
+      reducedInputSeq.map { tensors =>
+        val output = transformer.forward(tensors.flatten())
+        toTable(output)
+      }
+    } else {
+      Seq(toTable(transformer.forward(T())))
+    }
+  }
+
+  private def toTable(activity: Activity): Table = {
```
yiheng: why not use the toTable in Activity?

yangw1234: if this activity is actually a tensor, toTable will throw an exception

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/Tensor.scala
```diff
@@ -1006,6 +1006,11 @@ object Tensor {
     apply(Storage(matrix.toArray), 1, Array(matrix.numRows, matrix.numCols), strides)
   }
 
+  def scalar[T: ClassTag](value: T)(
```
yiheng: document

yangw1234:
pr validation pass

PR #1574, opened 2017-09-20
fix parse tensor

yangw1234:

fix parse tensor

## How was this patch tested?

unit tests


yiheng:
LGTM

PR #1575, opened 2017-09-20
Support load tf unet

yiheng:
1. add deconv operation
2. add bilinearresize layer
3. tensor add support broadcast

## How was this patch tested?
unit test


spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
```diff
@@ -2382,4 +2392,24 @@ object DenseTensor {
     }
     gauss
   }
+
+  private[tensor] def checkBroadcastDims[@specialized T: ClassTag](tensor: Tensor[T],
```
yangw1234: In tensorflow, it also support more types of broadcasting, like:
```python
a = tf.random_normal((2, 2, 2, 2), dtype=tf.float32)
b = tf.random_normal((1, 2, 1, 1), dtype=tf.float32)

c = a + b
```
It is used in BatchNorm nchw to broadcasting the channel dim.

yiheng:
PR valdiation pass

PR #1582, opened 2017-09-22
refactor predictor

zhangxiaoli73:

1. add batchsize parameter for predict
2. add parameter repeatMemory to appoint whether clone the predict results

## How was this patch tested?
unit test

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
@@ -478,17 +478,24 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag, 
   /**
    * module predict, return the probability distribution
    * @param dataset dataset for prediction
+   * @param batchSize total batchSize for all partitions.
+   *                  if -1, default is 4 * partitionNumber of datatset
+   * @param repeatMemory whether to use same memory for each batch predict results
    */
-  def predict(dataset: RDD[Sample[T]]): RDD[Activity] = {
-    Predictor(this).predict(dataset)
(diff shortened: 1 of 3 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -470,17 +470,24 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
...
    * @param dataset dataset for prediction
+   * @param batchSize total batchSize for all partitions.
+   *                  if -1, default is 4 * partitionNumber of datatset
+   * @param repeatMemory whether to use same memory for each batch predict results
    */
-  def predict(dataset: RDD[Sample[T]]): RDD[Activity] = {
-    Predictor(this).predict(dataset)
+  def predict(dataset: RDD[Sample[T]],
+              batchSize: Int = -1,
+              repeatMemory: Boolean = false): RDD[Activity] = {
```
jason-dai: why false by default?

zhangxiaoli73: In our user group, a user uses rdd.take() to get repeated predict results. And he thinks it is confused for him. Other users may have such confuse too. So I use default "false" to get no repeat results.

jason-dai: ok - change the name to sth like "shareBuffer"

zhangxiaoli73: ok

zhangxiaoli73:
jekins pass

PR #1587, opened 2017-09-25
load from definition

wzhongyuan:
Sometime we have the case to load weight&bias from persisted module for different net structure, this PR is to add this function
Copy params for pre-defined module from persisted module

## How was this patch tested?

Unit test

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/ModuleLoader.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/ModuleLoader.scala
@@ -19,16 +19,26 @@ import java.io._
 
 import scala.collection.JavaConverters._
 import com.google.protobuf.CodedInputStream
+import com.intel.analytics.bigdl.nn.Container
 import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, Activity}
+import com.intel.analytics.bigdl.tensor.Tensor
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
-import com.intel.analytics.bigdl.utils.File
+import com.intel.analytics.bigdl.utils.{File, Table}
 import serialization.Bigdl.BigDLModule
 
+import scala.collection.mutable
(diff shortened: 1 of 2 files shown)


wzhongyuan:
@i8run  please help review as it may also impact quantization

wzhongyuan:
@i8run  Thanks for the review, PR validation passed 
http://10.239.47.211:18888/job/BigDL-PR-Validation/1896/

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/ModuleLoader.scala
```diff
@@ -39,10 +49,97 @@ object ModuleLoader {
...
+  }
+
+  private def copyParam[T : ClassTag](params : Table,
+                                      copyParams : Table, paraName : String) : Unit = {
+    if (params.contains(paraName)) {
+      // this is for quantization tensors where the weight might be an array
+      if (copyParams.get(paraName).get
+        .isInstanceOf[Array[Tensor[T]]]) {
+        require(params.get(paraName).get
+          .isInstanceOf[Array[Tensor[T]]], "param type mismatch!")
```
jason-dai: Add more info to the error message

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/ModuleLoader.scala
```diff
@@ -39,10 +49,97 @@ object ModuleLoader {
...
+  }
+
+  private def copyParams[T : ClassTag](definition : AbstractModule[Activity, Activity, T],
+                                     mirror : AbstractModule[Activity, Activity, T],
+                                      layers : mutable.HashSet[String]) : Unit = {
+    val parameterTable = definition.getParametersTable()
+    val copiedParameterTable = mirror.getParametersTable()
+    layers.foreach(name => {
+      if (parameterTable.contains(name)) {
+        require(copiedParameterTable.contains(name), s"$name does not exist in loaded module")
```
jason-dai: Add more info to the error message

PR #1591, opened 2017-09-25
Add multiple cell

dding3:
Gives a multiCell layer which enable stack several rnn cells, similar to multirnncell in tensorflow

## How was this patch tested?
ut


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/MultiCell.scala
```diff
@@ -0,0 +1,142 @@
...
+ */
+class MultiCell[T : ClassTag](val cells: Array[Cell[T]])(implicit ev: TensorNumeric[T])
+  extends Cell[T](
+    hiddensShape = cells.last.hiddensShape,
+    regularizers = cells.flatMap(_.regularizers)) {
+  // inputDim and hidDim must be the same with Recurrent
+  private val inputDim = Recurrent.inputDim
+  private val hidDim = Recurrent.hidDim
+
+  override var preTopology: AbstractModule[Activity, Activity, T] = null
```
zhangxiaoli73: In recurrent layer, preTopology = Sequential[T]().add(preTopology).add(batchNormalization())
You set preTopology=null, how can you support batchNorm in MultiCell?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/MultiCell.scala
```diff
@@ -0,0 +1,142 @@
...
+    while (i < cells.length) {
+      result(hidDim) = states(i)
+
+      if (cells(i).preTopology != null) {
+//        val inputTmp = result(inputDim).asInstanceOf[Tensor[T]].clone()
+//        val sizes = 1 +: inputTmp.size()
+//        inputTmp.resize(sizes)
+//        val outputTmp = cells(i).preTopology.forward(inputTmp).toTensor[T]
+//        result(inputDim) = outputTmp.select(1, 1)
+        result(inputDim) = cells(i).preTopology.forward(result(inputDim)).toTensor[T]
```
zhangxiaoli73: before, preTopology will be only caculated once, what about the performance of your change?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -146,6 +148,22 @@ class Recurrent[T : ClassTag](var batchNormParams: BatchNormParams[T] = null)
       cells.head.hidResize(hidden = hidden, batchSize = batchSize, imageSize)
       gradHidden = hidden
     }
+  }
+
+  private def initHiddens(sizes: Array[Int]): Unit = {
+    val batchSize = sizes(0)
```
zhangxiaoli73: sizes(batchDim - 1)

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/MultiCell.scala
```diff
@@ -0,0 +1,142 @@
...
+    var error = T()
+    error(inputDim) = gradOutput(inputDim)
+
+    val nextInput = T()
+    while (i >= 1) {
+      val input0: Tensor[T] = if (i > 1) {
+        cells(i - 2).output.toTable(inputDim)
+      } else input(inputDim)
+      nextInput(inputDim) = if (cells(i - 1).preTopology != null) {
+        cells(i - 1).preTopology.forward(input0)
```
zhangxiaoli73: Why caculate  preTopology forward again? I think you can store the results in forward, and use here.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -280,7 +340,7 @@ class Recurrent[T : ClassTag](var batchNormParams: BatchNormParams[T] = null)
     while (i >= 1) {
       currentGradOutput(inputDim) = Recurrent.selectCopy(gradOutput, i, gradBuffer)
       _input(hidDim) = if (i > 1) cells(i - 2).output.toTable(hidDim)
-        else hidden
+      else hidden
```
zhangxiaoli73: also "else if (initState == null) hidden else initState"

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -316,7 +376,7 @@ class Recurrent[T : ClassTag](var batchNormParams: BatchNormParams[T] = null)
...
```
zhangxiaoli73: also "else if (initState == null) hidden else initState"

docs/docs/APIGuide/Layers/Recurrent-Layers.md
```diff
@@ -175,6 +175,9 @@ With RecurrentDecoder, inputsize and hiddensize of the cell must be the same.
 
 Different types of rnn cells can be added using add() function.
 
+RecurrentDecoder supports returning state and cell of its rnn cells at last time step by using getState.
+If contained cell is MultiCell, use getStates/setStates instead.
```
jason-dai: Can we unify the interface?

docs/docs/APIGuide/Layers/Recurrent-Layers.md
```diff
@@ -9,7 +9,7 @@ val module = Recurrent()
 module = Recurrent()
 ```
 
-Recurrent module is a container of rnn cells. Different types of rnn cells can be added using add() function.  
+Recurrent module is a container of rnn cells. Different types of rnn cells can be added using add() function. Don't support multiCell for performance issue. Use several Recurrent(cell) instead.
```
jason-dai: How do you enforce that?

docs/docs/APIGuide/Layers/Recurrent-Layers.md
```diff
@@ -1558,4 +1561,232 @@ array([[[[ 0.1       ,  0.2       ],
         [[ 0.30000001,  0.40000001],
          [ 0.30000001,  0.40000001]]]], dtype=float32)
 ```
+---
+
+## MultiCell ##
```
jason-dai: MultiRNNCell?

docs/docs/APIGuide/Layers/Recurrent-Layers.md
```diff
@@ -1558,4 +1561,232 @@ array([[[[ 0.1       ,  0.2       ],
...
+```python
+model = MultiCell(cells = multiCells)
+```
+
+A cell that stack multiple simple rnn cells. Only works with RecurrentDecoder. If you want to
+stack multiple cells with Recurrent. Use Sequential().add(Recurrent(cell)).add(Recurrent(cell))... instead
+
+Parameters:
+
+* `cells` list of RNNCells that will be composed in this order.
```
jason-dai: Can it be a list of LSTM cells? Or a list of cells mixing RNN/LSTM/GRU etc.?

docs/docs/APIGuide/Layers/Recurrent-Layers.md
dding3: get/setStates use an array of state while get/setState use a single state. To unify the interface, one thought is using interface of get/setStates, in this way for non-multirnn cell, we have to wrap a state to an array before pass it to get/setStates. I am not sure if it's straight forward enough for non-multirnn cell.

docs/docs/APIGuide/Layers/Recurrent-Layers.md
dding3: In recurrent.add(), it will check the cell type, if it's multirnncell, it will throw exception and ask user use sequential().add(Recurrent.add(cell))... instead

docs/docs/APIGuide/Layers/Recurrent-Layers.md
dding3: It can be a list of different rnn cells, if output of last cell can be can be processed by the following cell.

docs/docs/APIGuide/Layers/Recurrent-Layers.md
```diff
@@ -1558,4 +1561,232 @@ array([[[[ 0.1       ,  0.2       ],
         [[ 0.30000001,  0.40000001],
          [ 0.30000001,  0.40000001]]]], dtype=float32)
 ```
+---
+
+## MultiRNNCell ##
+
+**Scala:**
+```scala
+val model = MultiRNNCell(cells = multiRNNCells)
```
jason-dai: type of cells?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/LSTMPeephole.scala
```diff
@@ -76,11 +76,9 @@ class LSTMPeephole[T : ClassTag] (
         .add(SelectTable(1))
         .add(NarrowTable(2, 2)))
 
-  override def preTopology: AbstractModule[Activity, Activity, T] =
-    Sequential()
-    .add(Dropout(p))
-    .add(TimeDistributed(Linear(inputSize, hiddenSize * 4, wRegularizer = wRegularizer,
-      bRegularizer = bRegularizer)))
+  override var preTopology: TensorModule[T] =
+    Linear(inputSize, hiddenSize * 4, wRegularizer = wRegularizer,
```
jason-dai: if p is non zero, the pre-topology should be null (just as like LSTM)

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -73,10 +73,15 @@ class Recurrent[T : ClassTag](var batchNormParams: BatchNormParams[T] = null)
    */
   override def add(module: AbstractModule[_ <: Activity, _ <: Activity, T]): Recurrent.this.type = {
     require(module.isInstanceOf[Cell[T]],
-      "Recurrent: contained module should be Cell type")
+      "Recurrent: added module should be Cell type!")
+    require(!module.isInstanceOf[MultiRNNCell[T]],
+      "Recurrent: added module cannot be MultiRNNCell due to performance issue," +
```
jason-dai: remove due to perf issue

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
```diff
@@ -41,26 +41,79 @@ class RecurrentDecoder[T : ClassTag](seqLength: Int)
   (implicit ev: TensorNumeric[T]) extends Recurrent[T] {
 
   times = seqLength
-  private val newInput = Tensor[T]()
+  private var initStates: Array[Activity] = null
+  private var states: Array[Activity] = null
+  private var gradStates: Array[Activity] = null
+
+  def setStates(states: Array[Activity]): Unit = {
```
jason-dai: Unify state & states using a table?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
```diff
@@ -41,26 +41,79 @@ class RecurrentDecoder[T : ClassTag](seqLength: Int)
...
+    require(topology.isInstanceOf[MultiRNNCell[T]], "setStates only support for MultiCell")
+    initStates = states
+  }
+
+  def getStates(): Array[Activity] = {
+    require(topology.isInstanceOf[MultiRNNCell[T]], "getStates only support for MultiCell")
+    states
+  }
+
+  private def initHiddens(sizes: Array[Int]): Unit = {
```
jason-dai: unify hidden and hiddens using a table?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/LSTMPeephole.scala
dding3: I am worried if we set pre-topology is null, there would be a perf downgrade(LSTM didn't have this problem because its implementation is different). Change here is moving dropout from before linear to after linear, like update dropout -> linear to linear -> dropout. They may be not exactly the same, but both of them randomly drop values which will be feeded to next layer.

docs/docs/APIGuide/Layers/Recurrent-Layers.md
dding3: Updated

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
```diff
@@ -41,26 +42,84 @@ class RecurrentDecoder[T : ClassTag](seqLength: Int)
...
+    batchSize = input.size(batchDim)
     val featureSizes = outputSize.drop(1)
     output.resize(Array(batchSize, times) ++ featureSizes)
-    // Clone N modules along the sequence dimension.
-    extend(featureSizes)
-    if (preTopology != null) newInput.resize(output.size())
-    else outputCell.resize(output.size())
+    if (preTopology != null) {
+      val sizes = output.size()
+      sizes(2) = hiddenSize * 4
```
jason-dai: why hiddenSize * 4?

dding3: The preTopology gathered all linear computing with input. and there is 4 linears (i2g, f2g, c2g, o2g). So the expected hiddensize after gathering is hiddenSize * 4

jason-dai: you cannot assume LSTM here; what about RNN, GRU, etc?

docs/docs/APIGuide/Layers/Recurrent-Layers.md
jason-dai: You need to specify that cells should be an array of Cell

docs/docs/APIGuide/Layers/Recurrent-Layers.md
```diff
@@ -1558,5 +1558,233 @@ array([[[[ 0.1       ,  0.2       ],
...
+
+```
+**Python:**
+```python
+model = MultiRNNCell(cells = multiRNNCells)
```
jason-dai: You need to specify that cells should be an array of Cell

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/MultiRNNCell.scala
```diff
@@ -0,0 +1,165 @@
...
+      i += 1
+    }
+
+    result(hidDim) = outputStates
+    this.output = result
+    output
+  }
+
+  override def updateGradInput(input: Table, gradOutput: Table): Table = {
+    var i = cells.length
```
jason-dai: better to set i to cells.length-1, so as to align with the updateOutput implemetation

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/MultiRNNCell.scala
```diff
@@ -0,0 +1,165 @@
...
+      i -= 1
+    }
+
+    this.gradInput = error
+    gradInput(hidDim) = outputGradStates
+    gradInput
+  }
+
+  override def accGradParameters(input: Table, gradOutput: Table): Unit = {
+    var i = cells.length
```
jason-dai: better to set i to cells.length-1, so as to align with the updateOutput implemetation

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/MultiRNNCell.scala
```diff
@@ -0,0 +1,165 @@
...
+      nextInput(hidDim) = states(i - 1)
+      error(hidDim) = gradStates(i - 1)
+      cells(i - 1).accGradParameters(nextInput, error)
+      error(inputDim) = cells(i - 1).gradInput.toTable(inputDim)
+      i -= 1
+    }
+  }
+
+  override def backward(input: Table, gradOutput: Table): Table = {
+    var i = cells.length
```
jason-dai: better to set i to cells.length-1, so as to align with the updateOutput implemetation

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -146,6 +149,31 @@ class Recurrent[T : ClassTag](var batchNormParams: BatchNormParams[T] = null)
...
+//      while (i < multiCells.size) {
+//        //        hiddenStates(i) = multiCells(i).hidResize(null, batchSize, stepSizes)
+//        //        gradHiddenStates(i) = multiCells(i).hidResize(null, batchSize, stepSizes)
+//        hidden.toTable(i) = multiCells(i).hidResize(null, batchSize, stepSizes)
+//        gradHidden.toTable(i) = multiCells(i).hidResize(null, batchSize, stepSizes)
+//        i += 1
+//      }
+//    } else initHidden(sizes)
+//  }
+
```
jason-dai: Delete these

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
```diff
@@ -44,10 +44,31 @@ class RecurrentDecoder[T : ClassTag](val seqLength: Int)
 
   times = seqLength
 
+  override def initHidden(sizes: Array[Int]): Unit = {
+    val stepSizes = sizes
+
+    if (containMultiRNNCell) {
+      if (hidden == null) {
```
jason-dai: I think you should implement these logic inside MultiRNNCell.hidResize

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
```diff
@@ -217,6 +236,13 @@ class RecurrentDecoder[T : ClassTag](val seqLength: Int)
     state.map(_.hashCode()).foldLeft(0)((a, b) => 31 * a + b)
   }
 
+  private def cloneStates(states: Table, res: Table): Unit = {
```
jason-dai: where is it used?

spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDL.scala
```diff
@@ -1991,14 +1991,40 @@ class PythonBigDL[T: ClassTag](implicit ev: TensorNumeric[T]) extends Serializab
     layer.setInitMethod(weightInitMethod, biasInitMethod)
   }
 
-  def getHiddenStates(rec: Recurrent[T]): JList[JTensor] = {
+  def getHiddenStates(rec: Recurrent[T]): JList[JList[JTensor]] = {
```
jason-dai: This is too complex; instead of doing this, maybe we can have automatic mapping of Table to JList?

zhichao-li: @jason-dai  issue created for this: https://github.com/intel-analytics/BigDL/issues/1937

docs/docs/APIGuide/Layers/Recurrent-Layers.md
```diff
@@ -1558,5 +1558,233 @@ array([[[[ 0.1       ,  0.2       ],
...
+```
+
+A cell that stack multiple rnn cells(simpleRNN/LSTM/LSTMPeephole/GRU/ConvLSTMPeephole/ConvLSTMPeephole3D).
+Only works with RecurrentDecoder. If you want to stack multiple cells with Recurrent. Use Sequential().add(Recurrent(cell)).add(Recurrent(cell))... instead
+
+Parameters:
+
+* `cells` list of RNNCells that will be composed in this order.
```
jason-dai: RNNCell?

pyspark/bigdl/nn/layer.py
```diff
@@ -1326,18 +1326,8 @@ def get_hidden_state(self):
         
         :return: list of hidden state and cell
         """
-        state = callBigDlFunc(self.bigdl_type, "getHiddenState", self.value)
-        for idx, tensor in enumerate(state):
-            state[idx] = tensor.to_ndarray()
-
-        return state
-
-    def set_hidden_state(self, states):
```
jason-dai: Need to update the doc too

dding3: Updated

jason-dai:
LGTM

dding3:
Thanks for your review! Jekins passed.

PR #1593, opened 2017-09-26
Refine tensorflow loader code strcuture

yiheng:
This PR refine tensorflow load relate code. Make it easy to add new tensorflow operation load code.
## How was this patch tested?
unit test


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/loaders/Adapter.scala
```diff
@@ -0,0 +1,108 @@
...
+import com.intel.analytics.bigdl.utils.{T, Table}
+
+import scala.reflect.ClassTag
+
+/**
+ * Sometimes parameters pass as tensor from some previous nodes in graph. So we can't construct
+ * module at module define time. In such case, you can use the AdapterOperation to wrapper the
+ * module and create it in runtime.
+ *
+ * Please note you must guarantee the input parameter won't change each time.
```
yangw1234: Should we check this in our code? As it will be very hard to debug if the input parameter do change.

yiheng: fixed

PR #1594, opened 2017-09-27
sparse tensor

### LINKED ISSUE #979: Support Sparse Input/Tensor
Neural nets are not only used for dense inputs such as image and audio. Sparse inputs should also be supported. 

Now there's a trend to use NN for recommendation (e.g. wide&deep recommendation, ncf, rnn-based recommender). Many recommender works on dataframes (columns of features) and in such cases, sparse inputs are quite common because of heavily used one-hot encoding. 

We need to support Sparse Tensor as input for mostly used layers.  As for which layers we need to support for Sparse Input, I think for now we can just support the layers that are used in Wide&Deep recommender.  Then we can add support for new layers per request from customers.

qiuxin2012:

SparseTensor and coomm coomv

To describe an SparseTensor, we need indices, values, and shape:
Indices means non-zero elements' indices; values means the values of the non-zero elements;
Shape means the dense shape of this SparseTensor.

For example, an 2D 3x4 DenseTensor:
 1, 0, 0, 4
 0, 2, 0, 0
 0, 0, 3, 0

 it's sparse representation should be
 indices(0) = Array(0, 0, 1, 2)
 indices(1) = Array(0, 3, 1, 2)
 values     = Array(1, 4, 2, 3)
 shape      = Array(3, 4)

## How was this patch tested?

unit tests,

fixed #979


spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
```diff
@@ -2478,19 +2478,46 @@ object DenseTensor {
...
-    while(i >= 0) {
+    while (i >= 0) {
       size(i) = longTensor.size(i + 1)
       i -= 1
     }
 
     size
   }
+
+  def apply[T: ClassTag](
```
zhangxiaoli73: add some java doc

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/SparseTensor.scala
```diff
@@ -0,0 +1,1067 @@
...
+    }
+  }
+
+  override def hashCode(): Int = {
+    val state = Seq(_indices, _values, _storageOffset, _nElement, _shape, nDimension)
+    state.map(_.hashCode()).foldLeft(0)((a, b) => 31 * a + b)
+  }
+
+  override def isEmpty: Boolean = {
+    throw new UnsupportedOperationException(s"Unimplemented")
```
zhangxiaoli73: add class name

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/SparseTensor.scala
```diff
@@ -0,0 +1,1067 @@
...
+  }
+
+  override def cast[D: ClassTag](
+        castTensor: Tensor[D])(implicit ev: TensorNumeric[D]): Tensor[D] = {
+    throw new UnsupportedOperationException(s"Unimplemented")
+  }
+}
+
+object SparseTensor{
```
zhangxiaoli73: private?

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/Tensor.scala
```diff
@@ -1112,4 +1115,123 @@ object Tensor {
...
+   * @param sparseTensor a sparse tensor
+   * @param res if defined, override to res, else will generate a new tensor.
+   * @param ev
+   * @tparam T
+   * @return a DenseTensor.
+   */
+  def dense[T: ClassTag](
+        sparseTensor: Tensor[T],
+        res: Tensor[T] = null)(implicit ev: TensorNumeric[T]): Tensor[T] = {
+    DenseTensor(sparseTensor, res)
```
zhangxiaoli73: res.dense(sparseTensor)

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
```diff
@@ -2481,19 +2481,46 @@ object DenseTensor {
...
+    while (i >= 0) {
       size(i) = longTensor.size(i + 1)
       i -= 1
     }
 
     size
   }
+
+  private[tensor] def apply[T: ClassTag](
+        sparseTensor: Tensor[T],
```
zhangxiaoli73: SparseTensor

qiuxin2012:
jenkins passed

PR #1606, opened 2017-09-28
add new DLEstimator, DLClassifier, DLmodel, DLClassifierModel doc into API Guide

megaSpoon:

add new DLEstimator, DLClassifier, DLmodel, DLClassiferModel doc into API Guide
remove the old use SparkML doc

python examples will be finished and added at the end of this week.

Pls leave comments if you found typo or inappropriate words of choice.

@jason-dai @hhbyyh


docs/docs/APIGuide/MLPipeline/DLEstimator/DLClassifier.md
```diff
@@ -0,0 +1,71 @@
+## **DLClassifier**
```
jason-dai: Is this redundant?

jason-dai:
We should emphasize that DLModel extends mllib Transformer. The we can use an example to show how one can wrap a pre-trained BigDL model in DLModel or DLClassifierModel, and then use it to predict the result using Spark ML pipeline.

docs/mkdocs.yml
```diff
@@ -72,6 +72,9 @@ pages:
     - Optimization Algorithms: APIGuide/Optimizers/Optim-Methods.md
   - Triggers: APIGuide/Triggers.md
   - Metrics:  APIGuide/Metrics.md
+  - Spark ML:
```
jason-dai: I think we can merge the DLEstimator and DLClassifier files into one file

docs/docs/APIGuide/MLPipeline/DLEstimator/DLEstimator.md
```diff
@@ -0,0 +1,84 @@
+## DLEstimator ##
+
+**Scala:**
+
+```scala
+val estimator = new DLEstimator(model: Module[T], criterion: Criterion[T], val featureSize: Array[Int], val labelSize: Array[Int])
+```
+
+Source code [here](https://github.com/intel-analytics/BigDL/blob/585818a3fb0e7339eb4e3831f08da82b7d5e47ba/spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala#L53).
+
```
jason-dai: Make it clear in the beginning that DLEstimator extends MLlib Estimator and can be directly used in Spark ML pipelines (and add a link to the Spark ML pipelines doc).

docs/docs/APIGuide/MLPipeline/DLEstimator/DLClassifier.md
```diff
@@ -0,0 +1,71 @@
...
+|[0.0, 1.0]|1.0  |1.0       |
+|[1.0, 0.0]|2.0  |2.0       |
+
+
+&nbsp;
+
+---
+
+
+## DLClassifierModel ##
```
jason-dai: Please add API code

jason-dai:
@hhbyyh please take a look

docs/docs/APIGuide/MLPipeline/DLEstimator/DLEstimator.md
```diff
@@ -0,0 +1,84 @@
...
+**Scala:**
+
+```scala
+val estimator = new DLEstimator(model: Module[T], criterion: Criterion[T], val featureSize: Array[Int], val labelSize: Array[Int])
+```
+
+Source code [here](https://github.com/intel-analytics/BigDL/blob/585818a3fb0e7339eb4e3831f08da82b7d5e47ba/spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala#L53).
+
+In machine learning context, "Estimation" is regarded as a mechanism for choosing the relatively "best model" according to the observation of reality. The API design of the `Estimator` abstracts the concept of a learning algorithm or any algorithm that fits or trains on data to make the training pipeline more clear and convinient for our users. Technically, an Estimator takes the user-presumed model and criterion with specification on feature and label dimension to prepare for the training. Within its class definition, it implements a method `fit()`, which accepts dataset to start training and produce a optimized DLModel with more accurate prediction.
```
hhbyyh: Merge the above paragraph with the following one. Maybe for explaining the `Estimator`, it's better if we just add a link to the Spark doc of Estimator here.

docs/docs/APIGuide/MLPipeline/DLEstimator/DLEstimator.md
```diff
@@ -0,0 +1,84 @@
...
+---
+## DLModel ##
+**Scala:**
+```scala
+val dlModel = new DLModel[T](model: Module[T], featureSize: Array[Int])
+```
+
+Source code [here](https://github.com/intel-analytics/BigDL/blob/585818a3fb0e7339eb4e3831f08da82b7d5e47ba/spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala#L155).
+
+You can see an internal method named [`wrapBigDLModel`](https://github.com/intel-analytics/BigDL/blob/master/spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala#L139) is called in our [`fit()`](https://github.com/intel-analytics/BigDL/blob/585818a3fb0e7339eb4e3831f08da82b7d5e47ba/spark/spark-version/2.0/src/main/scala/org/apache/spark/ml/DLEstimatorBase.scala#L74) method within the DLEstimatorBase class. It is designed to wrap the BigDL Module into a DLModel which is compatible with both spark 1.5-plus and 2.0 by entending the Spark's ML Transformer. It greatly improves the experience of Spark users because now you can use your BigDL DLModel as one of the transformers in your Spark ML pipeline stages.
```
hhbyyh: I would not mention the internal method here. Since code change will happen.

hhbyyh:
@jason-dai 
Suggestion to document style. I don't see it's necessary to put the texts for one paragraph in one line in the raw md files, which make it harder to view or edit. Markdown will automatically treat the consecutive lines in md file as one paragraph, and the line breaks among them will not affect the generated doc. 

In general, I would suggest to keep each line in md file less than 100 columns, to keep consistent with the code style.

docs/docs/APIGuide/MLPipeline/DLClassifier/DLClassifier.md
```diff
@@ -0,0 +1,71 @@
+## **DLClassifier**
+Source code [here](https://github.com/intel-analytics/BigDL/blob/585818a3fb0e7339eb4e3831f08da82b7d5e47ba/spark/dl/src/main/scala/org/apache/spark/ml/DLClassifier.scala#L36).
+
```
jason-dai: Add API definition and description of the APIs

docs/docs/APIGuide/MLPipeline/DLClassifier/DLClassifier.md
```diff
@@ -0,0 +1,71 @@
...
+
+&nbsp;
+
+---
+
+
+## DLClassifierModel ##
+Source code [here](https://github.com/intel-analytics/BigDL/blob/585818a3fb0e7339eb4e3831f08da82b7d5e47ba/spark/dl/src/main/scala/org/apache/spark/ml/DLClassifier.scala#L63).
+
```
jason-dai: Add API definition and description of the APIs

jason-dai:
And yes, we should just provide a summary and then link to the relevant pages (e.g., MLPipline page in the programming guide, and Spark ML doc on Spark website) in the doc, instead of duplicating the contents.

jason-dai:
@megaSpoon any update?

jason-dai:
@megaSpoon please update according to the comments.

docs/mkdocs.yml
megaSpoon: Done.

megaSpoon:
Python API for ML pipeline support is finished and tested. The DLEstimator and DLClassifier LR examples have the desired results compared to the examples written in scala by @hhbyyh. However, in this release, the result of scala-version example [DLEstimator(https://github.com/intel-analytics/BigDL/blob/master/spark/dl/src/main/scala/com/intel/analytics/bigdl/example/MLPipeline/DLEstimatorMultiLabelLR.scala)] can't be reproduced compared to the correct one in BigDL version 0.2.0. The prediction outcome becomes wildly large(Can anyone experiment with this example in bigDL version  0.3.0 to see if the problem did exist?). The codes remain the same so I guess something might be wrong elsewhere.

docs/mkdocs.yml
jason-dai: It's still two md files, no?

jason-dai:
Please split the PR into multiple, smaller and manageable PRs:
1) One for Python APIs
2) One for API docs
3) Possible one for examples

And please update the docs according to the previous comments.

megaSpoon:
Closed.

PR #1612, opened 2017-09-29
feat: serialization for QuantizedTensor

i8run:

QuantizedTensor has a byte array which stores the quantized value.
So we add byte supporting in bigdl.proto.

QuantizedTensor is not like other DenseTensor, so the logic in
DataConverter should be modified for supporting serializing it.

## How was this patch tested?

Jenkins.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/DataConverter.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/DataConverter.scala
@@ -258,114 +258,33 @@ object DataConverter extends DataConverter{
       if (storages.contains(tensorId)) {
         return storages.get(tensorId).get.asInstanceOf[AnyRef]
       }
-      val tensorType = serializedTensor.getTensorType
       val dataType = serializedTensor.getDatatype
+      val tensorType = serializedTensor.getTensorType
       val sizes = serializedTensor.getSizeList.asScala.toArray.map(_.intValue())
       val strides = serializedTensor.getStrideList.asScala.toArray.map(_.intValue())
       val offSet = serializedTensor.getOffset
       val isScalr = serializedTensor.getIsScalar
       val serializedStorage = serializedTensor.getStorage
       val storageId = serializedStorage.getId
-
       val created = if (storages.contains(storageId)) {
         storages.get(storageId).get
       } else {
         null
       }
-      
-      def dense(): Tensor[T] = {
-        val tensor = dataType match {
-          case DataType.FLOAT =>
-            val storage : Storage[Float] = if (created == null ) {
-              val data = serializedStorage.getFloatDataList.asScala.toArray.map(_.floatValue())
-              val newStorage = Storage[Float](data)
-              storages(storageId) = newStorage
-              newStorage
-            } else created.asInstanceOf[Storage[Float]]
-            Tensor[Float](storage, offSet, sizes, strides)
-          case DataType.DOUBLE =>
-            val storage : Storage[Double] = if (created == null ) {
-              val data = serializedStorage.getDoubleDataList.asScala.toArray.map(_.doubleValue())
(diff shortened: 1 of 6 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/DataConverter.scala
```diff
@@ -352,20 +405,56 @@ object DataConverter extends DataConverter{
...
         } else {
           val storageBuilder = TensorStorage.newBuilder
           if (tensorNumeric == NumericFloat) {
-            tensorBuilder.setDatatype(DataType.FLOAT)
-            storageBuilder.setDatatype(DataType.FLOAT)
-            tensor.storage().array().asInstanceOf[Array[Float]].
-              foreach(data => storageBuilder.addFloatData(data))
+            tensor.getTensorType match {
+              case DenseType =>
+                tensorBuilder.setDatatype(DataType.FLOAT)
```
wzhongyuan: why put setDataType inside case clause ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/DataConverter.scala
```diff
@@ -352,20 +405,56 @@ object DataConverter extends DataConverter{
...
+                storageBuilder.setDatatype(DataType.FLOAT)
+                tensor.storage().array().asInstanceOf[Array[Float]].
+                  foreach(data => storageBuilder.addFloatData(data))
+              case QuantizedType =>
+                tensorBuilder.setDatatype(DataType.FLOAT)
```
wzhongyuan: why put setDataType inside case clause ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/DataConverter.scala
```diff
@@ -431,17 +520,21 @@ object DataConverter extends DataConverter{
...
         } else {
-          val tensorNumeric = tensor.getTensorNumeric()
-          val offSet = tensor.storageOffset()
           val totalElement = tensor.nElement()
           val dimension = tensor.dim()
-          val isScalar = tensor.isScalar
           val tensorBuilder = BigDLTensor.newBuilder
           tensorBuilder.setId(tensorId)
           tensorBuilder.setDimension(dimension)
-          tensorBuilder.setOffset(offSet)
```
wzhongyuan: why remove this ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/DataConverter.scala
```diff
@@ -431,17 +520,21 @@ object DataConverter extends DataConverter{
           attributeBuilder.setTensorValue(storages.get(tensorId).get
             .asInstanceOf[BigDLTensor])
         } else {
-          val tensorNumeric = tensor.getTensorNumeric()
-          val offSet = tensor.storageOffset()
           val totalElement = tensor.nElement()
           val dimension = tensor.dim()
-          val isScalar = tensor.isScalar
```
wzhongyuan: why remove this ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/DataConverter.scala
i8run: For `QuantizedTensor`, there's no such methods supporting. So there're in next lines for `DenseTensor`.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/DataConverter.scala
i8run: It seems as previous.

PR #1619, opened 2017-09-29
RecurrentDecoder enhancement

dding3:
1. Add more scala doc for RecurrentDecoder
2. Small code change

## How was this patch tested?
ut


docs/docs/APIGuide/Layers/Recurrent-Layers.md
```diff
@@ -164,12 +164,12 @@ RecurrentDecoder module is a container of rnn cells which used to make
 a prediction of the next timestep based on the prediction we made from
 the previous timestep.
 
-Input for RecurrentDecoder has to be batch x ???(depends on cell type) without time information. 
+Input for RecurrentDecoder has to be batch x recoderShape without time information.
```
jason-dai: batch x recordShape, where stepShape if the shape of the input of a single time step.

docs/docs/APIGuide/Layers/Recurrent-Layers.md
```diff
@@ -164,12 +164,12 @@ RecurrentDecoder module is a container of rnn cells which used to make
 a prediction of the next timestep based on the prediction we made from
 the previous timestep.
 
-Input for RecurrentDecoder has to be batch x ???(depends on cell type) without time information. 
+Input for RecurrentDecoder has to be batch x stepShape(shape of the input at a single time).
```
jason-dai: at a single time step

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Cell.scala
```diff
@@ -80,7 +80,9 @@ abstract class Cell[T : ClassTag](
    * Please refer to SimpleRNN or LSTM for reference.
    * @return
    */
-  def preTopology: AbstractModule[Activity, Activity, T] = null
+  var preTopology: TensorModule[T]
+
+  private[nn] var combinePreTopology: Boolean = false
```
jason-dai: includePreTopology

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Cell.scala
```diff
@@ -117,47 +121,66 @@ abstract class Cell[T : ClassTag](
...
   }
 
   override def backward(input: Table, gradOutput: Table): Table = {
-    gradInput = cell.backward(input, gradOutput).toTable
+    if (combinePreTopology) {
+      val inputTensor = input.toTable[Tensor[T]](Recurrent.inputDim)
+      input(Recurrent.inputDim) = preTopology.output
+      gradInput = cell.backward(input, gradOutput)
+      gradInput(Recurrent.inputDim) =
+        preTopology.updateGradInput(inputTensor, gradInput.toTable[Tensor[T]](Recurrent.inputDim))
```
jason-dai: preTopology.backward()?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Cell.scala
```diff
@@ -117,47 +121,66 @@ abstract class Cell[T : ClassTag](
...
+    } else {
+      gradInput = cell.updateGradInput(input, gradOutput).toTable
+    }
     gradInput
   }
 
   override def accGradParameters(input: Table, gradOutput: Table): Unit = {
-    cell.accGradParameters(input, gradOutput)
+    throw new Exception("Should not enter Cell accGradParameters" +
+      "as it has override backward")
```
jason-dai: I think you still need to implement this, as it might be called from other layers, e.g., ParalleTable

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Cell.scala
```diff
@@ -80,7 +80,9 @@ abstract class Cell[T : ClassTag](
    * Please refer to SimpleRNN or LSTM for reference.
    * @return
    */
-  def preTopology: AbstractModule[Activity, Activity, T] = null
+  var preTopology: TensorModule[T]
```
jason-dai: Do we need to handle the new variable in model serialization? @wzhongyuan

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/LSTMPeephole.scala
```diff
@@ -76,11 +76,9 @@ class LSTMPeephole[T : ClassTag] (
         .add(SelectTable(1))
         .add(NarrowTable(2, 2)))
 
-  override def preTopology: AbstractModule[Activity, Activity, T] =
-    Sequential()
-    .add(Dropout(p))
-    .add(TimeDistributed(Linear(inputSize, hiddenSize * 4, wRegularizer = wRegularizer,
-      bRegularizer = bRegularizer)))
+  override var preTopology: TensorModule[T] =
+    Linear(inputSize, hiddenSize * 4, wRegularizer = wRegularizer,
```
jason-dai: I think this is a bug - we should turn off preTopology if p is non zero, just like other cells.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Cell.scala
```diff
@@ -117,47 +121,66 @@ abstract class Cell[T : ClassTag](
...
         hidden
       }
     }
   }
 
   override def updateOutput(input: Table): Table = {
-    output = cell.forward(input).toTable
+    if (combinePreTopology) {
+      val inputTensor = input.toTable[Tensor[T]](Recurrent.inputDim)
+      input(Recurrent.inputDim) = preTopology.updateOutput(inputTensor)
```
jason-dai: what if preTopology is null?

dding3: combinePreTopology is set at RecurrentDecoder, if preTopology is null, combinePreTopology is using default value(false)

jason-dai: So add assertion here?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Cell.scala
wzhongyuan: yes, we need to handle it, otherwise the behaviour will be different after deserialization for a 'true' value

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Cell.scala
```diff
@@ -117,47 +123,76 @@ abstract class Cell[T : ClassTag](
...
   }
 
   override def updateGradInput(input: Table, gradOutput: Table): Table = {
-    gradInput = cell.updateGradInput(input, gradOutput).toTable
+    if (includePreTopology) {
+      val inputTensor = input.toTable[Tensor[T]](Recurrent.inputDim)
+      input(Recurrent.inputDim) = preTopology.output
+      gradInput = cell.updateGradInput(input, gradOutput).toTable
+      gradInput2PreTopology.resizeAs(gradInput.toTable[Tensor[T]](Recurrent.inputDim))
+      gradInput2PreTopology.copy(gradInput.toTable[Tensor[T]](Recurrent.inputDim))
```
jason-dai: why copy?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Cell.scala
```diff
@@ -117,47 +123,76 @@ abstract class Cell[T : ClassTag](
...
     output
   }
 
   override def updateGradInput(input: Table, gradOutput: Table): Table = {
-    gradInput = cell.updateGradInput(input, gradOutput).toTable
+    if (includePreTopology) {
+      val inputTensor = input.toTable[Tensor[T]](Recurrent.inputDim)
+      input(Recurrent.inputDim) = preTopology.output
+      gradInput = cell.updateGradInput(input, gradOutput).toTable
+      gradInput2PreTopology.resizeAs(gradInput.toTable[Tensor[T]](Recurrent.inputDim))
```
jason-dai: it's gradOutput2PreTopology, no?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
```diff
@@ -121,18 +130,21 @@ class RecurrentDecoder[T : ClassTag](seqLength: Int)
 
   override def backward(input: Tensor[T], gradOutput: Tensor[T]): Tensor[T] = {
     val st = System.nanoTime
+    gradInput.resizeAs(output)
```
jason-dai: gradInput should be the same size as input; the gradInput of RecurrentDecoder is the gradInput of the first cell, no?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
```diff
@@ -121,18 +130,21 @@ class RecurrentDecoder[T : ClassTag](seqLength: Int)
 
   override def backward(input: Tensor[T], gradOutput: Tensor[T]): Tensor[T] = {
```
jason-dai: need to implement updateGradInput and accGradParameters too.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
dding3: In RecurrentDecoder, hiddensize is the same as inputsize and input doesn't have time dim while output and gradInput has. So i resize gradInput as output.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Cell.scala
dding3: Updated

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Cell.scala
dding3: Updated

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RecurrentDecoder.scala
dding3: Added

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Cell.scala
```diff
@@ -80,7 +80,11 @@ abstract class Cell[T : ClassTag](
    * Please refer to SimpleRNN or LSTM for reference.
    * @return
    */
-  def preTopology: AbstractModule[Activity, Activity, T] = null
+  var preTopology: TensorModule[T]
+
+  private[nn] var includePreTopology: Boolean = false
```
jason-dai: I think you also need to handle includePreTopology in the serialization

dding3: Added

jason-dai:
LGTM

dding3:
Jekins pass. Thanks for your review.

Issue #1622, opened 2017-09-30
Only Python2.7 is running on Jenkins recently


hkvision:
Python 3.5 is installed but running under python 3.5 has not been tested since mid Aug.

PR #1624, opened 2017-09-30
add python versions when run tests

hkvision:
Fix issue #1622


hkvision:
Seemingly still not working lol

Issue #1580, opened 2017-09-22
Refactor evaluation API in Python


zhichao-li:
1. Keep the evaluation api consistent between python and scala. 
Currently we use "evaluate" in scala but use "test" in Python
2. Provide a runnable quick example in the docs instead of pseudo code. 
https://bigdl-project.github.io/master/#APIGuide/Module/

hkvision:
#1641 Merged.

PR #1639, opened 2017-10-10
Refactor model test to model evaluation

hkvision:

Fix issue #1580


hkvision:
Error in calling function `evaluate` with the same name but different arguments
http://172.168.2.101:8080/job/BigDL-PR-Validation-2.2-python/6/console

Issue #1627, opened 2017-10-02
Python documentation of model test is not correct


jenniew:
In python documentation of model test (https://bigdl-project.github.io/0.2.0/#APIGuide/Module/ and https://bigdl-project.github.io/master/#APIGuide/Module/), the API example is not correct. The original example code is: evaluateResult = model.test(testSet, 1, [Top1Accuracy]). It should be evaluateResult = model.test(testSet, 1, [Top1Accuracy()]).

hkvision:
I fixed this problem in PR #1588 but this PR hasn't been merged to master yet.

hkvision:
#1588 merged. Fixed this issue.

hkvision:
Fix typo in branch-0.2 in pr #1638

hkvision:
Fix typo in master branch in pr #1641 and merged.

PR #1560, opened 2017-09-14
expose evaluate api to python

wzhongyuan:
## What changes were proposed in this pull request?

Expose evaluate() in Python

## How was this patch tested?

unit test and local test

PR #1534, opened 2017-09-04
Support running tensorflow model on BigDL (feeding data with placeholder, providing model output node)

yangw1234:
## What changes were proposed in this pull request?

Support running tensorflow model on BigDL.

Typical Usage
```python
opt_method = Adam(learningrate=0.001)
criterion = CrossEntropyCriterion()
init_engine()
model = Model.train(output_node, data, label, opt_method=opt_method, criterion=criterion, batch_size=128, end_when=MaxEpoch(10))
``` 

Refer to [notebook](https://github.com/yangw1234/BigDL-Tutorials-1/blob/runtf/notebooks/neural_networks/run_tensorflow_on_bigdl.ipynb) for complete code.

Limitations:
1. Users have to use exactly one placeholder to feed data and one placeholder to feed label.
2. Users have to pass in the model output, which he/she may not have a reference to

## How was this patch tested?

add unit tests and manually tested

PR #1641, opened 2017-10-10
Refactor Python API: model test to model evaluation

hkvision:

Requested by issue #1580 
- Model evaluation in API in scala and python are not consistent -> change `model.test` to `model.evaluate` on python side.
- Provide a runnable code in python model evaluation docs so that users can run directly.

At the same time, fix issue #1627 (correct typo in docs) in master branch.

## How was this patch tested?

Jenkins.


hkvision:
Remove duplicate definitions of evaluate(self) from pr #1534 and #1560

hkvision:
Python doesn't support multiple methods with the same name but different arguments while Scala supports.
Reference:
https://stackoverflow.com/questions/34546775/two-functions-same-name-python
https://stackoverflow.com/questions/23707644/is-it-legal-to-define-two-methods-with-the-same-name-but-different-returning-typ/23707683

hkvision:
Tested the updated docs locally.

hkvision:
PR Validation passed: http://172.168.2.101:8080/view/PR-Validation/job/BigDL-PR-Validation/1985/

yiheng:
LGTM

Issue #1585, opened 2017-09-22
Save bigdl model to torch model error


jenniew:
Load a caffe googlenet model to bigdl model, fine tune the model, and save back to torch mode, and get error.
Code:
model.saveTorch("/model/front_story_40/b.torchmodel", overWrite = true)
Log:
[test_story.txt](https://github.com/intel-analytics/BigDL/files/1323660/test_story.txt)

yiheng:
I think the error message is very clean. You model contains SpatialCrossMapLRN, but currently we havn't implent a saver for that kind of layer

jenniew:
Could we add this support? As this layer is common layer in googlenet.

PR #1642, opened 2017-10-11
Save bigdl model to torch model error corrected

Zefeng-Liu:
Made SpatialCrossMapLRN layer in BigDL compatible with Torch models (both read and write).

## How was this patch tested?
Used unit tests to do the test. First save a certain SpatialCrossMapLRN layer to a Torch model. Then load the Torch model and compare it with the former one. Also, used Lua code to run the produced Torch model and then compared the output with that of the former one.

fixed https://github.com/intel-analytics/BigDL/issues/1585


Zefeng-Liu:
@yiheng Hi Yiheng. Plz have a review on this PR, thx.

yiheng:
LGTM

PR #1643, opened 2017-10-11
feat: convert model between caffe, torch, tensorflow and bigdl.

i8run:
## How was this patch tested?

Manual testing for models.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/ConvertModel.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/ConvertModel.scala
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2016 The BigDL Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.intel.analytics.bigdl.utils
+
+import com.intel.analytics.bigdl.nn.Module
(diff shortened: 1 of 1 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/ConvertModel.scala
```diff
@@ -0,0 +1,95 @@
...
+        case "torch" =>
+          Module.loadTorch[Float](input)
+        case "caffe" =>
+          CaffeLoader.loadCaffe[Float](param.prototxt, input)._1
+        case _ =>
+          throw new UnsupportedOperationException(s"Unsupported model type.")
+      }
+
+      val model = if (param.quantize) {
+        require(param.to.toLowerCase == "bigdl",
```
jason-dai: Should report error instead of throwing exception

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/ConvertModel.scala
```diff
@@ -0,0 +1,95 @@
...
+
+      val loadedModel = param.from.toLowerCase match {
+        case "bigdl" =>
+          Module.loadModule[Float](input)
+        case "torch" =>
+          Module.loadTorch[Float](input)
+        case "caffe" =>
+          CaffeLoader.loadCaffe[Float](param.prototxt, input)._1
+        case _ =>
+          throw new UnsupportedOperationException(s"Unsupported model type.")
```
jason-dai: Should report error instead of throwing exception

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/ConvertModel.scala
```diff
@@ -0,0 +1,95 @@
...
+
+      param.to.toLowerCase match {
+        case "bigdl" =>
+          model.saveModule(output, overWrite = true)
+        case "torch" =>
+          model.saveTorch(output)
+        case "caffe" =>
+          model.saveCaffe(param.prototxt, output)
+        case _ =>
+          throw new UnsupportedOperationException(s"Unsupported model type.")
```
jason-dai: Should report error instead of throwing exception

jason-dai:
Should also support tensorflow models

PR #1644, opened 2017-10-11
handle reflection race conditions in 2.10

wzhongyuan:

Reflection in scala 2.10 is not thread safe , we need to handle it in serialization 

## How was this patch tested?

unit test


wzhongyuan:
PR validation passed
http://10.239.47.211:18888/job/BigDL-PR-Validation/2027/ to

PR #1647, opened 2017-10-12
move model freeze to AbstractModule

SeaOfOcean:
move model freeze to container

## How was this patch tested?

unit test

### DIFF
--- a/docs/docs/APIGuide/Module.md
+++ b/docs/docs/APIGuide/Module.md
@@ -170,6 +170,8 @@ its parameters(weight/bias, if exists) are not changed in training process
 * stop the input gradient of layers that match the given names. Their input gradient are not computed.
 And they will not contributed to the input gradient computation of layers that depend on them.
 
+Note that stopGradient is only supported in Graph model.
+
 **Python**
 ```python
 layer.freeze()
--- a/docs/docs/APIGuide/Module.md
+++ b/docs/docs/APIGuide/Module.md
@@ -156,17 +156,17 @@ val preductResult = model.predict(predictSet)
 To "freeze" a module means to exclude some layers of model from training.
 
 ```scala
-layer.freeze()
-layer.unFreeze()
-model.freeze(Array("layer1", "layer2"))
-model.unFreeze()
-model.stopGradient(Array("layer1"))
+module.freeze()
+module.unFreeze(scaleW = 2, scaleB = 1)
+module.freeze(Array("layer1", "layer2"))
+module.unFreeze(Array("layer1", "layer2"), scaleW = 2, scaleB = 1)
+module.stopGradient(Array("layer1"))
 ```
-* A single layer can be "freezed" by calling ```freeze()```. If a layer is freezed,
+* The whole module can be "freezed" by calling ```freeze()```. If a module is freezed,
 its parameters(weight/bias, if exists) are not changed in training process
-* A single layer can be "unFreezed" by calling ```unFreeze()```.
-* User can set freeze of list of layers in model by calling ```freeze```
-* User can unfreeze all layers by calling ```unFreeze```
+* The whole module can be "unFreezed" by calling ```unFreeze()```
+* User can freeze list of layers in model by calling ```freeze```
+* User can unfreeze list of layers by calling ```unFreeze```
 * stop the input gradient of layers that match the given names. Their input gradient are not computed.
 And they will not contributed to the input gradient computation of layers that depend on them.
 
@@ -174,11 +174,11 @@ Note that stopGradient is only supported in Graph model.
 
 **Python**
 ```python
-layer.freeze()
-layer.unfreeze()
-model.freeze(["layer1", "layer2"])
-model.unfreeze()
-model.stop_gradient(["layer1"])
+module.freeze()
+module.unfreeze(scale_w = 2, scale_b = 1)
+module.freeze(["layer1", "layer2"])
+module.unfreeze(["layer1", "layer2"], scale_w = 2, scale_b = 1)
+module.stop_gradient(["layer1"])
 ```
 
 **Scala**
--- a/docs/docs/APIGuide/Module.md
+++ b/docs/docs/APIGuide/Module.md
@@ -156,17 +156,16 @@ val preductResult = model.predict(predictSet)
 To "freeze" a module means to exclude some layers of model from training.
 
 ```scala
-module.freeze()
-module.unFreeze(scaleW = 2, scaleB = 1)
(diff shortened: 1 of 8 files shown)


SeaOfOcean:
pr validate pass

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -233,6 +233,22 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
   }
 
   /**
+   * set an array of layers that match the given ```names``` to be "freezed",
+   * i.e. their parameters(weight/bias, if exists) are not changed in training process
+   * @param names an array of layer names
+   * @return current graph model
+   */
+  def freeze(names: Array[String]): this.type = {
```
yiheng: def freeze() ? also support unfreeze(name ：String*)

yiheng: Also python API and doc update?

SeaOfOcean:
Now the freeze and unFreeze API all support parameter names: Array[String] (default is null, and will apply to all)

docs/docs/APIGuide/Module.md
```diff
@@ -156,27 +156,26 @@ val preductResult = model.predict(predictSet)
...
+
 **Python**
 ```python
-layer.freeze()
-layer.unfreeze()
-model.freeze(["layer1", "layer2"])
-model.unfreeze()
-model.stop_gradient(["layer1"])
+module.freeze(["layer1", "layer2"])
+module.unfreeze(scale_w = 2, scale_b = 1, names = ["layer1", "layer2"])
```
yiheng: should not let user input scale_w and scale_b, we should cache 1 in the model

docs/docs/APIGuide/Module.md
```diff
@@ -156,27 +156,26 @@ val preductResult = model.predict(predictSet)
...
-model.unFreeze()
-model.stopGradient(Array("layer1"))
-```
-* A single layer can be "freezed" by calling ```freeze()```. If a layer is freezed,
-its parameters(weight/bias, if exists) are not changed in training process
-* A single layer can be "unFreezed" by calling ```unFreeze()```.
-* User can set freeze of list of layers in model by calling ```freeze```
-* User can unfreeze all layers by calling ```unFreeze```
+module.freeze(Array("layer1", "layer2"))
+module.unFreeze(scaleW = 2, scaleB = 1, names = Array("layer1", "layer2"))
```
yiheng: should not let user input scale_w and scale_b, we should cache 1 in the model

yiheng: try to use var args instead of array. Then you only need one API

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -213,22 +213,50 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
...
+   * if names is not null,
+   * set an array of layers that match the given ```names``` to be "freezed",
+   *
+   * @param names an array of layer names
+   * @return current graph model
    */
-  def freeze(): this.type = {
-    setScaleW(0)
-    setScaleB(0)
+  def freeze(names: Array[String] = null): this.type = {
```
yiheng: ugly. Don't use null for some meaning in a public API! use var length String instead

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -213,22 +213,50 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
...
+    if (null == names) {
+      setScaleW(0)
```
yiheng: why not override this in container?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -213,22 +213,50 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
...
-    setScaleB(1)
+   * if names is not null, unfreeze layers that match given names
+   *
+   * @param names array of module names to unFreeze
+   * @param scaleW Set the scale of gradientWeight
+   * @param scaleB Set the scale of gradientBias
+   */
+  def unFreeze(scaleW: Double = 1, scaleB: Double = 1, names: Array[String] = null): this.type = {
+    if (null == names) {
+      setScaleW(scaleW)
```
yiheng: why not override this in the container

yiheng: Don't let user pass in scaleW and scaleB.

SeaOfOcean:
pr validate pass

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -212,23 +212,54 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
...
+   * i.e. their parameters(weight/bias, if exists) are not changed in training process
+   * if names is not empty,
+   * set an array of layers that match the given ```names``` to be "freezed",
+   *
+   * @param names an array of layer names
+   * @return current graph model
+   */
+  def freeze(names: String*): this.type = {
+    if (names.isEmpty) {
+      scaleWCache = scaleW
```
yiheng: if scaleW != 0?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -212,23 +212,54 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
...
+      scaleBCache = scaleB
```
yiheng: if scaleB != 0?

yiheng:
LGTM

SeaOfOcean:
pr validate pass

PR #1648, opened 2017-10-12
fix: QuantizeTensor supports for ModelBroadcast

i8run:

`QuantizeTensor` does not support storage method, so it needs specific judgement.

## How was this patch tested?
Jenkins.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
@@ -71,9 +71,9 @@ class ModelBroadcast[T: ClassTag](implicit ev: TensorNumeric[T]) extends Seriali
       if (parameters._1(i) != null) {
         val wb = parameters._1(i)
         wb match {
-          case value1: QuantizedTensor[T] =>
-            weightsBias(i) = QuantizedTensor[T](value1.getStorage, value1.maxOfRow,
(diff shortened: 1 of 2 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
```diff
@@ -70,8 +70,14 @@ class ModelBroadcast[T: ClassTag](implicit ev: TensorNumeric[T]) extends Seriali
     while (i < parameters._1.length) {
       if (parameters._1(i) != null) {
         val wb = parameters._1(i)
-        weightsBias(i) = Tensor[T](Storage(wb.storage().array()),
-          wb.storageOffset(), wb.size(), wb.stride())
+        wb match {
+          case value1: QuantizedTensor[T] =>
```
SeaOfOcean: change value1 to some meaningful name

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
```diff
@@ -80,11 +86,18 @@ class ModelBroadcast[T: ClassTag](implicit ev: TensorNumeric[T]) extends Seriali
       if (parameters._1(i) != null) {
         parameters._1(i).set()
       }
+      i += 1
+    }
+
+    // because in quantized mode, the parameters are not the same length.
```
SeaOfOcean: in quantized mode, the weight number may be different from gradWeight number

PR #1651, opened 2017-10-12
implement RandomShuffleQueue

yangw1234:

implement RandomShuffleQueue

## How was this patch tested?

modify existing unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/Session.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/Session.scala
@@ -34,7 +34,9 @@ import com.intel.analytics.bigdl.models.utils.ModelBroadcast
 import TFTensorNumeric.NumericByteString
 
 import scala.collection.mutable
+import scala.collection.mutable.ArrayBuffer
 import scala.reflect.ClassTag
+import scala.util.Random
 
 abstract class Session[T: ClassTag] {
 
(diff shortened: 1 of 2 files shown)


yangw1234:
pr validation pass

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/Session.scala
```diff
@@ -248,78 +250,33 @@ class BigDLSessionImpl[T: ClassTag](
...
-        val elem = table[Tensor[ByteString]](j + 1)
-        newTable.insert(elem(i + 1))
-        j = j + 1
-      }
-      result(i) = newTable
-      i = i + 1
-    }
-    result
-  }
+        override def hasNext: Boolean = buffer.nonEmpty || iter.hasNext
```
yiheng: so we can only train one Epoch?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/Session.scala
```diff
@@ -34,9 +34,11 @@ import com.intel.analytics.bigdl.models.utils.ModelBroadcast
 import TFTensorNumeric.NumericByteString
 
 import scala.collection.mutable
+import scala.collection.mutable.ArrayBuffer
 import scala.reflect.ClassTag
+import scala.util.Random
 
-abstract class Session[T: ClassTag] {
+abstract class Session[T](implicit ct: ClassTag[T]) {
```
yiheng: Is it equal to T: ClassTag?

yangw1234: Yes, they are the same. I want to assign this implicit parameter to a local variable to avoid serializing the whole Session.

yiheng: you can put that method to the object

yangw1234:
pr validation pass

PR #1654, opened 2017-10-12
Python api support

megaSpoon:
DLModel, DLClassifer, DLModel, DLClassifier Python wrapper codes
PythonBigDL class (add responsive methods from python wrapper)
Examples to test

## How was this patch tested?

test with writtern examples comparing the results with the scala version

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,4 +35,3 @@ nohup.out
 .ipynb_checkpoints/
 pyspark/dist/
 pyspark/build/
-
--- a/pyspark/test/bigdl/test_dl_classifier.py
+++ b/pyspark/test/bigdl/test_dl_classifier.py
@@ -1,0 +1,170 @@
+#
+# Copyright 2016 The BigDL Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from bigdl.nn.layer import *
+from bigdl.nn.criterion import *
+from bigdl.util.common import *
+from bigdl.models.ml_pipeline.dl_classifier import *
+from pyspark.sql.types import *
+from pyspark.context import SparkContext
+import numpy as np
+import pytest
+from numpy.testing import assert_allclose
+
+
+class TestDLClassifer():
+    def setup_method(self, method):
+        """ setup any state tied to the execution of the given method in a
+        class.  setup_method is invoked for every test method of a class.
+        """
+        sparkConf = create_spark_conf()
+        self.sc = SparkContext(master="local[1]", appName="test model",
+                               conf=sparkConf)
+        self.sqlContext = SQLContext(self.sc)
+        init_engine()
+
+    def teardown_method(self, method):
+        """ teardown any state that was previously setup with a setup_method
+        call.
+        """
+        self.sc.stop()
+
+    def test_all_set_get_methods(self):
+        """ run tests for all the set and get methods involved in DLEstimator, DLModel,
+            DLClassifier, DLClassifierModel
+        """
+
+        '''
+        use linear model and MSE criterion to test the DLEstimator and DLModel
+        '''
+        linear_model = Sequential().add(Linear(2, 2))
+        mse_criterion = MSECriterion()
+
+        '''
+        initialize a DLEstimator to test setter, getter for
+        batchSize, maxEpoch, learningRate in DLEstimator
+        '''
+        estimator = DLEstimator(model=linear_model, criterion=mse_criterion,
+                                feature_size=[2], label_size=[2])
+
+        # check the set and get methods for batchSize
+        assert estimator.setBatchSize(30).getBatchSize() == 30
+        # check the set and get methods for maxEpoch
+        assert estimator.setMaxEpoch(40).getMaxEpoch() == 40
+        # check the set and get methods for learningRate
+        assert estimator.setLearningRate(1e-4).getLearningRate() == 1e-4
+
+        '''
+        initialize a DLModel to test setter, getter for featureSize, batchSize in DLModel
+        '''
+        dl_model = DLClassifierModel(model=linear_model, featureSize=[1])
+
+        # check the set and get methods for featureSize in DLModel
+        assert dl_model.setFeatureSize([2]).getFeatureSize() == [2]
+        # check the set and get methods for batchSize in DLModel
+        assert dl_model.setBatchSize((20)).getBatchSize() == 20
+
+        '''
+        use linear model and ClassNLL criterion to test the DLClassifier and DLClassifierModel
+        '''
+        linear_model = Sequential().add(Linear(2, 2))
+        classNLL_criterion = ClassNLLCriterion()
+
+        '''
+        initialize a DLClassifier to test setter, getter for
+        batchSize, maxEpoch, learningRate in DLClassifier
+        '''
+        classifier = DLClassifier(model=linear_model, criterion=classNLL_criterion,
+                                  feature_size=[2])
+
+        # check the set and get methods for batchSize
+        assert classifier.setBatchSize(20).getBatchSize() == 20
+        # check the set and get methods for maxEpoch
+        assert classifier.setMaxEpoch(50).getMaxEpoch() == 50
+        # check the set and get methods for learningRate
+        assert classifier.setLearningRate(1e-5).getLearningRate() == 1e-5
+        '''
+        initialize a DLClassifierModel to test setter, getter
+        for featureSize, batchSize in DLClassifierModel
+        '''
+        dl_classifier_model = DLClassifierModel(model=linear_model, featureSize=[1])
+
+        # check the set and get methods for featureSize
+        assert dl_classifier_model.setFeatureSize([2]).getFeatureSize() == [2]
+
+        # check the set and get methods for batchSize
+        assert dl_classifier_model.setBatchSize((20)).getBatchSize() == 20
+
+    def test_dlestimator_fit_dlmodel_transform(self):
+        model = Sequential().add(Linear(2, 2))
+        criterion = MSECriterion()
+        estimator = DLEstimator(model, criterion, [2], [2]).setBatchSize(4)\
+            .setLearningRate(0.01).setMaxEpoch(1000)
+
+        data = self.sc.parallelize([
+            ((2.0, 1.0), (1.0, 2.0)),
+            ((1.0, 2.0), (2.0, 1.0)),
+            ((2.0, 1.0), (1.0, 2.0)),
+            ((1.0, 2.0), (2.0, 1.0))])
+
+        schema = StructType([
+            StructField("features", ArrayType(DoubleType(), False), False),
+            StructField("label", ArrayType(DoubleType(), False), False)])
+        df = self.sqlContext.createDataFrame(data, schema)
+        dlModel = estimator.fit(df)
+
+        dlModel.transform(df).createTempView("dlModelDF")
+        results = self.sqlContext.table("dlModelDF")
+
+        count = results.rdd.count()
+        data = results.rdd.collect()
+
+        for i in range(count):
+            row_label = data[i]['label']
+            row_prediction = data[i]['prediction']
+            assert_allclose(row_label[0], row_prediction[0], atol=0.01, rtol=0)
+            assert_allclose(row_label[1], row_prediction[1], atol=0.01, rtol=0)
+
+    def test_dlclassifier_fit_dlclassifiermodel_transform(self):
+        model = Sequential().add(Linear(2, 2))
+        criterion = ClassNLLCriterion()
+        classifier = DLClassifier(model, criterion, [2]).setBatchSize(4)\
+            .setLearningRate(0.01).setMaxEpoch(1000)
+        data = self.sc.parallelize([
+            ((2.0, 1.0), 1.0),
+            ((1.0, 2.0), 2.0),
+            ((2.0, 1.0), 1.0),
+            ((1.0, 2.0), 2.0)])
+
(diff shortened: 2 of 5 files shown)


spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
```diff
@@ -271,7 +271,7 @@ object DLModel {
 trait HasBatchSize extends Params {
 
   final val batchSize: Param[Int] = new Param[Int](this, "batchSize", "batchSize")
-  setDefault(batchSize -> 1)
+  setDefault(batchSize -> 4)
```
jason-dai: different default value in python and scala?

pyspark/bigdl/ml_piepeline/dl_classifier.py
```diff
@@ -0,0 +1,118 @@
...
+from pyspark.ml.param.shared import *
+from bigdl.util.common import *
+from pyspark.ml.wrapper import JavaParams
+
+if sys.version >= '3':
+    long = int
+    unicode = str
+
+
+class HasBatchSize(Params):
```
jason-dai: Do we need a HasMaxEpoch class too?

pyspark/bigdl/ml_piepeline/dl_classifier.py
```diff
@@ -0,0 +1,118 @@
...
+        return self
+
+    def getBatchSize(self):
+        """
+        Gets the value of batchSize or its default value.
+        """
+        return self.getOrDefault(self.batchSize)
+
+
+class DLEstimator(JavaParams, Estimator, HasFeaturesCol, HasLabelCol, HasPredictionCol, HasBatchSize):
```
jason-dai: Can we make it a subclass of JavaValue too? If so, there is no need to duplicate the jvm_class_constructor code below.

pyspark/bigdl/ml_piepeline/dl_classifier.py
```diff
@@ -0,0 +1,118 @@
...
+        return name
+
+    def __init__(self,  model, featureSize, jvalue=None, bigdl_type="float"):
+        super(DLModel, self).__init__()
+        self.value = jvalue if jvalue else callBigDlFunc(
+            bigdl_type, self.jvm_class_constructor(), model, featureSize)
+        self.bigdl_type = bigdl_type
+
+    def setFeatureSize(self, value):
+        callBigDlFunc(self.bigdl_type, "dlmodelSetFeatureSize", self.value, value)
```
jason-dai: Is dlmodelSetFeatureSize implemented? Maybe we don't need this method.

pyspark/bigdl/ml_piepeline/dl_classifier.py
```diff
@@ -0,0 +1,118 @@
...
+        return self
+
+    def getFeatureSize(self):
+        return callBigDlFunc(self.bigdl_type, "dlmodelGetFeatureSize", self.value)
```
jason-dai: Is dlmodelGetFeatureSize implemented? Maybe we don't need this method.

jason-dai:
Please add some unit tests (see https://github.com/intel-analytics/BigDL/tree/master/pyspark/test)

jason-dai:
@jenniew @hhbyyh please take a look

pyspark/bigdl/ml_piepeline/dl_classifier.py
megaSpoon: I delete the` jvm_class_constructor` method declaration, import `JavaValue` from `bigdl.util.common` and replace it with JavaValue.jvm_class_constructor now. JavaValue's init method needs to specify parameters DLEstimator doesn't have like `jvalue` which could be easier by just using the class method outside the JavaValue class.

megaSpoon:
For unit tests, I will use the codes in the LR examples for estimator and classifier to see if the prediction result and expected result are the same. But I have a question about the unit test for estimator since the  prediction result doesn't remain the same each time I run, probably due to the thing that in the training stage, the optimizer initializes some parameters randomly. Thus, what could be the best strategy for writing a unit test but the expected result of the test is unknown.

pyspark/bigdl/ml_piepeline/dl_classifier.py
jason-dai: that works for now

jason-dai:
In general for Python API we just need to test it has the same result as the corresponding Scala API.

spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
megaSpoon: changed back to the same

pyspark/bigdl/ml_piepeline/dl_classifier.py
megaSpoon: added and also create a HasFeatureSize class for DLModel

megaSpoon:
My confusion above related to the DLEstimator API is that when using DLEstimator, the prediction result is even unrepeatable. It cannot be tested by using scala DLEstimator to produce a result A, using python DLEstimator to produce a result B to see if A equals B due to the randomness. I am guessing we can only do it by initializing the same random seed in their optimizers, am I right? I am trying to figure out how to do it.

hhbyyh:
I thought @jenniew already had a PR on this...

Anyway, for unit tests, just some simple tests should be sufficient. maybe something like scala unit tests.

Also we should support Python DLEstimator/DLModel etc. in spark ml pipeline. please add unit test for that.

jason-dai:
@zhichao-li please take a look

pyspark/bigdl/ml_piepeline/dl_classifier.py
zhichao-li: You can put `None` for `jvalue` if its empty.
The code is duplicate if without inheritance:
```
+    def jvm_class_constructor(self):
+        name = "create" + self.__class__.__name__
+        print("creating: " + name)
+        return name
+
+    def __init__(self,  model, criterion, featureSize, labelSize, bigdl_type="float"):
+        super(DLEstimator, self).__init__()
+        self.value = callBigDlFunc(
+            bigdl_type, self.jvm_class_constructor(), model, criterion, featureSize, labelSize)
```

```

**Comment by megaSpoon** [comment]
Update:
Thanks for @hhbyyh @yiheng's reviews. 
I am using the [LR pipeline example](https://spark.apache.org/docs/2.1.1/ml-pipeline.html#pipeline) in our BigDL fashion to demonstrate our API's feasibility.

**Comment by hhbyyh** [comment]
That's nice to have. For unit test, just compose any Transformers in Spark (e.g. StandardScaler) and DLClassifier in a pipeline (fit and transform) should be sufficient.

File: pyspark/bigdl/ml_piepeline/dl_classifier.py [hunk 6]
```diff
@@ -0,0 +1,118 @@
...
+    def getBatchSize(self):
+        """
+        Gets the value of batchSize or its default value.
+        """
+        return self.getOrDefault(self.batchSize)
+
+
+class DLEstimator(JavaParams, Estimator, HasFeaturesCol, HasLabelCol, HasPredictionCol, HasBatchSize):
+    maxEpoch = Param(Params._dummy(), "maxEpoch", "number of max Epoch")
+    def jvm_class_constructor(self):
```
zhichao-li: remove jvm_class_constructor?

pyspark/bigdl/ml_piepeline/dl_classifier.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def _fit(self, dataset):
+        #self._transfer_params_to_java()
+        jmodel = callBigDlFunc(self.bigdl_type, "fitEstimator", self.value, dataset)
+        model = DLModel.of(jmodel, self.bigdl_type)
+        return model
+
+
+class DLModel(JavaParams, Model, HasFeaturesCol, HasPredictionCol, HasBatchSize):
+    def jvm_class_constructor(self):
```
zhichao-li: remove?

zhichao-li:
@megaSpoon  why there are two dl_classifier.py in this pr?

zhichao-li:
We need to add some unit-tests for this pipline api to ensure they are callable at the first stage.
i.e "test_pipline.py"

pyspark/bigdl/ml_piepeline/dl_classifier.py
megaSpoon: already fixed

spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDL.scala
```diff
@@ -1976,4 +1978,76 @@ class PythonBigDL[T: ClassTag](implicit ev: TensorNumeric[T]) extends Serializab
...
+
+  def fitClassifier(classifier: DLClassifier[T], dataSet: DataFrame): DLModel[T] = {
+    classifier.fit(dataSet)
+  }
+
+  def setBatchSizeDLEstimator(estimator: DLEstimator[T], batchSize: Int): DLEstimator[T] = {
+    estimator.setBatchSize(batchSize)
+  }
+
+  def setBatchSizeDLClassifier(classifier: DLClassifier[T], batchSize: Int): DLClassifier[T] = {
```
zhichao-li: For such simple setter, actually we can invoke it directly from python without adding java wrapper method.
but anyway, adding a java wrapper is ok as well.
i.e
```
    def set_name(self, name):
        """
        Give this model a name. There would be a generated name
        consist of class name and UUID if user doesn't set it.
        """
        callJavaFunc(get_spark_context(), self.value.setName, name)
        return self
```

spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDL.scala
```diff
@@ -1979,6 +1979,12 @@ class PythonBigDL[T: ClassTag](implicit ev: TensorNumeric[T]) extends Serializab
     module.quantize()
   }
 
+  def createDLEstimator() = {
```
zhichao-li: empty method?

.gitignore
```diff
@@ -35,4 +35,8 @@ nohup.out
 .ipynb_checkpoints/
 pyspark/dist/
 pyspark/build/
-
+pyspark/bigdl/models/ml_pipeline/*.ipynb_checkpoints/
+pyspark/bigdl/models/ml_pipeline/metastore_db
+pyspark/bigdl/models/ml_pipeline/derby.log
+pyspark/bigdl/models/ml_pipeline/*.ipynb
+pyspark/bigdl/models/ml_pipeline/run.sh
```
zhichao-li: useless for the other user?

pyspark/test/bigdl/test_dl_classifier.py
```diff
@@ -0,0 +1,163 @@
...
+from bigdl.util.common import *
+from bigdl.models.ml_pipeline.dl_classifier import *
+from pyspark.sql.types import *
+from pyspark.context import SparkContext
+import numpy as np
+import pytest
+from numpy.testing import assert_allclose
+
+
+class TestTensorflow():
```
jason-dai: why TestTensorflow?

spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDL.scala
```diff
@@ -1979,6 +1979,12 @@ class PythonBigDL[T: ClassTag](implicit ev: TensorNumeric[T]) extends Serializab
...
+  }
+
+  def createDLClassifier() = {
+  }
```
jason-dai: Are the above two methods used?

pyspark/test/bigdl/test_dl_classifier.py
megaSpoon: naming error, fixed!

spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDL.scala
megaSpoon: removed, originally used to test the javaValue.jvm_class_constructor

pyspark/test/bigdl/test_dl_classifier.py
megaSpoon: I run the run-all.sh in the pyspark/test/dev folder. It passed without error.

megaSpoon: @zhichao-li Is this the way above to run tests in pyspark/test/bigdl folder?

spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDL.scala
```diff
@@ -1976,4 +1978,76 @@ class PythonBigDL[T: ClassTag](implicit ev: TensorNumeric[T]) extends Serializab
...
+    classifier.setBatchSize(batchSize)
+  }
+
+  def setMaxEpochDLEstimator(estimator: DLEstimator[T], maxEpoch: Int): DLEstimator[T] = {
+    estimator.setMaxEpoch(maxEpoch)
+  }
+
+  def setMaxEpochDLClassifier(classifier: DLClassifier[T], maxEpoch: Int): DLClassifier[T] = {
+    classifier.setMaxEpoch(maxEpoch)
+  }
```
jason-dai: Is it possible to combine the above two methods?

jason-dai:
Please rebase and run a Jenkins validation test

spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
```diff
@@ -91,7 +91,7 @@ class DLEstimator[@specialized(Float, Double) T: ClassTag](
 
   /**
    * learning rate for the optimizer.
-   * Default: 1.0
+   * Default: 0.01
```
jason-dai: the comment is incorrect?

megaSpoon: change to 0.001 fixed.

megaSpoon:
The unit test #1429 on Jenkins already passed. The PR is ready to merge. @jason-dai

megaSpoon:
I will simplify the codes of set and get methods tomorrow. Test if combining the two set methods together as one or simply invoking them directly from python based on @zhichao-li 's advice works.

PR #1656, opened 2017-10-13
Support load more tensorflow operations

yiheng:
Support load more tensorflow operations

## How was this patch tested?
existing unit test


yiheng:
PR validation pass

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/loaders/Adapter.scala
```diff
@@ -46,6 +46,10 @@ class Adapter[T: ClassTag](
   private var initTensors: Array[Tensor[_]] = _
 
   override def updateOutput(input: Table): Activity = {
+    if (getName().contains("softmax_cross_entropy_loss/num_present")) {
+      println()
+    }
+
```
yangw1234: delete these?

PR #1658, opened 2017-10-13
Add BroadcastGradientArgs

psyyz10:
Add BroadcastGradientArgs

## How was this patch tested?
Unit tests


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/BroadcastGradientArgs.scala
```diff
@@ -0,0 +1,107 @@
...
+        ry.append(1)
+      }
+    }
+
+    val xReducedIndexBuffer = new ArrayBuffer[Int]()
+    val yReducedIndexBuffer = new ArrayBuffer[Int]()
+
+    val n = rx.length
+
+    for (i <- 0 until n) {
```
yiheng: while instead of for

yiheng:
LGTM

Issue #1601, opened 2017-09-28
Saving VGG16 which is converted from TF throw OutOfMemoryError


i8run:
Branch: master, 2017.09.27
VGG16 model: http://download.tensorflow.org/models/vgg_16_2016_08_28.tar.gz

Progress:

1. freeze the tf model
2. load with `Module.loadTF`
3. save with `Moduel.saveModule`

PR #1659, opened 2017-10-13
Fix duplicated msg proto

wzhongyuan:

Protobuf serializer is a tree-like serializer similar to XMLSerializer so it doesn't maintain the reference even two msgs pointing to the same msgs.

This PR is to fix this to avoid duplicated storage in serialization

## How was this patch tested?

Unit test and local test with VGG model

fixed https://github.com/intel-analytics/BigDL/issues/1601


wzhongyuan:
PR validation passed http://10.239.47.211:18888/job/BigDL-PR-Validation/2088/

PR #1666, opened 2017-10-16
Cache zeroGrad future instead of use a driver local object

yiheng:
Cache zeroGrad future instead of use a driver local object

## How was this patch tested?
add assert in training process. Existing unit test

### DIFF
--- a/docs/docs/PythonUserGuide/install-from-pip.md
+++ b/docs/docs/PythonUserGuide/install-from-pip.md
@@ -5,7 +5,7 @@
 - We've tested this package with __Python 2.7__ and __Python 3.5__. Only these two Python versions are supported for now.
 
 
-## **Install BigDL-0.2.0.dev2**
+## **Install BigDL-0.3.0.dev0**
 
 Install BigDL release via pip (we tested this on pip 9.0.1)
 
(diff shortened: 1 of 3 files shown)


yiheng:
PR validation pass

yiheng:
PR validation pass

dding3:
May I know why we make this change? The reason why we put zero gradients at the end of the first job is we want overlap zero gradients with existing steps as much as possible. Currently zero gradients is parallel executed with launching the first and second jobs, aggregate gradients etc.

yiheng:
PR validation pass

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -198,12 +199,18 @@ object DistriOptimizer {
               b += 1
             }
           })
-          Engine.default.sync(tasks)
+          if (cached.zeroGradStarted) {
```
dding3: Why do we need this logic here?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -198,12 +199,18 @@ object DistriOptimizer {
...
+            assert(cached.tasks.length == _subModelNumber + 1)
```
dding3: I think it would be better if we give explanation about what we expected and what should we do if assert failed.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
yiheng: to verify the logic

dding3:
LGTM

PR #1670, opened 2017-10-16
Adding ndarray_to_rdd function

gaarsmu:
This is my first pull request, so don't be too angry...

Adding a function, which converts numpy ndarray to rdd.

## How was this patch tested?
I used tested this function in gradient descent algorithm realization, and tested Errors, could be find here:
https://github.com/gaarsmu/BigDL_notebooks/tree/master/to_RDD_func

### DIFF
--- a/pyspark/bigdl/util/common.py
+++ b/pyspark/bigdl/util/common.py
@@ -434,6 +434,18 @@ def _test():
     if failure_count:
         exit(-1)
 
+class TransformationError(BaseException):
+    pass
+
+def ndarray_to_rdd(X, y):
+    if type(X) != np.ndarray or type(y) != np.ndarray:
+        raise TransformationError('Input arrays should be numpy ndarrays')
+    if X.shape[0] != y.shape[0]:
+        raise TransformationError('Input arrays should have the same lenght')
+    sc = get_spark_context()    
+    return sc.parallelize(X).zip(sc.parallelize(y)).map(
+            lambda x: Sample.from_ndarray(x[0], x[1]))
+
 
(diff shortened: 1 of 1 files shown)


pyspark/bigdl/util/common.py
```diff
@@ -434,6 +434,18 @@ def _test():
     if failure_count:
         exit(-1)
 
+class TransformationError(BaseException):
+    pass
+
+def ndarray_to_rdd(X, y):
```
zhichao-li: How about `to_sample_rdd` or `to_rdd_of_sample` ?

pyspark/bigdl/util/common.py
```diff
@@ -434,6 +434,18 @@ def _test():
...
+    if type(X) != np.ndarray or type(y) != np.ndarray:
+        raise TransformationError('Input arrays should be numpy ndarrays')
```
zhichao-li: I think simply use `Exception` is fine here as we haven't have uniform customized exception for now.

pyspark/bigdl/util/common.py
zhichao-li: add another parameter: `numSlices=None` to control the partition num?

gaarsmu: I've fixed two issues you've mentioned, but got a question about this one:
Is this partition number a pre-build function of Spark Contex, of you want to to simple add it up manually? (like slice ndarrays to partitions and parallelize each of them)

zhichao-li:
there's a similar function: `to_sample_rdd` to fix this.

PR #1675, opened 2017-10-18
support more tensor operations

yiheng:

This is a subPR of #1656. In this PR, we add more tensor operations

## How was this patch tested?
unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
@@ -932,7 +932,7 @@ private[tensor] class DenseTensor[@specialized T: ClassTag](
     this
   }
 
-  private def expandTensor(x: Tensor[T]): Tensor[T] = {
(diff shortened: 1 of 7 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
```diff
@@ -1218,7 +1218,7 @@ private[tensor] class DenseTensor[@specialized T: ClassTag](
     DenseTensorMath.addmm(this, v1, this, v2, mat1, mat2)
 
   override def mm(mat1: Tensor[T], mat2: Tensor[T]): Tensor[T] =
-    DenseTensorMath.addmm(this, ev.fromType[Int](1), this, ev.fromType[Int](1), mat1, mat2)
+    DenseTensorMath.addmm(this, ev.zero, this, ev.fromType[Int](1), mat1, mat2)
```
yangw1234: Why change **1** to **0**

yiheng: It's a bug of mm

yiheng:
PR validation pass

PR #1681, opened 2017-10-19
fix SparseTensor's resize

qiuxin2012:

sparseTensor's resize may throw exception

## How was this patch tested?

unit tests


qiuxin2012:
jenkins passed

PR #1692, opened 2017-10-20
fix: make the length of weight and gradWeight in quantized conv same length

i8run:
fix: the different length of weight and gradWeight in quantized conv is
very confused. So make them same length.

Jenkins

## How was this patch tested?


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/quantized/SpatialConvolution.scala
```diff
@@ -198,7 +198,7 @@ private[bigdl] class SpatialConvolution[T: ClassTag](
   }
 
   override def parameters(): (Array[Tensor[T]], Array[Tensor[T]]) = {
-    (weight :+ bias, Array(empty, empty))
+    (weight :+ bias, Array.fill[Tensor[T]](nGroup)(empty) :+ empty)
```
SeaOfOcean: Array.fill[Tensor[T]](nGroup + 1)(empty)

PR #1694, opened 2017-10-20
Add new API to session to allow user train a tensorflow graph directly on BigDL

yiheng:
As the title descript

## How was this patch tested?
unit test, manual test

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
@@ -289,8 +289,7 @@ object ArraySample {
         featureTensors: Array[Tensor[T]])(implicit ev: TensorNumeric[T]) : Sample[T] = {
     val data = new Array[T](featureTensors.map(_.nElement()).sum)
     copy(data, featureTensors)
-    new ArraySample[T](featureTensors.flatMap(_.storage().array()),
-      getSize(featureTensors), null)
+    new ArraySample[T](data, getSize(featureTensors), null)
   }
 
   private def copy[T: ClassTag](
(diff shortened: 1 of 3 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
```diff
@@ -289,8 +289,7 @@ object ArraySample {
         featureTensors: Array[Tensor[T]])(implicit ev: TensorNumeric[T]) : Sample[T] = {
     val data = new Array[T](featureTensors.map(_.nElement()).sum)
     copy(data, featureTensors)
-    new ArraySample[T](featureTensors.flatMap(_.storage().array()),
-      getSize(featureTensors), null)
+    new ArraySample[T](data, getSize(featureTensors), null)
```
yiheng: @qiuxin2012 Please take a look at this change.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/Session.scala
```diff
@@ -75,12 +123,63 @@ class BigDLSessionImpl[T: ClassTag](graph: Seq[NodeDef], context: Context[T],
...
+    context.setAssignGrads(weightsAndGrads)
+    val modelOutputs = if (loss.isDefined) {
+      Seq(loss.get) ++ weightsAndGrads.map(_._2).toSeq
+    } else {
+      weightsAndGrads.map(_._2).toSeq
+    }
+    val (model, input) = constructModel(modelOutputs, byteOrder, false, Some(context))
+    val a = getRDD(Seq(input.element.getName), sc, isDataBatch).first()
+    val data = BigDLSessionImpl.toSample[T](getRDD(Seq(input.element.getName), sc, isDataBatch))
+    val d = data.first()
```
yangw1234: Variable `a` and `d` is not used

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/TensorflowLoader.scala
```diff
@@ -311,8 +319,6 @@ object TensorflowLoader{
               (builder.build[T](n.element, byteOrder, context), Seq(n).asJava, Seq(n))
             } catch {
               case _ =>
-                println("com.intel.analytics.bigdl.utils.tf.loaders." +
-                  n.element.getOp)
                 throw new UnsupportedOperationException(errorMsg)
```
yangw1234: maybe we should also put this exception in the error log, like
```scala
case e: Exception =>
    throw new UnsupportedOperationException(errorMsg, e)
```

PR #1699, opened 2017-10-23
CaffeLoader supports multiple inputs and some additional outputs

SeaOfOcean:

CaffeLoader supports multiple inputs and some additional outputs

## How was this patch tested?

manual test with faster rcnn model

fixed https://github.com/intel-analytics/BigDL/issues/1691

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/caffe/CaffeLoader.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/caffe/CaffeLoader.scala
@@ -262,19 +262,67 @@ class CaffeLoader[T: ClassTag](prototxtPath: String, modelPath: String,
 /**
  * Load caffe model from prototxt file and binary pre-trained model and converted
  * to BigDL graph module
+ * @param outputNames additional output layer names besides the default(layers without next nodes)
  * @return BigDL model and criterion
  */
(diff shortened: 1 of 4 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/caffe/CaffeLoader.scala
```diff
@@ -262,19 +262,67 @@ class CaffeLoader[T: ClassTag](prototxtPath: String, modelPath: String,
...
+    addInputList(inputs, layers, top2LayerMap, layersMap)
+  }
+
+  private def addInputList(inputs: Seq[ModuleNode[T]],
+    layers: ArrayBuffer[ModuleNode[T]],
+    top2LayerMap: mutable.HashMap[String, String],
+    layersMap: mutable.HashMap[String, ModuleNode[T]]): Boolean = {
+    if (null != inputs) {
+      inputs.foreach(input => {
+        top2LayerMap(input.element.getName()) = input.element.getName()
```
wzhongyuan: So 'top' for input layer will be always the same as its name ?

wzhongyuan: Got it from you conveter logic below :)

wzhongyuan:
LGTM

SeaOfOcean:
pr validate pass

PR #1700, opened 2017-10-23
Fix fromCaffeEltwise and fromCaffeBatchNorm

SeaOfOcean:

Fix fromCaffeEltwise and fromCaffeBatchNorm

## How was this patch tested?

Manual test with Faster rcnn


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/caffe/Converter.scala
```diff
@@ -637,5 +637,6 @@ abstract class Converter[T: ClassTag](implicit ev: TensorNumeric[T]) {
     caffe2BigDL("ELTWISE") = fromCaffeEltwise
     caffe2BigDL("INPUT") = null
```
wzhongyuan: Need to merge with your previous change

wzhongyuan:
LGTM

SeaOfOcean:
pr validate pass

PR #1707, opened 2017-10-23
Refine textclassifier with data without headers and refactor to temporal convolution 

psyyz10:

Refine textclassifier with data without headers and refactor to temporal convolution

### DIFF
--- a/pyspark/bigdl/dataset/news20.py
+++ b/pyspark/bigdl/dataset/news20.py
@@ -19,17 +19,17 @@ from bigdl.dataset import base
 import os
 import sys
 
-NEWS20_URL = 'http://qwone.com/~jason/20Newsgroups/20news-19997.tar.gz'  # noqa
+NEWS20_URL = 'http://qwone.com/~jason/20Newsgroups/20news-18828.tar.gz'  # noqa
 GLOVE_URL = 'http://nlp.stanford.edu/data/glove.6B.zip'  # noqa
 
 CLASS_NUM = 20
 
 
 def download_news20(dest_dir):
-    file_name = "20news-19997.tar.gz"
+    file_name = "20news-18828.tar.gz"
     file_abs_path = base.maybe_download(file_name, dest_dir, NEWS20_URL)
     tar = tarfile.open(file_abs_path, "r:gz")
-    extracted_to = os.path.join(dest_dir, "20_newsgroups")
+    extracted_to = os.path.join(dest_dir, "20news-18828")
     if not os.path.exists(extracted_to):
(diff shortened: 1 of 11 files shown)


pyspark/bigdl/dataset/news20.py
```diff
@@ -19,17 +19,17 @@
 import os
 import sys
 
-NEWS20_URL = 'http://qwone.com/~jason/20Newsgroups/20news-19997.tar.gz'  # noqa
+NEWS20_URL = 'http://qwone.com/~jason/20Newsgroups/20news-18828.tar.gz'  # noqa
```
zhichao-li: so 18828 is the datasource without header and label?

pyspark/bigdl/nn/layer.py
```diff
@@ -1079,7 +1079,7 @@ class TemporalMaxPooling(Layer):
 
     def __init__(self,
                  k_w,
-                 d_w,
+                 d_w=-1,
```
zhichao-li: add docs for the meaning of -1?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/NNPrimitive.scala
```diff
@@ -1006,9 +1006,9 @@ private[nn] object NNPrimitive {
 
       var y = 0
       while (y < frameSize) {
-        val maxIndex = xp + y
+        val maxIndex = indices(xp + y).toInt - 1
```
zhichao-li: why changing this? do we need to add unitest to cover it?

pyspark/bigdl/dataset/news20.py
psyyz10: Yes

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/NNPrimitive.scala
psyyz10: We have unit test, this just a bug, I fixed it, keep float same as double

zhichao-li: ok. just wondering why the existing unittest didn't capture this bug.

zhichao-li:
LGTM

zhichao-li:
btw, do we need to update the docs accordingly? i.e: update the accuracy

zhichao-li:
any update for this?

zhichao-li:
LGTM pending on validation

PR #1708, opened 2017-10-23
Support reading TFRecord and raw binary files from HDFS in tensorflow loader

yangw1234:

Support reading TFRecord files from HDFS
Support reading raw binary files from HDFS

## How was this patch tested?

Add an integration test and manually tested

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/Session.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/Session.scala
@@ -33,6 +33,7 @@ import com.google.protobuf.ByteString
 import com.intel.analytics.bigdl.models.utils.ModelBroadcast
 import TFTensorNumeric.NumericByteString
 import com.intel.analytics.bigdl.utils.tf.BigDLSessionImpl.FakeCriterion
+import org.apache.hadoop.io.{BytesWritable, NullWritable}
 
 import scala.collection.mutable
 import scala.collection.mutable.ArrayBuffer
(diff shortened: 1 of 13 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/Session.scala
```diff
@@ -254,27 +255,59 @@ class BigDLSessionImpl[T: ClassTag](graph: Seq[NodeDef], context: Context[T],
...
-      iter
-    }.map { record =>
-      val table = T()
-      val key = Tensor[ByteString](Array(ByteString.copyFromUtf8("somekey")), Array[Int]())
-      val value = Tensor[ByteString](Array(ByteString.copyFrom(record)), Array[Int]())
-      table.insert(key)
-      table.insert(value)
-      table
+    if (filesTable.isEmpty) {
+      return sc.parallelize(Seq.empty[Table], numSlices = Engine.coreNumber())
```
yiheng: Engine.coreNumber() * Engine.nodeNumber()?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/Session.scala
```diff
@@ -254,27 +255,59 @@ class BigDLSessionImpl[T: ClassTag](graph: Seq[NodeDef], context: Context[T],
...
+    }
+    val fileNames = filesTable.map { t =>
+      require(t.length() == 1 && t(1).isInstanceOf[Tensor[ByteString]],
```
yiheng: even if t(1) is Tensor[Float, ]t(1).isInstanceOf[Tensor[ByteString]] will also return true. Try to use the getType to check the numeric type.

yiheng: Two error cause, but in the message you only give one.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/Session.scala
```diff
@@ -254,27 +255,59 @@ class BigDLSessionImpl[T: ClassTag](graph: Seq[NodeDef], context: Context[T],
...
+        "Reader can only read one file at a time")
+      val fileTensor = t[Tensor[ByteString]](1)
+      require(fileTensor.isScalar)
```
yiheng: error message

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/Session.scala
```diff
@@ -254,27 +255,59 @@ class BigDLSessionImpl[T: ClassTag](graph: Seq[NodeDef], context: Context[T],
...
+          val key = Tensor.scalar[ByteString](ByteString.copyFromUtf8("fake_key"))
+          val value = Tensor.scalar[ByteString](ByteString.copyFrom(k.copyBytes()))
+          table.insert(key)
+          table.insert(value)
+          table
+        }
+      }.reduceLeft(_.union(_))
+    } else {
+      // all files are local files
+      val result = fileNames.flatMap { file =>
```
yiheng: what if some are hdfs and some are local

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/Session.scala
```diff
@@ -254,27 +255,59 @@ class BigDLSessionImpl[T: ClassTag](graph: Seq[NodeDef], context: Context[T],
...
+        iter
+      }.map { record =>
+        val table = T()
+        val key = Tensor.scalar[ByteString](ByteString.copyFromUtf8("fake_key"))
+        val value = Tensor.scalar[ByteString](ByteString.copyFrom(record))
+        table.insert(key)
+        table.insert(value)
+        table
+      }
+      sc.parallelize(result, numSlices = Engine.coreNumber())
```
yiheng: corenumber * nodeNumber

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/Session.scala
```diff
@@ -285,31 +318,54 @@ class BigDLSessionImpl[T: ClassTag](graph: Seq[NodeDef], context: Context[T],
...
+          val table = T()
+          val key = Tensor.scalar[ByteString](ByteString.copyFromUtf8("fake_key"))
+          val value = Tensor.scalar[ByteString](ByteString.copyFrom(bytes))
+          table.insert(key)
+          table.insert(value)
+          table
+        }
+      }.reduceLeft(_.union(_))
+    } else {
+      val result = fileNames.flatMap { file =>
```
yiheng: some are local and some are hdfs?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/Session.scala
```diff
@@ -285,31 +318,54 @@ class BigDLSessionImpl[T: ClassTag](graph: Seq[NodeDef], context: Context[T],
...
+        iter
+      }.map { record =>
+        val table = T()
+        val key = Tensor[ByteString](Array(ByteString.copyFromUtf8("fake_key")), Array[Int]())
+        val value = Tensor[ByteString](Array(ByteString.copyFrom(record)), Array[Int]())
+        table.insert(key)
+        table.insert(value)
+        table
+      }
+      sc.parallelize(result, numSlices = Engine.coreNumber())
```
yiheng: corenumber * node number?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/Session.scala
yangw1234: It will throw an exception at line `val isHdfs = allHDFSFiles(fileNames)`. Is that alright?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/Session.scala
yangw1234: The same as above.

yangw1234:
@yiheng please review again

yangw1234:
pr validation pass

PR #1720, opened 2017-10-24
hot-fix jtensor

qiuxin2012:

serialize sparse jtensor will throw exception, when create RDD of sparse JTensor

## How was this patch tested?

manual tests


qiuxin2012:
jenkins passed

PR #1737, opened 2017-10-25
fix shape output tensor type from T to Int

yiheng:
Shape layer output should be Int instead of T

## How was this patch tested?
unit test


yangw1234:
LGTM

Issue #1445, opened 2017-08-11
add documentation for customizing layers and criterions


shane-huang:

lseffer:
Until these documentations are available, could you provide a minimal example of how to adding a custom layer and criterion? Using pyspark would it be enough to extend the Python classes?

yiheng:
@lseffer Currently, we do not support add customized layer by extending the Python classes. You need to extend the scala classes and wrapper it as Python class...

You need to extend ```AbstractModule``` class to define your own Layer. 
Here's an example of a layer which add a scalar to the input tensor.(Just need to look at the ```updateOutput``` and ```updateGradInput``` methods, they forward and backward the data)
https://github.com/intel-analytics/BigDL/blob/master/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/AddConstant.scala


To define your own loss function, you need to extend the ```AbstractCriterion``` class.
Here's an example of use avg of absolute value of the element-wise difference as loss
https://github.com/intel-analytics/BigDL/blob/master/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/AbsCriterion.scala

PR #1739, opened 2017-10-25
add layer criterion doc

yiheng:
add a document to descript how to add new layer/criterion

## How was this patch tested?
only doc change

fixed https://github.com/intel-analytics/BigDL/issues/1445

### DIFF
--- a/docs/docs/ProgrammingGuide/add-layer-criterion.md
+++ b/docs/docs/ProgrammingGuide/add-layer-criterion.md
@@ -193,6 +193,11 @@ Reset gradParameters to zero.
 
 ReInitialize the parameter and reset gradParameter to zero
 
+You may notice that in the layer we don't define how to use g to update w. Layers just provide
+gradient and parameters. How to use gradient to update parameter is handled by
+[optimize methods](APIGuide/Optimizers/Optim-Methods.md) and
+[optimizer](APIGuide/Optimizers/Optimizer.md).
+
 ## Create a criterion in scala
 To create your own criterion, you need to extend the **AbstractCriterion** class. The criterion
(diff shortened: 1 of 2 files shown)


docs/docs/ProgrammingGuide/add-layer-criterion.md
```diff
@@ -0,0 +1,236 @@
...
+    g.zero()
+  }
+
+  override def reset(): Unit = {
+      w.rand()
+      g.zero()
+  }
+```
+First we introduce a new tensor g, which has same size with parameter w. We call g gradParameter, as
+it stores the gradients which will be used to update w.
```
shane-huang: How is w updated using g? Maybe we should add some elaboration here (reference to optimizer related docs).

docs/mkdocs.yml
```diff
@@ -45,6 +45,7 @@ pages:
   - Tensorflow Support: ProgrammingGuide/tensorflow-support.md
   - Caffe Support: ProgrammingGuide/caffe-support.md
   - Quantization Support: ProgrammingGuide/quantization-support.md
+  - Create Layer/Criterion: ProgrammingGuide/add-layer-criterion.md
```
shane-huang: May be we should change the title of this section to "Define new Layer/Criterion" or "Customizated Layer/Criterion", and similar. "Create layer/criterion" is a little bit confusing as it may means to create a layer or criterion object.

docs/docs/ProgrammingGuide/add-layer-criterion.md
yiheng: done

Issue #1743, opened 2017-10-25
TreeLSTM example will throw exception and loss will be NaN


i8run:
```
Exception in thread "main" java.lang.IllegalArgumentException: requirement failed: copyStatus:container's length mismatchexcepted 1, but get 0
        at scala.Predef$.require(Predef.scala:224)                                                                                                                                             
        at com.intel.analytics.bigdl.nn.Container.copyStatus(Container.scala:143)
        at com.intel.analytics.bigdl.nn.Container.copyStatus(Container.scala:40)                                                                                                               
        at com.intel.analytics.bigdl.nn.Container.copyStatus(Container.scala:148)           
        at com.intel.analytics.bigdl.nn.Container.copyStatus(Container.scala:40)                                                                                                               
        at com.intel.analytics.bigdl.nn.Utils$.copyModule(Utils.scala:238)
        at com.intel.analytics.bigdl.optim.DistriOptimizer.optimize(DistriOptimizer.scala:859)                                                                                                 
        at com.intel.analytics.bigdl.example.treeLSTMSentiment.Train$.train(Train.scala:124)
        at com.intel.analytics.bigdl.example.treeLSTMSentiment.Train$.main(Train.scala:40)                                                                                                     
        at com.intel.analytics.bigdl.example.treeLSTMSentiment.Train.main(Train.scala)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)                                                                                                                         
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)                                                                                               
        at java.lang.reflect.Method.invoke(Method.java:497)
        at org.apache.spark.deploy.SparkSubmit$.org$apache$spark$deploy$SparkSubmit$$runMain(SparkSubmit.scala:729)                                                                            
        at org.apache.spark.deploy.SparkSubmit$.doRunMain$1(SparkSubmit.scala:185)
        at org.apache.spark.deploy.SparkSubmit$.submit(SparkSubmit.scala:210)                                                                                                                  
        at org.apache.spark.deploy.SparkSubmit$.main(SparkSubmit.scala:124)
        at org.apache.spark.deploy.SparkSubmit.main(SparkSubmit.scala)                                                 
```

PR #1763, opened 2017-10-26
Clone graph keep their original prevs nodes and next nodes sequence

SeaOfOcean:

Clone graph keep their original prevs nodes and next nodes order.
Current cloneGraph may change the next nodes order, thus will influence the update of tree-lstm (it has many shared component)

In the meantime, fix a bug in graph backward that caused by the following ut

"Graph backward" should "be correct when contains nested output"

## How was this patch tested?

unit test
test over tree lstm training

fixed https://github.com/intel-analytics/BigDL/issues/1743

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
@@ -509,14 +509,17 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
         activity.toTensor[T].add(other.toTensor[T])
       } else {
         val actTable = activity.toTable
-        val actLen = actTable.length()
         val otherTable = other.toTable
-        val otherLen = otherTable.length()
-        require(actLen == otherLen, "table length is not equal")
+        val maxActIndex = actTable.keySet.map(_.asInstanceOf[Int]).max
+        val maxOtherIndex = otherTable.keySet.map(_.asInstanceOf[Int]).max
+        val maxIndex = Math.max(maxActIndex, maxOtherIndex)
         var i = 1
-        while(i <= actLen) {
-          require(actTable[Activity](i) != null, "Invalid table element")
-          accActivity(actTable[Activity](i), otherTable[Activity](i))
+        while(i <= maxIndex) {
+          if (actTable.contains(i) && otherTable.contains(i)) {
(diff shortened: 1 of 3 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DirectedGraph.scala
```diff
@@ -148,12 +148,17 @@ class DirectedGraph[T](val source : Node[T], val reverse : Boolean = false) exte
     })
     bfs.foreach(node => {
       if (reverseEdge) {
-        node.prevNodesAndEdges.foreach(prevNodeAndEdge => {
-          oldToNew.get(node).add(oldToNew.get(prevNodeAndEdge._1), prevNodeAndEdge._2)
+        node.nextNodesAndEdges.foreach(nextNodeAndEdge => {
+          // some node may be ignored
```
yiheng: Please add more comments here. e.g.
// keep the order in the nextNodes array in the current node. As we go through all node in bfs, the prevNodes order should be preserved.
// Some next nodes may be not included in the graph

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -509,14 +509,17 @@ class Graph[T: ClassTag](val inputs : Seq[ModuleNode[T]],
         activity.toTensor[T].add(other.toTensor[T])
       } else {
         val actTable = activity.toTable
-        val actLen = actTable.length()
         val otherTable = other.toTable
-        val otherLen = otherTable.length()
-        require(actLen == otherLen, "table length is not equal")
+        val maxActIndex = actTable.keySet.map(_.asInstanceOf[Int]).max
```
yiheng: try to use length instead of keySet.map

SeaOfOcean:
pr validate pass

yiheng:
LGTM. Please merge it and backport to branch-0.3

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
yiheng: otherTable.keySet.foreach?

SeaOfOcean:
pr validate pass

yiheng:
LGTM. Please merge and backport

PR #1803, opened 2017-10-30
update pip install bigdl version

hkvision:

Update pip install bigdl version for branch-0.2.


jason-dai:
LGTM

PR #1811, opened 2017-10-31
checkin some test utils for tensorflow operation loader

yiheng:
checkin some test utils for tensorflow operation loader. It will compare the output from a single operation graph executed by BigDL and Tensorflow.

This PR also refine the test utility methods

## How was this patch tested?
unit test

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Table.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Table.scala
@@ -339,8 +339,8 @@ object T {
    * @param data
    * @return
    */
-  def array(data: Array[Any]): Table = {
-    new Table(data)
+  def array(data: Array[_]): Table = {
+    new Table(data.asInstanceOf[Array[Any]])
   }
 
   /**
@@ -351,8 +351,8 @@ object T {
    * @param data
    * @return
    */
-  def seq(data: Seq[Any]): Table = {
-    new Table(data.toArray)
+  def seq(data: Seq[_]): Table = {
+    new Table(data.toArray.asInstanceOf[Array[Any]])
   }
 
   /**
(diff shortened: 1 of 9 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Table.scala
```diff
@@ -339,8 +339,8 @@ object T {
    * @param data
    * @return
    */
-  def array(data: Array[Any]): Table = {
-    new Table(data)
+  def array(data: Array[_]): Table = {
```
yiheng: make the API easier to use

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/BigDLToTensorflow.scala
```diff
@@ -139,8 +139,7 @@ object TemporalConvolutionToTF extends BigDLToTensorflow {
     require(inputs.length == 1, "SpatialConvolution only accept one input")
     val spatialConv = module.asInstanceOf[TemporalConvolution[_]]
 
-    val const1 = const(Tensor[Float](1).setValue(1, 1.0f),
-      spatialConv.getName() + "/dim1", byteOrder, true, DataType.DT_INT32)
+    val const1 = const(Tensor.scalar[Int](1), spatialConv.getName() + "/dim1", byteOrder)
```
yiheng: refine the const method

yiheng:
PR validation pass.

spark/dl/src/test/scala/com/intel/analytics/bigdl/utils/tf/TensorflowSpecHelper.scala
```diff
@@ -57,4 +62,74 @@ class TensorflowSpecHelper extends FlatSpec with Matchers with BeforeAndAfter {
...
+    val bigdlOutput = runGraphBigDL(graphFile, nodeDefBuilder.getName)
+    val bigdlOutputTensor = if (bigdlOutput.isTensor) {
+      require(outputIndex == 0, s"invalid output index $outputIndex")
+      bigdlOutput.asInstanceOf[Tensor[_]]
+    } else {
+      bigdlOutput.toTable.apply[Tensor[_]](outputIndex + 1)
+    }
+    val tfOutput = runGraphTF(graphFile, nodeDefBuilder.getName + s":$outputIndex")
+    bigdlOutput.asInstanceOf[Tensor[NumericWildCard]]
+      .equals(tfOutput.asInstanceOf[Tensor[NumericWildCard]]) should be(true)
```
yangw1234: Maybe we should use almostEqual and let the caller pass in the tolerance?

yangw1234:
LGTM except one minor comment

Issue #1697, opened 2017-10-20
CMul layer will crash for certain configure  


SeaOfOcean:
```
val cmul = CMul[Float](Array[Int](1, 4096, 1, 1))
cmul.weight.randn()
val input = Tensor[Float](300, 4096).randn()
cmul.forward(input)
```
```
It is because the input size is not valid, need some check
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007f591c84b006, pid=24589, tid=140021184374528
#
# JRE version: Java(TM) SE Runtime Environment (7.0_79-b15) (build 1.7.0_79-b15)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (24.79-b02 mixed mode linux-amd64 compressed oops)
# Problematic frame:
# C  [dlNativeLoader1509284193629917815libjmkl.so+0x2d5006]  mkl_blas_avx_xsscal+0x236
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /home/jxy/code/analytics-zoo/pipeline/ssd/hs_err_pid24589.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#
```

PR #1822, opened 2017-11-02
Fix cmul

lopelopelope:
fixed https://github.com/intel-analytics/BigDL/issues/1697

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/CMul.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/CMul.scala
@@ -82,7 +82,10 @@ class CMul[T: ClassTag](
   }
 
   private def mulOneDimWeight(dim: Int, expand: Tensor[T], output: Tensor[T]): Unit = {
-    val (innerNum, outerNum) = Utils.getInnerOuterNum(dim, output)
+    val multiplyDimSize = expand.size(dim)
+    var dimOfoutput : Int = 1
+    while(output.size(dimOfoutput) != multiplyDimSize) dimOfoutput += 1
+    val (innerNum, outerNum) = Utils.getInnerOuterNum(dimOfoutput, output)
     val weightData = expand.storage().array()
     val weightOffset = expand.storageOffset() - 1
     var outer = 0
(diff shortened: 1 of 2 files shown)


spark/dl/src/test/scala/com/intel/analytics/bigdl/nn/CMulSpec.scala
```diff
@@ -44,6 +43,11 @@ class CMulSpec extends FlatSpec with Matchers {
     val gradInput1 = layer1.backward(input, gradOutput)
     val output2 = layer2.forward(input)
     val gradInput2 = layer2.backward(input, gradOutput)
+    val cmul = CMul[Float](Array[Int](1, 4096, 1, 1))
```
yiheng: This is not regular test code. You should verify the result.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/CMul.scala
```diff
@@ -82,7 +82,10 @@ class CMul[T: ClassTag](
   }
 
   private def mulOneDimWeight(dim: Int, expand: Tensor[T], output: Tensor[T]): Unit = {
-    val (innerNum, outerNum) = Utils.getInnerOuterNum(dim, output)
+    val multiplyDimSize = expand.size(dim)
+    var dimOfoutput : Int = 1
```
yiheng: dimOfoutput  -> dimOfOutput

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/CMul.scala
```diff
@@ -82,7 +82,10 @@ class CMul[T: ClassTag](
...
+    while(output.size(dimOfoutput) != multiplyDimSize) dimOfoutput += 1
```
yiheng: What if there's no multiplyDimSize in output.size()? TTry to catch the array outofbound error and throw a more meaningful error msg

yiheng:
What about the jenkins test result?

lopelopelope:
It is still under test. I will contact you when it is done @yiheng

lopelopelope:
pass jenkins @yiheng

Issue #1500, opened 2017-08-23
Make the jar file and python zip file name with consistent version in BigDL release


jenniew:
In the documentation of running local mode without pip install: https://bigdl-project.github.io/master/#PythonUserGuide/run-without-pip/#a-quick-launch-for-local-mode

cd $BIGDL_HOME/dist/lib 
BIGDL_VERSION=...
${SPARK_HOME}/bin/pyspark --master local[4] \
--conf spark.driver.extraClassPath=bigdl-${BIGDL_VERSION}-jar-with-dependencies.jar \
--py-files bigdl-${BIGDL_VERSION}-python-api.zip \
--properties-file ../conf/spark-bigdl.conf 

If a user install BigDL by downloading release, replacing BIGDL_VERSION is not working for both jar file and python zip file. I.E. jar is: bigdl-SPARK_1.6-0.2.0-jar-with-dependencies.jar, python zip is: bigdl-0.2.0-python-api.zip.
We need to make the jar and zip file name consistent in release.

yiheng:
I think it's the document problem instead of the version problem.

@zhichao-li @jason-dai  The python start command is too complex and error prone. Should we refactor them by providing some script?

See https://github.com/intel-analytics/BigDL-Tutorials/blob/master/start_notebook.sh

PR #1832, opened 2017-11-03
Refine python run doc

yiheng:
I notice that the pyspark commands are too long in the python run without pip page. User need to copy a lot of commands, and need to update them before run. It's not user friendly and error prone. 

I put these long commands into some scripts. User just need to set two env var SPARK_HOME and BIGDL_HOME. Then can run it. Also update the document of run python without pip.

I think this will save a lot of time for user and developer.

## How was this patch tested?
manual test

fixed https://github.com/intel-analytics/BigDL/issues/1500


docs/docs/PythonUserGuide/run-without-pip.md
```diff
@@ -1,101 +1,70 @@
-## **Precondition**
-First of all, you need to obtain the BigDL libs. Refer to [Install from pre built](../ScalaUserGuide/install-pre-built.md) or [Install from source code](../ScalaUserGuide/install-build-src.md) for more details
```
zhichao-li: Keep this? as pip install would not always meet user's requirement. i.e run in cluster mode (at least pip doesn't support cluster mode for now)

zhichao-li:
LGTM Thanks for the update.To some extend, it would make life more easier.

PR #1839, opened 2017-11-07
use shade to package jar-with-dependencies

### LINKED ISSUE #1833: BigDL throws exceptions if it run on a platform whose dependency conflict with BigDL's
Tried to run BigDL on customer's platform and found their dependent google protobuf is conflict with BigDL's. Although we have shaded protobuf, but our assembled jar still includes google protobuf which will cause the exception. Currently I need exclude them manually from bigdl jar to avoid the exception. I think it would be better if we exclude google protobuf when assembing bigdl jar.

qiuxin2012:

use shade to package jar-with-dependencies to exclude protobuf.

## How was this patch tested?

unit tests

fixed #1833

### DIFF
--- a/spark/dl/pom.xml
+++ b/spark/dl/pom.xml
@@ -167,11 +167,6 @@
                 <artifactId>maven-shade-plugin</artifactId>
                 <version>3.0.0</version>
                 <configuration>
-                    <artifactSet>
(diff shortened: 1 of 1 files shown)


qiuxin2012:
jenkins passed

PR #1858, opened 2017-11-09
Support load more tensorflow operations

yiheng:
Rint, All, Any, Range, Exp, Expm1, Round, FloorDiv, TruncateDiv, Mod, FloorMod, TruncateMod, IntopK, Round
## How was this patch tested?
Add more unit test


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/FloorDiv.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  output = Tensor[D]()
+
+  override def updateOutput(input: Table): Tensor[D] = {
+    val input1 = input[Tensor[D]](1)
+    val input2 = input[Tensor[D]](2)
+    output.resizeAs(input1).copy(input1)
+    output.div(input2)
+    output.floor()
+    output
```
yangw1234: Not really sure this is correct, because the result of `math.floor(5 / -2)` is `-2` and  the result of `floordiv` should be `-3`

yangw1234:
LGTM expect the `FloorDiv` op

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/FloorDiv.scala
yiheng: It works for float / double but I didn't test Int... Good catch.

yiheng:
PR validation pass

Issue #1813, opened 2017-11-01
Adapter layer does not support different batch size at runtime


wzhongyuan:
Adapter layer would throw exception as below if feed with inputs of different batch sizes.

```scala
java.lang.IllegalArgumentException: requirement failed: constant tensor is changed.
```
Below is the code to recreate this issue by loading inception V3 and inferring with two inputs which have different batch sizes(10 and 2), there would be no issue if all inputs have the same batch size

```scala
    val model = Sequential[Float]().add(ShuffleWithPermutation(Array(1, 3, 4, 2)))
      .add(Module.loadTF[Float](
      "/home/jerry/data/model_zoo/inceptionv3/frozen_inception_v3.pb",
      Seq("input"), Seq("InceptionV3/Predictions/Reshape_1")))
     model.forward(Tensor[Float](10, 3, 299, 299).randn())
    model.forward(Tensor[Float](2, 3, 299, 299).randn())
```

Issue #1815, opened 2017-11-01
Adaper layer cannot be serialized directly


wzhongyuan:
```
class Adapter[T: ClassTag](
  val configIndexes: Array[Int],
  val build: Array[Tensor[_]] => AbstractModule[Activity, Activity, T]
)(implicit ev: TensorNumeric[T])
```

build function needs to be converted to be a meaningful format (mapping or build result ?)

PR #1859, opened 2017-11-09
fix adapter issue

yiheng:
This PR fixes two issues of Adapter:
1. When the configuration tensors of the adapter changes, an error will be thrown. It should rebuild the module
2. The adapter can't be serialized and de-serialized, as it contains a function member. This is fixed by use class inherent instead of function member

## How was this patch tested?
unit test
fixed https://github.com/intel-analytics/BigDL/issues/1813 and https://github.com/intel-analytics/BigDL/issues/1815


yiheng:
jenkins pass

wzhongyuan:
@yiheng  Can we add serialization tested for tf related layer/ops ? Maybe we need to open another issue

yangw1234:
LGTM

yiheng:
@wzhongyuan I prefer to open a new issue to track this. https://github.com/intel-analytics/BigDL/issues/1868

PR #1862, opened 2017-11-10
support SAME padding in 3d conv and allows user config padding size in convlstm and convlstm3d

dding3:
1. allows user config padding size in convlstm and convlstm3d
2. support SAME padding in 3d conv
3. set name of conv layers in convlstm/convlstm

## How was this patch tested?
unit test


yangw1234:
LGTM

One minor thing is that, maybe we can make the utils function like `getOutSizeAndPadding` to calculate the padding of each dimension separately and this may simplify the code a little bit. Like the tensorflow function [GetWindowedOutputSizeVerboseV2](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/common_shape_fns.cc#L20). It's your call.

PR #1865, opened 2017-11-10
Support keras squared_hinge and sparse_categorial_crossentropy

yangw1234:

Support keras squared_hinge and sparse_categorial_cross_entropy

squared_hinge -> MarginCriterion(1.0, True, True)
sparse_categorial_crossentropy -> ClassNLLCriterion(null, True, False)

## How was this patch tested?

added unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ClassNLLCriterion.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ClassNLLCriterion.scala
@@ -30,12 +30,12 @@ import com.intel.analytics.bigdl.utils.Engine
  * classes. If provided, the optional argument weights should be a 1D Tensor assigning weight to
  * each of the classes. This is particularly useful when you have an unbalanced training set.
  *
- * The input given through a forward() is expected to contain log-probabilities of each class:
- * input has to be a 1D Tensor of size n. Obtaining log-probabilities in a neural network is easily
- * achieved by adding a LogSoftMax layer in the last layer of your neural network. You may use
- * CrossEntropyCriterion instead, if you prefer not to add an extra layer to your network. This
- * criterion expects a class index (1 to the number of class) as target when calling
- * forward(input, target) and backward(input, target).
+ * The input given through a forward() is expected to contain log-probabilities/probabilities of
+ * each class: input has to be a 1D Tensor of size n. Obtaining log-probabilities in a neural
+ * network is easily achieved by adding a LogSoftMax layer in the last layer of your neural
+ * network. You may use CrossEntropyCriterion instead, if you prefer not to add an extra layer
+ * to your network. This criterion expects a class index (1 to the number of class) as target
+ * when calling forward(input, target) and backward(input, target).
  *
  * The loss can be described as:
  *     loss(x, class) = -x[class]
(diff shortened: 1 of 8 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/MarginCriterion.scala
```diff
@@ -49,15 +56,23 @@ class MarginCriterion[@specialized(Float, Double) T: ClassTag]
   }
 
   override def updateGradInput(input: Tensor[T], target: Tensor[T]): Tensor[T] = {
-    val norm = ev.fromType(if (sizeAverage) -1.0 / input.nElement() else 1.0)
+    val norm = ev.fromType(if (sizeAverage) -1.0 / input.nElement() else -1.0)
```
yangw1234: It's a bug of the original code. `sizeAverage` should not change the sign.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ClassNLLCriterion.scala
```diff
@@ -111,7 +120,12 @@ class ClassNLLCriterion[@specialized(Float, Double) T: ClassTag]
           if (curTarget == -1) (ev.zero, ev.one)
           else {
             val curWeight = if (weights != null) weights.valueAt(curTarget) else ev.fromType[Int](1)
-            (ev.times(input.valueAt(_i, curTarget), curWeight), curWeight)
+            if (acceptProb) {
```
hkvision: The name of the parameter `acceptProb` seems a bit confusing?
If true, `loss = -log(x[class])`
Else, `loss = -x[class]`
Users may find ambiguous for this parameter? Since if it is false, we may assume that they are actually pass in probabilities but we don't accept probabilities?
Also, after adding the log, we assume x[class] is always positive? What if it is zero?

yangw1234: @hkvision I'll try to handle the zero case and what's your suggestions about the parameter name?

hkvision: @yangw1234 It seems in keras, when get zero, it will use precision probably treat it as ~1e-7 and calculate the result as the following:

>>> y_true = 3
>>> y_pred = np.array([0.1, 0.6, 0.3, 0])
>>> res = loss(K.variable(y_true),K.variable(y_pred))
>>> K.eval(res)
array([ 16.1180954], dtype=float32)

Do we have the same or similar mechanism in BigDL for this?

hkvision: @yangw1234 Also, in BigDL index starts from 1 and in Keras index starts from 0, will this be a problem?

hkvision: @yangw1234 For the name, I'm not sure lol. Maybe something like takeLog...?

yangw1234: @hkvision For the zero probability case, I'll clip the probability to make it between 0+epsilon and 1 - epsilon where epsilon is 1e-8. This is exactly how keras tensorflow backend handle it.
I think index inconsistence should be handled by the keras-bigdl converter.

spark/dl/src/test/scala/com/intel/analytics/bigdl/nn/ClassNLLCriterionSpec.scala
```diff
@@ -234,4 +234,52 @@ class ClassNLLCriterionSpec extends FlatSpec with Matchers {
...
+    val gradInput = softMax.backward(input, testedLayer.backward(prob, target))
+
+    math.abs(expectedLoss - loss) < 1e-5 should be (true)
+    expectedGradInput.almostEqual(gradInput, 1e-5) should be (true)
+  }
+
+  "A ClassNLL Criterion with probabilities input 1d" should "generate correct output and grad" in {
+
+    val input = Tensor[Float](Array(4)).rand()
+    val target = Tensor[Float](Array[Float](4), Array(1))
```
zhichao-li: wondering can we support one-hot format as the target value ?

yangw1234:
pr validation pass

Issue #1494, opened 2017-08-21
Need a remove/drop API to remove last module layer to support fine tune


jenniew:
After load model, we need to modify last layers of pre-trained model to fine tune with different dataset. Now we already have API to add module,  we need remove/drop module API to do that modification.  Another use case is that we need to remove last layers( i.e. Fully Connected layer, etc.) to only extract features without doing classification.

yiheng:
This request is ambiguous. What is the last module in a graph?

jenniew:
The top layer

jenniew:
Say, in topological sorting of directed graph, the last one.  Or we have a way to find a module with name, if it is possible to remove that module.

yiheng:
@jenniew @jason-dai @hhbyyh 

I think one solution to the problem is you can get the graph of the Graph model, modify the graph and create a new Graph model on it. We can add a getNodeByName method to Graph, then for your use case, we can

1. remove the last several layers, so now we get output from e.g. conv10
```scala
val output = graph.getNodeByName("conv10")
val input = graph.getNodeByName("input")
val newModel = Graph(input, output)
```

2. replace last layer 'fc10' with different fc layer, layer 'relu' point to 'fc10'
```scala
val node = graph.getNodeByName("relu")
val newOutput = Linear(...).inputs(node)
val input = graph.getNodeByName("input")
val newModel = Graph(intput newOutput)
```

PR #1871, opened 2017-11-10
Add an API to get node from graph model with given name

yiheng:
Add an API to get node from graph model with given name

## How was this patch tested?
unit test

fixed https://github.com/intel-analytics/BigDL/issues/1494


zhichao-li:
add it to python as well?

yiheng:
jenkins pass

PR #1874, opened 2017-11-13
Fix squeeze loader bug and some other refinement

yangw1234:

1. Fix squeeze loader bug and some other refinement
2. Fix saving tensorflow scalar variable error

## How was this patch tested?

added unit tests

### DIFF
--- a/pyspark/bigdl/util/tf_utils.py
+++ b/pyspark/bigdl/util/tf_utils.py
@@ -117,9 +117,14 @@ def save_variable_bigdl(tensors, target_path, bigdl_type="float"):
     :param bigdl_type: model variable numeric type
     :return: nothing
     """
+    import numpy as np
     jtensors = {}
     for tn in tensors.keys():
-        jtensors[tn] = JTensor.from_ndarray(tensors[tn])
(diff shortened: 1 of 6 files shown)


yangw1234:
pr validation pass

pyspark/bigdl/util/tf_utils.py
```diff
@@ -117,9 +117,14 @@ def save_variable_bigdl(tensors, target_path, bigdl_type="float"):
     :param bigdl_type: model variable numeric type
     :return: nothing
     """
+    import numpy as np
     jtensors = {}
     for tn in tensors.keys():
-        jtensors[tn] = JTensor.from_ndarray(tensors[tn])
+        if not isinstance(tensors[tn], np.ndarray):
```
yiheng: when tensors[tn] will be an array?

yangw1234: when the variable is a scalar, tensors[tn] is not a `np.ndarray` but a single number. `JTensor.from_ndarray()` will fail on the single number case.

natnaelino: Same here fix single number fail np.ndarary ?

PR #1878, opened 2017-11-14
Test utility to verify layer implementation against Keras

zhichao-li:
Method `checkOutputAndGrad` would try to compare forwarding output, gradInput, gradWeight against Keras. 
1. Ideally we only need to change the shape of `Input` and the model definition.
2. You can customize weight inintialization in keras which be used to reset the BigDL model. We use `uniform` in this example.
3. You need to define a `weightConverter` which try to convert keras parameters to BigDL format.
```
  "Layer with weights" should "be ok" in {
    val kerasCode =
      """
        |input_tensor = Input(shape=[3])
        |output_tensor = Dense(2, init='uniform')(input_tensor)
        |model = Model(input=input_tensor, output=output_tensor)
      """.stripMargin
    val linear = new Linear[Float](3, 2)
    def weightConverter(in: Array[Tensor[Float]]): Array[Tensor[Float]] = Array(in(0).t(), in(1))
    checkOutputAndGrad(linear, kerasCode, weightConverter)
  }
```
## How was this patch tested?
unittest


yiheng:
Anyone review this PR?

PR #1887, opened 2017-11-16
Update Recursive-Layers.md

recurse-id:
The first element in the Table is actually the input and the second element is the tree.

Minor inconsistency in the documentation.


psyyz10:
@recurse-id Thanks for your contribution.

PR #1894, opened 2017-11-16
add GaussianDropout, GaussianNoise and unit tests, python wrapper

shane-huang:

Added 2 noise layers for Keras:  GaussianDropout, GaussianNoise, with unit tests, python wrapper, serializatoin test.. 

## How was this patch tested?

PR validation test passed. 

Do we need compare with Keras? How to deal with random numbers?

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDL.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDL.scala
@@ -527,6 +527,16 @@ class PythonBigDL[T: ClassTag](implicit ev: TensorNumeric[T]) extends Serializab
     Dropout[T](initP, inplace, scale)
   }
 
+  def createGaussianDropout(rate: Double)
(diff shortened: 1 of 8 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDL.scala
```diff
@@ -527,6 +527,16 @@ class PythonBigDL[T: ClassTag](implicit ev: TensorNumeric[T]) extends Serializab
     Dropout[T](initP, inplace, scale)
   }
 
+  def createGaussianDropout(rate: Double)
```
zhichao-li: Expose this layer in python. please refer to `Linear` in "layer.py"

zhichao-li:
docs for bigdl-project.github.io ?

wzhongyuan:
LGTM

zhichao-li:
Wondering do we need to differentiate batch or non-batch input and add doc to specify the input format?
LGTM except for this minor one.

PR #1895, opened 2017-11-16
Add tensorflow transfer learning example

yangw1234:

Add tensorflow transfer learning example

## How was this patch tested?

existing tests and manually tested


spark/dl/src/main/scala/com/intel/analytics/bigdl/example/tensorflow/loadAndSave/Load.scala
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.intel.analytics.bigdl.example.tensorflow
+package com.intel.analytics.bigdl.example.tensorflow.loadAndSave
```
yiheng: package name should be all lower case

yiheng:
Look good to me. Also update this document https://bigdl-project.github.io/master/#ProgrammingGuide/tensorflow-support/?

yangw1234:
pr validation pass

PR #1899, opened 2017-11-17
Ignore keras test if environment is not ready and add test method for loss

zhichao-li:

Ignore keras test if environment is not ready and add test method for loss

## How was this patch tested?
unittest

### DIFF
--- a/spark/dl/src/test/scala/com/intel/analytics/bigdl/keras/KerasBaseSpec.scala
+++ b/spark/dl/src/test/scala/com/intel/analytics/bigdl/keras/KerasBaseSpec.scala
@@ -77,7 +77,7 @@ abstract class KerasBaseSpec extends BigDLSpecHelper {
       KerasRunner.run(kerasCode, is_loss = true)
 
     val boutput = bmodel.forward(input, target)
-    require(output.isScalar, s"output should be scalar, but we got: ${output}")
+    require(output.nElement() == 1, s"output should only contain 1 element, but we got: ${output}")
     NumericFloat.nearlyEqual(boutput, output.storage.array()(0), precision)
 
     val bgradInput = bmodel.backward(input, target.clone().fill(1))
--- a/spark/dl/src/test/scala/com/intel/analytics/bigdl/keras/KerasBaseSpec.scala
+++ b/spark/dl/src/test/scala/com/intel/analytics/bigdl/keras/KerasBaseSpec.scala
@@ -14,11 +14,28 @@
  * limitations under the License.
  */
 package com.intel.analytics.bigdl.keras
-import com.intel.analytics.bigdl.nn.abstractnn.AbstractModule
+import com.intel.analytics.bigdl.Criterion
+import com.intel.analytics.bigdl.nn.abstractnn.{AbstractCriterion, AbstractModule, Activity}
 import com.intel.analytics.bigdl.tensor.Tensor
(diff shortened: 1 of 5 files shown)


zhichao-li:
cc @yiheng  @wzhongyuan

spark/dl/src/test/scala/com/intel/analytics/bigdl/keras/KerasBaseSpec.scala
```diff
@@ -14,11 +14,28 @@
...
+import com.intel.analytics.bigdl.nn.abstractnn.{AbstractCriterion, AbstractModule, Activity}
 import com.intel.analytics.bigdl.tensor.Tensor
-import org.scalatest.{BeforeAndAfter, FlatSpec, Matchers}
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric.NumericFloat
+import org.scalatest.{BeforeAndAfterEach, FlatSpec, Matchers}
 
-class KerasBaseSpec extends FlatSpec with BeforeAndAfter with Matchers {
+import scala.sys.process._
+
+class KerasBaseSpec extends FlatSpec with BeforeAndAfterEach with Matchers {
```
yiheng: change to abstract? extend BigDLSpecHelper?

spark/dl/src/test/scala/com/intel/analytics/bigdl/keras/KerasBaseSpec.scala
```diff
@@ -14,11 +14,28 @@
...
+    // Skip unitest if environment is not ready
+    try {
+    Seq("python", "-c", "import keras; import tensorflow").!!
+    } catch {
+      case e: Throwable => cancel("python or keras or tensorflow is not installed", e)
+    }
+  }
+
+  override def beforeEach() {
+    ifskipTest()
```
yiheng: will this be run in each test. I remember if you call cancel before test case, the test will be marked as failure. It is marked as skipped when you call cancel in each test.

spark/dl/src/test/scala/com/intel/analytics/bigdl/keras/KerasBaseSpec.scala
```diff
@@ -57,5 +74,20 @@ class KerasBaseSpec extends FlatSpec with BeforeAndAfter with Matchers {
       }
     }
   }
+
+  def checkOutputAndGradForLoss(bmodel: AbstractCriterion[Tensor[Float], Tensor[Float], Float],
```
yiheng: only loss? how can we test layer?

yiheng:
Please check my comments.

spark/dl/src/test/scala/com/intel/analytics/bigdl/keras/KerasBaseSpec.scala
zhichao-li: oh, that's too bad which means we need to explicitly call this `ifskipTest` in each test.

zhichao-li: would change it if there's no other better option.

yiheng: verify it when run keras code?

wzhongyuan:
Looks good, tested on top of it locally, will add my test once it's merged :)

PR #1902, opened 2017-11-20
Keras support - add cosine proximity criterion

wzhongyuan:

add cosine proximity criterion

## How was this patch tested?

Local test, merged test with zhichao's unit test utility

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/CosineProximityCriterion.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/CosineProximityCriterion.scala
@@ -28,6 +28,8 @@ import scala.reflect.ClassTag
  * x'(i) = x(i) / sqrt(max(sum(x(i)^2), 1e-12))
  * y'(i) = y(i) / sqrt(max(sum(x(i)^2), 1e-12))
  * cosine_proximity(x, y) = mean(-1 * x'(i) * y'(i))
+ *
+ * Both batch and un-batched inputs are supported
  */
 class CosineProximityCriterion[@specialized(Float, Double) T: ClassTag]
 (implicit ev: TensorNumeric[T]) extends TensorCriterion[T] {
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/CosineProximityCriterion.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/CosineProximityCriterion.scala
@@ -27,7 +27,7 @@ import scala.reflect.ClassTag
  * The cosine proximity is defined as below:
  * x'(i) = x(i) / sqrt(max(sum(x(i)^2), 1e-12))
  * y'(i) = y(i) / sqrt(max(sum(x(i)^2), 1e-12))
- * cosine_proximity(x, y) = sum_i(-1 * x'(i) * y'(i))
+ * cosine_proximity(x, y) = mean(-1 * x'(i) * y'(i))
  */
 class CosineProximityCriterion[@specialized(Float, Double) T: ClassTag]
 (implicit ev: TensorNumeric[T]) extends TensorCriterion[T] {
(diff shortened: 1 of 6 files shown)


wzhongyuan:
@PR passed after refinement based on 1.2
http://172.168.2.101:8080/job/BigDL-PR-V/177/

spark/dl/src/test/scala/com/intel/analytics/bigdl/utils/serializer/ModuleSerializerSpec.scala
```diff
@@ -2058,6 +2058,20 @@ class ModuleSerializerSpec extends FlatSpec with Matchers {
     loadedModel.isTraining() should be (false)
   }
 
+  "HardSigmoid serialization" should "work properly" in {
+    val hardSigmoid = HardSigmoid()
```
zhichao-li: Wondering should we expose this as a generic test and put it into `beforeEach` for every `Layer` test? `BigDLSpecHelper` or `BigDLLayerSpec` would be a good place for this.

zhichao-li: @yiheng any suggestion? I think torchspec would like this as well.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/CosineProximityCriterion.scala
```diff
@@ -0,0 +1,137 @@
...
+      override def apply(v1: Array[T], v2: Int): Unit = {
+        loss = ev.plus(ev.negative(v1(v2)), loss)
+      }
+    }
+    DenseTensorApply.apply1(mul, func)
+    loss = ev.divide(loss, ev.fromType(mul.nElement()))
+    loss
+  }
+
+  private def l2Norm(input : Tensor[T]) : Tensor[T] = {
```
zhichao-li: Looks like a good utility method. not sure if we can move it to Tensor class. or maybe we can put it into a following PR.

zhichao-li:
Should we add `(val sizeAverage: Boolean = true)` as well? as all of the existing Criterion do support this option. Another thing is Keras1.2.2 only support `sizeAverage=true` and Keras2 support `sizeAverage=false`.  Maybe we can just stick with Keras1.2.2 in this PR.

zhichao-li:
just wondering how do we differentiate batch input and unbatch input? do we need to add some doc saying we support batch input only?

spark/dl/src/test/scala/com/intel/analytics/bigdl/utils/serializer/ModuleSerializerSpec.scala
```diff
@@ -2058,6 +2058,20 @@ class ModuleSerializerSpec extends FlatSpec with Matchers {
...
+    ModulePersister.saveToFile("/tmp/hardSigmoid.bigdl", hardSigmoid, true)
+    val loadedModel = ModuleLoader.loadFromFile("/tmp/hardSigmoid.bigdl")
+
+    val input = Tensor(2, 2).rand()
+
+    val res1 = hardSigmoid.forward(input)
+
+    val res2 = hardSigmoid.forward(input)
```
zhichao-li: should be `loadedModule` here?

wzhongyuan:
@zhichao-li   For serialization unit test, I do have some thoughts to refine it so that we can easily check if all layers have been covered,

wzhongyuan:
@zhichao-li   both batch and un-batched input are supported, Already tested it.

wzhongyuan:
@zhichao-li  Any other concerns ?

zhichao-li:
add some docs/comments for: "both batch and un-batched input are supported"?
LGTM except for this

wzhongyuan:
@zhichao-li  added comment and also run another PR successfully 
http://172.168.2.101:8080/job/BigDL-PR-V/238/

PR #1907, opened 2017-11-21
Support different types of tensor in the same sample

yangw1234:

Support different types of tensor in the same sample

## How was this patch tested?

Added unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
@@ -414,9 +414,9 @@ object MiniBatch {
   }
 
   private[bigdl] def copy[T: ClassTag](
-       samples: Seq[TensorSample[T]],
-       miniBatch: TableMiniBatch[T],
-       unlabeled: Boolean)(implicit ev: TensorNumeric[T]): MiniBatch[T] = {
+                                        samples: Seq[TableSample[T]],
+                                        miniBatch: TableMiniBatch[T],
+                                        unlabeled: Boolean)(implicit ev: TensorNumeric[T]): MiniBatch[T] = {
     val inputs = miniBatch.inputData
     val targets = miniBatch.targetData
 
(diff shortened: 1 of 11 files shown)


qiuxin2012:
Can you merge TensorSample and TableSample into one? I think they can be the same type of sample.

yangw1234:
pr validation pass

yangw1234:
@qiuxin2012 ready to be reviewed

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
```diff
@@ -396,34 +397,71 @@ object Sample {
...
 }
 
 /**
- * A kind of Sample who hold both DenseTensor and SparseTensor as features.
+ * A kind of Sample who holds different types of tensors as features.
  * @param features feature tensors
  * @param labels label tensors
  * @tparam T numeric type
  */
 private[bigdl] class TensorSample[T: ClassTag](
```
qiuxin2012: should we change the name to TableSample?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
```diff
@@ -396,34 +397,71 @@ object Sample {
       ArraySample(featureTensors)
     }
   }
+
+  def apply[T: ClassTag](
```
yiheng: The API looks not right. What's the meaning of T here?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
```diff
@@ -396,34 +397,71 @@ object Sample {
...
+      featureTensors: Table)(implicit ev: TensorNumeric[T]) : Sample[T] = {
+    new TensorSample(featureTensors, T())
+  }
+
+  def apply[T: ClassTag](
+      featureTensors: Table,
+      labelTensors: Table)(implicit ev: TensorNumeric[T]) : Sample[T] = {
+    new TensorSample[T](featureTensors, labelTensors)
```
yiheng: save question with the above

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/MiniBatch.scala
```diff
@@ -256,6 +256,10 @@ object MiniBatch {
     new ArrayTensorMiniBatch[T](input, target)
   }
 
+  def apply[T: ClassTag](input: Table, target: Table): MiniBatch[T] = {
```
yiheng: what's the meaning of T here?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Sample.scala
yangw1234: The base class Sample needs a type parameter T.

yiheng: I think we should change that. Or it's not a reasonable API.

Issue #1901, opened 2017-11-17
Support other activation in recurrent layers.


hkvision:
Currently in recurrent layers including SimpleRNN, LSTM, GRU and ConvLSTM, we hard code the activation to be `Tanh`, and inner_activation to be `Sigmoid`.
In Keras, users can pass in activation and inner_activation as parameters. Moreover, inner_activation is set to `hard_sigmoid` by default. Usually it turns out that when users add a recurrent layer, they won't specify the activation, namely they will use `hard_sigmoid` as inner_activation but we don't support it.
Maybe we can consider adding activation parameters in these layers.

zhichao-li:
cc @jason-dai @psyyz10 @yiheng  any thoughts?

PR #1909, opened 2017-11-21
add activation argument for recurrent layers

hkvision:

Users can pass in activation and inner_activation for recurrent layers like what keras does.

## How was this patch tested?

Jenkins

#1901

### DIFF
--- a/docs/docs/APIGuide/Layers/Recurrent-Layers.md
+++ b/docs/docs/APIGuide/Layers/Recurrent-Layers.md
@@ -254,11 +254,27 @@ output = module.forward(input)
 
 **Scala:**
 ```scala
-val rnnCell = RnnCell[Double](inputSize, hiddenSize, activation, wRegularizer, uRegularizer, bRegularizer)
+val rnnCell = RnnCell[Double](
+  inputSize,
+  hiddenSize,
+  activation,
+  isInputWithBias = true,
+  isHiddenWithBias = true,
+  wRegularizer = null,
+  uRegularizer = null,
+  bRegularizer = null)
 ```
 **Python:**
 ```python
-rnnCell = RnnCell(input_size, hidden_size, Tanh(), w_regularizer, u_regularizer, b_regularizer)
+rnnCell = RnnCell(
+  input_size,
+  hidden_size,
+  activation,
+  isInputWithBias=True,
+  isHiddenWithBias=True,
+  wRegularizer=None,
+  uRegularizer=None,
+  bRegularizer=None)
 ```
 
 Implementation of vanilla recurrent neural network cell
@@ -279,7 +295,10 @@ Parameters:
 
 * `inputSize` input size. Default: 4
 * `hiddenSize`  hidden layer size. Default: 3
-* `activation` activation function f for non-linearity
+* `activation` instance of activation function for non-linearity.
+  For Python, one can also pass the name of an existing activation as a string, eg. 'tanh', 'relu', 'sigmoid', 'hard_sigmoid', 'softmax' etc.
+* `isInputWithBias` boolean, whether to contain bias for input. Default: true
+* `isHiddenWithBias` boolean, whether to contain bias for hidden layer. Default: true
 * `wRegularizer` instance of `Regularizer`(eg. L1 or L2 regularization), applied to the input weights matrices. Default: null
 * `uRegularizer` instance of `Regularizer`(eg. L1 or L2 regularization), applied to the recurrent weights matrices. Default: null
 * `bRegularizer` instance of `Regularizer`(eg. L1 or L2 regularization), applied to the bias. Default: null
@@ -365,11 +384,27 @@ gradient = model.backward(input, grad_output)
 
 **Scala:**
 ```scala
-val lstm = LSTM(inputSize, hiddenSize)
+val lstm = LSTM(
+  inputSize,
+  hiddenSize,
+  p = 0.0,
+  activation = null,
+  innerActivation = null,
+  wRegularizer = null,
+  uRegularizer = null,
+  bRegularizer = null)
 ```
 **Python:**
 ```python
-lstm = LSTM(input_size, hidden_size)
+lstm = LSTM(
+  input_size,
+  hidden_size,
+  p=0.0,
+  activation=None,
+  inner_activation=None,
+  wRegularizer=None,
+  uRegularizer=None,
+  bRegularizer=None)
 ```
 
 Long Short Term Memory architecture.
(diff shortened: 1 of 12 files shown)


pyspark/bigdl/keras/converter.py
```diff
@@ -754,12 +752,8 @@ def create_convlstm2d(self, klayer, kclayer):
         self.check_constraint_in_config(config)
         activation = self.to_bigdl_activation(config["activation"],
                                               "%s_%s" % (config["name"], config["activation"]))
-        if not isinstance(activation, BLayer.Tanh):
```
zhichao-li: cc @psyyz10 any comments?

pyspark/bigdl/nn/layer.py
```diff
@@ -3232,21 +3259,6 @@ def __init__(self,
                                           index)
 
 
-class Sigmoid(Layer):
```
zhichao-li: why delete this?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ConvLSTMPeephole.scala
```diff
@@ -110,8 +116,12 @@ class ConvLSTMPeephole[T : ClassTag](
           .add(h2g))
     }
 
+    // make a new instance of innerActivation for each build gate
+    val innerActivation2 = innerActivation.cloneModule()
+    innerActivation2.setName(Integer.toHexString(java.util.UUID.randomUUID().hashCode()))
```
zhichao-li: This may impact the performance and why we need to re-create this every time?

hkvision: Actually in the original code, it uses .add(Sigmoid()), it is creating a new instance every time. Here I just clone the input activation to make a new instance.
I think that's basically the same?

pyspark/bigdl/nn/layer.py
hkvision: Not delete. Just remove it to the place before LSTM. 
Since in LSTM ut after adding activation paramater, I need Sigmoid() to create it. But if Sigmoid is defined after LSTM, then it will cause a notfound error.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ConvLSTMPeephole.scala
dding3: Yes, we need create a new sigmoid every time buildGate is called. I think we can try to move "if (innerActivation == null) innerActivation = Sigmoid[T]()" into buildGate, change innerActivation from a class member to a local variable, it will make sure the logic is the same and will not impact the performance.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ConvLSTMPeephole.scala
```diff
@@ -190,6 +200,9 @@ class ConvLSTMPeephole[T : ClassTag](
     buildCell()
     buildOutputGate()
 
+    val activation2 = activation.cloneModule()
+    activation2.setName(Integer.toHexString(java.util.UUID.randomUUID().hashCode()))
```
dding3: Similar to inneractivation.

spark/dl/src/main/scala/com/intel/analytics/bigdl/example/languagemodel/PTBModel.scala
```diff
@@ -49,7 +49,7 @@ object PTBModel {
                input: ModuleNode[Float]): ModuleNode[Float] = {
     if (depth == numLayers) {
       Recurrent[Float]()
-        .add(LSTM[Float](inputSize, hiddenSize, 0, null, null, null))
+        .add(LSTM[Float](inputSize, hiddenSize, 0, null, null, null, null, null))
```
dding3: I think two new added parameters have default value, we don't need change this file?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ConvLSTMPeephole.scala
```diff
@@ -54,6 +56,8 @@ class ConvLSTMPeephole[T : ClassTag](
   val kernelC: Int,
   val stride: Int = 1,
   val padding: Int = -1,
+  var activation: TensorModule[T] = null,
```
dding3: Looks like Keras defines activation as String and create an instance when using it. Maybe we can have the same behavior, one benefit is we don't need clone it every time.

docs/docs/APIGuide/Layers/Recurrent-Layers.md
```diff
@@ -383,6 +383,25 @@ Ref:
 3. http://arxiv.org/pdf/1503.04069v1.pdf
 4. https://github.com/wojzaremba/lstm
 
+
```
jason-dai: Specify which parameters are optional

pyspark/bigdl/nn/layer.py
```diff
@@ -1332,21 +1347,27 @@ class LSTM(Layer):
...
+    creating: createSigmoid
     creating: createL1Regularizer
     creating: createL1Regularizer
     creating: createL1Regularizer
     creating: createLSTM
     '''
 
-    def __init__(self, input_size, hidden_size, p=0.0, wRegularizer=None, uRegularizer=None, bRegularizer=None, bigdl_type="float"):
-        super(LSTM, self).__init__(None, bigdl_type, input_size, hidden_size, p, wRegularizer, uRegularizer, bRegularizer)
+    def __init__(self, input_size, hidden_size, p=0.0, activation=Tanh(), inner_activation=Sigmoid(),
```
jason-dai: For python API, just use string instead of class to specify the activations

hkvision: Supported name string as activation now.

spark/dl/src/main/scala/com/intel/analytics/bigdl/example/languagemodel/PTBModel.scala
hkvision: Yeah, no need to change this file. My bad :(

hkvision:
jenkins passed http://172.168.2.101:8080/view/PR-V-P/job/BigDL-PR-V/209/

hkvision:
@jason-dai @zhichao-li please have another quick review~

docs/docs/APIGuide/Layers/Recurrent-Layers.md
hkvision: DONE~

hkvision:
jenkins passed http://172.168.2.101:8080/view/PR-V-P/job/BigDL-PR-V/230/

PR #1917, opened 2017-11-22
feat: SReLU and ActivityRegularization support

i8run:

## How was this patch tested?

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SReLU.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SReLU.scala
@@ -162,12 +162,16 @@ class SReLU[T: ClassTag](shared_axes: Array[Int] = null)(
     while (batch < gradInput.size(1)) {
       val sliceInput = input.select(1, batch + 1)
       val sliceGradInput = gradInput.select(1, batch + 1)
+      val sliceGradOutput = gradOutput.select(1, batch + 1)
 
       val xArr = sliceInput.storage().array()
       var xOffset = sliceInput.storageOffset() - 1
 
       val yArr = sliceGradInput.storage().array()
       var yOffset = sliceGradInput.storageOffset() - 1
+
+      val zArr = sliceGradOutput.storage().array()
+      var zOffset = sliceGradOutput.storageOffset() - 1
 
       // if share axes array is not null, do some groups
       val groups = sliceInput.nElement() / weights(tLeft).nElement()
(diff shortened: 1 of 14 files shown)


yangw1234:
LGTM

zhichao-li:
@wzhongyuan  How can we make sure the new added layer match the Model serialization logic?

zhichao-li:
@i8run `ActivityRegularization` should be exposed as a Layer, then we can add it like the other Layer into the Model instead of just use it as Layer parameter.
i.e
```
    layer = ActivityRegularization(l1=0.01, l2=0.01)

    # test in functional API
    x = Input(shape=(3,))
    z = Dense(2)(x)
    y = layer(z)
    model = Model(input=x, output=y)
```

spark/dl/src/test/scala/com/intel/analytics/bigdl/nn/SReLUSpec.scala
```diff
@@ -0,0 +1,70 @@
...
+    srelu.forward(input)
+    println(srelu.output)
+  }
+
+  "SReLU serialize" should "work correctly" in {
+    val srelu = SReLU[Float]()
+    val input = Tensor[Float](5, 2, 3, 4).randn()
+    val res1 = srelu.forward(input)
+
+    ModulePersister.saveToFile[Float]("/tmp/srelu.bigdl", srelu, true)
```
wzhongyuan: Can we move it into 'ModuleSerialierSpec' with an separate unit test ?

zhichao-li: What I'm thinking is if we can make it be a common method, then the new added layer can be tested automatically for serialization which would make life more easier.
It's fine if move it to other spec.

wzhongyuan: @zhichao-li  That's the next plan, I will take care of it

wzhongyuan:
LGTM

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SReLU.scala
```diff
@@ -0,0 +1,370 @@
...
+  override def updateGradInput(input: Tensor[T], gradOutput: Tensor[T]): Tensor[T] = {
+    gradInput.resizeAs(input)
+
+    var batch = 0
+    while (batch < gradInput.size(1)) {
+      val sliceInput = input.select(1, batch + 1)
+      val sliceGradInput = gradInput.select(1, batch + 1)
+      val sliceGradOutput = gradOutput.select(1, batch + 1)
+
+      val xArr = sliceInput.storage().array()
```
yangw1234: Is it possible that the input tensor is not contiguous?

yangw1234:
LGTM

PR #1919, opened 2017-11-23
add model tag

wzhongyuan:

Add model tag in serialized message

## How was this patch tested?

Unit test

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Module.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Module.scala
@@ -24,7 +24,7 @@ import com.intel.analytics.bigdl.tensor.Tensor
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
 import com.intel.analytics.bigdl.utils.File
 import com.intel.analytics.bigdl.utils.caffe.CaffeLoader
-import com.intel.analytics.bigdl.utils.serializer.ModuleLoader
+import com.intel.analytics.bigdl.utils.serializer.{ModuleLoader, ModuleTag, ModuleWithTag}
 import com.intel.analytics.bigdl.utils.tf.{Session, TensorflowDataFormat, TensorflowLoader}
 
 import scala.reflect.ClassTag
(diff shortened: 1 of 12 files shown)


wzhongyuan:
@jason-dai  @SeaOfOcean   Please help review the change

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -473,6 +473,22 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
...
+  /**
+   * Save this module with class tag to path with protobuf format
+   * @param path path to save module, local file system, HDFS and Amazon S3 is supported.
+   *             HDFS path should be like "hdfs://[host]:[port]/xxx"
+   *             Amazon S3 path should be like "s3a://bucket/xxx"
+   * @param moduleTag  module tag
+   * @param overWrite if overwrite
+   * @return self
+   */
+  def saveModuleWithTag(path : String, moduleTag : ModuleTag, overWrite
```
jason-dai: Why do we need a new API? Just make moduleTag an optional argument in saveModule

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Module.scala
```diff
@@ -58,6 +58,20 @@ object Module {
...
+  /**
+   * Load model from path.
+   *
+   * @param path path to save module, local file system, HDFS and Amazon S3 is supported.
+   *             HDFS path should be like "hdfs://[host]:[port]/xxx"
+   *             Amazon S3 path should be like "s3a://bucket/xxx"
+   * @tparam T numeric type
+   * @return model loaded from path
+   */
+  def loadModuleWithTag[T: ClassTag](path : String)(implicit ev: TensorNumeric[T])
```
jason-dai: Don't add a new API; just add an optional argument for storing tag to loadModule.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/ModuleSerializer.scala
```diff
@@ -211,3 +211,9 @@ private case class GenericTypeWrapper[T: ClassTag](tensor : Tensor[T],
   ttpe : T
   )(implicit ev: TensorNumeric[T])
 
+// ModelTag is an unique identifier for a serialized model
+case class ModuleTag(publisher : String,
```
jason-dai: Do we need a ModuleTag class? Or maybe just use a string which can contain arbitrary tag, e.g., bigdl.v_0_3_0.ssd?

wzhongyuan: I discussed with @SeaOfOcean ,  Maybe with a ModuleTag we could handle the compatibility in a more elegant way in application (model_zoo or SSD model)

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Module.scala
wzhongyuan: @jason-dai   The tag is not a class field of 'AbstractModule', if we use the same API,  we may want to add that in AbstractModule, shall we ?

wzhongyuan:
As discussed offline, no need to add a separate tag, closing this one

PR #1925, opened 2017-11-23
Support same padding for keras atrousconv1d,2d and conv3d

hkvision:

Support same padding for keras atrousconv1d,2d and conv3d

## How was this patch tested?

Jenkins

### DIFF
--- a/pyspark/bigdl/keras/converter.py
+++ b/pyspark/bigdl/keras/converter.py
@@ -27,6 +27,7 @@ from keras.models import model_from_json
 from keras.models import Sequential, Model, Layer
 import keras
 import warnings
+from math import ceil
 
 
 def unsupport_exp(name):
@@ -255,7 +256,7 @@ class WeightsConverter:
         k_weights = weights[0]
         b_weights = k_weights[0].T
         for i in range(1, k_weights.shape[0]):
-            b_weights = np.concatenate((b_weights, k_weights[i].T))        
+            b_weights = np.concatenate((b_weights, k_weights[i].T))
         return [b_weights, weights[1]]
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


pyspark/bigdl/keras/converter.py
```diff
@@ -1030,21 +1025,30 @@ def create_atrousconvolution2d(self, klayer, kclayer):
...
+        w = input_shape[3]
+        kh = config["nb_row"]
+        kw = config["nb_col"]
+        dh = config["subsample"][0]
+        dw = config["subsample"][1]
+        dilation_h = config["atrous_rate"][0]
+        dilation_w = config["atrous_rate"][1]
+        pad_h = 0
+        pad_w = 0
+        if config["border_mode"] == "same":
```
zhichao-li: extract this as a function?

pyspark/bigdl/keras/converter.py
```diff
@@ -886,16 +888,6 @@ def to_bigdl_2d_padding(self, border_mode):
         else:
             raise Exception("Unsupported border mode: %s" % border_mode)
 
-    def to_bigdl_1d_padding(self, border_mode, kernel_w):
-        if border_mode == "same":
```
zhichao-li: why delete this?

pyspark/bigdl/keras/converter.py
hkvision: I consider extracting it as a function, but this function will need too many arguments but only does one calculation?

pyspark/bigdl/keras/converter.py
hkvision: This is never used lol. For all 1D layers, we use 2D layers to achieve. Thus, for padding, the manner is the same as 2D.

hkvision:
jenkins passed http://172.168.2.101:8080/view/PR-V-P/job/BigDL-PR-V/229/

hkvision:
Have a look on the calculation of padding? @yangw1234 Any comments or corner cases missing? Thanks

yangw1234:
LGTM

hkvision:
@yangw1234 thanks for the review!

hkvision:
jenkins passed http://172.168.2.101:8080/job/BigDL-PR-V-Spark-1.6/473/

PR #1939, opened 2017-11-27
Fix tensor set

SeaOfOcean:

Fix tensor set

## How was this patch tested?

unit test


SeaOfOcean:
bircnn tutorial now can work

SeaOfOcean:
pr validate pass

SeaOfOcean:
empty tensor size() will also throw null exception

PR #1941, opened 2017-11-27
Add include_container option for flattened_layers

zhichao-li:

Add include_container option for flattened_layers

## How was this patch tested?
unittest


zhichao-li:
test passed.

PR #1945, opened 2017-11-27
A keras backend wrapper

zhichao-li:
The idea is to provide some syntax sugar to  load/train/inference a keras model with bigdl 

## How was this patch tested?
unittest

### DIFF
--- a/pyspark/bigdl/keras/backend.py
+++ b/pyspark/bigdl/keras/backend.py
@@ -14,30 +14,12 @@
 # limitations under the License.
 #
 
-# import new
-# import types # TODO: support python3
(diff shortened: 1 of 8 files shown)


hkvision:
LGTM.
I will add more loss mapping (including loss function as parameter) afterwards.

zhichao-li:
@jason-dai  Shall we merge this once the Jenkins pass?  then we can enrich the loss function or the other testing base on that. 
The extra line still needed and haven't found a solution to get rid of it.  If take a step back, we also need some logic to hack the original `Model`.

jason-dai:
LGTM

PR #1951, opened 2017-11-29
Always averaged the returning value

zhichao-li:

Always averaged the returning value while comparing the result between bigdl and keras

## How was this patch tested?
unittest


hhbyyh:
LGTM pending on unit tests

zhichao-li:
jenkins passed

PR #1954, opened 2017-11-29
Add OpenCV Vision Transformer

SeaOfOcean:

Add OpenCV Vision Transformer structure
Will add the remaining transformers once the jenkins test pass

## How was this patch tested?

unit test


spark/dl/src/main/scala/com/intel/analytics/bigdl/transform/vision/image/opencv/OpenCVMat.scala
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  /**
+   * convert opencv mat to image bytes
+   *
+   * @param mat opencv mat
+   * @param encoding encoding type
+   * @return bytes that represent an image
+   */
+  def imencode(mat: Mat, encoding: String = "png"): Array[Byte] = {
```
jason-dai: OpenCVMat instead of Mat?

jason-dai:
Need to add more scala doc and python doc

And also add more image augmentation transformers (crop, brightness, colorjitter, etc.)

SeaOfOcean:
pr validate pass

PR #1956, opened 2017-11-30
add tensorflow operation SegmentSum

qiuxin2012:

add tensorflow operation SegmentSum

## How was this patch tested?
unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/SegmentSum.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/SegmentSum.scala
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2016 The BigDL Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/SegmentSum.scala
```diff
@@ -0,0 +1,55 @@
...
+import com.intel.analytics.bigdl.tensor.Tensor
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+import com.intel.analytics.bigdl.utils.Table
+
+import scala.reflect.ClassTag
+
+/**
+ * Computes the sum along segments of a tensor.
+ */
+class SegmentSum[T: ClassTag]()(implicit ev: TensorNumeric[T])
```
yiheng: Can input tensor has different numeric type of the parameter?

qiuxin2012: No, they are the same type.

yiheng:
looks good to me. Please merge it after pass the pr validation test

qiuxin2012:
jenkins passed

Issue #1937, opened 2017-11-27
Add new wrapper Table in python 


zhichao-li:
` def getHiddenStates(rec: Recurrent[T]): JList[JList[JTensor]] = { `
This is too complex; instead of doing this, maybe we can have automatic mapping of Table to python list? 

Not sure if we need to introduce a Table concept into Python. Maybe nested list is what we are looking for.

related pr: https://github.com/intel-analytics/BigDL/pull/1591#pullrequestreview-66305975
We can compare with the other framework and then think of this seriously.

PR #1962, opened 2017-11-30
An Activity wrapper for python to simplify the returning value

zhichao-li:

def getHiddenStates(rec: Recurrent[T]): JList[JList[JTensor]] = {
This is too complex; instead of doing this, maybe we can have automatic mapping of Table to python list?

## How was this patch tested?
unitest.

https://github.com/intel-analytics/BigDL/issues/1937

### DIFF
--- a/pyspark/bigdl/util/common.py
+++ b/pyspark/bigdl/util/common.py
@@ -118,6 +118,13 @@ def get_dtype(bigdl_type):
     # Always return float32 for now
     return "float32"
 
+
+class JActivity(object):
+
+    def __init__(self, value):
+        self.value = value
+
+
 class JTensor(object):
     """
     A wrapper to easy our work when need to pass or return Tensor to/from Scala.
@@ -362,7 +369,8 @@ _picklable_classes = [
     'LabeledPoint',
     'Sample',
     'EvaluatedResult',
-    'JTensor'
+    'JTensor',
+    'JActivity'
 ]
 
 
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/BigDLSerde.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/BigDLSerde.scala
@@ -23,7 +23,10 @@ import java.nio.charset.StandardCharsets
 import java.nio.{ByteBuffer, ByteOrder}
 import java.util.{ArrayList => JArrayList, HashMap => JHashMap, List => JList, Map => JMap}
 
-import com.intel.analytics.bigdl.python.api.{EvaluatedResult, JTensor, Sample}
+import com.intel.analytics.bigdl.nn.abstractnn.Activity
+import com.intel.analytics.bigdl.python.api._
+import com.intel.analytics.bigdl.tensor.Tensor
+import com.intel.analytics.bigdl.utils.Table
 import net.razorvine.pickle._
 import org.apache.spark.api.java.JavaRDD
 import org.apache.spark.api.python.SerDeUtil
@@ -223,6 +226,45 @@ object BigDLSerDe extends BigDLSerDeBase with Serializable {
     }
   }
 
+  private[python] class JActivityPickler extends BigDLBasePickler[JActivity] {
+    private def doConvertTable(table: Table): Any = {
+      val tMap = table.getState()
+      if (tMap.isEmpty) {
+        throw new RuntimeException("Found empty table")
+      }
+      if (tMap.values.iterator.next().isInstanceOf[Activity]) {
+        return tMap.values.map { value =>
+          doConvertActivity(value.asInstanceOf[Activity])
+        }.toList.asJava
+      } else {
+        throw new RuntimeException(s"""not supported type:
+          ${tMap.values.iterator.next().getClass.getSimpleName}""")
+      }
+    }
+    private def doConvertActivity(activity: Activity): Any = {
+      if (activity.isTable) {
+        return doConvertTable(activity.toTable)
+      } else if (activity.isTensor) {
+        return PythonBigDL.ofFloat().toJTensor(activity.toTensor)
+      } else {
+        throw new RuntimeException(s"""not supported type:
+          ${activity.getClass.getSimpleName}""")
+      }
+    }
+
+    def saveState(obj: Object, out: OutputStream, pickler: Pickler): Unit = {
+      val record = obj.asInstanceOf[JActivity]
+      saveObjects(out,
+        pickler,
+        doConvertActivity(record.value))
+    }
+
+    def construct(args: Array[Object]): Object = {
+      throw new RuntimeException("haven't be implemented")
+    }
+  }
+
+
   private[python] class TestResultPickler extends BigDLBasePickler[EvaluatedResult] {
 
     def saveState(obj: Object, out: OutputStream, pickler: Pickler): Unit = {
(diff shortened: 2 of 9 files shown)


zhichao-li:
Would split the wrapper testing code into an separated file.

zhichao-li:
cc @dding3 I think you only need to returning `JTable` in your case after this PR merged.

spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/BigDLSerde.scala
```diff
@@ -223,6 +226,60 @@ object BigDLSerDe extends BigDLSerDeBase with Serializable {
...
+
+    def construct(args: Array[Object]): Object = {
+      throw new RuntimeException("haven't be implemented")
+    }
+  }
+
+  private[python] class JActivityPickler extends BigDLBasePickler[JActivity] {
+
+    private def doConvert(jactivity: JActivity): Any = {
+      if (jactivity.value.isInstanceOf[JTable]) {
```
jason-dai: use isTable() instead

spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/BigDLSerde.scala
```diff
@@ -223,6 +226,60 @@ object BigDLSerDe extends BigDLSerDeBase with Serializable {
...
+        return jactivity.value.asInstanceOf[JTable]
+      } else if (jactivity.value.isInstanceOf[JTensor]) {
```
jason-dai: use isTensor() instead

spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDL.scala
```diff
@@ -65,6 +65,12 @@ case class Sample(features: JList[JTensor],
 case class JTensor(storage: Array[Float], shape: Array[Int],
                    bigdlType: String, indices: Array[Array[Int]] = null)
 
+// `Any` of the value can only be a JTensor or JTable
+case class JTable(value: JList[Any])
+
+// `Any` of the value can only be a JTensor or JTable
+case class JActivity(value: Any)
```
jason-dai: Why don't we use JList and JTensor?

zhichao-li: You mean the value of JActivity should be `JList` and `JTensor` ? in that case the semantic would not be the same as the original `Activity`. Anyway, what the python can see is just nested `List` or `JTensor`

jason-dai:
Do we need JTable? I think we can just add JActivity - which can be then mapped to either JTensor or JList.

zhichao-li:
JTable has been removed.

zhichao-li:
Jenkins passed. Pending on approval.

spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/BigDLSerde.scala
```diff
@@ -223,6 +226,45 @@ object BigDLSerDe extends BigDLSerDeBase with Serializable {
...
   }
 
+  private[python] class JActivityPickler extends BigDLBasePickler[JActivity] {
+    private def doConvertTable(table: Table): Any = {
+      val tMap = table.getState()
+      if (tMap.isEmpty) {
+        throw new RuntimeException("Found empty table")
+      }
+      if (tMap.values.iterator.next().isInstanceOf[Activity]) {
+        return tMap.values.map { value =>
```
jason-dai: For Activity, we need to keep the order of the elements in Table; I don't think it is guaranteed here.

pyspark/bigdl/util/common.py
```diff
@@ -118,6 +118,13 @@ def get_dtype(bigdl_type):
     # Always return float32 for now
     return "float32"
 
+
+class JActivity(object):
```
jason-dai: I don't think we need JActivity on the Python size?

zhichao-li: We do need that as kind of container otherwise there's no place to put those converting method.

spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/BigDLSerde.scala
zhichao-li: Shall we have an assumption that the key is always be `Int`? I can sort those values in that case, or we might need to return a nested Map instead of nested list.

jason-dai: Yes - for Activity, we always set/get the value of a Table t as t(1), t(2), etc. See https://github.com/intel-analytics/BigDL/blob/master/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Cell.scala#L109 for an example

zhichao-li: ok, let me update the logic accordingly.

spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/BigDLSerde.scala
```diff
@@ -223,6 +225,46 @@ object BigDLSerDe extends BigDLSerDeBase with Serializable {
     }
   }
 
+  private[python] class JActivityPickler extends BigDLBasePickler[JActivity] {
+    private def doConvertTable(table: Table): Any = {
+      val sortedList = table.getState().toList.sortWith(
```
jason-dai: Instead of sorting, I think you should do something like `table.toSeq().toList`

zhichao-li: Within `toSeq` it just retrieve the key value starting with 1 same as sorting, but it also add restriction that the value should only be `Tensor[T]` instead of `Activity`. 
Should we update that method `toSeq` to return `Seq[Activity]`? if so, I think we also need to refactor some code to adapt this changes.
```
def toSeq[T]: Seq[Tensor[T]] = {
```

zhichao-li: let me update `toSeq` method to return Seq[Activity]

jason-dai: You can either update `toSeq`, or retrieve each element using index 1 to N. Either way, we should avoid sorting.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Table.scala
```diff
@@ -295,7 +295,26 @@ class Table private[bigdl](
...
+    }
+  }
+
+  /**
+   * Return the elements of this table as a Seq.
+   * This method assumes the key of this table are all
+   * the integers between 1 to this.length(),
+   * the values are all Activity
+   */
+  def toSeqActivity[T]: Seq[Activity] = {
```
jason-dai: Why don't you have a toSeq[D], which can then support both toSeq[Tensor[T]] and toSeq[Activity] ?

zhichao-li: Good idea!

zhichao-li:
Jenkins passed.

jason-dai:
LGTM

PR #1966, opened 2017-12-01
Do not check spark-bigdl.conf if sc exists

zhichao-li:
Sometimes user create SparkContext with the require config settings, in that case we should not check and create "spark-bigdl.conf". 

## How was this patch tested?
unitest


jenniew:
LGTM

zhichao-li:
jenkins passed.

PR #1982, opened 2017-12-05
fix output allocation bug

### LINKED ISSUE #1980: Some layers with 2 TensorNumeric does not work properly
Some layers have multiple TensorNumeric, but it does not work correctly if different types are specified.

Below is one example
```java
    val abs = Abs[Float, Double]()
    val input = Tensor[Double](2).fill(-1.0)
    val res1 = abs.forward(input)
```
Exception will be thrown out
```java
java.lang.ClassCastException: [D cannot be cast to [F
	at com.intel.analytics.bigdl.tensor.TensorNumericMath$TensorNumeric$NumericFloat$.vAbs(TensorNumeric.scala:479)
	at com.intel.analytics.bigdl.tensor.DenseTensor.abs(DenseTensor.scala:1730)
	at com.intel.analytics.bigdl.nn.Abs.updateOutput(Abs.scala:34)
	at com.intel.analytics.bigdl.nn.Abs.updateOutput(Abs.scala:28)
	at com.intel.analytics.bigdl.nn.abstractnn.AbstractModule.forward(AbstractModule.scala:280)
```
Seems the first TensorNumeric is applied by default to output type.

yangw1234:

fix output allocation bug

fix #1980 

## How was this patch tested?

unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Abs.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Abs.scala
@@ -29,6 +29,10 @@ class Abs[T: ClassTag, D: ClassTag]
  (implicit ev: TensorNumeric[T], ev2: TensorNumeric[D])
   extends AbstractModule[Tensor[D], Tensor[D], T] {
 
+  output = Tensor[D]()
+
+  gradInput = Tensor[D]()
+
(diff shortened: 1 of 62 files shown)


yangw1234:
pr validation pass

wzhongyuan:
LGTM

PR #1983, opened 2017-12-05
code refactor to reduce code duplicate and avoid manual toBatch operation.

hhbyyh:

1. Support more data types for features and label columns.
2. Move DLParam from SparkVersion to dl folder, to avoid code duplicate between Spark 1.6 and 2.0+
3. add transient to model parameter in DLEstimator and DLClassifier, to avoid unnecessary serialization.
4. use Optimizer and Predictor for batch training and avoid manual toBatch operation.

Other new functions will be added in follow-up PRs.

## How was this patch tested?

existing unit tests

### DIFF
--- a/spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
+++ b/spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
@@ -41,7 +41,7 @@ private[ml] trait HasBatchSize extends Params {
  * Common trait for DLEstimator and DLModel
  */
 private[ml] trait DLParams[@specialized(Float, Double) T] extends HasFeaturesCol
-  with HasPredictionCol with VectorCompatibility {
+  with HasPredictionCol with VectorCompatibility with HasBatchSize {
 
   /**
    * optimization method to be used. BigDL supports many optimization methods like Adam,
@@ -75,6 +75,8 @@ private[ml] trait DLParams[@specialized(Float, Double) T] extends HasFeaturesCol
   final val learningRateDecay = new DoubleParam(this, "learningRateDecay", "learningRateDecay")
 
   def getLearningRateDecay: Double = $(learningRateDecay)
+
+  setDefault(batchSize -> 1)
 
   protected def validateDataType(schema: StructType, colName: String): Unit = {
     val dataTypes = Seq(
@@ -116,7 +118,6 @@ private[ml] trait DLParams[@specialized(Float, Double) T] extends HasFeaturesCol
         }
     }
   }
-
 }
 
 
@@ -147,7 +148,7 @@ class DLEstimator[@specialized(Float, Double) T: ClassTag](
     val featureSize : Array[Int],
     val labelSize : Array[Int],
     override val uid: String = "DLEstimator")(implicit ev: TensorNumeric[T])
-  extends DLEstimatorBase[DLEstimator[T], DLModel[T]] with DLParams[T] with HasBatchSize {
+  extends DLEstimatorBase[DLEstimator[T], DLModel[T]] with DLParams[T] {
 
   def setFeaturesCol(featuresColName: String): this.type = set(featuresCol, featuresColName)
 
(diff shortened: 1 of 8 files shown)


hhbyyh:
cc @wzhongyuan @yiheng @jason-dai

spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
```diff
@@ -67,68 +166,61 @@ class DLEstimator[@specialized(Float, Double) T: ClassTag](
...
-   * learning rate for the optimizer in the DLEstimator.
-   * Default: 0.001
-   */
-  val learningRate = new DoubleParam(this, "learningRate", "learningRate", ParamValidators.gt(0))
-
-  setDefault(learningRate -> 1e-3)
-
-
-  def getLearningRate: Double = $(learningRate)
+  setDefault(maxEpoch -> 100)
```
jason-dai: 100 epoch can be too large?

spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
```diff
@@ -209,52 +276,40 @@ class DLModel[@specialized(Float, Double) T: ClassTag](
...
-    val resultRDD = dataset.rdd.zip(predictRdd).map { case (row, predict) =>
-      Row.fromSeq(row.toSeq ++ Seq(predict))
-    }
-    val resultSchema = transformSchema(dataset.schema)
-    dataset.sqlContext.createDataFrame(resultRDD, resultSchema)
+    val prediction = model.asInstanceOf[Module[T]]
+      .predict(samplesRDD, $(batchSize))
+      .map { t => outputToPrediction(t.asInstanceOf[Tensor[T]]) }
+
+    val resultRDD = dataFrame.rdd.zipWithIndex().map(_.swap)
```
jason-dai: A potential issue is that dataFrame.rdd may need to be computed twice if it's not cached. One way to address that is to do something like https://github.com/intel-analytics/BigDL/blob/master/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Predictor.scala#L93

spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
hhbyyh: Just want to ensure that we got converged model on some smaller data set. 
I can change it to 50.

hhbyyh:
Please let me know if further change is required. Thanks.

hhbyyh:
http://172.168.2.101:8080/job/BigDL-PR-V/568/

hhbyyh:
@jason-dai we need the change to support Mastercard, is it OK to merge this into 0.4?

spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
```diff
@@ -281,25 +283,31 @@ class DLModel[@specialized(Float, Double) T: ClassTag](
     val featureType = dataFrame.schema($(featuresCol)).dataType
     val featureColIndex = dataFrame.schema.fieldIndex($(featuresCol))
     val featureFunc = getConvertFunc(featureType)
+    val sc = dataFrame.sqlContext.sparkContext
+    val modelBroadCast = ModelBroadcast[T]().broadcast(sc, model)
+    val localBatchSize = $(batchSize)
+
+    val resultRDD = dataFrame.rdd.mapPartitions { rowIter =>
+      val rows = rowIter.toArray
```
jason-dai: What if rows do not fit in memory?

hhbyyh: Updated to use group iterator

jason-dai:
LGTM

hhbyyh:
http://172.168.2.101:8080/job/BigDL-PR-V/584/ passed.

PR #1985, opened 2017-12-06
Support passing JActivity from python to java

zhichao-li:

i.e:
```
        value = np.random.random_sample([2, 3])
        nested = JActivity([ [value, value],  [value, value]])
        back = callBigDlFunc("float", "testPyToJavaActivity",
                             nested)
```
## How was this patch tested?
unittest

### DIFF
--- a/pyspark/bigdl/util/common.py
+++ b/pyspark/bigdl/util/common.py
@@ -122,7 +122,34 @@ def get_dtype(bigdl_type):
 class JActivity(object):
 
     def __init__(self, value):
-        self.value = value
+        self.value = self.__to_np(value)  # self.value is a ndarray or nested list of ndarray
+
+    @staticmethod
+    def __to_np(v):
+        """
+        This method would convert the JTensor to ndarray
+        :return: ndarray or nested list of ndarray
+        """
+        if isinstance(v, JTensor):
+            return v.to_ndarray()
+        elif isinstance(v, list):
+            return [JActivity.__to_np(i) for i in v]
+        elif isinstance(v, np.ndarray):
+            return v
+        else:
+            raise Exception("Unsupported type: %s", type(v))
+
+    @staticmethod
+    def __do_convertion(v):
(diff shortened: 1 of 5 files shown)


zhichao-li:
@jason-dai  How about this style?
By default,  we only return `JActivity` as JavaObject without Serialization and we can use `toPy` method to enforce the serialization to convert it to ndarray or nested list of ndarray.

Java Side:
```
  def testPyToJavaActivity(jactivity: JActivity): JActivity = {
    jactivity
  }
```

Python Side:
```
        back = callBigDlFunc("float", "testActivityWithTableOfTable")  # back is a JActivity from Java which is a JavaObject without serialization
        assert isinstance(back, JavaObject)
        back_again = callBigDlFunc("float", "testPyToJavaActivity", back)  # back_again is also a JavaObject withou serialization
        assert isinstance(back_again, JavaObject)
        back_again = back_again.toPy()
        assert isinstance(back_again.value, list)
        assert isinstance(back_again.value[0], list)
        assert isinstance(back_again.value[0][0], np.ndarray)
```

zhichao-li:
jenkins passed

jason-dai:
Please check with @dding3 on how this will be used?

dding3:
I usually set state which returned from get_hidden_state directly in set_hidden_state. And I tried this PR with simplernn/lstm/multirnncell, it works good.

jason-dai:
Why do you need to do that in python @dding3 ? I think this is needed in seq2seq training, and that's done in scala side?

dding3:
Yes, it's done in scala side. 

However in scala Recurrent, we support get/setHiddenState and they are public APIs. Besides, both tf and torch support user set init state. I am thinking it would be better if we support these functionality in python as well.

jason-dai:
I don't think we need to add that at this moment if no one is expected use that.

zhichao-li:
BTW, the there's no overhead for get/set JActivity if without calling `toPy()` on this PR.

Issue #1948, opened 2017-11-28
make it more convinient to reuse optimizer


shane-huang:
sub issue seperated from issue #1872. 
We need to expose 2 sets of new interfaces:
1. setDataset
1. setCriterion
setDataset needs a sets of overloaded methods to support several arguments as RDD[Sample], batch size,etc. 
Also add python support.

PR #1986, opened 2017-12-06
add setTrainData and setCriterion for optimizer reuse

shane-huang:
add setTrainData and setCriterion for optimizer reuse, both scala and python api supported

Unit test and PR-validatoin passed.

### DIFF
--- a/docs/docs/APIGuide/Optimizers/Optimizer.md
+++ b/docs/docs/APIGuide/Optimizers/Optimizer.md
@@ -6,36 +6,33 @@ An optimizer is in general to minimize any function with respect to a set of par
 
 ***Factory method***
 
-```scala
-val optimizer = Opimizer[T: ClassTag](
-  model: Module[T],
-  sampleRDD: RDD[Sample[T]],
-  criterion: Criterion[T],
-  batchSize: Int)
-```
-`T`: the numeric type(Float/Double).  
-`model`: the model will be optimized.  
-`sampleRDD`: an RDD of training Sample.  
-`criterion`: the Loss function.  
-`batchSize`: size of minibatch. 
- <br>
+In summary, you need to supply 3 kinds of paramters to create an optimizer:
+1) train data: You could supply 
+   a) a sampleRDD and batchSize (with optional featurePadding and labelPadding), or 
+   b) a sampleRDD and batchSize and a customized implemenation of trait MiniBatch, or
+   c) a DataSet - the type of optimizer created will be dermined by the type of Dataset. 
+2) a model
+3) a criterion (i.e. loss)
+as shown in below interfaces:
 
 ```scala
-val optimizer = Opimizer[T: ClassTag](
-  model: Module[T],
-  sampleRDD: RDD[Sample[T]],
-  criterion: Criterion[T],
-  batchSize: Int,
-  miniBatch: MiniBatch[T]
+val optimizer = Opimizer[T: ClassTag, D](
+      model: Module[T],
+      sampleRDD: RDD[Sample[T]],
+      criterion: Criterion[T],
+      batchSize: Int,
+      featurePaddingParam: PaddingParam[T]=null,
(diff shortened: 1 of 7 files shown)


shane-huang:
fixed issue #1948

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Optimizer.scala
```diff
@@ -232,6 +232,88 @@ abstract class Optimizer[T: ClassTag, D](
...
   }
 
+  /**
+   * Set new train dataset
+   *
+   * @param sampleRDD training Samples
+   * @param batchSize mini batch size
+   * @return the Optimizer
+   */
+  def setTrainData(sampleRDD: RDD[Sample[T]],
```
jason-dai: Shall we combine these methods, using None as default for optional arguments?

docs/docs/APIGuide/Optimizers/Optimizer.md
```diff
@@ -136,6 +154,41 @@ Function setModel will set a new model to the optimizer.
...
+
+                 batchSize: Int,
+                 featurePaddingParam: PaddingParam[T],
+                 labelPaddingParam: PaddingParam[T])
+
+optimizer.setTrainData(sampleRDD: RDD[Sample[T]],
+                        batchSize: Int,
+                        featurePaddingParam: PaddingParam[T])
+```
+the overloaded set of methods `setTrainData` allows user to replace the trainig data before calling optimize(). The mearning of arguments are the same as in the Factory methods:
```
yiheng: mearning -> meaning

yiheng: can we replace the training data after calling optimize?

docs/docs/APIGuide/Optimizers/Optimizer.md
```diff
@@ -136,6 +154,41 @@ Function setModel will set a new model to the optimizer.
 `newModel`: a model will replace the old model in optimizer.  
  <br>
 ```scala
+optimizer.setTrainData(sampleRDD: RDD[Sample[T]],
+                 batchSize: Int)
+
+optimizer.setTrainData(sampleRDD: RDD[Sample[T]],
+                 batchSize: Int,
+                 miniBatch: MiniBatch[T])
```
yiheng: why sampleRDD and miniBatch are both here?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Optimizer.scala
```diff
@@ -350,9 +442,9 @@ object Optimizer {
       )(implicit ev: TensorNumeric[T]): Optimizer[T, MiniBatch[T]] = {
     new DistriOptimizer[T](
       _model = model,
-      dataset = (DataSet.rdd(sampleRDD) -> SampleToMiniBatch(batchSize))
+      _dataset = (DataSet.rdd(sampleRDD) -> SampleToMiniBatch(batchSize))
```
yiheng: where is the _dataset from?

docs/docs/APIGuide/Optimizers/Optimizer.md
shane-huang: why we need so ? it would be used like setTrainData, optimize(), setTrainData, optimize()

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Optimizer.scala
shane-huang: I follow the same interface as the apply() function of optimizer.  Probabaly it's designed this way to avoid conditional clauses.

shane-huang: changed both apply and setTrainData interface and merged several.

yiheng:
LGTM

PR #1997, opened 2017-12-08
Fix some issues found when save bigdl model to tensorflow format file

yiheng:
This PR fixes some issues when saving bigdl model to tensorflow format file
1. BatchNormalization without affine
2. SpatialConvolution without bias
2. SpatialConvolution with conv group
3. Support save Scale layer
4. Support save LRN layer
5. get padding information issue in pooling layer

## How was this patch tested?
unit test, manual test


yangw1234:
LGTM

PR #1999, opened 2017-12-08
Add Keras Init Methods

hkvision:

Add some simple init methods that supported by keras.
Just put it here if necessary.

## How was this patch tested?

Jenkins

### DIFF
--- a/pyspark/bigdl/nn/initialization_method.py
+++ b/pyspark/bigdl/nn/initialization_method.py
@@ -22,11 +22,13 @@ if sys.version >= '3':
     long = int
     unicode = str
 
+
 class InitializationMethod(JavaValue):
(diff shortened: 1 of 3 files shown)


zhichao-li:
The problem it's not easy to validate the random result.  It's OK to merge this if just want to enrich the init methods. cc @jason-dai  @yiheng  any thoughts?

hkvision:
Yes, it's not easy to verify random results. But according to the existing implementation of `Xavier`, these initialization methods are much alike and should be correct (just uniform or normal distributions)

hkvision:
If we believe `Xavier` in BigDL is equivalent to `glorot_uniform`  in Keras, then these would be correct as well.

PR #2001, opened 2017-12-09
Support gradient clipping

dding3:
1. support clip by value
2. support l2 norm clip

## How was this patch tested?
manual test

### DIFF
--- a/pyspark/bigdl/optim/optimizer.py
+++ b/pyspark/bigdl/optim/optimizer.py
@@ -690,6 +690,25 @@ class Optimizer(JavaValue):
         callBigDlFunc(self.bigdl_type, "setCheckPoint", self.value,
                       checkpoint_trigger, checkpoint_path, isOverWrite)
 
+    def set_constant_clipping(self, min_value, max_value):
+        """
+        Configure constant clipping settings.
+
+
+        :param min_value: the minimum value to clip by
+        :param max_value: the maxmimum value to clip by
+        """
+        callBigDlFunc(self.bigdl_type, "setConstantClip", self.value, min_value, max_value)
+
+    def set_l2norm_clipping(self, clip_norm):
+        """
+        Configure L2 norm clipping settings.
+    
+    
+        :param clip_norm: gradient L2-Norm threshold        
+        """
+        callBigDlFunc(self.bigdl_type, "setL2NormClip", self.value, clip_norm)
+
     # return a module
     def optimize(self):
         """
--- a/pyspark/bigdl/optim/optimizer.py
+++ b/pyspark/bigdl/optim/optimizer.py
@@ -709,6 +709,12 @@ class Optimizer(JavaValue):
         """
         callBigDlFunc(self.bigdl_type, "setL2NormClip", self.value, clip_norm)
 
+    def disable_gradclip(self):
+        """
+        disable clipping.
+        """
+        callBigDlFunc(self.bigdl_type, "disableClip", self.value)
+
     # return a module
     def optimize(self):
         """
--- a/pyspark/bigdl/optim/optimizer.py
+++ b/pyspark/bigdl/optim/optimizer.py
@@ -690,7 +690,7 @@ class Optimizer(JavaValue):
         callBigDlFunc(self.bigdl_type, "setCheckPoint", self.value,
                       checkpoint_trigger, checkpoint_path, isOverWrite)
 
-    def set_constant_clipping(self, min_value, max_value):
+    def set_gradclip_const(self, min_value, max_value):
         """
         Configure constant clipping settings.
 
@@ -700,7 +700,7 @@ class Optimizer(JavaValue):
         """
         callBigDlFunc(self.bigdl_type, "setConstantClip", self.value, min_value, max_value)
 
-    def set_l2norm_clipping(self, clip_norm):
+    def set_gradclip_l2norm(self, clip_norm):
         """
         Configure L2 norm clipping settings.
     
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
@@ -319,8 +319,7 @@ object DistriOptimizer {
               squares(tid) = ev.toType[Double](
                 parameters.gradientPartition.narrow(1, offset + 1, length).sumSquare())
             }))
-            var sum = 0.0
-            squares.foreach(sum += _)
+            var sum = 0.0            
             var i = 0
             while (i < parallelNum) {
               sum += squares(i)
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
@@ -18,7 +18,7 @@ package com.intel.analytics.bigdl.optim
 
 import com.intel.analytics.bigdl.{Module, _}
 import com.intel.analytics.bigdl.dataset.{DataSet, DistributedDataSet,
-MiniBatch, SampleToMiniBatch, Sample, PaddingParam}
+                        MiniBatch, SampleToMiniBatch, Sample, PaddingParam}
 import com.intel.analytics.bigdl.nn.{Module, Utils}
 import com.intel.analytics.bigdl.parameters.AllReduceParameter
 import com.intel.analytics.bigdl.tensor.Tensor
@@ -56,15 +56,15 @@ object DistriOptimizer {
    * @tparam T Tensor element type
    */
   case class Cache[T](
-     localModels: Array[Module[T]],
-     modelWeights: Array[Tensor[T]],
-     modelGradients: Array[Tensor[T]],
-     localCriterions: Array[Criterion[T]],
-     localStates: Array[Table],
-     var moduleTimeList: Array[Long] = null,
-     localMethods: Array[Option[Array[ValidationMethod[T]]]],
-     optimMethod: OptimMethod[T]
-   )
+    localModels: Array[Module[T]],
+    modelWeights: Array[Tensor[T]],
+    modelGradients: Array[Tensor[T]],
+    localCriterions: Array[Criterion[T]],
+    localStates: Array[Table],
+    var moduleTimeList: Array[Long] = null,
+    localMethods: Array[Option[Array[ValidationMethod[T]]]],
+    optimMethod: OptimMethod[T]
+  )
 
   /**
    * Train the model.
@@ -85,8 +85,10 @@ object DistriOptimizer {
    * @param trainSummary train summary
    * @param validationSummary validation summary
    * @param isOverWrite  if overwrite the checkpoint
+   * @param clippingParams  gradient clipping configurations
    */
   private[optim] def optimize[T: ClassTag](
+    trainingModel: Module[T],
     dataset: DistributedDataSet[MiniBatch[T]],
     coresPerNode: Int,
     state: Table,
@@ -445,7 +447,8 @@ object DistriOptimizer {
             summary,
             models,
             driverState,
-            parameters
+            parameters,
+            trainingModel
           )
         }
 
@@ -457,7 +460,8 @@ object DistriOptimizer {
           models,
           driverState,
           parameters,
-          optimMethod
+          optimMethod,
+          trainingModel
         )
 
       } else {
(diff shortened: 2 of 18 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/parameters/AllReduceParameter.scala
```diff
@@ -292,16 +292,17 @@ class AllReduceParameter[T: ClassTag](id: Long, partitionNum: Int, size: Int) ex
    */
   def sendWeightPartition(): Unit = {
     val blockId = getWeightBlockId(partitionId)
+    val localBuffer = BlockManagerWrapper.getLocalBytes(blockId).getOrElse {
+      throw new RuntimeException(s"Didn't find weight block $blockId in the block " +
+        s"manager. Did you initialize this AllReduceParameter on every executor?")
+    }
+    SerializerInstance.serialize(localBuffer).compress(weightPartition)
```
jason-dai: The `serialize(ByteBuffer)` method is confusing; maybe change to a different name such as create

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/LocalOptimizer.scala
```diff
@@ -152,8 +152,19 @@ class LocalOptimizer[T: ClassTag] (
           })
       )
       val loss = lossSum / parallelism
-      grad.div(ev.fromType(parallelism))
+      var scale = ev.fromType(parallelism)
+      if (gradientClippingParams.enableL2NormClipping) {
```
jason-dai: If the model (e.g., AlexNet or VGG) has large parameter sizes, the overhead can be large? Shall we do this using multi-threading?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -289,20 +297,43 @@ object DistriOptimizer {
         numFinishedModelUpdates >= driverSubModelNum * (1.0 - maxDropPercentage)) {
         // enough records were processed for this batch, so update the model
         val value = lossSum.value / numFinishedModelUpdates
+
+        var l2Norm = 0.0f
+        var scale = ev.fromType(numFinishedModelUpdates)
+        if (normClippingEnable) {
```
jason-dai: If the model (e.g., AlexNet or VGG) has large parameter sizes, the overhead can be large? Shall we do this using multi-threading?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Optimizer.scala
```diff
@@ -452,3 +481,10 @@ object Optimizer {
     }
   }
 }
+
+case class GradientClippingParams(
```
jason-dai: make this a trait, and implement two subclass for l2norm and constant clipping respectively

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
```diff
@@ -866,6 +866,34 @@ private[tensor] class DenseTensor[@specialized T: ClassTag](
...
   }
 
+  protected def sumNSquare(n: T): T = {
+    var res: T = ev.fromType(0)
+    val func = new TensorFunc2[T] {
+      override def apply(data1: Array[T], offset1: Int): Unit = {
+        res = ev.plus(res, ev.pow(ev.abs(data1(offset1)), n))
+      }
+    }
+    DenseTensorApply.apply1[T](this, func)
```
jason-dai: Instead of using apply1, how about using sth like vsMul?

jason-dai:
Please verify the performance overhead on models with large parameter sizes (such as AlexNet or LSTM)

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -148,6 +149,13 @@ object DistriOptimizer {
     var dropModelNumBatch = 0
     var lossArray = new Array[Double](_subModelNumber)
 
+    // gradient clip settings
+    val constantClippingEnable = clippingParams.enableConstantClipping
+    val normClippingEnable = clippingParams.enableL2NormClipping
```
jason-dai: what if both l2norm and constant clipping are set to true?

dding3: It will use l2norm first and then constant clipping

jason-dai: Does it make sense?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/LocalOptimizer.scala
dding3: Updated to use multi-thread to compute the square sum

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
dding3: Updated

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
dding3: Updated to use MKL

pyspark/bigdl/optim/optimizer.py
```diff
@@ -669,6 +669,25 @@ def set_checkpoint(self, checkpoint_trigger,
         callBigDlFunc(self.bigdl_type, "setCheckPoint", self.value,
                       checkpoint_trigger, checkpoint_path, isOverWrite)
 
+    def set_constant_clipping(self, min_value, max_value):
```
shane-huang: shall we change the name to gradient clipping? We can not know what it is clipping only by the method name. And add some docs?

shane-huang: if the name too long, u can use sth. like set_gradclip_const, and set_gradclip_l2norm

jason-dai: Sounds good

shane-huang:
BTW, is it possible to make this pr more general to allow more customization of gradients handing? Actually I need a little bit of more processing besides clipping. Can I set a callback Tensor -> Tensor to define my own processing of gradients before they are passed to OptimMethod? Better be a trait though, so I will be able to set some member variables.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
shane-huang: or shall we make it a single interface and the method of clipping as parameters? e.g. interface setGradClip(['const']) or setGradClip(['l2']) if I want one of the two, and if I want to stack those I shall write setGradClip(['const', 'l2']), the order also matters.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Optimizer.scala
```diff
@@ -284,6 +286,33 @@ abstract class Optimizer[T: ClassTag, D](
...
 
   def prepareInput(): Unit = {}
+
+  /**
+   * Set constant gradient clipping
+   * @param min the minimum value to clip by
+   * @param max the maximum value to clip by
+   * @return
+   */
+  def setConstantGradientClipping(min: Float, max: Float)
```
shane-huang: any way to disable clipping? People might use clipping in some stages of training and then diable it later. Also optimizer could be reused. Need a way to set to defaults.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Optimizer.scala
```diff
@@ -284,6 +286,33 @@ abstract class Optimizer[T: ClassTag, D](
...
+    gradientClippingParams.maxValueClip = max
+    this
+  }
+
+  /**
+   * Clip gradient to a maximum L2-norm
+   * @param clipNorm gradient L2-Norm threshold
+   * @return
+   */
+  def setGradientClippingByNorm(clipNorm: Float)
```
shane-huang: method name should include L2, because there're several kinds of norms (e.g. l1)

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Optimizer.scala
shane-huang: agree

shane-huang:
Please check my comments.

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
```diff
@@ -866,6 +866,38 @@ private[tensor] class DenseTensor[@specialized T: ClassTag](
...
+  }
+
+  override def clamp(min: Float, max: Float): Tensor[T] = {
+    val maxT = ev.fromType[Float](max)
+    val minT = ev.fromType[Float](min)
+    val func = new TensorFunc2[T] {
+      override def apply(data1: Array[T], offset1: Int): Unit = {
+        if (ev.isGreater(data1(offset1), maxT)) data1(offset1) = maxT
+        else if (ev.isGreater(minT, data1(offset1))) data1(offset1) = minT
+      }
```
shane-huang:
```scala
if (ev.isGreater(data1(offset1), maxT)) data1(offset1) = maxT
else if (ev.isGreater(minT, data1(offset1))) data1(offset1) = minT
```
could be 
```scala
ev.min(ev.max(data1(offset, minT)), maxT)
```

pyspark/bigdl/optim/optimizer.py
dding3: Updated

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Optimizer.scala
dding3: Added disable clipping method.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Optimizer.scala
dding3: Updated

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
dding3: The latter code is much cleaner than the previous one. But I think it may affect performance since it need compare twice every time

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/LocalOptimizer.scala
```diff
@@ -152,8 +152,31 @@ class LocalOptimizer[T: ClassTag] (
...
+        }
+        val l2Norm = (math.sqrt(sum) / parallelism).toFloat
+        
+        if (l2Norm > gradientClippingParams.normValueClip) {
+          scale = ev.fromType[Float]((l2Norm * parallelism) / gradientClippingParams.normValueClip)
+        }
+      }
+      grad.div(scale)
 
+      if (gradientClippingParams.enableConstantClipping) {
```
shane-huang: possible to make it a trait? so that people could define their own clipping.

jason-dai: I think we can leave it for future improvement.

spark/dl/src/main/scala/com/intel/analytics/bigdl/parameters/Parameter.scala
```diff
@@ -58,14 +58,14 @@ object SerializerInstance {
     this.pm = pm
   }
 
-  def serialize[T: ClassTag](data: Tensor[T]): CompressedTensor[T] = {
+  def create[T: ClassTag](data: Tensor[T]): CompressedTensor[T] = {
```
jason-dai: this is actually to serialize the tensor, not create

jason-dai:
LGTM

dding3:
Thanks for your review. Jekins passed.

Issue #1987, opened 2017-12-06
Model Load/Save document need to be update


yiheng:
See https://bigdl-project.github.io/master/#APIGuide/Module/

PR #1944, opened 2017-11-27
Support SperableConv2D and one hot crossentropy

yiheng:
## What changes were proposed in this pull request?
1. Add a new layer type SpatialSeperableConvolution
2. Add a new criterion type cross entropy which accepts onehot tensor as target input
3. Support to load tf depthwise operations

## How was this patch tested?
unit test

PR #2005, opened 2017-12-11
Add documentation for loading keras models

hkvision:

Add doc for keras model loading.


docs/docs/ProgrammingGuide/keras-support.md
```diff
@@ -0,0 +1,24 @@
+# **Keras Support**
```
zhichao-li: It can be moved to https://bigdl-project.github.io/master/#APIGuide/Module/

zhichao-li: related issue: https://github.com/intel-analytics/BigDL/issues/1987

hkvision:
@jason-dai any comments on the doc?

docs/docs/ProgrammingGuide/keras-support.md
```diff
@@ -0,0 +1,108 @@
...
+keras_model.add(Activation('relu'))
+keras_model.add(MaxPooling2D(pool_size=(2, 2)))
+keras_model.add(Dropout(0.25))
+
+keras_model.add(Flatten())
+keras_model.add(Dense(128))
+keras_model.add(Activation('relu'))
+keras_model.add(Dropout(0.5))
+keras_model.add(Dense(10))
+keras_model.add(Activation('softmax'))
```
zhichao-li: We should use `LogSoftMax` together with `ClassNLL`, or we can refactor this later after this PR (https://github.com/intel-analytics/BigDL/pull/1944 )merged

docs/docs/ProgrammingGuide/keras-support.md
```diff
@@ -0,0 +1,108 @@
...
+
+# Save the keras model to JSON file
+model_json = keras_model.to_json()
+def_path = "lenet.json"
+with open(def_path, "w") as json_file:
+    json_file.write(model_json)
+
```
zhichao-li: Divide this into 3 :  1) Pure keras model definition + data processing 2) save keras model 3)load keras model and train with BigDL optimizer. Maybe extract 3 methods to make your intention more clear.

docs/docs/ProgrammingGuide/keras-support.md
```diff
@@ -0,0 +1,108 @@
...
+y_train = sc.parallelize(y_train)
+train_data = X_train.zip(y_train).map(lambda t: Sample.from_ndarray(t[0], t[1] + 1))
+X_test = sc.parallelize(X_test)
+y_test = sc.parallelize(y_test)
+test_data = X_test.zip(y_test).map(lambda t: Sample.from_ndarray(t[0], t[1] + 1))
+optimizer = Optimizer(
+    model=bigdl_model,
+    training_rdd=train_data,
+    criterion=ClassNLLCriterion(),
+    optim_method=SGD(learningrate=0.01),
```
zhichao-li: tunning this?

docs/docs/APIGuide/Module.md
```diff
@@ -120,6 +120,24 @@ model = Model.load_tensorflow(model_def, inputs, outputs, byte_order = "little_e
...
+for more details. The Keras version we support and test is [__Keras 1.2.2__](https://faroit.github.io/keras-docs/1.2.2/).
+
+A Keras model definition in __JSON__ file can be loaded as a BigDL model.
+Saved weights in __HDF5__ file can also be loaded together with the architecture of a model.
+
+You can directly use the API `load_keras` to load the Keras model into BigDL.
+
+__Remark__: Packages `tensorflow`, `keras==1.2.2` and `h5py` are required. They can be installed via `pip` easily.
+
+```python
```
yiheng: scala API example?

docs/docs/APIGuide/Module.md
```diff
@@ -120,6 +120,24 @@ model = Model.load_tensorflow(model_def, inputs, outputs, byte_order = "little_e
...
+
+BigDL also supports loading pre-defined Keras models. See [keras support](https://bigdl-project.github.io/master/#ProgrammingGuide/keras-support/)
+for more details. The Keras version we support and test is [__Keras 1.2.2__](https://faroit.github.io/keras-docs/1.2.2/).
+
+A Keras model definition in __JSON__ file can be loaded as a BigDL model.
+Saved weights in __HDF5__ file can also be loaded together with the architecture of a model.
+
+You can directly use the API `load_keras` to load the Keras model into BigDL.
+
+__Remark__: Packages `tensorflow`, `keras==1.2.2` and `h5py` are required. They can be installed via `pip` easily.
```
yiheng: why tensorflow is required?

docs/docs/ProgrammingGuide/keras-support.md
yiheng: keep it here. That issue is not related with this PR.

docs/docs/APIGuide/Module.md
hkvision: tensorflow is the backend used by keras

yiheng: is one of the backend. For keras model save/load, I don't think we need tensorflow

docs/docs/ProgrammingGuide/keras-support.md
```diff
@@ -0,0 +1,81 @@
...
+
+## **Load a Keras model into BigDL**
+
+A Keras model definition in __JSON__ file can be loaded as a BigDL model.
+Saved weights in __HDF5__ file can also be loaded together with the architecture of a model.
+See [here](https://faroit.github.io/keras-docs/1.2.2/getting-started/faq/#how-can-i-save-a-keras-model) on how to save the architecture and weights of a Keras model.
+
+You can directly use the API `load_keras` to load the Keras model into BigDL.
+
+__Remark__: Packages `tensorflow`, `keras==1.2.2` and `h5py` are required. They can be installed via `pip` easily.
```
yiheng: don't need tensorflow?

docs/docs/ProgrammingGuide/keras-support.md
```diff
@@ -0,0 +1,81 @@
...
+
+```python
```
yiheng: where is the scala example?

docs/docs/ProgrammingGuide/keras-support.md
```diff
@@ -0,0 +1,81 @@
...
+# Load the JSON file to a BigDL model
+from bigdl.nn.layer import *
+bigdl_model = Model.load_keras(def_path=def_path)
+```
+After loading the model into BigDL, you can train it with MNIST dataset. See [here](https://github.com/intel-analytics/BigDL/blob/master/pyspark/bigdl/keras/examples/keras_lenet.py) for the full example code which includes the training and validation after model load. After 12 epochs, accuracy >97% can be achieved.
+
+### **Limitations**
+We have tested the model loading functionality with several standard [Keras applications](https://faroit.github.io/keras-docs/1.2.2/applications/) and [examples](https://github.com/fchollet/keras/tree/1.2.2/examples).
+
+However, there still exist some arguments for Keras layers that are not supported in BigDL for now. Also we haven't supported self-defined Keras layers, but one can still define your customized layer converter and weight converter method for new layers if you wish.
```
yiheng: Can we create a page provide the list of keras layer we have supported? like this: https://github.com/tf-coreml/tf-coreml#supported-ops

hkvision:
@yiheng Due to the limitation of our implementation, loading a keras model into BigDL need to rely on keras and its backend (mainly used for the calculation of input shape). In other words, if users need to use this functionality, they need to install keras and tensorflow beforehand. Otherwise if they don't load keras models, BigDL won't depend on keras. In addition, the API is only available on Python side.

hkvision:
jenkins passed http://172.168.2.101:8080/job/BigDL-PR-V/532/

hkvision:
Due to the limitation of our implementation, loading a keras model into BigDL need to rely on keras (mainly used for the calculation of input shape). In other words, if users need to use this functionality, they need to install `keras 1.2.2` beforehand.  `tensorflow` is not required now. Keras can use `theano` as backend as well. Tested with using `theano` to load a keras model. In addition, the API is only available on Python side.

hkvision:
Since keras needs a backend to construct the model, in the future we may consider using bigdl backend (only implement the minimum necessary methods) as suggested by Jason.

yiheng:
LGTM

jason-dai:
LGTM

Issue #1912, opened 2017-11-22
support models stored in HDFS, S3, etc


zhichao-li:
Currently we only support loading from local file.

wzhongyuan:
You may want to use FileWriter and FileReader which hide the underlying file systems

PR #2007, opened 2017-12-11
Support HDFS and S3 for Keras model loading

zhichao-li:

By leverage `sc.textfile` and `sc.binaryfile` we can easily support all Spark compatible file path.

## How was this patch tested?
unitest and manual test

fixed https://github.com/intel-analytics/BigDL/issues/1912

### DIFF
--- a/pyspark/bigdl/util/common.py
+++ b/pyspark/bigdl/util/common.py
@@ -589,6 +589,25 @@ def create_tmp_path():
     return tmp_file.name
 
 
+def text_from_path(path):
+    sc = get_spark_context()
+    return sc.textFile(path).collect()[0]
+
+
+def get_local_file(a_path):
+    if not is_distributed(a_path):
+        return a_path
+    path, data = get_spark_context().binaryFiles(a_path).collect()[0]
+    local_file_path = create_tmp_path()
+    with open(local_file_path, 'w') as local_file:
+        local_file.write(data)
+    return local_file_path
+
+
+def is_distributed(path):
+    return "://" in path
+
+
 def get_activation_by_name(activation_name, activation_id=None):
     """ Convert to a bigdl activation layer
         given the name of the activation as a string  """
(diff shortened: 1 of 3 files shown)


pyspark/bigdl/util/common.py
```diff
@@ -589,6 +589,25 @@ def create_tmp_path():
...
+    if not is_distributed(a_path):
+        return a_path
+    path, data = get_spark_context().binaryFiles(a_path).collect()[0]
+    local_file_path = create_tmp_path()
+    with open(local_file_path, 'w') as local_file:
+        local_file.write(data)
+    return local_file_path
+
+
+def is_distributed(path):
```
hkvision: Wondering if we can say a file is distributed if we find `://` in path.  Since for local files, we can also use `file:///`.

pyspark/test/bigdl/keras/test_load_model.py
```diff
@@ -64,5 +64,21 @@ def test_load_def_weights_kmodel_seq_lenet_mnist(self):
...
+        WeightLoader.load_weights_from_kmodel(bmodel, kmodel)
+        self.assert_allclose(bmodel.forward(input_data), kmodel.predict(input_data))
+
+    def test_load_weights(self):
+        kmodel, input_data, output_data = TestModels.kmodel_graph_1_layer()
+        keras_model_json_path, keras_model_hdf5_path = self._dump_keras(kmodel, dump_weights=True)
+        bmodel = DefinitionLoader.from_json_path(keras_model_json_path)
+        kmodel.set_weights([kmodel.get_weights()[0] + 100, kmodel.get_weights()[1]])
+        WeightLoader.load_weights_from_hdf5(bmodel, kmodel, filepath=keras_model_hdf5_path)
+        self.assert_allclose(bmodel.forward(input_data), kmodel.predict(input_data))
```
hkvision: Why adding these two tests instead of adding a test related to loading definition from HDFS?

pyspark/bigdl/util/common.py
zhichao-li: This is a just weak checking to improve performance.  It would works fine if the path starts with `file:///` in this case.

pyspark/test/bigdl/keras/test_load_model.py
zhichao-li: We don't have HDFS environment in Unitest and `sc.textfile` should be good to handle those different data storage, so what we need to do here is to ensure local storage works  fine. As for the other storage we can test it manually or add it into integration test.

zhichao-li:
Jenkins passed

PR #2015, opened 2017-12-12
Add object detection related layers for model zoo and keras support

SeaOfOcean:

Add object detection related layers

## How was this patch tested?
unit test


SeaOfOcean:
pr validate pass

PR #2016, opened 2017-12-12
Lookup table for multivalue

### LINKED ISSUE #1949: Layer which can process multi-value LookupTable is necessary
The layer "LookupTable" is useful to handle one-hot inputs, such as ID, category....

But,  it can't handle inputs having multi-hots, like: tags, histories...

Now, multi-hots' inputs can be handled by "SparseLinear”, but it's really inefficient.

So, a layer which is able to process multi-hots is necessary.

qiuxin2012:

Lookup table for multivalue

## How was this patch tested?

unit tests

fixed #1949

### DIFF
--- a/docs/docs/APIGuide/Layers/Embedding-Layers.md
+++ b/docs/docs/APIGuide/Layers/Embedding-Layers.md
@@ -94,7 +94,7 @@ gradInput = layer.backward(input, output)
 ```
 
 ---
-## LookupTable ##
+## LookupTableSparse ##
 
 **Scala:**
 ```scala
@@ -115,7 +115,7 @@ layer = LookupTableSparse(nIndex, nOutput,
 LookupTable for multi-values. 
 Also called embedding_lookup_sparse in TensorFlow. 
 
-The input of LookupTableSparse should be a 2D SparseTensor or a Table. If the input is a SparseTensor, the values are positive integer ids, values in each row of this SparseTensor will be turned into a dense vector. If the input is a Table, the first tensor in this table should be the integer ids, just like the SparseTensor input. And the second tensor in this table is the corresponding weights of the integer ids.
+The input of LookupTableSparse should be a 2D SparseTensor or two 2D SparseTensors. If the input is a SparseTensor, the values are positive integer ids, values in each row of this SparseTensor will be turned into a dense vector. If the input is a Table, the first tensor in this table should be the integer ids, just like the SparseTensor input. And the second tensor in this table is the corresponding weights of the integer ids.
 
 @param nIndex Indices of input row
 @param nOutput the last dimension size of output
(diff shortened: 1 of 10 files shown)


docs/docs/APIGuide/Layers/Embedding-Layers.md
```diff
@@ -93,3 +93,83 @@ gradInput = layer.backward(input, output)
...
+layer = LookupTableSparse(nIndex, nOutput,
+    combiner,
+    maxNorm,
+    wRegularizer)
+```
+
+LookupTable for multi-values. 
+Also called embedding_lookup_sparse in TensorFlow. 
+
+The input of LookupTableSparse should be a 2D SparseTensor of a Table. If the input is a SparseTensor, the values are positive integer ids, values in each row of this SparseTensor will be turned into a dense vector. If the input is a Table, the first tensor in this table should be the integer ids, just like the SparseTensor input. And the second tensor in this table is the corresponding weights of the integer ids.
```
zhangxiaoli73: or a Table

zhangxiaoli73:
it's better to add more docs or equations to explain your code

spark/dl/src/test/scala/com/intel/analytics/bigdl/nn/LookupTableSparseSpec.scala
```diff
@@ -0,0 +1,323 @@
...
+    exceptedGradWeight.select(1, 2).copy(Tensor(T(
+      29.20300217f, 34.20300217f, 39.20300217f, 44.20300217f)))
+    exceptedGradWeight.select(1, 4).copy(Tensor(T(
+      0.55000001f, 1.04999995f, 1.54999995f, 2.04999995f)))
+
+    output should be (exceptedOutput)
+    layer1.gradWeight should be (exceptedGradWeight)
+  }
+
+  "A LookupTableSparse sqrtn" should "generate correct output" in {
```
zhangxiaoli73: no sqrtn without weight?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/LookupTableSparse.scala
```diff
@@ -0,0 +1,300 @@
...
+
+  /**
+   * Compute the l2 norm clipping scale of the indices-th frame in tensor's first dimension.
+   * @return a HashMap contains l2 norm clipping scale
+   */
+  protected def norm2ScaleWithIndices[T: ClassTag](
+      tensor: Tensor[T],
+      indices: Tensor[T],
+      maxNorm: T)(implicit ev: TensorNumeric[T]): mutable.HashMap[Int, T] = {
+    val scale = mutable.HashMap[Int, T]()
```
zhangxiaoli73: use a buffer and clear each time

qiuxin2012:
jenkins passed

Issue #1568, opened 2017-09-19
Report proper error when caffe prototxt file and model file don't match in model loading


jason-dai:
Report proper error when caffe prototxt file and model file don't match in model loading

PR #2020, opened 2017-12-13
Add proper exception handling

wzhongyuan:
Add proper exception handling to make it more user friendly
## How was this patch tested?

Unit tests and local tests

fixed https://github.com/intel-analytics/BigDL/issues/1568

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/caffe/V1LayerConverter.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/caffe/V1LayerConverter.scala
@@ -41,10 +41,7 @@ class V1LayerConverter[T: ClassTag](implicit ev: TensorNumeric[T]) extends Conve
     val param = getConvolutionParam(layer).get
     val weightBlob = getBlob(layer, 0).get
     val biasBlob = getBlob(layer, 1)
-    if (!biasBlob.isDefined) {
-      throw new CaffeConversionException(s"${getLayerName(layer)} " +
-        s"without bias is not supported now")
-    }
+    val withBias = biasBlob.isDefined
     val group = if (param.getGroup == 0)  1 else param.getGroup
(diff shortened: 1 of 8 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/caffe/V1LayerConverter.scala
```diff
@@ -42,7 +42,8 @@ class V1LayerConverter[T: ClassTag](implicit ev: TensorNumeric[T]) extends Conve
     val weightBlob = getBlob(layer, 0).get
     val biasBlob = getBlob(layer, 1)
     if (!biasBlob.isDefined) {
-      throw new RuntimeException(s"${getLayerName(layer)} without bias is not supported now")
+      throw new CaffeConversionException(s"${getLayerName(layer)} " +
+        s"without bias is not supported now")
```
SeaOfOcean: It should be supported now

SeaOfOcean:
What if there is layer in prototxt definition but not in binary file?

wzhongyuan:
@SeaOfOcean  Refined on top of your comments. for the question 'what if there is layer in prototxt definition but not in binary',  I just  print a warning to user, the reason was maybe someone modify the prototxt and add some more layers which do not have weights or weight is not necessary on loading (maybe continue training after loading)

SeaOfOcean:
yes it is reasonable for user to modify some layers.
The problem is that if they append a new linear or convolution (layers with parameters),
since you need to use its weight blob to determine the nOutput, it will throw exception.

wzhongyuan:
@SeaOfOcean  Makes sense,  we should do sanity check where needed, will come back with that shortly

wzhongyuan:
@SeaOfOcean  pls review the updates

wzhongyuan:
PR validation passed http://172.168.2.101:8080/job/BigDL-PR-V/528/

PR #2026, opened 2017-12-14
 Add local model predictImage 

SeaOfOcean:

Add local model predictImage 

## How was this patch tested?

unit test


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -561,7 +561,7 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
    * @param predictKey key to store predicted result
    * @return
    */
-  def predictImage(imageFrame: ImageFrame,
+  def predictImage(imageFrame: DistributedImageFrame,
```
wzhongyuan: why not using just `predictImage` for both local and distributed and dispatching it to different handle according to ImageFrame type ? this would reduce the API exposed to users

wzhongyuan:
@SeaOfOcean  LGTM exception for the LocalModule & LocalPredictor usage but i think we should discuss in separate thread.

SeaOfOcean:
pr validate pass

PR #2029, opened 2017-12-14
set random seed on loaders spec

yangw1234:

set random seed on loaders spec

## How was this patch tested?

existing tests


yiheng:
I don't think we should set it in the base class. Please set it for specific case

yiheng:
LGTM

yangw1234:
pr validation pass

PR #2032, opened 2017-12-14
lable should be label

hkiang01:


yiheng:
LGTM. Thanks for the correction.

Issue #2027, opened 2017-12-14
Concate layer should pop proper infor to users if size mismatch


wzhongyuan:
Concat  requires only size of `dimension` for each output variable and other size should be the same among each output, we should do the size checking otherwise user may get unexpected exception

PR #2034, opened 2017-12-15
Concat check input size check

wzhongyuan:

Each module output inside concat is supposed to have the same dim size exception the one for concatenating, this change is to do the sanity check to make it more user friendly when exception happens

## How was this patch tested?

Unit test

fixed https://github.com/intel-analytics/BigDL/issues/2027

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Concat.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Concat.scala
@@ -64,6 +64,16 @@ class Concat[T: ClassTag](val dimension: Int)(
         this.size = currentOutput.size()
       } else {
         this.size(this.dimension - 1) += currentOutput.size(this.dimension)
+        require(this.size.length == currentOutput.size.length,
+        s"${this.modules(i).getName} output size mismatch, expected : ${this.size.length}," +
+          s"actual ${currentOutput.size.length}")
+        this.size.zipWithIndex.foreach(size => {
+          if (size._2 != dimension -1) {
+            require(size._1 == currentOutput.size(size._2 + 1),
+            s"${this.modules(i).getName} output size at dimension ${size._2 + 1} mismatch," +
+              s"expected ${size._1}, actual : ${currentOutput.size(size._2)}")
+          }
+        })
       }
       i += 1
     }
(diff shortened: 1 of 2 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Concat.scala
```diff
@@ -108,6 +110,27 @@ class Concat[T: ClassTag](val dimension: Int)(
     this.output
   }
 
+  private def checkOutputs(outputs : Array[Tensor[T]], dim : Int) : Unit = {
```
yiheng: Can we do this check at line 66? when update this.size also check other dimensinos

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Concat.scala
```diff
@@ -108,6 +110,27 @@ class Concat[T: ClassTag](val dimension: Int)(
...
+    if (outputs == null || outputs.length == 0) {
+      return
+    }
+    val size = outputs(0).size()
+    val dim = size.length
+    for ( i <- 1 until outputs.length) {
```
yiheng: avoid to use for here.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Concat.scala
wzhongyuan: @yiheng  makes sense,  we don't need to wait all module to compute before check, coming back shortly with refinement

wzhongyuan:
@yiheng  pls review the updates

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Concat.scala
```diff
@@ -64,6 +64,16 @@ class Concat[T: ClassTag](val dimension: Int)(
         this.size = currentOutput.size()
       } else {
         this.size(this.dimension - 1) += currentOutput.size(this.dimension)
+        require(this.size.length == currentOutput.size.length,
+        s"${this.modules(i).getName} output size mismatch, expected : ${this.size.length}," +
+          s"actual ${currentOutput.size.length}")
+        this.size.zipWithIndex.foreach(size => {
```
yiheng: try to use while instead of foreach for performance concern...

yiheng:
also add a unit test to verfy the exception?

wzhongyuan:
@yiheng  Refinement on top of your comments, please review

yiheng:
LGTM

wzhongyuan:
PR validation passed http://172.168.2.101:8080/job/BigDL-PR-V/560/

Issue #2014, opened 2017-12-12
Runtime.getRuntime().availableProcessors() / 2 in AllReduceParameter.scala may equals 0


glorysdj:
when I run bigdl toturials in machine with only one core in AWS, may encounter below error.
%%time
# Start to train
trained_model = optimizer.optimize()
print "Optimization Done."'

---------------------------------------------------------------------------
Py4JJavaError                             Traceback (most recent call last)
<ipython-input-6-7d568eb79eb3> in <module>()
----> 1 get_ipython().run_cell_magic(u'time', u'', u'# Start to train\ntrained_model = optimizer.optimize()\nprint "Optimization Done."')

/usr/local/lib/python2.7/dist-packages/IPython/core/interactiveshell.pyc in run_cell_magic(self, magic_name, line, cell)
   2115             magic_arg_s = self.var_expand(line, stack_depth)
   2116             with self.builtin_trap:
-> 2117                 result = fn(magic_arg_s, cell)
   2118             return result
   2119 

<decorator-gen-60> in time(self, line, cell, local_ns)

/usr/local/lib/python2.7/dist-packages/IPython/core/magic.pyc in <lambda>(f, *a, **k)
    186     # but it's overkill for just that one bit of state.
    187     def magic_deco(arg):
--> 188         call = lambda f, *a, **k: f(*a, **k)
    189 
    190         if callable(arg):

/usr/local/lib/python2.7/dist-packages/IPython/core/magics/execution.pyc in time(self, line, cell, local_ns)
   1191         else:
   1192             st = clock2()
-> 1193             exec(code, glob, local_ns)
   1194             end = clock2()
   1195             out = None

<timed exec> in <module>()

/tmp/spark-60c382ee-86b6-4e70-a807-a3bc54929b50/userFiles-2ddf2cb9-ff7e-4a6b-bc06-28146f94d7c2/bigdl-0.3.0-python-api.zip/bigdl/optim/optimizer.py in optimize(self)
    596         Do an optimization.
    597         """
--> 598         jmodel = callJavaFunc(get_spark_context(), self.value.optimize)
    599         from bigdl.nn.layer import Layer
    600         return Layer.of(jmodel)

/tmp/spark-60c382ee-86b6-4e70-a807-a3bc54929b50/userFiles-2ddf2cb9-ff7e-4a6b-bc06-28146f94d7c2/bigdl-0.3.0-python-api.zip/bigdl/util/common.py in callJavaFunc(sc, func, *args)
[...]

PR #2036, opened 2017-12-15
fix core number is 0 where there's only one core in system

yiheng:
fix core number is 0 where there's only one core in system

## How was this patch tested?
unit test

fixed https://github.com/intel-analytics/BigDL/issues/2014


yiheng:
jenkins pass

PR #2047, opened 2017-12-18
use flatmap instead of next in predictImage

SeaOfOcean:

use flatmap instead of next in predictImage

## How was this patch tested?

unit test


jason-dai:
LGTM

SeaOfOcean:
pr validate pass

PR #2060, opened 2017-12-19
Map Keras LocallyConnected1D

hkvision:

locallyconnected1d

## How was this patch tested?

Jenkins

### DIFF
--- a/pyspark/bigdl/keras/converter.py
+++ b/pyspark/bigdl/keras/converter.py
@@ -276,6 +276,13 @@ class WeightsConverter:
         return weights
 
     @staticmethod
+    def convert_locallyconnected1d(klayer, weights):
+        bweights1 = np.transpose(weights[0], (0, 2, 1))
+        if len(weights) == 1:  # if without bias
+            return [bweights1]
+        return[bweights1, weights[1]]
+
+    @staticmethod
     def convert_locallyconnected2d(klayer, weights):
         bweights1 = np.transpose(weights[0], (0, 2, 1))
         if len(weights) == 1:  # if without bias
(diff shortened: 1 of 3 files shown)


pyspark/bigdl/keras/converter.py
```diff
@@ -1614,7 +1621,7 @@ def create_highway(self):
         else:
             activation = get_activation_by_name(self.config["activation"],
                                                 "%s_%s" % (self.config["name"], self.config["activation"]))
-        blayer = BLayer.Highway(size=self.input_shape[1],
+        blayer = BLayer.Highway(size=int(self.input_shape[1]),
```
zhichao-li: is this a must that surround with int?

pyspark/bigdl/keras/converter.py
```diff
@@ -1190,7 +1197,7 @@ def create_atrousconvolution1d(self):
         if not self.config["bias"]:
             raise Exception("Only bias=True is supported for AtrousConvolution1D")
 
-        h = self.input_shape[2]
+        h = int(self.input_shape[1])
```
zhichao-li: add a dedicated unitest for this which would fail if we use `self.input_shape[2]`

zhichao-li:
LGTM pending on adding unit test.

pyspark/bigdl/keras/converter.py
hkvision: I don't know. But you surround everything related to shape with int and thus I do this for everything to be consistent.

hkvision:
jenkins passed http://172.168.2.101:8080/job/BigDL-PR-V/652/

PR #2074, opened 2017-12-20
[branch-0.4] Avoid annoying log after calling `engine.init`

### LINKED ISSUE #2070: call init_engine first time will create a SGD
Will see following log.
```
creating: createMaxEpoch
creating: createDefault
creating: createSGD
```

zhichao-li:

Avoid annoying log after calling `engine.init`

## How was this patch tested?
unittest

fixed #2070

### DIFF
--- a/pyspark/bigdl/optim/optimizer.py
+++ b/pyspark/bigdl/optim/optimizer.py
@@ -668,9 +668,9 @@ class Optimizer(BaseOptimizer):
     def create(model,
                training_set,
                criterion,
-               end_trigger=MaxEpoch(1),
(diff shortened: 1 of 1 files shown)


pyspark/bigdl/optim/optimizer.py
```diff
@@ -668,9 +668,9 @@ def __init__(self,
     def create(model,
                training_set,
                criterion,
-               end_trigger=MaxEpoch(1),
+               end_trigger=None,
```
hkvision: Wondering why `init_engine` will call this method?

PR #2092, opened 2017-12-22
Add transformer doc and some fix

SeaOfOcean:

Add transformer doc and some fix

## How was this patch tested?

unit test


SeaOfOcean:
jenkins pass

PR #2097, opened 2017-12-25
Add TransformerCriterion

yangw1234:

Add TransformerCriterion:

The criterion that takes two modules to transform input and target, and take
one criterion to compute the loss with the transformed input and target.

## How was this patch tested?

unit test


yangw1234:
pr validation pass

docs/docs/APIGuide/Losses.md
```diff
@@ -2186,3 +2186,62 @@ loss = criterion.forward(input, target)
...
+**Scala:**
+```scala
+val criterion = TransformerCriterion(criterion, Some(inputTransformer), Some(targetTransformer))
+```
+**Python:**
+```python
+criterion = TransformerCriterion(criterion, input_transformer, targetTransformer)
+```
+
+The criterion that takes two modules (optional) to transform input and target, and take
```
shane-huang: Elaborate the purpose of this criterion in doc. e.g. "This criterion is used for ..." or in what situations.

yangw1234: @shane-huang Added

yangw1234:
pr validation pass

PR #2099, opened 2017-12-25
add batchsize option for keras examples

hkvision:

In case when using spark-submit, batch size should be a multiple of core number

## How was this patch tested?

Manual test and jenkins


hkvision:
jenkins passed http://172.168.2.101:8080/job/BigDL-PR-V/678/

PR #2106, opened 2017-12-26
fix one node filter issue

wzhongyuan:

Fix filter if there is only one layer

## How was this patch tested?

unit test


wzhongyuan:
@SeaOfOcean  As discussed offline, check the output names when doing filtering . please help review, thanks

wzhongyuan:
PR validation passed
http://172.168.2.101:8080/job/BigDL-PR-V/682/

PR #2110, opened 2017-12-27
optimize localpredictor for small dataset

SeaOfOcean:

optimize localpredictor for small dataset

## How was this patch tested?

unit test


spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/LocalPredictor.scala
```diff
@@ -50,6 +50,14 @@ class LocalPredictor[T: ClassTag] private[optim](model: Module[T], weightsBias:
 
   private val batchPerCore = 4
 
+  private val weightsBias = Util.getAndClearWeightBias(model.cloneModule().parameters())
```
qiuxin2012: don't use cloneModule()

SeaOfOcean:
jenkins validation pass

qiuxin2012:
LGTM

Issue #1855, opened 2017-11-09
Error creating RnnCell in python when using activation=softplus or softsign


hkvision:
When I run the python unit test for creating rnn cell on master,
The following works with activation such as Tanh, Sigmoid, ReLU, SoftMax:
`layer= RnnCell(4, 3, Tanh(), True, True, L1Regularizer(0.5), L1Regularizer(0.5), L1Regularizer(0.5))`

However, if I change `Tanh()` to `SoftPlus()` or `SoftSign()`, I got the following error:

Py4JError: An error occurred while calling o27.createRnnCell. Trace:
py4j.Py4JException: Method createRnnCell([class java.lang.Integer, class java.lang.Integer, class com.intel.analytics.bigdl.nn.SoftPlus, class java.lang.Boolean, class java.lang.Boolean, class com.intel.analytics.bigdl.optim.L1Regularizer, class com.intel.analytics.bigdl.optim.L1Regularizer, class com.intel.analytics.bigdl.optim.L1Regularizer]) does not exist
    at py4j.reflection.ReflectionEngine.getMethod(ReflectionEngine.java:318)
    at py4j.reflection.ReflectionEngine.getMethod(ReflectionEngine.java:326)
    at py4j.Gateway.invoke(Gateway.java:272)
    at py4j.commands.AbstractCommand.invokeMethod(AbstractCommand.java:132)
    at py4j.commands.CallCommand.execute(CallCommand.java:79)
    at py4j.GatewayConnection.run(GatewayConnection.java:214)
    at java.lang.Thread.run(Thread.java:748)

hkvision:
@yiheng @yangw1234 now `SoftPlus` takes another `D` as tag but in RNN activation it only accepts `T`?

PR #2122, opened 2017-12-28
Wrap layers that requires two ClassTags into TensorModuleWrapper and move to tf package

### LINKED ISSUE #2115: TimeDistributed(Sum) is not correct
Run the code:
```
from bigdl.nn.layer import TimeDistributed,Sum
import numpy as np
layer = TimeDistributed(Sum(1, squeeze = False, n_input_dims = 2))
```
Will get the following error:
```
py4j.Py4JException: Method createTimeDistributed([class com.intel.analytics.bigdl.nn.Sum]) does not exist
	at py4j.reflection.ReflectionEngine.getMethod(ReflectionEngine.java:318)
	at py4j.reflection.ReflectionEngine.getMethod(ReflectionEngine.java:326)
	at py4j.Gateway.invoke(Gateway.java:272)
	at py4j.commands.AbstractCommand.invokeMethod(AbstractCommand.java:132)
	at py4j.commands.CallCommand.execute(CallCommand.java:79)
	at py4j.GatewayConnection.run(GatewayConnection.java:214)
	at java.lang.Thread.run(Thread.java:748)
```
Probably because of PR #1609  Other layers affected by this PR may also have this problem.

yangw1234:

Layers such as Sum, Abs, SoftPlus used to be TensorModule[T]. To support different types of Tensor in one module, we changed them to AbstractModule[Tensor[D], Tensor[D], T]. However, this caused some problems (some of which are hard to fix like #2115 #1855 ), confusions and complaints.

In this pr, I change those layers back to TensorModule[T], and add a new wrapper layer TensorModuleWrapper[T, D] in ops package that takes a TensorModule[D] to perform the computation of different types of tensors other than T.

## How was this patch tested?

unit tests

fixed #2115 #1855


wzhongyuan:
@yangw1234   Please refer to https://github.com/intel-analytics/BigDL/pull/2130 for the DataConverter fix. you may want to rebase it

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/ModuleToOperation.scala
```diff
@@ -43,3 +43,32 @@ object ModuleToOperation {
     (implicit ev: TensorNumeric[T]): ModuleToOperation[T] =
     new ModuleToOperation(model.asInstanceOf[AbstractModule[Activity, Activity, T]])
 }
+
+class TensorModuleToTFModule[T: ClassTag, D: ClassTag] private
+(val module: AbstractModule[Activity, Activity, D])
```
yiheng: should the module be TensorModule?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/ModuleToOperation.scala
```diff
@@ -43,3 +43,32 @@ object ModuleToOperation {
     (implicit ev: TensorNumeric[T]): ModuleToOperation[T] =
     new ModuleToOperation(model.asInstanceOf[AbstractModule[Activity, Activity, T]])
 }
+
+class TensorModuleToTFModule[T: ClassTag, D: ClassTag] private
```
yiheng: rename it to TensorModuleToOperation? Also please add java document to explain why we need this.

yangw1234:
pr validation pass

PR #2124, opened 2017-12-28
fix resize bilinear

yangw1234:

fix resize bilinear and add grad operation

## How was this patch tested?

unit tests


yangw1234:
pr validation pass

PR #2128, opened 2017-12-28
Add keras init_method & warning for unsupported init_method

hkvision:

For unsupported init methods, it's not reasonable if we just raise Exception and stop. Init_methods doesn't matter if users also load with weights.
So, raise a warning first for now if there's unsupported init_method.
To be discussed whether the init_methods in #1999 to be added in the next release.

## How was this patch tested?

Jenkins.

### DIFF
--- a/pyspark/bigdl/keras/converter.py
+++ b/pyspark/bigdl/keras/converter.py
@@ -1763,8 +1763,13 @@ class LayerConverter:
     def combo_parameter_layer(self, blayer, config):
         blayer.set_name(config["name"])
         if hasattr(blayer, "set_init_method"):
-            blayer.set_init_method(self.to_bigdl_init(config["init"]),
-                                   BInit.Zeros())  # Keras always set this to be zeros
+            try:
+                blayer.set_init_method(self.to_bigdl_init(config["init"]),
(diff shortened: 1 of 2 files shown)


pyspark/bigdl/keras/converter.py
```diff
@@ -1763,8 +1763,13 @@ def create_locallyconnected2d(self):
...
         if hasattr(blayer, "set_init_method"):
-            blayer.set_init_method(self.to_bigdl_init(config["init"]),
-                                   BInit.Zeros())  # Keras always set this to be zeros
+            try:
+                blayer.set_init_method(self.to_bigdl_init(config["init"]),
+                                       BInit.Zeros())  # Keras always set this to be zeros
+            except Exception:
+                warning_msg = "We don't support initialization " + config["init"] + " for now. " \
+                    + "Using the default instead."
+                warnings.warn(warning_msg)
```
zhichao-li: +                warnings.warn( "We don't support initialization %s for now. Using the default instead." %  config["init"] )

zhichao-li:
LGTM

hkvision:
jenkins passed http://172.168.2.101:8080/job/BigDL-PR-V/703/

PR #2130, opened 2017-12-29
support layer with different type

wzhongyuan:

Support nested layer with different types

## How was this patch tested?
unit test


wzhongyuan:
PR validation passed http://172.168.2.101:8080/job/BigDL-PR-V/706/

PR #2147, opened 2018-01-02
Fix Wrong labelSize at Sample.scala 411

sperlingxx:
"val labelSize = features.map(_.size())"   -> "val labelSize = labels.map(_.size())


qiuxin2012:
jenkins passed

PR #2150, opened 2018-01-03
Provide API to support sharing weights in Graph

### LINKED ISSUE #2080: Provide a general API to support weight sharing
Weight sharing is a very popular technology. BigDL has provided weight sharing in some container(e.g. recurrent). It's better if we convert this to a general API.

The API can refer to keras
```python
layer = Layer(...)
output1 = layer(input1)
output2 = layer(input2)
```

In the backend, when graph check one layer exists in multiple nodes, it can auto clone the layer and share their weights.

yangw1234:

Provide API to support sharing weights in Graph as suggested in #2080 

## How was this patch tested?
unit tests

fixed #2080

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
@@ -185,9 +185,10 @@ abstract class Graph[T: ClassTag](
     val layerToNodes = forwardNodes.filter(n => !n.eq(dummyOutput))
       // use System.identityHashCode to make sure two layers are actually the
       // same instance.
-      .groupBy(n => (System.identityHashCode(n.element), n.element))
-    layerToNodes.foreach { case ((_, layer), nodes) =>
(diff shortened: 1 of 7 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -75,6 +75,33 @@ abstract class Graph[T: ClassTag](
...
+
+  // Check all inputs of the graph should be passed in
+  checkRoots
+
+  protected var dummyOutputGrad: ModuleNode[T] = _
+  protected var backwardGraph: DirectedGraph[AbstractModule[Activity, Activity, T]] = _
+  protected var backwardNodes: Array[Node[AbstractModule[Activity, Activity, T]]] = _
+
+  private var stopGradientLayers: util.HashSet[String] = _
+
```
yangw1234: I move all the initialization stuff here.

yiheng:
Can we do the module clone and share in the inputs function?

PR #2151, opened 2018-01-03
Fix SparseMinibatch when batching DenseTensor

yangw1234:

When batch multiple tensors, it should add another dimension instead of concat on the original dimension.

## How was this patch tested?

unit tests


yangw1234:
pr validation pass

PR #2159, opened 2018-01-04
Fix bug of SparseTensor.resize(size, nElement)

sperlingxx:
According to the comment, if size.length > dimension,  additional indices should be in front of _indices array instead of behind it.

So, at SparseTensor.scala Line 482:
  >    _indices = _addIndices ++ _indices
should replace 
  >    _indices ++= _addIndices

### DIFF
--- a/spark/dl/src/test/scala/com/intel/analytics/bigdl/tensor/SparseTensorSpec.scala
+++ b/spark/dl/src/test/scala/com/intel/analytics/bigdl/tensor/SparseTensorSpec.scala
@@ -19,6 +19,8 @@ package com.intel.analytics.bigdl.tensor
 import org.scalatest.{FlatSpec, Matchers}
 import com.intel.analytics.bigdl.numeric.NumericFloat
 
+import scala.util.Random
+
 @com.intel.analytics.bigdl.tags.Parallel
 class SparseTensorSpec  extends FlatSpec with Matchers {
   "dim, shape, nElement" should "return right result" in {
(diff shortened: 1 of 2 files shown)


spark/dl/src/test/scala/com/intel/analytics/bigdl/tensor/SparseTensorSpec.scala
```diff
@@ -99,6 +101,14 @@ class SparseTensorSpec  extends FlatSpec with Matchers {
...
   }
 
+  "resize tensor to higher dim when nElement < sum(size)" should "return right result" in {
+    val indices = Array(Array(0, 4, 5, 7, 9))
+    val values = Array.fill(5)(Random.nextFloat())
+    val sTensor = Tensor.sparse(indices, values, Array(10))
+    sTensor.resize(Array(1, 10), 5)
+    Tensor.dense(sTensor).squeeze().toArray().sum should be (values.sum)
+  }
+
```
qiuxin2012: style check failed with following error message:
```
error file=/var/jenkins_home/workspace/BigDL-PR-V-StyleCheck/spark/dl/src/test/scala/com/intel/analytics/bigdl/tensor/SparseTensorSpec.scala message=Whitespace at end of line line=111 column=0
```

sperlingxx: Sorry, it has been fixed now.

qiuxin2012:
jenkins passed

PR #2162, opened 2018-01-04
Tensor.dense() support narrowed SparseTensor

sperlingxx:
Current implementation can not work properly when nElement  is not equal to _indices.length. It usually happens after narrow a SparseTensor. 
This commit solve this problem by taking account of offsets(_indicesOffset, _storageOffset) and nElement.

### DIFF
--- a/spark/dl/src/test/scala/com/intel/analytics/bigdl/tensor/SparseTensorSpec.scala
+++ b/spark/dl/src/test/scala/com/intel/analytics/bigdl/tensor/SparseTensorSpec.scala
@@ -109,7 +109,7 @@ class SparseTensorSpec  extends FlatSpec with Matchers {
     sTensor.storage().array.length should be (30)
     sTensor.storageOffset() should be (6)
   }
-  
+
   "Tensor.dense narrowed tensor" should "return right result" in {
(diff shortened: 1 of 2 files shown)


spark/dl/src/test/scala/com/intel/analytics/bigdl/tensor/SparseTensorSpec.scala
```diff
@@ -107,5 +109,13 @@ class SparseTensorSpec  extends FlatSpec with Matchers {
     sTensor.storage().array.length should be (30)
     sTensor.storageOffset() should be (6)
   }
+
```
qiuxin2012: Style check failed.
```
error file=/var/jenkins_home/workspace/BigDL-PR-V-StyleCheck/spark/dl/src/test/scala/com/intel/analytics/bigdl/tensor/SparseTensorSpec.scala message=Whitespace at end of line line=112 column=0
```

sperlingxx: Sorry, it has been fixed now.

qiuxin2012:
jenkins passed

PR #2163, opened 2018-01-08
Add sizes for input layer to support caffe to specify input dimensions

wzhongyuan:

Add sizes for input layer to support caffe to specify input dimensions

## How was this patch tested?

unit test


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Input.scala
```diff
@@ -36,7 +36,7 @@ import scala.reflect.ClassTag
  * @tparam T The numeric type in the criterion, usually which are [[Float]] or [[Double]]
  */
 @SerialVersionUID(- 8525406230282608924L)
-class Input[T: ClassTag]()(implicit ev: TensorNumeric[T])
+class Input[T: ClassTag](val sizes: Array[Int] = null)(implicit ev: TensorNumeric[T])
```
yiheng: I suggest defining a new Shape class. We will use it in shape infer

yiheng: How do we deal with unfixed size, e.g. batch size dimension

wzhongyuan:
@yiheng   yes, but we still need this PR , i will rebase and update

wzhongyuan:
@yiheng   rebased and updated, and refined the test files to apply the change
PR validation passed http://172.168.2.101:8080/job/BigDL-PR-Validation-v3/90/

Issue #1782, opened 2017-10-27
Move add method from Container to specific subclass


yiheng:
Container has an add method, while some container should not have add, e.g. graph. Should we remove the add method from container and implement it only in subclass which really need it?

jason-dai:
we can have a BaseContainer class with no add method

PR #2166, opened 2018-01-08
Refine container

yiheng:
Refine the implementation of containers, include:
1. Move the add method to a subclass dynamic container, as some containers may not have an `add` action

## How was this patch tested?
unit test

fixed https://github.com/intel-analytics/BigDL/issues/1782

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Container.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Container.scala
@@ -29,7 +29,7 @@ import scala.reflect.ClassTag
  * [[Container]] is an abstract [[AbstractModule]] class which
  * declares methods defined in all containers. A container usually
  * contain some other modules in the `modules` variable. It overrides
- * many module methods such that calls are propogated to the contained
+ * many module methods such that calls are propagated to the contained
  * modules.
  *
  * @tparam A Input data type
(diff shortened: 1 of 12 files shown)


jason-dai:
Maybe we can add another class, such as DynamicContainer, which extends Container and support add?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Container.scala
```diff
@@ -29,33 +29,21 @@ import scala.reflect.ClassTag
...
  * @tparam T Numeric type. Only support float/double now
  */
 @SerialVersionUID(- 2120105647780417237L)
-abstract class Container[A <: Activity : ClassTag,
-    B <: Activity : ClassTag, T: ClassTag](
+abstract class Container[A <: Activity : ClassTag, B <: Activity : ClassTag, T: ClassTag](
   implicit ev: TensorNumeric[T]) extends AbstractModule[A, B, T] {
 
   // list of sub modules
   val modules: ArrayBuffer[AbstractModule[Activity, Activity, T]]
```
zhichao-li: user can still add module if we expose `ArrayBuffer` here or we can put it into constructor and make it immutable.

yiheng: It is not an immutable object... I think maybe we can make a defense copy here. The user can get a copy of modules but cannot change the original one.

However, this still breaks the original API behavior. Maybe break some client code.

PR #2167, opened 2018-01-08
fix s3 integration test

wzhongyuan:
fix s3 integration test

## How was this patch tested?

unit test and local test


wzhongyuan:
PR validation passed http://172.168.2.101:8080/job/BigDL-PR-V/740/

PR #2169, opened 2018-01-08
Keras-like API for training

hkvision:

First version, subject to modifications.
Like keras,
model.compile
model.fit
cc @jason-dai @zhichao-li 

## How was this patch tested?

Jenkins

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
@@ -816,5 +816,47 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag, 
   def getClassTagNumerics() : (Array[ClassTag[_]], Array[TensorNumeric[_]]) = {
     (Array(scala.reflect.classTag[T]), Array(ev))
   }
+
+  protected var optimMethod: OptimMethod[T] = null
+  protected var criterion: Criterion[T] = null
+  protected var vMethods: Array[ValidationMethod[T]] = null
+
+  /**
+   * Configures the learning process.
+   * Must call this before fit.
+   */
+  def compile(optimizer: OptimMethod[T], loss: Criterion[T],
+              metrics: Array[ValidationMethod[T]] = null): Unit = {
+    // TODO: shape inference
+    this.optimMethod = optimizer
+    this.criterion = loss
+    this.vMethods = metrics
+  }
+
+  /**
+   * Trains the model for a fixed number of epochs.
+   */
+  def fit(x: RDD[Sample[T]], batchSize: Int = 32, epochs: Int = 10,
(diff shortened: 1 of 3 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -816,5 +816,47 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
...
+    // TODO: shape inference
+    this.optimMethod = optimizer
+    this.criterion = loss
+    this.vMethods = metrics
+  }
+
+  /**
+   * Trains the model for a fixed number of epochs.
+   */
+  def fit(x: RDD[Sample[T]], batchSize: Int = 32, epochs: Int = 10,
```
jason-dai: Need to check if the model has been compiled

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -816,5 +816,47 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
   def getClassTagNumerics() : (Array[ClassTag[_]], Array[TensorNumeric[_]]) = {
     (Array(scala.reflect.classTag[T]), Array(ev))
   }
+
+  protected var optimMethod: OptimMethod[T] = null
+  protected var criterion: Criterion[T] = null
+  protected var vMethods: Array[ValidationMethod[T]] = null
```
jason-dai: Maybe add one single object that contains all these variables?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -816,5 +816,47 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
...
+
+  /**
+   * Configures the learning process.
+   * Must call this before fit.
+   */
+  def compile(optimizer: OptimMethod[T], loss: Criterion[T],
+              metrics: Array[ValidationMethod[T]] = null): Unit = {
+    // TODO: shape inference
```
jason-dai: Can we compile the model multiple times?

hkvision: Currently the behavior is that if compiled multiple times, the last compile configurations will take effect.

zhichao-li:
I'm thinking maybe this keras-like API should be with the keras.Model or keras.Sequential instead of mix with the original `AbstractModule`.

hkvision:
@zhichao-li Yeah. After adding more layers, I will remove it into keras packages. cc @jason-dai

hkvision:
Opened a new one. Closed this one.

PR #2177, opened 2018-01-10
Add default zeroGradParameters to AbstractModule

SeaOfOcean:
training will go wrong if layer does not zeroGradParameters

Solution: Add default zeroGradParameters to AbstractModule,
the default behavior is to zero gradParameters
It can avoid missing override zeroGradParameters

## How was this patch tested?
unit test

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
@@ -258,10 +258,6 @@ object DistriOptimizer {
             val taskSize = gradLength / _subModelNumber
             val extraTask = gradLength % _subModelNumber
 
-            (0 until _subModelNumber).diff(finishedThreads).foreach(i =>
-              cached.modelGradients(i).zero()
-            )
-
             // Aggregate multi-model's gradient to the first model's gradient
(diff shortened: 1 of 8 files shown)


SeaOfOcean:
jenkins pass

yiheng:
Can you give more details about the root cause of the issue and how do you fix it? It's not easy to understand from the code change.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -258,6 +258,10 @@ object DistriOptimizer {
             val taskSize = gradLength / _subModelNumber
             val extraTask = gradLength % _subModelNumber
 
+            (0 until _subModelNumber).diff(finishedThreads).foreach(i =>
```
yiheng: Please give more description of the fix. I don't know why make this change without more background.

yiheng:
I think this doesn't solve the problem, e.g. a new layer with parameters may still miss override the zeroGradient.

You should change the behavior of the zeroGradient in the AbstractModule

yiheng:
Please add more information in the description of the PR. E.g. what's the problem, what's the root cause, how to fix this

SeaOfOcean:
updated @yiheng please help have a review, thanks

yiheng:
LGTM

PR #2178, opened 2018-01-11
Fix optimizer state messed up when calling optimize() multiple times in DistriOptimizer

yangw1234:

This pr fixed the following issues:
1. Originally, when the call `optimize()` is ended , the "epoch" and 
"neval" state in optimMethod is not update for the last iteration.  So when `optimize` is called again, it will run for one more iteration or epoch then expected.

2. When `optimize()` is called multiple times, there will be multiple copy of models in executor.

3. When `setTrainData` is called and current epoch is not finished, the current implementation does not start a new epoch.

## How was this patch tested?

unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
@@ -111,12 +111,24 @@ object DistriOptimizer {
     val partitionNum = dataset.originRDD().partitions.length
     var wallClockTime = 0L
     var lastEpochTime = 0L
-    state("epoch") = state.getOrElse("epoch", 1)
-    state("neval") = state.getOrElse("neval", 1)
-    state("Loss") = state.getOrElse("Loss", Float.PositiveInfinity)
-    state("score") = state.getOrElse("score", 0f)
-    state("recordsProcessedThisEpoch") = state.getOrElse("recordsProcessedThisEpoch", 0)
-    var recordsProcessedThisEpoch = state[Int]("recordsProcessedThisEpoch")
+
+    // driverState is needed to prevent serializing the whole optimizer
+    if (!optimMethod.state.contains("epoch")) optimMethod.state.update("epoch", 1)
+    if (!optimMethod.state.contains("neval")) optimMethod.state.update("epoch", 1)
+    if (!optimMethod.state.contains("Loss")) {
+      optimMethod.state.update("epoch", Float.PositiveInfinity)
+    }
+    if (!optimMethod.state.contains("score")) optimMethod.state.update("epoch", 0f)
+    if (!optimMethod.state.contains("recordsProcessedThisEpoch")) {
+      optimMethod.state.update("recordsProcessedThisEpoch", 0)
+    }
+    val driverState = T(
+      "epoch" -> optimMethod.state("epoch"),
+      "neval" -> optimMethod.state("neval"),
+      "Loss" -> optimMethod.state("Loss"),
+      "score" -> optimMethod.state("score")
+    )
+
     val _subModelNumber = Engine.getEngineType() match {
       case MklBlas => coresPerNode
     }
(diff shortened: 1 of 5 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -111,12 +111,12 @@ object DistriOptimizer {
     val partitionNum = dataset.originRDD().partitions.length
     var wallClockTime = 0L
     var lastEpochTime = 0L
-    val driverState = T(
-      "epoch" -> optimMethod.state.get[Int]("epoch").getOrElse(1),
-      "neval" -> optimMethod.state.get[Int]("neval").getOrElse(1),
-      "Loss" -> optimMethod.state.get[Float]("Loss").getOrElse(Float.PositiveInfinity),
-      "score" -> optimMethod.state.get[Float]("score").getOrElse(0f)
-    )
+    state("epoch") = state.getOrElse("epoch", 1)
```
yiheng: why remove the driverState? I think there should be some reason to use driverState in the old code

yiheng:
I can't understand what problem do you solve. Please add more information into the description.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -161,8 +156,16 @@ object DistriOptimizer {
 
     var epochStart = System.nanoTime()
     var dataRDD = dataset.data(train = true)
-    var recordsProcessedThisEpoch = 0
-    while (!endWhen(driverState)) {
+
+    while (!endWhen(state)) {
+      if (recordsProcessedThisEpoch == 0) {
```
yiheng: why move this block into while loop?

yiheng:
any unit test for your new change in scala?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -809,13 +813,22 @@ class DistriOptimizer[T: ClassTag] (
     }).count()
   }
 
+  private def endEpoch(): Unit = {
+    val records = this.state.get[Int]("recordsProcessedThisEpoch")
+    if (records.isDefined && records.get != 0) {
+      this.state("epoch") = this.state[Int]("epoch") + 1
```
yiheng: why epoch add 1?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -946,6 +960,7 @@ class DistriOptimizer[T: ClassTag] (
 
     // Reset some internal states, so this or other optimizers can run optimize again
     clearState()
+    models.unpersist()
```
yiheng: why unpersisit model?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
yiheng: please add some comments

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
yiheng: please add some comments

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
yiheng: Please add a comment why use driverState

yiheng:
LGTM

yangw1234:
pr validation pass

PR #2179, opened 2018-01-11
Set weight ref to support share params at tensor level

wzhongyuan:

Set weight ref to support share params at tensor level

## How was this patch tested?
unit test

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Add.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Add.scala
@@ -32,7 +32,7 @@ import scala.reflect.ClassTag
 class Add[T: ClassTag](val inputSize: Int
   )(implicit ev: TensorNumeric[T]) extends TensorModule[T] with Initializable {
 
-  val bias = Tensor[T](inputSize)
+  var bias = Tensor[T](inputSize)
 
   val ones : Tensor[T] = Tensor[T]()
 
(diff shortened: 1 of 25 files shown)


wzhongyuan:
@yangw1234   Please review the change, thanks

jason-dai:
I think it is somewhat dangerous to change val to var.

wzhongyuan:
@yangw1234  close it for now

yiheng:
@jason-dai maybe we can set it to a private var and the modification method can only be accessed in a limited scope. I think there's potential issues if we share weight only share storage

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -354,6 +354,9 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
    */
   def parameters(): (Array[Tensor[T]], Array[Tensor[T]]) = null
 
+
+  def setParameters(params : Array[Tensor[T]]): Unit = {}
```
yiheng: private[bigdl]?

yiheng: should we group the parameter relate API to a trait? As the AbstractModule is too large now...

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Add.scala
```diff
@@ -32,7 +32,7 @@ import scala.reflect.ClassTag
 class Add[T: ClassTag](val inputSize: Int
   )(implicit ev: TensorNumeric[T]) extends TensorModule[T] with Initializable {
 
-  val bias = Tensor[T](inputSize)
+  var bias = Tensor[T](inputSize)
```
yiheng: -> private var _bias
def bias = _bias

yiheng: same as the below

yangw1234:
[https://github.com/torch/nn/blob/master/doc/module.md#sharemlps1s2sn](https://github.com/torch/nn/blob/master/doc/module.md#sharemlps1s2sn)
Can we do something like this?

1. Make parameter private
2. Do **not** provide set method, so that the parameter can only be modified by `clone`/`share`

PR #2183, opened 2018-01-11
Module DenseToSparse support backward disable (or not) setting

sperlingxx:
This feature is useful when DenseToSparse layer is placed on the top of (sequential) model, received gradient from a SparseLinear layer. 

After setting isBackward false(default: true), there is no need to do gradient backpropagation at SparseLinear layer, which will make whole backward procedure more efficient and convenient.

## How was this patch tested?
 unit test

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/DenseToSparse.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/DenseToSparse.scala
@@ -24,11 +24,11 @@ import scala.reflect.ClassTag
 
 /**
  * Convert DenseTensor to SparseTensor.
- * @param ev$1
- * @param ev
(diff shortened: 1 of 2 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/DenseToSparse.scala
```diff
@@ -24,29 +24,31 @@ import scala.reflect.ClassTag
 
 /**
  * Convert DenseTensor to SparseTensor.
- * @param ev$1
- * @param ev
+ * @param isBackward whether update gradInput during backward
```
qiuxin2012: Please change `isBackward` to `val propagateBack: Boolean = true, // propagate gradient back`, like SpatialConvolution.

PR #2184, opened 2018-01-12
Load imagenet sequence file to ImageFrame

SeaOfOcean:

Load imagenet sequence file to ImageFrame

## How was this patch tested?

unit test

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/DataSet.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/DataSet.scala
@@ -16,12 +16,15 @@
 
 package com.intel.analytics.bigdl.dataset
 
+import java.nio.ByteBuffer
 import java.nio.file.{Files, Path, Paths}
 import java.util.concurrent.atomic.AtomicInteger
 
 import com.intel.analytics.bigdl.DataSet
 import com.intel.analytics.bigdl.dataset.image.{LabeledBGRImage, _}
-import com.intel.analytics.bigdl.utils.{Engine, RandomGenerator}
+import com.intel.analytics.bigdl.tensor.Tensor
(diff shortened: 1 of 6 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/DataSet.scala
```diff
@@ -548,6 +551,32 @@ object DataSet {
...
+     * @param url sequence files folder path
+     * @param sc spark context
+     * @param classNum class number of data
+     * @param partitionNum partition number, default: Engine.nodeNumber() * Engine.coreNumber()
+     * @return
+     */
+    private[bigdl] def filesToImageFrame(url: String, sc: SparkContext,
+      classNum: Int, partitionNum: Option[Int] = None): ImageFrame = {
+      val num = partitionNum.getOrElse(Engine.nodeNumber() * Engine.coreNumber())
+      val rawData = sc.sequenceFile(url, classOf[Text], classOf[Text], num).map(image => {
```
jenniew: suggested to change the value type to ByteWritable. Also change the value type in BGRImgToLocalSeqFile to ByteWritable. Load Text value in python seq file API would get encoding error if the true value type is byte although we only load seq file in scala.

spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDL.scala
```diff
@@ -2889,6 +2893,12 @@ class PythonBigDL[T: ClassTag](implicit ev: TensorNumeric[T]) extends Serializab
     ImageFrameToSample[T](inputKeys.asScala.toArray, targets, sampleKey)
   }
 
+  def filesToImageFrame(url: String, sc: JavaSparkContext,
```
jenniew: suggest to change name to seqfilesToImageFrame, etc. Avoid to misunderstand to change image files to imageframe.

PR #2192, opened 2018-01-16
[WIP] add IO feature for DLModel/DLClassifierModel

sperlingxx:

add IO feature for DLModel/DLClassifierModel

## How was this patch tested?

unit test

### DIFF
--- a/spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
+++ b/spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
@@ -36,6 +36,7 @@ import org.json4s.JsonDSL._
 import org.json4s.{DefaultFormats, JObject}
 
 import scala.reflect.ClassTag
+import scala.util.Try
 
 private[ml] trait HasBatchSize extends Params {
 
@@ -65,7 +66,7 @@ private[ml] trait OptParams[@specialized(Float, Double) T] extends Params {
   /**
    * When to stop the training, passed in a [[Trigger]]. E.g. Trigger.maxIterations
    */
-  final val endWhen = new Param[Trigger](this, "endWhen", "Trigger to stop the training")
+  final val endWhen = new TriggerParam(this, "endWhen", "Trigger to stop the training")
 
   def getEndWhen: Trigger = $(endWhen)
 
@@ -99,9 +100,9 @@ private[ml] trait OptParams[@specialized(Float, Double) T] extends Params {
    * SGD and LBFGS. Refer to package com.intel.analytics.bigdl.optim for all the options.
    * Default: SGD
    */
-  final val optimMethod = new Param[OptimMethod[T]](this, "optimMethod", "optimMethod")
+  final val optimMethod = new OptimMethodParam(this, "optimMethod", "optimMethod")
 
-  def getOptimMethod: OptimMethod[T] = $(optimMethod)
+  def getOptimMethod: OptimMethod[T] = $(optimMethod).asInstanceOf[OptimMethod[T]]
 }
 
 
@@ -334,7 +335,7 @@ class DLEstimator[@specialized(Float, Double) T: ClassTag](
     val endTrigger = if (isSet(endWhen)) $(endWhen) else Trigger.maxEpoch($(maxEpoch))
     val optimizer = Optimizer(model, trainingSamples, criterion, $(batchSize))
       .setState(state)
-      .setOptimMethod($(optimMethod))
+      .setOptimMethod($(optimMethod).asInstanceOf[OptimMethod[T]])
       .setEndWhen(endTrigger)
 
     if (validationTrigger.isDefined) {
@@ -369,6 +370,7 @@ class DLEstimator[@specialized(Float, Double) T: ClassTag](
     copyValues(new DLEstimator(model, criterion, featureSize, labelSize), extra)
   }
 }
+
 
 /**
  * [[DLModel]] helps embed a BigDL model into a Spark Transformer, thus Spark users can
--- a/spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
+++ b/spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
@@ -15,82 +15,96 @@
  */
 package org.apache.spark.ml
 
-import java.net.URI
-
 import com.intel.analytics.bigdl.dataset._
 import com.intel.analytics.bigdl.models.utils.ModelBroadcast
 import com.intel.analytics.bigdl.optim._
-import com.intel.analytics.bigdl.tensor.Tensor
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+import com.intel.analytics.bigdl.tensor.{Tensor, DoubleType => TensorDouble, FloatType => TensorFloat}
+import com.intel.analytics.bigdl.utils.T
 import com.intel.analytics.bigdl.utils.serializer.ModuleLoader
-import com.intel.analytics.bigdl.utils.{File, T}
 import com.intel.analytics.bigdl.visualization.{TrainSummary, ValidationSummary}
 import com.intel.analytics.bigdl.{Criterion, Module}
-import opennlp.tools.cmdline.ModelLoader
-import org.apache.commons.lang3.SerializationUtils
 import org.apache.hadoop.fs.Path
+import org.apache.spark.SparkContext
 import org.apache.spark.ml.param.shared.{HasFeaturesCol, HasPredictionCol}
 import org.apache.spark.ml.param.{IntParam, ParamMap, ParamValidators, _}
-import org.apache.spark.ml.util.SchemaUtils
+import org.apache.spark.ml.util._
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.types._
 import org.apache.spark.sql.{DataFrame, Row}
+import org.json4s.{DefaultFormats, JObject}
 
 import scala.reflect.ClassTag
 
 private[ml] trait HasBatchSize extends Params {
 
-  final val batchSize: Param[Int] = new Param[Int](this, "batchSize", "batchSize")
+  final val batchSize: IntParam = new IntParam(this, "batchSize", "batchSize")
 
   def getBatchSize: Int = $(batchSize)
 }
 
-/**
- * Common trait for DLEstimator and DLModel
- */
-private[ml] trait DLParams[@specialized(Float, Double) T] extends HasFeaturesCol
-  with HasPredictionCol with VectorCompatibility with HasBatchSize {
+private[ml] trait IsOverWrite extends Params {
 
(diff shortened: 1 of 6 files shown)


spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
```diff
@@ -425,14 +431,70 @@ class DLModel[@specialized(Float, Double) T: ClassTag](
   }
 
   override def copy(extra: ParamMap): DLModel[T] = {
-    val copied = new DLModel(model, featureSize, uid).setParent(parent)
+    val copied = new DLModel(_model, featureSize, uid).setParent(parent)
```
hhbyyh: can we just use model?

spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
```diff
@@ -425,14 +431,70 @@ class DLModel[@specialized(Float, Double) T: ClassTag](
...
+  protected def setModule(module: Module[T]): Unit
 
+  /**
+   * Save a [[DLModelBase]] to a dictionary, which contains 3 files(meta, module, weight).
+   * @param path path of model dictionary, it should NOT exists as a File.
+   * @param isOverWrite whether overwrite earlier files
+   */
+  def save(path: String, isOverWrite: Boolean): Unit = {
+    val (metaPath, modulePath, weightPath) = DLModelBase.getPaths(path)
+    val meta = SerializationUtils.serialize(this)
```
hhbyyh: Is it based on Java serialization? I'm a little concerned since it may not be compatible across different Java/scala version.

spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
```diff
@@ -425,14 +431,70 @@ class DLModel[@specialized(Float, Double) T: ClassTag](
...
+  (path: String)(implicit ev: TensorNumeric[T]): DLModel[T] = {
+    DLModelBase.load[T, DLModel[T]](path).asInstanceOf[DLModel[T]]
+  }
+}
+
+/**
+ * Base class of DLModels, such as DLClassifierModel, provide (pipeline) model IO features.
+ * Please make sure working with module which returned by [[getModule]] and set by [[setModule]].
+ */
+abstract class DLModelBase[@specialized(Float, Double) T: ClassTag, M <: DLTransformerBase[M]]
```
hhbyyh: Is it possible to merge this into DLModel?

spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
sperlingxx: If  using model, model should be var instead of val, then it will be un-extendable.

spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
sperlingxx: I think DLModelBase can be extend by other model like transformer based on a bigdl.Module which need IO supporting.  
Actually, I am working on DAGEstimator/DAGModel which supporting feature engineering preprocess before training。

spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
sperlingxx: Yes, it is based on Java Serialization. Spark use Java Serialization(kyro is another option, but java is default) to do all serialization works. And BigDL Module serialized by [[SerializationUtils.serialize]] at DistriOptimizer.initThreadModels in eariler version.

spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
hhbyyh: I don't quite understand why we want to change model in a DLModel.

spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
hhbyyh: That's separate topic. Still I would vote to put the save/load code in DLModel for now, and we can migrate it to DLModelBase in your future PR.
As long as we keep the API stable, it will not impact user experience.

spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
hhbyyh: ML Model and Estimator does not use Java serialization for save/load. And I would recommend we follow the implementation in ML and leverage the existing infrastructure code.

sperlingxx: Okay,  so,  is it a good way to do saving/loading  by extending MLReader/MLWriter :)

spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
sperlingxx: DLClassifierModel.model overiding DLModel.model, if we modify DLModel.model to var, it will lose the ability of be overided (by DLClassifierModel.model).

sperlingxx: I think modelImpl is better name than _model :)

hhbyyh: Still I don't understand why model in DLModel needs to be altered.

spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
hhbyyh: Another reminder, We need to support both Spark 1.6 and Spark 2.X.

hhbyyh:
Hi @sperlingxx, Thanks for making the contribution.

Just want to confirm, is this an effort to support save/load for DLEstimator, DLModel to be compatible with Spark ML pipeline, or just to allow the save/load for DLModel (in any format). Since DLEstimator/DLModel is primarily designed for support of Spark DataFrame interface and compatibility with Spark ML Pipeline, we'd prefer the save/load function would be compatible with Spark ML pipeline. That means if user use DLModel as one step of their pipeline, they should be able to intuitively invoke the save/load function of ML Pipeline. 

The long term target is to support save load for both Spark 1.6 and 2.x. For now, we can start with one of them. Please also include unit test with Spark ML pipeline. Thanks.

spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
sperlingxx: Since we decide to implement [load] by MLReader, there is no need to alter [model] after deserializing [DLModel].

sperlingxx:
For now, save/load function is compatible with Spark ML pipeline.  It passed unit tests for spark_2.x, but when I built a version for spark_1.6, it came out to be spark_1.5.1.  MLReader/MLWriter are supported since 1.6.

yiheng:
what about adding -Dspark.version=1.6.2 to your mvn build option. It will use Spark 1.6.2.

@jason-dai @hhbyyh do we still need to maintain Spark 1.5? I think Spark2 should be main stream now.

hhbyyh:
@yiheng I'm OK with removing support for 1.5.

spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
```diff
@@ -15,36 +15,53 @@
...
 }
 
 /**
- * Common trait for DLEstimator and DLModel
+ *  Whether overwrite existed files while saving models.
+ *  Default: false
  */
-private[ml] trait DLParams[@specialized(Float, Double) T] extends HasFeaturesCol
-  with HasPredictionCol with VectorCompatibility with HasBatchSize {
+private[ml] trait IsOverWrite extends Params {
```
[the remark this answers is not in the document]
sperlingxx: Yes, I am using that. But it's a trait of MLWriter, while DLModel extends MLWritable. So, we need a trigger param for MLWriter.overwrite() which can be set by users.

spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
```diff
@@ -161,12 +186,12 @@ private[ml] trait DLParams[@specialized(Float, Double) T] extends HasFeaturesCol
...
-    val featureSize : Array[Int],
-    val labelSize : Array[Int],
-    override val uid: String = "DLEstimator")(implicit ev: TensorNumeric[T])
-  extends DLEstimatorBase[DLEstimator[T], DLModel[T]] with DLParams[T] {
+  @transient val model: Module[T],
+  val criterion : Criterion[T],
+  val featureSize : Array[Int],
+  val labelSize : Array[Int],
+  override val uid: String = "DLEstimator")(implicit ev: TensorNumeric[T])
+  extends DLEstimatorBase[DLEstimator[T], DLModel[T]] with DLParams[T] with OptParams[T] {
```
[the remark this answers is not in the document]
sperlingxx: It's a good idea :)

sperlingxx:
@hhbyyh

sperlingxx:
@hhbyyh 
1.   __Param__ wrapper for __optimMethod__ and __Trigger__
2.   __DLEstimatorSpec__ and __DLClassifierSpec__ support both spark_1.6 and spark_2.x, and all unit tests are passed in both version.
3.  Non-default values have been set in unit test for saving/loading function of DLModel.
4.  Saving/Loading a __DLEstimator__/__DLEClassifier__ is on _TODO List_, it will take other efforts.  I think maybe we can convert all member paras to __Param__(just like __GBTClassifier__)?

spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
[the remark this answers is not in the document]
hhbyyh: User can write something like dlModel.write.overwrite.save...

hhbyyh:
Thanks for the update. 

I'm OK with supporting save/load for DLModel only. In that case, maybe we can separate the customized params into another PR. 

And we still need to wait for the removal of Spark 1.5 support.

drod331:
Any update on this PR?  I was thinking of forking and adding this functionality myself, but I was glad to see this PR already existed.  I can take a look at implementing saving/loading for DLEstimator/DLClassifier in this PR or in a separate PR as well.

sperlingxx:
@drod331 After discussion, spark-1.5 will be not longer supported, we will finish this PR and merge it to master in recent  :-)

sperlingxx:
I've created another PR(https://github.com/intel-analytics/BigDL/pull/2514), which is compatible to latest master, for saving/loading features of DLModel.

qiuxin2012:
We can close this one.

PR #2194, opened 2018-01-16
bug fix: DLModel prediction

sperlingxx:

1.  Make sure DLModel.train=False when predicting in pipeline API
2.  Broadcast transformer(SampleToMiniBatch) rather than initializing them on executors, which will cause exception in spark-cluster mode because of calling _Engine.nodeNumber()_ on executors.

## How was this patch tested?

manual test

### DIFF
--- a/spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
+++ b/spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
@@ -391,7 +391,8 @@ class DLModel[@specialized(Float, Double) T: ClassTag](
     val localBatchSize = $(batchSize)
 
     val resultRDD = dataFrame.rdd.mapPartitions { rowIter =>
-      val localModel = modelBroadCast.value()
+      // call the evaluate method to enable DLModel.train=False during the predict process
+      val localModel = modelBroadCast.value().evaluate()
       rowIter.grouped(localBatchSize).flatMap { rowBatch =>
         val samples = rowBatch.map { row =>
           val features = featureFunc(row, featureColIndex)
--- a/spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
+++ b/spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
@@ -387,12 +387,13 @@ class DLModel[@specialized(Float, Double) T: ClassTag](
     val featureColIndex = dataFrame.schema.fieldIndex($(featuresCol))
     val featureFunc = getConvertFunc(featureType)
     val sc = dataFrame.sqlContext.sparkContext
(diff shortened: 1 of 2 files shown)


spark/dl/src/main/scala/org/apache/spark/ml/DLEstimator.scala
```diff
@@ -391,7 +391,8 @@ class DLModel[@specialized(Float, Double) T: ClassTag](
     val localBatchSize = $(batchSize)
 
     val resultRDD = dataFrame.rdd.mapPartitions { rowIter =>
-      val localModel = modelBroadCast.value()
+      // call the evaluate method to enable DLModel.train=False during the predict process
+      val localModel = modelBroadCast.value().evaluate()
```
hhbyyh: Is it better if we move evaluate before model broadcasting?

spark/dl/src/test/scala/com/intel/analytics/bigdl/optim/DLEstimatorSpec.scala
```diff
@@ -91,6 +91,27 @@ class DLEstimatorSpec extends FlatSpec with Matchers with BeforeAndAfter {
...
+  "An DLEstimator" should "throws exception when DLModel is predicting with DLModel.train=True" in {
+    val model = new Sequential().add(Linear[Float](6, 2))
+      .add(Dropout[Float](initP = 0.5))
+      .add(LogSoftMax[Float])
+    val criterion = ClassNLLCriterion[Float]()
+    val estimator = new DLEstimator[Float](model, criterion, Array(6), Array(1))
+      .setBatchSize(nRecords)
+      .setOptimMethod(new LBFGS[Float]())
+      .setLearningRate(0.1)
+      .setMaxEpoch(maxEpoch)
```
hhbyyh: let's set this to 1.

spark/dl/src/test/scala/com/intel/analytics/bigdl/optim/DLEstimatorSpec.scala
```diff
@@ -91,6 +91,27 @@ class DLEstimatorSpec extends FlatSpec with Matchers with BeforeAndAfter {
     assert(correct > nRecords * 0.8)
   }
 
+  "An DLEstimator" should "throws exception when DLModel is predicting with DLModel.train=True" in {
```
hhbyyh: Shall we use intercept exception in this unit test?

sperlingxx:
Hi @hhbyyh,
 Modifications are made about this PR.

hhbyyh:
Thanks for the update.

I'm not sure how much difference does it make to broadcast the transformer. Can you please elaborate?

Also an unit test is still nice to have. I can add that in a follow-up PR.

qiuxin2012:
jenkins passed

PR #2197, opened 2018-01-17
Add PGCriterion to compute the negative policy gradient given action distribution, sampled action and reward

yangw1234:

Add PGCriterion to compute the negative policy gradient given action distribution, sampled action and reward.

The input to PGCriterion is a vector (or batch of vectors) representing a multinomial distribution and the target is vector of the same size representing the sampled action and reward. Typically the target vector has only one non-zero entry whose index is the sampled action and value is the reward. The target can be SparseTensor.

Optionally, you can pass in a transformer to manipulate the input distribution like clipping or add negative entropy penalty to encourage exploration.

Result on @shane-huang 's  REINFORCE notebook:

criterion

## How was this patch tested?

unit test and manual test


yangw1234:
pr validation pass

yangw1234:
pr validation pass

PR #2202, opened 2018-01-19
Fix recurrent forward throwing exception after reset() 

yangw1234:

Fix recurrent forward throwing exception after reset() 

The following code would throw an exception:
```scala
val model = Sequential[Double]()
      .add(Recurrent[Double]()
        .add(RnnCell[Double](inputSize, hiddenSize, Tanh())))
      .add(Select(1, 1))
      .add(Linear[Double](hiddenSize, outputSize))
model.forward(input)
model.reset()
model.forward(input)
```
This pr fixes this issue.

## How was this patch tested?

unit tests


psyyz10:
LGTM

yangw1234:
pr validation pass

PR #2205, opened 2018-01-19
Add NCHW support for resizebilinear

yangw1234:
### Add nchw support for resizebilinear

ResizeBilinear layer is commonly used in computer vision models for upsampling. However, current implementation only support nhwc format with float data.

This pr extends ResizeBilinear to support NCHW format (which is the default format in BigDL) and double data.

## How was this patch tested?

unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ResizeBilinear.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ResizeBilinear.scala
@@ -403,19 +403,18 @@ object ResizeBilinear {
 
   @inline
   private def resizeImageBackprop(
-                                 batchSize: Int,
-                                 channels: Int,
-                                 inHeight: Int,
-                                 inWidth: Int,
-                                 outputHeight: Int,
-                                 outputWidth: Int,
-                                 heightScale: Float,
-                                 widthScale: Float,
-                                 gradInputData: Array[Float],
-                                 gradInputOffset: Int,
-                                 gradOutputData: Array[Float],
-                                 gradOutputOffset: Int
-                                 ): Unit = {
+    batchSize: Int,
+    channels: Int,
+    inHeight: Int,
+    inWidth: Int,
+    outputHeight: Int,
+    outputWidth: Int,
+    heightScale: Float,
+    widthScale: Float,
+    gradInputData: Array[Float],
+    gradInputOffset: Int,
+    gradOutputData: Array[Float],
+    gradOutputOffset: Int): Unit = {
     val inRowSize = inWidth * channels
     val inBatchNum = inHeight * inRowSize
     val outRowSize = outputWidth * channels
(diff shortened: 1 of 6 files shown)


yangw1234:
pr validation pass

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ResizeBilinear.scala
```diff
@@ -164,6 +260,19 @@ object ResizeBilinear {
...
   }
 
+  private def computeLERP(
+    topLeft: Double,
+    topRight: Double,
+    bottomLeft: Double,
+    bottomRight: Double,
+    xLERP: Double,
+    yLERP: Double
+                         ): Double = {
```
yiheng: indent

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ResizeBilinear.scala
```diff
@@ -32,18 +32,74 @@ import scala.reflect.ClassTag
...
+  val dataFormat: DataFormat = DataFormat.NCHW)(implicit ev: TensorNumeric[T])
+  extends AbstractModule[Tensor[T], Tensor[T], T]{
 
   private val ys = (1 to (outputHeight + 1)).map(i => InterpolationWeight(0, 0, 0)).toArray
   private val xs = (1 to (outputWidth + 1)).map(i => InterpolationWeight(0, 0, 0)).toArray
 
   import ResizeBilinear._
 
-  override def updateOutput(input: Tensor[Float]): Tensor[Float] = {
+  override def updateOutput(input: Tensor[T]): Tensor[T] = {
```
yiheng: why change to T? Is there anywhere use double image input?

spark/dl/src/test/scala/com/intel/analytics/bigdl/nn/ResizeBilinearSpec.scala
```diff
@@ -104,4 +105,29 @@ class ResizeBilinearSpec extends FlatSpec with Matchers {
     )))
     output should be(expectOutput)
   }
+
+  "ResizeBilinear forward and backward" should "be correct with NCHW" in {
```
yiheng: please add more unit tests

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ResizeBilinear.scala
yangw1234: Hi, @yiheng . I don't know the exact case where double image is used as input, but I thought it is make sense to support double, as other layers dealing with image like SpatialConvolution, SpatialMaxPooling, .etc all supports double.

yiheng: I think maybe no people want to use double tensor as image input. So we need not support this. If there's such require in the future, we can add then.

yiheng:
others looks good to me

spark/dl/src/test/scala/com/intel/analytics/bigdl/nn/ResizeBilinearSpec.scala
yangw1234: done

yangw1234:
pr validation pass

PR #2207, opened 2018-01-19
 support imageframe in python optimizer 

SeaOfOcean:

support imageframe in python optimizer 

## How was this patch tested?
integration test


pyspark/bigdl/optim/optimizer.py
```diff
@@ -732,8 +733,12 @@ def set_validation(self, batch_size, val_rdd, trigger, val_method=None):
         """
         if val_method is None:
             val_method = [Top1Accuracy()]
-        callBigDlFunc(self.bigdl_type, "setValidation", self.value, batch_size,
+        if isinstance(val_rdd, RDD):
```
zhichao-li: if else should only responsible for the method name instead of the entire callBigDLFunc

jenniew:
Please also update model.evaluate to support ImageFrame.

jenniew:
Get null pointer error when training, caused by no mat in ImageFeature. 
Add these lines to filesToImageFrame() in DataSet.scala
val mat = OpenCVMat.fromPixelsBytes(bytes, height, width)
        imf(ImageFeature.mat) = mat
Please check if it is right.

jenniew:
Can you add python support for reading image folder to ImageFrame, that Imageframe transformer can be applied to?

SeaOfOcean:
You should have transformer PixelBytesToMat() after filesToImageFrame(), which will convert bytes to mat

SeaOfOcean:
ImageFrame currently has python API to read a folder of images
see https://bigdl-project.github.io/master/#APIGuide/Data/#imageframe

SeaOfOcean:
@jenniew

SeaOfOcean:
pr validate pass

jenniew:
LGTM

SeaOfOcean:
jenkins pass http://172.168.2.101:8080/view/PR-V(new)/job/BigDL-PR-Validation-v3/219/

PR #2212, opened 2018-01-22
Refine doc script

yiheng:
1. The build doc script spends a lot of time on git clone latest doc repo. As the latest doc repo style files(css, html template) won't change frequently, this PR allow user to skip download bigdl doc repo to speed up build doc time
2. Automatically add a title to each document page. Users needn't add a doc title to each page. (remove the existed doc title on each md file)
3. Refine the doc style. Document content align left(refer other frameworks doc style) and use a different color as background, which makes it easier to read
4. hide the panel toggle button in large screen, just like the search button.

## How was this patch tested?
manual test


helenlly:
LGTM

PR #2214, opened 2018-01-22
Keras-like API Some layers and LeNet example

hkvision:

Layers supported:
Convolution2D
MaxPooling2D
Activation
Dropout
Flatten
Reshape

The LeNet example using the new API is compatible with our original LeNet example.

## How was this patch tested?

Jenkins


zhichao-li:
Depending on the progress of : https://github.com/intel-analytics/BigDL/pull/2211, We need to rebase and support serialization in our unittest.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialConvolution.scala
```diff
@@ -173,16 +173,40 @@ class SpatialConvolution[T: ClassTag](
...
+      s"output size is too small. outputWidth: $outputWidth, outputHeight: $outputHeight")
+    val outputShape = getOutputShape(outputHeight, outputWidth, input(0))
+    Shape(outputShape)
+  }
+
+  private def getOutputShape(oh: Int, ow: Int, batchSize: Int = -2): Array[Int] = {
     format match {
       case DataFormat.NCHW =>
-        if (batchSize == -1) {
+        if (batchSize == -2) {
```
zhichao-li: add doc here why changing this.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialDilatedConvolution.scala
```diff
@@ -147,6 +148,15 @@ class SpatialDilatedConvolution[T: ClassTag](
     }
   }
 
+  override def computeOutputShape(inputShape: Shape): Shape = {
+    val input = inputShape.toSingle().toArray
```
zhichao-li: add inputShape checking here.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialConvolution.scala
hkvision: ok =  =
-1 means no batch in this method, which conflicts with our inputshape

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialMaxPooling.scala
```diff
@@ -87,6 +87,39 @@ class SpatialMaxPooling[T: ClassTag](
...
+        require(kW / 2 >= padW && kH / 2 >= padH,
+          "pad should be smaller than half of kernel size. " +
+          s"current pad size($padW, $padH), " + s"kernel size($kW, $kH)")
+        Utils.getOutSizeAndPadding(inputHeight, inputWidth, dH, dW, kH, kW, padH, padW, ceilMode)
+      }
+    val oHeight = sizes(4)
+    val oWidth = sizes(5)
+
+    val outputShape = format match {
+      case DataFormat.NCHW =>
```
zhichao-li: remember to add unitest for both format

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialDilatedConvolution.scala
hkvision: Do we need checking? As calling `shapeCheck` need a tensor not shape

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TemporalMaxPooling.scala
```diff
@@ -101,6 +102,17 @@ class TemporalMaxPooling[T: ClassTag](
     }
   }
 
+  override def computeOutputShape(inputShape: Shape): Shape = {
```
zhichao-li: checking for inputshape

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialMaxPooling.scala
hkvision: Already there. https://github.com/intel-analytics/BigDL/pull/2214/files#diff-710faeb0d79adba6f8f90c251eff80bc

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Activation.scala
```diff
@@ -0,0 +1,43 @@
...
+import com.intel.analytics.bigdl.tensor.Tensor
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+import com.intel.analytics.bigdl.utils.Shape
+
+import scala.reflect.ClassTag
+
+class Activation[T: ClassTag](val activation: String,
+                              var inputShape: Shape = null)(implicit ev: TensorNumeric[T])
+  extends KerasLayer[Tensor[T], Tensor[T], T](KerasLayer.addBatch(inputShape)) {
+  require(activation != null)
```
zhichao-li: this checking can move to KerasUtils.getActivation(activation)

zhichao-li: add require message

hkvision: But `KerasUtils.getActivation(activation)` can return null in my implementation so that in factory methods, if `activation: String = null`, it will return null and still works.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Convolution2D.scala
```diff
@@ -0,0 +1,110 @@
...
+    inputShape: Shape = null)
+    (implicit ev: TensorNumeric[T]) : Convolution2D[T] = {
+    new Convolution2D[T](nbFilter, nbRow, nbCol, init, KerasUtils.getActivation(activation),
+      borderMode, subsample, wRegularizer, bRegularizer, format, bias, inputShape)
+  }
+}
+
+// Alias
+object Conv2D {
+  def apply[@specialized(Float, Double) T: ClassTag](
```
zhichao-li: reuse `Convolution2D ` here. no need to duplicate code.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasUtils.scala
```diff
@@ -0,0 +1,52 @@
...
+    if (borderMode == "same") {
+      // padH, padW
+      (-1, -1)
+    } else {
+      (0, 0)
+    }
+  }
+
+  private[keras] def getActivation[T : ClassTag] (activation: String)
+    (implicit ev: TensorNumeric[T]): TensorModule[T] = {
```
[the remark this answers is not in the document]
hkvision: I will change it~ previously forgot the case syntax

spark/dl/src/test/scala/com/intel/analytics/bigdl/keras/KerasRunner.scala
```diff
@@ -109,11 +109,18 @@ object KerasRunner {
                               keyName: String): Array[Tensor[Float]] = {
     if (!pvalues.keySet.filter(key => key.contains(keyName)).isEmpty) {
       val weightNum = pvalues.keySet.filter(key => key.contains(keyName)).size / 2
-      Range(0, weightNum).map {i =>
-        Tensor[Float](
-          data = pvalues(s"${keyName}_${i}_value"),
-          shape = pvalues(s"${keyName}_${i}_shape").map(_.toInt))
-      }.toArray
+      if (weightNum > 1) { // with bias
```
zhichao-li: why changing this?

hkvision: I discussed this in the previous PR into your branch.
If the layer has no bias,
data = pvalues(s"${keyName}_${i}_value") will get an error as there's only 1 weight, it will be just ${keyName} instead of ${keyName}_0
Changing this will take care of the case if there's no bias.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Reshape.scala
```diff
@@ -0,0 +1,91 @@
...
+import com.intel.analytics.bigdl.utils.Shape
+
+import scala.reflect.ClassTag
+
+class Reshape[T: ClassTag](
+   val targetShape: Array[Int],
+   var inputShape: Shape = null)(implicit ev: TensorNumeric[T])
+  extends KerasLayer[Tensor[T], Tensor[T], T](KerasLayer.addBatch(inputShape)) {
+
+  private var infer = false
```
hkvision: `Reshape` in Keras supports -1 to infer. https://faroit.github.io/keras-docs/1.2.2/layers/core/#reshape
The implementation here should be the same as Keras. 
Also update the mapping on Python side.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialConvolution.scala
hkvision: Updated.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Convolution2D.scala
hkvision: Not sure how to use alias... help?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Activation.scala
zhichao-li: or you can check it after getActivation. i.e input an empty string

spark/dl/src/test/scala/com/intel/analytics/bigdl/keras/KerasRunner.scala
zhichao-li: maybe we can change the logic that generating the `pvalues` to unify the naming.

hkvision: Yeah. This is only one option of change. We can discuss tomorrow.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Reshape.scala
```diff
@@ -0,0 +1,91 @@
...
+      infer = true
+      var i = 0
+      var count = 0
+      while (i < targetShape.length) {
+        if (targetShape(i) == -1) {
+          inferIndex = i
+          count += 1
+        }
+        // don't consider 0 here, same as Keras
+        else require(targetShape(i) >= 1, "wrong reshape size")
```
zhichao-li: add invalid index and value to the error message

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Reshape.scala
```diff
@@ -0,0 +1,91 @@
...
+
+  private var infer = false
+  private var inferIndex = -1
+  init()
+
+  private def init(): Unit = {
+    if (targetShape.contains(-1)) {
+      infer = true
+      var i = 0
+      var count = 0
```
zhichao-li: infer_count?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Reshape.scala
```diff
@@ -0,0 +1,91 @@
...
+class Reshape[T: ClassTag](
+   val targetShape: Array[Int],
+   var inputShape: Shape = null)(implicit ev: TensorNumeric[T])
+  extends KerasLayer[Tensor[T], Tensor[T], T](KerasLayer.addBatch(inputShape)) {
+
+  private var infer = false
+  private var inferIndex = -1
+  init()
+
+  private def init(): Unit = {
```
zhichao-li: enrichInferShape?

spark/dl/src/test/scala/com/intel/analytics/bigdl/keras/KerasRunner.scala
zhichao-li: val keyPrefix = if (weightNum > 1) {${keyName}_${i}} else {${keyName}}
key = keyPrefix + "_value"

zhichao-li:
You can use package object for the object alias: Conv2D etc....

hkvision:
We don't need SerialUID as said in https://github.com/intel-analytics/BigDL/issues/2224

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialDilatedConvolution.scala
hkvision: Updated.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TemporalMaxPooling.scala
hkvision: Updated

hkvision:
jenkins passed http://172.168.2.101:8080/job/BigDL-PR-V/850/
Need to add serialization test.

yiheng:
BTW, I think the missing serialization test error is not so friendly on Jenkins... @wzhongyuan

wzhongyuan:
@yiheng   The exception "layer ... not included in the test!" should be thrown out, i did see it in log, I will try to see if it is

zhichao-li:
Pls update the description for the added layers.

hkvision:
jenkins passed http://172.168.2.101:8080/job/BigDL-PR-Validation-v3/48/

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Reshape.scala
```diff
@@ -0,0 +1,92 @@
...
+        if (targetShape(i) == -1) {
+          inferIndex = i
+          inferCount += 1
+        }
+        // We don't consider 0 here, same as Keras
+        else require(targetShape(i) >= 1,
+          s"wrong reshape size at index $i: ${targetShape(i)}")
+        i += 1
+      }
+      require(inferCount == 1, "can only specify one unknown dimension")
```
zhichao-li: should we simplify the logic same as python?

pyspark/test/bigdl/keras/test_layer.py
```diff
@@ -309,6 +309,8 @@ def test_reshape(self):
         input_data = np.random.random_sample([1, 3, 5, 4])
         layer = Reshape(target_shape=(3, 20), input_shape=(3, 5, 4))
         self.modelTestSingleLayer(input_data, layer)
+        layer2 = Reshape(target_shape=(-1, 2, 3), input_shape=(3, 5, 4))
```
zhichao-li: what if there's multiple -1 ?

zhichao-li:
Please update the description. i.e: SpatialDilatedConvolution.scala

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialMaxPooling.scala
```diff
@@ -87,6 +87,39 @@ class SpatialMaxPooling[T: ClassTag](
     this
   }
 
+  override def computeOutputShape(inputShape: Shape): Shape = {
+    val input = inputShape.toSingle().toArray
+    require(input.length == 4,
+      "SpatialMaxPooling: " + ErrorInfo.constrainInputAs3DOrBatch)
```
zhichao-li: We should add the input length into the error message in respect of debugging. Please update the other require error message as well.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Activation.scala
```diff
@@ -0,0 +1,44 @@
...
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+import com.intel.analytics.bigdl.utils.Shape
+
+import scala.reflect.ClassTag
+
+class Activation[T: ClassTag](val activation: String,
+                              var inputShape: Shape = null)(implicit ev: TensorNumeric[T])
+  extends KerasLayer[Tensor[T], Tensor[T], T](KerasLayer.addBatch(inputShape)) {
+
+  require(activation != null)
```
zhichao-li: add error msg here.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasUtils.scala
```diff
@@ -0,0 +1,67 @@
...
+    if (activation == null) null
+    else {
+      activation.toLowerCase() match {
+          case "tanh" => Tanh[T]()
+          case "sigmoid" => Sigmoid[T]()
+          case "relu" => ReLU[T]()
+          // TODO: softmax 3D input is different from Keras
+          case "softmax" => SoftMax[T]()
```
zhichao-li: should we add a converter or throw exception here? otherwise user would get a surprise result here.

zhichao-li:
one more thing, add scala doc for those new added layers.

zhichao-li:
add test for the new added layers in `src/test/scala/com/intel/analytics/bigdl/utils/serializer/KerasModuleSerializerSpec.scala` otherwise jenkins validation would fail

pyspark/test/bigdl/keras/test_layer.py
hkvision: This is a Keras layer, and what we do on python side is under the assumption that we write correct code in Keras. If there are multiple -1, Keras will handle the expection.

spark/dl/src/main/scala/com/intel/analytics/bigdl/example/keras/Train.scala
```diff
@@ -0,0 +1,85 @@
+/*
+ * Copyright 2016 The BigDL Authors.
```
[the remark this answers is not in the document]
hkvision: Added.

hkvision:
Added serialization test. @zhichao-li

zhichao-li:
Add softmax  unitest for input.length == 2, then we can merge this once pass the jenkins.

hkvision:
Added ut for 2d softmax @zhichao-li

hkvision:
jenkins passed http://172.168.2.101:8080/job/BigDL-PR-Validation-v3/105/

PR #2215, opened 2018-01-23
Fix th test concurrent failure

yiheng:
In torch compare test, the name of the file which stores the torch compute result is not changed. It will cause a problem when multiple tests run in parallel on s single machine.

This change append a suffix to the torch result file name

## How was this patch tested?
unit test


glorysdj:
http://172.168.2.101:8080/job/BigDL-PR-V/820/ passed

please merge

PR #2217, opened 2018-01-23
add [[Log]] like operations

sperlingxx:

add Log/Log1p/LogN/LogN1p operations and logSpec

## How was this patch tested?

unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/Log.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/Log.scala
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2016 The BigDL Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.intel.analytics.bigdl.nn.ops
+
+import com.intel.analytics.bigdl.tensor.Tensor
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+
+import scala.reflect.ClassTag
+
+class Log[T: ClassTag, D: ClassTag](implicit ev: TensorNumeric[T], ev2: TensorNumeric[D])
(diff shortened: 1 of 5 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/Log1p.scala
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+package com.intel.analytics.bigdl.nn.ops
+
+import com.intel.analytics.bigdl.tensor.Tensor
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+
+import scala.reflect.ClassTag
+
+class Log1p[T: ClassTag, D: ClassTag](implicit ev: TensorNumeric[T], ev2: TensorNumeric[D])
```
yiheng: is this duplicate with **com.intel.analytics.bigdl.nn.tf.Log1p**?

It's a useful layer. Maybe we should move that class to nn package.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/Log.scala
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+package com.intel.analytics.bigdl.nn.ops
+
+import com.intel.analytics.bigdl.tensor.Tensor
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+
+import scala.reflect.ClassTag
+
+class Log[T: ClassTag, D: ClassTag](implicit ev: TensorNumeric[T], ev2: TensorNumeric[D])
```
yiheng: Is it duplicate with **com.intel.analytics.bigdl.nn.Log**? Maybe append a generic type D to that layer?

tosky001: the `nn.ops.Log` only contains **forward** function for Data Preprocessing. there is another ops **Exp** which are also defined  as `nn.ops.Exp` and `nn.Exp`. So, we have always had a confusion: what is  the specific distinguish between `nn.ops.*` and `nn.*`.  Moreover, how to deal with the common Components like **Log** and **Exp** etc.

yiheng: nn.ops.* cannot do backpropagation, so cannot be used to build the model(e.g. add to sequential and static graph).

Currently, we only use nn.ops.* when load and run tensorflow models. Use operation for data preprocessing is an interesting idea. But we may need to improve the logic in containers. We're open for discussion about this.

yiheng: If the logic of operation Log and layer Log are totally same, I think we can use the layer Log in the place of operation Log, right?

PR #2218, opened 2018-01-23
add backwardGraphPruning feature

sperlingxx:
Add backwardGraph pruning feature for [[StaticGraph]], which means remove all pre-processing nodes from [[backwardGraph]] whose element is an [[Operation]] or who is simply depended on [[Operation]] based nodes during backward.

I think this feature will be useful when building a graph with numerous operations for data pre-processing. Futhermore, it make possible to retreat modules with parameters as operations(such as  nn.Abs), as long as they are full depended on operation nodes during backward.

And we are working on data pre-processing operations, mainly for data-mining.  :)

## How was this patch tested?

unit test


sperlingxx:
@yiheng  Is this feature reasonable?

yiheng:
Hi @sperlingxx, yes I think this is a very useful feature. If I understand it correctly, I think there may be a simpler way to get what you want.

Please check the ```buildBackwardGraph``` function(https://github.com/intel-analytics/BigDL/blob/master/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala#L203)

Generally speaking, when we build the back graph, we will
1. clone the forward graph
2. modify the graph if user set some nodes as stop gradient
3. collect backward target

We use this piece of code to collect backward target
```scala
val backwardTargets = backwardNodes.filter(n => (n.element.parameters() != null && n.element.parameters()._1.length != 0) || inputNames.contains(n.element.getName()))
```

As you see, we will use input nodes(to generate gradInput of the model) and nodes with parameters(to generate the parameterGradient of the model) as target.

So I think maybe we can set a flag, if it's true, the graph won't use input nodes as backward target. So in the backward, the preprocessing nodes won't be invoked.  What do you think :-)

sperlingxx:
Hi @yiheng , thank you for replying me so patient.

First of all, I have a bit confused about whether `    val backwardTargets = backwardNodes
      .filter(n => (n.element.parameters() != null && n.element.parameters()._1.length != 0)
` this piece of code will remove all nodes without parameters.  Since, topologySort will do a depth-first search to find out all nodes related to the graph.  Is there any possibility for appending non-para nodes back(which have already been filtered before, but they may still have a linkage with other nodes in backwardTargets)?

Secondly, backGraphPruning feature will remove nodes with parameters, when all their prev_nodes are non-para nodes. For instance,  at Line 1281 of GraphSpec(https://github.com/intel-analytics/BigDL/pull/2218/files#diff-f186ad8eccbbc92e5ab957adf1ecf58b),  ll is a [Linear], but ll will still be removed by pruning because the only prev node of it is a [Round] which parameters() return null.

My propose of backGraphPruning is set users free from doing stopGradient manually :)

yiheng:
@sperlingxx 

For your first question, in backpropagation, we only compute two kinds of gradients:
1. gradients of the inputs
2. gradients of the weight/bias

So we only remove non-para nodes which are not related to the above gradients. For example, in
```
Input -> ReLU1 -> Join -> ReLU2 -> Output
Const1 -> Linear ------|
Const2 -------------------|
```
we won't compute the gradients of Const1/Const2, as they're not related to input gradients and Linear(para-node) gradients. But we still compute gradients of ReLU1 and ReLU2(non-para), as they're the parent of the target gradients.

So
>Although they are not included in backwardTargets, they still have a linkage with other nodes in backwardTargets.

We should include these nodes in the gradient computing.

---

I understand what do you want is when detecting operation in the graph, auto stop gradients backpropagation. I think it's a useful feature. My concern is

* From implementation, do we need to impl the invToplogy? Another solution is we can add all operations of the graph to the stopGradientLayers list when building the graph. It seems to meet what you want.

* Do we need to add the backGraphPruning option? I think less option, more friendly to the user. If the graph contains operations, it should stop gradients on the operation, or we should throw an exception, right?

* The last concern is the mixed use of layer and operation. Currently, the operations are only used in load tensorflow models. You have a new requirement, use operation in preprocessing. It's an interesting idea, but beyond current scope for operation. I'd like to discuss this with other members.

@jason-dai @hhbyyh @SeaOfOcean @yangw1234 What do you think we use operation for data preprocessing and mix operation and layer when build the model.

sperlingxx:
Hi @yiheng , thanks for concerning about the issue: 

>  use operation for data preprocessing and mix operation and layer when build the model

To your first concern, I think there exists a little bit difference between pruning nodes and gradient stopping.  

According to _Line 142 StaticGraph_ `curNode.element.accGradParameters(curInput, curGradOutput) `, 
_accGradParameters_ method will be called even if _isStopGradient_ equals _true_. While _accGradParameters_ is unsupported in _Operation_  by default.

yiheng:
@sperlingxx Yes. It's kind of stop gradient without accGradParameters :nerd_face:

yangw1234:
I think the `backwardGraphPruning` option should be able to be changed after the graph's construction.

When a graph is created, we should deactivate this feature in all the other graphs serving as its nodes, because we need their `gradInput`s for back propagation.

In implementation, we could keep the whole backward topology and only mark the node being pruned. When executing the backward graph, we just skip those pruned nodes.

yangw1234:
Or we could provide anther `backward` that only compute the gradients w.r.t its parameters, so when a layer is served as a intermediate computation node, we used the original `backward` to compute the gradients w.r.t its parameters as well as all its inputs; when it is served as a model, we use the second `backward` to only compute the gradient w.r.t its parameters. And we implement the backwardGraphPruning feature in the second `backward`

yiheng:
@sperlingxx For data preprocessing, there're other solutions. One is use Spark MLlib pipeline https://spark.apache.org/docs/2.2.0/ml-pipeline.html. Another is our own transformer https://bigdl-project.github.io/master/#APIGuide/Transformer/.

Is there some gap between these solutions and your requirement?

sperlingxx:
@yiheng We have considered about __pipeline-style__ API, and we have tried  _Transformer_ API of BigDL. After some trails, we found that __pipeline__ is less convenient than __DAG__. 

And then we have tried to build  DAGOperators/DAGScheduler on the basis of __Transformer__. It's far more friendly than __pipeline__. I can show a small sample here:
      
    val rows: Iterator[Row] = (1 to 10).map(_ => randRow(3, 20, 3)).toIterator
    val denseFea1 = ToDenseTensor[Float](ColumnSelector[Float]((3 to 12).toArray), 2)
    val denseFea2 = TensorOp[Float](Tensor[Float](randFloats(5), Array(1, 5)))
    val sparseFea1 = KVToTensor[Float]("sparse1", 100)
    val sparseFea2 = KVToTensor[Float]("sparse2", 100)
    val lbl = ToDenseTensor[Float](ColumnSelector[Float](13))
    val feature = DAGResult(dim = 2) ++=
      (Seq(denseFea1, denseFea2, sparseFea1, sparseFea2): _*)
    val label = DAGResult(dim = 2) += lbl
    val r2s: Transformer[Row, Sample[Float]] = RowToSample[Float](feature, label)

We build a tiny __DAG__ in a __Transformer__ named __RowToSample__.

After that, we found data pre-processing procedure is still isolated from main model, which means users need to define it separately, apart from main model. 
But, in my opinion, in most of the time, users just want a single DAG to define the whole procedure rather than distinguishing data pre-processing steps from main model. 
For instance, _hash_embedding_ operation contains a pre-processing action(hashing a string to an index) and a network layer(LookupTable). In _Tensorflow_, users can use __tf.feature_column__  API to do jobs like this easily.  

After reading codes about __Graph__ and __Operation__,  I think maybe we can provide a similar way to define the procedure like _Tensorflow_ by them? Maybe I can add  an __Activity__ wrapper for __spark.sql.Row__ as a start point of the whole procedure(graph)?

sperlingxx:
hi @yiheng,

can we add a _RowActivity_ which extends _Activity_ to wrap _Row_ with batch? 
Correspondingly, _RowSample_ and _RowMiniBatch_ are necessary for slicing _Rows_ into batches.

After that, we can use _Graph_ to build a end to end procedure.

yiheng:
@yangw1234 I don't think we need that complexity. Maybe just throw a friendly error message when user add a Graph(without gradInput) into another container

yiheng:
@sperlingxx Is it possible to convert Row to a Tensor at the first step, then we still implement the preprocess algorithms on Tensor?

sperlingxx:
@yiheng 
Yes, it is a good idea. Maybe we can create a Transformer[Row, Table(key, Tensor)] to do this job ?

yiheng:
@sperlingxx I work out a PR based on your change. Please check https://github.com/intel-analytics/BigDL/pull/2241. Do you think it meets your requirement?

sperlingxx:
@yiheng 
Yes, tks :-)   I will close this PR.

PR #2227, opened 2018-01-24
Fix a random failure in tf fused batch norm test

yiheng:
Fix a random failure in tf fused batch norm test by setting the random seed before generating input data. The failure is due to a high accuracy comparison.

## How was this patch tested?
unit test


glorysdj:
LGTM, please merge

PR #2232, opened 2018-01-25
Enable serialization for keras layer

zhichao-li:

Serialization for new added Keras layer is disable temporary, and this PR try to fix and enable that.
## How was this patch tested?
unittest

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
@@ -18,26 +18,25 @@ package com.intel.analytics.bigdl.nn
 import java.util
 
 import com.intel.analytics.bigdl.Module
-
-import scala.collection.JavaConverters._
 import com.intel.analytics.bigdl.nn.Graph.ModuleNode
-import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, Activity, TensorModule}
+import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, Activity}
 import com.intel.analytics.bigdl.nn.ops._
 import com.intel.analytics.bigdl.nn.tf.{ControlDependency, WithoutInput}
+import com.intel.analytics.bigdl.serialization.Bigdl.{AttrValue, BigDLModule}
 import com.intel.analytics.bigdl.tensor.Tensor
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
-import com.intel.analytics.bigdl.utils.serializer._
 import com.intel.analytics.bigdl.utils._
+import com.intel.analytics.bigdl.utils.serializer._
 import com.intel.analytics.bigdl.utils.serializer.converters.DataConverter
 import com.intel.analytics.bigdl.utils.tf.Tensorflow
-import com.intel.analytics.bigdl.serialization.Bigdl.{AttrValue, BigDLModule}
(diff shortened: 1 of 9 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -452,85 +451,10 @@ abstract class Graph[T: ClassTag](
...
-   * Build multiple inputs, multiple outputs graph container.
-   * @param input input node
-   * @param output output node
-   * @return a graph container
-   */
-  def apply[T: ClassTag](
-    input : Array[ModuleNode[T]],
-    output : Array[ModuleNode[T]],
-    variables: Option[(Array[Tensor[T]], Array[Tensor[T]])] = None
-  )(implicit ev: TensorNumeric[T]) : Graph[T] = {
```
zhichao-li: @yiheng  no functional changes. just extract a method to make it extendable by the subclass.

hkvision:
Have you enabled the serialization checking?

zhichao-li:
@hkvision  pls take a look at `KerasModuleSerializerSpec.scala`

zhichao-li:
cc @yiheng @wzhongyuan  any comments?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -72,8 +71,8 @@ import org.tensorflow.framework.GraphDef
 @SerialVersionUID(- 2896121321564992779L)
 abstract class Graph[T: ClassTag](
   val inputs : Seq[ModuleNode[T]],
-  protected val outputs : Seq[ModuleNode[T]],
-  private val variables: Option[(Array[Tensor[T]], Array[Tensor[T]])] = None
+  private[bigdl] val outputs : Seq[ModuleNode[T]],
+  private[bigdl] val variables: Option[(Array[Tensor[T]], Array[Tensor[T]])] = None
```
yiheng: why change the scope of the variables and outputs?

spark/dl/src/test/scala/com/intel/analytics/bigdl/utils/serializer/SerializerSpecHelper.scala
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+
+  override protected def afterAll() = {
+    var total = 0
+    getExpected().foreach(exp => {
+      require(tested.contains(exp), s" $exp not included in the test!")
+      total += 1
+    })
+    println(s"total $total, remaining ${getExpected().size - total}")
```
wzhongyuan: should put this line before the foreach cluase otherwise it will never get executed if test for some layers are not added

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
zhichao-li: `variables` or `variables` cannot be access by  `GraphSerializable` from the original scope as it's extracted from object `Graph`. Not sure if there's any other good solution for this, or I guess we can change the scope if no specific concern.

spark/dl/src/test/scala/com/intel/analytics/bigdl/utils/serializer/SerializerSpecHelper.scala
zhichao-li: ok

wzhongyuan:
LGTM

zhichao-li:
jenkins passed.

PR #2234, opened 2018-01-25
add [[Kv2Tensor]] Operation

tosky001:

add Kv2Tensor Ops which support convert kvString to Sparse/Dense Tensor

## How was this patch tested?

unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/Kv2Tensor.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/Kv2Tensor.scala
@@ -61,7 +61,8 @@ class Kv2Tensor[T: ClassTag, D: ClassTag](
     val rows = kvTensor.size(dim = 1)
     val shape = Array(rows, feaLen)
 
-    for(i <- 1 to rows) {
+    var i = 1
+    while(i<=rows) {
       val kvFeaString = kvTensor.select(1, i).valueAt(1)
       kvFeaString.split(kvDelimiter).foreach { kv =>
         indices0 += i-1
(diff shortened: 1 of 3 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/Kv2Tensor.scala
```diff
@@ -0,0 +1,107 @@
...
+  override def updateOutput(input: Table): Tensor[D] = {
+    val kvTensor = input[Tensor[String]](1)
+    val feaLen = input[Tensor[Int]](2).value()
+    val indices0 = new ArrayBuffer[Int]()
+    val indices1 = new ArrayBuffer[Int]()
+    val values = new ArrayBuffer[D]()
+    val rows = kvTensor.size(dim = 1)
+    val shape = Array(rows, feaLen)
+
+    for(i <- 1 to rows) {
```
yiheng: As this is executed in model forward, try to use while instead of for. As for in scala is not so efficient as while, which is different from Java.

tosky001: Thanks for your advice and I have replaced `for` with `while` and run the unit tests successfully.

yiheng:
run Jenkins test and find that the serialization test of the new operation is missed. Can you add a unit test in the https://github.com/intel-analytics/BigDL/blob/master/spark/dl/src/test/scala/com/intel/analytics/bigdl/utils/serializer/ModuleSerializerSpec.scala file. Just like other model serialization test in that class.

tosky001:
ok，i will do it as soon as possible. and Thanks for your advice!

tosky001:
the Serialization-test has finished and run successfully.

yiheng:
PR validation pass

PR #2241, opened 2018-01-29
Combine pre-processor and trainable model into one model

yiheng:
This PR provide a new method to create a Graph which includes data pre-processing operations.

This PR includes following change
1. Operation will throw an exception when invoking its accGradParameter method. However, other layers which don't have parameter won't have this behavior. This PR will make the operation accGradParameter behavior consistent with other layers.

2. Don't allow use operation in Dynamic Containers(e.g. Sequential, Concat)

3. Dynamic Graph should be an internal class and creating dynamic graph become an internal API.

4. If a graph model doesn't generate gradInput, an EmptyGradInput class will be return if you try to get the gradInput, which will throw error when you want to use it.

5. A new API to create Graph which combines preprocessor and trainable model

## How was this patch tested?
unit test


yiheng:
Hi @sperlingxx 
I have some discussion about this feature with other team members. A concern is it maybe make user confused if we auto remove the operation from the backward graph if user is not familiar with this.

To help user conveniently handle the data preprocessing operations/layers, we suggest that we can provide a new graph construct API, something like
```scala
val preprocessGraph = Graph(...)
val modelGraph = Graph(...)
val model = Graph(preprocessGraph, modelGraph)
```
Does this look good to you?

sperlingxx:
Hi @yiheng , 

> val model = Graph(preprocessGraph, modelGraph)

Does this mean regarding all outputs of  _preprocessGraph_ as inputs of _modelGraph_?  So, the nodes in _preprocessGraph_ just work in _forward_.

yiheng:
@sperlingxx Yes

jason-dai:
Please update the discription

yiheng:
@sperlingxx can you help review this PR?

sperlingxx:
@yiheng Okay, I am pleasure to do this job.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/DynamicContainer.scala
```diff
@@ -37,6 +38,9 @@ abstract class DynamicContainer[A <: Activity : ClassTag, B <: Activity : ClassT
    * @return this container
    */
   def add(module: AbstractModule[_ <: Activity, _ <: Activity, T]): this.type = {
+    require(!module.isInstanceOf[Operation[_, _, _]],
+      "Add operations to dynamic container is not allowed, as operations don't have backward. " +
+        "Operation is only allowed used in Graph")
```
jason-dai: Operation is can only be used in Graph

yiheng:
jenkins test passed

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -212,6 +216,17 @@ abstract class Graph[T: ClassTag](
...
     backwardGraph = dummyBackwardEnd.graph(true)
+
+    // Check if gradInput is empty for each input
+    isGradInputAvailable = inputs.map(_ => false).toArray
+    inputs.zipWithIndex.map{ case (n, i) =>
+      backwardGraph.DFS.foreach(curNode => {
+        if (curNode.element.getName() == n.element.getName() && !isStopGradient(n.element)) {
+          isGradInputAvailable(i) = true
+        }
+      })
```
yangw1234: Maybe it does not matter that much, but I think we are able to do this in one DFS pass.

yiheng:
PR validation pass

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Graph.scala
```diff
@@ -212,6 +216,17 @@ abstract class Graph[T: ClassTag](
         || inputNames.contains(n.element.getName()))
     backwardTargets.foreach(_ -> dummyBackwardEnd)
     backwardGraph = dummyBackwardEnd.graph(true)
+
+    // Check if gradInput is empty for each input
+    isGradInputAvailable = inputs.map(_ => false).toArray
+    backwardGraph.DFS.foreach(curNode => {
+      inputs.zipWithIndex.map { case (n, i) =>
+        if (curNode.element.getName() == n.element.getName() && !isStopGradient(n.element)) {
+          isGradInputAvailable(i) = true
```
jason-dai: This looks expensive. In addition, one can stop gradient in the middle (instead of on the input node) to make the gradInput of the input node empty, no?

yiheng: It happens in graph build time.

If one stop gradient in the middle, which makes gradient cannot be back propagated to some inputs, these inputs won be in the backwardGraph. So the gradInput of such inputs are empty.

yiheng:
jenkins pass

PR #2247, opened 2018-01-30
Keras-like API AtrousConv1D2D, LocallyConnected1D, deconv, convlstm

hkvision:

AtrousConvolution1D
AtrousConvolution2D
Deconvolution2D
ConvLSTM2D
LocallyConnected1d

## How was this patch tested?

Jenkins


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/AtrousConvolution2D.scala
```diff
@@ -0,0 +1,77 @@
...
+import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, TensorModule}
+import com.intel.analytics.bigdl.nn.{InitializationMethod, SpatialDilatedConvolution, Xavier, Zeros}
+import com.intel.analytics.bigdl.optim.Regularizer
+import com.intel.analytics.bigdl.tensor.Tensor
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+import com.intel.analytics.bigdl.utils.Shape
+
+import scala.reflect.ClassTag
+
+class AtrousConvolution2D[T: ClassTag](
```
zhichao-li: What's the expected order for this class? NCHW? or both?

hkvision: Only NCHW is supported. Will write this in the documentation.

zhichao-li: is there any parameters like dim_order here? or we also need to add this into the scala doc.

zhichao-li:
If the layer only support NCHW or NHWC, we should explicitly add scala doc to clarify it.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/AtrousConvolution2D.scala
hkvision: Since we only support one format, it is strange to have a parameter `format`. I will add it into scala doc.

zhichao-li: Just check with Keras, there should be a `dim_ordering` in the parameter list. let's add it and throw exception if `channel-first` doesn't support for now.

hkvision: I think on scala side, we don't need such an option. It's really awkward that we provide an argument but it has to be only one value but no other choices. On python side, we can provide an argument just as Keras.

hkvision: Added dimOrdering

zhichao-li:
LGTM

hkvision:
jenkins passed http://172.168.2.101:8080/view/PR-V(new)-Pipeline/job/BigDL-PR-Validation-v3/170/

PR #2251, opened 2018-01-31
fix getParameters() in submodule will create new storage when parameters has been shared by parent module.

qiuxin2012:

fix getParameters() in submodule will create new storage when parameters has been shared by parent module.

## How was this patch tested?

unit tests


qiuxin2012:
jenkins passed

PR #2242, opened 2018-01-29
Add python api for keras layer

zhichao-li:
## What changes were proposed in this pull request?

1.Add python api for keras layer.
2.Support extra JavaCreator via `add_creator_class`.
## How was this patch tested?
unittest

PR #2225, opened 2018-01-24
Fix python ut with wrong format and precision

hkvision:
## What changes were proposed in this pull request?

Use np.testing.assert_allclose instead of docstring.

## How was this patch tested?

Jenkins

## Related links or issues (optional)
fixed https://github.com/intel-analytics/BigDL/issues/XXX

PR #2252, opened 2018-01-31
Fix incompatiable syntax

zhichao-li:
Fix the syntax error that doesn't compatible with python2.7 and python3.6
Related PRs:#2242 #2225
## How was this patch tested?
unittest


zhichao-li:
http://172.168.2.101:8080/job/BigDL-PR-Validation-v3/62/

PR #2257, opened 2018-01-31
fix: initializing of weights for SReLU should not in forward.

i8run:
Otherwise, the Keras will not convert successfully.

## How was this patch tested?

### DIFF
--- a/pyspark/bigdl/keras/converter.py
+++ b/pyspark/bigdl/keras/converter.py
@@ -1600,16 +1600,16 @@ class LayerConverter:
         if "shared_axes" not in self.config:
             warnings.warn("Cannot find shared_axes from json definition. Using shared_axes=None instead.")
         shape = self.input_shape[1:]
-        t_left_init = to_bigdl_init(config["t_left_init"])
-        a_left_init = to_bigdl_init(config["a_left_init"])
-        t_right_init = to_bigdl_init(config["t_right_init"])
-        a_right_init = to_bigdl_init(config["a_right_init"])
+        t_left_init = to_bigdl_init(self.config["t_left_init"])
+        a_left_init = to_bigdl_init(self.config["a_left_init"])
+        t_right_init = to_bigdl_init(self.config["t_right_init"])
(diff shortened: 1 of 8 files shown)


pyspark/bigdl/nn/layer.py
```diff
@@ -3690,22 +3688,29 @@ class SReLU(Layer):
...
             set `shared_axes=[1, 2]`.
 
-    >>> srelu = SReLU()
+    >>> srelu = SReLU((2, 3))
     creating: createSReLU
-    >>> srelu = SReLU((1, 2))
+    >>> srelu = SReLU((2, 2), (1, 2))
     creating: createSReLU
+    >>> from bigdl.nn.initialization_method import Xavier
+    >>> tt = srelu.set_init_method(tLeftInit=Xavier(), aLeftInit=Xavier(), tRightInit=Xavier(), aRightInit=Xavier())
```
hkvision: Replace `tt` = =

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SReLU.scala
```diff
@@ -35,6 +35,7 @@ import scala.reflect.ClassTag
  *
  * [Deep Learning with S-shaped Rectified Linear Activation Units](http://arxiv.org/abs/1512.07030)
  *
+ * @param shape shape for tleft, aleft, tright, aright, for a 4-D input, the shape is the last 3-D
```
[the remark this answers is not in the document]
i8run: 4-D is an example.

pyspark/bigdl/keras/converter.py
```diff
@@ -1599,10 +1599,18 @@ def create_masking(self):
     def create_srelu(self):
         if "shared_axes" not in self.config:
             warnings.warn("Cannot find shared_axes from json definition. Using shared_axes=None instead.")
+        shape = self.input_shape[1:]
+        t_left_init = to_bigdl_init(config["t_left_init"])
```
hkvision: jenkins failed... `self.config`

PR #2260, opened 2018-02-01
Add [[CategoricalColHashBucket]] Operation

tosky001:

the Operation is aimed to distribute your inputs into a finite number of buckets by hashing.

## How was this patch tested?

unit test


tosky001:
@yiheng  May i have your review please and thank you very much.

yiheng:
sure. Can we split the operations into different PRs? It's easier to review and we can merge them as early as possible

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/CategoricalColHashBucket.scala
```diff
@@ -0,0 +1,110 @@
...
+import com.intel.analytics.bigdl.utils.{T, Table}
+
+import scala.collection.mutable.ArrayBuffer
+import scala.reflect.ClassTag
+import scala.util.hashing.MurmurHash3
+
+/**
+ * CategoricalColHashBucket operation can convert feature string to a Sparse/Dense Tensor
+ *
+ * DenseTensor if transType = 0
```
yiheng: Maybe use enum classes or isSparse instead of the number can make code easier to read?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/CategoricalColHashBucket.scala
```diff
@@ -0,0 +1,110 @@
...
+      max_fea_len = math.max(max_fea_len, feaStrArr.length)
+      var j = 0
+      while(j < feaStrArr.length) {
+        val hashVal = MurmurHash3.stringHash(feaStrArr(j)) % hashBucketSize match {
+          case v if v < 0 => v + hashBucketSize
+          case v => v
+        }
+        indices0 += i-1
+        indices1 += j
+        ev.getType() match {
```
yiheng: ev.fromType()?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/CategoricalColHashBucket.scala
```diff
@@ -0,0 +1,110 @@
...
+class CategoricalColHashBucket[T: ClassTag](
+  val hashBucketSize: Int,
+  val strDelimiter: String = ",",
+  val transType: Int = 1
+  )(implicit ev: TensorNumeric[T])
+  extends Operation[Table, Tensor[T], T] {
+
+  output = Activity.allocate[Tensor[T], T]()
+
+  override def updateOutput(input: Table): Tensor[T] = {
```
yiheng: what about set the input type to Tensor[String]? For Integer, may be we can add a cast operation before the operation

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/CategoricalColHashBucket.scala
```diff
@@ -0,0 +1,110 @@
...
+    val column = input[Tensor[_]](1)
+    val rows = column.size(dim = 1)
+    val indices0 = new ArrayBuffer[Int]()
+    val indices1 = new ArrayBuffer[Int]()
+    val values = new ArrayBuffer[T]()
+    var i = 1
+    var max_fea_len = 0
+    while(i <= rows) {
+      val feaStrArr = column.select(1, i).valueAt(1).toString.split(strDelimiter)
```
yiheng: why not column.valueAt(1, i)?

tosky001:
ok, no problem, i will close this PR and create a new one.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/CategoricalColHashBucket.scala
```diff
@@ -0,0 +1,110 @@
...
+          case FloatType =>
+            values += hashVal.toFloat.asInstanceOf[T]
+        }
+        j += 1
+      }
+      i += 1
+    }
+    val indices = Array(indices0.toArray, indices1.toArray)
+    val shape = Array(rows, max_fea_len)
+    output = transType match {
```
yiheng: what about set the output type to Tensor[Int]? Maybe we can cast it after the operation.

yiheng:
You may rename or edit the PR

tosky001:
@yiheng the Operation and PR description has been re-edit, and Thanks for your very meaningful advice.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/CategoricalColHashBucket.scala
```diff
@@ -0,0 +1,102 @@
...
+          case v => v
+        }
+        indices0 += i-1
+        indices1 += j
+        values += hashVal
+        j += 1
+      }
+      i += 1
+    }
+    val indices = Array(indices0.toArray, indices1.toArray)
```
yiheng: Looks like the output of the operation is dense, right? Why we put it in a sparse tensor here?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/CategoricalColHashBucket.scala
```diff
@@ -0,0 +1,102 @@
...
+
+  override def updateOutput(input: Tensor[String]): Tensor[Int] = {
+    val rows = input.size(dim = 1)
+    val indices0 = new ArrayBuffer[Int]()
+    val indices1 = new ArrayBuffer[Int]()
+    val values = new ArrayBuffer[Int]()
+    var i = 1
+    var max_fea_len = 0
+    while(i <= rows) {
+      val feaStrArr = input.valueAt(i, 1).toString.split(strDelimiter)
```
yiheng: toString is not needed, right?

yiheng:
Can you resolve the conflict? So I can run a jenkins test on your PR :smile:

tosky001:
@yiheng  the confict is due to the changes of  `ModuleSerializerSpec.scala`. the serializerTest of nn.ops.*  has been moved to `TFSerializerSpec.scala`. So, i want to confirm where all the follow-up operations should be tested in the new file `TFSerializerSpec.scala` instead of origin file `ModuleSerializerSpec.scala`.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/CategoricalColHashBucket.scala
tosky001: the output of the operation is better sparse. because feature column may have multi-value, and  the dense will be a waste of memory. There, i offer an option of dense just because some methods in sparseTensor have not been implemented.

yiheng: I see

yiheng:
@wzhongyuan can you help answer the questions?

I think we should have a OperationSerializerSpec to hold the operation serializer test, right?

wzhongyuan:
@tosky001   @yiheng   Makes sense to rename it , it should work,

tosky001:
@wzhongyuan So the follow-up operations as well as the accepted operation **Kv2Tensor** should be tested in a new file `OperationSerializerSpec.scala`?

wzhongyuan:
@tosky001  yes, just rename toe TFSerializerSepec to OperationSerializerSpec

wzhongyuan:
@yiheng   seems I put bigdl.tf package into the same class, since we rename it to OperationSerializerSpec, I will split tf package into a new one, but we can leave it for now

yiheng:
jenkins test pass. merge

Issue #966, opened 2017-06-05
Enrich function doc for getParameters()


zhichao-li:
1) Add ```final``` to getParameters() to prevent overriding in subclass
2) We might need to add a checking to ensure not invoking this method multiple times but alert that in comments maybe enough as it would slightly effect the performance.
2) Enrich the comments with:

This function returns two tensors. One for the flattened learnable parameters flatParameters and another for the gradients of the energy wrt to the learnable parameters flatGradParameters.

__Custom modules should not override this function.__  They should instead override parameters(...) which is, in turn, called by the present function.

This function will go over all the weights and gradWeights and make them view into a single tensor (one for weights and one for gradWeights). Since the storage of every weight and gradWeight is changed, this function __should be called only once on a given network.__

yiheng:
+1

PR #2262, opened 2018-02-01
Refine AbstractModule methods

yiheng:
This PR is code refactor, include:
1. make getParamter method as final and private[bigdl]. That method should only used in BigDL optimization. If user accident call it outside, the optimization result maybe incorrect
2. Remove updateParameters method. It's only used for debug, should not be a public API. User should use optimizer instead of updateParameter method to change the model weights
3. ZeroGrad should be final, remove the redundant implementations
4. add final to necessary methods
5. reorder the methods, put public method first

## How was this patch tested?
unit test

fixed https://github.com/intel-analytics/BigDL/issues/966

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
@@ -341,8 +341,15 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag, 
   }
 
   /**
-   * This method compact all parameters and gradients of the model into two tensors. So it's easier
-   * to use optim method
+   *This function returns two tensors. One for the flattened learnable parameters flatParameters
+   * and another for the gradients of the energy wrt to the learnable parameters flatGradParameters.
+   *
+   * Custom modules should not override this function. They should instead override parameters(...)
+   * which is, in turn, called by the present function.
+   *
+   * This function will go over all the weights and gradWeights and make them view into a single
+   * tensor (one for weights and one for gradWeights). Since the storage of every weight and
+   * gradWeight is changed, this function should be called only once on a given network.
    *
    * @return
    */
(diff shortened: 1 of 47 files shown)


zhichao-li:
One idea is we can provide a common method for `getGrad`, then  `zeroGrad` doent's need to be override in every place as it just zero grads....

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -340,16 +340,26 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
...
-   * to use optim method
+   * This function returns two tensors. One for the flattened trainable parameters flatParameters
+   * and another for the gradients of the energy wrt to the trainable parameters flatGradParameters.
+   *
+   * Custom modules should not override this function. They should instead override parameters(...)
+   * which is, in turn, called by the present function.
+   *
+   * This function will go over all the weights and gradWeights and make them view into a single
+   * tensor (one for weights and one for gradWeights). Since the storage of every weight and
+   * gradWeight is changed.
```
yangw1234: Feels like this sentence is not finished.

yiheng:
@zhichao-li This PR is removing the unnecessary zeroGrad methods. ZeroGrad is using parameter() method to get the grads. We needn't an extra getGrad methods

yiheng:
jenkins test pass

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -849,30 +777,206 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
...
+  final def evaluate(
+    dataSet: LocalDataSet[MiniBatch[T]],
+    vMethods: Array[ValidationMethod[T]]
+  ): Array[(ValidationResult, ValidationMethod[T])] = {
     Validator(this, dataSet).test(vMethods)
   }
 
-  def quantize(): Module[T] = {
+  /**
+   * Quantize this module, whcih reduces the precison of the parameter. Get a higher speed with a
```
yangw1234: which and precision are spelled wrong.

yangw1234:
LGTM

yiheng:
jenkins pass.merge

Issue #1333, opened 2017-07-21
Training Resnet not support HDFS file dir


ZeweiChen11:
`spark-submit --class com.intel.analytics.bigdl.models.resnet.Train --master spark://hsx-node1:7077 --driver-memory=30g --executor-cores=28 --num-executors=8 --total-executor-cores=224 --executor-memory=220g --driver-java-options="-Dlog4j.configuration=/opt/package/BigDL_713_2/log4j.conf" /opt/package/BigDL_713_2/dist/lib/bigdl-0.2.0-SNAPSHOT-jar-with-dependencies.jar --batchSize 896 -f hdfs:///user/root/cifar-10 --optnet true --depth 20 --classes 10 --shortcutType A --nEpochs 1 --learningRate 0.1`
It reports:
`Exception in thread "main" java.nio.file.NoSuchFileException: hdfs:/user/root/cifar-10/data_batch_1.bin `

It can train after changing to cifar-10 local dir.

helenlly:
@JerryYanWan  pls have a look

yiheng:
cifar10 is too small. we need not put it on hdfs...

ibingoogle:
straightforward solution: 
First, replace three functions loadTrain loadTest load in resnet/Utils.scala with counterparts in vgg/Utils.scala
Then, recompile bigdl
Works for both local file and hdfs

yiheng:
@ibingoogle thanks for providing the solution. I will look at your PR

PR #2266, opened 2018-02-01
support hdfs-cifar10 on resnet

ibingoogle:
1. changes three functions (loadTrain loadTest load) in com.intel.analytics.bigdl.models.resnet/Util.scala so that we can load cifar10 from hdfs when training resnet.
2. I changed the source code according to com.intel.analytics.bigdl.models.vgg/Util.scala since the vgg/Util.scala supports hdfs-cifar10 loading.

## How was this patch tested?
Runing Spark local, Spark standalone and Spark yarn client models with both local dir and hdfs dir

fixed https://github.com/intel-analytics/BigDL/issues/1333
BTW: yiheng closed this issue because of the size of cifar10. But I think we can support this in resnet/Util.scala just like vgg/Util.scala

Thanks

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/resnet/Utils.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/resnet/Utils.scala
@@ -20,6 +20,7 @@ import java.nio.ByteBuffer
 import java.nio.file.{Files, Path, Paths}
 
 import com.intel.analytics.bigdl.dataset.ByteRecord
+import com.intel.analytics.bigdl.utils.File
 import scopt.OptionParser
 
 import scala.collection.mutable.ArrayBuffer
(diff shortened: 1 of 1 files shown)


yiheng:
manual test looks good. Run jenkins

yiheng:
There's a style error
```
error file=/opt/work/jenkins/workspace/BigDL-PRVN-StyleCheck/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/resnet/Utils.scala message=Use Javadoc style indentation for multiline comments line=141 column=0
```

Please use java doc style instead of scala doc style in the multiple line comments. Thanks

ibingoogle:
Hi, @yiheng , I have changed the doc style. You can check again. Thanks

yiheng:
jenkins test pass. Merge

PR #2267, opened 2018-02-02
Refine the operation and tf folder

yiheng:
The ops and tf package under nn is a little mess. Some modules should belong to tf but are placed under ops. The PR will clean up such code.

The difference between tf and ops is that modules under ops package are a part of public API. which means we should wrap them in python and add their document; modules under tf package can be operations or layer, they are only for internal usage.

The difference between ops and nn modules. Modules under ops are not derivable, which we can write a backward method.

This PR includes:
1. Set operation backward method to final, in case that developers override them.
2. Move control and dataflow ops to tf package

## How was this patch tested?
unit test


spark/dl/src/test/scala/com/intel/analytics/bigdl/utils/serializer/OperationSerializerSpec.scala
```diff
@@ -50,7 +51,7 @@ class OperationSerializerSpec extends SerializerSpecHelper {
     })
   }
 
-  "All serializer" should "work properly" in {
+  /* "All serializer" should "work properly" in {
```
yangw1234: Are you commenting out these intentionally?

tmp
```diff
@@ -0,0 +1,930 @@
+[INFO] Scanning for projects...
```
yangw1234: delete this file?

yiheng:
jenkins passed

PR #2274, opened 2018-02-05
Refine module serialization tests code

yiheng:
The original serialization test put all module serialization test code into three files. It has some problems:
1. It's too large, which is not easy to maintain
2. User need to modify the same file when adding a new module, easy conflict
3. If a user forgets to add the serial test for the newly added module, the error message is not easy to understand. And it will not be displayed in the error reports, the user has to find it in the log.

This has been fixed in this PR:
1. We split the model serialization test into small classes. User only needs to implement ```ModuleSerializationTest``` in their own ModuleSpec file. It's easier to maintain and less conflict.
2. We use class reflection to load such classes and dynamically generate test cases. So if user forgets to implement serialization test of a new module, the error message is easier to read and can be displayed in Jenkins test error report.

## How was this patch tested?
unit test


wzhongyuan:
This actually looks good, I also tried previously but not sure why it didn't work for nested tests
Just two questions
1) This requires the unit test should have similar naming conventions as corresponding layer with suffix appended, that means for those layers that the tests are not following this , we should change it first especially for keras @zhichao-li @hkvision 
2) what if we want to write multiple test cases for one layer ? are we going to put all into one ?

hkvision:
@wzhongyuan For keras layers, I do think the tests are put wrongly in the folder...

yiheng:
We have a map to record the special mapping(e.g. many to one and keras layers).

We have to put multiple tests for one module into one test class

yiheng:
Looks like the change works. http://172.168.2.101:8080/job/BigDL-PRVN-scalatest-others-Spark-1.6/128/

Now change the error to cancel until all old serial test code has been converted

yiheng:
jenkins test pass

wzhongyuan:
@yiheng   LGTM, please merge it first and then we can continue adding other tests and finally decon old ones

PR #2276, opened 2018-02-06
add notebook token to docker and update jdk8 downloading

glorysdj:
add notebook token to docker which is required by AWS marketplace
update jdk8 downloading

## How was this patch tested?
have tested for both 0.4.0 and 0.5.0-SNAPSHOT


yiheng:
LGTM

PR #2282, opened 2018-02-07
add [[CategoricalColVocaList]] Operation

tosky001:

add CategoricalColVocaList operation to map a feature string to a integer id

## How was this patch tested?

unit tests


tosky001:
@yiheng May i have your review and Thanks.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/CategoricalColVocaList.scala
```diff
@@ -0,0 +1,134 @@
...
+    val rows = input.size(dim = 1)
+
+    val cols = if (numOovBuckets==0) {
+      if (isSetDefault) vocaLen + 1 else vocaLen
+    }
+    else {
+      vocaLen + numOovBuckets
+    }
+    val shape = Array(rows, cols)
+    val indices0 = new ArrayBuffer[Int]()
```
yiheng: Looks like the length of indices0, indices1 and values are known. Can we use Array instead of ArrayBuffer here?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/CategoricalColVocaList.scala
```diff
@@ -0,0 +1,134 @@
...
+    val indices1 = new ArrayBuffer[Int]()
+    val values = new ArrayBuffer[Int]()
+
+    var i = 1
+    while (i <= rows) {
+      var feaStrArr = input.valueAt(i).split(strDelimiter)
+      if (!isSetDefault && numOovBuckets == 0) {
+        feaStrArr = feaStrArr.filter(x => vocaMap.contains(x))
```
yiheng: I don't quite understand here. feaStrAttr contain "true" or "false"? Can you add some comments to explain why to do this? Thanks

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/CategoricalColVocaList.scala
```diff
@@ -0,0 +1,134 @@
...
+      }
+      var j = 0
+      while (j < feaStrArr.length) {
+        val mapVal = numOovBuckets==0 match {
+          case true =>
+            vocaMap.getOrElse(feaStrArr(j), vocaMap.size)
+          case false =>
+            vocaMap.getOrElse(feaStrArr(j),
+              MurmurHash3.stringHash(feaStrArr(j)) % numOovBuckets match {
```
yiheng: Have seen this in another PR. Maybe consider create a util method for this function?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/CategoricalColVocaList.scala
tosky001: @yiheng  well, there are three different ways to deal with the out-of vocabulary values: ignored, default_value and numOovBuckets. the `isSetDefault=false && numOvvBuckets==0` means **ignore**, So the Operation here is aimed to filter the values not in vocabulary list.  it seems better that having a specific explanation in scalaDoc.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/CategoricalColVocaList.scala
tosky001: @yiheng the rows of original input is  certain, but the each row is a string and the length of the string is uncertain after splitting the string into an array. So, the total length is also uncertain.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/CategoricalColVocaList.scala
tosky001: a good idea! but i'm not sure where to create the util method, `nn.ops.*` or `nn.utils.Util.scala`?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/CategoricalColVocaList.scala
yiheng: I see

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/CategoricalColVocaList.scala
yiheng: My bad. Just ignore this comment...

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/CategoricalColVocaList.scala
yiheng: ```nn.Utils``` should be fine

tosky001:
@yiheng  add the nn.utils.HashFunc.scala, which is used to implement some hash methods, and modify the Operation. all the tests run successfully.

yiheng:
jenkins test pass. merged

PR #2283, opened 2018-02-07
refine tests for some layers

wzhongyuan:

Refine some tests with new approach

## How was this patch tested?

unit test


wzhongyuan:
@yiheng  Refined the testes gradually to avoid conflicts with others

hkvision:
Currently Keras side is still adding more layers and serialization tests. Will this influence you?

spark/dl/src/test/scala/com/intel/analytics/bigdl/keras/nn/PermuteSpec.scala
```diff
@@ -41,3 +44,11 @@ class PermuteSpec extends KerasBaseSpec {
   }
 
 }
+class PermuteSerialTest extends ModuleSerializationTest {
```
hkvision: Add a new line after the previous bracket?

spark/dl/src/test/scala/com/intel/analytics/bigdl/keras/nn/ReshapeSpec.scala
```diff
@@ -58,3 +61,12 @@ class ReshapeSpec extends KerasBaseSpec {
...
 }
+
+class ReshapeSerialTest extends ModuleSerializationTest {
+  override def test(): Unit = {
+    val layer = Reshape[Float](Array(4, 15), inputShape = Shape(3, 4, 5))
+    layer.build(Shape(2, 3, 4, 5))
+    val input = Tensor[Float](2, 3, 4, 5).apply1(_ => Random.nextFloat())
+    runSerializationTest(layer, input)
+  }
+}
```
hkvision: Need a new line at the end of the file. Or style test will fail?

wzhongyuan:
@hkvision  it doesn't matter for now, as we still keep the old tests before we updated for all layers
BTW, new lines added

wzhongyuan:
PR validation passed http://172.168.2.101:8080/job/BigDL-PR-Validation-v3/158/

PR #2284, opened 2018-02-07
Support Calling Java Function in Python Executor and ModelBroadcast in Python

yangw1234:

Support Calling Java Function in Python Executor and ModelBroadcast in Python

To enable this feature, user should explicitly call
```python
init_executor_gateway(sc)
```
The approach we took is 
1. launch a JavaGateway server on each executor and write the listening port to a fine in the executor's working directory (typical the ${SPARK_HOMW}/work/app-id/executor-id)
2. in python, if running on executor, find the port on the above file and create a gateway client to connect the executor JVM.

## How was this patch tested?

add a unit test to for spark local mode
manually tested on spark standalone cluster


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -109,6 +112,58 @@ object Engine {
   private var physicalCoreNumber = -1
   private var nodeNum: Int = -1
 
+  private var gatewayServer: py4j.GatewayServer = null
+
+
+  private[bigdl] def createJavaGateway(driverPort: Int): Unit = {
```
zhichao-li: should it be a singleton here? i.e createJavaGateway is invoked multiple times on the same Executor.

yangw1234: On java side, Engine is already a singleton, so it won't create multiple servers.

pyspark/bigdl/util/common.py
```diff
@@ -610,9 +634,14 @@ def _java2py(sc, r, encoding="bytes"):
 
 def callJavaFunc(sc, func, *args):
     """ Call Java Function """
-    args = [_py2java(sc, a) for a in args]
+    if sc is None:
```
zhichao-li: Move this checking into `GatewayWrapper` ? 
```
if run_on_driver:
    return get_spark_context()._jvm
else:
   return create_gateway()....
```

pyspark/bigdl/util/common.py
```diff
@@ -533,6 +541,8 @@ def get_spark_context(conf=None):
     :param conf: combining bigdl configs into spark conf
     :return: SparkContext
     """
+    if SparkFiles._is_running_on_worker:
```
zhichao-li: remove this?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -109,6 +112,58 @@ object Engine {
...
+    }
+    if (gatewayServer != null) return
+    this.synchronized {
+      if (gatewayServer == null) {
+        gatewayServer = new py4j.GatewayServer(null, 0)
+      }
+    }
+    val thread = new Thread(new Runnable() {
+      override def run(): Unit = try {
+        gatewayServer.start()
```
zhichao-li: should we add a hooker to stop the gatewayServer once the thread is closed? i.e:
`Runtime.getRuntime().addShutdownHook`

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -109,6 +112,64 @@ object Engine {
   private var physicalCoreNumber = -1
   private var nodeNum: Int = -1
 
+  @volatile
+  private var gatewayServer: py4j.GatewayServer = null
+
+
+  private[bigdl] def createJavaGateway(driverPort: Int): Unit = {
+    if (SparkUtils.isDriver) {
+      val file = new java.io.File(SparkFiles.getRootDirectory(), "gateway_port")
```
yiheng: Should we make sure it's thread safe when check and write the file?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -109,6 +112,64 @@ object Engine {
...
+      if (file.exists()) {
+        file.delete()
+      }
+      file.createNewFile()
```
yiheng: Please add some logs when initializing gateway server, makes it easier to monitor

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -109,6 +112,64 @@ object Engine {
...
+      val out = new PrintWriter(file)
```
yiheng: should we try to catch the file create exception and throw a more meaningful error message?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -109,6 +112,64 @@ object Engine {
...
+      return
+    }
+    if (gatewayServer != null) return
+    this.synchronized {
+      if (gatewayServer != null) return
+      gatewayServer = new py4j.GatewayServer(null, 0)
+    }
+    val thread = new Thread(new Runnable() {
+      override def run(): Unit = try {
+        gatewayServer.start()
```
yiheng: If the gatewayServer crashs, will it output some log and where can we find it?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -109,6 +112,64 @@ object Engine {
...
+      } catch {
+        case ct: ControlThrowable =>
+          throw ct
+        case t: Throwable =>
+          throw new Exception(s"Uncaught exception in thread ${Thread.currentThread().getName}", t)
```
yiheng: Please replace with a more meaning error message

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -109,6 +112,64 @@ object Engine {
...
+      }
+    })
+    thread.setName("py4j-executor-gateway-init")
+    thread.setDaemon(true)
+    thread.start()
+
+    thread.join()
```
yiheng: why thread.join here?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -109,6 +112,64 @@ object Engine {
...
+
+    Runtime.getRuntime().addShutdownHook(new Thread {
+      override def run(): Unit = {
+        gatewayServer.shutdown()
+      }
+    })
+
+    val file = new java.io.File(SparkFiles.getRootDirectory(), "gateway_port")
```
yiheng: should it be thread safe?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -109,6 +112,64 @@ object Engine {
...
+    if (file.exists()) {
+      file.delete()
+    }
+    file.createNewFile()
```
yiheng: try to catch the exceptions and throw a more meaningful error

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -109,6 +112,64 @@ object Engine {
...
+      out.flush()
+      out.close()
+      return
+    }
+    if (gatewayServer != null) return
+    this.synchronized {
+      if (gatewayServer != null) return
+      gatewayServer = new py4j.GatewayServer(null, 0)
+    }
+    val thread = new Thread(new Runnable() {
```
yiheng: what if another thread try to use this gateway while the gateway havn't inited

spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDL.scala
```diff
@@ -2940,6 +2935,11 @@ class PythonBigDL[T: ClassTag](implicit ev: TensorNumeric[T]) extends Serializab
     maxIteration: Int): SequentialSchedule = {
     seq.add(scheduler, maxIteration)
   }
+
+  def initExecutorGateway(sc: JavaSparkContext, driverPort: Int): Unit = {
```
yiheng: private bigdl?

pyspark/bigdl/util/common.py
```diff
@@ -438,6 +443,9 @@ def uniform(self, a, b, size):
 def init_engine(bigdl_type="float"):
     callBigDlFunc(bigdl_type, "initEngine")
 
+def init_executor_gateway(sc, bigdl_type="float"):
```
yiheng: when call this method?

yiheng:
In each execution, between the env and agent, there will be a tensor serial and deserial, right?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
yangw1234: Only one thread can reach this part of code.

yangw1234:
@yiheng yes, but I can't think of any way around this.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
yangw1234: Done

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
yangw1234: Done

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
yangw1234: done

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
yangw1234: done

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
yangw1234: done

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
yangw1234: done

spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDL.scala
yangw1234: done

pyspark/test/bigdl/test_simple_integration.py
```diff
@@ -533,7 +533,7 @@ def test_save_jtensor_dict(self):
         tensors["tensor1"] = JTensor.from_ndarray(np.random.rand(3, 2))
         tensors["tensor2"] = JTensor.from_ndarray(np.random.rand(3, 2))
         # in old impl, this will throw an exception
-        _py2java(self.sc, tensors)
+        _py2java(self.sc._gateway, tensors)
```
zhichao-li: add unittest for the new added ModelBroadcast?

pyspark/bigdl/models/utils/model_broadcast.py
```diff
@@ -0,0 +1,74 @@
...
+import sys
+import gc
+from tempfile import NamedTemporaryFile
+
+from pyspark.cloudpickle import print_exec
+from pyspark.broadcast import Broadcast
+from bigdl.nn.layer import Model
+
+def _from_id_and_type(bid, bigdl_type):
+    from pyspark.broadcast import _broadcastRegistry
```
zhichao-li: call method `_from_id` directly to avoid duplicating the code here?

pyspark/test/bigdl/test_simple_integration.py
yangw1234: added

zhichao-li:
LGTM

yangw1234:
PR Validation pass

pyspark/bigdl/util/common.py
```diff
@@ -556,16 +564,33 @@ def get_spark_sql_context(sc):
     else:
         return SQLContext(sc)  # Compatible with Spark1.5.1
 
+def _get_port():
+    root_dir = SparkFiles.getRootDirectory()
+    path = os.path.join(root_dir, "gateway_port")
+    f = open(path)
```
jason-dai: what if this fails?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -109,6 +113,83 @@ object Engine {
...
+      out.print(port)
+      out.flush()
+    } finally {
+      out.close()
+    }
+  }
+
+  private[bigdl] def createJavaGateway(driverPort: Int): Unit = {
+    if (SparkUtils.isDriver) {
+      if (driverPortFileCreated.compareAndSet(false, true)) {
```
jason-dai: Why is it needed on driver?

PR #2285, opened 2018-02-07
[WIP] Module Initializer

sperlingxx:

Module Initializer, a tool for initializing(reflecting) Modules with a Table.
The propose of this tool is providing a parametric way to build Networks. 
On the basis of this tool, maybe we can build whole network structures by configurations in future. 
For now, it can parse default values of parameters in _Constructor_. So, it's not too hard to use. 

This is just a preliminary version :-)

## How was this patch tested?

unit test

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/ModuleInitializer.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/ModuleInitializer.scala
@@ -25,22 +25,23 @@ import scala.reflect.runtime.universe._
 import scala.util.Try
 
 /**
- *
+ *  ======== ModuleInitializer ========
+ *  ModuleInitializer, a tool for initializing(reflecting) Modules with a Table.
  */
 object ModuleInitializer {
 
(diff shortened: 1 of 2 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/ModuleInitializer.scala
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ *  ======== ModuleInitializer ========
+ *  ModuleInitializer, a tool for initializing(reflecting) Modules with a Table.
+ */
+object ModuleInitializer {
+
+  private val prefix = "com.intel.analytics.bigdl.nn."
+
+  def init[T: ClassTag](state: Table)(implicit ev: TensorNumeric[T]): Module[T] = {
```
qiuxin2012: I think you can move this function to `Module.apply`

qiuxin2012: Add Scala doc, Add markdown for this new feature.

Issue #2281, opened 2018-02-07
Some code need to be refactor for keras implementation


zhichao-li:
- Change the default value of isCompatibleWithKeras to be false and override it if needed.
- Change the outputShape from Array[Shape] to be Shape.
- seq.add(dense), check if the outputShape of seq equals to the manually specified inputShape of dense or not. // and also seq.add(linear). // if there's getInputShape then we would trigger such checking? or check the isBuilt?
- `linear -> seq -> linear` can be put into a KModel for now. we need to iterate the whole graph on the constructing stage to see if all of the node is compatible with keras or vice versus.
- final `getInputShape` and `getOutputShape` as we do not go through building process while reload model from file. `InputShapeValue` and `outputShapeValue` is the final answer.

hkvision:
We need to validate outputshape all entries >0 after building and calculation.

Issue #1934, opened 2017-11-24
SoftMax doesn't calculate correct values for 3D inputs


hkvision:
Seems that SoftMax is not working properly for 3D inputs
```
>>> layer = SoftMax()
creating: createSoftMax
>>> input = np.random.random([1, 2, 3])
>>> input
array([[[ 0.39775828,  0.11849257,  0.32446772],
        [ 0.13931134,  0.99062448,  0.67649278]]])
>>> layer.forward(input)
array([[[ 1.,  1.,  1.],
        [ 1.,  1.,  1.]]], dtype=float32)

>>> input_data = np.array([[[1, 2, 3], [4, 5, 6]], [[7, 8, 9], [10, 11, 12]]])
>>> layer.forward(input_data)
array([[[ 0.00247262,  0.00247262,  0.00247262],
        [ 0.00247262,  0.00247262,  0.00247262]],

       [[ 0.99752742,  0.99752742,  0.99752742],
        [ 0.99752742,  0.99752742,  0.99752742]]], dtype=float32)
```

qiuxin2012:
SoftMax's implement follows torch, it's same result with torch.

hkvision:
For 3D input, the result doesn't match with keras for the axis.

hkvision:
Now use transpose first before applying softmax for 3D input for keras model loading so that the result is the same with keras.

PR #2293, opened 2018-02-09
Enhance and refactor the logic of InferShape

zhichao-li:
- Enhance the checking for mix use of keras layer and torch layer.
- Change the outputShape from Array[Shape] to be Shape.
- final `getInputShape` and `getOutputShape` as we do not go through building process while reload model from file. `InputShapeValue` and `outputShapeValue` is the final answer.
- Compare the inputShape from user with the shape from the building stage. 
- Add flag into `InferShape` to identify if this layer being used or not
- `Sequential` should be frozen once being used in `add` or `inputs` 
## How was this patch tested?
unittest
fixed https://github.com/intel-analytics/BigDL/issues/2281

### DIFF
--- a/spark/dl/src/main/resources/serialization/bigdl.proto
+++ b/spark/dl/src/main/resources/serialization/bigdl.proto
@@ -1,5 +1,5 @@
 syntax = "proto3";
-package serialization;
+package com.intel.analytics.bigdl.serialization;
 import "google/protobuf/any.proto";
 message BigDLModule
 {
@@ -16,7 +16,7 @@ message BigDLModule
     string namePostfix = 11; // name post fix
     int32 id = 12; // unique ID of this module , used for shared modules
     Shape inputShape = 13; // input shape
-    repeated Shape outputShape = 14; //output shape
+    Shape outputShape = 14; //output shape
 }
 enum VarFormat {
     EMPTY_FORMAT = 0;
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Container.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Container.scala
@@ -16,7 +16,6 @@
 
 package com.intel.analytics.bigdl.nn
 
-import com.intel.analytics.bigdl.Module
 import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, Activity}
 import com.intel.analytics.bigdl.tensor.Tensor
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
@@ -45,12 +44,6 @@ abstract class Container[A <: Activity : ClassTag,
   // list of sub modules
   val modules: ArrayBuffer[AbstractModule[Activity, Activity, T]]
   = ArrayBuffer[AbstractModule[Activity, Activity, T]]()
-
-  override private[bigdl] def isCompatibleWithKeras(): Boolean = false
-
-  override private[bigdl] def isCompatibleWithTorch(): Boolean = {
-    modules.filter(!_.isCompatibleWithTorch()).length <= 0
-  }
 
   override def reset(): Unit = {
     modules.foreach(_.reset())
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/DynamicContainer.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/DynamicContainer.scala
@@ -41,7 +41,7 @@ abstract class DynamicContainer[A <: Activity : ClassTag, B <: Activity : ClassT
     require(!module.isInstanceOf[Operation[_, _, _]],
       "Add operations to dynamic container is not allowed, as operations don't have backward. " +
         "Operation can only be used in Graph")
-    Util.excludeNotTorch[T](Seq(module))
+    validateInput[T](Seq(module))
     modules += module.asInstanceOf[AbstractModule[Activity, Activity, T]]
     checkDuplicate()
     this
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Linear.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Linear.scala
@@ -80,10 +80,6 @@ class Linear[T: ClassTag](
     zeroGradParameters()
   }
 
-  override def computeOutputShape(inputShape: Shape): Shape = {
-    inputShape.copyAndUpdate(-1, outputSize)
-  }
-
   override def updateOutput(input: Tensor[T]): Tensor[T] = {
     require(input.dim() == 1 || input.dim() == 2,
       "Linear: " + ErrorInfo.constrainInputAsVectorOrBatch +
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Maxout.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Maxout.scala
@@ -62,12 +62,12 @@ class Maxout[T: ClassTag](val inputSize: Int, val outputSize: Int, val maxoutNum
   }
 
   override def updateOutput(input: Tensor[T]): Tensor[T] = {
-    output = layer.updateOutput(input)
+    output = layer.updateOutput(input).toTensor
     output
   }
 
   override def updateGradInput(input: Tensor[T], gradOutput: Tensor[T]): Tensor[T] = {
-    gradInput = layer.updateGradInput(input, gradOutput)
+    gradInput = layer.updateGradInput(input, gradOutput).toTensor
     gradInput
   }
 
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialFullConvolution.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialFullConvolution.scala
@@ -18,15 +18,13 @@ package com.intel.analytics.bigdl.nn
 
 import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, Activity, Initializable}
 import com.intel.analytics.bigdl.optim.Regularizer
+import com.intel.analytics.bigdl.serialization.Bigdl.{AttrValue, BigDLModule}
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
 import com.intel.analytics.bigdl.tensor._
-import com.intel.analytics.bigdl.utils.{Shape, T, Table, serializer}
-import com.intel.analytics.bigdl.utils.RandomGenerator._
 import com.intel.analytics.bigdl.utils.serializer._
 import com.intel.analytics.bigdl.utils.serializer.converters.DataConverter
-import com.intel.analytics.bigdl.serialization.Bigdl.{AttrValue, BigDLModule}
+import com.intel.analytics.bigdl.utils.{Shape, T, Table}
 
-import scala.concurrent.Future
 import scala.reflect.ClassTag
 import scala.reflect.runtime.universe
 
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/StaticGraph.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/StaticGraph.scala
@@ -36,7 +36,7 @@ class StaticGraph[T: ClassTag](
   private val _inputs : Seq[ModuleNode[T]],
   private val _outputs : Seq[ModuleNode[T]],
   private val _variables: Option[(Array[Tensor[T]], Array[Tensor[T]])] = None,
-  private val excludeKeras: Boolean = true
+  private val enableExcludeChecking: Boolean = true
 )(implicit ev: TensorNumeric[T]) extends Graph[T](_inputs, _outputs, _variables) {
   private val forwardExecution = forwardGraph.topologySort.reverse
   private var backwardExecution: Array[Node[AbstractModule[Activity, Activity, T]]] = _
@@ -44,9 +44,8 @@ class StaticGraph[T: ClassTag](
   private var backId2ForwardId: Array[Int] = _
   private var gradOutputCache: Array[Activity] = _
 
-  if (excludeKeras) {
-    Util.excludeNotTorch(inputs.map(_.element))
-    Util.excludeNotTorch(outputs.map(_.element))
+  if (enableExcludeChecking) {
+    excludeInvalidLayers(forwardExecution.map {_.element})
   }
 
   buildBackwardGraph()
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/UpSampling1D.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/UpSampling1D.scala
@@ -16,8 +16,6 @@
 
 package com.intel.analytics.bigdl.nn
 
-import java.util
-
 import com.intel.analytics.bigdl.nn.abstractnn.TensorModule
 import com.intel.analytics.bigdl.tensor.Tensor
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/InferShape.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/InferShape.scala
@@ -27,10 +27,6 @@ class InvalidLayer(msg: String) extends RuntimeException(msg)
 
 trait InferShape {
 
-  private[bigdl] var usedAsSrc = false
-
-  private[bigdl] var usedAsDest = false
-
   private[bigdl] var _inputShapeValue: Shape = null
 
   private[bigdl] var _outputShapeValue: Shape = null
@@ -87,41 +83,6 @@ trait InferShape {
     throw new RuntimeException("Haven't been implemented yet. Do not use it with Keras Layer")
   }
 
-
-
-  private def ensureNotShared[T: ClassTag](modules : Seq[AbstractModule[_, _, T]]): Unit = {
-    def skip(infer: InferShape): Boolean = {
-      infer.isInstanceOf[TInput[_]] || infer.isInstanceOf[KInput[_]]
-    }
-    // We can add module into Sequential multiple times.
-    if (this.isInstanceOf[KSequential[_]]) {
-      val submodules = this.asInstanceOf[KSequential[_]].getSubModules()
-      if (!submodules.isEmpty) {
-        if (!skip(submodules.last) && submodules.last.usedAsSrc == true) {
-          throw new RuntimeException(s"Reuse module as src is not allowed: $this")
-        }
-        submodules.last.usedAsSrc = true
-      }
-      modules.foreach{module =>
-        if (!skip(this) && module.usedAsDest == true) {
-          throw new RuntimeException(s"Reuse module as dest is not allowed: $this")
-        }
-        module.usedAsDest = true
-      }
-    } else {
-      if (!skip(this) && this.usedAsDest == true) {
-        throw new RuntimeException(s"Reuse module as dest is not allowed: $this")
-      }
-      this.usedAsDest = true
-      modules.foreach{module =>
-        if (!skip(module) && module.usedAsSrc == true) {
-          throw new RuntimeException(s"Reuse module as src is not allowed: $this")
-        }
-        module.usedAsSrc = true
-      }
-    }
-  }
-
   private[bigdl] def excludeInvalidLayers[T: ClassTag]
   (modules : Seq[AbstractModule[_, _, T]]): Unit = {
     val invalidNodes = if (this.isKerasStyle()) {
@@ -139,7 +100,6 @@ trait InferShape {
   private[bigdl] def validateInput[T: ClassTag](modules : Seq[AbstractModule[_, _, T]]): Unit = {
     if (this.isKerasStyle()) {
       require(modules != null && !modules.isEmpty, "Empty input is not allowed")
-      ensureNotShared(modules)
     }
     excludeInvalidLayers(modules)
   }
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/InferShape.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/InferShape.scala
@@ -16,11 +16,11 @@
 
 package com.intel.analytics.bigdl.nn.abstractnn
 
-import com.intel.analytics.bigdl.nn.keras.{Sequential => KSequential, Input => KInput}
+import com.intel.analytics.bigdl.nn.keras.{Input => KInput, Sequential => KSequential}
 import com.intel.analytics.bigdl.nn.{Input => TInput}
 import com.intel.analytics.bigdl.utils.Shape
-import scala.language.existentials
 
+import scala.language.existentials
 import scala.reflect.ClassTag
 
 class InvalidLayer(msg: String) extends RuntimeException(msg)
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/InferShape.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/InferShape.scala
@@ -19,6 +19,7 @@ package com.intel.analytics.bigdl.nn.abstractnn
 import com.intel.analytics.bigdl.nn.keras.{Sequential => KSequential, Input => KInput}
 import com.intel.analytics.bigdl.nn.{Input => TInput}
 import com.intel.analytics.bigdl.utils.Shape
+import scala.language.existentials
 
 import scala.reflect.ClassTag
 
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/InferShape.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/InferShape.scala
@@ -16,24 +16,34 @@
 
 package com.intel.analytics.bigdl.nn.abstractnn
 
+import com.intel.analytics.bigdl.nn.keras.{Sequential => KSequential, Input => KInput}
+import com.intel.analytics.bigdl.nn.{Input => TInput}
 import com.intel.analytics.bigdl.utils.Shape
+
+import scala.reflect.ClassTag
+
+class InvalidLayer(msg: String) extends RuntimeException(msg)
 
 trait InferShape {
 
+  private[bigdl] var usedAsSrc = false
+
+  private[bigdl] var usedAsDest = false
+
   private[bigdl] var _inputShapeValue: Shape = null
 
-  private[bigdl] var _outputShapeValue: Array[Shape] = Array[Shape]()
+  private[bigdl] var _outputShapeValue: Shape = null
 
   private[bigdl] def inputShapeValue: Shape = _inputShapeValue
 
-  private[bigdl] def outputShapeValue: Array[Shape] = _outputShapeValue
+  private[bigdl] def outputShapeValue: Shape = _outputShapeValue
 
   // scalastyle:off
   private[bigdl] def inputShapeValue_=(value: Shape): Unit = {
     _inputShapeValue = value
   }
 
-  private[bigdl] def outputShapeValue_=(value: Array[Shape]): Unit = {
+  private[bigdl] def outputShapeValue_=(value: Shape): Unit = {
     _outputShapeValue = value
   }
   // scalastyle:on
@@ -41,28 +51,15 @@ trait InferShape {
   /**
    * We suppose the first dim is batch
    */
-  private[bigdl] def getInputShape(): Shape = {
+  private[bigdl] final def getInputShape(): Shape = {
     _inputShapeValue
-  }
-
-  /**
-   * Get the outputshape by index.
-   * @param index start from 0
-   * @return
-   */
-  private[bigdl] def getOutputShapeFor(index: Int): Shape = {
-    _outputShapeValue(index)
   }
 
   /**
    * We suppose the first dim is batch
    */
-  private[bigdl] def getOutputShape(): Shape = {
-    if (_outputShapeValue.length > 1) {
-      throw new RuntimeException(
-        "There are multiple outputs for this layer. Please use getInputShapeFor instead")
-    }
-    outputShapeValue(0)
+  private[bigdl] final def getOutputShape(): Shape = {
+    outputShapeValue
   }
 
   /**
@@ -71,24 +68,79 @@ trait InferShape {
    */
   private[bigdl] def build(inputShape: Shape): Shape = {
     val outputShape = computeOutputShape(inputShape)
-    this._outputShapeValue ++ Array(outputShape)
-    this._inputShapeValue = inputShape
-    isBuilt = true
+    this.outputShapeValue = outputShape
+    this.inputShapeValue = inputShape
     outputShape
   }
 
-  private[bigdl] var isBuilt: Boolean = false
+  private[bigdl] def isBuilt(): Boolean = outputShapeValue != null
 
+  private[bigdl] def isKerasStyle(): Boolean = false
 
-  private[bigdl] def isCompatibleWithKeras(): Boolean = true
-
-  private[bigdl] def isCompatibleWithTorch(): Boolean = true
+  private[bigdl] def allowRebuilt(): Boolean = false
 
   /**
    * We suppose the first dim is batch
    */
   private[bigdl] def computeOutputShape(inputShape: Shape): Shape = {
     throw new RuntimeException("Haven't been implemented yet. Do not use it with Keras Layer")
+  }
+
+
+
+  private def ensureNotShared[T: ClassTag](modules : Seq[AbstractModule[_, _, T]]): Unit = {
+    def skip(infer: InferShape): Boolean = {
+      infer.isInstanceOf[TInput[_]] || infer.isInstanceOf[KInput[_]]
+    }
+    // We can add module into Sequential multiple times.
+    if (this.isInstanceOf[KSequential[_]]) {
+      val submodules = this.asInstanceOf[KSequential[_]].getSubModules()
+      if (!submodules.isEmpty) {
+        if (!skip(submodules.last) && submodules.last.usedAsSrc == true) {
+          throw new RuntimeException(s"Reuse module as src is not allowed: $this")
+        }
+        submodules.last.usedAsSrc = true
+      }
+      modules.foreach{module =>
+        if (!skip(this) && module.usedAsDest == true) {
+          throw new RuntimeException(s"Reuse module as dest is not allowed: $this")
+        }
+        module.usedAsDest = true
+      }
+    } else {
+      if (!skip(this) && this.usedAsDest == true) {
+        throw new RuntimeException(s"Reuse module as dest is not allowed: $this")
+      }
+      this.usedAsDest = true
+      modules.foreach{module =>
+        if (!skip(module) && module.usedAsSrc == true) {
+          throw new RuntimeException(s"Reuse module as src is not allowed: $this")
+        }
+        module.usedAsSrc = true
+      }
+    }
+  }
+
+  private[bigdl] def excludeInvalidLayers[T: ClassTag]
+  (modules : Seq[AbstractModule[_, _, T]]): Unit = {
+    val invalidNodes = if (this.isKerasStyle()) {
+      modules.filter{!_.isKerasStyle()}
+    } else {
+      modules.filter{_.isKerasStyle()}
+    }
+    if (invalidNodes.length > 0) {
+      throw new InvalidLayer(s"""Do not mix ${this}(isKerasStyle=${isKerasStyle()}) with Layer
+                           (isKerasStyle=${invalidNodes(0).isKerasStyle()}):
+         ${invalidNodes.mkString(",")}""")
+    }
+  }
+
+  private[bigdl] def validateInput[T: ClassTag](modules : Seq[AbstractModule[_, _, T]]): Unit = {
+    if (this.isKerasStyle()) {
+      require(modules != null && !modules.isEmpty, "Empty input is not allowed")
+      ensureNotShared(modules)
+    }
+    excludeInvalidLayers(modules)
   }
 }
 
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Activation.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Activation.scala
@@ -16,7 +16,7 @@
 
 package com.intel.analytics.bigdl.nn.keras
 
-import com.intel.analytics.bigdl.nn.abstractnn.AbstractModule
+import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, IdentityOutputShape}
 import com.intel.analytics.bigdl.tensor.Tensor
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
 import com.intel.analytics.bigdl.utils.Shape
@@ -37,15 +37,14 @@ import scala.reflect.ClassTag
 class Activation[T: ClassTag](
    val activation: String,
    val inputShape: Shape = null)(implicit ev: TensorNumeric[T])
-  extends KerasLayer[Tensor[T], Tensor[T], T](KerasLayer.addBatch(inputShape)) {
+  extends KerasLayer[Tensor[T], Tensor[T], T](KerasLayer.addBatch(inputShape))
+    with IdentityOutputShape {
 
   require(activation != null, "The name of an activation function as a string is required")
 
   override def doBuild(inputShape: Shape): AbstractModule[Tensor[T], Tensor[T], T] = {
-    val model = Sequential[T]()
-    model.add(InputLayer(inputShape = KerasLayer.removeBatch(inputShape)))
-    val layer = KerasUtils.getActivation(activation)
-    model.add(layer).asInstanceOf[AbstractModule[Tensor[T], Tensor[T], T]]
+    val kerasActivation = KerasUtils.getKerasActivation(activation)
+    kerasActivation.doBuild(inputShape)
   }
 }
 
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/AtrousConvolution1D.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/AtrousConvolution1D.scala
@@ -16,7 +16,7 @@
 
 package com.intel.analytics.bigdl.nn.keras
 
-import com.intel.analytics.bigdl.nn.abstractnn.AbstractModule
+import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, Activity}
 import com.intel.analytics.bigdl.nn.{InitializationMethod, Xavier, Zeros}
 import com.intel.analytics.bigdl.nn.{SpatialDilatedConvolution, Squeeze, Transpose, Sequential => TSequential}
 import com.intel.analytics.bigdl.optim.Regularizer
@@ -57,7 +57,7 @@ class AtrousConvolution1D[T: ClassTag](
    val nbFilter: Int,
    val filterLength: Int,
    val init: InitializationMethod = Xavier,
-   val activation: AbstractModule[Tensor[T], Tensor[T], T] = null,
+   val activation: KerasLayer[Tensor[T], Tensor[T], T] = null,
    val subsampleLength: Int = 1,
    val atrousRate: Int = 1,
    var wRegularizer: Regularizer[T] = null,
@@ -95,7 +95,7 @@ class AtrousConvolution1D[T: ClassTag](
     model.add(Transpose(Array((2, 3))))
     model.add(Squeeze(4))
     if (activation != null) {
-      model.add(activation)
+      model.add(activation.doBuild(inputShape))
     }
     model.asInstanceOf[AbstractModule[Tensor[T], Tensor[T], T]]
   }
@@ -113,7 +113,7 @@ object AtrousConvolution1D {
     bRegularizer: Regularizer[T] = null,
     inputShape: Shape = null)(implicit ev: TensorNumeric[T]): AtrousConvolution1D[T] = {
     new AtrousConvolution1D[T](nbFilter, filterLength, KerasUtils.getInitMethod(init),
-      KerasUtils.getActivation(activation), subsampleLength, atrousRate,
+      KerasUtils.getKerasActivation(activation), subsampleLength, atrousRate,
       wRegularizer, bRegularizer, inputShape)
   }
 }
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasLayer.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasLayer.scala
@@ -18,16 +18,17 @@ package com.intel.analytics.bigdl.nn.keras
 
 import com.intel.analytics.bigdl._
 import com.intel.analytics.bigdl.nn.Graph._
-import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, Activity, TensorModule}
+import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, Activity}
 import com.intel.analytics.bigdl.nn.keras.{Sequential => KSequential}
-import com.intel.analytics.bigdl.nn.{Container, Sequential => TSequential}
+import com.intel.analytics.bigdl.nn.{Container => TContainer, Sequential => TSequential}
+import com.intel.analytics.bigdl.serialization.Bigdl.{AttrValue, BigDLModule}
 import com.intel.analytics.bigdl.tensor.Tensor
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
 import com.intel.analytics.bigdl.utils.serializer._
 import com.intel.analytics.bigdl.utils.serializer.converters.DataConverter
-import com.intel.analytics.bigdl.utils.{Shape, SingleShape, Util}
-import serialization.Bigdl.{AttrValue, BigDLModule}
+import com.intel.analytics.bigdl.utils.{MultiShape, Shape, SingleShape}
 
+import scala.collection.JavaConverters._
 import scala.collection.mutable.ArrayBuffer
 import scala.reflect.ClassTag
 
@@ -41,7 +42,21 @@ private[bigdl] trait TKerasSerializerHelper {
   }
 }
 
-object KerasLayerSerializer extends ContainerSerializable with TKerasSerializerHelper{
+object KerasLayerSerializer extends KerasLayerSerializable
+
+trait KerasLayerSerializable extends ContainerSerializable with TKerasSerializerHelper{
+
+  override def loadSubModules[T: ClassTag](context : DeserializeContext,
+      module : AbstractModule[Activity, Activity, T])
+    (implicit ev: TensorNumeric[T]) : Unit = {
+    val klayer = module.asInstanceOf[KerasLayer[Activity, Activity, T]]
+    val subModules = context.bigdlModule.getSubModulesList.asScala
+    subModules.foreach(module => {
+      val subModuleData = ModuleSerializer.load(DeserializeContext(module,
+        context.storages, context.storageType, _copyWeightAndBias))
+      klayer.labor = subModuleData.module
+    })
+  }
 
   override def doSerializeModule[T: ClassTag](context: SerializeContext[T],
                                               moduleBuilder : BigDLModule.Builder)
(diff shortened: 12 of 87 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/InferShape.scala
```diff
@@ -71,7 +64,7 @@ trait InferShape {
    */
   private[bigdl] def build(inputShape: Shape): Shape = {
```
jason-dai: what if it is already built?

jason-dai: Or what if inputShape is different from _inputShapeValue?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/InferShape.scala
```diff
@@ -80,7 +73,7 @@ trait InferShape {
   private[bigdl] var isBuilt: Boolean = false
```
jason-dai: We can remove this.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/InferShape.scala
jason-dai: Or what if inputShape is different from the one user specified for Torch-style layers?

zhichao-li: `ensureNotShared ` added to ensure a layer can only be built once.
`checkWithCurrentInputShape ` added to ensure the inputshape from "graph" is the same with the user specified one.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasLayer.scala
```diff
@@ -186,12 +195,14 @@ abstract class KerasLayer[A <: Activity: ClassTag, B <: Activity: ClassTag, T: C
...
-    Util.excludeNotKeras(nodes.map(_.element))
+    excludeInvalidLayers(nodes.map(_.element))
+    ensureNotShared(nodes.map{_.element})
     if (!nodes.isEmpty) { // as there's  Identity().inputs() within Graph
     val inputShape = Shape(nodes.map{_.element.getOutputShape()}.toList)
       this.build(inputShape)
     }
-    super.inputs(nodes: _*)
+
+    processInputs(nodes)
```
zhichao-li: this can be replaced with super.inputs() only that the `excludeInvalidLayers` would be executed twice

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Cropping2D.scala
```diff
@@ -43,7 +43,7 @@ class Cropping2D[T: ClassTag](
     val heightCrop: Array[Int],
     val widthCrop: Array[Int],
     val dataFormat: DataFormat = DataFormat.NCHW
-  )(implicit ev: TensorNumeric[T]) extends TensorModule[T] {
```
hkvision: Don't need to change this file?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Cropping3D.scala
```diff
@@ -15,7 +15,7 @@
  */
 package com.intel.analytics.bigdl.nn
 
-import com.intel.analytics.bigdl.nn.abstractnn.TensorModule
+import com.intel.analytics.bigdl.nn.abstractnn.{TensorModule}
```
hkvision: ??

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/LocallyConnected2D.scala
```diff
@@ -168,23 +169,6 @@ class LocallyConnected2D[T: ClassTag](
...
-          Array(batchSize, nOutputPlane, oh, ow)
-        }
-      case DataFormat.NHWC =>
-        if (batchSize == -1) {
-          Array(oh, ow, nOutputPlane)
-        } else {
-          Array(batchSize, oh, ow, nOutputPlane)
-        }
-    }
-  }
```
hkvision: Also need to move unnecessary import Shape from old layers.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/LocallyConnected2D.scala
```diff
@@ -72,6 +73,44 @@ class LocallyConnected2D[T: ClassTag](
...
+      } else {
+        Utils.getOutSizeAndPadding(inputHeight, inputWidth, strideH, strideW,
+          kernelH, kernelW, padH, padW, ceilMode = false)
+      }
+    val padTop = sizes(0)
+    val padBottom = sizes(1)
+    val padLeft = sizes(2)
+    val padRight = sizes(3)
+    val outputHeight = sizes(4)
+    val outputWidth = sizes(5)
```
hkvision: If we put compute output logic in the new layer, we can refer to Keras and have a much cleaner and shorter code logic for this layer and other layers as well.
Can use utils KerasUtils.computeConvOutputLength.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/LocallyConnected2D.scala
```diff
@@ -921,6 +894,28 @@ class LocallyConnected2D[T: ClassTag](
 }
 
 object LocallyConnected2D {
+
+  private[bigdl] def getOutputShape(format: DataFormat,
```
hkvision: Why remove this to be a static method?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Maxout.scala
```diff
@@ -16,7 +16,7 @@
 
 package com.intel.analytics.bigdl.nn
 
-import com.intel.analytics.bigdl.nn.abstractnn.TensorModule
+import com.intel.analytics.bigdl.nn.abstractnn.{TensorModule}
```
hkvision: ??

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/ThresholdedReLU.scala
```diff
@@ -39,7 +39,8 @@ import scala.reflect.ClassTag
 class ThresholdedReLU[T: ClassTag](
    val theta: Double = 1.0,
    val inputShape: Shape = null)(implicit ev: TensorNumeric[T])
-  extends KerasLayer[Tensor[T], Tensor[T], T](KerasLayer.addBatch(inputShape)) {
+  extends KerasLayer[Tensor[T], Tensor[T], T](KerasLayer.addBatch(inputShape))
+    with IdentityOutputShape{
```
hkvision: Add a space between IdentityOutputShape and {
So that every new layer is formatted.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/LocallyConnected2D.scala
```diff
@@ -16,7 +16,9 @@
 
 package com.intel.analytics.bigdl.nn.keras
 
-import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, DataFormat, TensorModule}
+import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, Activity, DataFormat}
+import com.intel.analytics.bigdl.nn.{LocallyConnected2D => TLocallyConnected2D}
+import com.intel.analytics.bigdl.nn.{Container => TContainer}
```
hkvision: Are you changing this file?

spark/dl/src/test/scala/com/intel/analytics/bigdl/keras/nn/LocallyConnected2DSpec.scala
```diff
@@ -54,10 +54,10 @@ class LocallyConnected2DSpec extends KerasBaseSpec {
         |model = Model(input=input_tensor, output=output_tensor)
       """.stripMargin
     val seq = KSequential[Float]()
-    val layer = LocallyConnected2D[Float](32, 2, 2,
-      activation = "relu", inputShape = Shape(12, 24, 24))
+    val layer = LocallyConnected2D[Float](32, 2, 2, inputShape = Shape(12, 24, 24))
     seq.add(layer)
-    checkOutputAndGrad(seq.asInstanceOf[AbstractModule[Tensor[Float], Tensor[Float], Float]],
+    layer.build(KerasLayer.addBatch(Shape(12, 24, 24)))
```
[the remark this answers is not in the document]
zhichao-li: typo for testing. let me revert it back.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialConvolution.scala
```diff
@@ -70,7 +70,8 @@ class SpatialConvolution[T: ClassTag](
   val initGradBias: Tensor[T] = null,
   val withBias: Boolean = true,
   val format: DataFormat = DataFormat.NCHW
-)(implicit ev: TensorNumeric[T]) extends TensorModule[T] with Initializable {
+)(implicit ev: TensorNumeric[T]) extends TensorModule[T] with Initializable
+ {
```
hkvision: Need to change this file?
Are you moving compute output shape to new layers?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialFullConvolution.scala
```diff
@@ -86,7 +86,7 @@ class SpatialFullConvolution[T: ClassTag](
   var wRegularizer: Regularizer[T] = null,
   var bRegularizer: Regularizer[T] = null
   )(implicit ev: TensorNumeric[T])
-  extends AbstractModule[Activity, Tensor[T], T] with Initializable {
+  extends AbstractModule[Activity, Tensor[T], T] with Initializable{
```
hkvision: ??

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/UpSampling1D.scala
```diff
@@ -18,7 +18,7 @@ package com.intel.analytics.bigdl.nn
 
 import java.util
 
-import com.intel.analytics.bigdl.nn.abstractnn.TensorModule
+import com.intel.analytics.bigdl.nn.abstractnn.{TensorModule}
```
hkvision: ??

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/UpSampling3D.scala
```diff
@@ -16,7 +16,7 @@
 
 package com.intel.analytics.bigdl.nn
 
-import com.intel.analytics.bigdl.nn.abstractnn.TensorModule
+import com.intel.analytics.bigdl.nn.abstractnn.{TensorModule}
```
hkvision: ??

hkvision:
For output shape in old layers, are you moving to the new layers? Currently seems that only activation layers have been moved but not other layers such as SpatialConvolution, SpatialDilatedConvolution, etc.

hkvision:
If shape inference is only for new layers, we should move `InferShape` and `IdentityOutputShape` under the folder `nn/keras`

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasUtils.scala
```diff
@@ -17,13 +17,14 @@
...
-import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, DataFormat}
+import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, Activity, DataFormat}
 import com.intel.analytics.bigdl.tensor.Tensor
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
 
+import scala.collection.mutable
 import scala.reflect.ClassTag
 
-object KerasUtils {
+private[keras] object KerasUtils {
```
hkvision: private[bigdl]? As some utilities will be used in PythonKerasBigDL

zhichao-li:
The existing `computeOutputShape` in torch layer would not be touched in this PR.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/InferShape.scala
zhichao-li: Would check `isBuilt()` to ensure a KerasLayer being built only once.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialConvolution.scala
zhichao-li: The existing `computOutputShape` would not be touched.

hkvision:
Since you don't touch existing compute outputshape logic, why are you moving `with IdentityShape` to new layers?

zhichao-li:
The idea is we would delete all useless code and only keep it if being in use. They can be added back if we want to support "InferShape" for Torch in the future.
For those layers like SpatialConvolution, SpatialDilatedConvolution, etc, the `calcOutputShape` still be used by keras. We can rewrite those in the keras layer in a separated PR if needed.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialConvolution.scala
hkvision: So for `SpatialConvolution.scala` and `SpatialDilatedConvolution.scala`, I don't think these files should be touched in this PR?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Activation.scala
```diff
@@ -37,15 +37,13 @@ import scala.reflect.ClassTag
 class Activation[T: ClassTag](
    val activation: String,
    val inputShape: Shape = null)(implicit ev: TensorNumeric[T])
-  extends KerasLayer[Tensor[T], Tensor[T], T](KerasLayer.addBatch(inputShape)) {
+  extends KerasLayer[Tensor[T], Tensor[T], T](KerasLayer.addBatch(inputShape))
+    with IdentityOutputShape{
```
hkvision: Add a space between Shape and {

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Activation.scala
```diff
@@ -37,15 +37,13 @@ import scala.reflect.ClassTag
...
 
   require(activation != null, "The name of an activation function as a string is required")
 
   override def doBuild(inputShape: Shape): AbstractModule[Tensor[T], Tensor[T], T] = {
-    val model = Sequential[T]()
-    model.add(InputLayer(inputShape = KerasLayer.removeBatch(inputShape)))
-    val layer = KerasUtils.getActivation(activation)
-    model.add(layer).asInstanceOf[AbstractModule[Tensor[T], Tensor[T], T]]
+    KerasUtils.getActivation(activation)
```
hkvision: This modification will cause a problem for softmax as softmax is considered as a new layer.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/LocallyConnected2D.scala
```diff
@@ -16,7 +16,8 @@
 
 package com.intel.analytics.bigdl.nn.keras
 
-import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, DataFormat, TensorModule}
+import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, DataFormat}
+import com.intel.analytics.bigdl.nn.{Container => TContainer, LocallyConnected2D => TLocallyConnected2D}
```
hkvision: This file should also be unchanged?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/InferShape.scala
jason-dai: What if a Model is added to a Sequential? checkWithCurrentInputShape is only used in KerasLayer

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Topology.scala
```diff
@@ -18,30 +18,24 @@ package com.intel.analytics.bigdl.nn.keras
...
+  extends StaticGraph[T](_inputs, _outputs, None) {
 
   this.inputShapeValue = Shape(inputs.map{n => n.element.getInputShape()}.toList)
 
-  this.outputShapeValue = Array(outputs.map{_.element.getOutputShape()}: _*)
-
-  isBuilt = true
-
-  override private[bigdl] def isCompatibleWithKeras(): Boolean = true
+  this.outputShapeValue = Shape(outputs.map{_.element.getOutputShape()}.toList)
```
jason-dai: What if inputs or outputs are not Keras style?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Topology.scala
```diff
@@ -18,30 +18,24 @@ package com.intel.analytics.bigdl.nn.keras
...
 
-  override private[bigdl] def isCompatibleWithTorch(): Boolean = false
+  override private[bigdl] def isKerasStyle(): Boolean = true
 
   override def computeOutputShape(inputShape: Shape): Shape = {
```
jason-dai: What if inputShape is not compatible with this.inputShapeValue?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Topology.scala
```diff
@@ -114,38 +108,27 @@ trait ModelSerializer extends GraphSerializable with TKerasSerializerHelper{
...
-class Sequential[T: ClassTag](val stopInferShape: Boolean = false)
-(implicit ev: TensorNumeric[T]) extends TSequential[T] {
-
-  override private[bigdl] def isCompatibleWithKeras(): Boolean = true
-
-  override private[bigdl] def isCompatibleWithTorch(): Boolean = false
+class Sequential[T: ClassTag]()
+(implicit ev: TensorNumeric[T]) extends KerasLayer[Activity, Activity, T] {
 
   private[bigdl] var frozen: Boolean = false
```
jason-dai: Is frozen same as used?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasLayer.scala
```diff
@@ -149,33 +153,52 @@ abstract class KerasLayer[A <: Activity: ClassTag, B <: Activity: ClassTag, T: C
...
-  override def build(inputShape: Shape): Shape = {
-    this.labor = doBuild(inputShape)
-    val outputShape = computeOutputShape(inputShape)
-    this.outputShapeValue ++= Array(outputShape)
-    this.inputShapeValue = inputShape
-    isBuilt = true
-    outputShape // we cannot use getOutputShape here as it may containing multiple value
+  override def build(calcInputShape: Shape): Shape = {
+    this match {
+      case ks: com.intel.analytics.bigdl.nn.keras.Sequential[T] =>
```
jason-dai: Why don't we just override build() method in keras.Sequential?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasLayer.scala
```diff
@@ -200,12 +224,24 @@ abstract class KerasLayer[A <: Activity: ClassTag, B <: Activity: ClassTag, T: C
    * @return node containing current module
    */
   override def inputs(nodes : Array[ModuleNode[T]]): ModuleNode[T] = {
-    Util.excludeNotKeras(nodes.map(_.element))
-    if (!nodes.isEmpty) { // as there's  Identity().inputs() within Graph
+    validateInput(nodes.map(_.element))
```
jason-dai: What if we have:
```
l1 = n.inputs(n1)
l2 = n.inputs(n2)
```

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasLayer.scala
```diff
@@ -149,33 +153,52 @@ abstract class KerasLayer[A <: Activity: ClassTag, B <: Activity: ClassTag, T: C
...
+  override def isKerasStyle(): Boolean = true
 
-  override def getInputShape(): Shape = {
-    if (batchInputShape != null) {
-      batchInputShape
-    } else if (this.labor == null) {
-      null
-    } else {
-      this.labor.getInputShape()
+  override def computeOutputShape(inputShape: Shape): Shape = {
```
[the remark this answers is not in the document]
zhichao-li: In the previously design, we've implement lots of `computeOutputShape` in torch layers in case we want to support "InferShape" for Torch.
This is just a default implementation, the new added Keras layer can depend on this if it meets the requirement or override it with it's own logic.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Topology.scala
zhichao-li: There's a hooker: `checkWithCurrentInputShape` for this
``` scala
  private def checkWithCurrentInputShape(calcInputShape: Shape): Unit = {
    if (getInputShape() != null) {
      val withoutBatchInputShape = KerasLayer.removeBatch(getInputShape())
      val withoutBatchCalcInputShape = KerasLayer.removeBatch(calcInputShape)
      require(withoutBatchInputShape == withoutBatchCalcInputShape,
        s"InputShape from constructor ${withoutBatchInputShape}" +
          s"should be the same with the calculated inputShape: ${withoutBatchCalcInputShape}")
    }
  }

  override def inferShape(calcInputShape: Shape): Shape = {
    checkWithCurrentInputShape(calcInputShape)
    super.inferShape(calcInputShape)
  }
```

zhichao-li:
There's a problem once we disable the mix useage of torch and keras style layer as currently we mix the Activation (from torch or from keras) into a Keras Sequential to wrap the Layer like: `Dense(10, activation="relu")`.  
Maybe we can either 1)rewrite the keras activation into torch style or 2)We provide a utility to wrap a torch layer into keras and depend on Keras Sequential only.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Topology.scala
zhichao-li: Would add unittest for that.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasLayer.scala
zhichao-li: Humm, good catch. It's more about how to prevent stopping "shareWeights". Adding another flag to count? @yiheng any idea?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Topology.scala
jason-dai: These are defined in KerasLayer, but Model is not a KerasLayer

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasLayer.scala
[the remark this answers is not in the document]
jason-dai: It's not clear to me if the default behavior is correct

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasLayer.scala
jason-dai: Maybe we can have both usedAsSrc and usedAsDest or somthing similar.

jason-dai:
Just have a method in KerasLayer to wrap Torch activations?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Topology.scala
jason-dai: You can check if other layers are already used when they are added to the Sequential, yes?

zhichao-li:
Would introduce a `KerasLayerWrapper` and we should always use KSequential in `doBuild` method to address this issue. Ideally, it can wrap any Torch layer into Keras style. :
```scala
class KerasLayerWrapper[A <: Activity, B <: Activity, T: ClassTag]
(torchLayer: AbstractModule[Activity, Activity, T],
    inputShape: Shape)(implicit ev: TensorNumeric[T])
  extends KerasLayer[Activity, Activity, T](KerasLayer.addBatch(inputShape)) {
```

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasLayer.scala
zhichao-li: ok

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Topology.scala
zhichao-li: Let's think of bellowing scenario:
```
model = Sequential()
seq = Sequential()
model.add(seq)
seq.add(Dense(10)) // we should throw exception here, but "isUsed" is false for Dense(10) 
```

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasLayer.scala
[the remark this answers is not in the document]
zhichao-li: would simplify this logic once we introduce `KerasLayerWrapper`

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/LocallyConnected2D.scala
zhichao-li: I think it's just delete the useless imnport

hkvision: Yes...

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/StaticGraph.scala
```diff
@@ -35,19 +35,16 @@ import scala.reflect.ClassTag
...
   private val inputCache = new Array[Activity](forwardExecution.length)
   private var backId2ForwardId: Array[Int] = _
   private var gradOutputCache: Array[Activity] = _
 
-  if (excludeKeras) {
-    Util.excludeNotTorch(inputs.map(_.element))
-    Util.excludeNotTorch(outputs.map(_.element))
-  }
+  // StaticGraph would append Identity, so we would need to ignore it here.
+  excludeInvalidLayers(forwardExecution.map {_.element}.filter{!_.isInstanceOf[Identity[T]]})
```
jason-dai: Why to filter out Identity here? It is not Keras style, yes?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Container.scala
```diff
@@ -45,11 +44,7 @@ abstract class Container[A <: Activity : ClassTag,
   val modules: ArrayBuffer[AbstractModule[Activity, Activity, T]]
   = ArrayBuffer[AbstractModule[Activity, Activity, T]]()
 
-  override private[bigdl] def isCompatibleWithKeras(): Boolean = false
-
-  override private[bigdl] def isCompatibleWithTorch(): Boolean = {
-    modules.filter(!_.isCompatibleWithTorch()).length <= 0
-  }
+  override private[bigdl] def isKerasStyle(): Boolean = false
```
jason-dai: no need to override, yes?

jason-dai:
Some general comments for the latest code below; see other comments inline.
1.	allowRebuilt -> we should disallow rebuilt except for Keras Model and Sequential

2.	build vs. inferShape -> only one of them need to be overridden; it’s somewhat confusing if both get overridden

3.	Keras Model and Sequential will need similar implementations for build(), allowRebuilt(), inputs(), etc. Maybe we should have some class hierarchy like:
       * Model/Squential <-Topology <- KerasStyle <- InferShape
       * Keraslayer <- KerasStyle <- InferShape

4.	Keraslayer.inputs() and Sequential.add() methods have different semantics and shouldn’t use the same ValidateInput 

5.	`KerasIdentityActivationWrapper` is somewhat complex. We can just have a simple `KerasWrapper`, and have `KerasIdentityWrapper` to extend `KerasWrapper`; `KerasIdentityActivationWrapper` can be simply replaced by KSequential(KerasWrapper(layer), activation)

spark/dl/src/main/resources/serialization/bigdl.proto
```diff
@@ -16,7 +16,7 @@ message BigDLModule
     string namePostfix = 11; // name post fix
     int32 id = 12; // unique ID of this module , used for shared modules
     Shape inputShape = 13; // input shape
-    repeated Shape outputShape = 14; //output shape
```
jason-dai: both inputShape and outputShape should be optional (default to null) to ensure backward compatibility?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/DynamicContainer.scala
```diff
@@ -41,7 +41,7 @@ abstract class DynamicContainer[A <: Activity : ClassTag, B <: Activity : ClassT
     require(!module.isInstanceOf[Operation[_, _, _]],
       "Add operations to dynamic container is not allowed, as operations don't have backward. " +
         "Operation can only be used in Graph")
-    Util.excludeNotTorch[T](Seq(module))
+    validateInput[T](Seq(module))
```
jason-dai: validateInput is correct for inputs(), but not for container.add()

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/InferShape.scala
```diff
@@ -16,79 +16,119 @@
...
    * Execute building logic and return the outputShape for the given inputShape.
    * NB: the first dim of inputShape is batch
    */
   private[bigdl] def build(inputShape: Shape): Shape = {
-    val outputShape = computeOutputShape(inputShape)
-    this._outputShapeValue ++ Array(outputShape)
-    this._inputShapeValue = inputShape
-    isBuilt = true
-    outputShape
+    inferShape(inputShape)
```
jason-dai: Which of the two methods, `build` and `inferShape`, is the subclass expected to override? We should avoid overriding both of them.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/InferShape.scala
```diff
@@ -16,79 +16,119 @@
...
 
-  private[bigdl] var isBuilt: Boolean = false
-
+  private[bigdl] def isBuilt(): Boolean = outputShapeValue != null
 
-  private[bigdl] def isCompatibleWithKeras(): Boolean = true
+  private[bigdl] def isKerasStyle(): Boolean = false
 
-  private[bigdl] def isCompatibleWithTorch(): Boolean = true
+  private[bigdl] def allowRebuilt(): Boolean = false
```
jason-dai: Only Keras Model and Sequential should set this to true, yes?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/InferShape.scala
```diff
@@ -16,79 +16,119 @@
...
 
   /**
    * We suppose the first dim is batch
    */
   private[bigdl] def computeOutputShape(inputShape: Shape): Shape = {
     throw new RuntimeException("Haven't been implemented yet. Do not use it with Keras Layer")
   }
+
+  private def ensureNotShared[T: ClassTag](modules : Seq[AbstractModule[_, _, T]]): Unit = {
```
jason-dai: This is to for using this as dest, and modules as src - correct for `inputs()`, and not current for `Sequential.add()`

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/InferShape.scala
```diff
@@ -16,79 +16,119 @@
...
+
+  private[bigdl] def excludeInvalidLayers[T: ClassTag]
+  (modules : Seq[AbstractModule[_, _, T]]): Unit = {
+    val invalidNodes = if (this.isKerasStyle()) {
+      modules.filter{!_.isKerasStyle()}
+    } else {
+      modules.filter{_.isKerasStyle()}
+    }
+    if (invalidNodes.length > 0) {
+      throw new InvalidLayer(s"Do not mix with Layer: ${invalidNodes.mkString(",")}")
```
jason-dai: Give a better error message

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/AtrousConvolution1D.scala
```diff
@@ -93,7 +93,7 @@ class AtrousConvolution1D[T: ClassTag](
     model.add(Transpose(Array((2, 3))))
     model.add(Squeeze(4))
     if (activation != null) {
-      model.add(activation)
+      model.add(activation.doBuild(inputShape))
```
jason-dai: Do we need this? Shouldn't `TSequential.add()` handle the activation building?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Topology.scala
```diff
@@ -18,30 +18,24 @@ package com.intel.analytics.bigdl.nn.keras
...
+import scala.collection.JavaConverters._
 import scala.reflect.ClassTag
 
 class Model[T: ClassTag](private val _inputs : Seq[ModuleNode[T]],
       private val _outputs : Seq[ModuleNode[T]])(implicit ev: TensorNumeric[T])
-  extends StaticGraph[T](_inputs, _outputs, None, false) {
-
-  Util.excludeNotKeras(inputs.map(_.element))
-  Util.excludeNotKeras(outputs.map(_.element))
+  extends StaticGraph[T](_inputs, _outputs, None) {
```
jason-dai: We need the implementation of `allowRebuild`, `build`, etc. for Model too

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Topology.scala
```diff
@@ -154,33 +137,37 @@ class Sequential[T: ClassTag](val stopInferShape: Boolean = false)
...
     }
     if (module.isInstanceOf[Sequential[T]]) {
       module.asInstanceOf[Sequential[T]].frozen = true
     }
-    Util.excludeNotKeras[T](Seq(module))
-    if (!stopInferShape) {
-      triggerBuilding(module)
-    }
-    modules += module.asInstanceOf[AbstractModule[Activity, Activity, T]]
+    validateInput[T](Seq(module))
```
jason-dai: validateInput is not correct for `add`

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Topology.scala
```diff
@@ -154,33 +137,37 @@ class Sequential[T: ClassTag](val stopInferShape: Boolean = false)
...
+
+    triggerBuilding(module)
+
+    labor.asInstanceOf[TSequential[T]].modules +=
+      module.asInstanceOf[AbstractModule[Activity, Activity, T]]
     this
   }
+
+  override def computeOutputShape(inputShape: Shape): Shape = {
```
jason-dai: Will this be ever used?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasLayer.scala
```diff
@@ -51,23 +67,94 @@ object KerasLayerSerializer extends ContainerSerializable with TKerasSerializerH
   }
 }
 
+private[bigdl] class KerasIdentityActivationWrapper[T: ClassTag]
```
jason-dai: This is somewhat complex. We just need a KerasWrapper, and this can be implemented as KSequential(KerasWrapper(layer), activation), yes?

KerasIdentityWrapper can simply extend KerasWrapper.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasLayer.scala
```diff
@@ -51,23 +67,94 @@ object KerasLayerSerializer extends ContainerSerializable with TKerasSerializerH
...
+  extends KerasLayer[Activity, Activity, T](null) {
+  if (layer.isKerasStyle()) {
+    throw new RuntimeException(s"We only accept torch layer here, but got: $layer")
+  }
+  override def computeOutputShape(inputShape: Shape): Shape = {
+    inputShape
+  }
+  override def doBuild(inputShape: Shape): AbstractModule[Activity, Activity, T] = layer
+
+  override def allowRebuilt(): Boolean = true
```
jason-dai: Why should we allow rebuilt here? No reason to do that.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasLayer.scala
```diff
@@ -51,23 +67,94 @@ object KerasLayerSerializer extends ContainerSerializable with TKerasSerializerH
...
+    val dummyOutTensor =
+      torchLayer.forward(Tensor[T](
+        (List(2) ++ KerasLayer.removeBatch(calcInputShape).toSingle()).toArray).rand())
+    val outSize = dummyOutTensor.toTensor.size()
+    KerasLayer.addBatch(Shape(outSize.slice(1, outSize.length)))
+  }
+
+  override def doBuild(inputShape: Shape): AbstractModule[Activity, Activity, T] = torchLayer
+
+  override def allowRebuilt(): Boolean = true
```
jason-dai: Why do we need to support rebuilt here?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasLayer.scala
```diff
@@ -149,35 +223,50 @@ abstract class KerasLayer[A <: Activity: ClassTag, B <: Activity: ClassTag, T: C
...
+      val withoutBatchCalcInputShape = KerasLayer.removeBatch(calcInputShape)
+      require(withoutBatchInputShape == withoutBatchCalcInputShape,
+        s"InputShape from constructor ${withoutBatchInputShape}" +
+          s"should be the same with the calculated inputShape: ${withoutBatchCalcInputShape}")
+    }
   }
 
-  override def getOutputShape(): Shape = labor.getOutputShape()
+  override def inferShape(calcInputShape: Shape): Shape = {
+    checkWithCurrentInputShape(calcInputShape)
```
jason-dai: Shall we check in `build` or `inferShape`?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasUtils.scala
```diff
@@ -46,15 +47,28 @@ object KerasUtils {
     }
   }
 
-  private[keras] def getActivation[T : ClassTag] (activation: String)
+  private[bigdl] def getKerasActivation[T : ClassTag] (activation: String)
+    (implicit ev: TensorNumeric[T]): KerasLayer[Tensor[T], Tensor[T], T] = {
+    if (activation == null) { return null }
+    if (activation.toLowerCase() == "softmax") {
```
jason-dai: what's the difference between Keras softmax and Torch softmax?

hkvision: @jason-dai Torch Softmax is different from Keras Softmax for __3D__ input. They will have the same result for 2D input. https://github.com/intel-analytics/BigDL/issues/1934 
For 3D, currently to be compatible, we first do a transpose before applying softmax.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/InferShape.scala
```diff
@@ -16,79 +16,119 @@
...
 
   /**
    * We suppose the first dim is batch
    */
   private[bigdl] def computeOutputShape(inputShape: Shape): Shape = {
     throw new RuntimeException("Haven't been implemented yet. Do not use it with Keras Layer")
   }
+
+  private def ensureNotShared[T: ClassTag](modules : Seq[AbstractModule[_, _, T]]): Unit = {
+    def validateUsage(module: InferShape): Unit = {
```
hkvision: This is an empty method? What's the purpose?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/InferShape.scala
```diff
@@ -16,79 +16,119 @@
...
-    outputShape
+    inferShape(inputShape)
   }
 
-  private[bigdl] var isBuilt: Boolean = false
-
+  private[bigdl] def isBuilt(): Boolean = outputShapeValue != null
 
-  private[bigdl] def isCompatibleWithKeras(): Boolean = true
+  private[bigdl] def isKerasStyle(): Boolean = false
```
hkvision: Shouldn't `isKerasStyle` and `allowRebuilt` be a field not a method? ie `isKerasStyle()` => `isKerasStyle`

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/InferShape.scala
```diff
@@ -16,79 +16,119 @@
...
+      modules.filter{_.isKerasStyle()}
+    }
+    if (invalidNodes.length > 0) {
+      throw new InvalidLayer(s"Do not mix with Layer: ${invalidNodes.mkString(",")}")
+    }
+  }
+
+  private[bigdl] def validateInput[T: ClassTag](modules : Seq[AbstractModule[_, _, T]]): Unit = {
+    if (this.isKerasStyle()) {
+      require(modules != null && !modules.isEmpty, "Empty input is not allow")
```
hkvision: `allow` -> `allowed`

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Linear.scala
```diff
@@ -51,7 +51,8 @@ class Linear[T: ClassTag](
   private val initBias: Tensor[T] = null,
   private val initGradWeight: Tensor[T] = null,
   private val initGradBias: Tensor[T] = null
-)(implicit ev: TensorNumeric[T]) extends TensorModule[T] with Initializable {
+)(implicit ev: TensorNumeric[T]) extends TensorModule[T] with Initializable
+ {
```
hkvision: ??

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialDilatedConvolution.scala
```diff
@@ -68,7 +68,8 @@ class SpatialDilatedConvolution[T: ClassTag](
   val dilationH: Int = 1,
   var wRegularizer: Regularizer[T] = null,
   var bRegularizer: Regularizer[T] = null
-)(implicit ev: TensorNumeric[T]) extends TensorModule[T] with Initializable {
+)(implicit ev: TensorNumeric[T]) extends TensorModule[T] with Initializable
+ {
```
hkvision: ??

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialFullConvolution.scala
```diff
@@ -18,15 +18,13 @@ package com.intel.analytics.bigdl.nn
...
 import com.intel.analytics.bigdl.optim.Regularizer
+import com.intel.analytics.bigdl.serialization.Bigdl.{AttrValue, BigDLModule}
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
 import com.intel.analytics.bigdl.tensor._
-import com.intel.analytics.bigdl.utils.{Shape, T, Table, serializer}
-import com.intel.analytics.bigdl.utils.RandomGenerator._
 import com.intel.analytics.bigdl.utils.serializer._
 import com.intel.analytics.bigdl.utils.serializer.converters.DataConverter
-import com.intel.analytics.bigdl.serialization.Bigdl.{AttrValue, BigDLModule}
+import com.intel.analytics.bigdl.utils.{Shape, T, Table}
```
hkvision: ??

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasLayer.scala
```diff
@@ -18,17 +18,19 @@ package com.intel.analytics.bigdl.nn.keras
 
 import com.intel.analytics.bigdl._
 import com.intel.analytics.bigdl.nn.Graph._
-import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, Activity, TensorModule}
+import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, Activity}
+import com.intel.analytics.bigdl.nn.keras.Sequential._copyWeightAndBias
```
hkvision: is copyweightandbias used?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/LocallyConnected2D.scala
```diff
@@ -72,6 +73,7 @@ class LocallyConnected2D[T: ClassTag](
   require(subsample.length == 2,
     s"For LocallyConnected2D, subsample should be of length 2 but got length ${subsample.length}")
 
+
```
hkvision: remove

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/SimpleRNN.scala
```diff
@@ -16,7 +16,7 @@
 
 package com.intel.analytics.bigdl.nn.keras
 
-import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, TensorModule}
+import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, Activity, TensorModule}
```
hkvision: No need to import AbstractModule and Activity now?

hkvision:
Can we add validation after computing outputshape that require the computed outputshape all entries >0?
Also, for those files that ought not to be touched such as `Linear`, `SpatialDilatedConvolution`, `UpSampling`, etc, please keep them unchanged.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/InferShape.scala
zhichao-li: `Model` would be inherited from `KerasLayer` same as `KSequential`

spark/dl/src/main/resources/serialization/bigdl.proto
zhichao-li: We just remove "repeated" here and it would be backward compatible with the old model. cc @wzhongyuan

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/InferShape.scala
zhichao-li: In general, is there's no overhead for building then we can mark it as "allowRebuilt" . ie: `KerasIdentityWrapper, Model, Sequential, Input`

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/InferShape.scala
zhichao-li: would remove `inferShape`

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/InferShape.scala
zhichao-li: would be removed.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasLayer.scala
zhichao-li: There's a little bit overhead for `KerasLayerWrapper ` as it would need to `forward` once to get the `outputshape` although ideally it can wrap any Torch Layer.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasLayer.scala
zhichao-li: `doBuild` just simply return parameter `layer`. In other words, there's no overhead to invoke it multiple times, I think it's safe to allow this.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasLayer.scala
zhichao-li: would do that same as `Model`

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Topology.scala
zhichao-li: done

spark/dl/src/main/resources/serialization/bigdl.proto
wzhongyuan: @jason-dai   it's by default optional in proto3

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasLayer.scala
zhichao-li: There's no overhead for rebuilding this layer.

zhichao-li: `KerasLayerWrapper`  is used to combine activity and layer and it wold be built within `doBuild` method, so we need to enable `allowRebuilt` here.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Topology.scala
zhichao-li: It would be used when `labor` of a KerasLayer is `KSequential` as the default implementation for `computeOutputShape` is `labor.computeOutputShape`.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasLayer.scala
```diff
@@ -51,23 +66,80 @@ object KerasLayerSerializer extends ContainerSerializable with TKerasSerializerH
...
+ * Wrap a torch style layer to keras style layer.
+ * This layer can be built multiple times.
+ * @param torchLayer a torch style layer
+ *   i.e If the input data is (2, 3, 4) and 2 is the batch size, you should input: (3, 4) here.
+ * @return a keras compatible layer
+ */
+class KerasLayerWrapper[T: ClassTag]
+(val torchLayer: AbstractModule[Activity, Activity, T],
+    val inputShape: Shape = null,
+    computeOutputShapeFunc: Option[(Shape => Shape)] = None)(implicit ev: TensorNumeric[T])
```
zhichao-li: @jason-dai function cannot be serialized to `proto` for now, so we may need to think of another way to wrap the activation.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Topology.scala
jason-dai: `labor` can be a torch layer (e.g., Linear for Dense), and has no `computeOutputShape` or `inferShape` deined, yes?

In addition, it's somewhat confusing to call `InferShape` in `KerasLayer.build`: `InferShape` should be called when the output shape needs to be inferred; however, for `KerasLayer`, its labor is already built, and therefore it should call `labor.getOutputShape`  instead.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasLayer.scala
jason-dai: There can be three different wrappers:
1) Identity shape
2) Torch layer w/ no computeOutputshape - just call forward
3) Torch layer w/ computeOutputshape defined

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Topology.scala
```diff
@@ -18,37 +18,55 @@ package com.intel.analytics.bigdl.nn.keras
...
+import com.intel.analytics.bigdl.utils.Shape
 import com.intel.analytics.bigdl.utils.serializer._
-import com.intel.analytics.bigdl.utils.{Shape, Util}
 
+import scala.collection.JavaConverters._
+import scala.collection.mutable.ArrayBuffer
 import scala.reflect.ClassTag
 
+abstract class KerasModel[T: ClassTag](implicit ev: TensorNumeric[T])
+  extends KerasLayer[Activity, Activity, T] {
```
hkvision: where to put Trainable?

hkvision:
For those files that ought not to be touched such as `Linear`, `SpatialDilatedConvolution`, `UpSampling`, etc, please keep them unchanged.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Topology.scala
zhichao-li: fixed in: https://github.com/intel-analytics/BigDL/pull/2293/files#diff-887accbfc37525584b59f19718396065R97

zhichao-li:
@jason-dai  jenkins passed http://172.168.2.101:8080/job/BigDL-PR-Validation-v3/362/

PR #2294, opened 2018-02-09
add missing link

wzhongyuan:

add doc link for serialization

## How was this patch tested?
review

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -48,8 +48,8 @@ pages:
   - Keras Support: ProgrammingGuide/keras-support.md
   - Quantization Support: ProgrammingGuide/quantization-support.md
   - Define Customized Layer/Criterion:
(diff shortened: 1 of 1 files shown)


wzhongyuan:
@yiheng  please review, thanks

docs/mkdocs.yml
```diff
@@ -37,6 +37,7 @@ pages:
   - Define a Model:
     - Using Sequential API: ProgrammingGuide/Model/Sequential.md
     - Using Functional API: ProgrammingGuide/Model/Functional.md
+    - Implement customized serializer: ProgrammingGuide/Model/Serialization.md
```
yiheng: I don't think we should put the link here. It has nothing to do with how you define a model. I think we should put it together with ```add-layer-criterion```. Maybe create a group named "Define Customized Layer/Criterion" to hold them.

wzhongyuan:
@yiheng  makes sense, refined

PR #2296, opened 2018-02-11
Keras-like API Python Wrapper for layers + lenet example

hkvision:

Wrap all keras layers to python side with doc and ut.
Wrap graph, get input output shape.
Some fix in doc format or wordings. Fix SReLU missing parameters.
LeNet model using the new API. Refine duplicate code of lenet examples.

## How was this patch tested?

Jenkins


pyspark/bigdl/examples/lenet/lenet.py
```diff
@@ -0,0 +1,114 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from optparse import OptionParser
+from bigdl.dataset import mnist
+from bigdl.dataset.transformer import *
+from bigdl.nn.keras.layer import *
+from bigdl.nn.layer import Model
```
zhichao-li: remove nn.layer.Model?

hkvision: It is used in test for model loading https://github.com/intel-analytics/BigDL/pull/2296/files/c43ca0927cae8a33ff207aeeb67adaba55813354#diff-bbc826873e146fd0ef3de8f1cf5aa024R110
Or we can first remove test option. @zhichao-li

zhichao-li:
There's already a lenet5.py under "models/lenet" with model definition, training and prediction.
We should think of a good way to avoid code duplication as the only difference here is the `buildModel` method.

hkvision:
jenkins passed http://172.168.2.101:8080/job/BigDL-PR-Validation-v3/227/

hkvision:
I have extracted some common methods for lenet to avoid code duplication. Please have a look @zhichao-li

pyspark/bigdl/models/lenet/utils.py
```diff
@@ -0,0 +1,68 @@
...
+    :return: RDD of (features: ndarray, label: ndarray)
+    """
+    (images, labels) = mnist.read_data_sets(location, data_type)
+    images = sc.parallelize(images)
+    labels = sc.parallelize(labels + 1) # Target start from 1 in BigDL
+    record = images.zip(labels)
+    return record
+
+
+def process_mnist_rdd(sc, options):
```
zhichao-li: `get_mnist_rdd`?

zhichao-li:
Let's update the other LeNet example to remove duplication base on the `utils` as well

pyspark/bigdl/models/lenet/utils.py
hkvision: But in the above method `get_mnist` it already returns an RDD and this function does normalization and turn into Sample RDD...

pyspark/bigdl/nn/keras/layer.py
```diff
@@ -16,33 +16,68 @@
...
 
+class InferShape(JavaValue):
+    def __init__(self, bigdl_type="float"):
+        self.bigdl_type = bigdl_type
+
+    @classmethod
+    def __process_shape(cls, output_shape):
+        return tuple([None] + output_shape[1:])
+
+    def get_input_shape(self):
```
zhichao-li: unitest for this two methods?

pyspark/bigdl/nn/keras/layer.py
```diff
@@ -16,33 +16,68 @@
...
+                              self.value)
+        if len(input) == 1:
+            return self.__process_shape(input(0))
+        else:
+            res = []
+            for i in input:
+                res.append(self.__process_shape(i))
+            return res
+
+    def get_output_shape(self):
```
zhichao-li: add doc

pyspark/bigdl/nn/keras/layer.py
```diff
@@ -135,17 +238,1683 @@ def get_running_std(self):
...
+            raise ValueError("For ConvLSTM2D, only border_mode='same' is supported for now")
+        if subsample[0] != subsample[1]:
+            raise ValueError("For ConvLSTM2D, only equal strides is supported for now")
+        super(ConvLSTM2D, self).__init__(None, bigdl_type,
+                                         nb_filter,
+                                         nb_row,
+                                         activation,
+                                         inner_activation,
+                                         dim_ordering,
+
```
zhichao-li: `alpha` should be wrap with float: `float(alpha)`. .i.e if user input `ELU(1)`, there would be "method xxx not found" error

pyspark/bigdl/nn/keras/layer.py
hkvision: Added in `test_newapi.py`

pyspark/bigdl/nn/keras/layer.py
hkvision: Added.

hkvision:
@zhichao-li I have fixed all the appearances of the float issue.

hkvision:
@zhichao-li lenet examples refined.

pyspark/bigdl/models/local_lenet/local_lenet.py
```diff
@@ -67,6 +49,7 @@ def get_mnist(data_type="train", location="/tmp/mnist"):
     (X_train, Y_train) = get_mnist("train", options.dataPath)
```
zhichao-li: `get_mnist` should return NDarray here same as the previous logic.

spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDLKeras.scala
```diff
@@ -56,11 +57,34 @@ class PythonBigDLKeras[T: ClassTag](implicit ev: TensorNumeric[T]) extends Pytho
...
+      list.asScala.toArray
+    }
+  }
+
   def createKerasInputLayer(
     inputShape: JList[Int] = null): Input[T] = {
     InputLayer(inputShape = toScalaShape(inputShape))
   }
 
+  def getOutputShape(module: Container[Activity, Activity, T]): JList[Int] = {
```
zhichao-li: There's possible that `outputshape` is MultiShape.

hkvision:
jenkins passed http://172.168.2.101:8080/job/BigDL-PR-Validation-v3/228/

spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDLKeras.scala
```diff
@@ -56,11 +57,34 @@ class PythonBigDLKeras[T: ClassTag](implicit ev: TensorNumeric[T]) extends Pytho
...
+
+  def getInputShape(module: Container[Activity, Activity, T]): JList[JList[Int]] = {
+    val input = module.getInputShape()
+    val inputs = if (input.isInstanceOf[SingleShape]) {
+      MultiShape(List(input))
+    }
+    else {
+      input
+    }
+    inputs.toMulti().map(single => single.toSingle().toList.asJava).toList.asJava
```
zhichao-li: From the definition, nested `MultiShape` is possible, but simply process like this is OK for now. Let's add a TODO here.

spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDLKeras.scala
hkvision: In what case outputShape is a MultiShape?

pyspark/test/bigdl/keras/test_newapi.py
```diff
@@ -119,6 +119,14 @@ def test_merge_cos(self):
         input_data = [np.random.random([2, 3]), np.random.random([2, 3])]
         self.compare_newapi(klayer, blayer, input_data)
 
+    def test_input_output_shape(self):
+        from bigdl.examples.lenet.lenet import build_model
+        model = build_model(10)
+        input_shape = model.get_input_shape()
+        np.testing.assert_allclose((28, 28, 1), input_shape[1:])
+        output_shape = model.get_output_shape()
+        np.testing.assert_allclose((10, ), output_shape[1:])
```
zhichao-li: test for `MultiShape` output and `MultiShape` input?

pyspark/bigdl/nn/keras/layer.py
```diff
@@ -16,33 +16,71 @@
...
+        Return one shape tuple otherwise.
+        """
+        input = callBigDlFunc(self.bigdl_type, "getInputShape",
+                              self.value)
+        if len(input) == 1:
+            return self.__process_shape(input[0])
+        else:
+            res = []
+            for i in input:
+                res.append(self.__process_shape(i))
```
zhichao-li: return [self.__process_shape(i) for i in input]

hkvision:
http://172.168.2.101:8080/job/BigDL-PR-Validation-v3/236/ jenkins passed

hkvision:
jenkins passed http://172.168.2.101:8080/job/BigDL-PR-Validation-v3/237/

PR #2297, opened 2018-02-11
add [[IndicatorCol]] Operation

tosky001:

add the **Indicator** to convert the sparse tensor(eg. the output of CategoricalCol* Ops) to multi-hot dense tensor, which can be feed into DNN.

## How was this patch tested?

unit tests


tosky001:
@yiheng  may i have your review please and thanks.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/IndicatorCol.scala
```diff
@@ -0,0 +1,91 @@
...
+  val feaLen: Int,
+  val isCount: Boolean = true
+) (implicit ev: TensorNumeric[T])
+  extends Operation[Tensor[Int], Tensor[T], T]{
+
+  output = Tensor[T]()
+
+  override def updateOutput(input: Tensor[Int]): Tensor[T] = {
+
+    val rows = input.size(dim = 1)
```
yiheng: require(input.isInstanceOf[SparseTensor[_]], "Only sparse input is supported") ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/IndicatorCol.scala
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * The output tensor should be a DenseTensor with shape (batch, feaLen).
+ *
+ * For example, A input SparseTensor as follows:
+ *  indices(0) = Array(0, 0, 1, 2, 2)
+ *  indices(1) = Array(0, 3, 1, 1, 2)
+ *  values     = Array(1, 2, 2, 3, 3)
+ *  shape      = Array(3, 4)
+ *
+ *  the output tensor should be an 2D 3x4 DenseTensor with isCount = true
```
yiheng: I think here is a little misleading. Should we mention 3 x 4 result tensor is because we set feaLen = 4? BTw, in this case, what if we set feaLen = 2(some value cannot be mapped to multi-hot feature space)

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/IndicatorCol.scala
yiheng: do we need to require input is a 2D tensor? If so, should we mention this in the java doc of the class.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/IndicatorCol.scala
```diff
@@ -0,0 +1,91 @@
...
+    var i = 1
+    while (i <= rows) {
+      val narrowTensor = input.narrow(1, i, 1)
+      val tempArr = narrowTensor.storage().array().slice(
+        narrowTensor.storageOffset()-1, narrowTensor.storageOffset() - 1 + narrowTensor.nElement())
+      var j = 0
+      while (j < tempArr.length) {
+        isCount match {
+          case false =>
+            resTensor.setValue(i, tempArr(j) + 1, ev.fromType[Int](1))
```
yiheng: ev.fromType\[Int\](1) -> ev.one

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/IndicatorCol.scala
tosky001: well, this Operation is used after `CategoricalCol*` Ops. So the **setValue of feaLen** should **avoid** the case you mentioned. For example, if the Operation used after the `CategoricalColHashBucket`, then the **feaLen** should be set *bucketSize*. This is our solution by far. Looking forward your better advice and thanks.

tosky001:
@yiheng the Operation has been modified and  test successfully.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ops/IndicatorCol.scala
yiheng: what will happen if the user sets an incorrect value? For example, if the operation is used after the ```CategoricalColHashBucket```, and the user set the ```feaLen``` to bucketSize + 1 or bucketSize -1

tosky001: @yiheng well, if the user sets an incorrect value, it will not work. a alternative solution is that we can get the `feaLen` from the `input.size`, but we should ensure the size of input tensor correctly, which means the input from `CategoricalColHashBucket` should be a SparseTensor with shape(batch_size,**bucketSize**). Then the config parameter `feaLen` can be cancelled. So what's your opinion? Thanks.

yiheng: I think we can let user set the ```feaLen```.

If the ```feaLen``` > ```bucketSize```, looks like it won't affect the multi-hot representation, right? (just the extra bits of the multi-hot vector are always zero)

if the ```feaLen``` < ```bucketSize```, looks like exception will be thrown at line 71 or line 74. Should we check if the tempArr(j) > feaLen before set the value, and throw a more meaningfull error message there?

tosky001: @yiheng  that's a good idea and the **validation** of parameter `feaLen`, pelase check it and thanks.

yiheng:
Jenkins test pass and merge the PR. Thanks for the contribution:)

PR #2300, opened 2018-02-12
Prediction service

sperlingxx:

Add a prediction service interface for concurrent calls.

In this service, several `model instances` sharing weights/bias will be built.
And a `BlockingQueue` will be built to maintain available `model instances`.

When predict method called, service will try to take an instance from `BlockingQueue`,
which means if all instances are on serving, the predicting request will be blocked until
some instances are released.

The release method will be executed asynchronously when predictions finished.

## How was this patch tested?

unit test

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/PredictionService.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/PredictionService.scala
@@ -16,7 +16,7 @@
 
 package com.intel.analytics.bigdl.optim
 
-import java.util.concurrent.{Executors, LinkedBlockingQueue}
+import java.util.concurrent.LinkedBlockingQueue
 
 import com.intel.analytics.bigdl.Module
 import com.intel.analytics.bigdl.nn.abstractnn.Activity
@@ -31,9 +31,9 @@ import com.intel.analytics.bigdl.utils.{T, Table}
 
 import scala.collection.JavaConverters._
 import scala.collection.mutable
-import scala.concurrent.{ExecutionContext, ExecutionContextExecutor, Future}
 import scala.reflect.ClassTag
 import scala.reflect.runtime.universe.Type
+import scala.util.{Failure, Success, Try}
 
 /**
  * <h6>Prediction Service Interface for Concurrent Calls</h6>
@@ -45,27 +45,18 @@ import scala.reflect.runtime.universe.Type
  * which means if all instances are on serving, the predicting request will be blocked until
  * some instances are released.
  * <br>
- * The release method will be executed asynchronously when predictions finished.
+ * If exceptions caught during predict,
+ * a scalar Tensor[String] will be returned with thrown message.
  *
  * @param model BigDL model used to do predictions
  * @param nInstances number of model instances
- * @param nBackendThreads number of threads to release instances after predictions.
  */
 class PredictionService[T: ClassTag] private[optim](
   model: Module[T],
-  nInstances: Int = 10,
-  nBackendThreads: Int = 2
+  nInstances: Int = 10
 )(implicit ev: TensorNumeric[T]) {
 
   require(nInstances > 1, "At least two instances are needed!")
-
-  require(nBackendThreads >= 1, "At least one backend threads are needed!")
-
-  // ThreadPool for scala.concurrent.Future,
-  // which is build for keeping threads available for unlocking to avoid dead locks.
-  private implicit val ec: ExecutionContextExecutor = {
-    ExecutionContext.fromExecutor(Executors.newFixedThreadPool(nBackendThreads))
-  }
 
   protected val instQueue: LinkedBlockingQueue[Module[T]] = {
     val shallowCopies = (1 to nInstances)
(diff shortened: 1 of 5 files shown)


sperlingxx:
@yiheng 
Could you help reviewing this PR  :- )

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/PredictionService.scala
```diff
@@ -0,0 +1,258 @@
...
+
+  def predict(request: Activity): Activity = {
+    val module = fetchInstance()
+    // cloned values after prediction finished
+    val output = module.forward(request) match {
+      case tensor: Tensor[_] =>
+        tensor.clone()
+      case table: Table =>
+        val clonedMap = mutable.HashMap[Any, Any]()
+        table.getState().foreach { case (k, v) =>
```
qiuxin2012: what if output is a table of table?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/PredictionService.scala
```diff
@@ -0,0 +1,258 @@
...
+  // it will cause a thread blocking when no instance is available.
+  protected def fetchInstance(): Module[T] = {
+    instQueue.take()
+  }
+
+  // Asynchronously release module instance back to blocking queue.
+  protected def releaseInstance(inst: Module[T]): Future[Boolean] = {
+    Future {
+      val isSuccess = instQueue.offer(inst)
+      println(s"Release instance ${if (isSuccess) "successfully" else "failed"}!" +
```
qiuxin2012: use logger?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/PredictionService.scala
```diff
@@ -0,0 +1,258 @@
...
+    val shallowCopies = (1 to nInstances)
+      .map(_ => model.clone(false).evaluate()).asJava
+
+    new LinkedBlockingQueue[Module[T]](shallowCopies)
+  }
+
+  def predict(request: Activity): Activity = {
+    val module = fetchInstance()
+    // cloned values after prediction finished
+    val output = module.forward(request) match {
```
qiuxin2012: forward may throw exception if the request is illegal, please catch the exception and releaseInstance anyway.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/PredictionService.scala
```diff
@@ -0,0 +1,258 @@
...
+    bytesOut
+  }
+
+  // Take an instance from blocking queue,
+  // it will cause a thread blocking when no instance is available.
+  protected def fetchInstance(): Module[T] = {
+    instQueue.take()
+  }
+
+  // Asynchronously release module instance back to blocking queue.
```
qiuxin2012: Why not just use Synchronously release

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/PredictionService.scala
sperlingxx: Nested table is not supported so far.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/PredictionService.scala
sperlingxx: code pieces added to catch exceptions

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/PredictionService.scala
```diff
@@ -75,43 +66,55 @@ class PredictionService[T: ClassTag] private[optim](
...
+          }
         new Table(clonedMap)
+
+      case Failure(e) =>
+        errorTensor("running forward", e)
     }
-    releaseInstance(module)
+
+    // Release module instance back to blocking queue
+    instQueue.offer(module)
```
qiuxin2012: put this line into `finnally`?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/PredictionService.scala
```diff
@@ -121,10 +124,9 @@ object PredictionService {
 
   def apply[T: ClassTag](
     model: Module[T],
-    nInstances: Int = 10,
-    nBackendThreads: Int = 2
+    nInstances: Int = 10
```
qiuxin2012: why default is 10? Set default Engine.coreNumber?

sperlingxx: I think 10 looks like a reasonable number. In my opinion, while serving online, Engine.init() is unnecessary.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/PredictionService.scala
```diff
@@ -0,0 +1,274 @@
...
+  }
+
+  protected val instQueue: LinkedBlockingQueue[Module[T]] = {
+    val shallowCopies = (1 to numThreads)
+      .map(_ => model.clone(false).evaluate()).asJava
+
+    new LinkedBlockingQueue[Module[T]](shallowCopies)
+  }
+
+  def predict(request: Activity): Activity = {
```
qiuxin2012: All public method should have scala doc

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/PredictionService.scala
```diff
@@ -0,0 +1,274 @@
...
+      case e: Throwable => errorTensor("Clone Result", e)
+    } finally {
+      // Release module instance back to blocking queue
+      instQueue.offer(module)
+    }
+
+    output
+  }
+
+  def predict(request: Array[Byte]): Array[Byte] = {
```
qiuxin2012: All public method should have scala doc

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -35,6 +35,7 @@ sealed trait EngineType
 
 case object MklBlas extends EngineType
 
+case object LocalEngine extends EngineType
```
qiuxin2012: what's this? Engine type is the native libraries like mklblas and mkldnn...

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/PredictionService.scala
```diff
@@ -0,0 +1,274 @@
...
+ */
+class PredictionService[T: ClassTag] private[optim](
+  model: Module[T],
+  numThreads: Int
+)(implicit ev: TensorNumeric[T]) {
+
+  initThreads()
+
+  private def initThreads(): Unit = {
+    Engine.setEngineType(LocalEngine)
```
qiuxin2012: You can init an localMode Engine
```
def init: Unit = this.synchronized {
    if (localMode) {
      logger.info("Detect bigdl.localMode is set. Run workload without spark")
      // The physical core number should have been initialized
      // by java property -Dbigdl.coreNumber=xx
      setNodeAndCore(1, getCoreNumberFromProperty)
    } else {
      logger.info("Auto detect executor number and executor cores number")
      val (nExecutor, executorCores) = sparkExecutorAndCore().get
      logger.info(s"Executor number is $nExecutor and executor cores number is $executorCores")
      setNodeAndCore(nExecutor, executorCores)
      checkSparkContext
    }
  }
```

sperlingxx: If localMode, the modelPoolSize always equals to 1 when engineType == MKlBlas by defaults.

    private def initThreadPool(core : Int) : Unit = {
        (core * 50).toString).toInt
      }
  
      val modelPoolSize: Int = if (engineType == MklBlas) {
        1
      } else {
        core
      }
  
      if(_model == null || _model.getPoolSize != modelPoolSize) {
        _model = new ThreadPool(modelPoolSize)
        _model.setMKLThread(MKL.getMklNumThreads)
      }
    }

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
sperlingxx: As mentioned above, a EngineType other than _MklBlas_ has been built to make _modelPoolSize_ == _nCores_.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/PredictionService.scala
qiuxin2012: If modelPoolSize=1, will just run the model in current thread.

qiuxin2012:
jenkins passed

PR #2304, opened 2018-02-13
Fix a failure in nightly build

yiheng:
Nightly build always failed at some step recently, while PR validation runs well. The root cause is some TensorArray operations use a shared static object for output. And this output will be reset during graph building. This PR override the clearState method of these TensorArray operations

## How was this patch tested?
unit test


yiheng:
jenkins pass. merge

yiheng:
jenkins pass

PR #2324, opened 2018-02-24
Optimize padding mechanism

psyyz10:
Optimize padding mechanism
This PR make the paddings or zeros make no side effect on the training.

`LookupTable`: if `maskZero` is set to true, the input whose value equals `paddingValue`,  the output will be masked to zero vector.

`Recurrent`: The recurrent includes some mask mechanisms. if the `maskZero` variable is set to true, the `Recurrent` module will not consider zero vector inputs. For each time step input, if a certain row is a zero vector (all the elements of the vector equals zero), then output of certain row of this time step would be a zero vector, and the hidden state of the certain row of this time step would be the same as the corresponding row of the hidden state of the previous step.

`TimeDistributed': if `maskZero` is set to true, if the input including zero vectors, the corresponding output will be set to zero vecotrs.

`TimeDistributedMaskCriterion`: it supports padding mask. eg. if the target is [ [-1, 1, 2, 3, -1], [5, 4, 3, -1, -1] ],  and set the paddingValue property to -1, then the loss of -1 would not  be accumulated and the loss is only divided by 6 (not including the amount of -1, in this case, we are only interested in 1, 2, 3, 5, 4, 3)

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
@@ -72,11 +72,11 @@ class Recurrent[T : ClassTag](
   private val timeBuffer =
     new ArrayBuffer[(AbstractModule[_ <: Activity, _ <: Activity, T], Long, Long)]
   private var layer: TensorModule[T] = null
-  private var maskBuffer: Tensor[T] = _
-  private var gradOutputBuff: Table = _
-  private var indexBuffer: Tensor[T] = _
-  private var inputBuffer: Tensor[T] = _
-  private var outputBuffers: ArrayBuffer[Tensor[T]] = _
+  private var maskBuffer: Tensor[T] = Tensor()
+  private var gradOutputBuff: Table = T()
+  private var indexBuffer: Tensor[T] = Tensor()
+  private var inputBuffer: Tensor[T] = Tensor()
+  private var outputBuffers: ArrayBuffer[Tensor[T]] = ArrayBuffer(Tensor())
   private var minLength: Int = 0
 
   /**
@@ -166,26 +166,6 @@ class Recurrent[T : ClassTag](
     } else {
       cells.head.hidResize(hidden = hidden, batchSize = batchSize, stepShape)
       gradHidden = hidden
-    }
-
-    if (maskBuffer == null) {
-      maskBuffer = Tensor()
-    }
-
-    if (gradOutputBuff == null) {
-      gradOutputBuff = T()
-    }
-
-    if (indexBuffer == null) {
-      indexBuffer = Tensor()
-    }
-
-    if (inputBuffer == null) {
-      inputBuffer = Tensor()
-    }
-
-    if (outputBuffers == null) {
-      outputBuffers = ArrayBuffer(Tensor())
     }
   }
 
@@ -299,6 +279,9 @@ class Recurrent[T : ClassTag](
       if (maskZero && i > minLength) {
         val curMask = maskBuffer.select(2, i)
         val curOut = curOutput[Table](hidDim)[Tensor[T]](1)
+        // Copy output to a new new tensor as output, because for some cells
+        // such as LSTM the hidden h and ouput o refer to the same tensor.
+        // But in this case, we want h and o have difference values.
         curOutput.update(inputDim, outputBuffers(i - 1).resizeAs(curOut).copy(curOut))
         for (b <- 1 to curMask.size(1)) {
           if (curMask(Array(b, 1)) == ev.zero) {
@@ -460,6 +443,7 @@ class Recurrent[T : ClassTag](
 
   override def backward(input: Tensor[T], gradOutput: Tensor[T]): Tensor[T] = {
     val st = System.nanoTime
+    currentGradOutput(hidDim) = gradHidden
     var i = times
 
     while (i >= 1) {
@@ -505,6 +489,7 @@ class Recurrent[T : ClassTag](
       } else {
         cells(i - 1).backward(_input, currentGradOutput)
       }
+      currentGradOutput(hidDim) = cells(i - 1).gradInput.toTable(hidDim)
       i -= 1
     }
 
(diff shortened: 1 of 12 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TimeDistributedMaskCriterion.scala
```diff
@@ -0,0 +1,164 @@
...
+    mask.resizeAs(target)
+    mask.applyFun[T](target, x =>
+      if (x != ev.fromType[Int](padValue)) ev.one else ev.zero)
+    val outputBuff = Tensor()
+    outputBuff.resizeAs(mask)
+    (0 until nstep).foreach(b => {
+      //      output = ev.plus(output, cells(b).output)
+      outputBuff
+        .select(2, b + 1)
+        .cmul(mask.select(2, b + 1), cells(b).asInstanceOf[ClassNLLCriterion[T]].sparseOutput)
```
jason-dai: Why is cell always ClassNLLCriterion?

jason-dai:
Please add more details (in both PR descriptions and code comments) on what the changes are.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TimeDistributedMaskCriterion.scala
jason-dai: Why is cell always ClassNLLCriterion?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TimeDistributedMaskCriterion.scala
```diff
@@ -0,0 +1,160 @@
...
+    mask.applyFun[T](target, x =>
+      if (x != ev.fromType[Int](paddingValue)) ev.one else ev.zero)
+    val outputBuff = Tensor()
+    outputBuff.resizeAs(mask)
+    (0 until nstep).foreach(b => {
+      outputBuff
+        .select(2, b + 1)
+        .cmul(mask.select(2, b + 1),
+          cells(b).sparseOutput)
+    })
```
zhangxiaoli73: I think sparseOutput is not needed. 
You can define getSizeAverage in criterion which can determine the compute mode.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractCriterion.scala
```diff
@@ -52,6 +52,8 @@ abstract class AbstractCriterion[A <: Activity: ClassTag, B <: Activity: ClassTa
   var gradInput: A = Activity.allocate[A, T]()
   var output: T = ev.fromType[Int](0)
 
+  var sparseOutput: Tensor[T] = Tensor()
```
zhangxiaoli73: can replaced by getSizeAverage()

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -149,6 +167,26 @@ class Recurrent[T : ClassTag](var batchNormParams: BatchNormParams[T] = null)
       cells.head.hidResize(hidden = hidden, batchSize = batchSize, stepShape)
       gradHidden = hidden
     }
+
+    if (maskBuffer == null) {
```
zhangxiaoli73: just create maskBuffer = Tensor()

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -245,14 +284,34 @@ class Recurrent[T : ClassTag](var batchNormParams: BatchNormParams[T] = null)
...
 
     while (i <= times) {
       currentInput(inputDim) = Recurrent.selectCopy(input2Cell, i, stepInput2CellBuf)
       cells(i - 1).forward(currentInput)
-      currentInput(hidDim) = cells(i - 1).output.toTable(hidDim)
+      val curOutput = cells(i - 1).output
+      if (maskZero && i > minLength) {
+        val curMask = maskBuffer.select(2, i)
+        val curOut = curOutput[Table](hidDim)[Tensor[T]](1)
+        curOutput.update(inputDim, outputBuffers(i - 1).resizeAs(curOut).copy(curOut))
```
zhangxiaoli73: add some comments to explain why you have to do update

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -353,13 +469,42 @@ class Recurrent[T : ClassTag](var batchNormParams: BatchNormParams[T] = null)
...
+            val newState = cells(i - 1).gradInput[Table](hidDim)
+            val originState = currentGradOutput[Table](hidDim)
+            for (j <- 1 to newState.length()) {
+              newState[Tensor[T]](j).select(1, b).copy(originState[Tensor[T]](j).select(1, b))
+            }
+          }
+        }
+      } else {
+        cells(i - 1).backward(_input, currentGradOutput)
+      }
```
zhangxiaoli73: lack “currentGradOutput(hidDim) = cells(i - 1).gradInput.toTable(hidDim)” ??

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -245,14 +284,34 @@ class Recurrent[T : ClassTag](var batchNormParams: BatchNormParams[T] = null)
     // Clone N modules along the sequence dimension.
     initHidden(outputSize.drop(2))
     cloneCells()
+    if (maskZero) {
```
zhangxiaoli73: add requirement to appoint that input should be 3 dimensions if using  "maskZero"

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
```diff
@@ -299,6 +279,9 @@ class Recurrent[T : ClassTag](
       if (maskZero && i > minLength) {
         val curMask = maskBuffer.select(2, i)
         val curOut = curOutput[Table](hidDim)[Tensor[T]](1)
+        // Copy output to a new new tensor as output, because for some cells
+        // such as LSTM the hidden h and ouput o refer to the same tensor.
+        // But in this case, we want h and o have difference values.
```
zhangxiaoli73: difference -> different

jason-dai:
Please also update the correspoding md docs.

PR #2325, opened 2018-02-26
DataFrame-based image reader and transformer

hhbyyh:

Provides DataFrame-based image reader and feature transformer. 
Both Scala and Python interface are provided.

More unit tests will be added.

## How was this patch tested?

new unit tests.

### DIFF
--- a/pyspark/test/bigdl/dlframes/test_dl_image_reader.py
+++ b/pyspark/test/bigdl/dlframes/test_dl_image_reader.py
@@ -27,7 +27,7 @@ class TestDLImageReader():
     def setup_method(self, method):
         """
         setup any state tied to the execution of the given method in a
-        class.  setup_method is invoked for every test method of a class.
+        class. setup_method is invoked for every test method of a class.
         """
         sparkConf = create_spark_conf().setMaster("local[1]").setAppName("test model")
         self.sc = get_spark_context(sparkConf)
@@ -36,7 +36,7 @@ class TestDLImageReader():
 
     def teardown_method(self, method):
         """
-         teardown any state that was previously setup with a setup_method
+        teardown any state that was previously setup with a setup_method
         call.
         """
         self.sc.stop()
--- a/pyspark/test/bigdl/dlframes/test_dl_image_reader.py
+++ b/pyspark/test/bigdl/dlframes/test_dl_image_reader.py
@@ -1,0 +1,58 @@
+#
+# Copyright 2016 The BigDL Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import pytest
+from bigdl.dlframes.dl_image_reader import *
+from bigdl.transform.vision.image import *
+from bigdl.util.common import *
+from pyspark.sql.types import *
+
+
+class TestDLImageReader():
+
+    def setup_method(self, method):
+        """
+        setup any state tied to the execution of the given method in a
+        class.  setup_method is invoked for every test method of a class.
+        """
+        sparkConf = create_spark_conf().setMaster("local[1]").setAppName("test model")
+        self.sc = get_spark_context(sparkConf)
+        self.sqlContext = SQLContext(self.sc)
+        init_engine()
+        resource_path = os.path.join(os.path.split(__file__)[0], "../resources")
+        self.image_path = os.path.join(resource_path, "pascal/000025.jpg")
+
+    def teardown_method(self, method):
+        """
+         teardown any state that was previously setup with a setup_method
+        call.
+        """
+        self.sc.stop()
+
+    def test_get_image(self):
+        image_frame = DLImageReader.readImages(self.image_path, self.sc)
+        assert image_frame.count() == 1
+        assert type(image_frame).__name__ == 'DataFrame'
+        first_row = image_frame.take(1)[0][0]
+        assert first_row[0].endswith("pascal/000025.jpg")
+        assert first_row[1] == 375
+        assert first_row[2] == 500
+        assert first_row[3] == 3
+        assert first_row[4] == 16
+        assert len(first_row[5]) == 95959
+
+if __name__ == "__main__":
+    pytest.main([__file__])
--- a/pyspark/test/bigdl/dlframes/test_dl_image_reader.py
+++ b/pyspark/test/bigdl/dlframes/test_dl_image_reader.py
@@ -28,7 +28,7 @@ class TestDLImageReader():
         setup any state tied to the execution of the given method in a
         class.  setup_method is invoked for every test method of a class.
         """
-        sparkConf = create_spark_conf().setMaster("local[1]").setAppName("test model")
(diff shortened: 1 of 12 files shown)


hhbyyh:
Please help review @songhappy

spark/dl/src/main/scala/com/intel/analytics/bigdl/dlframes/DLImageReader.scala
```diff
@@ -0,0 +1,143 @@
...
+      StructField("height", IntegerType, false) ::
+      StructField("width", IntegerType, false) ::
+      StructField("nChannels", IntegerType, false) ::
+      // OpenCV-compatible type: CV_8UC3, CV_32FC3 in most cases
+      StructField("mode", IntegerType, false) ::
+      // floats in OpenCV-compatible order: row-wise BGR in most cases
+      StructField("data", new ArrayType(FloatType, false), false) :: Nil)
+
+  private[dlframes] def imf2Row(imf: ImageFeature): Row = {
+    val (mode, data) = if (imf.contains(ImageFeature.imageTensor)) {
```
jason-dai: need to handle ImageFeature.floats as well

spark/dl/src/main/scala/com/intel/analytics/bigdl/dlframes/DLImageTransformer.scala
```diff
@@ -0,0 +1,93 @@
...
+
+      if (!result.contains(ImageFeature.imageTensor)) {
+        MatToTensor[Float]().transform(result)
+      }
+      DLImageSchema.imf2Row(result)
+    }
+
+    val transformUDF = SqlAdapter.getUDF(transFunc, DLImageSchema.floatSchema)
+
+    dataFrame.withColumn($(outputCol), transformUDF(dataFrame($(inputCol))))
```
jason-dai: 1. Use the transform in Transformer or FeatuerTransformer instead, as it will handle broadcast, clone, etc.; someting like:
   ```
   r1 = dataFrame.rdd().map(DLImageSchema.row2IMF(_)))
   r2 = localTransformer(r1).map(DLImageSchema.imf2Row(_))
   ```

2. Do we want to replicate the image contents in both inputCol and outputCol here?

jason-dai:
@wzhongyuan please take a look

spark/dl/src/main/scala/com/intel/analytics/bigdl/dlframes/DLImageTransformer.scala
hhbyyh: I'm thinking we should support the case the there're multiple columns in the input DataFrame. To use the RDD interface, we need to extract, transform then zip back with the other columns, which can bring extra performance cost and instability during runtime.

For 2, user may set inputCol and outputCol to the same name. I'll add a unit test.

spark/dl/src/main/scala/com/intel/analytics/bigdl/dlframes/DLImageTransformer.scala
```diff
@@ -0,0 +1,93 @@
...
+    val validTypes = Array(DLImageSchema.floatSchema, DLImageSchema.byteSchema)
+
+    require(validTypes.exists(t => SchemaUtils.sameType(inputType, t)),
+      s"Bad input type: $inputType. Requires ${validTypes.mkString(", ")}")
+  }
+
+  override def transformSchema(schema: StructType): StructType = {
+    val inputType = schema($(inputCol)).dataType
+    validateInputType(inputType)
+    if (schema.fieldNames.contains($(outputCol))) {
```
wzhongyuan: So each transformer in the chain will have its own output column ?  that means the data will be duplicated.

spark/dl/src/main/scala/com/intel/analytics/bigdl/dlframes/DLImageReader.scala
```diff
@@ -0,0 +1,143 @@
...
+import com.intel.analytics.bigdl.transform.vision.image.{BytesToMat, ImageFeature, ImageFrame}
+import org.apache.spark.SparkContext
+import org.apache.spark.sql.types._
+import org.apache.spark.sql.{DataFrame, Row, SQLContext}
+import org.opencv.core.CvType
+
+/**
+ * Definition for image data in a DataFrame
+ */
+object DLImageSchema {
```
wzhongyuan: make it private to  package `dlframes` ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dlframes/DLImageTransformer.scala
hhbyyh: Only one output column for the chained transformers. Please check the unit test for example

spark/dl/src/main/scala/com/intel/analytics/bigdl/dlframes/DLImageReader.scala
hhbyyh: Though it may not be the major use case, but it's possible that user want to manually compose a image column.
The byteSchema is public in Spark Image schema so we probably will keep it stable.

hhbyyh:
Any comments?

pyspark/test/bigdl/dlframes/test_dl_image_reader.py
```diff
@@ -0,0 +1,58 @@
...
+from bigdl.util.common import *
+from pyspark.sql.types import *
+
+
+class TestDLImageReader():
+
+    def setup_method(self, method):
+        """
+        setup any state tied to the execution of the given method in a
+        class.  setup_method is invoked for every test method of a class.
```
wzhongyuan: remove unnecessary space ?

pyspark/test/bigdl/dlframes/test_dl_image_reader.py
```diff
@@ -0,0 +1,58 @@
...
+        """
+        sparkConf = create_spark_conf().setMaster("local[1]").setAppName("test model")
+        self.sc = get_spark_context(sparkConf)
+        init_engine()
+        self.resource_path = os.path.join(os.path.split(__file__)[0], "../resources")
+
+    def teardown_method(self, method):
+        """
+         teardown any state that was previously setup with a setup_method
```
wzhongyuan: indent ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/dlframes/DLImageReader.scala
wzhongyuan: Makes sense

hhbyyh:
Thanks for the comments. Fixed the format issue and rebased.

PR validation passed: http://172.168.2.101:8080/view/PR-V(new)-Pipeline/job/BigDL-PR-Validation-v3/341/

spark/dl/src/main/scala/com/intel/analytics/bigdl/dlframes/DLImageTransformer.scala
jason-dai: The current usage of transformer in the code is incorrect; you also need to clone the transformer after broadcast; see https://github.com/intel-analytics/BigDL/blob/master/spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/Transformer.scala#L67

hhbyyh: Thanks for the reminder. I'll update and run some tests.

hhbyyh:
Updated the code to use mapPartitions and clone.

jason-dai:
LGTM

hhbyyh:
http://172.168.2.101:8080/view/PR-V(new)-Pipeline/job/BigDL-PR-Validation-v3/365/ passed. Merging to master

PR #2327, opened 2018-02-27
fix sparse tensor dot product bug

yangw1234:

fix sparse tensor dot product bug

## How was this patch tested?

add unit tests


yangw1234:
pr validation pass

PR #2332, opened 2018-02-28
[Bug fix] Use new dump_model method in the export_tf_checkpoint.py

yiheng:
In this commit(https://github.com/intel-analytics/BigDL/commit/ac64b7bd3b969dacbc3181d2688c79e024dc87a4#diff-3d29fec1deddd4fa984992de26dcf901), the dump_model method is changed while export_tf_checkpoint.py is not changed. This PR fixes this issue.

## How was this patch tested?
manual test. unit test.


yangw1234:
LGTM

PR #2333, opened 2018-02-28
support getting node and core number in Python

yangw1234:

support getting node and core number in Python

## How was this patch tested?

manually tested

### DIFF
--- a/pyspark/bigdl/util/common.py
+++ b/pyspark/bigdl/util/common.py
@@ -446,6 +446,9 @@ def init_engine(bigdl_type="float"):
 def init_executor_gateway(sc, bigdl_type="float"):
     callBigDlFunc(bigdl_type, "initExecutorGateway", sc, sc._gateway._gateway_client.port)
 
+def get_node_and_core_number(bigdl_type="float"):
+    result = callBigDlFunc(bigdl_type, "getNodeAndCoreNumber")
+    return result[0], result[1]
 
 def redire_spark_logs(bigdl_type="float", log_path=os.getcwd()+"/bigdl.log"):
     """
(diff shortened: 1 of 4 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDL.scala
```diff
@@ -2338,6 +2338,10 @@ class PythonBigDL[T: ClassTag](implicit ev: TensorNumeric[T]) extends Serializab
     Engine.init
   }
 
+  def getNodeAndCoreNumber(): Array[Int] = {
```
hkvision: JList[Int]? Will Array[Int] work?

pyspark/bigdl/util/common.py
```diff
@@ -446,6 +446,9 @@ def init_engine(bigdl_type="float"):
 def init_executor_gateway(sc, bigdl_type="float"):
     callBigDlFunc(bigdl_type, "initExecutorGateway", sc, sc._gateway._gateway_client.port)
 
+def get_node_and_core_number(bigdl_type="float"):
+    result = callBigDlFunc(bigdl_type, "getNodeAndCoreNumber")
```
hkvision: Add a unit test?

yangw1234: done

spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDL.scala
yangw1234: yes, it works

yangw1234:
pr validation pass

PR #2337, opened 2018-03-01
fix maxout ser issue

wzhongyuan:
Fix maxout ser issue

## How was this patch tested?

unit test


wzhongyuan:
PR validation passed http://172.168.2.101:8080/job/BigDL-PR-Validation-v3/276/

PR #2351, opened 2018-03-06
[Enhancement] Check duplicate layers in the container

yiheng:
If there're duplicate layers in the container, an exception will be thrown.

The duplicate check happens when:
1. user add a new layer into a container(this still cannot find some cases)
2. build a graph
3. create an optimizer
4. change the model in an optimizer

## How was this patch tested?
new unit test


yiheng:
PR validation pass

yiheng:
@qiuxin2012 please take a look at this PR

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -1029,5 +1029,13 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
   private[bigdl] def getClassTagNumerics() : (Array[ClassTag[_]], Array[TensorNumeric[_]]) = {
     (Array(scala.reflect.classTag[T]), Array(ev))
   }
+
+  /**
+   * Check if some module is duplicated in the model. For a layer it cannot be duplicated.
+   * Container should override this method
+   */
+  private[bigdl] def checkDuplicate(
+    record: mutable.HashMap[Int, Boolean] = mutable.HashMap()
```
qiuxin2012: Why not HashSet?

yiheng:
Jenkins pass

Issue #2353, opened 2018-03-06
Python add extra jar will fail for spark below 2.2


hkvision:
```
from bigdl.util.common import Configuration
Configuration.add_extra_jars(jars)
Configuration.add_extra_python_modules(packages)
```
When we primarily provide this feature, we suppose it will be used after pip install (namely using pyspark 2.2.0).
However, the above code will fail for adding extra jars when using spark <2.2.
Need to refactor to a unified way for this no matter in what way user start the program as suggested by @shane-huang .

Issue #2350, opened 2018-03-06
jupyter launch script needs to be extended to support adding extra jars and python zip


shane-huang:
juptyer-with-bigdl.sh should support adding extra jars and python zip (user-defined). Or it's not possible to add extra jars after SparkContext is created. 

specifically, the configurations needs to change as below:
```
       --py-files ${PYTHON_API_ZIP_PATH},${user_PYTHON_API_ZIP_PATH} \
       --jars ${BigDL_JAR_PATH},${user_JAR_PATH} \
       --conf spark.driver.extraClassPath=${BigDL_JAR_PATH}:${user_JAR_PATH} \
       --conf spark.executor.extraClassPath=bigdl-${BigDL_lib_version}-jar-with-dependencies.jar:${user_JAR_PATH}
```

hkvision:
I tried adding `--jars --py-files` when starting the script and it works. Jupyter launching scripts in this way should be able to add configurations as normal. I will update the run-without-pip doc to point out this. cc @yiheng @shane-huang

hkvision:
Updated docs here: https://github.com/intel-analytics/BigDL/pull/2358/files#diff-ddd31288395e39dc7e89f92bf41bf453R64

Issue #2346, opened 2018-03-05
add items to FAQ and Trouble Shooting


shane-huang:
1. put this to higher level of headings. This is now barried too deep in docs https://bigdl-project.github.io/0.4.0/#PythonUserGuide/run-without-pip/#faq. 
1. add TypeError: 'JavaPackage' object is not callable may also be caused by mismatching BigDL self-built version and spark version and if any additional lib, the verison of the lib it compiles against. 
1. also a new error: "Py4JError: com.intel.analytics.bigdl.python.api.PythonBigDLKeras.ofFloat does not exist in the JVM" could be caused by version mismatch too.

PR #2358, opened 2018-03-07
Fix and refactor Python adding extra jars; fix doc issues

hkvision:

1. Refactor adding extra jars/python packages. `Configuration` related code is REMOVED.
Now only need to set env variable `BIGDL_JARS` & `BIGDL_PACKAGES`, which works not only for pip install but also for building the project directly.
Modified the docs accordingly.
2. Fix incompatibility of `set_creator_class` between 0.4 and master
3. Tested for spark 2.1 and 2.2 #2353 
4. Fix a bug in set up python package.
5. Move Python FAQ from run-without-pip to Python User Guide/Run/FAQ (the same level as run-from-pip and run-without-pip) so that users can find it more easily. Also enrich FAQ as suggested in #2346 
6. Update docs for running with script #2350 

## How was this patch tested?

Jenkins and manual test.

#2346 #2350 #2353


yiheng:
LGTM

hkvision:
jenkins passed http://172.168.2.101:8080/job/BigDL-PR-Validation-v3/336/

hkvision:
@yiheng Thanks for the review!

PR #2363, opened 2018-03-08
[Fix]add race condtion fix in spark 1.6

wzhongyuan:
To make it compatible with spark_1.6 and scala 2.10 where scala.reflect is not thread safe

## How was this patch tested?
unit test and model validation job

### DIFF
--- a/pyspark/test/bigdl/test_simple_integration.py
+++ b/pyspark/test/bigdl/test_simple_integration.py
@@ -665,13 +665,6 @@ class TestSimple():
         assert node == 1
         assert core == 4
 
-    def tes_read_image_frame(self):
-        init_engine()
-        resource_path = os.path.join(os.path.split(__file__)[0], "resources")
-        image_path = os.path.join(resource_path, "pascal/000025.jpg")
-        image_frame = ImageFrame.read(image_path, self.sc)
-        count = image_frame.get_image().count()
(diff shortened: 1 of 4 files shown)


wzhongyuan:
@yiheng   please help review, I successfully ran the example without any issue now

wzhongyuan:
PR validation passed http://172.168.2.101:8080/job/BigDL-PR-Validation-v3/337/

wzhongyuan:
@yiheng   Any comments ? I just ran another PR to test against latest master code just in case.
http://172.168.2.101:8080/job/BigDL-PR-Validation-v3/345/

yiheng:
Is it possible we add a sync lock outside ```doSerializeModule```

wzhongyuan:
@yiheng   I know what you mean, that will make the code less to get changed, but we will loose kind of efficiency and not that flexible, right ?

yiheng:
It's not only about less change. I think user may forget to add the lock when define the ser code in a new module

wzhongyuan:
@yiheng   Refined as per discussion,

Ran another PR successfully http://172.168.2.101:8080/job/BigDL-PR-Validation-v3/358/

wzhongyuan:
@yiheng   Any other comments ? thanks

PR #2367, opened 2018-03-08
Keras-like API Validate Outputshape

hkvision:

When computing output shape for new layers, ensure that the output shape has all positive entries (excluding batch). This can prevent cases such as too small input for cropping at the time of building a layer instead of at the time of doing forward.

After removing all computing output shape logic to new layers, we can put `computeOutputShape` before creating `labor` and this will save effort if the model will lead to an invalid output.

## How was this patch tested?

Jenkins


hkvision:
jenkins passed http://172.168.2.101:8080/view/PR-V(new)/job/BigDL-PR-Validation-v3/339/

Issue #2310, opened 2018-02-15
Resnet Visualization for Graph Topology and biases


SnehalA:
How to view the topology and parameters visualization.
1. Integration of model.saveGraphTopology("logpath") with Resnet model - [DONE]
2. Viewing “weights”, “bias”, “gradientWeights” and “gradientBias” under the DISTRIBUTIONS and HISTOGRAMS tabs .

Only able to integrate the training and validation summaries

jason-dai:
I believe you need to look at the GRAPH tab for viewing the topology.

zhangxiaoli73:
default model in resenet/Train.scala is non-graph resnet, if you want to use graph model, you can add "--graphModel" to get graph model.

SnehalA:
I have tried -g option, but it isn't working. Can the topology and parameters for Resnet/Inception models be viewed under tensorboard? Thanks.

zhangxiaoli73:
If you have set -g, please try "model.asInstanceOf[Graph[Float]].saveGraphTopology("logpath")" to save topology. @SnehalA

SnehalA:
Any tips for viewing “weights”, “bias”, “gradientWeights” and “gradientBias” under the DISTRIBUTIONS and HISTOGRAMS tabs  ?

qiuxin2012:
@SnehalA 
I think you have known how to [create TrainSummary](https://bigdl-project.github.io/0.4.0/#ProgrammingGuide/visualization/).
Add a SummaryTrigger for TrainSummary like following code:
```
val optimizer = Optimizer(...)
...
val logdir = "mylogdir"
val appName = "myapp"
val trainSummary = TrainSummary(logdir, appName)
trainSummary.setSummaryTrigger("Parameters", Trigger.severalIteration(10))  // enable collecting “weights”, “bias”, “gradientWeights” and “gradientBias”
val validationSummary = ValidationSummary(logdir, appName)
optimizer.setTrainSummary(trainSummary)
optimizer.setValidationSummary(validationSummary)
...
val trained_model = optimizer.optimize()
```

SnehalA:
Thank you, closing this issue.

PR #2368, opened 2018-03-08
Model Visualization enabled - Resnet & Inception

SnehalA:
-Enabled Graph topology, training summaries in tensorboard for Resnet and inception models.
-Added parameter "visualization"
-Added Top5 accuracy capture for Resnet

## How was this patch tested?
Manual tests

fixed https://github.com/intel-analytics/BigDL/issues/2310

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/resnet/Train.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/resnet/Train.scala
@@ -24,6 +24,8 @@ import com.intel.analytics.bigdl.utils.{Engine, LoggerFilter, T, Table}
 import org.apache.log4j.{Level, Logger}
 import org.apache.spark.SparkContext
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric._
+import com.intel.analytics.bigdl.nn.Graph
+import com.intel.analytics.bigdl.visualization.{TrainSummary,ValidationSummary,Summary}
 
 object Train {
   LoggerFilter.redirectSparkInfoLogs()
@@ -105,7 +107,7 @@ object Train {
       optimizer
         .setOptimMethod(optimMethod)
         .setValidation(Trigger.everyEpoch,
-          validateSet, Array(new Top1Accuracy[Float]))
+          validateSet, Array(new Top1Accuracy[Float], new Top5Accuracy[Float]))
         .setEndWhen(Trigger.maxEpoch(maxEpoch))
         .optimize()
       sc.stop()


spark/dl/src/main/scala/com/intel/analytics/bigdl/models/inception/Options.scala
```diff
@@ -39,7 +39,8 @@ object Options {
     warmupEpoch: Option[Int] = None,
     gradientL2NormThreshold: Option[Double] = None,
     gradientMin: Option[Double] = None,
-    gradientMax: Option[Double] = None
+    gradientMax: Option[Double] = None,
+    visualization: Boolean = true
```
yiheng: Can we change the ```visualization:Boolean``` to something like ```summary:Option[String]``` and let user input the visualization summary location? So if user inputs such value, we can save the graph visualization file in that location instead of hard code it in our code; if user doesn't input, it means he/she doesn't want to enable the visualization

yiheng: Also please update the ReadMe.md file in the same folder to let user know how to use this new option.

yiheng: I think you may miss the change in the trainParser?

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/inception/Train.scala
```diff
@@ -118,6 +120,19 @@ object TrainInceptionV1 {
       if (param.gradientL2NormThreshold.isDefined) {
         optimizer.setGradientClippingByl2Norm(param.gradientL2NormThreshold.get.toFloat)
       }
+      
+      if (param.visualization) {
+        val logdir = "bigdlSummary"
```
yiheng: As mentioned above, can you let user input the logdir instead of hard code one in the code

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/inception/Train.scala
```diff
@@ -118,6 +120,19 @@ object TrainInceptionV1 {
...
+        val timestamp: Long = System.currentTimeMillis / 1000
```
yiheng: can we convert the timestamp into a more human readable format? I have a similiar change in my local. I use the human readable time as the suffix. For your reference
```scala
val sdf = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss")
val timeStamp = sdf.format(new Date())
```

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/inception/Train.scala
```diff
@@ -118,6 +120,19 @@ object TrainInceptionV1 {
...
+        val appName = "inception_"+timestamp
+        val trainSummary = TrainSummary(logdir, appName)
+        val validationSummary = ValidationSummary(logdir, appName)
+        trainSummary.setSummaryTrigger("Parameters", Trigger.severalIteration(10))
+        optimizer.setValidationSummary(validationSummary)
+        optimizer.setTrainSummary(trainSummary)
+        model.asInstanceOf[Graph[Float]].saveGraphTopology(logdir)
+        //print("class Num "+param.classNumber)
```
yiheng: please remove this commented line

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/resnet/Train.scala
```diff
@@ -89,11 +91,23 @@ object Train {
       if (param.checkpoint.isDefined) {
         optimizer.setCheckpoint(param.checkpoint.get, Trigger.everyEpoch)
       }
+      
+      if (param.visualization) {
+        val logdir = "bigdlSummary"
```
yiheng: Please check my comments in the above

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/resnet/Utils.scala
```diff
@@ -42,7 +42,8 @@ object Utils {
     momentum: Double = 0.9,
     dampening: Double = 0.0,
     nesterov: Boolean = true,
-    graphModel: Boolean = false)
+    graphModel: Boolean = false,
+    visualization: Boolean = true)
```
yiheng: please check my comments in the inception code

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/inception/Train.scala
```diff
@@ -118,6 +120,19 @@ object TrainInceptionV1 {
...
+      if (param.visualization) {
+        val logdir = "bigdlSummary"
+        val timestamp: Long = System.currentTimeMillis / 1000
+        val appName = "inception_"+timestamp
+        val trainSummary = TrainSummary(logdir, appName)
+        val validationSummary = ValidationSummary(logdir, appName)
+        trainSummary.setSummaryTrigger("Parameters", Trigger.severalIteration(10))
+        optimizer.setValidationSummary(validationSummary)
+        optimizer.setTrainSummary(trainSummary)
+        model.asInstanceOf[Graph[Float]].saveGraphTopology(logdir)
```
yiheng: we should check if this is real a graph model.
```scala
if (model.isInstanceOf[Graph[Float]]) ...
```

PR #2372, opened 2018-03-12
Fix deploy failure in nightly build

yiheng:
On the deployment server, we use JDK7 to build the package. However, Spark2 is built with JDK8. A warning is thrown when building the package. As we turn on the warning as failure now, the nightly build job is failed.

Split the original deploy script into two scripts. One for spark1 and another for spark2. So user can change the JDK setting between the two scripts.

## How was this patch tested?
unit tests, manual test


yiheng:
Jenkins pass

glorysdj:
LGTM

PR #2376, opened 2018-03-12
[Enhancement] make SparseMiniBatch supporting TensorDataTypes other than T

sperlingxx:

As Modules(Operations) with Tensor types other than [T] added, it is necessary to support batching samples which contains Tensors of multi-types(some types other than [T]). 
For instances, 

     val t1 = Tensor[Double](3).rand()
     val t2 = Tensor[Float](3).rand()
     val sample = TensorSample.create[Float](Array(t1, t2))
     val miniBatch = SparseMiniBatch[Float](2, 0)
     miniBatch.set(Seq(sample))
   
## How was this patch tested?

unit test

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/SparseTensor.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/SparseTensor.scala
@@ -1146,7 +1146,7 @@ object SparseTensor{
       val curLength = currentTensor.nElement()
       val curTensorOffset = currentTensor.storageOffset() - 1
       // copy to concat _values
-      ev.arraycopy(currentTensor.storage().array(), curTensorOffset,
+      System.arraycopy(currentTensor.storage().array(), curTensorOffset,
         res.storage().array(), offset, curLength)
       // make new Indices
       var indicesIndex = 0
--- a/spark/dl/src/test/scala/com/intel/analytics/bigdl/dataset/MiniBatchSpec.scala
+++ b/spark/dl/src/test/scala/com/intel/analytics/bigdl/dataset/MiniBatchSpec.scala
@@ -109,4 +109,34 @@ class MiniBatchSpec extends FlatSpec with Matchers {
     target should be (expectedTarget)
   }
 
+  "SparseTensorMiniBatch with different TensorTypes" should "return right result" in {
+    val a1 = Tensor.sparse(Tensor[Double](4).range(1, 4, 1)).asInstanceOf[Tensor[Float]]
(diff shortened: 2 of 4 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/SparseTensor.scala
```diff
@@ -1146,7 +1146,7 @@ object SparseTensor{
       val curLength = currentTensor.nElement()
       val curTensorOffset = currentTensor.storageOffset() - 1
       // copy to concat _values
-      ev.arraycopy(currentTensor.storage().array(), curTensorOffset,
+      System.arraycopy(currentTensor.storage().array(), curTensorOffset,
```
yiheng: why change this?

sperlingxx: Because `ev.arraycopy(src: Array[T], srcPos: Int, dest: Array[T], destPos: Int, length: Int)` is inherited by  _NumericFloat_ and _NumericDouble_, which requiring Array[Float] or Array[Double].

sperlingxx:
@yiheng Yes, I've fixed it.

spark/dl/src/test/scala/com/intel/analytics/bigdl/dataset/MiniBatchSpec.scala
```diff
@@ -109,4 +109,34 @@ class MiniBatchSpec extends FlatSpec with Matchers {
     target should be (expectedTarget)
   }
 
+  "SparseTensorMiniBatch with different TensorTypes" should "return right result" in {
+    val a1 = Tensor.sparse(Tensor[Double](4).range(1, 4, 1)).asInstanceOf[Tensor[Float]]
```
yiheng: should we hide this piece of code into the TensorSample.apply()?

sperlingxx: It looks great!

sperlingxx: Since __Array[Tensor[T]]__ is __not__ a sub type of __Array[Tensor[_]]__,  it will cause compile error if changing apply method directly.
So, I'm trying to add a _TensorSample.create_ method instead.

yiheng:
LGTM. Running jenkins

yiheng:
Jenkins passed. Merge. Thanks for the contribution!

PR #2380, opened 2018-03-13
[BugFix] TensorOp & SelectTensor

sperlingxx:

1.  __ClassNotFoundException__ were caught occasionally when deserializing TensorOp$.ClosureWrapper, which may be caused by _latestUserDefinedloader_  in _ObjectInputStream.resolveClass_ .
  So, re-implement a Java deserialize Method , in which _ObjectInputStream.resolveClass_ is overridden. 

2.  __SelectTensor__ couldn't be used to fetch _Tensor_ from _Table_ with incremental keys. 
   So,  add an option(_isTensorKey_) to let user decide whether  the key is a primitive type or a scalar tensor. 

## How was this patch tested?

unit test


yiheng:
The change looks good to me. Can you resolve the conflict so I can run a jenkins test?

sperlingxx:
@yiheng  No conflict now :)

yiheng:
Jenkins passed. Merge. Thanks for the contribution.

Issue #2168, opened 2018-01-08
The type of return model by optimizer.optimze() will change to StaticGraph.


qiuxin2012:
The type of return model by optimizer.optimze() will change to StaticGraph, if I put a graph model to the optimizer.
As StaticGraph's type is a layer not a Model, I couldn't call `.save_graph_topology` method.

zhichao-li:
We need to check and wrap the a concrete type(Model, Sequential or Layer) according to the type of `jmodel`
```   
 def optimize(self):
        """
        Do an optimization.
        """
        jmodel = callJavaFunc(get_spark_context(), self.value.optimize)
        from bigdl.nn.layer import Layer
        return Layer.of(jmodel)
```

PR #2382, opened 2018-03-14
Fix reload model in python

### LINKED ISSUE #2362: Python Model.load will return a Layer
Python Model.load / Model.load_caffe and some other methods will always return a Layer, which doesn't make sense for a Sequential or Graph.
#2168 also points out this issue. cc @qiuxin2012

zhichao-li:
- This code
```
        model_loaded = Model.load(tmp_path)
        model_loaded.layers 
```
would fail as model_loaded is treated as `Layer` instead of `Model` or other subclass.

Fix #2362 #2168

- Fix exception in: #2388 
```
bmodel = with_bigdl_backend(kmodel)
bmodel.predict(xxx)
```
## How was this patch tested?

unittest

### DIFF
--- a/pyspark/bigdl/nn/layer.py
+++ b/pyspark/bigdl/nn/layer.py
@@ -102,10 +102,8 @@ class SharedStaticUtils():
             base_module = importlib.import_module('bigdl.nn.keras.topology')
         elif "com.intel.analytics.bigdl.nn.keras" == jpackage_name:
             base_module = importlib.import_module('bigdl.nn.keras.layer')
-        elif "com.intel.analytics.bigdl.nn" == jpackage_name:
-            base_module = importlib.import_module('bigdl.nn.layer')
         else:
-            raise Exception("Not supported type: %s" % jname)
+            base_module = importlib.import_module('bigdl.nn.layer')
 
         realClassName = "Layer" # The top base class
         if pclass_name in dir(base_module):
--- a/pyspark/bigdl/nn/layer.py
+++ b/pyspark/bigdl/nn/layer.py
@@ -16,6 +16,7 @@
 
 
 import sys
+import importlib
 
 import numpy as np
 import six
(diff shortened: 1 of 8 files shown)


zhichao-li:
jenkins passed http://172.168.2.101:8080/job/BigDL-PR-Validation-v3/372/

pyspark/bigdl/nn/layer.py
```diff
@@ -1060,8 +1081,19 @@ class Sequential(Container):
 
     '''
 
-    def __init__(self, bigdl_type="float"):
-        super(Sequential, self).__init__(None, bigdl_type)
+    def __init__(self, jvalue=None, bigdl_type="float"):
+        super(Sequential, self).__init__(jvalue, bigdl_type)
+
+    @staticmethod
+    def from_jvalue(jvalue, bigdl_type="float"):
```
hkvision: Why there are two `from_jvalue`?

pyspark/bigdl/nn/layer.py
```diff
@@ -115,14 +115,35 @@ def __call__(self, x=None):
...
         :return: A Python Layer
         """
-        model = Layer(jvalue, bigdl_type)
+        realClassName = callBigDlFunc(bigdl_type,
+                      "getRealClassNameOfJValue",
+                      jvalue)
+        if realClassName == "StaticGraph" or realClassName == "DynamicGraph":
+            realClassName = "Model"
+        elif realClassName == "Input":
+            realClassName = "Layer"
```
hkvision: What if the real class is a `KerasModel`? Or in other words, if `Sequential`, how do you identify whether it is a Torch Sequential or a Keras Sequential?

zhichao-li:
jenkins passed  http://172.168.2.101:8080/job/BigDL-PR-Validation-v3/385/

PR #2401, opened 2018-03-19
Public getInputShape and getOutputShape

zhichao-li:

Open two methods for user `getInputShape` and `getOutputShape`. 

## How was this patch tested?
unittest


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/InferShape.scala
```diff
@@ -46,16 +45,20 @@ trait InferShape {
...
-  private[bigdl] final def getInputShape(): Shape = {
+  final def getInputShape(): Shape = {
+    require(this.isKerasStyle(),
+      "Torch style definition doesn't support getInputShape for now.")
     _inputShapeValue
   }
 
   /**
-   * We suppose the first dim is batch
+   * Return the outputShape for the current Layer and the first dim is batch.
```
hkvision: Can it return the output shape of a layer before it being added into a container?
It seems calling `Dense(10, inputShape=Shape(5)).getOutputShape()` will give null.

zhichao-li:
http://172.168.2.101:8080/job/BigDL-PR-Validation-v3/397/  jenkins passed

PR #2403, opened 2018-03-19
[Bug Fix] Fix duplicate check sometimes should be suspend

yiheng:
We find that sometimes duplicate check should be suspended. Here's an example
```scala
    val bx1 = Input(Shape(4))
    val bx2 = Input(Shape(5))
    val by1 = Dense(6, activation = "sigmoid").inputs(bx1)
    val bbranch1 = Model(bx1, by1).inputs(bx1)
    val bbranch2 = Dense(8).inputs(bx2)
    val bz = merge(List(bbranch1, bbranch2), mode = "concat")
    val bmodel = Model(Array(bx1, bx2), bz)
```
This kind of code will be allowed in keras but will thrown exception in BigDL. The root cause is that the ```bx1``` is added to ```bbranch1``` and ```bmodel```. So bmodel will throw exeption when do the duplicate check.

However, this kind of usage should be allowed as
1. Keras can do it
2. The ```bx1``` is the first node in both ```bbranch1``` and ```bmode```, which won't cause any error

The PR fix this by adding a flag to the module API. So when the user uses an input node, the duplicate check will be suspended.

This PR also fix a bug that the duplicate check is missed in Keras Seq topology.

## How was this patch tested?
unit tests

### DIFF
--- a/pyspark/test/bigdl/keras/test_keras_api.py
+++ b/pyspark/test/bigdl/keras/test_keras_api.py
@@ -30,7 +30,7 @@ from keras.models import Sequential as KSequential
 np.random.seed(1337)  # for reproducibility
 
 
-class TestNewAPI(BigDLTestCase):
+class TestKerasAPI(BigDLTestCase):
 
     def test_embedding(self):
         input_data = np.random.randint(1000, size=(32, 10))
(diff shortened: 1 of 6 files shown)


yiheng:
jenkins passed

hkvision:
Previously in `test_newapi.py` https://github.com/intel-analytics/BigDL/blob/master/pyspark/test/bigdl/keras/test_newapi.py#L203 , I use one more input node to avoid this problem in the last two unit tests. Now we can refine these python tests as well. Can either do this PR or I will update the python side after this PR is merged. :)

hkvision:
Also thanks for fixing checkDuplicate on Keras Sequential add @yiheng ><

yiheng:
@hkvision can we rename the ```test_newapi.py``` to ```test_keras_api.py```

hkvision:
@yiheng Yeah. `test_keras_api` will be better.

hkvision:
@yiheng  Much thanks for the update!

Issue #414, opened 2017-01-24
Add epoch# to the accuracy info


zhichao-li:
2017-01-24 15:17:02 INFO  DistriOptimizer$:461 - top1 accuracy is Accuracy(correct: 47311, count: 55000, accuracy: 0.8602)

PR #2417, opened 2018-03-21
[Enhancement] Add epoch# to the accuracy info to validation output

yiheng:
Add epoch# to the accuracy info to validation output

## How was this patch tested?
manual test, unit test

fixed https://github.com/intel-analytics/BigDL/issues/414


yiheng:
jenkins pass

Issue #1947, opened 2017-11-28
empty tensor call size() stride() will throw null exception


SeaOfOcean:
empty tensor call size() stride() will also throw null exception

```
val t = Tensor[Float]()
t.size()
t.stride()
```

PR #2418, opened 2018-03-21
[Bug Fix] Empty tensor call size() and stride() not throw null exception

yiheng:
Empty tensor call size() and stride() not throw null exception

## How was this patch tested?
new unit test and jenkins

fixed https://github.com/intel-analytics/BigDL/issues/1947


yiheng:
jenkins passed

PR #2440, opened 2018-03-22
LeNet using Keras one-hot labels

hkvision:

Use Keras MNIST dataset and one-hot label for training.
But for validation, only label-based is supported for now.

## How was this patch tested?

Manual running the example.


zhichao-li:
Numpy is 0-based index, Tensor in scala is one-based index and the validation only support scalar value instead of one-hot. We might need to solve those inconsistent behavior before merging this.

Issue #2384, opened 2018-03-15
Python DataSet of ImageFrame needs to support predict, evaluation or have API to get ImageFrame


jenniew:
Python DataSet of ImageFrame can not do prediction or evaluation. Either provide APIs to do prediction/evaluation, or provide API to get ImageFrame from Dataset, then use imageFrame to do prediction/evaluation. Also, ImageFrame does not have API to do evaluation, we may need to provide this API or provide API to extract Sample RDD from imageframe to do evaluation.

PR #2464, opened 2018-03-27
[Enhancement] ImageFrame adding more APIs

wzhongyuan:

Add imageframe based module evaluation and related python API
 
## How was this patch tested?

Model validation with inception v1

fixed https://github.com/intel-analytics/BigDL/issues/2384

### DIFF
--- a/pyspark/bigdl/dataset/dataset.py
+++ b/pyspark/bigdl/dataset/dataset.py
@@ -26,10 +26,12 @@ if sys.version >= '3':
 
 class DataSet(JavaValue):
 
-    def __init__(self, jvalue=None, bigdl_type="float"):
+    def __init__(self, jvalue=None, image_frame = None, bigdl_type="float"):
         self.bigdl_type = bigdl_type
         if jvalue:
             self.value = jvalue
+        if image_frame:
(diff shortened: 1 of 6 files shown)


wzhongyuan:
@jenniew  please help review, thanks

wzhongyuan:
I already tested it by running model validation example using Python API

jenniew:
Please add test case of write_parquet.

wzhongyuan:
@jenniew  added. also found a bug in `read_parquet`, we should passed in a spark context instead of sql context, because there is no java version sql context, we should initialize sql context on inside in `PythonBigDL`

wzhongyuan:
PR validation passed http://172.168.2.101:8080/job/BigDL-PR-Validation-v3/475/

jenniew:
I found this issue too.  There is another way, for the py2java method, if isinstance(obj, SQLContext): obj = obj._ssql_ctx to get scala sql context.

wzhongyuan:
@jenniew  ok, but  i don't want to change the common method ,  changing this method only is preferable

jenniew:
OK

wzhongyuan:
Any other concerns？

PR #2473, opened 2018-03-29
fix BCE return Nan

### LINKED ISSUE #2471: BCECriterion may return NaN
BCECriterion may return NaN, caused by #2231 .
#2231 speed up BCECriterion, but import a new bug may return Nan.
The changing in line 84,
```
old : buffer.fill(ev.fromType(1.0)).sub(input).add(ev.fromType(eps)).log()
new:  buffer.fill(ev.fromType(1.0 + eps)).sub(input).log()
```
For float version BCE, `ev.fromType(1.0 + eps)` (while `eps = 1e-12`) will convert the Double to Float, but `(1.0 + 1e-12).toFloat`'s result is `1.0`, eps won't work for float version BCE.

qiuxin2012:

fix BCE return Nan

## How was this patch tested?

unit tests

fixed #2471


yangw1234:
LGTM

qiuxin2012:
jenkins passed

Issue #2292, opened 2018-02-09
Save/Load module should not use getParameterTable method


yiheng:
As the getParameterTable will mix weight and gradients together, there's no way to distinguish them. Currently the save/load method only save parameter with name "weight" and "bias", which is wrong as there's no restrict to the name and one layer may have more than three weights parameter with a different name.

wzhongyuan:
@yiheng   Previously  there was not layer breaking this convention and i saw other utility uses this way as well, also this API is much friendly to use

Currently there are couple of APIs in `AbstractModule` like `parameters` `getParameters` `setWeightsAndBias` , `setWeights`, etc. we should also open an issue to consolidate them to avoid confusion, it's kind of messed up...

PR #2477, opened 2018-03-30
[Fix] using parameters() instead of getParameterTable() to get weight and bias

wzhongyuan:

using parameters() instead of getParameterTable() to get weight and bias

1) update the bigdl proto file adding repeated tensor field and update generated file accordingly
2) refine copy2bigdl and copyfrombigdl to load parameters using parameters() instead of getparametertable()
3) keep legacy way to load old versioned model

## How was this patch tested?

unit test and model validation test

fixed https://github.com/intel-analytics/BigDL/issues/2292

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/ModuleSerializable.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/ModuleSerializable.scala
@@ -353,10 +353,10 @@ trait ModuleSerializable extends Loadable with Savable{
    */
   protected def copyFromBigDL[T: ClassTag](context : SerializeContext[T],
     modelBuilder : BigDLModule.Builder)(implicit ev : TensorNumeric[T]) : Unit = {
-    val parameters = context.moduleData.module.parameters._1
-    if (parameters != null) {
+    val parameters = context.moduleData.module.parameters
+    if (parameters != null && parameters._1 != null) {
       modelBuilder.setHasParameters(true)
-      parameters.foreach(parameter => {
+      parameters._1.foreach(parameter => {
         val tensorAttr = AttrValue.newBuilder
         TensorConverter.setAttributeValue(context, tensorAttr, parameter)
         modelBuilder.addParameters(tensorAttr.getTensorValue)
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/ModuleSerializable.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/ModuleSerializable.scala
@@ -21,6 +21,7 @@ import com.intel.analytics.bigdl.nn.Container
 
 import scala.collection.JavaConverters._
 import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, Activity}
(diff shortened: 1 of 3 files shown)


wzhongyuan:
@i8run   please help check if any impact on the quantization and the layers with multiple weights

wzhongyuan:
PR validation passed http://172.168.2.101:8080/job/BigDL-PR-Validation-v3/486/

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/ModuleSerializable.scala
```diff
@@ -285,6 +289,43 @@ trait ModuleSerializable extends Loadable with Savable{
...
+  private def copyParameters2BigDL[T: ClassTag]
+    (context: DeserializeContext, module : ModuleData[T])
+    (implicit ev: TensorNumeric[T]): Unit = {
+
+    val serializedParameters = context.bigdlModule.getParametersList.asScala.toArray
+
+    val arrayValue = ArrayValue.newBuilder
+    arrayValue.setDatatype(DataType.TENSOR)
+    serializedParameters.foreach(param => arrayValue.addTensor(param))
+    arrayValue.setSize(serializedParameters.length)
```
i8run: What will happen if we set `hasParameters` to true, but without the parameters?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/ModuleSerializable.scala
```diff
@@ -285,6 +289,43 @@ trait ModuleSerializable extends Loadable with Savable{
    */
   protected def copy2BigDL[T: ClassTag](context: DeserializeContext, module : ModuleData[T])
                                        (implicit ev: TensorNumeric[T]): Unit = {
+
+    if (context.bigdlModule.getHasParameters) {
+      copyParameters2BigDL(context, module)
+    } else {
+      // for legacy format models
+      copyWeightAndBias(context, module)
```
i8run: If we set `weight`, `bias` while `hasParameters` is `true`. What should we do?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/ModuleSerializable.scala
wzhongyuan: this variable will only be set to true when model has parameters(false by default), check below code : 
```scala
if (parameters != null && parameters._1 != null) {
      modelBuilder.setHasParameters(true)
      parameters._1.foreach(parameter => {
        val tensorAttr = AttrValue.newBuilder
        TensorConverter.setAttributeValue(context, tensorAttr, parameter)
        modelBuilder.addParameters(tensorAttr.getTensorValue)
      })
}
```

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/ModuleSerializable.scala
wzhongyuan: it could not happen, this `else` is only for those legacy format models already existing in model zoo

i8run:
LGTM.

PR #2488, opened 2018-04-09
[Fix] Fix python API for random split

wzhongyuan:
 Fix python API for random split, the default deserializer would return a dict

## How was this patch tested?

unit test and manual test


wzhongyuan:
PR validation passed http://172.168.2.101:8080/job/BigDL-PR-Validation-v3/489/

PR #2492, opened 2018-04-13
[Feature] - support group serializer

wzhongyuan:

Support group serializer so thay layers of the same hierarchy could share the same serializer
it could reduce the effort, users do not need to register multiple times for each module

## How was this patch tested?
unit test


wzhongyuan:
@i8run @hkvision  please help review, thanks

wzhongyuan:
PR validation passed http://172.168.2.101:8080/job/BigDL-PR-Validation-v3/490/

hkvision:
Thanks for the help!

PR #2506, opened 2018-04-24
[bug fix]refine getTimes and time counting.

### LINKED ISSUE #2499: getTimes() in Container is inconsistent in most of the Containers.
Some container's himself time cost contains submodules' forward/backward time, like StaticGraph.
Some container doesn't return container's time cost, like Concat.
Some container's himself time cost doesn't contain submodules' forward/backward time, like Recurrent.

We need to apply a general rule for these containers, or it's hard to breakdown the time cost in each layer.

qiuxin2012:
Refine time counting:
* Fix time counting doesn't work in some of the modules, like BiRecurrent.
* Fix Cell's `getTimes()` will return a wrong result if call getTimes more than two times.
* Fix `getTimes()` in some Containers doesn't contains himself, like Sequential, Concat.

## How was this patch tested?

unit tests, manual tests on deepspeech2

fixed #2499

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialDivisiveNormalization.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialDivisiveNormalization.scala
@@ -137,7 +137,7 @@ class SpatialDivisiveNormalization[T: ClassTag](
   private var localstds: Tensor[T] = _
 
   override def updateOutput(input: Tensor[T]): Tensor[T] = {
-    localstds = stdestimator.updateOutput(input).toTensor[T]
+    localstds = stdestimator.forward(input).toTensor[T]
 
     // compute side coefficients
     val dim = input.dim()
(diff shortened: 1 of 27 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialDivisiveNormalization.scala
```diff
@@ -137,7 +137,7 @@ class SpatialDivisiveNormalization[T: ClassTag](
   private var localstds: Tensor[T] = _
 
   override def updateOutput(input: Tensor[T]): Tensor[T] = {
-    localstds = stdestimator.updateOutput(input).toTensor[T]
+    localstds = stdestimator.forward(input).toTensor[T]
```
zhangxiaoli73: you can not get time for each layer in stdestimator

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialSubtractiveNormalization.scala
```diff
@@ -116,20 +116,20 @@ class SpatialSubtractiveNormalization[T: ClassTag](
       if (dim == 4) {
         // batch mode
         ones.resizeAs(input(1)).fill(ev.fromType[Int](1))
-        val _coef = meanestimator.updateOutput(ones).toTensor[T]
+        val _coef = meanestimator.forward(ones).toTensor[T]
```
zhangxiaoli73: same, you can not get time for each layer in meanestimator

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/TanhShrink.scala
```diff
@@ -33,7 +33,7 @@ class TanhShrink[T: ClassTag](
   private val tanh = new Tanh[T]()
 
   override def updateOutput(input: Tensor[T]): Tensor[T] = {
-    val th = tanh.updateOutput(input)
+    val th = tanh.forward(input)
```
zhangxiaoli73: maybe no need, you can not get tanh time

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -168,6 +168,18 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
     Array((this, forwardTime, backwardTime))
   }
 
+  /**
+   * Get the forward/backward cost time for the module or its submodules
+   * and group by module type.
+   * @return (module type name, forward time, backward time)
+   */
+  final def getTimesGroupByModuleType():
```
zhangxiaoli73: when will use this function

qiuxin2012: Breaking down the performance in a model.

qiuxin2012:
jenkins passed

PR #2513, opened 2018-05-02
[New feature]share Const in ModelBroadcast

qiuxin2012:

share values of Const in ModelBroadcast

## How was this patch tested?

unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
@@ -60,10 +60,11 @@ class ModelBroadcast[T: ClassTag](applyProtoBuffer: Boolean = false)
         val moduleConsts = getAndClearConsts(model.asInstanceOf[Container[_, _, T]])
         broadcastConsts = sc.broadcast(moduleConsts)
       }
-      // broadcast weight
+      // broadcast weight and model
       val weightsBias = getAndClearWeightBias(model.parameters())
       broadcastModel = sc.broadcast(model.cloneModule())
       broadcastParameters = sc.broadcast(weightsBias)
+
       putWeightBias(weightsBias, model)
       initGradWeightBias(weightsBias, model)
     }
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
@@ -17,6 +17,7 @@
 package com.intel.analytics.bigdl.models.utils
 
 import com.intel.analytics.bigdl.Module
+import com.intel.analytics.bigdl.nn.{Container, Graph}
 import com.intel.analytics.bigdl.tensor.{QuantizedTensor, QuantizedType, Storage, Tensor}
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
 import org.apache.spark.SparkContext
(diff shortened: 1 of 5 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Util.scala
```diff
@@ -134,6 +136,84 @@ object Util {
...
+      .map(v => (v, newTensor(v.value)))
+    moduleConsts.foreach(_._1.value.set())
+    val result = moduleConsts.map(v => (v._1.getName(), v._2)).toMap[String, Tensor[_]]
+    require(result.size == moduleConsts.length, s"${model}'s Const node's name is duplicated," +
+      s"please check your model.")
+    result
+  }
+
+  protected def newTensor(tensor: Tensor[_]): Tensor[_] = {
+    tensor.getType() match {
```
qiuxin2012: @yiheng Is there a better approach to instead of this matching?

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
```diff
@@ -53,9 +55,16 @@ class ModelBroadcast[T: ClassTag](applyProtoBuffer: Boolean = false)
     if (applyProtoBuffer) {
       broadcastModel = sc.broadcast(model)
     } else {
+      // broadcast Consts
```
yiheng: Update the method doc

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
```diff
@@ -53,9 +55,16 @@ class ModelBroadcast[T: ClassTag](applyProtoBuffer: Boolean = false)
...
+      if (model.isInstanceOf[Container[_, _, T]]) {
+        val moduleConsts = getAndClearConsts(model.asInstanceOf[Container[_, _, T]])
+        broadcastConsts = sc.broadcast(moduleConsts)
```
yiheng: should we broadcast once?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Util.scala
```diff
@@ -134,6 +136,84 @@ object Util {
     }
   }
 
+  private[bigdl] def getAndClearConsts[T: ClassTag](
+        model: Container[_, _, T])(implicit ev: TensorNumeric[T]): Map[String, Tensor[_]] = {
+    val moduleConsts = model.findModules("Const")
+      .map(_.asInstanceOf[Const[T, _]])
+      .map(v => (v, newTensor(v.value)))
```
yiheng: why not use clone tensor?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Util.scala
```diff
@@ -134,6 +136,84 @@ object Util {
...
+    moduleConsts.foreach(_._1.value.set())
+    val result = moduleConsts.map(v => (v._1.getName(), v._2)).toMap[String, Tensor[_]]
```
yiheng: Can we use array of tensor instead of map?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Util.scala
qiuxin2012: Reuse the storage array to prevent memory copy and GC.

yiheng:
LGTM

qiuxin2012:
jenkins passed

PR #2514, opened 2018-05-02
add save/load support for DLModels

sperlingxx:
Add save/load support for DLModels.
This PR is built on the basis of  MLWriter/MLReader, so it only works on _**spark 1.6+**_.

## How was this patch tested?
unit test

This PR is reorganized from https://github.com/intel-analytics/BigDL/pull/2192


spark/dl/src/main/scala/org/apache/spark/ml/MLParams.scala
```diff
@@ -0,0 +1,368 @@
...
+      case "TensorFloat" =>
+        decode[Float]((js \ "objectData").asInstanceOf[JObject])
+      case tt =>
+        throw new IllegalArgumentException(s"Wrong TensorDataType: $tt")
+    }
+  }
+
+  override def jsonEncode(value: OptimMethod[_]): String = {
+    // infer tensor data type by suffix of ClassName
+    val json = value.getClass.getName match {
```
yiheng: This is a little weird... Should we consider adding class tag to OptimMethod? @hhbyyh Do you have any commets about this?

yiheng:
Run a jenkins

spark/dl/src/main/scala/org/apache/spark/ml/MLParams.scala
hhbyyh: I haven't tried with the json serialization much myself, yet I'm curious if the current implementation can provide backward compatibility. E.g. if we add a new field for SGD() in a new release, will the current implementation with jObject be able to load the optimMethod saved with previous version?

sperlingxx: @hhbyyh It's a potential problem, maybe we can add decoding part for new field with default, like : __Option(...).getOrElse(<default_values>)__ ?

hhbyyh: I don't yet have an easy solution for all the customized parameters (like optimMethod, Trigger, Train/validation summary). For now I would suggest excluding those fields from model serialization. That's just my personal opinion.

Please confirm with @yiheng or @qiuxin2012

yiheng:
The job failed for spark 1.5 and spark 1.6...

```
[ERROR] /opt/work/jenkins/workspace/BigDL-PRVN-scalatest-optim-Spark-1.6/spark/dl/src/test/scala/com/intel/analytics/bigdl/dlframes/DLClassifierSpec.scala:244: not found: value linalg
[ERROR]         val data = sc.parallelize(smallData.map(p => (new linalg.DenseVector(p._1), p._2)))
[ERROR]                                                           ^
[ERROR] /opt/work/jenkins/workspace/BigDL-PRVN-scalatest-optim-Spark-1.6/spark/dl/src/test/scala/com/intel/analytics/bigdl/dlframes/DLEstimatorSpec.scala:348: not found: value linalg
[ERROR]         val data = sc.parallelize(smallData.map(p => (linalg.Vectors.dense(p._1), p._2)))
[ERROR]                                                       ^
[ERROR] two errors found
```

yiheng:
can we skip the test on Spark 1.5 and Spark 1.6?

sperlingxx:
@yiheng  It doesn't work on spark 1.5.  And the test pieces for 1.6 and 2.x depend on different __DenseVector  Interface__. The below code will cause an exception when building on spark 1.6, I don't know how to do optional build/compilation on Scala :

> 

	val appSparkVersion = org.apache.spark.SPARK_VERSION
	val df: DataFrame = appSparkVersion.trim match {
	  case v if v.startsWith("1") =>
	    val data = sc.parallelize(smallData.map(p => (new DenseVector(p._1), p._2)))
	    sqlContext.createDataFrame(data).toDF("features", "label")
	  case v if v.startsWith("2") =>
	    val data = sc.parallelize(smallData.map(p => (new linalg.DenseVector(p._1), p._2)))
	    sqlContext.createDataFrame(data).toDF("features", "label")
	   }

jamesprinc3:
Is there anything fundamental still holding up this PR (other than the merge conflicts/excluding Spark 1.x tests?). Doesn't seem desirable to hold up a new feature due to backwards compatibility concerns?

jason-dai:
@jamesprinc3 you may wan to try `NNFrames` (https://analytics-zoo.github.io/master/#APIGuide/PipelineAPI/nnframes/)

jamesprinc3:
Thanks for the suggestion, seem like Analytics Zoo has a bunch of features that might be useful. As for this PR, is there any incentive to have this functionality in BigDL given that it's already in Analytics Zoo?

PR #2518, opened 2018-05-07
[python API]add batchSize support in model.predict

### LINKED ISSUE #2516: [python API] model.predict can't set batch size

qiuxin2012:

batchSizes support in model.predict

## How was this patch tested?

unit tests, manual tests

fixed #2516

### DIFF
--- a/pyspark/test/bigdl/test_simple_integration.py
+++ b/pyspark/test/bigdl/test_simple_integration.py
@@ -524,8 +524,8 @@ class TestSimple():
         for i in range(0, total_length):
             assert_allclose(p[i], ground_label[i], atol=1e-6, rtol=0)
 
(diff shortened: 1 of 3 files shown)


pyspark/test/bigdl/test_simple_integration.py
```diff
@@ -320,7 +320,7 @@ def gen_rand_sample():
 
         assert parameters["linear1"] is not None
         print("parameters %s" % parameters["linear1"])
-        predict_result = trained_model.predict(trainingData)
+        predict_result = trained_model.predict(trainingData, batch_size = 4)
```
yiheng: should we need to add a new test case instead of replace the original test batch_size = -1 case

qiuxin2012:
jenkins passed

PR #2520, opened 2018-05-08
[New Feature] Allow User to Specify Input Port when load Tensorflow model

yiheng:
The user can feed data into a middle operation in Tensorflow graph. However, if the operation has multiple inputs, the user needs to feed data for all inputs.

This PR allows the user to specify an input to feed data. The user can do this by using ```node_name:port``` format name when loading tensorflow model.

## How was this patch tested?
New unit test.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/TensorflowLoader.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/TensorflowLoader.scala
@@ -268,7 +268,7 @@ object TensorflowLoader{
                     s" ${inputNum - 1}"))
                 var i = 0
                 while (i < inputNum) {
-                  if(selectInputs.contains(i)) {
+                  if (selectInputs.contains(i)) {
                     val name = s"input$inputCounter"
                     val placeholder = NodeDef.newBuilder()
                       .setName(name)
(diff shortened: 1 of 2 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/TensorflowLoader.scala
```diff
@@ -44,7 +44,9 @@ object TensorflowLoader{
   /**
    * Load tensorflow model from a prototxt file
    * @param graphPrototxt where is the tensorflow protobuf file
-   * @param inputs input node names
+   * @param inputs input node names, where feed in the data. You can feed data into an internal
+   *               node. If the internal node has multiple dependency, you can use name:n to specify
```
qiuxin2012: what if user whats 2 of the inputs?

qiuxin2012: Array(name1:1, name1:2) ?

yiheng: yes

yiheng:
Jenkins pass

qiuxin2012:
LGTM

PR #2525, opened 2018-05-17
use google-analytics in bigdl doc website

yiheng:
Add google-analytics code to generate page view report of the document website

## How was this patch tested?
doc change. manual test


jason-dai:
Does it work for page view from PRC?

yiheng:
@jason-dai I think so

PR #2557, opened 2018-06-14
fix: memory leak in `model.predictImageSet`.

### LINKED ISSUE #2535: Memory leak when using quantized model for prediction
The memory consumption grows steadily as multiple inferences running with quantized model. 

After some initial investigation, found the heap memory was actually small and consistent while the used physical memory is steadily increased.

i8run:
There're three reasons of memory leak.

1. repeat allocations in bigquant, which will be fixed in BigDL-core.
2. repeat clone module but no release. `model.predictImageSet` will new
   Predictor again and again.
3. share weights.

Because of multi models architecture, every `LocalPredictor` will clone models.
For efficiency all cloned models and origin model will share the same parameters.
If we iterate and release all cloned models, the parameters of origin model will
be released also. So we should clone a new model first in `LocalPredictor`.

This patch add a `StorageManager` which contains a concurrent hash map
to maintain all allocations of native memory/resources and prevent
duplicate release. It's also helpful for debug.

## How was this patch tested?

Add a new test case and run Jenkins.

fixed #2535

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
@@ -47,7 +47,7 @@ class ModelBroadcast[T: ClassTag](applyProtoBuffer: Boolean = false)
   private var broadcastConsts: Broadcast[Map[String, Tensor[_]]] = _
   private var broadcastParameters: Broadcast[Array[Tensor[T]]] = _
 
-  val uuid: String = UUID.randomUUID().toString
+  private[bigdl] val uuid: String = UUID.randomUUID().toString
 
   /**
    * broadcast the model
@@ -182,7 +182,7 @@ object ModelBroadcast {
   }
 }
 
-class ModelInfo[T: ClassTag](val uuid: String, @transient var model: Module[T])(
+private[bigdl] class ModelInfo[T: ClassTag](val uuid: String, @transient var model: Module[T])(
   implicit ev: TensorNumeric[T]) extends Serializable {
   @throws(classOf[IOException])
   private def writeObject(out: ObjectOutputStream): Unit = {
@@ -193,7 +193,7 @@ class ModelInfo[T: ClassTag](val uuid: String, @transient var model: Module[T])(
   }
 }
 
-object ModelInfo {
+private[bigdl] object ModelInfo {
   def apply[T: ClassTag](uuid: String, model: Module[T])(
     implicit ev: TensorNumeric[T]): ModelInfo[T] = new ModelInfo[T](uuid, model)
 }
@@ -204,61 +204,31 @@ object CachedModels {
   import scala.collection._
   import scala.collection.convert.decorateAsScala._
 
-  val lock = new AnyRef
+  type Modles = ArrayBuffer[Module[_]]
 
-  type FloatValues = ArrayBuffer[Module[Float]]
-  type DoubleValues = ArrayBuffer[Module[Double]]
-
-  private val cachedFloatModels: concurrent.Map[String, FloatValues] =
-    new ConcurrentHashMap[String, FloatValues]().asScala
-  private val cachedDoubleModels: concurrent.Map[String, DoubleValues] =
-    new ConcurrentHashMap[String, DoubleValues]().asScala
+  private val cachedModels: concurrent.Map[String, Modles] =
+    new ConcurrentHashMap[String, Modles]().asScala
 
   def add[T: ClassTag](uuid: String, model: Module[T])( implicit ev: TensorNumeric[T]): Unit =
-    lock.synchronized {
-      ev.getType() match {
-        case FloatType =>
-          val models = cachedFloatModels.get(uuid) match {
-            case Some(values) => values += model.asInstanceOf[Module[Float]]
-            case _ => ArrayBuffer(model.asInstanceOf[Module[Float]])
-          }
-          cachedFloatModels.put(uuid, models)
-        case DoubleType =>
-          val models = cachedDoubleModels.get(uuid) match {
-            case Some(values) => values += model.asInstanceOf[Module[Double]]
-            case _ => ArrayBuffer(model.asInstanceOf[Module[Double]])
-          }
-          cachedDoubleModels.put(uuid, models)
-        case _ => throw new UnsupportedOperationException(s"unsupported type")
+    CachedModels.synchronized {
+      val models = cachedModels.get(uuid) match {
+        case Some(values) => values += model
+        case _ => ArrayBuffer(model)
       }
     }
 
   def deleteAll[T: ClassTag](currentKey: String)(implicit ev: TensorNumeric[T]): Unit =
-    lock.synchronized {
-      ev.getType() match {
-        case FloatType =>
-          val keys = cachedFloatModels.keys
-          for (key <- keys) {
-            if (key != currentKey) {
-              val models = cachedFloatModels(key)
-              println(s"delete key = $key ${models.length}")
-              for (model <- models) { model.release()
-                println(StorageManager.get().count(!_._2.isFreed))
-              }
-              cachedFloatModels.remove(key)
-              println(StorageManager.get().count(!_._2.isFreed))
-            }
+    CachedModels.synchronized {
+      val keys = cachedModels.keys
+      for (key <- keys) {
+        if (key != currentKey) {
+          val models = cachedModels(key)
+          println(s"delete key = $key ${models.length}")
+          for (model <- models) {
+            model.release()
           }
-        case DoubleType =>
-          val keys = cachedDoubleModels.keys
-          for (key <- keys) {
-            if (key != currentKey) {
-              val models = cachedDoubleModels(key)
-              for (model <- models) { model.release() }
-              cachedDoubleModels.remove(key)
-            }
-          }
-        case _ => throw new UnsupportedOperationException(s"unsupported type")
+          cachedModels.remove(key)
+        }
       }
     }
 }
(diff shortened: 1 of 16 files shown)


jason-dai:
Does distributed model predict have similar issues?

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/Tensor.scala
```diff
@@ -804,6 +804,8 @@ trait Tensor[T] extends Serializable with TensorMath[T] with Activity {
     }
     return false
   }
+
+  def toQuantizedTensor: QuantizedTensor[T]
```
yiheng: What's this for?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Util.scala
```diff
@@ -166,6 +166,10 @@ object Util {
     var i = 0
     while (i < tensors.length) {
       if (tensors(i) != null) {
+        if (tensors(i).getTensorType == QuantizedType) {
+          tensors(i).toQuantizedTensor.release()
```
yiheng: asInstanceof[...]?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Util.scala
```diff
@@ -179,10 +183,23 @@ object Util {
...
     while (i < localWeightBias.length) {
       if (localWeightBias(i) != null) {
-        localWeightBias(i).set(broadcastWeightBias(i))
+        clearAndSet(localWeightBias(i), broadcastWeightBias(i))
       }
       i += 1
     }
+
+    def clearAndSet(old: Tensor[T], other: Tensor[T]): Unit = {
+      if (old.getTensorType == QuantizedType && other.getTensorType == QuantizedType) {
```
yiheng: do this in the set method of quantized tensor?

i8run: It can't do `release` in the quentized tensor set method. Because if we have tensor A, B, C, than we do `A.set(B)` and `A.set(C)`. At the end, the `A` and `B` will be all released. This is not we want because we may want to reserve tensor `B`. So we only release the tensor carefully when needed.

i8run:
@jason-dai Yes, it also has this problem. We're working on it. But it can not be fixed with the same method of `LocalPredictor`.  Because RDD is lazy. We can't release the resources in `mapPartition`, otherwise the `release` will be called before main part of `mapPartition`.

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
```diff
@@ -53,26 +59,45 @@ class ModelBroadcast[T: ClassTag](applyProtoBuffer: Boolean = false)
    * @return this
    */
   def broadcast(sc: SparkContext, model: Module[T]): this.type = {
+    CachedModels.deleteAll(uuid) // delete the models on driver
+
     if (applyProtoBuffer) {
-      broadcastModel = sc.broadcast(model)
+      broadcastModel = sc.broadcast(ModelInfo(uuid, model))
```
wzhongyuan: Why not add into `CachedModels` ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
```diff
@@ -141,13 +173,92 @@ class ModelBroadcast[T: ClassTag](applyProtoBuffer: Boolean = false)
...
 
-
 object ModelBroadcast {
   def apply[@specialized(Float, Double) T: ClassTag](applyProtoBuffer: Boolean = false)
         (implicit ev: TensorNumeric[T]) : ModelBroadcast[T] = {
     new ModelBroadcast(applyProtoBuffer)
   }
 }
+
+class ModelInfo[T: ClassTag](val uuid: String, @transient var model: Module[T])(
```
wzhongyuan: case class ?

wzhongyuan: Why transient ? will it be ignored while serialized and deserialized between driver and executor ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
```diff
@@ -141,13 +173,92 @@ class ModelBroadcast[T: ClassTag](applyProtoBuffer: Boolean = false)
...
+  def deleteAll[T: ClassTag](currentKey: String)(implicit ev: TensorNumeric[T]): Unit =
+    lock.synchronized {
+      ev.getType() match {
+        case FloatType =>
+          val keys = cachedFloatModels.keys
+          for (key <- keys) {
+            if (key != currentKey) {
+              val models = cachedFloatModels(key)
+              println(s"delete key = $key ${models.length}")
+              for (model <- models) { model.release()
```
wzhongyuan: Does `release` compatible for those models not quantized ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
```diff
@@ -141,13 +173,92 @@ class ModelBroadcast[T: ClassTag](applyProtoBuffer: Boolean = false)
...
+                println(StorageManager.get().count(!_._2.isFreed))
```
wzhongyuan: remove the `println` clause ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
```diff
@@ -141,13 +173,92 @@ class ModelBroadcast[T: ClassTag](applyProtoBuffer: Boolean = false)
...
+              }
+              cachedFloatModels.remove(key)
+              println(StorageManager.get().count(!_._2.isFreed))
```
wzhongyuan: remove the `println` clause ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/quantized/Desc.scala
```diff
@@ -178,3 +181,28 @@ object QuantParams {
   val THRESHOLD = 127.0f
 }
 
+case class StorageInfo(descType: DescType, isFreed: Boolean)
+
+object StorageManager {
+  import java.util.concurrent.ConcurrentHashMap
+  private val nativeStorages: ConcurrentHashMap[Long, StorageInfo] = new ConcurrentHashMap()
+
+  def isFreed(nativeStorage: Long): Boolean = {
```
wzhongyuan: Add comments.. also `isReleased` makes more sense ?

yiheng:
Do you miss the change in bigdl core?

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
```diff
@@ -38,10 +43,11 @@ import scala.reflect.ClassTag
 class ModelBroadcast[T: ClassTag](applyProtoBuffer: Boolean = false)
   (implicit ev: TensorNumeric[T]) extends Serializable {
 
-  private var broadcastModel: Broadcast[Module[T]] = _
+  private var broadcastModel: Broadcast[ModelInfo[T]] = _
   private var broadcastConsts: Broadcast[Map[String, Tensor[_]]] = _
   private var broadcastParameters: Broadcast[Array[Tensor[T]]] = _
 
+  val uuid: String = UUID.randomUUID().toString
```
yiheng: private?

yiheng: why it's a `val`, for different predictions, should we use different uuid?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/quantized/Desc.scala
```diff
@@ -178,3 +181,28 @@ object QuantParams {
   val THRESHOLD = 127.0f
 }
 
+case class StorageInfo(descType: DescType, isFreed: Boolean)
+
+object StorageManager {
```
yiheng: private[bigdl]?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/quantized/Desc.scala
```diff
@@ -178,3 +181,28 @@ object QuantParams {
   val THRESHOLD = 127.0f
 }
 
+case class StorageInfo(descType: DescType, isFreed: Boolean)
```
yiheng: private[bigdl]?

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
```diff
@@ -141,13 +173,92 @@ class ModelBroadcast[T: ClassTag](applyProtoBuffer: Boolean = false)
...
+
+  import scala.collection._
+  import scala.collection.convert.decorateAsScala._
+
+  val lock = new AnyRef
+
+  type FloatValues = ArrayBuffer[Module[Float]]
+  type DoubleValues = ArrayBuffer[Module[Double]]
+
+  private val cachedFloatModels: concurrent.Map[String, FloatValues] =
```
yiheng: Why there're two arraybuffer here?

yiheng:
should we also release memory in DistOptimizer and LocalOptimizer?

wzhongyuan:
I refined per review and merged to @i8run 's branch, please run a PR job

wzhongyuan:
PR validation passed http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/72/

PR #2559, opened 2018-06-21
add parameter sync for batchnorm

wzhongyuan:
1) Add parameter synchronizer
2) apply in batch normalization layers so that we can use small batch size to leverage more computing resources
## How was this patch tested?

Unit test, Resnet-50 training


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/ParameterSynchronizer.scala
```diff
@@ -0,0 +1,97 @@
...
+
+import scala.reflect._
+
+object ParameterSynchronizer {
+
+  val fEvents = new java.util.concurrent.ConcurrentHashMap[String, Event[Float]]
+
+  val dEvents = new java.util.concurrent.ConcurrentHashMap[String, Event[Double]]
+
+  def register[T: ClassTag](eventKey: String, threadNum: Int): Unit = {
```
qiuxin2012: all public methods should have comments

qiuxin2012:
What will happen if one of the executors is lost?

wzhongyuan:
@qiuxin2012  it has nothing to do with executor lost, as the synchronization is inside executor , we don't do cross-executor sync

wzhongyuan:
PR validation passed , http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/55/

spark/dl/src/test/scala/com/intel/analytics/bigdl/nn/SpatialBatchNormalizationSpec.scala
```diff
@@ -17,13 +17,116 @@ package com.intel.analytics.bigdl.nn
...
+
+    val gradInput = bn.backward(input, gradOutput)
+
+    val saveMean = bn.saveMean
+    val saveStd = bn.saveStd
+    val runningMean = bn.runningMean
+    val runningVar = bn.runningVar
+
+    val bn1 = SpatialBatchNormalization[Float](2)
+    bn1.setParallism(Some(2))
```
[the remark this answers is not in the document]
wzhongyuan: Yes, it's an optimization, previously we just use Enginde,coreNumber by default, but it relies on Engine and does not make sense to expose to layer itselft , also it's not flexible,

wzhongyuan:
@qiuxin2012  Please let me know if you have other concerns, thanks

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/BatchNormalization.scala
```diff
@@ -61,6 +61,19 @@ class BatchNormalization[T: ClassTag](
 
   require(nOutput > 0, "output feature map number must be greater than zero")
 
+  private var parallism : Option[Int] = None
+
+  def setParallism(parallism: Option[Int]): Unit = {
```
qiuxin2012: comments for public methods and variables

yiheng: Maybe
```scala
def setParallism(parallism: Int): Unit = {
    this.parallism = Some(parallism)
}
```
is more user friendly. And should it be private?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/BatchNormalization.scala
```diff
@@ -209,21 +246,37 @@ class BatchNormalization[T: ClassTag](
     saveMean.resizeAs(runningMean).zero
     saveStd.resizeAs(runningVar).fill(ev.zero)
 
+    val nChannels = _input.size(2)
```
yiheng: should we requirement the format is NCHW?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/BatchNormalization.scala
```diff
@@ -243,25 +296,39 @@ class BatchNormalization[T: ClassTag](
...
+    val gmKeyWithId = s"${this.gmKey}_${this.getId}"
+    val gxmKeyWithId = s"${this.gxmKey}_${this.getId}"
+    val needSync = getParallism() != None && getParallism().get > 1
     _gradOutput.set(gradOutput)
     makeBatch(_gradOutput)
     _gradOutput.addSingletonDimension(_gradOutput, 3)
     _gradOutput.addSingletonDimension(_gradOutput, 4)
     gxMean.zero()
     gMean.zero()
+    val nChannel = _gradOutput.size(2)
```
yiheng: should we require the input is NCHW?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/ParameterSynchronizer.scala
```diff
@@ -0,0 +1,132 @@
...
+
+import com.intel.analytics.bigdl.tensor.Tensor
+import org.apache.spark.TaskContext
+
+import scala.reflect._
+
+/**
+ * A parameter synchronizer among threads per task
+ */
+object ParameterSynchronizer {
```
yiheng: private[bigdl] ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/ParameterSynchronizer.scala
```diff
@@ -0,0 +1,132 @@
...
+
+  val fEvents = new java.util.concurrent.ConcurrentHashMap[String, Event[Float]]
```
yiheng: why there're two kind of events?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/ParameterSynchronizer.scala
```diff
@@ -0,0 +1,132 @@
...
+      if (event == null) {
+        event = new Event[Float](threadNum)
+        fEvents.putIfAbsent(eventKey, event)
+      }
+    } else if (classTag[T] ==  classTag[Double]) {
+      var event = dEvents.get(eventKey)
+      if (event == null) {
+        event = new Event[Double](threadNum)
+        dEvents.putIfAbsent(eventKey, event)
+      }
```
yiheng: else throw exception?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/ParameterSynchronizer.scala
```diff
@@ -0,0 +1,132 @@
...
+      fEvents.get(eventKey).data.asInstanceOf[java.util.Map[String, Tensor[T]]]
+    } else if (classTag[T] ==  classTag[Double]) {
+      dEvents.get(eventKey).data.asInstanceOf[java.util.Map[String, Tensor[T]]]
+    } else {
+      null
+    }
+  }
+}
+
+class Event[T: ClassTag](threadNum: Int) {
```
yiheng: private?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/BatchNormalization.scala
wzhongyuan: i's already separated in NCHW related methods, not impact on others

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/BatchNormalization.scala
wzhongyuan: it's already separated in NCHW related methods, not impact on others

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/ParameterSynchronizer.scala
wzhongyuan: we cannot throw exception here ,  it's for race condition and make sure the thread safety

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/ParameterSynchronizer.scala
wzhongyuan: For different types, as `SpatialBatchNormalization` implements explicitly type specific methods, this way would reduce the implicit conversion cost

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/BatchNormalization.scala
wzhongyuan: Makes sense, we cannot make it private , it's an exposed interface

wzhongyuan:
@yiheng ,  Ran another PR job and passed after refinement http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/61/ 

Please let me know if any other issue

wzhongyuan:
Per review refinement PR validation passed 
http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/65/

PR #2275, opened 2018-02-05
multi optimMethods in Optimizer

qiuxin2012:
## What changes were proposed in this pull request?

multi optimMethods support in Optimizer

Add a new interface `def setOptimMethods(method: Map[String, OptimMethod[T]])` to set multi OptimMethods in Optimizer.
With this API, user can set different optimMethods for each subModules with the subModules' name and optimMethod pair. 

## How was this patch tested?

unit tests

## Related links or issues (optional)
fixed #2171

PR #2560, opened 2018-06-21
[new feature] multi optimMethods support in Optimizer

### LINKED ISSUE #2171: Support multi-OptimMethod in one model.
For model like wide&deep can use more than one OptimMethods in model's different parts. Like wide&deep model,  ftrl for wide part, adagrad for deep part.

qiuxin2012:

multi optimMethods support in Optimizer

Add a new interface def setOptimMethods(method: Map[String, OptimMethod[T]]) to set multi OptimMethods in Optimizer.
With this API, user can set different optimMethods for each subModules with the subModules' name and optimMethod pair.

Updated #2275

## How was this patch tested?

unit test

fixed #2171

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
@@ -997,6 +997,8 @@ class DistriOptimizer[T: ClassTag] (
             lastFailureTimestamp = System.nanoTime()
 
             val modelFile = getLatestFile(checkpointPath.get, "model")
+            clearState()
+            models.unpersist()
             val newModel = if (modelFile != null) {
               DistriOptimizer.logger.info("Model recover from last snapshot")
               Module.load[T](modelFile)
@@ -1006,8 +1008,6 @@ class DistriOptimizer[T: ClassTag] (
             }
             optimMethods = optimMethods.map { case (moduleName, optimMethod) =>
               val methodFile = getLatestFile(checkpointPath.get, s"optimMethod-$moduleName")
-              clearState()
-              models.unpersist()
 
               val newOptimMethod = if (methodFile != null) {
                 DistriOptimizer.logger.info(s"$moduleName's OptimMethod recover from last snapshot")
@@ -1018,9 +1018,6 @@ class DistriOptimizer[T: ClassTag] (
               }
               newOptimMethod.clearHistory()
               (moduleName, newOptimMethod)
-            }
-            optimMethods.values.foreach{
-              _.clearHistory()
             }
             models = DistriOptimizer.initThreadModels(newModel, distDataset, criterion, state,
               nodeNumber, coresPerNode, checkSingleton, parameters, validationMethods,
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
@@ -343,7 +343,7 @@ object DistriOptimizer {
               System.nanoTime() - getG)
           }
           parameters.foreach { p =>
-            parameterProcessers.foreach(_.collectGlobalData(models, p._2, metrics, driverState))
+            parameterProcessers.foreach(_.processParameters(p._2, modelCache, driverState))
           }
           modelCache.optimMethod.foreach{ case (name, optimMethod) =>
             var time = System.nanoTime()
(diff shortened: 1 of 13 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -62,7 +65,7 @@ object DistriOptimizer {
     localStates: Array[Table],
     var moduleTimeList: Array[Long] = null,
     localMethods: Array[Option[Array[ValidationMethod[T]]]],
-    optimMethod: OptimMethod[T]
+    optimMethod: Map[String, OptimMethod[T]]
```
wzhongyuan: optimMethods ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -112,24 +118,26 @@ object DistriOptimizer {
...
-    if (!optimMethod.state.contains("epoch")) optimMethod.state.update("epoch", 1)
-    if (!optimMethod.state.contains("neval")) optimMethod.state.update("neval", 1)
-    if (!optimMethod.state.contains("Loss")) {
-      optimMethod.state.update("Loss", Float.PositiveInfinity)
-    }
-    if (!optimMethod.state.contains("score")) optimMethod.state.update("score", 0f)
-    if (!optimMethod.state.contains("recordsProcessedThisEpoch")) {
-      optimMethod.state.update("recordsProcessedThisEpoch", 0)
+    optimMethods.values.foreach{ optimMethod =>
+      if (!optimMethod.state.contains("epoch")) optimMethod.state.update("epoch", 1)
```
wzhongyuan: Call optimMethods share states as all seesm to be duplicated here

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -409,13 +435,14 @@ object DistriOptimizer {
...
 
-        optimMethod.state.update("recordsProcessedThisEpoch", recordsProcessedThisEpoch)
-
-        optimMethod.state.update("epoch", driverState[Int]("epoch"))
-        optimMethod.state.update("neval", driverState[Int]("neval"))
-        optimMethod.state.update("Loss", driverState[Float]("Loss"))
-        if (validationMethods.isDefined) {
-          optimMethod.state.update("score", driverState[Float]("score"))
+        optimMethods.map { case (moduleName, optimMethod) =>
+          optimMethod.state.update("recordsProcessedThisEpoch", recordsProcessedThisEpoch)
```
wzhongyuan: Can optimMethods share the same state ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Optimizer.scala
```diff
@@ -227,22 +236,42 @@ abstract class Optimizer[T: ClassTag, D](
...
+    if (optimMethods.size == 1 && optimMethods.contains(model.getName())) {
+      if (newModel.getName() != model.getName()) {
+        optimMethods = Map(newModel.getName() -> optimMethods(model.getName()))
+      }
+      logger.info(s"Optimizer.setModel: Detect current optimMethod is a global optimMethod." +
+        s" Automatically associate the current optimMethod with the new model.")
+    } else {
+      logger.warn(s"Optimizer.setModel: Detect current optimMethod is not a global optimMethod." +
+        s" Clearing the old optimMethod.")
+      optimMethods = Map[String, OptimMethod[T]]()
```
yiheng: should we build a newModelName -> optimMethod map here?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/LocalOptimizer.scala
```diff
@@ -87,8 +87,20 @@ class LocalOptimizer[T: ClassTag] (
...
     var wallClockTime = 0L
     var count = 0
-    optimMethod.clearHistory()
-    optimMethod.loadFromTable(state)
+    optimMethods.values.foreach{ optimMethod =>
+      optimMethod.clearHistory()
+    }
+
+    // To be compatible with the old usage that user define hyperparameters in a table.
+    if (optimMethods.size == 1) {
```
yiheng: should we throw exception or print warnings if the size is not 1?

yiheng:
please check my comments

wzhongyuan:
Can we run a full training on a specific model to check the compatibility ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
qiuxin2012: We can't, the state also saves some internal buffer, the buffer will conflict if we share states.

qiuxin2012:
jenkins passed

PR #2561, opened 2018-06-26
Use a separated java gateway on local mode

yiheng:
In Spark 2.3, the py4j is upgraded. The gateway server authentication is enabled, which is not supported in the old py4j gateway server.

To keep the compatibility with Spark 2.0/2.1/2.2, we still use the old way to create py4j gateway server. It works fine in most cases. As on distributed mode, we'll create our own gateway server on worker JVM. In driver python process, we'll use the gateway client created by Spark.

However, in local mode, the worker python process will create its own gateway client, which connect to the gateway server created by Spark. This PR fixes this. It will create a separated gateway server for usage.

## How was this patch tested?
unit test


qiuxin2012:
LGTM

yiheng:
jenkins pass

PR #2580, opened 2018-07-05
fix: distribute predictor will send model twice when do `mapPartition`.

i8run:
The `predictClass`, `predict` and `predictImage` will call the relevant methods
of object `Predictor`. Why we do this? Because every these methods uses the
ClassTag `T`. If we do these jobs in the methods of class`Predictor`, when we do
`mapPartition`, Spark will find all used values and do serialization. The `T` is
the argument of constructor, the serialization will package the whole
`Predictor` class, which contains the`model`.  Then it will send a duplicate model
to the workers. So we should move these methods to object `Predictor`.

## How was this patch tested?

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Predictor.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Predictor.scala
@@ -16,18 +16,15 @@
 
 package com.intel.analytics.bigdl.optim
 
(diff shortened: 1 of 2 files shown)


yiheng:
How do you verify your change?

yiheng:
Where's the code release the native resource?

i8run:
If we send twice of the model, some of existed test cases will be failure.

Issue #2569, opened 2018-06-29
`MM` layer cannot be forward multiple times.


zhichao-li:
```
out1 = mm.forward(input)
out2 = mm.forward(input)
```

out1 and out2 do not share the same value.

wzhongyuan:
@zhichao-li , i ran test but got the same result

```scala
  "MM forward multi times" should "work properly" in {
    val mm = MM[Float]()
    val input1 = Tensor[Float](3, 3).randn()
    val input2 = Tensor[Float](3, 3).randn()
    val input = T(1 -> input1, 2 -> input2)

    val res1 = mm.forward(input)

    println(res1)

    val res2 = mm.forward(input)

    println(res2)
  }
```

zhichao-li:
try to test with batched input

PR #2583, opened 2018-07-06
[Issue fix] - Fix MM layer multi forward/backward issue

wzhongyuan:

Fix the issue so that MM can be forward/backward multi times

## How was this patch tested?

unit test

fixed https://github.com/intel-analytics/BigDL/issues/2569

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/MM.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/MM.scala
@@ -53,6 +53,8 @@ class MM[T: ClassTag](
   override def updateOutput(input: Table): Tensor[T] = {
     var (ma, mb) = checkInputFormat(input)
 
+    output.set
+
     if (ma.dim() == 2) {
       require(mb.dim() == 2, "second input tensor must be 2D" +
         s"second input dim ${mb.dim()}")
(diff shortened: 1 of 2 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/MM.scala
```diff
@@ -53,6 +53,8 @@ class MM[T: ClassTag](
   override def updateOutput(input: Table): Tensor[T] = {
     var (ma, mb) = checkInputFormat(input)
 
+    output.set
```
yiheng: set will change memory. should we avoid that? Another fix is using ```baddbmm``` instead of ```bmm```. We can set the ```beta == 0``` and ```alpha == 1```.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/MM.scala
```diff
@@ -91,6 +93,10 @@ class MM[T: ClassTag](
   }
 
   override def updateGradInput(input: Table, gradOutput: Tensor[T]): Table = {
+
+    gradInput[Tensor[T]](1).set
```
yiheng: same with the above comments

wzhongyuan:
@yiheng  That would be better. I did not know the referred method...

I am running another PR

wzhongyuan:
PR validation passed http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/93/

PR #2585, opened 2018-07-06
[Bug Fix]clear preTopology's output while cloneCells

qiuxin2012:

Recurrent cost too many memories, find it's caused by cloneCells. 
clear preTopology's output while cloneCells.

With this fix, 9-depth deepspeech2 model forward a input tensor of size  (1, 1, 13, 3000) cost only 1GB memory.
Without this fix, the model will cost about 80GB memory.

## How was this patch tested?

unit tests, manual tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Recurrent.scala
@@ -174,7 +174,11 @@ class Recurrent[T : ClassTag](
       cloneCell.parameters()._1.map(_.set())
       cloneCell.parameters()._2.map(_.set())
       // preTopology's output is useless here, clear it.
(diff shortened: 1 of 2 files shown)


qiuxin2012:
@yiheng failed, some times preTopology is null.

qiuxin2012:
jenkins passed

zhichao-li:
@qiuxin2012 @yiheng 
Would it be better if we set preTopology to be null in `add` method as there's weights, grad or other stuff within a preTopology ` other than `output`

```
  override def add(module: AbstractModule[_ <: Activity, _ <: Activity, T]): this.type = {
    if (topology.preTopology != null) {
      val tmp = topology.preTopology.cloneModule()
      topology.preTopology = null 
      preTopology = TimeDistributed(tmp, maskZero = maskZero)
    }
```

PR #2586, opened 2018-07-06
delete logger variable  in LocalPredictor

### LINKED ISSUE #2578: LocalPredictor can not be serialized because of field logger

qiuxin2012:

delete logger variable in LocalPredictor

## How was this patch tested?

unit tests

fixed #2578


yiheng:
why remove the logger? I don't think this is the right way to fix the issue

qiuxin2012:
The right way to use logger is new a logger in the Object, and import the logger in the corresponding Class.
So making the logger a member of the Class is not a good idea. @yiheng

yiheng:
Yes, you should move the logger into the object.

qiuxin2012:
jenkins passed

PR #2598, opened 2018-07-18
Dnn model serialization supports.

i8run:

Support serialization for native dnn model. Especially make the primitive and primitive desc to transient.
Because it only influences the layer which has weights. So only add some test cases for these layers.

## How was this patch tested?

Jenkins.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DnnStorage.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DnnStorage.scala
@@ -15,6 +15,8 @@
  */
 package com.intel.analytics.bigdl.tensor
 
+import java.io.{IOException, ObjectInputStream, ObjectOutputStream}
+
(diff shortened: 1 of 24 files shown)


i8run:
Jenkins has passed.

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DnnStorage.scala
```diff
@@ -31,7 +33,7 @@ private[tensor] class DnnStorage[T: ClassTag](size: Int) extends Storage[T] {
   private var _isReleased: Boolean = false
 
   // Hold the address of the native array
-  val ptr: Pointer = new Pointer(allocate(size))
+  @transient lazy val ptr: Pointer = new Pointer(allocate(size))
```
yiheng: please remove the lazy.

i8run:
Jenkins passed with modified code based on comments. :)

Issue #2539, opened 2018-05-23
Sometimes python code may keep printing 'Adding ... to SPARK_CLASSPATH'


hkvision:
```
2018-05-18 17:43:33 INFO  DistriOptimizer$:871 - caching training rdd ...
[Stage 0:>                                                          (0 + 8) / 8]Adding /home/lizhichao/bin/god/zoo/dist/lib/analytics-zoo-0.1.0-SNAPSHOT-jar-with-dependencies.jar to BIGDL_JARSAdding /home/lizhichao/bin/god/zoo/dist/lib/analytics-zoo-0.1.0-SNAPSHOT-jar-with-dependencies.jar to BIGDL_JARS

Adding /home/lizhichao/bin/god/zoo/dist/lib/analytics-zoo-0.1.0-SNAPSHOT-jar-with-dependencies.jar to BIGDL_JARSAdding /home/lizhichao/bin/god/zoo/dist/lib/analytics-zoo-0.1.0-SNAPSHOT-jar-with-dependencies.jar to BIGDL_JARS

Adding /home/lizhichao/bin/god/zoo/dist/lib/analytics-zoo-0.1.0-SNAPSHOT-jar-with-dependencies.jar to BIGDL_JARSAdding /home/lizhichao/bin/god/zoo/dist/lib/analytics-zoo-0.1.0-SNAPSHOT-jar-with-dependencies.jar to BIGDL_JARS

Adding /home/lizhichao/bin/god/zoo/dist/lib/analytics-zoo-0.1.0-SNAPSHOT-jar-with-dependencies.jar to SPARK_CLASSPATH
Adding /home/lizhichao/bin/god/zoo/dist/lib/analytics-zoo-0.1.0-SNAPSHOT-jar-with-dependencies.jar to SPARK_CLASSPATHAdding /home/lizhichao/bin/god/zoo/dist/lib/analytics-zoo-0.1.0-SNAPSHOT-jar-with-dependencies.jar to SPARK_CLASSPATH

Adding /home/lizhichao/bin/god/zoo/dist/lib/analytics-zoo-0.1.0-SNAPSHOT-jar-with-dependencies.jar to SPARK_CLASSPATH
Adding /home/lizhichao/bin/god/zoo/dist/lib/analytics-zoo-0.1.0-SNAPSHOT-jar-with-dependencies.jar to SPARK_CLASSPATH
[...]

PR #2603, opened 2018-07-28
quick fix for 2539

megaSpoon:

fix issue 2539(https://github.com/intel-analytics/BigDL/issues/2539)
quick workaround. But the root cause is not revealed.

fix issue 2539(https://github.com/intel-analytics/BigDL/issues/2539)

### DIFF
--- a/pyspark/bigdl/util/engine.py
+++ b/pyspark/bigdl/util/engine.py
@@ -67,9 +67,9 @@ def __prepare_bigdl_env():
 
     def append_path(env_var_name, path):
         try:
-	    if path not in os.environ[env_var_name]:
-	            print("Adding %s to %s" % (path, env_var_name))
-	            os.environ[env_var_name] = path + ":" + os.environ[env_var_name]  # noqa
+            if path not in os.environ[env_var_name]:
+	        print("Adding %s to %s" % (path, env_var_name))
+	        os.environ[env_var_name] = path + ":" + os.environ[env_var_name]  # noqa
         except KeyError:
             os.environ[env_var_name] = path
 
--- a/pyspark/bigdl/util/engine.py
+++ b/pyspark/bigdl/util/engine.py
@@ -41,6 +41,16 @@ def check_spark_source_conflict(spark_home, pyspark_path):
         warnings.warn(warning_msg)
 
 
+def __sys_path_insert(file_path):
+    try:
+        if file_path not in sys.path.split(":"):
+            print("Prepending %s to sys.path" % file_path)
+            sys.path.insert(0, file_path)
+    except:
+        raise RuntimeError("The %s has already been inserted into the sys.path" % file_path)
+
+
+
 def __prepare_spark_env():
     spark_home = os.environ.get('SPARK_HOME', None)
     if exist_pyspark():
@@ -56,8 +66,8 @@ def __prepare_spark_env():
         print("Using %s" % spark_home)
         py4j = glob.glob(os.path.join(spark_home, 'python/lib', 'py4j-*.zip'))[0]
         pyspark = glob.glob(os.path.join(spark_home, 'python/lib', 'pyspark*.zip'))[0]
-        sys.path.insert(0, py4j)
-        sys.path.insert(0, pyspark)
+        __sys_path_insert(py4j)
+        __sys_path_insert(pyspark)
 
 
 def __prepare_bigdl_env():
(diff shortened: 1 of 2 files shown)


pyspark/bigdl/util/engine.py
```diff
@@ -67,8 +67,9 @@ def __prepare_bigdl_env():
 
     def append_path(env_var_name, path):
         try:
-            print("Adding %s to %s" % (path, env_var_name))
-            os.environ[env_var_name] = path + ":" + os.environ[env_var_name]  # noqa
+            if path not in os.environ[env_var_name]:
```
yiheng: what if path is a substring of another in the env_var_name, e.g. your path is "/ABC/DEF", and the env_var_name is "/ABC/DEF/GHI:/DDD/XXX". The path is not in the env but your check will consider it in the env, right?

yiheng:
Please add unit tests for your change.

yiheng:
@hkiang01 any comments?

hkiang01:
@yiheng I think you got the wrong guy

yiheng:
@hkiang01 Sorry...

yiheng:
@hkvision any comments?

hkvision:
LGTM

hkvision:
Just to be safe, can we check this as well when inserting into sys.path? i.e. https://github.com/megaSpoon/BigDL/blob/19437a436dc5772a71aa360fd9a7e5b94ab05998/pyspark/bigdl/util/engine.py#L59
https://github.com/megaSpoon/BigDL/blob/19437a436dc5772a71aa360fd9a7e5b94ab05998/pyspark/bigdl/util/engine.py#L82
https://github.com/megaSpoon/BigDL/blob/19437a436dc5772a71aa360fd9a7e5b94ab05998/pyspark/bigdl/util/engine.py#L90

pyspark/bigdl/util/engine.py
hkvision: What you doubt is reasonable. But in actual case, it is used to adding jar to SPARK_CLASSPATH, will this scenario happen?

yiheng: that should be fine. But the code could not tell that. Should we change the path to jar_path to make it more clearer?

hkvision: I'm fine for changing the name to `jar_path` as `append_path` is used for adding jars only.

megaSpoon: we can solve it by checking path split by ":" to avoid the substring flaw.

megaSpoon:
add codes to check insertion into sys.path and unit tests for it.

about the unit test:
The prepare_env() will be called when we import bigdl so I just called those functions in the engine which  append or insert path to the environment variables to see if they raise the expected exception.

pyspark/test/bigdl/test_engine_env.py
```diff
@@ -0,0 +1,49 @@
...
+import glob
+import pyspark
+
+
+class TestEngineEnv():
+    def setup_method(self, method):
+        """ setup any state tied to the execution of the given method in a
+        class.  setup_method is invoked for every test method of a class.
+        """
+        self.spark_home = '/home/megaspoon/Documents/spark-2.0.1-bin-hadoop2.7'
```
hkvision: change the spark home?

pyspark/bigdl/util/engine.py
```diff
@@ -41,6 +41,16 @@ def check_spark_source_conflict(spark_home, pyspark_path):
         warnings.warn(warning_msg)
 
 
+def __sys_path_insert(file_path):
+    try:
+        if file_path not in sys.path.split(":"):
+            print("Prepending %s to sys.path" % file_path)
+            sys.path.insert(0, file_path)
+    except:
```
hkvision: Why will this raise an exception? In the try block even if file_path is in sys.path, I suppose it won't throw an error but just not go into the if branch?

hkvision: I mean just add the if condition as well when inserting to sys.path...

pyspark/bigdl/util/engine.py
```diff
@@ -41,6 +41,16 @@ def check_spark_source_conflict(spark_home, pyspark_path):
         warnings.warn(warning_msg)
 
 
+def __sys_path_insert(file_path):
+    try:
+        if file_path not in sys.path.split(":"):
```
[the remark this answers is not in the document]
megaSpoon: Yep.

pyspark/test/bigdl/test_engine_env.py
```diff
@@ -0,0 +1,46 @@
...
+    def teardown_method(self, method):
+        """ teardown any state that was previously setup with a setup_method
+        call.
+        """
+        pass
+
+    def test___sys_path_insert(self):
+        py4j = glob.glob(os.path.join(self.spark_home, 'python/lib', 'py4j-*.zip'))[0]
+        pyspark = glob.glob(os.path.join(self.spark_home, 'python/lib', 'pyspark*.zip'))[0]
+        with pytest.raises(RuntimeError):
```
hkvision: will this still raise an error?

pyspark/test/bigdl/test_engine_env.py
```diff
@@ -0,0 +1,46 @@
...
+        pass
+
+    def teardown_method(self, method):
+        """ teardown any state that was previously setup with a setup_method
+        call.
+        """
+        pass
+
+    def test___sys_path_insert(self):
+        py4j = glob.glob(os.path.join(self.spark_home, 'python/lib', 'py4j-*.zip'))[0]
```
hkvision: is self.spark_home defined?

yiheng:
@megaSpoon  any update of Kais's comments?

megaSpoon:
Jenkins PR test #138 passed

PR #2605, opened 2018-07-30
Distributed asynchronizer based on block manager

wzhongyuan:

Design and implement a distributed synchronizer using block manager
1) support priority
2) aggregation on partition (return partition specific aggregation result), applicable for gradient scenario 
3) all in asynchronous processing

## How was this patch tested?

unit test

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
@@ -1,0 +1,210 @@
+/*
+ * Copyright 2016 The BigDL Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.intel.analytics.bigdl.utils
+
+import java.nio.ByteBuffer
+import java.util.concurrent._
+
+import com.intel.analytics.bigdl.parameters.AllReduceParameter.{computePoolSize, logger, syncPool, syncPoolSize}
+import com.intel.analytics.bigdl.parameters.{CompressedTensor, FP16CompressedTensor, SerializerInstance}
+import com.intel.analytics.bigdl.tensor.Tensor
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+import org.apache.commons.lang.exception.ExceptionUtils
+import org.apache.spark.sparkExtension.SparkExtension
+import org.apache.spark.storage.{BlockId, BlockManagerWrapper, StorageLevel}
+
+import scala.collection.mutable
+import scala.reflect.{ClassTag, classTag}
+import scala.collection.JavaConverters._
+
+trait DistriParameterSynchronizer[T] {
+
+  /**
+   * Init synchronization context for new layer
+   * @param name  identifier for parameter
+   * @param globalSize toal size of parameter
+   */
+  def init(name: String, globalSize: Int): Unit
+
+  /**
+   * put parameter to global
+   * @param name identifier for parameter
+   * @param parameter  paraemter to put
+   */
+  def put(name: String, parameter: Tensor[T]): Unit
+
+  /**
+   * get parameter for specific identifier
+   * @param name  identifier for parameter
+   * @return  parameter
+   */
+  def get(name: String): Tensor[T]
+}
+
+class BlockManagerParameterSynchronizer[T: ClassTag](partitionID: Int,
+                                                     totalPartition: Int)
+                                                    (implicit ev: TensorNumeric[T])
+  extends DistriParameterSynchronizer[T] with Serializable {
+
+  private val syncResults: mutable.HashMap[String, Future[Tensor[T]]]
+    = new mutable.HashMap[String, Future[Tensor[T]]]()
+
+  private val workerPoolSize: Int = System.getProperty("bigdl.Parameter.syncPoolSize", "4").toInt
+
+  private val syncPoolSize: Int = Math.max(System.getProperty("bigdl.Parameter.computePoolSize",
+    (Runtime.getRuntime().availableProcessors() / 2).toString).toInt, 1)
+
+  private lazy val workerPool: ExecutorService =
+    Executors.newFixedThreadPool(workerPoolSize, new ThreadFactory {
+    override def newThread(r: Runnable): Thread = {
+      val t = Executors.defaultThreadFactory().newThread(r)
+      t.setDaemon(true)
+      t
+    }
+  })
+
+  private lazy val syncPool: ExecutorService = Executors.newFixedThreadPool(syncPoolSize,
+    new ThreadFactory {
+      override def newThread(r: Runnable): Thread = {
+        val t = Executors.defaultThreadFactory().newThread(r)
+        t.setDaemon(true)
+        t
+      }
+    })
+
+
+  private val syncMetaMap = new ConcurrentHashMap[String, SyncMeta[T]]
+
+  override def init(name: String, globalSize: Int): Unit = {
+    syncMetaMap.putIfAbsent(name, SyncMeta(name, 0, globalSize,
+      new ConcurrentHashMap[Int, CompressedTensor[T]]()))
+  }
+
+  override def put(name: String, parameter: Tensor[T]): Unit = {
+    val syncMeta = syncMetaMap.get(name)
+    val ayncTask = new AyncTask(syncMeta, parameter)
+    val future = workerPool.submit(ayncTask)
+    syncResults.put(name, future)
+    syncMeta.counter = syncMeta.counter + 1
+  }
+
+  override def get(name: String): Tensor[T] = {
+    require(syncResults.contains(name), "put must be done before get")
+    syncResults.get(name).get.get()
+  }
+
+  private class AyncTask(syncMeta: SyncMeta[T],
+    parameter: Tensor[T]) extends Callable[Tensor[T]] {
+    override def call(): Tensor[T] = {
+
+      // clear last status
+
+      syncMeta.stateOfWorld.clear
+      val _classTag = classTag[T]
+      val size = syncMeta.globalSize
+      val taskSize = size / totalPartition
+      val extraSize = size % totalPartition
+
+      // step 1 : put all local partitioned parameter to global
+
+      val putThreads = (0 until totalPartition).map { pid =>
+        new Callable[Int] {
+          override def call(): Int = {
+            try {
+              val offset = parameter.storageOffset() + pid * taskSize + math.min(pid, extraSize)
+              val length = taskSize + (if (pid < extraSize) 1 else 0)
+              val partitionParam = parameter.narrow(1, offset, length)
+              val parameterBlockId = getParameterBlockId(syncMeta.name, partitionID, pid)
+              println(s"${partitionID} put : ${syncMeta.name}paraBytes${partitionID}${pid}")
+              val fp16param = new FP16CompressedTensor[T](length)(_classTag)
+              fp16param.compress(0, parameter, offset - 1, length)
+              BlockManagerWrapper.putBytes(parameterBlockId,
+                fp16param.bytes(), StorageLevel.MEMORY_ONLY_SER)
+              pid
+            } catch {
+              case t: Throwable =>
+                logger.error("Error: " + ExceptionUtils.getStackTrace(t))
+                throw t
+            }
+          }
+        }
+      }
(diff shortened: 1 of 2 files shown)


wzhongyuan:
@yiheng @qiuxin2012  @i8run   Please help check and review, your comments and suggestions are highly appreciated.

this PR generally implements a distributed synchronizer based on block manager, the challenge here is to support priority queue, as well as the status checking (sync over storage)

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
```diff
@@ -0,0 +1,366 @@
...
+import scala.collection.mutable
+import scala.reflect.{ClassTag, classTag}
+import scala.collection.JavaConverters._
+
+trait DistriParameterSynchronizer[T] {
+
+  /**
+   * Init synchronization context for new parameter
+   * @param name  identifier for parameter
+   * @param globalSize toal size of parameter
```
yiheng: total

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
```diff
@@ -0,0 +1,366 @@
...
+   * @param priority priority for this parameter
```
yiheng: larger is higher?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
```diff
@@ -0,0 +1,366 @@
...
+   */
+  def init(name: String, globalSize: Int, priority: Int = 1): Unit
```
yiheng: what does the globalSize mean? Say the tensor size is 1M, there're 10 nodes. GlobalSize = 10M or 1M?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
```diff
@@ -0,0 +1,366 @@
...
+    ".fetchCompletionPoolSize", "1").toInt
+
+  private val asyncTaskWaitingQueue : PriorityBlockingQueue[AsyncFutureTask] =
+    new PriorityBlockingQueue[AsyncFutureTask](taskSize)
+
+  private val blockFetchRequestQueue: PriorityBlockingQueue[BlockFetchRequest] =
+    new PriorityBlockingQueue[BlockFetchRequest](taskSize)
+
+  // thread pool to remove expired blocks
+  private lazy val clearPool: ExecutorService =
```
yiheng: transient?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
```diff
@@ -0,0 +1,366 @@
...
+
+  import com.intel.analytics.bigdl.utils.BlockManagerParameterSynchronizer.logger
+
+  private val syncResults: mutable.HashMap[String, FutureTask[Tensor[T]]]
+    = new mutable.HashMap[String, FutureTask[Tensor[T]]]()
+
+  private val taskSize: Int = System.getProperty("bigdl.ParameterSynchronier." +
+    "asyncTaskSize", "100").toInt
+
+  private val clearPoolSize: Int = System.getProperty("bigdl.ParameterSynchronier." +
```
yiheng: why multiple pools?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
```diff
@@ -0,0 +1,366 @@
...
+        val t = Executors.defaultThreadFactory().newThread(r)
+        t.setDaemon(true)
+        t
+      }
+    })
+
+  (0 until syncPoolSize).foreach(th => {
+    fetchPool.submit(new Runnable {
+      override def run(): Unit = {
+        while (true) {
```
yiheng: can we shutdown the service?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
```diff
@@ -0,0 +1,366 @@
...
+   */
+  def get(name: String): Tensor[T]
+
+  /**
+   * clear the synchronizer
+   */
+  def clear(): Unit
+}
+
+class BlockManagerParameterSynchronizer[T: ClassTag](partitionID: Int,
```
yiheng: why Serializable ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
wzhongyuan: ok, fixed the typo

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
wzhongyuan: yes

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
wzhongyuan: the global size should be 1m in this case, each node handles 1k

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
wzhongyuan: three reasons for this :  
1) There are some long running threads like clear threads and request handling threads
2) Because we do internally and asynchronously, some threads are blocked to wait for other threads result, thus we should do in different pools to avoid blocking
3) Easy to tune the pool size to make it high efficient

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
wzhongyuan: Not required, I will remove this

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
wzhongyuan: We could, but the threads already running will be terminated until executor lifecycle ends, I will add some logic to make it more graceful

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
```diff
@@ -0,0 +1,366 @@
...
+                throw t
+            }
+          }
+        }
+      }
+      syncPool.invokeAll(syncThreads.asJava)
+
+      // step 4: aggregation
+
+      val length = taskSize + (if (partitionID < extraSize) 1 else 0)
```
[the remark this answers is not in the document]
wzhongyuan: we can make the syncpool smaller, the worst case should be 1 which is equivalant to single thread

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
```diff
@@ -0,0 +1,369 @@
...
+        val t = Executors.defaultThreadFactory().newThread(r)
+        t.setDaemon(true)
+        t
+      }
+    })
+
+  (0 until syncPoolSize).foreach(th => {
+    fetchPool.submit(new Runnable {
+      override def run(): Unit = {
+        while (true && !shutdown) {
```
yiheng: while(true && !shutdown) -> while(!shutdown)...

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
```diff
@@ -0,0 +1,369 @@
...
+          val fetchRequest = blockFetchRequestQueue.take
+          val syncMeta = fetchRequest.syncMeta
+          val pid = fetchRequest.futureTask.fetchOnCompletion.fromPartition
+          val parameterBlockId = getParameterBlockId(syncMeta.name,
+          syncMeta.counter, pid, partitionID)
+          val block = BlockManagerWrapper.getLocalOrRemoteBytes(parameterBlockId)
+          if (block == None) {
+            // promote the priporty in next fetch
+            fetchRequest.priority += 1
```
yiheng: why increase the priority if the data is not ready?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
```diff
@@ -0,0 +1,369 @@
...
+                throw t
+            }
+          }
+        }
+      }
+      syncPool.invokeAll(putThreads.asJava)
+
+      // step 3: get all remote paritioned parameter to local
+
+      val syncThreads = (0 until totalPartition).map { pid =>
```
yiheng: So we only get 1/n all reduced block?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
wzhongyuan: every block is supposed to be fetched in the right order, so the one that was not fetched should be prioritized as it has highest possibility in next fetch

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
wzhongyuan: refined with all paritions

wzhongyuan:
Ran a PR successfully  http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/124/

wzhongyuan:
Added small size parameter sync method (size < totalPartition), running another PR now

wzhongyuan:
PR validation passed http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/125/

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
```diff
@@ -0,0 +1,429 @@
...
+      val t = Executors.defaultThreadFactory().newThread(r)
+      t.setDaemon(true)
+      t
+    }
+  })
+
+  // long running thread to fetch the request
+  workerPool.submit(new Runnable {
+    override def run(): Unit = {
+      val asyncFutureTask = asyncTaskWaitingQueue.take
```
yiheng: long running?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
```diff
@@ -0,0 +1,429 @@
...
+      workerPool.submit(asyncFutureTask.task)
```
yiheng: why use recursion here?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
yiheng: long running?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
```diff
@@ -0,0 +1,429 @@
...
+        t.setDaemon(true)
+        t
+      }
+    })
+
+  (0 until syncPoolSize).foreach(th => {
+    fetchPool.submit(new Runnable {
+      override def run(): Unit = {
+        while (!shutdown) {
+          val fetchRequest = blockFetchRequestQueue.take
```
yiheng: even you set shutdown == true, if the thread is blocked at the `take`, the thread will not shutdown, right?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
wzhongyuan: it's a bug, thanks for pointing out, just fixed it

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/DistriParameterSynchronizer.scala
wzhongyuan: add a graceful mechanism to exit the thread after shutdown

wzhongyuan:
Ran another PR successfully after refinement https://github.com/intel-analytics/spark-dl/pull/2605
merging it now

PR #2608, opened 2018-08-10
Modify the thread pool to adopt mkldnn models

i8run:

This patch will modify the threadpools.

1. The `default` will support single thread including `LocalOptimizer` and `DistriOptimizer`. For supporting single thread version of `invokeWait2` method in `ThreadPool`, it will make the `threadpool` to current thread.
2. For dnn model, it will use the affinity to bind omp thread. And for performance issue, the `default` thread must use current main thread.
3. For `MTLabeldBGRImgToBatch` will use another new threads pool which is called `io`. So it will not be blocked when the `default` thread pool is single thread.
4. For `FileWriter`, it will not use `default`, otherwise the whole app will stuck at creating summary.

## How was this patch tested?

Manual tests and Jenkins.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
@@ -23,7 +23,7 @@ import java.util.concurrent.atomic.AtomicBoolean
 import org.apache.log4j.Logger
 import org.apache.spark._
 import com.intel.analytics.bigdl.mkl.MKL
-import com.intel.analytics.bigdl.mkl.hardware.Affinity
+import com.intel.analytics.bigdl.mkl.hardware.CpuInfo
 import org.apache.spark.utils.SparkUtils
 import py4j.GatewayServer
 
@@ -40,9 +40,8 @@ case object MklDnn extends EngineType
 
 object Engine {
 
-  // initialize some properties. Do not call it by manually.
-  // Because when we create some variables, they will set relevant properties, such as `_model`.
-  // If we do, any change of property for mkldnn will have no effect.
+  // Initialize some properties for mkldnn engine. We should call it at the beginning.
+  // Otherwise some properties will have no effect.
   setMklDnnEnvironments()
 
   @deprecated(
@@ -333,7 +332,6 @@ object Engine {
       (core * 50).toString).toInt
     if(_default == null || _default.getPoolSize != defaultPoolSize) {
       _default = new ThreadPool(defaultPoolSize)
-      _default.setMKLThread(MKL.getMklNumThreads)
     }
 
     if (_io == null) {
@@ -351,11 +349,7 @@ object Engine {
       _model.setMKLThread(MKL.getMklNumThreads)
     }
 
-    if (engineType == MklDnn) {
-      MKL.setNumThreads(MKL.getMklNumThreads)
-      com.intel.analytics.bigdl.mkl.MklDnn.setNumThreads(MKL.getMklNumThreads)
-      Affinity.setOmpAffinity()
-    }
+    ThreadPool.setThreadsOfBackend(MKL.getMklNumThreads)
   }
 
   /**
@@ -545,9 +539,9 @@ object Engine {
     }
   }
 
-  def setMklDnnEnvironments(): Unit = {
-    if (System.getProperty("bigdl.engineType") == "mkldnn") { // do not use engineType
-      val threadsNumber = Math.ceil(Runtime.getRuntime.availableProcessors() / 2).toInt
+  private def setMklDnnEnvironments(): Unit = {
+    if (System.getProperty("bigdl.engineType") == "mkldnn") { // do not use Engine.engineType
+      val threadsNumber = Math.ceil(CpuInfo.getPhysicalProcessorCount).toInt
 
       System.setProperty("bigdl.disable.mklBlockTime", "true")
       System.setProperty("bigdl.mklNumThreads", s"$threadsNumber")
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
@@ -38,6 +38,12 @@ case object MklDnn extends EngineType
 
 
 object Engine {
+
+  // initialize some properties. Do not call it by manually.
+  // Because when we create some variables, they will set relevant properties, such as `_model`.
+  // If we do, any change of property for mkldnn will have no effect.
+  setMklDnnEnvironments()
+
   @deprecated(
     "See https://bigdl-project.github.io/master/#APIGuide/Engine/",
     "0.1.0")
@@ -314,7 +320,7 @@ object Engine {
     val defaultPoolSize: Int = System.getProperty("bigdl.utils.Engine.defaultPoolSize",
       (core * 50).toString).toInt
     if(_default == null || _default.getPoolSize != defaultPoolSize) {
-      _default = new ThreadPool(defaultPoolSize)
+      _default = new ThreadPool(defaultPoolSize, withAffinity = true)
     }
 
     val modelPoolSize: Int = if (engineType == MklBlas) {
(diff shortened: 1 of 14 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/ThreadPool.scala
```diff
@@ -41,6 +43,7 @@ class ThreadPool(private var poolSize: Int) {
 
   private def spawnThreadPool(poolSize: Int): ExecutionContext = {
     if (poolSize == 1) {
+      threadPool = MoreExecutors.sameThreadExecutor()
```
yiheng: looks like the threadPool is not used

yiheng: Looks like the sameThreadExecutor is deprecated?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -515,4 +544,15 @@ object Engine {
       throw new IllegalArgumentException(s"Engine.init: Unsupported master format $master")
     }
   }
+
+  def setMklDnnEnvironments(): Unit = {
```
yiheng: is this a private method?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -515,4 +544,15 @@ object Engine {
...
+    if (System.getProperty("bigdl.engineType") == "mkldnn") { // do not use engineType
```
yiheng: what do you mean about "do not use engineType" in the comment?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -515,4 +544,15 @@ object Engine {
...
+      val threadsNumber = Math.ceil(Runtime.getRuntime.availableProcessors() / 2).toInt
```
yiheng: what if the machine core number is 1

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -327,6 +350,12 @@ object Engine {
       _model = new ThreadPool(modelPoolSize)
       _model.setMKLThread(MKL.getMklNumThreads)
     }
+
+    if (engineType == MklDnn) {
+      MKL.setNumThreads(MKL.getMklNumThreads)
+      com.intel.analytics.bigdl.mkl.MklDnn.setNumThreads(MKL.getMklNumThreads)
```
yiheng: why not import com.intel.analytics.bigdl.mkl.MklDnn?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -310,11 +320,24 @@ object Engine {
...
+    }
+    _io
+  }
+
   private def initThreadPool(core : Int) : Unit = {
     val defaultPoolSize: Int = System.getProperty("bigdl.utils.Engine.defaultPoolSize",
       (core * 50).toString).toInt
     if(_default == null || _default.getPoolSize != defaultPoolSize) {
       _default = new ThreadPool(defaultPoolSize)
+      _default.setMKLThread(MKL.getMklNumThreads)
```
yiheng: will this affect the MKLBLAS model execution?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -38,6 +39,12 @@ case object MklDnn extends EngineType
 
 
 object Engine {
+
+  // initialize some properties. Do not call it by manually.
+  // Because when we create some variables, they will set relevant properties, such as `_model`.
+  // If we do, any change of property for mkldnn will have no effect.
+  setMklDnnEnvironments()
```
yiheng: It looks a little confused. what about if(System.property("...") == mkldnn) setMklDnnEnvironments?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -327,6 +350,12 @@ object Engine {
       _model = new ThreadPool(modelPoolSize)
       _model.setMKLThread(MKL.getMklNumThreads)
     }
+
+    if (engineType == MklDnn) {
```
jason-dai: The code is similar to what `setMKLThread` does for each thread; maybe put it into a method and then invoke the method

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/ThreadPool.scala
i8run: `invokeAndWait2` will use `threadPool` and the `default` pool size maybe 1.

i8run: The `sameThreadExecutor` is deprecated for new version of guava and is replaced with `directExecutor`. But we use 11.0.2 which has only the `sameThreadExecutor`.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
i8run: It has no effects about the MKLBLAS model.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
i8run: Yes. I'll move them to a new methods in `ThreadPool` object.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
i8run: Yes.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -515,4 +538,15 @@ object Engine {
       throw new IllegalArgumentException(s"Engine.init: Unsupported master format $master")
     }
   }
+
+  private def setMklDnnEnvironments(): Unit = {
+    if (System.getProperty("bigdl.engineType") == "mkldnn") { // do not use Engine.engineType
```
yiheng: why not put the if statement outside the function?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -310,13 +319,25 @@ object Engine {
     _default
   }
 
+  private[bigdl] def io: ThreadPool = {
+    if (_io == null) {
+      throw new IllegalStateException(s"Engine.init: Thread engine is not " +
```
yiheng: why not init _io here?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -515,4 +538,15 @@ object Engine {
       throw new IllegalArgumentException(s"Engine.init: Unsupported master format $master")
     }
   }
+
+  private def setMklDnnEnvironments(): Unit = {
+    if (System.getProperty("bigdl.engineType") == "mkldnn") { // do not use Engine.engineType
+      val threadsNumber = Math.ceil(CpuInfo.getPhysicalProcessorCount).toInt
```
yiheng: not / 2?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/ThreadPool.scala
yiheng: I think the code structure is weird. And we should not use a deprecated method

PR #2609, opened 2018-08-10
feature: add shutdown for optimizer which will release the native resources

i8run:

This patch add an API `shutdown` for `Optimizer` to release native resources when the training end.
When should only implement them for dnn layers and optimizer. For container, it have been implemented in `Container`.

## How was this patch tested?
Manual and Jenkins.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
@@ -205,7 +205,7 @@ private[bigdl] object ModelInfo {
     implicit ev: TensorNumeric[T]): ModelInfo[T] = new ModelInfo[T](uuid, model)
 }
 
-object CachedModels {
+private[bigdl] object CachedModels {
   import java.util.concurrent.ConcurrentHashMap
 
   import scala.collection._
@@ -232,7 +232,6 @@ object CachedModels {
       for (key <- keys) {
         if (key != currentKey) {
           val models = cachedModels(key)
-          println(s"delete key = $key ${models.length}")
           for (model <- models) {
             model.release()
           }
@@ -247,7 +246,6 @@ object CachedModels {
       for (k <- keys) {
         if (k == key) {
           val models = cachedModels(key)
-          println(s"delete key = $key ${models.length}")
           for (model <- models) {
             model.release()
           }
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
@@ -240,4 +240,19 @@ object CachedModels {
         }
       }
     }
+
+  def deleteKey[T: ClassTag](key: String)(implicit ev: TensorNumeric[T]): Unit =
+    CachedModels.synchronized {
+      val keys = cachedModels.keys
+      for (k <- keys) {
+        if (k == key) {
+          val models = cachedModels(key)
+          println(s"delete key = $key ${models.length}")
+          for (model <- models) {
+            model.release()
+          }
+          cachedModels.remove(key)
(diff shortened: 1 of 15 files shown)


yiheng:
Please add some unit tests

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -1062,6 +1063,8 @@ class DistriOptimizer[T: ClassTag] (
 
     // unpersist the model because the next time optimize is called, new `models` will be
     // created
+    shutdown()
+    CachedModels.deleteKey(modelBroadcast.uuid)
```
yiheng: should this be a part of shutdonw()?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -963,9 +960,11 @@ class DistriOptimizer[T: ClassTag] (
 
     prepareInput()
 
-    models = DistriOptimizer.initThreadModels(model, distDataset, criterion, state,
+    val modelsAndBroadcast = DistriOptimizer.initThreadModels(model, distDataset, criterion, state,
```
yiheng: (models, modelBroadcast) = ... ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnBase.scala
```diff
@@ -250,6 +250,26 @@ trait MklDnnLayer extends AbstractModule[Activity, Activity, Float] with MklDnnM
   def parametersWithShape(): (Array[MemoryData], Array[MemoryData]) = {
     (null, null)
   }
+
+  override def release(): Unit = {
+    val tensors: ArrayBuffer[DnnTensor[Float]] = ArrayBuffer.empty
+    List(output, gradInput).filter(_ != null).foreach { t =>
+      if (t.isTensor && t.toTensor[Float].getTensorType == MklDnnType) {
```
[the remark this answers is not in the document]
i8run: Yes, we have the `MklDnnType`. We can't use `isInstanceOf` here, compiler will say it's type erasure.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
i8run: Scala doesn't support this style. We only can init two variables like this, such as `val (a, b) = (1, 2)`. But can't assign to two `var`s, such as `var c:Int = 0; var d: Int = 0; (c, d) = (1, 2)`.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
i8run: Yes. Moving to `shutdown` makes it simple.

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
```diff
@@ -240,4 +240,19 @@ object CachedModels {
         }
       }
     }
+
+  def deleteKey[T: ClassTag](key: String)(implicit ev: TensorNumeric[T]): Unit =
```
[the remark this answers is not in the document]
i8run: I'll move this whole `object` to `private[bigdl]`.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnBase.scala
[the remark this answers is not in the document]
yiheng: DnnTensor[_] is type erasure?

PR #2612, opened 2018-08-16
New parallel optimizer for multinode training

wzhongyuan:

1) new parallel optimizer
2) layerwise training
3) communication overlapping with block manager based synchronizer
## How was this patch tested?

unit test, integration test with lenet

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/ParallelOptimizer.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/ParallelOptimizer.scala
@@ -275,7 +275,7 @@ object ParallelOptimizer {
           driverState[Int]("neval"), wallClockTime)
         logger.info(s"${_header} Trained ${localRecordsNum} records in ${(end - start) / 1e9} " +
           s"seconds. Throughput is ${driverState("Throughput")} records/second. Loss is ${
-            driverState("Loss")}. ${getHyperParameterLog(optimMethods)}")
+            driverState("Loss")}.")
         logger.debug("\n" + metrics.summary())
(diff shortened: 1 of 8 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/ParallelOptimizer.scala
```diff
@@ -0,0 +1,1009 @@
...
+         */
+          val end = System.nanoTime()
+          wallClockTime += end - start
+          Iterator.single(finishedThreadSize, lossSum, recordsNum)
+        }
+      }.reduce((a, b) => (a._1 + b._1, a._2 + b._2, a._3 + b._3))
+
+      dropModelNumBatch += (driverSubModelNum - numFinishedModelUpdates)
+
+      if (dropPercentage == 0.0 ||
```
yiheng: let's don't support the backup worker feature in the first version?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/ParallelOptimizer.scala
```diff
@@ -0,0 +1,1009 @@
...
+          epochStart = System.nanoTime()
+          logger.info(s"${_header} Epoch finished. Wall clock time is ${wallClockTime / 1e6} ms")
+
+          driverState("epoch") = driverState[Int]("epoch") + 1
+          dataset.shuffle()
+          dataRDD = dataset.data(train = true)
+          recordsProcessedThisEpoch = 0
+        }
+
+        optimMethods.map { case (moduleName, optimMethod) =>
```
yiheng: do we still need this?

PR #2613, opened 2018-08-17
fix: should load dnn relevant library if enabled dnn engine.

i8run:

It should load the dnn relevant library when enable dnn engine. Because currently we has not supported windows so should not do if disabled dnn engine.

## How was this patch tested?
Jenkins and manual on windows.


yiheng:
LGTM

PR #2615, opened 2018-08-17
fix: set the encoding of input and output files is UTF-8

i8run:

On windows the scala style check will failed. Need the input encoding to UTF-8.

## How was this patch tested?

Manual test on Windows 10.


yiheng:
LGTM

Issue #2558, opened 2018-06-15
Predict image in spark local mode get IllegalArgumentException


jenniew:
We met errors when using predictImage API to predict inception v1 w/ ImageNet val. But it reported java.lang.IllegalArgumentException for 5k images and java.lang.ArrayIndexOutOfBoundsException for 4999 images. Predicting 1000 images can pass.

Code:
     val conf = Engine.createSparkConf()
     conf.setAppName("Predict with trained model")
      val sc = new SparkContext(conf)
      Engine.init
      val model = Module.loadModule[Float](params.model)
      val images = ImageFrame.read(params.imageFolder, sc, params.nPartition)
      val partitionNum = images.toDistributed().rdd.partitions.length
      val partitions = images.toDistributed().rdd.partitions

      logger.info(s"partition number is: ${partitionNum}")
      images.toDistributed().rdd.mapPartitions(partition => {
        println(s"partition size: ${partition.size}")
        partition
      }).collect()

      val inceptionPreprocessing = Resize(256, 256) -> CenterCrop(224, 224) ->
        ChannelNormalize(123, 117, 104) ->
        MatToTensor[Float]() ->
        ImageFrameToSample[Float]()

      val data = inceptionPreprocessing(images)
      val output = model.predictImage(data)
      val result = output.toDistributed().rdd.collect

Execution script:

#!/bin/sh
spark-submit --master local[28]  --conf spark.driver.maxResultSize=20g --driver-memory 20g --class com.intel.analytics.bigdl.example.imageclassification.ImageClassifier ./dist/lib/bigdl-0.5.0-jar-with-dependencies.jar -f /root/val_5k --model /root/bigdl_inception-v1_imagenet_0.4.0.model  -p 4 

Exception:
java.lang.IllegalArgumentException: requirement failed: Input must have 4 dimensions, corresponding to (batch, channels, height, width)input dimension 3
	at scala.Predef$.require(Predef.scala:224)
	at com.intel.analytics.bigdl.nn.SpatialCrossMapLRN.updateOutput(SpatialCrossMapLRN.scala:97)
	at com.intel.analytics.bigdl.nn.SpatialCrossMapLRN.updateOutput(SpatialCrossMapLRN.scala:43)
	at com.intel.analytics.bigdl.nn.abstractnn.AbstractModule.forward(AbstractModule.scala:243)
	at com.intel.analytics.bigdl.nn.StaticGraph.updateOutput(StaticGraph.scala:59)
[...]

PR #2616, opened 2018-08-20
[Bug Fix] Handle grey image correctly if model require a 3 channel tensor input

yiheng:
The OpenCV based transformer will convert the grey image into 1 channel tensor, which causes some problem when the model requires a 3 channel tensor. The PR provides a conversion for such situation.

## How was this patch tested?
New unit test, existing ut and manual test

fixed https://github.com/intel-analytics/BigDL/issues/2558


yiheng:
jenkins pass

spark/dl/src/main/scala/com/intel/analytics/bigdl/transform/vision/image/ImageFeature.scala
```diff
@@ -296,6 +298,27 @@ class ImageFeature extends Serializable {
...
+      while(c < 3) {
+        var i = 0
+        while(i < frameLength) {
+          storageFloat(i + c * frameLength + offset) = data(i)
+          i += 1
+        }
+        c += 1
+      }
+    } else {
+      throw new IllegalArgumentException("Not support type")
```
wzhongyuan: supported

wzhongyuan:
LGTM

Issue #2604, opened 2018-07-30
Use a factory for model broadcast (instead of hard coding)


jason-dai:
Instead of hard coding how to broadcast model (`ModelBroadcast`), use a factory that can be customized by the users (e.g., see broadcast in Spark https://github.com/apache/spark/blob/branch-1.6/core/src/main/scala/org/apache/spark/broadcast/BroadcastManager.scala)

yiheng:
We'll look into it.

PR #2618, opened 2018-08-22
[Feature] Allow user to customized how model is broadcast in distributed training

yiheng:
In this PR, we define a ModelBroadcast factory. It will create a ModelBroadcast class to control how model is broadcasted in training. The user can extend the ModelBroadcast factory to change it. 

## How was this patch tested?
unit tests

fixed https://github.com/intel-analytics/BigDL/issues/2604

### DIFF
--- a/docs/docs/ScalaUserGuide/configuration.md
+++ b/docs/docs/ScalaUserGuide/configuration.md
@@ -34,7 +34,7 @@ java -cp xxx.jar -DFOO=BAR your.main.class.name
 **Mode**
 
 - `bigdl.localMode`: Whether BigDL is running as a local Java/Scala program. Default is false.
-- `bigdl.engineType`: Default is **mkldnn**. When you run model contains mkl dnn layers, you should set it to **mkldnn** to get better performance.
+- `bigdl.engineType`: Default is **mklblas**. When you run model contains mkl dnn layers, you should set it to **mkldnn** to get better performance.
 
 **Multi-threading**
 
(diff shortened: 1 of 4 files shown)


docs/docs/ScalaUserGuide/configuration.md
```diff
@@ -34,6 +34,7 @@ java -cp xxx.jar -DFOO=BAR your.main.class.name
 **Mode**
 
 - `bigdl.localMode`: Whether BigDL is running as a local Java/Scala program. Default is false.
+- `bigdl.engineType`: Default is **mkldnn**. When you run model contains mkl dnn layers, you should set it to **mkldnn** to get better performance.
```
jason-dai: default is mklblas or mkldnn?

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
```diff
@@ -31,6 +31,41 @@ import org.apache.spark.broadcast.Broadcast
...
+   * @param initGradient If create a tensor for gradient when fetch the model. Please note that
+   *                     the gradient is not needed in model inference
+   * @return model
+   */
+  def value(initGradient: Boolean = false): Module[T]
+
+  def uuid(): String = UUID.randomUUID().toString
+}
+
+private[bigdl] object ModelBroadcast {
```
jason-dai: why private[bigdl]? I think this is needed in user applications.

docs/docs/ScalaUserGuide/configuration.md
yiheng: Er.. a typo. will fix it

yiheng:
Jenkins pass

Issue #2617, opened 2018-08-22
Error 'assertion failed: total weight must larger than 0' - occasional failures when using NLLCriterion with some unlabeled data


griffinkardos:
I'm training on an dataset with some unlabeled data, using a parallel criterion, where one of the criteria is a reconstruction loss, and the other is a classification loss, which should be ignored for unlabeled data. When I call optimizer.optimize() with this criterion, it trains for some number of iterations (i.e, it doesn't fail on every iteration), before crashing with this error:

```
2018-08-21 23:56:02 ERROR ThreadPool$:115 - Error: java.lang.AssertionError: assertion failed: total weight must larger than 0
	at scala.Predef$.assert(Predef.scala:170)
	at com.intel.analytics.bigdl.nn.ClassNLLCriterion.updateGradInput(ClassNLLCriterion.scala:169)
	at com.intel.analytics.bigdl.nn.CrossEntropyCriterion.updateGradInput(CrossEntropyCriterion.scala:48)
	at com.intel.analytics.bigdl.nn.CrossEntropyCriterion.updateGradInput(CrossEntropyCriterion.scala:31)
	at com.intel.analytics.bigdl.nn.ParallelCriterion.updateGradInput(ParallelCriterion.scala:74)
	at com.intel.analytics.bigdl.nn.ParallelCriterion.updateGradInput(ParallelCriterion.scala:35)
	at com.intel.analytics.bigdl.nn.abstractnn.AbstractCriterion.backward(AbstractCriterion.scala:84)
	at com.intel.analytics.bigdl.optim.DistriOptimizer$$anonfun$4$$anonfun$6$$anonfun$apply$2.apply$mcI$sp(DistriOptimizer.scala:258)
	at com.intel.analytics.bigdl.optim.DistriOptimizer$$anonfun$4$$anonfun$6$$anonfun$apply$2.apply(DistriOptimizer.scala:249)
	at com.intel.analytics.bigdl.optim.DistriOptimizer$$anonfun$4$$anonfun$6$$anonfun$apply$2.apply(DistriOptimizer.scala:249)
	at com.intel.analytics.bigdl.utils.ThreadPool$$anonfun$1$$anon$4.call(ThreadPool.scala:112)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
[...]

PR #2620, opened 2018-08-22
fix: Don't crash NLL classifier if all targets are unlabeled

griffinkardos:
If all the targets are unlabeled (or all target classes have a weight of zero), changes total_weight from 0 to 1, so it doesn't crash in backwards() when it asserts the total weight is > 0. This doesn't change the output of forward or backwards to anything unexpected.

fixed https://github.com/intel-analytics/BigDL/issues/2617

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ClassNLLCriterion.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ClassNLLCriterion.scala
@@ -153,6 +153,9 @@ class ClassNLLCriterion[@specialized(Float, Double) T: ClassTag]
         total_weight = ev.plus(total_weight, w)
         i += 1
       }
+      if (total_weight == 0) {
+        total_weight = ev.fromType[Int](1)
+      }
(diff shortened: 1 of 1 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ClassNLLCriterion.scala
```diff
@@ -153,6 +153,9 @@ class ClassNLLCriterion[@specialized(Float, Double) T: ClassTag]
         total_weight = ev.plus(total_weight, w)
         i += 1
       }
+      if(total_weight == 0){
```
yiheng: there's a style check failure.

```
StyleCheck/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ClassNLLCriterion.scala message=
      Single Space between ')' and `{`.
     line=156 column=26
```

yiheng:
Thanks for the contribution. Only a small issue. The Jenkins job failed at style check. Please check my comments.

griffinkardos:
Okay thanks. I amended that

PR #2622, opened 2018-08-27
fix: the inference performance regression of mkldnn

i8run:

We should copy weights when `updateOutput` at training. The weights are loaded before and will not be changed when do inference.

## How was this patch tested?

Manual tests.


yiheng:
LGTM

PR #2624, opened 2018-08-28
feat: training ResNet50 w/ dnn backend on Spark.

i8run:

This patch supports training resnet-50 model w/ dnn backend on Spark.

## How was this patch tested?
Manual tests on Spark cluster with 64 nodes, the loss and accuracy is right.


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -340,11 +340,10 @@ object Engine {
...
 
-    val modelPoolSize: Int = if (engineType == MklBlas) {
-      1
-    } else {
-      core
-    }
+    // for dnn model we should set the pool size to 1 also.
+    // otherwise, it will downgrade the performance and
+    // FIXME make the loss to NaN.
+    val modelPoolSize = 1
```
yiheng: why remove the if ... else ...

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DnnTensor.scala
```diff
@@ -95,6 +95,11 @@ class DnnTensor[T: ClassTag](
     Memory.Axpby(this.nElement(), a, x, b, y)
   }
 
+  def scale(from: DnnTensor[T], scal: Float): Unit = {
+    val length = this.nElement()
+    Memory.Scale(length, scal, from._storage.ptr.address, this._storage.ptr.address)
```
yiheng: Should we create a new Class to handle all math operations instead of adding them to Memory?

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DnnStorage.scala
```diff
@@ -104,17 +104,23 @@ private[tensor] class DnnStorage[T: ClassTag](size: Int) extends Storage[T] {
...
     in.defaultReadObject()
-    ptr = new Pointer(allocate(this.size))
-    val elements = in.readObject().asInstanceOf[Array[Float]]
-    Memory.CopyArray2Ptr(elements, 0, ptr.address, 0, size, DnnStorage.FLOAT_BYTES)
+    if (!_isReleased) {
+      ptr = new Pointer(allocate(this.size))
+      val elements = in.readObject().asInstanceOf[Array[Float]]
+      Memory.CopyArray2Ptr(elements, 0, ptr.address, 0, size, DnnStorage.FLOAT_BYTES)
+    } else {
+      ptr = new Pointer(allocate(this.size))
```
yiheng: if it's released, should we use a null for ptr?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/ValidationMethod.scala
```diff
@@ -167,16 +167,21 @@ class TreeNNAccuracy[T: ClassTag]()(
...
   implicit ev: TensorNumeric[T])
   extends ValidationMethod[T] {
   override def apply(output: Activity, target: Activity):
   ValidationResult = {
     var correct = 0
     var count = 0
 
-    val _output = output.asInstanceOf[Tensor[T]]
     val _target = target.asInstanceOf[Tensor[T]]
+    val _output = if (output.toTensor[T].size().head != _target.size().head) {
```
yiheng: why should we change this?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/ValidationMethod.scala
```diff
@@ -215,11 +220,18 @@ class Top1Accuracy[T](
...
-class Top5Accuracy[T] extends ValidationMethod[T] {
+class Top5Accuracy[T: ClassTag](
+  implicit ev: TensorNumeric[T]) extends ValidationMethod[T] {
   override def apply(output: Activity, target: Activity):
   AccuracyResult = {
-    val _output = output.asInstanceOf[Tensor[T]]
-    val _target = target.asInstanceOf[Tensor[T]].squeezeNewTensor()
+    var _target = target.asInstanceOf[Tensor[T]].squeezeNewTensor()
+
+    val _output = if (output.toTensor[T].size().head != _target.size().head) {
```
yiheng: why should we change this?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
i8run: The thread pool size of all the two cases is 1. We need not the `if else`. If the thread pool size of `_model` more than 1, it will downgrade the performance of Resnet50.

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DnnTensor.scala
i8run: Yes. We can do it. But we should refactor `BigDL-core` first.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/ValidationMethod.scala
i8run: For dnn model, the model's output is consistent. For example, we have `32x3x224x224` input and `32x1000` as output. When the input's batch size is less than 32, the output will output `32x1000` also, but the target will be not. So we should narrow the useful elements.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/ValidationMethod.scala
i8run: The same as above.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -148,6 +148,10 @@ object DistriOptimizer {
     logger.info("Count dataset")
     val countBefore = System.nanoTime()
     val numSamples = dataset.data(train = false).map(_.size()).reduce(_ + _)
+    // FIXME dnn model must cache val dataset first, otherwise there will be a segment fault.
+    if (Engine.getEngineType() == MklDnn) {
+      validationDataSet.get.toDistributed().data(train = false).map(_.size()).reduce(_ + _)
```
jason-dai: what if validationDataSet is None?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -907,6 +911,8 @@ class DistriOptimizer[T: ClassTag] (
     if (!dataset.asInstanceOf[DistributedDataSet[MiniBatch[T]]].isCached) {
       logger.info("caching training rdd ...")
       dataset.asInstanceOf[DistributedDataSet[MiniBatch[T]]].cache()
+      // FIXME dnn model must cache val dataset first, otherwise there will be a segment fault.
+      validationDataSet.get.toDistributed().cache()
```
jason-dai: what if validationDataSet is None?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
jason-dai: And why do we need this if you already cached validationDataSet in prepareInput?

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DnnStorage.scala
i8run: Yes. You are right. I'll check the `ModelBroadcast`. It should be the bug of last fix of quantized tensor memory leak.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
i8run: Yes. We should not include this.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
i8run: I'll check whether it's `None`.

PR #2627, opened 2018-09-03
feat: add dnn vgg model.

i8run:

Add vgg16 model w/ dnn backend.

## How was this patch tested?
Manual and jenkins.


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/ResNet50Perf.scala
```diff
@@ -36,7 +37,10 @@ object ResNet50Perf {
 
   val logger = Logger.getLogger(getClass)
 
-  val parser = new OptionParser[ResNet50PerfParams]("BigDL Local ResNet-50 Performance Test") {
+  val parser = new OptionParser[ResNet50PerfParams]("BigDL w/ Dnn Local Model Performance Test") {
```
yiheng: should we rename the file and classes to Perf instead of ResNet50Perf?

i8run: Yes.

yiheng:
LGTM

Issue #2592, opened 2018-07-13
Predictor does not support model whose output has table within a table


yangw1234:
which is common if the the model contains two graph each output a table

yiheng:
We'll look into it.

PR #2629, opened 2018-09-11
Fix issue 2592

megaSpoon:

fix issue 2592 for the problem that previous splitBatch() function in the Predictor is not correctly implemented. And in the end, the Predictor should support output type with table inside a table.

## How was this patch tested?
put new unit tests in the PredictorSpec class

fixed https://github.com/intel-analytics/BigDL/issues/2592


megaSpoon:
passed local unit test in PredictorSpec. Ready for Jenkins.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Predictor.scala
```diff
@@ -68,57 +68,51 @@ object Predictor {
...
     val out = if (output.isTensor) {
-      val result = if (shareBuffer) output.toTensor[T] else output.toTensor[T].clone()
-      if (result.dim() == 1) {
-        require(batchSize == 1,
-          s"If result dim == 1, the batchSize is required to be 1, while actual is $batchSize")
-        Array(result)
-      } else {
-        result.split(1)
-      }
+      splitTensor(output.toTensor, shareBuffer, batchSize)
```
yiheng: output may be a table. Why output.toTensor?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Predictor.scala
```diff
@@ -68,57 +68,51 @@ object Predictor {
...
+    (implicit ev: TensorNumeric[T]): Array[Activity] = {
+    val result = if (shareBuffer) output else output.clone()
+    val size = result.size(1)
+    require(batchSize == size,
+      s"The batchSize is required to be $size, while actual is $batchSize")
+    val out = result.split(1)
+    out.asInstanceOf[Array[Activity]]
+  }
+
+  private[optim] def splitTable[T: ClassTag](output: Activity, shareBuffer: Boolean, batchSize: Int)
```
yiheng: The name is a little confusing. As you also split the tensor inside this function.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Predictor.scala
```diff
@@ -68,57 +68,51 @@ object Predictor {
...
           val table = T()
           tables(i - 1) = table
           (1 to result.length()).foreach(key => {
-            val split = result[Tensor[T]](key)(i)
-            if (shareBuffer) {
-              table.insert(split)
-            } else {
-              table.insert(split.clone())
-            }
+            val split = splitTable(result(key), shareBuffer, batchSize)
```
yiheng: move this outside the while loop?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Predictor.scala
```diff
@@ -68,57 +68,51 @@ object Predictor {
...
+      val tables = new Array[Table](batchSize)
+
+      (1 to result.length()).foreach(key => {
+
         var i = 1
-        while (i <= batch) {
+        while (i <= batchSize) {
           val table = T()
           tables(i - 1) = table
           (1 to result.length()).foreach(key => {
```
yiheng: remove this line?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Predictor.scala
megaSpoon: Yes. As concluded,  the table represents the output structure and the split operation happened on tensor. When we find a tensor output, just use splitTensor to split it. I should remove splitTable here since there is no literally table split here and we just recursively probe into the deeper layers of the table until we find a tensor.

megaSpoon: Fix: remove spliTable and simply use splitBtach to do the recursive calls

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Predictor.scala
megaSpoon: Yes, I should remove this line. I was struggling to decide whether batchsize or result.length should be the outermost to be looped and forgot to remove the extra line. It turns out that both ways are fine.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Predictor.scala
megaSpoon: same issue with the last comment

megaSpoon:
Jekins test #183 passed

yiheng:
It looks good to me. @yangw1234 please take a look at this.

PR #2631, opened 2018-09-19
feature: vgg-16 with mkldnn backend

i8run:

This patch add vgg-16 on imagenet example.

## How was this patch tested?

Tested on 32-node bdw cluster.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/vgg/TrainImageNet.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/vgg/TrainImageNet.scala
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2016 The BigDL Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.intel.analytics.bigdl.models.vgg
+
+import com.intel.analytics.bigdl.nn
+import com.intel.analytics.bigdl.nn.{Module, SoftmaxWithCriterion}
+import com.intel.analytics.bigdl.optim.SGD.{Poly, SequentialSchedule, Warmup}
+import com.intel.analytics.bigdl.optim._
+import com.intel.analytics.bigdl.utils.{Engine, LoggerFilter, MklBlas, MklDnn}
+import com.intel.analytics.bigdl.visualization.TrainSummary
+import org.apache.log4j.{Level, Logger}
+import org.apache.spark.SparkContext
+
+object TrainImageNet {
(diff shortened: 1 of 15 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/InitializationMethod.scala
```diff
@@ -278,12 +278,23 @@ case class ConstInitMethod(value: Double) extends InitializationMethod {
  *  (http://jmlr.org/proceedings/papers/v9/glorot10a/glorot10a.pdf)
  */
 case object Xavier extends InitializationMethod {
+  private var version2: Boolean = false
+
+  def setVersion2(v: Boolean): this.type = {
+    version2 = v
```
jason-dai: when should one use version2?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/SGD.scala
```diff
@@ -306,7 +306,7 @@ object SGD {
     override def updateHyperParameter[T](optimMethod: SGD[T]): Unit = {
       val nevals = optimMethod.state.get[Int]("evalCounter").getOrElse(0)
       val lr = optimMethod.learningRate
-      val polyIter = nevals - excludeIterations
+      val polyIter = nevals // fix: should have no exclude iterations.
```
jason-dai: this is no longer consistent with line#295

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/InitializationMethod.scala
i8run: If the model is defined by dnn backend, we should use version2, such as the vgg16 in the example.

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/vgg/TrainImageNet.scala
```diff
@@ -0,0 +1,115 @@
...
+          nesterov = param.nesterov, learningRateSchedule = lrSchedules)
+      }
+
+      val logdir = "vgg16-imagenet"
+      val appName = s"${sc.applicationId}"
+      val trainSummary = TrainSummary(logdir, appName)
+      trainSummary.setSummaryTrigger("LearningRate", Trigger.severalIteration(1))
+      trainSummary.setSummaryTrigger("Parameters", Trigger.severalIteration(10))
+
+      val criterion = Engine.getEngineType() match {
```
wzhongyuan: why different engines need different criterion functions?  do they have different topologies?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/InitializationMethod.scala
wzhongyuan: version2 seems quite confusing, can we have a meaningful name?

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/vgg/TrainImageNet.scala
i8run: Yes. They have different topologies.

wzhongyuan:
LGTM

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/SGD.scala
qiuxin2012: @i8run  This change lead the learningRate of TrainInceptionV1 inconsistent, if warmupEpoch and maxLR is set.

PR #2632, opened 2018-09-19
perf: need not narrow the gradients and zero gradients for dnn backend

i8run:

In `DistriOptimizer`, we need not narrow the gradients and zero gradients for dnn backend.

## How was this patch tested?
Tested on VGG-16 training.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Linear.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Linear.scala
@@ -67,7 +67,8 @@ class Linear(
       bias.copy(initBias)
     }
 
-    zeroGradParameters()
+    gradWeight.zero()
+    gradBias.zero()
   }
 
   override private[mkldnn] def initFwdPrimitives(inputs: Array[MemoryData], phase: Phase) = {
@@ -279,8 +280,6 @@ class Linear(
   }
 
   override def zeroGradParameters(): Unit = {
-    gradWeight.zero()
-    gradBias.zero()
   }
 
   override def release(): Unit = {
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/SpatialBatchNormalization.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/SpatialBatchNormalization.scala
@@ -94,6 +94,8 @@ class SpatialBatchNormalization(
 
     runningMean.zero()
     runningVariance.zero()
+
+    gradWeightAndBias.zero()
   }
 
   private object Index extends Serializable {
@@ -303,9 +305,6 @@ class SpatialBatchNormalization(
   }
 
   override def zeroGradParameters(): Unit = {
-    if (affine) {
-      gradWeightAndBias.zero()
-    }
   }
 
   override def parameters(): (Array[Tensor[Float]], Array[Tensor[Float]]) = {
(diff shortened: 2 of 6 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -315,7 +317,9 @@ object DistriOptimizer {
             (0 until _subModelNumber).map { i =>
               () => {
                 cached.localModels(i).training()
-                cached.localModels(i).zeroGradParameters()
+                if (Engine.getEngineType() != MklDnn) {
+                  cached.localModels(i).zeroGradParameters()
```
yiheng: can we override dnn layer zeroGradParameter with an empty method?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/SpatialBatchNormalization.scala
```diff
@@ -94,6 +94,8 @@ class SpatialBatchNormalization(
 
     runningMean.zero()
     runningVariance.zero()
+
+    gradWeightAndBias.zero()
```
yiheng: do not check affine?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/SpatialConvolution.scala
```diff
@@ -368,8 +369,6 @@ class SpatialConvolution(
   }
 
   override def zeroGradParameters(): Unit = {
```
yiheng: maybe we can just remove the method. I remember the default impl is do nothing, right?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Linear.scala
```diff
@@ -279,8 +280,6 @@ class Linear(
   }
 
   override def zeroGradParameters(): Unit = {
```
yiheng: maybe we can just remove the method. I remember the default impl is do nothing, right?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/SpatialBatchNormalization.scala
```diff
@@ -303,9 +305,6 @@ class SpatialBatchNormalization(
   }
 
   override def zeroGradParameters(): Unit = {
```
yiheng: maybe we can just remove the method. I remember the default impl is do nothing, right?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/SpatialConvolution.scala
i8run: No. The default implementation will get parameters and zero gradients.

```scala
  def zeroGradParameters(): Unit = {
    if (parameters() != null) {
      parameters()._1.zip(parameters()._2)foreach{ case (weight, grad) =>
        grad.resizeAs(weight).zero()
      }
    }
  }
```

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/SpatialBatchNormalization.scala
i8run: This BatchNorm only support the true of affine.

Issue #2606, opened 2018-08-08
not supported tensorflow estimator, supported tensorflow version is too low?


d0evi1:
My application is train NN model on tensorflow(GPU) platform with middle-level dataset, use this model  with spark(cpu) to inference large predictSet. BigDL is just what i wanted.

After I test BigDL-0.6 several days, i was frustrated。。。There are some issues that great affects tensorflow's usecase:

1.not supported tensorflow estimator（can export pb/bin, but can't load because of some unsupported op types）
2.supported tensorflow version is too low. I've tested the example from: https://github.com/intel-analytics/BigDL/tree/master/spark/dl/src/main/scala/com/intel/analytics/bigdl/example/tensorflow. 
this example is only supported with tensorflow version=1.2, when the version>=1.4,  the forward() function fails.

jason-dai:
@d0evi1 Maybe you can take a look at https://github.com/intel-analytics/analytics-zoo/blob/master/apps/tfnet/image_classification_inference.ipynb?

yiheng:
@d0evi1 Thanks for the feedback. We'll look into the issues on priority.

Issue #2511, opened 2018-04-28
Two tensorflow examples doesn't work with latest tensorflow


yiheng:
These two tensorflow examples doesn't work with latest tensorflow

https://github.com/intel-analytics/BigDL/tree/branch-0.5/spark/dl/src/main/scala/com/intel/analytics/bigdl/example/tensorflow/transferlearning

https://github.com/intel-analytics/BigDL/tree/branch-0.5/spark/dl/src/main/scala/com/intel/analytics/bigdl/example/tensorflow/loadandsave

It works fine with tensorflow==1.2.0

Need to 
1. Update the document
2. Fix the incompatiable issue

PR #2637, opened 2018-09-27
change Reshape to InferReShape in reshapeLoadTF

megaSpoon:
1. change Reshape to InferReShape in reshapeLoadTF to allow -1 in shape size
2. modified the docs for the loadandsave example in the tensorflow support
3. modified the unit test in the reshapeLoadTFSpec to test the case where the shape size contains a '-1'

## How was this patch tested?
1. Use modified old unit tests to run Jenkins build. No new unit tests need to be added.
2. Run the loadandsave example successfully to support both TF1.2 and TF1.11 which is the latest version during the test

fixed loadandsave in https://github.com/intel-analytics/BigDL/issues/2511 and in https://github.com/intel-analytics/BigDL/issues/2606

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/loaders/Reshape.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/loaders/Reshape.scala
@@ -18,7 +18,8 @@ package com.intel.analytics.bigdl.utils.tf.loaders
 import java.nio.ByteOrder
 
 import com.intel.analytics.bigdl.Module
-import com.intel.analytics.bigdl.nn.{InferReshape => ReshapeOps}
+import com.intel.analytics.bigdl.nn.{Reshape => ReshapeOps}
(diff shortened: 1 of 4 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/tf/loaders/Reshape.scala
```diff
@@ -52,7 +50,8 @@ class ReshapeLoadTF[T: ClassTag]()(implicit ev: TensorNumeric[T]) extends Adapte
       k += 1
       i += 1
     }
-    ReshapeOps[T](size = arraySize, Some(batchMode))
+    if (infer) ReshapeOps[T](size = arraySize, Some(batchMode))
+    else InferReshape[T](size = arraySize, batchMode)
```
yiheng: switch the if else

yiheng:
Please also add unit test

Issue #2640, opened 2018-09-29
Calling Optimizer.optimize() multiple times will reset some state in OptimMethod


yangw1234:
Every time `Optimizer.optimize()` is called, `OptimMethod.clearHistory()` will be called. ([this line](https://github.com/intel-analytics/BigDL/blob/master/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala#L921))
This operation will delete some state (such as the 1st and 2nd moment in adam) in optimMethod.
It is wrong if user wants to resume training from the previous state.

yangw1234:
maybe move this to Optimizer.clearState()? and also clear states like "epoch" "neval"?

yiheng:
@qiuxin2012 Please take a look at this.

qiuxin2012:
Aha, I also find this issue while writing NCF.

qiuxin2012:
Another question, when we [save a optimMethod](https://github.com/intel-analytics/BigDL/blob/d8ad65511d0030744e412b4cb070607a3a0bbe37/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/OptimMethod.scala#L93) , the history will be clear. And this clearing will make the serialized OptimMethod much smaller. So shall we clear the history or not while saving OptimMethod?

qiuxin2012:
Wall clock time shouldn't be clear either.

```
2018-09-21 15:43:20 INFO  NCFOptimizer2$:226 - [Epoch 1 99307520/99309568][Iteration 48490][Wall Clock 347.99537632s] loss is 0.13749460453566695, iteration time is 0.006456804s train time 0.00644507s. Thr
oughput is 317762.2585945537 record / second. 
2018-09-21 15:43:20 INFO  NCFOptimizer2$:226 - [Epoch 1 99309568/99309568][Iteration 48491][Wall Clock 348.002623711s] loss is 0.12566054140084557, iteration time is 0.007247391s train time 0.00723512s. Th
roughput is 283063.7225090945 record / second. 
2018-09-21 15:43:20 INFO  NCFOptimizer2$:293 - [Epoch 1 99309568/99309568][Iteration 48491][Wall Clock 348.002623711s] Validate model...
[...]

Issue #2636, opened 2018-09-26
Print validation throughput in DistriOptimizer


qiuxin2012:

zhangxiaoli73:
ok, we can just print average throught for validation.

PR #2648, opened 2018-10-11
Reserve all state in OptimMethod when calling Optimizer.optimize() multiple times

zhangxiaoli73:

1. Print validation throughput when doing distribute training
2. Support user to reserve optimMethod on each worker and reuse in next training task

## How was this patch tested?

unit tests

fixed https://github.com/intel-analytics/BigDL/issues/2640
fixed https://github.com/intel-analytics/BigDL/issues/2636

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
@@ -707,6 +707,14 @@ class DistriOptimizer[T: ClassTag] (
    * If you want to reserve optimMethod for each worker, and reuse those methods in
    * next training task, you can call it.
    */
+
+  /**
+   * If you want to reserve optimMethod for each worker and reuse those methods in
+   * next training task, please set reserve = true
+   * Otherwise, if just using optimMethod you set in optimizer, please set reserve = false
+   * @param reserve whether to reserve optim method for each worker
+   * @return
+   */
   def reserveOptim(reserve: Boolean): this.type = {
     reserveOptimMethod = reserve
     this
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
@@ -707,8 +707,8 @@ class DistriOptimizer[T: ClassTag] (
    * If you want to reserve optimMethod for each worker, and reuse those methods in
    * next training task, you can call it.
    */
-  def reserveOptim(): this.type = {
-    reserveOptimMethod = true
+  def reserveOptim(reserve: Boolean): this.type = {
+    reserveOptimMethod = reserve
     this
   }
 
@@ -923,11 +923,12 @@ class DistriOptimizer[T: ClassTag] (
 
     // reserve optimMethod internal state for each worker if need
     if (reserveOptimMethod) {
-      previousOptim = models.map(m => m.optimMethods)
+      previousOptim = models.map(m => m.optimMethods).cache()
+      previousOptim.count()
     } else {
-      previousOptim = null
-      models.unpersist()
+      if (previousOptim != null) previousOptim.unpersist()
     }
+    models.unpersist()
 
     model
   }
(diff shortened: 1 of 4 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -887,7 +918,14 @@ class DistriOptimizer[T: ClassTag] (
     // unpersist the model because the next time optimize is called, new `models` will be
     // created
     shutdown()
-    models.unpersist()
+
+    // reserve optimMethod internal state for each worker if need
+    if (reserveOptimMethod) {
+      previousOptim = models.map(m => m.optimMethods)
```
yangw1234: Should we make a deep copy here?

yangw1234: I think we should always unpersist models, as initThreadModels will always create new ones.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -786,7 +811,13 @@ class DistriOptimizer[T: ClassTag] (
     val modelsAndBroadcast = DistriOptimizer.initThreadModels(model, distDataset, criterion, state,
       nodeNumber, coresPerNode, checkSingleton, parameters, validationMethods,
       optimMethods, parameterProcessors)
-    models = modelsAndBroadcast._1
+
+    models = if (reserveOptimMethod && previousOptim != null) {
+      // replace optimMethods with previous ones
+      resetOptimMethods(modelsAndBroadcast._1, previousOptim)
```
yangw1234: If we go into this branch, the current `models` will depends on `previsousOptim` which depends on the `model` on last call of `optimize()`. This may create long dependencies if we call `optimize` many times.

yangw1234: Maybe we can cache the last previousOptim and release all previous ones.

zhangxiaoli73: Yeah, changed. Cache the last previousOptim and release all previous ones.

zhangxiaoli73:
@yangw1234 @qiuxin2012 Do you have any other comments about this pr?

yangw1234:
If user call setOptimMethods, it may not take effect.

yangw1234:
If all the optimMethods are supposed to have exactly the same state, how about just got one back and broadcasted it during the next call of optimize

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -73,7 +72,7 @@ object DistriOptimizer extends AbstractOptimizer {
     localStates: Array[Table],
     var moduleTimeList: Array[Long] = null,
     localMethods: Array[Option[Array[ValidationMethod[T]]]],
-    optimMethods: Map[String, OptimMethod[T]],
+    var optimMethods: Map[String, OptimMethod[T]],
```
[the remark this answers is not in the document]
zhangxiaoli73: as optimMethods may be changed when user wants to use previous optim methods.

zhangxiaoli73:
@yangw1234 for reserving one worker optimMethod and broadcasting it, I have discussed with qiuxin before. this solution may fail when broadcasting. 
your another concern is about reusing setOptimMethods, but it may no take effect. Yeah, how about add a parameter for reserveOptim to decide whether using previous one? or do you have any other better idea?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -889,6 +928,14 @@ class DistriOptimizer[T: ClassTag] (
     // unpersist the model because the next time optimize is called, new `models` will be
     // created
     shutdown()
+
+    // reserve optimMethod internal state for each worker if need
+    if (reserveOptimMethod) {
+      previousOptim = models.map(m => m.optimMethods).cache()
```
qiuxin2012: Do we need to unpersist before cache?

zhangxiaoli73:
jenkins pass： http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/331/

PR #2649, opened 2018-10-12
Fix transfer learning (#2645)

megaSpoon:


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialBatchNormalization.scala
```diff
@@ -317,17 +317,38 @@ object SpatialBatchNormalization {
...
-        var c = 0
-        while (c < nChannels) {
-          val invStd = 1 / Math.sqrt(varData(varOffset + c) + eps).toFloat
-          outputData(i + outputOffset + c) = (inputData(i + inputOffset + c) -
-            meanData(c + meanOffset)) * invStd * scaleData(scaleOffset + c) +
-            offsetData(offsetOffset + c)
-          c += 1
+      var isIdenticalScale = false
+      var identicalScale = 0.0f
+      if (scale.stride().length == 1 && scale.stride()(0) == 0 && scaleData.length == 1) {
```
yiheng: use the scaleData instead of scale

PR #2654, opened 2018-10-13
Feature/tidy test

starcolon:

Tidy up the redundant Spark life cycle setup and tear down of some tests.

## How was this patch tested?

The patch doesn't change any of the logics. So the existing unit tests should not break.

n/a

### DIFF
--- a/spark/dl/src/test/scala/com/intel/analytics/bigdl/dataset/DataSetSpec.scala
+++ b/spark/dl/src/test/scala/com/intel/analytics/bigdl/dataset/DataSetSpec.scala
@@ -22,7 +22,7 @@ import java.util.concurrent.{Callable, Executors}
 
 import com.intel.analytics.bigdl.dataset.image._
 import com.intel.analytics.bigdl.tensor.Tensor
-import com.intel.analytics.bigdl.utils.{Engine, RandomGenerator, TestUtils}
+import com.intel.analytics.bigdl.utils.{Engine, RandomGenerator, TestUtils, SparkContextLifeCycle}
 import org.apache.hadoop.io.Text
 import org.apache.spark.{SparkConf, SparkContext}
 import org.scalatest.{BeforeAndAfter, FlatSpec, Matchers}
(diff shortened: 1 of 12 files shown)


spark/dl/src/test/scala/com/intel/analytics/bigdl/utils/SparkTestLifeCycle.scala
```diff
@@ -0,0 +1,38 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.intel.analytics.bigdl.utils
+
+import org.apache.spark.{SparkConf, SparkContext}
+import org.scalatest.{BeforeAndAfter, FlatSpec, Matchers}
+
+trait SparkTestLifeCycle extends FlatSpec with BeforeAndAfter {
```
zhangxiaoli73: SparkContextLifeCycle

spark/dl/src/test/scala/com/intel/analytics/bigdl/utils/SparkTestLifeCycle.scala
```diff
@@ -0,0 +1,38 @@
...
+  var sc: SparkContext = null
+  def nodeNumber: Int
+  def coreNumber: Int
+  def appName: String
```
zhangxiaoli73: set a default  name

spark/dl/src/test/scala/com/intel/analytics/bigdl/utils/SparkTestLifeCycle.scala
```diff
@@ -0,0 +1,38 @@
...
+
+  before {
+    Engine.init(nodeNumber, coreNumber, true)
+    val conf = new SparkConf().setMaster("local[1]").setAppName(appName)
```
zhangxiaoli73: should be local[$coreNumber], not local[1]

zhangxiaoli73:
Hi ,thank you for your pr. Maybe you can do some changes:
1. Change "SparkTestLifeCycle" to "SparkContextLifeCycle" and add some comments about how to use it
2. Add default values for variables, as users may not assign values by themselves
3. Maybe your trait should just replace "BeforeAndAfter" in other unit tests.

starcolon:
Hi, I agree with your suggestions. Some of other compatible tests which also make use of [[sc]] now also use [[SparkContextLifeCycle]]. I leave some incompatible tests which [[SparkContextLifeCycle]] cannot apply as they are.

zhangxiaoli73:
ok. Please merge your nine commits into one.

starcolon:
okay, squashed :)

zhangxiaoli73:
PR has code style errors. Please use "mvn clean verify -DskipTests" to check code style.

starcolon:
fixed. just noticed `mvn test....` wont trigger the scala style checker. it should build find now

PR #2667, opened 2018-10-24
fix: uuid() will return a new uuid every call

i8run:

This patch fix two bugs.

1. ModelBroadcast will generate a new uuid when call `uuid()`. So `value()` will delete all models because their `uuid` are different with the new `uuid`.
2. If multiple partitions are mapped to the same worker, it will delete the previous cached model.

## How was this patch tested?

Manual and Jenkins.


spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
```diff
@@ -160,6 +162,10 @@ private[bigdl] class ModelBroadcastImp[T: ClassTag](applyProtoBuffer: Boolean =
     }
   }
 
+  private def getFullUId(uuid: String): String = {
```
wzhongyuan: full seems confusing,  probably getPartitionUID ?

wzhongyuan:
Add new test case for multipartion to same work?

wzhongyuan:
LGTM please merge it after jenkins passes

PR #2670, opened 2018-10-26
fix: uuid() will return a new uuid every call

i8run:

This patch fix a bug of `ModelBroadcast`. ModelBroadcast will generate a new uuid when call uuid(). So value() will delete all models because their uuid are different with the new uuid.

## How was this patch tested?

Jenkins and manual.


i8run:
Has passed Jenkins.

PR #2679, opened 2018-11-05
Add a splittable DataSet to help handle huge train data inputs.

bethunebtj:

BigDL read all input data into memory and coalesce them into a few partitions as a default manner, thus will cause OOM exception on huge train data. We will find that `rdd.repartition()` takes no effect, which is very confusing.

I think it would be better if we can provide a 'BIG DATA' mode, in which we can avoid OOM on huge data inputs by splitting them into small pieces.

1. Add a switch to object DataSet. 
If set to true, nothing will happen and everything works as usual.
If set to false, the input data will be splitted into batches of ( input partitions / executor numbers).
```
  var declineRepartitionedRdd = true
```
2. Add a new splittable `DataSet` implementation to DataSet.scala
When the switch is turned on, the `DataSet.rdd()` method will return a `SplittableTransformedDistributedDataSet` rather than a ` new DistributedDataSet` , with a new method called `getSplits()`. This method will return a much smaller RDD (and you can control its size by repartitioning the input data) instead of the original huge one.

3. DistriOptimizer will train all the input splits now. 

4. There is still a problem, that it will modify the original `Trigger` if set to `maxEpoch` or `maxIteration` ( e.g. the user will set the endwhen trigger to `Trigger.maxEpoch(10)` , and we split the whole data to 4 parts. In that way, the whole train procedure will run 40 epochs ) . 

## How was this patch tested?

1. Passes all tests if `declineRepartitionedRdd = true`
2. Test on WideAndDeep , NeuralCF model with 100GB input data, passed( will cause OOM before turning on the switch ) .

fixed https://github.com/intel-analytics/analytics-zoo-internal/issues/1225

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Trigger.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Trigger.scala
@@ -37,7 +37,7 @@ object TriggerType extends Enumeration {
 
   type TriggerType = Value
 
-  val EveryEpoch, SeveralIteration, MaxEpoch, MaxIteration, MaxScore, MinLoss, And, Or = Value
+  val EveryEpoch, SeveralIteration, MaxEpoch, MaxIteration, MaxScore, MinLoss, And, Or, Anonymous = Value
 }
 
 object Trigger {
--- a/spark/dl/src/test/scala/com/intel/analytics/bigdl/optim/OptimizerSpec.scala
+++ b/spark/dl/src/test/scala/com/intel/analytics/bigdl/optim/OptimizerSpec.scala
@@ -22,6 +22,7 @@ import com.intel.analytics.bigdl.dataset.{DistributedDataSet, LocalDataSet, Samp
 import com.intel.analytics.bigdl.nn.{ClassNLLCriterion, Linear, Sequential}
 import com.intel.analytics.bigdl._
 import com.intel.analytics.bigdl.example.loadmodel.AlexNet
+import com.intel.analytics.bigdl.optim.TriggerType.TriggerType
 import com.intel.analytics.bigdl.tensor.Tensor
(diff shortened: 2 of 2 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/dataset/DataSet.scala
```diff
@@ -357,13 +443,65 @@ object DataSet {
...
+            val rddBatches = mutable.ArrayBuffer[CachedDistriDataSet[T]]()
+            for (i <- 0 until splits) {
+              val splittedData = originalRdd
+                .mapPartitionsWithIndex((index, iter) => {
+                  val upperLimit = (i + 1) * nodeNum
+                  val downLimit = upperLimit - nodeNum
+                  if (index < upperLimit && index >= downLimit) iter
+                  else Iterator.empty
+                })
+              rddBatches.append(new CachedDistriDataSet(
```
zhichao-li: @songhappy I guess this solution is similar with what you did which split a "big" RDD into smaller pieces and then cache it in turn. One concern is we need to go through the whole dataset to filter the data each time when switching to a new "smaller" RDD. It might not be a issue as it only happen after several epochs and also depends on how much time it would take.

bethunebtj: @zhichao-li yes it will trigger rescanning the original data. When we have enough executor, it only takes a short time. But more executor means more model and longer all-reduce procedure, thus  should be deliberately considered. (btw, we can easily cache all train data（100GB） in memory to accelerate the procedure, as it's much smaller than `Sample` or `Tensor`)

jamesprinc3:
I've come across a similar issue which has led me here, what things need to be fixed (other than the merge conflicts of course) before this PR can get a positive review?

Issue #2672, opened 2018-10-29
endWhen and maxEpoch cannot be both set in DLEstimator


zhangxiaoli73:
In DLEstimator, "endWhen" and "maxEpoch" cannot be both set , otherwise there will be error

zhangxiaoli73:
@qiuxin2012 you can have a look, and give better work

qiuxin2012:
fixed in #2675

PR #2682, opened 2018-11-12
add Trigger and/or python API

qiuxin2012:

add and/or python API

## How was this patch tested?

unit tests

#2672

### DIFF
--- a/docs/docs/APIGuide/Triggers.md
+++ b/docs/docs/APIGuide/Triggers.md
@@ -107,7 +107,7 @@ s reached.
 ``` 
 A trigger contains other triggers and triggers when all of them trigger (logical AND). 
 
-For example, TriggerAnd(MinLoss(0.01), MaxEpoch(2)), means optimizer should stop when loss < 0.01 and epoch > 2.
+For example, TriggerAnd(MinLoss(0.01), MaxEpoch(2)), means training should stop when loss < 0.01 and epoch > 2.
 
 ---
 ## Or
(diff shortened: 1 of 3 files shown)


pyspark/bigdl/optim/optimizer.py
```diff
@@ -253,6 +253,45 @@ def __init__(self, min, bigdl_type="float"):
         """
         JavaValue.__init__(self, None, bigdl_type, min)
 
+class And(JavaValue):
```
zhangxiaoli73: Change another name, TriggerAnd?

pyspark/bigdl/optim/optimizer.py
```diff
@@ -253,6 +253,45 @@ def __init__(self, min, bigdl_type="float"):
...
+        """
+        Create a And trigger.
+
+
+        :param first: first Trigger
+        :param other: other Trigger
+        """
+        JavaValue.__init__(self, None, "float", first, list(other))
+
+class Or(JavaValue):
```
zhangxiaoli73: Same

docs/docs/APIGuide/Triggers.md
```diff
@@ -95,3 +95,26 @@ s reached.
...
+## Or
+**Scala:**
+```scala
+  val trigger = Trigger.and(Trigger.minLoss(0.01), Trigger.maxEpoch(2))
+```
+**Python:**
+```python
+  trigger = Or(MinLoss(0.01), MaxEpoch(2))
+``` 
+A trigger contains other triggers and triggers when all of them trigger (logical OR)
```
zhangxiaoli73: Comments not clear. Actually, user can use this api to do early stop.

qiuxin2012: As early stop is just a specific use case of Or, so I add an example to explain this usage.

qiuxin2012:
jenkins passed

PR #2690, opened 2018-11-23
add BlasWrapper

zhangxiaoli73:

wrapper blas layer to dnn layer, then we can use blas layer in dnn graph

## How was this patch tested?

unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
@@ -38,7 +38,7 @@ private[bigdl] class BlasWrapper(val module: AbstractModule[Activity, Activity, 
   gradInput = module.gradInput
 
   private def inferFormats(inputs: Array[MemoryData]): Int = {
-    // remind: here assume all shapes in inputs should be same
+    // reminder: here assume all shapes in inputs should be same
     inputs.foreach(in =>
       require(in.shape.length == 2 || in.shape.length == 4,
       s"just support input shape dim is 2 or 4, but get ${in.shape.length}"))
@@ -51,10 +51,9 @@ private[bigdl] class BlasWrapper(val module: AbstractModule[Activity, Activity, 
   }
 
   override private[mkldnn] def initFwdPrimitives(inputs: Array[MemoryData], phase: Phase) = {
-    // todo: only support model having implemented computeOutputShape
+    // reminder: only support model having implemented computeOutputShape
     val inputShape = inputs.map(in => Shape(in.shape))
     val outputShape = if (inputShape.length == 1) {
-      // single shape
       List(module.computeOutputShape(inputShape(0)))
     } else {
       // multi shape
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
@@ -41,7 +41,7 @@ private[bigdl] class BlasWrapper(val module: AbstractModule[Activity, Activity, 
     // reminder: here assume all shapes in inputs should be same
     inputs.foreach(in =>
       require(in.shape.length == 2 || in.shape.length == 4,
-      s"just support input shape dim is 2 or 4, but get ${in.shape.length}"))
+      s"only input shape dim 2 and 4 supported, but get ${in.shape.length}"))
 
     inputs(0).layout match {
       case Memory.Format.nhwc => Memory.Format.nhwc
@@ -62,7 +62,7 @@ private[bigdl] class BlasWrapper(val module: AbstractModule[Activity, Activity, 
     }
     val outDim = outputShape(0).toSingle().length
     require(outDim == 4 || outDim == 2,
-      s"just support output shape dim is 2 or 4, but get ${outDim}")
+      s"only output shape dim 2 and 4 supported, but get ${outDim}")
 
     val inputFormats = inferFormats(inputs)
     val outputFormats = if (outDim == 4) inputFormats else Memory.Format.nc
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
@@ -21,45 +21,58 @@ import com.intel.analytics.bigdl.mkl.Memory
 import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, Activity, TensorModule}
 import com.intel.analytics.bigdl.tensor.Tensor
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
-import com.intel.analytics.bigdl.utils.Shape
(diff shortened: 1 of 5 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
```diff
@@ -0,0 +1,125 @@
...
+
+import com.intel.analytics.bigdl.Module
+import com.intel.analytics.bigdl.mkl.Memory
+import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, Activity, TensorModule}
+import com.intel.analytics.bigdl.tensor.Tensor
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+import com.intel.analytics.bigdl.utils.Shape
+import spire.syntax.module
+
+// only support wrap bigdl to support nchw or nc
```
wzhongyuan: what else that we cannot support ？

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
```diff
@@ -0,0 +1,125 @@
...
+private[bigdl] class BlasWrapper(val module: TensorModule[Float]) extends MklDnnLayer {
```
wzhongyuan: private val module ?

wzhongyuan: and why it's hard-coded to float ? what if int or fp16?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
zhangxiaoli73: Now only support 2 or 4 dims inputs. Other cases will be error.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
zhangxiaoli73: Now BlasWrapper is only used when converting to dnn graph. As tests are going on, I think not expose BlasWrapper  to users is better.

zhangxiaoli73: As for  hard-coded to float, dnn layer now only supports float tensor, so here tensor data type should be float.
For int or fp16 type, I am not sure how to define the model.

wzhongyuan: I mean the parameter `module`  should be private, right ?

wzhongyuan: In future, we may have in8, fp16 training/inference, so probably we need to support them, just make it generic ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
```diff
@@ -0,0 +1,139 @@
...
+private[bigdl] class BlasWrapper(val module: AbstractModule[Activity, Activity, Float])
+  extends MklDnnLayer {
+
+  require(!module.isInstanceOf[MklDnnModule], "Only support wrapper blas layer to dnn layer")
+
+  output = module.output
+  gradInput = module.gradInput
+
+  private def inferFormats(inputs: Array[MemoryData]): Int = {
+    // remind: here assume all shapes in inputs should be same
```
wzhongyuan: reminder

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
```diff
@@ -0,0 +1,139 @@
...
+
+    inputs(0).layout match {
+      case Memory.Format.nhwc => Memory.Format.nhwc
+      case Memory.Format.nc => Memory.Format.nc
+      case _ => Memory.Format.nchw
+    }
+  }
+
+  override private[mkldnn] def initFwdPrimitives(inputs: Array[MemoryData], phase: Phase) = {
+    // todo: only support model having implemented computeOutputShape
```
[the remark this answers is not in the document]
zhangxiaoli73: sorry, here should not be "todo", but "reminder". Only having implemented computeOutputShape blas layer can wrapped to dnn layer.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
```diff
@@ -0,0 +1,139 @@
...
+    val inputShape = inputs.map(in => Shape(in.shape))
+    val outputShape = if (inputShape.length == 1) {
+      // single shape
```
wzhongyuan: Do we need this comment ?  the if condition will tell us obviously right ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
```diff
@@ -0,0 +1,139 @@
...
+      List(module.computeOutputShape(inputShape(0)))
+    } else {
+      // multi shape
+      val out = module.computeOutputShape(MultiShape(inputShape.toList))
+      if (out.isInstanceOf[MultiShape]) out.toMulti() else List(out)
+    }
+    val outDim = outputShape(0).toSingle().length
+    require(outDim == 4 || outDim == 2,
+      s"just support output shape dim is 2 or 4, but get ${outDim}")
```
wzhongyuan: Only shap dim 2 and 4 supported ?

wzhongyuan:
Only some small comments, other looks good to me

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
zhangxiaoli73: yeah, dim 3 has no corresponding memory format in mkl-dnn.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
zhangxiaoli73: ok, just a reminder.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
wzhongyuan: I mean  make a slight change to the description to make it aligned with other exception

zhangxiaoli73: ok, changed

zhangxiaoli73:
Jenkins pass:
http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/301/

PR #2691, opened 2018-11-23
add dnn output layer

zhangxiaoli73:

add dnn output layer

## How was this patch tested?

unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Output.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Output.scala
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2016 The BigDL Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.intel.analytics.bigdl.nn.mkldnn
+
+import com.intel.analytics.bigdl.mkl.{Memory, MklDnn}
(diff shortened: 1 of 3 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Output.scala
```diff
@@ -0,0 +1,93 @@
...
+    val outputShape = getShape(inputs(0).layout, inputs(0).shape, _outputLayOut)
+    // remind: output memory storage should be heapData
+    _outputFormats = Array(HeapData(outputShape, outputLayOut))
+    _inputFormats = _outputFormats
+
+    (_inputFormats, _outputFormats)
+  }
+
+  override def updateOutput(input: Activity): Activity = {
+    output = input
```
i8run: Why we not reorder the output or copy back to java heap?

i8run: I saw the test case, it solve our problems. But I can't understand why it can work ?

zhangxiaoli73: Just appoint input & output formats, then ReorderManager in DnnGraph will handle the reorder.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Output.scala
```diff
@@ -0,0 +1,93 @@
...
+
+import com.intel.analytics.bigdl.mkl.{Memory, MklDnn}
+import com.intel.analytics.bigdl.nn.abstractnn.Activity
+import com.intel.analytics.bigdl.tensor.{DnnTensor, Tensor}
+
+/**
+ * Convert output to user defined layout and appoint gradOutput layout
+ * @param outputLayOut output memory layout
+ * @param gradOutputLayout gradoutput memory layout,
+ *                         if it is -1, that means gradOutputLayout same with outputLayOut
```
wzhongyuan: same on what ? I think you should clarify it to make it more clear to user,  is here you mean layout format right ?

zhangxiaoli73: ok, changed to more clear  comment

zhangxiaoli73:
jenkins pass:
http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/303/

Issue #2699, opened 2018-12-04
Scala compiler version vulnerability 


wzhongyuan:
https://github.com/intel-analytics/BigDL/network/alert/pom.xml/org.scala-lang:scala-compiler/open

PR #2700, opened 2018-12-04
[Fix] Fix scala compiler security issue

wzhongyuan:
Per https://github.com/intel-analytics/BigDL/network/alert/pom.xml/org.scala-lang:scala-compiler/open
we should upgrade scala compiler issue

## How was this patch tested?
UT and integration test

fixed https://github.com/intel-analytics/BigDL/issues/2699


wzhongyuan:
PR passed
http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/305/

PR #2711, opened 2018-12-25
convert static graph to IR graph and build

zhangxiaoli73:

support converting static graph to ir graph, and build it to dnn graph with dnn engine type

## How was this patch tested?
unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/StaticGraph.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/StaticGraph.scala
@@ -15,13 +15,19 @@
  */
 package com.intel.analytics.bigdl.nn
 
+import com.intel.analytics.bigdl.Module
+import com.intel.analytics.bigdl.mkl.Memory
 import com.intel.analytics.bigdl.nn.Graph.ModuleNode
 import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, Activity}
 import com.intel.analytics.bigdl.nn.tf.ControlDependency
 import com.intel.analytics.bigdl.tensor.Tensor
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+import com.intel.analytics.bigdl.utils.Engine._
+import com.intel.analytics.bigdl.utils.intermediate.{BlasToIR, IRGraph}
 import com.intel.analytics.bigdl.utils.{Node, Util}
+import com.intel.analytics.bigdl.optim.DistriOptimizer._
 
+import scala.collection.mutable.ArrayBuffer
 import scala.reflect.ClassTag
 
 /**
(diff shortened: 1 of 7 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/StaticGraph.scala
```diff
@@ -147,4 +153,59 @@ class StaticGraph[T: ClassTag](
     gradInput = fetchModelGradInput()
     gradInput
   }
+
+  private var inputFormats: Seq[Int] = null
```
wzhongyuan: why seq instead of array ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/StaticGraph.scala
```diff
@@ -147,4 +153,59 @@ class StaticGraph[T: ClassTag](
...
+  }
+
+  /**
+   * convert static graph to ir graph and build according to engine type
+   * @return
+   */
+  def toIRgraph() : IRGraph[T] = {
+    val inFormats = if (inputFormats == null) {
+      logger.warn(s"Please set input formats for graph model, otherwise we" +
+        s"will use default input format NCHW")
```
wzhongyuan: "Input format NCHW by default, Please set explicitly if needed" ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/StaticGraph.scala
```diff
@@ -147,4 +153,59 @@ class StaticGraph[T: ClassTag](
...
+    } else inputFormats
+
+    val outFormats = if (outputFormats == null) {
+      logger.warn(s"Please set output formats for graph model, otherwise we" +
+        s"will use default output format NC")
+      Seq(Memory.Format.nc)
+    } else outputFormats
+
+    val allNodes = forwardExecution
+    if (!BlasToIR[T].convertingCheck(allNodes)) return null
```
wzhongyuan: Do we check the null value where we call this method ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/StaticGraph.scala
wzhongyuan: inputFormats -> inputFormat ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/StaticGraph.scala
```diff
@@ -147,4 +153,59 @@ class StaticGraph[T: ClassTag](
...
+    inFormats.foreach(in => if (in == Memory.Format.nhwc) return null)
```
wzhongyuan: print a warning message saying we cannot support nhwc for now ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/StaticGraph.scala
zhangxiaoli73: As inputs nodes are seq, and inputFormats are corresponding to inputs nodes. 
So I use seq instead of array.

zhangxiaoli73: I think "inputFormats" is more clear as there maybe multi inputs

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/StaticGraph.scala
zhangxiaoli73: ok, changed

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/StaticGraph.scala
zhangxiaoli73: Yeah. I will do check.
As this api is to converting static graph to ir graph. So if converting failed, return null is proper.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/StaticGraph.scala
zhangxiaoli73: ok, changed.

zhangxiaoli73:
jenkins pass: http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/319/

PR #2712, opened 2018-12-27
fix: update mkldnn version to v0.17 issues.

i8run:
Upgrade mkldnn to v0.17.

There're two issues,

1. the padding tensor required. mkl-dnn will use a padding tensor which
   will use more memory, such as 4x1x28x28 to 4x8x28x28(avx2). It will
   pad to times of simd width.
2. the TensorMMap between DenseTensor and DnnTensor. Previous impl
    will allocate DnnTensor when model is created, which will cost too much
    space. So this patch will allocate it at runtime.

## How was this patch tested?

Test on ResNet-50 training. Test on Jenkins.


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/ReorderMemory.scala
```diff
@@ -84,6 +85,11 @@ class ReorderMemory(inputFormat: MemoryData, outputFormat: MemoryData,
 
     // recover to original data
     output = initTensor(realOutput(0))
+
+    if (outputLayout == Memory.Format.nhwc && realOutput(0).isInstanceOf[HeapData]) {
+      output.toTensor[Float].resize(_outputFormats(0).shape)
```
zhangxiaoli73: if not heapdata, output may not have real size, is it right?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/ThreadPool.scala
```diff
@@ -85,6 +94,10 @@ class ThreadPool(private var poolSize: Int) {
       val tid = Thread.currentThread().getId()
       logger.info(s"Set mkl threads to $size on thread $tid")
     }(context)).foreach(Await.result(_, Duration.Inf))
+
+    invokeAndWait2 ((0 until poolSize).map( _ => () => {
```
zhangxiaoli73: Why do you need this?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/ReorderMemory.scala
i8run: The tensor size will always be right.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Linear.scala
```diff
@@ -138,8 +136,8 @@ class Linear(
     updateWithNewTensor(updateOutputTensors, 0, input)
 
     if (isTraining()) {
-      weight.syncToNative()
-      bias.syncToNative()
+      weight.sync()
```
zhangxiaoli73: if not training, when to sync weight and bias to native?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/SpatialConvolution.scala
```diff
@@ -253,8 +250,8 @@ class SpatialConvolution(
     updateWithNewTensor(updateOutputTensors, 0, input)
 
     if (isTraining()) {
-      weight.syncToNative()
-      bias.syncToNative()
+      weight.sync()
```
zhangxiaoli73: if not training, when to sync?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Linear.scala
```diff
@@ -58,26 +58,24 @@ class Linear(
...
   }
 
   override private[mkldnn] def initFwdPrimitives(inputs: Array[MemoryData], phase: Phase) = {
-    val weightShape = inputs(0).shape.length match {
-      case 4 => Array(weight.size(1)) ++ inputs(0).shape.slice(1, 4)
-      case _ => weight.size()
+    val (weightShape, weightLayout) = inputs(0).shape.length match {
+      case 4 =>
+        (Array(weight.size(1)) ++ inputs(0).shape.slice(1, 4),
+          Memory.Format.oihw)
```
wzhongyuan: what's oihw for

i8run: Linear (mkldnn version) will support 4-D input, so the relevant weight will be 4-D. We make all 4-D weight with the format oihw.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Linear.scala
i8run: the previous commit will sync weight and bias at `setMemoryData`. Have pushed a new commit which will sync weight hard code when compiling.

i8run:
Jenkins has passed all test cases. And distributed training of ResNet-50 is right.

PR #2713, opened 2019-01-07
[Enhancement] - Enhance unig test to avoid dynamic resource allocation issue by docker

wzhongyuan:

Java runtime will get available processors to set the executor cores in Spark, but it will not match the one specified in command line of ran from docker. this enhancement is to set fixed number for spark to avoid the inconsistency 

## How was this patch tested?

unit test


wzhongyuan:
Jenkins passed http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/322

PR #2721, opened 2019-01-18
Fix Add operation error when type is Double importing Tensorflow graph

EmiCareOfCell44:
Loading wan Tensor flow graph with and Add operation gives the following result:

"Cannot convert the given tensorflow operation graph to BigDL model. The convert fails
 at node Linear33c12dce/add. Operation type is Add". 

 I had to add the "else if" clause to support the Double Type in this kind of operation importing graph to Bigdl format.

## How was this patch tested?

As the change is so minimal I don´t include a specific unit test. 

I didn´t open issue


wzhongyuan:
@EmiCareOfCell44   Thanks for the fix, we will be running a testing job and then merge it after the job passes

wzhongyuan:
Jenkins passed http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/421/

PR #2724, opened 2019-01-22
fix: move mkldnn computing to a single thread pool

i8run:

Move the computing of mkldnn to a single thread.

  Because if we use the parent thread directly, there will be two bugs,
    1. The child threads forked from parent thread will be bound to core 0
       because of the affinity settings.
    2. The native thread has some unknown thread local variables. So if
       the parent thread exits and is recreated, such as the thread from
       Executors.newFixedThreadPool. The whole app will be segment fault.
  The parent thread means the main thread (Local Mode) or worker thread of
  `mapPartition` (Distributed Mode).

## How was this patch tested?
Jenkins.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Perf.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Perf.scala
@@ -17,7 +17,7 @@
 package com.intel.analytics.bigdl.nn.mkldnn
 
 import com.intel.analytics.bigdl.Module
-import com.intel.analytics.bigdl.mkl.{Memory, MklDnn}
+import com.intel.analytics.bigdl.mkl.{MKL, Memory, MklDnn}
 import com.intel.analytics.bigdl.nn.Graph._
 import com.intel.analytics.bigdl.nn._
 import com.intel.analytics.bigdl.nn.abstractnn.Activity
@@ -28,7 +28,7 @@ import com.intel.analytics.bigdl.numeric.NumericFloat
 import com.intel.analytics.bigdl.tensor.Tensor
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
 import com.intel.analytics.bigdl.utils.RandomGenerator._
-import com.intel.analytics.bigdl.utils.{Engine, T, Table}
+import com.intel.analytics.bigdl.utils.{Engine, T, Table, ThreadPool}
 import org.apache.log4j.Logger
 import scopt.OptionParser
 
@@ -86,34 +86,39 @@ object Perf {
 
       val criterion = CrossEntropyCriterion()
 
-      if (training) {
-        if (model.isInstanceOf[MklDnnContainer]) {
-          model.asInstanceOf[MklDnnContainer]
-            .compile(TrainingPhase, Array(HeapData(inputShape, inputFormat)))
-        } else if (model.isInstanceOf[DnnGraph]) {
-          model.asInstanceOf[DnnGraph].compile(TrainingPhase)
+      Engine.computing.invokeAndWait2(Array(1).map(_ => () => {
+        if (training) {
+          if (model.isInstanceOf[MklDnnContainer]) {
+            model.asInstanceOf[MklDnnContainer]
+              .compile(TrainingPhase, Array(HeapData(inputShape, inputFormat)))
+          } else if (model.isInstanceOf[DnnGraph]) {
+            model.asInstanceOf[DnnGraph].compile(TrainingPhase)
+          }
+          model.training()
+        } else {
+          if (model.isInstanceOf[MklDnnContainer]) {
+            model.asInstanceOf[MklDnnContainer]
+              .compile(InferencePhase, Array(HeapData(inputShape, inputFormat)))
+          } else if (model.isInstanceOf[DnnGraph]) {
+            model.asInstanceOf[DnnGraph].compile(InferencePhase)
+          }
+          model.evaluate()
         }
-        model.training()
-      } else {
-        if (model.isInstanceOf[MklDnnContainer]) {
-          model.asInstanceOf[MklDnnContainer]
-            .compile(InferencePhase, Array(HeapData(inputShape, inputFormat)))
-        } else if (model.isInstanceOf[DnnGraph]) {
-          model.asInstanceOf[DnnGraph].compile(InferencePhase)
-        }
-        model.evaluate()
-      }
+      }))
 
       var iteration = 0
       while (iteration < iterations) {
         val start = System.nanoTime()
-        val output = model.forward(input)
 
-        if (training) {
-          val _loss = criterion.forward(output, label)
-          val errors = criterion.backward(output, label).toTensor
-          model.backward(input, errors)
-        }
+        Engine.computing.invokeAndWait2(Array(1).map(_ => () => {
+          val output = model.forward(input)
+
+          if (training) {
+            val _loss = criterion.forward(output, label)
+            val errors = criterion.backward(output, label).toTensor
+            model.backward(input, errors)
+          }
+        }))
 
         val takes = System.nanoTime() - start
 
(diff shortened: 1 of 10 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -220,8 +220,20 @@ object Engine {
...
+  //   2. The native thread has some unknown thread local variables. So if
+  //      the parent thread exits and is recreated, such as the thread from
+  //      Executors.newFixedThreadPool. The whole app will be segment fault.
+  // The parent thread means the main thread (Local Mode) or worker thread of
+  // `mapPartition` (Distributed Mode).
+  // --------------------------------------------------------------------------
+  // We will only use the `threadPool` in ThreadPool, which is a ExecutorService.
+  // For `context` in ThreadPool, it is the called thread when poolSize is 1.
+  // So many usages of that thread, we will not change it for now.
+  val computing: ThreadPool = new ThreadPool(1)
```
zhangxiaoli73: maybe give more clear name to illustrate this thread is just for mkldnn model.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/ThreadPool.scala
```diff
@@ -81,7 +90,8 @@ class ThreadPool(private var poolSize: Int) {
   def setMKLThread(size: Int): this.type = this.synchronized {
     mklPoolSize = Some(size)
     (1 to poolSize).map(i => Future {
-      ThreadPool.setThreadsOfBackend(size)
+      require(MKL.isMKLLoaded)
+      MKL.setNumThreads(size)
```
zhangxiaoli73: there are two kinds of backends. I think it's better to define two fun setThreadsOfMklBackend and setThreadsOfMkldnnBackend. And different threads call different backends according to their need.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -255,10 +255,20 @@ object DistriOptimizer extends AbstractOptimizer {
...
-              localModel.backward(input, errors)
+              Engine.getEngineType() match {
+                case MklBlas =>
+                  val output = localModel.forward(input)
+                  lossArray(i) = ev.toType[Double](localCriterion.forward(output, target))
+                  val errors = localCriterion.backward(output, target)
+                  localModel.backward(input, errors)
+                case MklDnn =>
+                  Engine.computing.invokeAndWait2(Array(0).map(_ => () => {
+                    val output = localModel.forward(input)
```
zhangxiaoli73: if localModel is IRgraph, there will be computing.invokeAndWait2 calling computing.invokeAndWait2 in IRGraph. Is it ok?

spark/dl/src/test/scala/com/intel/analytics/bigdl/nn/mkldnn/ReorderMemorySpec.scala
```diff
@@ -16,10 +16,76 @@
...
 
 import com.intel.analytics.bigdl.mkl.Memory
+import com.intel.analytics.bigdl.nn.mkldnn.Phase.{InferencePhase, TrainingPhase}
 import com.intel.analytics.bigdl.tensor.Tensor
-import com.intel.analytics.bigdl.utils.BigDLSpecHelper
+import com.intel.analytics.bigdl.utils.Engine
+import org.scalatest.{BeforeAndAfter, FlatSpec, Matchers}
+
+class ReorderMemorySpec extends FlatSpec with Matchers with BeforeAndAfter {
+  "lenet5 with reorder" should "works no segment fault" in {
```
zhangxiaoli73: what's this spec for? I think it should be in pr about updating to 0.17

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Perf.scala
```diff
@@ -86,34 +86,39 @@ object Perf {
...
-        model.training()
-      } else {
-        if (model.isInstanceOf[MklDnnContainer]) {
-          model.asInstanceOf[MklDnnContainer]
-            .compile(InferencePhase, Array(HeapData(inputShape, inputFormat)))
-        } else if (model.isInstanceOf[DnnGraph]) {
-          model.asInstanceOf[DnnGraph].compile(InferencePhase)
+      Engine.computing.invokeAndWait2(Array(1).map(_ => () => {
+        if (training) {
+          if (model.isInstanceOf[MklDnnContainer]) {
```
wzhongyuan: what's difference between dnngraph and mkldnngraph ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -255,10 +255,20 @@ object DistriOptimizer extends AbstractOptimizer {
...
-              val errors = localCriterion.backward(output, target)
-              localModel.backward(input, errors)
+              Engine.getEngineType() match {
+                case MklBlas =>
+                  val output = localModel.forward(input)
+                  lossArray(i) = ev.toType[Double](localCriterion.forward(output, target))
+                  val errors = localCriterion.backward(output, target)
+                  localModel.backward(input, errors)
+                case MklDnn =>
+                  Engine.computing.invokeAndWait2(Array(0).map(_ => () => {
```
wzhongyuan: By using computing thread pool, can we make sure the underlying will use mult-omp thread ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
wzhongyuan: agreed, like `dnnComputing` or just `dnn`

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -321,24 +333,13 @@ object Engine {
...
     val defaultPoolSize: Int = System.getProperty("bigdl.utils.Engine.defaultPoolSize",
       (core * 50).toString).toInt
     if(_default == null || _default.getPoolSize != defaultPoolSize) {
       _default = new ThreadPool(defaultPoolSize)
     }
-
-    if (_io == null) {
-      _io = new ThreadPool(core * 50)
-    }
+    _default.setMKLThread(1)
```
zhangxiaoli73: no need

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Perf.scala
i8run: DnnGraph inherits from StaticGraph directly. And is not the type of MklDnnContainer.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
i8run: No. Here we only consider the raw dnn model.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
i8run: OK.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
i8run: Yes.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/ThreadPool.scala
i8run: Yes. It will be more clear.

spark/dl/src/test/scala/com/intel/analytics/bigdl/nn/mkldnn/ReorderMemorySpec.scala
i8run: It can works on the two pr. In this case, it will tests run the dnn model in the same thread. The result should be no segment fault.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
i8run: I will add some condition to handle the `IRGraph`.

zhangxiaoli73:
LGTM

PR #2736, opened 2019-02-18
fix IRGraph init & Add regualizer

zhangxiaoli73:

1. fix init primitives in IR graph, all primitives should be initialized together, otherwise there will be result error.

2. Add regualizer for linear and convolution dnn layer

## How was this patch tested?

unit tests and training test

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/View.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/View.scala
@@ -19,6 +19,7 @@ package com.intel.analytics.bigdl.nn
 import com.intel.analytics.bigdl.nn.abstractnn.TensorModule
 import com.intel.analytics.bigdl.tensor.Tensor
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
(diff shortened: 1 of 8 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/View.scala
```diff
@@ -99,13 +98,18 @@ class View[T: ClassTag](val sizes: Array[Int])(
...
     } else {
       bse
     }
   }
 
+  private def batchSize(
+    input: Tensor[T], size: Array[Int], numberInputDims: Int, numElements: Int): Int = {
+    batch(input.nDimension(), input.size(), size, numberInputDims, numElements)
+  }
+
```
i8run: Use only 1 api is much better i think.

zhangxiaoli73: ok, changed

i8run:
LGTM.

zhangxiaoli73:
jenkins pass:
http://mlt-bddl01.sh.intel.com:8080/view/PR/job/BigDL-PRVN-lite/358/

PR #2743, opened 2019-02-26
Remove JBoss repo out from pom

LeicongLi:
Remove JBoss repository out from pom.xml, since no dependency is relaying on it.

## How was this patch tested?


wzhongyuan:
No need to do that. it's different from the blacklist jboss library

Issue #2744, opened 2019-03-04
SSD throughput performance issue with blas model running under mkl-dnn


wzhongyuan:
If we have big batch size,  SSD BLAS model will not get performance benefit with mkl-dnn runtime

PR #2748, opened 2019-03-11
[Enhancement] - Speed up BlasWrapper performance under MKL-DNN

zhangxiaoli73:
 Current mkl-dnn implementation is that we only run single model on top of multi-thread enabled mkl-dnn, so if we switch to blas model , we will lack the ability to leverage multi threading. In order to solve the performance drop, we check the layer type and do parallel computing for blas wrapper. 

the underlying layer must be computed in batch (both input and output have same batch size)

## How was this patch tested?

unit tests and performance tests by hand

fixed https://github.com/intel-analytics/BigDL/issues/2744

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
@@ -16,16 +16,19 @@
 
 package com.intel.analytics.bigdl.nn.mkldnn
 
+import breeze.linalg.Axis._1
+import breeze.linalg.dim
 import com.intel.analytics.bigdl.Module
 import com.intel.analytics.bigdl.dataset.MiniBatch
 import com.intel.analytics.bigdl.mkl.{MKL, Memory}
+import com.intel.analytics.bigdl.nn.{DetectionOutputSSD, PriorBox}
 import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, Activity, TensorModule}
 import com.intel.analytics.bigdl.nn.mkldnn.Phase.{InferencePhase, TrainingPhase}
 import com.intel.analytics.bigdl.tensor.Tensor
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+import com.intel.analytics.bigdl.utils.Engine._
 import com.intel.analytics.bigdl.utils._
-import com.intel.analytics.bigdl.utils.intermediate.IRGraph
-import spire.syntax.module
+import org.apache.log4j.Logger
 
 /**
  * wrap blas module to be dnn module,
@@ -52,12 +55,15 @@ private[bigdl] class BlasWrapper(val module: AbstractModule[Activity, Activity, 
   }
 
   private[mkldnn] var needOutputFormats: Boolean = true
-  @transient private var workingModels: Array[Module[Float]] = _
-  @transient private var _subModelNumber : Int = 1
+  private val logger = Logger.getLogger(getClass)
+
+  @transient private var subModels: Array[Module[Float]] = _
+  @transient private var subModelNumber : Int = 1
   @transient private var withMultiThread: Boolean = false
   @transient private var inputBuffer : Array[Activity] = _
   @transient private var tensorBuffer : Array[Tensor[Float]] = _
   @transient private var batchSize : Int = _
+  @transient private var initEnv: Boolean = false
 
   private def inferInputFormats(inputs: Array[MemoryData]): Array[MemoryData] = {
     inputs.map(in => HeapData(in.shape, getFormats(in.shape.length)))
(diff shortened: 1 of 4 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
```diff
@@ -16,13 +16,19 @@
...
 import com.intel.analytics.bigdl.tensor.Tensor
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
-import com.intel.analytics.bigdl.utils.{MultiShape, Shape}
-import spire.syntax.module
+import com.intel.analytics.bigdl.utils.Engine._
+import com.intel.analytics.bigdl.utils._
+import org.apache.log4j.Logger
 
 /**
  * wrap blas module to be dnn module,
```
wzhongyuan: delete `be`

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
```diff
@@ -69,10 +84,82 @@ private[bigdl] class BlasWrapper(val module: AbstractModule[Activity, Activity,
     }).toArray
   }
 
+  /**
+   * Blas layers may not have good performance with mkldnn backend,
```
wzhongyuan: => Blas layers normally does not have competitive performance when running under mkldnn ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
```diff
@@ -69,10 +84,82 @@ private[bigdl] class BlasWrapper(val module: AbstractModule[Activity, Activity,
...
+   * so we can use java multi thread to run blas layers and thread number is
+   * related with batch size and core number.
```
wzhongyuan: => we can leverage multi-threading to resolve bottleneck brought by one model only for dnn,  the parallelism is determined by both bath size and core number, with restrictions that both input and output format must be batched

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
```diff
@@ -69,10 +84,82 @@ private[bigdl] class BlasWrapper(val module: AbstractModule[Activity, Activity,
...
+   * input and output for this module must be in batch.
```
wzhongyuan: => we can leverage multi-threading to resolve bottleneck introduced by one model only for mkl-dnn, the parallelism is determined by both bath size and core number, with restrictions that both input and output format must be batched

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
```diff
@@ -69,10 +84,82 @@ private[bigdl] class BlasWrapper(val module: AbstractModule[Activity, Activity,
...
+   */
+  private def setMultiThreadEnv(input: Activity): Unit = {
+    initEnv = true
+    val multiThread = System.getProperty("multiThread", "false").toBoolean
+    if (this.train && multiThread) {
+      throw new IllegalArgumentException("Please not set multiThread to true for model training")
```
wzhongyuan: what's will be the issue if we set it in training ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
```diff
@@ -87,8 +174,62 @@ private[bigdl] class BlasWrapper(val module: AbstractModule[Activity, Activity,
...
+    } else {
+      if (module.isInstanceOf[DetectionOutputSSD[Float]]) {
+        T(tensorBuffer(0).narrow(dim, index, size),
+          tensorBuffer(1).narrow(dim, index, size), tensorBuffer(2))
+      } else {
+        T.array(tensorBuffer.map(_.narrow(dim, index, size)))
+      }
+    }
+  }
+  private def forwardWithMultiThread(input: Activity): Activity = {
```
wzhongyuan: maybe a better name with `forwardInParallel` ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -220,6 +220,9 @@ object Engine {
   // Thread pool for layer use
   @volatile private var _model: ThreadPool = new ThreadPool(1)
 
+  // Thread pool for blaw wrapper layer
```
wzhongyuan: typo:  `blaw` => `blas`

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
```diff
@@ -69,10 +84,82 @@ private[bigdl] class BlasWrapper(val module: AbstractModule[Activity, Activity,
...
+    if (inputDepth == 1) {
+      tensorBuffer(0) = input.toTensor[Float]
+    } else {
+      val in = input.toTable
+      for (i <- 1 to in.length()) {
+        if (in.get(i).get.isInstanceOf[Table]) return false
+        tensorBuffer(i - 1) = in.get[Tensor[Float]](i).get
+        if (i == 1) batch = tensorBuffer(i - 1).size(1)
+        if (batch != tensorBuffer(i - 1).size(1)
+          && !module.isInstanceOf[DetectionOutputSSD[Float]]) {
```
wzhongyuan: Add a utility as a private class maintain a hashset for all these kinds of layers ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
zhangxiaoli73: ok, changed

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
zhangxiaoli73: ok, good comments

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
zhangxiaoli73: ok, good comments

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
zhangxiaoli73: ok, changed

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
zhangxiaoli73: ok, forwardInParallel

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
zhangxiaoli73: Setting mkl omp thread to 1 may be not suitable for training. We need more tests to verify it.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
```diff
@@ -69,10 +84,82 @@ private[bigdl] class BlasWrapper(val module: AbstractModule[Activity, Activity,
...
+      || !multiThread
+      || (_outputFormats != null && _outputFormats.length != 1)
+      || (_outputFormats != null && _inputFormats != null
+      && _inputFormats(0).shape(0) != _outputFormats(0).shape(0))
+      || !flattenInput(input)
+    ) {
+      return
+    }
+    batchSize = tensorBuffer(0).size(1)
+    val t = batchSize % Engine.coreNumber()
```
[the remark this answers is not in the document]
zhangxiaoli73: ok, how about residue to present the remain number after division.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
```diff
@@ -69,10 +84,82 @@ private[bigdl] class BlasWrapper(val module: AbstractModule[Activity, Activity,
...
 
+  /**
+   * Blas layers may not have good performance with mkldnn backend,
+   * so we can use java multi thread to run blas layers and thread number is
+   * related with batch size and core number.
+   * input and output for this module must be in batch.
+   */
+  private def setMultiThreadEnv(input: Activity): Unit = {
+    initEnv = true
+    val multiThread = System.getProperty("multiThread", "false").toBoolean
```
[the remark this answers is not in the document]
zhangxiaoli73: if moving to class level, I have to use "def" instead of "val". Besides, no other codes need this property in this class

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/BlasWrapper.scala
zhangxiaoli73: Now there is only one special layer, I think add some comments is better. If we have more such cases in the future, we can consider how to design a set to record.

wzhongyuan:
LGTM

zhangxiaoli73:
jenkins pass: http://mlt-bddl01.sh.intel.com:8080/view/PR/job/BigDL-PRVN-lite/432/

PR #2750, opened 2019-03-11
[New Feature] Calculating Scales

LeicongLi:
1. Implement Trait 'MklInt8Convertible', which enriches modules with this trait having the API to calculate scales for input/output/weight, preparing data to perform quantize and converting model from FP32 to INT8.

2. Added extra fields information to save and load from protobuf. 
   1) Flag of whether the feature (FP32 to INT8) is enabled or not
   2) Input dimension mask
   3) Output dimension mask
   4) Weight dimension mask
   5) Input Scales
   6) Output Scales
   7) Weight Scales
  
## How was this patch tested?
1. Enable this feature to Linear, Spatial Convolution, Sequential, ConcatTable
2. Construct a testing model for each type of them
3. Select a sample set of data as testing input
4. Run forward on the testing model with testing input to populate model's output
5. Invoke newly added API calcScales from MklInt8COnvertible trait to populate scales
6. Compared the populated scales with expected values
7. Save the updated model using protocol buffer definition
8. Re-load the model
9. Check if the newly added fields exist in the loaded model  

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/MklInt8Convertible.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/MklInt8Convertible.scala
@@ -18,9 +18,10 @@ package com.intel.analytics.bigdl.nn
 
 import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, Activity}
 import com.intel.analytics.bigdl.tensor.Tensor
+import com.intel.analytics.bigdl.utils.Table
 
 import scala.collection.mutable.ArrayBuffer
-import scala.reflect.ClassTag
+
 
 /**
  * Trait which provides MKL-DNN functionality to convert FP32 model to INT8 model
@@ -49,13 +50,14 @@ trait MklInt8Convertible {
    * 4) ConcatTable: calculate scales for input, output as well as the scales of submodules
    * @param inActivity
    */
-  def calcScales(inputActvt: Activity): Unit = {
+  private[bigdl]def calcScales(inputActvt: Activity): Unit = {
 
     if (inputActvt != null) {
       val module = this.asInstanceOf[AbstractModule[_, _, Float]]
       val outputActvt = module.forward(inputActvt)
 
       module match {
+        case graph: Graph[Float] => calcGraphScales(inputActvt, outputActvt)
         // handlers for BLAS modules
         case linear: Linear[Float@unchecked] =>
           calcModuleScales(inputActvt, outputActvt, linear.weight)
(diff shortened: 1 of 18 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/MklInt8Convertible.scala
```diff
@@ -0,0 +1,427 @@
...
+  // weight scales
+  private[nn] var weightScalesBuffer: ArrayBuffer[Array[Float]] = ArrayBuffer.empty[Array[Float]]
+
+
+  /**
+   * Calculate the required scales for converting int8 modules
+   * Currently there are four type of modules should be supported:
+   * 1) Linear: calculate scales for input, output and weight
+   * 2) Spatial Convolution: calculate scales for input, output and weight
+   * 3) Sequential: calculate scales for input, output as well as the scales of submodules
```
wzhongyuan: How about other containers like graph

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/MklInt8Convertible.scala
```diff
@@ -0,0 +1,427 @@
...
+   * 4) ConcatTable: calculate scales for input, output as well as the scales of submodules
+   * @param inActivity
+   */
+  def calcScales(inputActvt: Activity): Unit = {
```
wzhongyuan: make it invisible out of BigDL, and do a module type check for only supporting Float

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/MklInt8Convertible.scala
```diff
@@ -0,0 +1,427 @@
...
+    if (activity.isTensor) {
+      Array(calcTensorScale(activity.toTensor[Float], mask))
+    } else if (activity.isTable) {
+      activity.toTable.map[Array[Float]](elem => {
+        val index: Any = elem._1
+        val tensor: Tensor[Float] = elem._2.asInstanceOf[Tensor[Float]]
+        calcTensorScale(tensor, mask)
+      }).toArray
+    } else {
+      throw new IllegalArgumentException()
```
wzhongyuan: Add comments in exception

wzhongyuan:
Generally looks good to me, please add graph support and then test fully with a pre-trained model

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/MklInt8Convertible.scala
wzhongyuan: Add comments for graph

LeicongLi:
Thanks for the comments!
Fixed the issues mentioned above. 

Jenkins run result:
http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/426/

PR #2752, opened 2019-03-12
remove some unit tests

zhangxiaoli73:

remove some unit tests
mkldnn spec may be run parallel with other spec, which may not support mkldnn engine type.

## How was this patch tested?

jenkins


wzhongyuan:
How are we going to handle the tests then?

zhangxiaoli73:
Most of mkldnn unit tests have to run with mkldnn, so we have to handle those cases.
Have discussed with Yanzhang, we can move all mkldnn tests in one package and separate a jenkins job for mkldnn from others job, like for torch or python. 

For now, in order to not block others pr in jenkins , I suggest to move those tests and then have time buffer to fix. @wzhongyuan

wzhongyuan:
let's set up a new job for it

PR #2754, opened 2019-03-13
recursively update mask for container module

LeicongLi:
For modules of type Container, when update their mask, their sub-modules should also update the mask accordingly.

## How was this patch tested?
1. Create a module of Container, e.g Sequential
2. Verify the value of mask equal to default value
3. Update mask to a valid non-default value
4. Verify the mask value of main module and its sub-module


LeicongLi:
Jenkins job ran successfully.
http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/429/

PR #2756, opened 2019-03-14
feat: reorder for int8 supports

i8run:

   Supports the reorder between int8 and fp32 format.
   
   1. Because the new data type, we should add a new attribute called dataType
      to the `MemoryData`.
   2. Because we should transfer the scales between FP32->int8 and Int8->FP32.
      we should add two new attributes called `mask` and `scales`.

## How was this patch tested?

Unit tests and the full model tests by manual.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Dropout.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Dropout.scala
@@ -39,6 +39,7 @@ class Dropout(
   override private[mkldnn] def initFwdPrimitives(inputs: Array[MemoryData], phase: Phase) = {
     _inputFormats = inputs.map(x => HeapData(x.shape, format(x.shape)))
     _outputFormats = inputs.map(x => HeapData(x.shape, format(x.shape)))
+    // we should genereate the primitives here, otherwise the initTensor can't get the padding shape
     _outputFormats.map(_.getPrimitive(runtime))
     output = initTensor(_outputFormats.head)
     (_inputFormats, _outputFormats)
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Dropout.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Dropout.scala
@@ -39,6 +39,7 @@ class Dropout(
   override private[mkldnn] def initFwdPrimitives(inputs: Array[MemoryData], phase: Phase) = {
     _inputFormats = inputs.map(x => HeapData(x.shape, format(x.shape)))
     _outputFormats = inputs.map(x => HeapData(x.shape, format(x.shape)))
+    _outputFormats.map(_.getPrimitive(runtime))
     output = initTensor(_outputFormats.head)
     (_inputFormats, _outputFormats)
   }
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/MemoryData.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/MemoryData.scala
@@ -15,14 +15,23 @@
  */
 package com.intel.analytics.bigdl.nn.mkldnn
 
-import com.intel.analytics.bigdl.mkl.{DataType, Memory, MklDnn, Query}
-import com.intel.analytics.bigdl.tensor.{DnnTensor, Tensor}
+import com.intel.analytics.bigdl.mkl._
 
 sealed trait MemoryData extends Serializable {
   def shape: Array[Int]
   def layout: Int
+  def dataType: Int
   def setShape(shape: Array[Int]): Unit
   def setLayout(layout: Int): Unit
+  def setDataType(dataType: Int): Unit
+
+  private var _mask: Int = -1
+  private var _scales: Array[Float] = Array.emptyFloatArray
+
+  def mask: Int = _mask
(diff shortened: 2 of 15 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Dropout.scala
```diff
@@ -39,6 +39,7 @@ class Dropout(
   override private[mkldnn] def initFwdPrimitives(inputs: Array[MemoryData], phase: Phase) = {
     _inputFormats = inputs.map(x => HeapData(x.shape, format(x.shape)))
     _outputFormats = inputs.map(x => HeapData(x.shape, format(x.shape)))
+    _outputFormats.map(_.getPrimitive(runtime))
```
zhangxiaoli73: why need this?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/MemoryData.scala
```diff
@@ -128,14 +152,16 @@ case class HeapData(private var _shape: Array[Int], private var _layout: Int) ex
     s"HeapData([${shape.mkString("x")}], ${layout})"
   }
 
-  override def cloneFormat(): MemoryData = new HeapData(_shape, _layout)
+  override def cloneFormat(): MemoryData = new HeapData(_shape, _layout, _dataType)
```
zhangxiaoli73: func name is not clear.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/ReorderMemory.scala
```diff
@@ -85,17 +90,45 @@ class ReorderMemory(inputFormat: MemoryData, outputFormat: MemoryData,
       }
     }
 
-    val fwdReorderPrimDesc = MklDnn.ReorderPrimitiveDescCreate(
-      realInput(0).getPrimitiveDescription(runtime),
-      realOutput(0).getPrimitiveDescription(runtime))
+    val needNotAttr = inputFormats()(0).dataType == DataType.F32 &&
```
zhangxiaoli73: you can add more comments for those operations

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/ReorderMemory.scala
```diff
@@ -85,17 +90,45 @@ class ReorderMemory(inputFormat: MemoryData, outputFormat: MemoryData,
...
+
     val fwdReorderPrim = MklDnn.PrimitiveCreate2(fwdReorderPrimDesc,
       Array(realInput(0).getPrimitive(runtime)), Array(0), 1,
       Array(realOutput(0).getPrimitive(runtime)), 1)
 
     updateOutputPrimitives = Array(fwdReorderPrim)
 
     // recover to original data
-    output = initTensor(realOutput(0))
+    output = initActivity(realOutput)
```
zhangxiaoli73: why change to initActivity?

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DnnStorage.scala
```diff
@@ -47,7 +47,7 @@ private[tensor] class DnnStorage[T: ClassTag](size: Int) extends Storage[T] {
   // Hold the address of the native array
   @transient var ptr: Pointer = new Pointer(allocate(size))
 
-  override def length(): Int = size
+  override def length(): Int = size / bytes
```
zhangxiaoli73: add more comments to explain this length is not same with blas tensor lenght

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/MemoryData.scala
i8run: Yes. but it's a legacy code. I will put the refactor into another pr.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/ReorderMemory.scala
i8run: OK.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Dropout.scala
i8run: Because the new reorder need them  to get the real/actual size of memory. We should create them here, so the `ReorderMemmory` can call `initActivity` or `initTensor` to get the real/actual native memory size.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/ReorderMemory.scala
i8run: It's the same. `initActivity` will call `initTensor`. I'll delete the change.

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DnnStorage.scala
i8run: It's a wrong change. I'll restore it.

i8run:
Jenkins has passed. http://mlt-bddl01.sh.intel.com:8080/view/PR/job/BigDL-PRVN-lite/453/

PR #2759, opened 2019-03-15
feature: mkldnn int8 layer wise supports

i8run:

This patch adds the support for mkldnn int8. The all changes surround the convolution.
Int8 version of convolution will have signed int_8 or unsigned int_8 as the input. If the
next layer/operation is relu, it will have unsigned int_8 as output, otherwise as signed
int8 output.

The 8 bit (singed) weights will be generated at runtime. Based
on the documents of  [MKL-DNN Int8 Example](https://intel.github.io/mkl-dnn/ex_int8_simplenet.html),
we should give the mkldnn the scales. both for input and output. 

Currently, we generate scales offline (PR, #2750 ,Leicong). As a result, the scales will
be maintained in the model. If you wants to use mkldnn int8 acceleration,  you can call
`model.quantize()`. In other words, there're 2 steps,

1. Generate the scales of the model. This step is case by case because of the different
transformers of different models. An example to be added in separate PR
2. Call `model.quantize()`. You should choose the mkldnn backend through `-Dbigdl.engineType=mkldnn`. 

Now you can use the model like a normal model.

Others,

1. the #2762 have enable the layer wise int8 supports. this patch will focus the change of fusion.

## How was this patch tested?

This patch tested by the unit tests and manual test on VGG16 and ResNet-50.


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnBase.scala
```diff
@@ -69,6 +69,8 @@ trait MklDnnModule extends MklDnnModuleHelper {
   private[mkldnn] def gradOutputFormats(): Array[MemoryData]
 
   private[mkldnn] def gradOutputWeightFormats(): Array[MemoryData]
+
+  def setQuantizeFlag(value: Boolean): this.type
```
wzhongyuan: `setQuantization` ?  no need `flag`

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/BatchNormalization.scala
```diff
@@ -57,7 +57,7 @@ class BatchNormalization[T: ClassTag](
   private val initBias: Tensor[T] = null,
   private val initGradWeight: Tensor[T] = null,
   private val initGradBias: Tensor[T] = null
-)(implicit ev: TensorNumeric[T]) extends TensorModule[T] with Initializable {
+)(implicit ev: TensorNumeric[T]) extends TensorModule[T] with Initializable with MklInt8Convertible {
```
zhangxiaoli73: please sync in dnn layer

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/MklInt8Convertible.scala
```diff
@@ -32,7 +32,7 @@ trait MklInt8Convertible {
   // output dimension mask
   protected var outputDimMask: Int = 0
   // weight dimension mask
-  protected var weightDimMask: Int = 0
+  protected var weightDimMask: Int = 2
```
zhangxiaoli73: why change?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ReLU.scala
```diff
@@ -31,7 +31,7 @@ import scala.reflect.ClassTag
  */
 @SerialVersionUID(1208478077576570643L)
 class ReLU[T: ClassTag](ip: Boolean = false)(
-  implicit ev: TensorNumeric[T]) extends Threshold[T](0, 0, ip) {
+  implicit ev: TensorNumeric[T]) extends Threshold[T](0, 0, ip) with MklInt8Convertible {
```
zhangxiaoli73: please sync in dnn layer

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/AvgPooling.scala
```diff
@@ -77,13 +77,20 @@ class AvgPooling(
 
     paddingTL = Array(pt, pl)
     paddingBR = Array(pb, pr)
-    val outputMD = MklDnn.MemoryDescInit(4, Array(n, c, oh, ow), DataType.F32, Memory.Format.any)
+    val outputMD = MklDnn.MemoryDescInit(4, Array(n, c, oh, ow), inputs(0).dataType,
+      Memory.Format.any)
     val description = MklDnn.PoolingForwardDescInit(
-      PropKind.Forward, algKind,
+      PropKind.ForwardScoring, algKind,
```
zhangxiaoli73: why do you change it?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnGraph.scala
```diff
@@ -40,7 +40,7 @@ class DnnGraph(
   private val _variables: Option[(Array[Tensor[Float]], Array[Tensor[Float]])] = None,
   private val enableExcludeChecking: Boolean = true)
   extends Graph[Float](_inputs, _outputs, _variables)
-  with MklDnnLayer {
+  with MklDnnLayer with MklInt8Convertible {
```
zhangxiaoli73: When will you use scales in dnn graph

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Fusion.scala
```diff
@@ -122,14 +134,30 @@ private[mkldnn] object Fusion {
       if (conv != null) {
         node.element = conv.element
         val element = node.element.asInstanceOf[SpatialConvolution]
-        element.setSumOp(previousNodes(otherNumber).element, otherNumber + 1)
+        element.setSumOp(previousNodes(otherNumber).element, otherNumber)
```
zhangxiaoli73: why delete "+1" operation?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Fusion.scala
```diff
@@ -167,11 +196,15 @@ private[mkldnn] object Fusion {
       weight.mul(alpha)
 
       val bias = convBias.storage().array()(j)
-      val mean = originMean.storage().array()(j)
+      val mean = originMean.storage().array()(j) / bn.scaleFactor.storage().array()(0)
```
zhangxiaoli73: no need to div scaleFactor, as mean is the real running mean.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Fusion.scala
```diff
@@ -151,7 +179,8 @@ private[mkldnn] object Fusion {
     val bnWeight = Tensor[Float].resizeAs(bn.weightAndBias.dense).copy(bn.weightAndBias.dense)
 
     (0 until bn.nOutput).foreach { j =>
-      val variance = originVar.storage().array()(j + originVar.storageOffset() - 1)
+      val variance = originVar.storage().array()(j + originVar.storageOffset() - 1) /
```
zhangxiaoli73: same

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/LRN.scala
```diff
@@ -33,21 +34,52 @@ class LRN(
...
+
     val description = MklDnn.LRNForwardDescInit(
-      PropKind.ForwardTraining, AlgKind.LrnAcrossChannels,
+      kind, AlgKind.LrnAcrossChannels,
       _inputFormats(0).getMemoryDescription(), size, alpha.toFloat, beta.toFloat, k.toFloat)
     fwdPrimDesc = MklDnn.PrimitiveDescCreate(description, runtime.engine, 0L)
-    _outputFormats = Array(MemoryData.primitiveOutput(fwdPrimDesc))
-    workSpaceFormat = MemoryData.primitiveWorkSpace(fwdPrimDesc)
-    workSpace = initTensor(workSpaceFormat)
+    if (inputFormats()(0).scales.nonEmpty) {
```
zhangxiaoli73: duplicate code with other layers

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/MaxPooling.scala
```diff
@@ -36,6 +37,8 @@ class MaxPooling(
   @transient private var paddingBR: Array[Int] = _
   @transient private var fwdPD: Long = _
 
+  var isOnlyForInference = false
```
zhangxiaoli73: private?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Perf.scala
```diff
@@ -402,7 +458,7 @@ object ResNet {
       val output = ReorderMemory(HeapData(Array(batchSize, classNum), Memory.Format.nc)).inputs(fc)
 
       val model = DnnGraph(Array(input), Array(output))
-      modelInit(model)
+      model.reset()
```
zhangxiaoli73: wrong change

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/ReLU.scala
```diff
@@ -45,14 +52,34 @@ class ReLU(value: Float = 0.0f) extends MklDnnLayer {
...
+    _gradInputFormats = Array(MemoryData.operationWant(primDesc, Query.DiffSrcPd))
     updateGradInputPrimitives = Array(
       MklDnn.PrimitiveCreate2(primDesc, Array(_inputFormats(0),
         _gradOutputFormats(0)).map(_.getPrimitive(runtime)), Array(0), 2,
         _gradInputFormats.map(_.getPrimitive(runtime)), _gradInputFormats.length))
     gradInput = initTensor(_gradInputFormats(0))
     (_gradOutputFormats, _gradInputFormats)
   }
+
+  override def calcScales(input: Activity): Unit = {
```
zhangxiaoli73: duplicate

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/SelectTable.scala
```diff
@@ -80,12 +80,12 @@ class SelectTable(val index: Int)(implicit ev: TensorNumeric[Float]) extends Mkl
 
   override private[mkldnn] def initFwdPrimitives(inputs: Array[MemoryData], phase: Phase) = {
     _inputFormats = inputs
-    _outputFormats = Array(inputs(index))
+    _outputFormats = Array(inputs(index - 1))
```
zhangxiaoli73: double check

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Fusion.scala
```diff
@@ -97,6 +100,15 @@ private[mkldnn] object Fusion {
...
 
+  private def findNext(node: Node[AbstractModule[Activity, Activity, Float]])
+  : Seq[Node[AbstractModule[Activity, Activity, Float]]] = {
+    if (node.element.isInstanceOf[Identity]) {
+      node.nextNodes.flatMap(n => findNext(n))
+    } else {
+      Seq(node)
+    }
+  }
+
```
LeicongLi: May I ask the logics of this method?
We are trying to find next nodes of the node passed in as an argument, 
if it is Identity Node, return the next nodes of it's next nodes
else return the node as Sequence.
The logic is kind of unclear.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/BatchNormalization.scala
jason-dai: MklInt8Convertible should only be applicable to MKLDNN layers, no?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Sequential.scala
```diff
@@ -369,6 +397,26 @@ class Sequential extends MklDnnContainer with MklInt8Convertible {
...
+      require(mklDnnModules(i).outputFormats().length == mklDnnModules(i + 1).inputFormats().length)
+      lastOutput = reorderManager.infer(
+        mklDnnModules(i).outputFormats(),
+        mklDnnModules(i + 1).inputFormats(),
+        curOutput
+      )
+
+      i += 1
+    }
+  }
```
LeicongLi: Think about the case:  The Sequential module has only 1 sub-module.
The first time we are trying to enter the while loop, condition (0 < 1 - 1) is false.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/BatchNormalization.scala
i8run: @jason-dai We only save one version of model which is based on the orignal BLAS backend. MKL-DNN layer is not visible to user. and can't use `saveModule` also. With @zhangxiaoli73 work, we can convert the original BLAS model at runtime based on the `engineType` property. 

But Int8 in MKL-DNN needs a model specific infomation (scales). They should be saved somewhere. This trait as a bridge between the two version of layers is a solution.  Then bigdl can save the scales in the model (protobuf).  and will sync them to dnn models when user call `quantize` and the engine is mkldn. 

Of cause, the initial version, only `Convolution` and `Linear` need the scales. But because of the optimization, the scales should be updated based on the fusion rules, such as conv + bn fusion will update conv's output scales to bn's output scales.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/MklInt8Convertible.scala
i8run: OK. I will change it back.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ReLU.scala
i8run: Yes. done

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/AvgPooling.scala
i8run: The `Forwardscoring` will be better performance than `Forward`.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnBase.scala
i8run: Yes.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Fusion.scala
i8run: If current node is not `Identity`, it will return immediately. Otherwise, it will find the next recursively. I use a `flatMap` to flat all arrays to a single seq here.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Fusion.scala
i8run: OK. I have changed back. For original version there's a bug. And currently, there's no.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnGraph.scala
i8run: For test. Same as the `Graph`. I need to make sure the `DnnGraph` can worked fine first.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Fusion.scala
i8run: OK. I will delete them.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/LRN.scala
i8run: Yes. Refactor them now.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/MaxPooling.scala
i8run: Delete this variable now to prevent bugs like `SpatialBatchNormalization`.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Perf.scala
i8run: Yes. I forgot to reset it after test.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/ReLU.scala
i8run: Yes. Refactor them now.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/SelectTable.scala
i8run: Change it back now.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Sequential.scala
i8run: Yes. You are right. I step into the same river again.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/ConcatTable.scala
```diff
@@ -130,6 +130,10 @@ class ConcatTable extends MklDnnContainer with MklInt8Convertible {
     _gradOutputWeightFormats
   }
 
+  private[mkldnn] def reconstruct(): Unit = {
```
zhangxiaoli73: where do you use it?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Dropout.scala
```diff
@@ -39,6 +39,8 @@ class Dropout(
   override private[mkldnn] def initFwdPrimitives(inputs: Array[MemoryData], phase: Phase) = {
     _inputFormats = inputs.map(x => HeapData(x.shape, format(x.shape)))
     _outputFormats = inputs.map(x => HeapData(x.shape, format(x.shape)))
+    // we should genereate the primitives here, otherwise the initTensor can't get the padding shape
+    _outputFormats.map(_.getPrimitive(runtime))
```
zhangxiaoli73: if so, please add same change in BlasWrapper layer

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnGraph.scala
```diff
@@ -397,11 +392,13 @@ class DnnGraph(
    */
   private def fusion(): Unit = {
     if (!this.train) {
-      for (j <- 0 to 1) {
+      for (j <- 0 to 2) {
         var i = forwardExecution.length - 1
         while (i >= 0) {
           if (j == 0) Fusion.fuseModule(forwardExecution(i))
-          if (j == 1) Fusion.fuseCAdd(forwardExecution(i))
+          // we should do this before sum fusion, because it will change the structure of graph
```
zhangxiaoli73: change to more clear comments

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/SelectTable.scala
i8run: For SelectTable, it requires the start index is 1. So it shoud be like this here.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnBase.scala
wzhongyuan: @jason-dai  One driving motivation is we plan to decouple layer implementation from underlying engine and make layer unified, this is planned along with mkl-dnn upgrade soon. Also it should be better to keep format only and runtime engine irrelevant layer definition.  That's why we gave up the option to save DNN defined models

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/BatchNormalization.scala
wzhongyuan: @jason-dai One driving motivation is we plan to decouple layer implementation from underlying engine and make layer unified, this is planned along with mkl-dnn upgrade soon. Also it should be better to keep format only and runtime engine irrelevant layer definition. That's why we gave up the option to save DNN defined models

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnGraph.scala
```diff
@@ -392,11 +392,13 @@ class DnnGraph(
    */
   private def fusion(): Unit = {
     if (!this.train) {
-      for (j <- 0 to 1) {
+      for (j <- 0 to 2) {
         var i = forwardExecution.length - 1
```
wzhongyuan: why 3 times

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Sequential.scala
```diff
@@ -25,10 +25,21 @@ import scala.collection.mutable.ArrayBuffer
 
 class Sequential extends MklDnnContainer with MklInt8Convertible {
 
-  val fuseConvBn = System.getProperty("bigdl.mkldnn.fusion.convbn", "false").toBoolean
-  val fuseBnRelu = System.getProperty("bigdl.mkldnn.fusion.bnrelu", "false").toBoolean
-  val fuseConvRelu = System.getProperty("bigdl.mkldnn.fusion.convrelu", "false").toBoolean
-  val fuseConvSum = System.getProperty("bigdl.mkldnn.fusion.convsum", "false").toBoolean
+  def fuseConvBn: Boolean = {
```
wzhongyuan: I don't think we need to control at layer granularity for fusion,

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/SpatialBatchNormalization.scala
```diff
@@ -62,12 +64,10 @@ class SpatialBatchNormalization(
   val weightAndBias = new TensorMMap(Array(nOutput * 2))
   val gradWeightAndBias = new TensorMMap(Array(nOutput * 2))
 
-  var scaleFactor: Float = 1.0f
-  var biasFactor: Float = 1.0f
-
   private val runningMeanScaled = Tensor[Float].resizeAs(runningMean.dense)
   private val runningVarianceScaled = Tensor[Float].resizeAs(runningVariance.dense)
 
+  // the blank shoud be here, otherwise the runningVarianceScaled will be a method
```
wzhongyuan: what does this mean

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/intermediate/ConversionUtils.scala
```diff
@@ -60,4 +69,27 @@ private[bigdl] object ConversionUtils {
       dataset.coalesce(Engine.nodeNumber(), false)
     } else dataset
   }
+
+  private def getInt8ModelIfNeeds[T: ClassTag](model: Module[T],
```
wzhongyuan: needed

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnGraph.scala
i8run: Should have 3 passes. the second will set the flag about the input whether is negative or not.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Sequential.scala
i8run: I have add the option same as graph fusion.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/SpatialBatchNormalization.scala
i8run: like the comment. otherwise scala compiler will think this is a method.

i8run:
Jenkins passed. http://mlt-bddl01.sh.intel.com:8080/view/PR/job/BigDL-PRVN-lite/474/

PR #2760, opened 2019-03-17
fix conversion accuracy

zhangxiaoli73:
1.  when running blas model with mkldnn backend, we should convert it to graph model first.
2.  running mean /variance of mkldnn batch_norm is not the same meaning in  mklblas, we  should divide running mean /variance with a scaler.

## How was this patch tested?

unit tests and example test


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnGraph.scala
```diff
@@ -481,6 +481,11 @@ class DnnGraph(
     )
     checkDuplicate()
   }
+
+  override def release(): Unit = {
+    super.release()
+    reorderManager.release()
```
i8run: the `super.release()` will not release the whole memory.

zhangxiaoli73: dnn graph is a container, and "super.release()" will release all modules.

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/resnet/TestImageNet.scala
```diff
@@ -47,7 +47,7 @@ object TestImageNet {
       val sc = new SparkContext(conf)
       Engine.init
 
-      val model = Module.loadModule[Float](param.model)
+      val model = Module.load[Float](param.model)
```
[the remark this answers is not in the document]
zhangxiaoli73: ok, changed

zhangxiaoli73:
jenkins pass:
http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/456/

PR #2762, opened 2019-03-18
feature: pr-split of layer wise supports of int8

i8run:

This pr is splited from #2759.  It main focus the layer wise supporting, and not fusion relative logics.
Which will make the code clean.

## How was this patch tested?

Jenkins and unit tests.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Helper.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Helper.scala
@@ -1,0 +1,141 @@
+/*
+ * Copyright 2016 The BigDL Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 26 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Helper.scala
```diff
@@ -0,0 +1,141 @@
+/*
```
wzhongyuan: I think other similar class are named `Util` ?

i8run: Updated. Renamed to `Utils` now.

i8run:
Jenkins passed. http://mlt-bddl01.sh.intel.com:8080/view/PR/job/BigDL-PRVN-lite/458/

wzhongyuan:
LGTM

PR #2769, opened 2019-03-20
[Branch-0.8] fix: the influence of default value of fusion

i8run:

When writing these test cases, the default value of fusion is `false`. But now the default value is `true`. But the test case should use the `false`.

## How was this patch tested?

Jenkins and NB.


i8run:
http://mlt-bddl01.sh.intel.com:8080/view/PR/job/BigDL-PRVN-lite/489/

PR #2771, opened 2019-03-20
[Branch-0.8] fix: performance regression on resnet50

i8run:
This patch fix the performance regression issue of ResNet-50 after refactor.
The u8 to s8 or s8 to u8 needs no reorder on this case.

## How was this patch tested?
Jenkins and example with manual tests.


wzhongyuan:
Please syn in master as well

PR #2772, opened 2019-03-20
[Branch-0.8] fix: inplace of input/output and weight dimension error

i8run:

This patch fix the accuracy issue of non bn models after the refactor.

Some layer's input and output use the same memory. We can't do forward in the
`calcScales`. Because at that time, the input has been changed, its scales maybe
not right. Such as,

Seqeuntail().add(Conv).add(ReLU)

it will do two steps, seq.forward(input) first. and when go into the ReLU, it
will do another forward, so the input will be the output. And scales will be
wrong.

For convolution's weight, the dimension always is 5, although the group number
is 1. But for dnn convolution, if there's no group, the weight's dimension
should be 4.

## How was this patch tested?
Jenkins and example tests.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/MklInt8Convertible.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/MklInt8Convertible.scala
@@ -55,15 +55,17 @@ trait MklInt8Convertible {
 
     if (inputActvt != null) {
       val module = this.asInstanceOf[AbstractModule[_, _, Float]]
-      val outputActvt = mkldnn.Utils.getOutput(module, inputActvt)
+      // do not forward here, because the input maybe not the real input
+      // such as th ReLU(true) will do the computing inplace
+      val outputActvt = module.output.asInstanceOf[Activity]
 
       module match {
         case graph: Graph[Float] => calcGraphScales(inputActvt, outputActvt)
         // handlers for BLAS modules
         case linear: Linear[Float@unchecked] =>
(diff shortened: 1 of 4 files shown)


wzhongyuan:
please update in master as well

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/MklInt8Convertible.scala
```diff
@@ -282,7 +283,14 @@ trait MklInt8Convertible {
   private def getWeight(module: AbstractModule[_, _, Float]): Tensor[Float] = {
     if (module != null) {
       // the getParameters will flatten the weight and bias, it's wrong
-      module.parameters()._1(0)
+      val weight = module.parameters()._1(0)
+      // if the weight is came from nn.SpatialConvolution and the nGroup is 1,
+      // we need to skip the first dimension
+      if (module.isInstanceOf[SpatialConvolution[Float]] && weight.size(1) == 1) {
```
LeicongLi: Hi Yanzhang, could you please explain the reason why we need to skip the first dimension of nn.SpatialConv with nGroup 1? So we could take this into consideration in future development, thanks!

i8run: Got it.

i8run:
Jenkins passed. http://mlt-bddl01.sh.intel.com:8080/view/PR/job/BigDL-PRVN-lite/493/

PR #2773, opened 2019-03-20
[Branch-0.8] fix: the blas wrapper has no scales

i8run:

We should check the `from` and `to` when copy the `MemoryData`. If the model is mixed with blas layer. One of them maybe null.

## How was this patch tested?

Jenkins and manual.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Utils.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Utils.scala
@@ -25,13 +25,19 @@ import com.intel.analytics.bigdl.utils.T
 
 private[bigdl] object Utils {
   def copyMaskAndScales(from: MemoryData, to: MemoryData): Unit = {
-    if (to.scales.isEmpty) {
+    // here we check the from and to, they should not be null ideally
+    // but if the model is mixin with blas layer, it may be null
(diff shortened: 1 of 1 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Utils.scala
```diff
@@ -25,13 +25,19 @@ import com.intel.analytics.bigdl.utils.T
 
 private[bigdl] object Utils {
   def copyMaskAndScales(from: MemoryData, to: MemoryData): Unit = {
-    if (to.scales.isEmpty) {
+    // here we check the from and to, they should not be null ideally
+    // but if the model is mixin with blas layer, it may be null
+    if (from != null && to != null && to.scales.isEmpty) {
```
zhangxiaoli73: what will happen if "from" has empty scales?

i8run: Nothing. The  code is be compatible with empty from scales.

zhangxiaoli73:
LGTM

i8run:
Jenkins has passed. http://mlt-bddl01.sh.intel.com:8080/view/PR/job/BigDL-PRVN-lite/494/

PR #2774, opened 2019-03-21
fix: performance regression on resnet50

i8run:

This patch fix the performance regression issue of ResNet-50 after refactor.
The u8 to s8 or s8 to u8 needs no reorder on this case.

## How was this patch tested?
Jenkins and example with manual tests.


i8run:
Jenkins has passed. http://mlt-bddl01.sh.intel.com:8080/view/PR/job/BigDL-PRVN-lite/530/

PR #2777, opened 2019-03-21
[Backport to master] fix softmax with narrowed input

zhangxiaoli73:

softmax will use input array storage directly without considering its storage offset.

## How was this patch tested?

unit test


i8run:
Jenkins has passed.

PR #2778, opened 2019-03-21
fix: the blas wrapper has no scales

i8run:

We should check the from and to when copy the `MemoryData`. If the model is mixed with blas layer. One of them maybe `null`.

## How was this patch tested?
Jenkins and manual.


i8run:
Jenkins has passed. http://mlt-bddl01.sh.intel.com:8080/view/PR/job/BigDL-PRVN-lite/526/

PR #2779, opened 2019-03-21
fix: inplace of input/output and weight dimension error

i8run:
This patch fix the accuracy issue of non bn models after the refactor.

Some layer's input and output use the same memory. We can't do forward in the
calcScales. Because at that time, the input has been changed, its scales maybe
not right. Such as,

Seqeuntail().add(Conv).add(ReLU)

it will do two steps, seq.forward(input) first. and when go into the ReLU, it
will do another forward, so the input will be the output. And scales will be
wrong.

For convolution's weight, the dimension always is 5, although the group number
is 1. But for dnn convolution, if there's no group, the weight's dimension
should be 4.
## How was this patch tested?

Jenkins and example tests.


i8run:
Jenkins has passed. http://mlt-bddl01.sh.intel.com:8080/view/PR/job/BigDL-PRVN-lite/524/

PR #2783, opened 2019-03-24
fix: allocate too much memory of mkldnn models

i8run:

The `realSize` of DnnTensor means the element number of this tensor. But mkldnn will return the bytes of memory, we should make it to the elements number.

## How was this patch tested?
Jenkins and model training.


i8run:
jenkins has passed. http://mlt-bddl01.sh.intel.com:8080/view/PR/job/BigDL-PRVN-lite/522/

PR #2791, opened 2019-04-02
[Fix] doc : some changes for scalaUserGuide and release links according to …

louie-tsai:
…v0.8.0 release

1. issues for v0.8.0 download link
2. add bigdl-core github info in the "Build BigDL-core" section
3. setting ICC environment variables before users compile BigDL-core with ICC
4. remove descriptions about missing dist/bin/bigdl.sh file
5. add some mkl-dnn verbose for mkldnn validation

## How was this patch tested?

manual tests on a Cascade Lake platform.

### DIFF
--- a/docs/docs/ScalaUserGuide/build-bigdl-core.md
+++ b/docs/docs/ScalaUserGuide/build-bigdl-core.md
@@ -135,7 +135,6 @@ can use `brew`.
 
 We use maven profile to control the build process. For different platforms has different profiles.
 
-
 | Platform      | Profile | Command                      |
 | -----         | :--:    | :--:                         |
 | Linux         | linux   | `mvn clean package -P linux` |
@@ -156,4 +155,3 @@ There two ways to deploy. You can use `mvn deploy -P deploy` at the end.
 ```
 mvn clean deploy -P 'linux' -pl 'bigquant/bigquant-java-x86_64-linux'
 ```
-
--- a/docs/docs/ScalaUserGuide/build-bigdl-core.md
+++ b/docs/docs/ScalaUserGuide/build-bigdl-core.md
@@ -135,8 +135,6 @@ can use `brew`.
 
 We use maven profile to control the build process. For different platforms has different profiles.
 
-$source ${Parallel_Studio_XE_HOME}/compilers_and_libraries/linux/bin/compilervars.sh
- intel64
 
 | Platform      | Profile | Command                      |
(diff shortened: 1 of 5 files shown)


docs/docs/ScalaUserGuide/build-bigdl-core.md
```diff
@@ -125,6 +135,9 @@ can use `brew`.
 
 We use maven profile to control the build process. For different platforms has different profiles.
 
+$source ${Parallel_Studio_XE_HOME}/compilers_and_libraries/linux/bin/compilervars.sh
```
wzhongyuan: why this line ?

louie-tsai: without sourcing ICC environment variable from Parallel Studio installation, I would see an compilation error due to missing icc command.

wzhongyuan: but this is for your environment specific right ? we don't want this from framework's perspective

louie-tsai: Hi Jerry.  the only difference among my environment and other users' environment should be where they install Parallel Studio XE. therefore, I make the parallel studio installation path as a variable {Parallel_Studio_XE_HOME}, and users need to fill path themselves.  
we can add some explanation there. We could also remove this line, and let users handle it by themselves. what is your preference?

wzhongyuan: My understanding is  Parallel Studio is not required for compiling bigdl core right ?  I would recommend to remove it to avoid confusion

wzhongyuan: My understanding is  Parallel Studio is not required for compiling bigdl core right ?  I would recommend to remove it to avoid confusion

louie-tsai: sure. will do.

louie-tsai: remove the line about setting ICC env variable

wzhongyuan:
@louie-tsai   I ran a successful PASS for you http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/548/, merging it now.

wzhongyuan:
@louie-tsai  can you raise the same for branch-0.8

PR #2794, opened 2019-04-09
test: should compare the right grad input

i8run:

This test case error is found when do mkldnn update. Because the grad input of convolution sometimes will be paded, so we should compare the nchw format gradinput.

## How was this patch tested?
Jenkins with v0.18 mkldnn library.


i8run:
Jenkins passed. http://mlt-bddl01.sh.intel.com:8080/view/PR/job/BigDL-PRVN-lite/554/

PR #2804, opened 2019-05-05
[New Feature] Enable Dilation in MKL-DNN Convolution Layer

LeicongLi:
Enable Dilation in MKL-DNN Convolution Layer

## How was this patch tested?
1. Create convolution layers both in BLAS and MKLDNN with dilation of one, which is just the regular convolution, and compare their output and weight
2. Create convolution layers both in BLAS and MKLDNN with dilation of two, and compare their output and weight.


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/SpatialConvolution.scala
```diff
@@ -69,7 +69,9 @@ class SpatialConvolution[T: ClassTag](
   val initGradWeight: Tensor[T] = null,
   val initGradBias: Tensor[T] = null,
   val withBias: Boolean = true,
-  val format: DataFormat = DataFormat.NCHW
+  val format: DataFormat = DataFormat.NCHW,
+  private final val dilationW: Int = 1,
+  private final val dilationH: Int = 1
```
wzhongyuan: can we try to load a model to see if it works ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/SpatialConvolution.scala
```diff
@@ -91,6 +138,20 @@ class SpatialConvolution(
   private var _sumInput = false
 
   def relu: Boolean = _relu
+
+  // get padding type
+  private val paddingType: PaddingType.Value = getPaddingType()
+
+  // Parameters for Dilated Convolution
```
wzhongyuan: use `/***
*`  for multi line comments

Issue #2805, opened 2019-05-06
bug: weight mask of Int8 will be overrided by container's setWeightDimMask.


i8run:
For example, a seq model like,

```
val model = seq.add(conv1).add(conv2)
model.setWeightDimMask(1)
conv2.setWeightDimMask(3)
model.saveModule("xxx")

val loadedModel = Module.loadModule("xxx")
```

The conv2's weight mask will be 1.

PR #2807, opened 2019-05-06
bugfix - set mask for container

LeicongLi:
Add an argument 'recursiveFlag' in setMask series api
when set to true, it would set dimension mask for this module and all of its submodules,
otherwise set dimension mask only to itself.

P.S. 
In the previous version, the setMask api include recursion, and 'recursiveFlag' is false by default,
therefore I need to set the flag to true for existing code. 

## How was this patch tested?
For setInputDimMask, setOutputDimMask & setWeightDimMask
1. create a container module (a Sequential Module in my testcase)
2. add 2 non-container module (Convolution module)
3. Verify the mask value of Sequential and Convolution module inside is the default value 0
4. Modify Sequential's mask by calling set*XXX*DimMask functions with recursiveFlag off, verify only the Sequential's mask changed, and submodules remain unchanged
5. Make set*XXX*DimMask call on Sequential Module with recursive flag, verify all the modules including Sequential and Convolution modules are changed
6. Make set mask function calls to Convolution modules individually, verify only themselves are effected afterwards.

Jenkins result url:
http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/574/

fixed https://github.com/intel-analytics/BigDL/issues/2805


wzhongyuan:
`setMask` is not a friendly API to users, we need to find a way to hide these from users @i8run

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/MklInt8Convertible.scala
```diff
@@ -311,14 +311,14 @@ trait MklInt8Convertible {
    * @param mask value of input dimension mask to be set
    * @return Unit
    */
-  def setInputDimMask(mask: Int) : Unit = {
+  def setInputDimMask(mask: Int, recursiveFlag: Boolean = false) : Unit = {
```
wzhongyuan: `recursiveFlag` is not a meaningful name

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/ModuleSerializable.scala
```diff
@@ -353,17 +353,17 @@ trait ModuleSerializable extends Loadable with Savable{
      val weightScales = protobufModel.getWeightScalesList.iterator().asScala
       .map(attrValueToFloatArray)
 
-     module.setInputDimMask(protobufModel.getInputDimMasks)
+     module.setInputDimMask(protobufModel.getInputDimMasks, false)
```
wzhongyuan: it's false by fault right ? no need to make change if so

wzhongyuan:
LGTM

PR #2808, opened 2019-05-09
[Enhancement]: Scala Reflection: get default value for constructor parameters

LeicongLi:
There would be an error if we try to load a module from existing protobuf file while some modules get updated with new constructing parameters.
To solve this, we include the functionality to get parameter's default value if this parameter has not been saved in protobuf model. 

## How was this patch tested?
To be tested

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/ModuleSerializable.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/ModuleSerializable.scala
@@ -162,7 +162,6 @@ trait ModuleSerializable extends Loadable with Savable{
         }
         args(i) = pvalue
       }
-      println(i, args(i))
       i += 1
     })
    constructorMirror.apply(args : _*).
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/ModuleSerializable.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/ModuleSerializable.scala
@@ -128,10 +128,10 @@ trait ModuleSerializable extends Loadable with Savable{
     /*
     https://www.scala-lang.org/api/2.10.7/#scala.reflect.api.Symbols$Symbol
     this line tries to get companion object of the class;
-    through the companion, default values can be access by calling
+    through the companion, default values can be accessed by calling
     some static methods created by scala compiler, however it does not work when
     the class is not a case class or has not defined a companion, which in this case,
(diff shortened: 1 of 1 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/ModuleSerializable.scala
```diff
@@ -121,26 +121,49 @@ trait ModuleSerializable extends Loadable with Savable{
...
+          val attrValue = modelAttributes.get(pname)
+          DataConverter.getAttributeValue(context, attrValue)
+        } else { // parameter not found, get its default value
+          getPrimCtorDefaultParamValue(instanceMirror, param, i)
         }
-        i += 1
-      })
+        args(i) = pvalue
+      }
+      println(i, args(i))
```
wzhongyuan: remove this line

LeicongLi: removed

LeicongLi:
jenkins result: http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/580/

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/ModuleSerializable.scala
```diff
@@ -121,26 +121,48 @@ trait ModuleSerializable extends Loadable with Savable{
...
-          val value = DataConverter.getAttributeValue(context, attribute)
-          args(i) = value
+
+    val clsMirror = universe.runtimeMirror(cls.getClassLoader)
+    val clsSymbol = clsMirror.classSymbol(cls)
+
+    /*
+    https://www.scala-lang.org/api/2.10.7/#scala.reflect.api.Symbols$Symbol
+    this line tries to get companion object of the class;
+    through the companion, default values can be access by calling
```
wzhongyuan: accessed

wzhongyuan:
Have your tested with both 2.10 and 2.11 ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/ModuleSerializable.scala
LeicongLi: fixed typo

LeicongLi:
> Have your tested with both 2.10 and 2.11 ?

Yes.
Tested using both 2.10 and 2.11 to load existing trained models, such as ResNet, DenseNet, MobileNet, etc.

PR #2802, opened 2019-04-28
[New feature & fix] Add layer-wise adaptive rate scaling optimizer

Menooker:
## What changes were proposed in this pull request?

 * Add LARS optimizer: Layer-wise scaled. Also with utility functions to build a set of LARS optim for a container.
 * Bug fix: The gradient block id of AllReduceParameter is originally composed of {id}{pidTo}gradientBytes{pidFrom}. But the combination of {id}{pidTo} will cause ambiguity. e.g., "112" can be {1}{12} or {11}{2}. Now a "_" is added to separate id from pidTo

## How was this patch tested?

UT

PR #2810, opened 2019-05-13
enhancement: enable whole gradient optim methods in DistriOptimizer

Menooker:

 * For multiple layers, use one shared AllReduceParameter instead of different AllReduceParameters for each layer.
 * Add a new flag in the optimizer class, to indicate that whether updating the weights should be delayed until the next iteration. If the flag is enabled, DistriOptimizer will update the weights at the beginning of each iterations, rather than at the end of each iteration. The flag is useful for the optim methods that require the global weights/gradient, such as LARS. 

## How was this patch tested?

UT

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
@@ -16,40 +16,32 @@
 
 package com.intel.analytics.bigdl.optim
 
-import com.intel.analytics.bigdl.{Module, _}
-import com.intel.analytics.bigdl.dataset.{DataSet, DistributedDataSet, MiniBatch, PaddingParam, Sample, SampleToMiniBatch}
-import com.intel.analytics.bigdl.nn.{Container, Module, Utils}
+import com.intel.analytics.bigdl.dataset.{DistributedDataSet, MiniBatch, PaddingParam, Sample}
+import com.intel.analytics.bigdl.models.utils.{CachedModels, ModelBroadcast}
+import com.intel.analytics.bigdl.nn.mkldnn.Phase.TrainingPhase
+import com.intel.analytics.bigdl.nn.mkldnn.{DnnGraph, MklDnnContainer, MklDnnModule}
+import com.intel.analytics.bigdl.nn.{Container, Graph, Module, Utils}
 import com.intel.analytics.bigdl.parameters.{AllReduceParameter, ParameterProcessor}
-import com.intel.analytics.bigdl.nn.{Container, Module, Utils}
-import com.intel.analytics.bigdl.parameters.AllReduceParameter
-import com.intel.analytics.bigdl.tensor.{FloatType, Tensor}
+import com.intel.analytics.bigdl.tensor.Tensor
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
 import com.intel.analytics.bigdl.utils._
+import com.intel.analytics.bigdl.utils.intermediate.{ConversionUtils, IRGraph}
+import com.intel.analytics.bigdl.visualization.{TrainSummary, ValidationSummary}
+import com.intel.analytics.bigdl.{Module, _}
 import java.io.{File, FilenameFilter}
 import java.text.SimpleDateFormat
 import java.util.Calendar
-
-import com.intel.analytics.bigdl.models.utils.{CachedModels, ModelBroadcast}
-import com.intel.analytics.bigdl.nn.abstractnn.Activity
-import com.intel.analytics.bigdl.nn.mkldnn.Phase.TrainingPhase
-import com.intel.analytics.bigdl.nn.mkldnn.{DnnGraph, MklDnnContainer, MklDnnLayer, MklDnnModule}
-import com.intel.analytics.bigdl.utils.intermediate.IRGraph
 import org.apache.commons.lang.exception.ExceptionUtils
-import com.intel.analytics.bigdl.visualization.{TrainSummary, ValidationSummary}
 import org.apache.log4j.Logger
-import org.apache.spark.network.netty.SparkTransportConf
-import org.apache.spark.{SparkContext, TaskContext}
-import org.apache.spark.rdd.{RDD, ZippedPartitionsWithLocalityRDD}
-
+import org.apache.spark.TaskContext
+import org.apache.spark.rdd.RDD
 import scala.collection.mutable
 import scala.collection.mutable.ArrayBuffer
 import scala.concurrent.Future
 import scala.reflect.ClassTag
-import com.intel.analytics.bigdl.nn.{Container, Graph, Module, Utils}
-import com.intel.analytics.bigdl.nn.mkldnn.{DnnGraph, MklDnnContainer, MklDnnLayer}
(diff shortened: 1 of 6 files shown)


wzhongyuan:
@qiuxin2012   Please help take a look as well

jason-dai:
1) Don't merge two changes (using one shared AllReduceParameter and supporting LARS) in one PR; instead, split them into two separate PRs

2) Why can't you collect the global info for LARS using `ParameterProcessor` (similar to gradient clipping)? Defining a new flag like `isWeightUpdateDeffered` makes the code complex and hard to maintain, and can introduce potential performance degradation.

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -808,15 +890,18 @@ class DistriOptimizer[T: ClassTag] (
     state("computeThresholdbatchSize") = computeThresholdbatchSize
     state("maxDropPercentage") = maxDropPercentage
     state("isLayerwiseScaled") = Utils.isLayerwiseScaled(_model)
+    state("isWeightUpdateDeferred") = isWeightUpdateDeferred
```
qiuxin2012: What does `isWeightUpdateDeferred` mean?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
```diff
@@ -212,13 +218,21 @@ object DistriOptimizer extends AbstractOptimizer {
...
+          }
+          val gradientBuffer = localGradientBuffer.asInstanceOf[Tensor[T]]
           /*
             Note: All models in `cached` share the same storage for weights, so we only need to
             copy the weights from parameter server into the first model's weights.
            */
-          val weightsResults = parameters.values.map(p =>
-            p.getWeights(cached.modelWeights.head.narrow(1, p.paramOffset, p.size))
-          ).toArray
+          val weightsResults = if (!isWeightUpdateDeferred || iteration == 0) {
```
qiuxin2012: What will happen if we resume a training?

Menooker:
A new PR for using one shared AllReduceParameter only.
 
https://github.com/intel-analytics/BigDL/pull/2814

The new PR is the preliminary changes for "deferring updates to weights"

Menooker:
@jason-dai 
> 1. Don't merge two changes (using one shared AllReduceParameter and supporting LARS) in one PR; instead, split them into two separate PRs

Another PR is submitted for "Using one shared allreduceparameter". Please review that, thanks.

> 2. Why can't you collect the global info for LARS using `ParameterProcessor` (similar to gradient clipping)? Defining a new flag like `isWeightUpdateDeffered` makes the code complex and hard to maintain, and can introduce potential performance degradation.

`ParameterProcessor` is not helpful in this case, since `ParameterProcessor` seems not to be designed to alter the model weights. The weights should be updated by `DistriOptimizer`, as is done currently. Also I think that `ParameterProcessor` in BigDL does not collect the whole global gradients. e.g., `L2NormParameterProcessor` only has aggregated gradients of the node's own partition.

The updated code in this PR removed the flag `isWeightUpdateDeffered` and simplified the logic of fetching the "whole gradients". There is a internal state "requiresWholeGradients" which is automatically inferred by the trait "WholeGradientOptimMethod"

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizer.scala
Menooker: The updated code can fall back to fetching global weights if the local model cache does not have the updated weights

jason-dai:
Fetching all gradients in every partition will potentially be a big scalability bottleneck, which can defeat the purpose of LARS.  

You may simply define a new `ParameterProcessor`, such as `LarsProcessor`, which can collect the global info (such as the trust for each layer) in collectGlobalData; then you can update the weight in the same way as other optim methods.

Menooker:
@jason-dai  Hi, 
> Fetching all gradients in every partition will potentially be a big scalability bottleneck, which can defeat the purpose of LARS.

LARS algrithm needs the whole gradients for each layer. Our problem is that a layer's gradient may lay over different partitions. But nodes can only get aggragated gradients of its local partition efficiently.  Splitting a layer into different partitions may not ensure the correctness of LARS. So we need to fetch the whole gradients.

Note that the changes in this PR does not add communication costs. If the optimizer fetches all gradients in every partition, it does not need to fetch weights in the next iteration (as is done in current code). A vector of the size of weights is always transferred in original code or in this PR. The only performance concern is that every node needs to compute the whole weights. However, it has no impact on scalability.

I have tested the performance. The new optimizer does not have significant degeneration on performance.

Menooker:
@jason-dai 
> You may simply define a new `ParameterProcessor`, such as `LarsProcessor`, which can collect the global info (such as the trust for each layer) in collectGlobalData; then you can update the weight in the same way as other optim methods.

LARS may need global gradients to get the scale for each layer. If you fetch global gradient in `ParameterProcessor.collectGlobalData` and use the old code path to update the weights, you may transferred two vectors (gradients & weights) rather than one.

jason-dai:
You just need to compute the per layer "trust" for LARS, no?

Menooker:
Yes. But when a layer lies across several partitions, it will be more complicated. E.g. the index range of a large layer (say, "Linear1") is [1500, 2100) and the whole gradients of the model is in index range [0, 10000). There are 10 nodes(partitions).  So the layer "Linear1" lies in partitions of nodes 0,1,2. None of a single node in 0~2 knows what the gradient "Linear1" is like, unless they do some communications. And LARS needs the norm of the whole layer "Linear1" to scale the gradients.  For node 1, it only knows the range [1000,2000).  Nodes 0~2 needs to communicate with each other to get the full range of layer gradient in [1500,2100).

And yes, you can just let node 1 fetch gradients from nodes 0 and 2, but fine-grained communications may cause more complexity in the code and imbalance of communication workload. Also, there are more data to be transferred - node 1 fetches data from nodes 0 and 2.

jason-dai:
You can simply compute the `trust` for all layers in `ParameterProcessor.collectGlobalData`, yes?

Menooker:
Yes I can. But 1) I think the ParameterProcessor design may not be "simple" if considering a layer may lie across partitions 2) it involves more communications.

Please first note that this PR has already used one shared AllreduceParameter for all layers. 
In ParameterProcessor, if a layer lies across the partitions of three nodes, each of the three nodes needs to  **manually** fetch the remaining parts of the layer from the other two nodes, to get the whole gradient tensor. That is because each node has only direct access to the slice of gradients within its own partition. 

This communication brings **imbalance** of the workload, since some nodes don't need to fetch gradients from other nodes, and others have to. Also, **more data** is transferred than in this PR. If ParameterProcessor is used to calculate LARS trust, the global weights still needs to be fetched by `getWeights()` in next iteration. 

If we are not considering the strict correctness of LARS (in other words, if we allow the layers to be truncated), the current implementation of LarsSGD (https://github.com/intel-analytics/BigDL/pull/2802) (Already merged) and using one shared AllreduceParameter (https://github.com/intel-analytics/BigDL/pull/2814) will do the job.

Menooker:
Maybe I get your idea... A reduceByKey can do the job with key as the layer name.

PR #2813, opened 2019-05-13
[New feature] Add transformer for language model

zhangxiaoli73:

Add transformer to support language model, including:
. Layer normalization implement
. Prepare decode implement
. Transformer implement (only support LM now)
. Expand dims when handing two tensors with different size
. Optimize backward graph generation to keep topology order
. python api support

## How was this patch tested?

unit tests


pyspark/bigdl/nn/layer.py
```diff
@@ -930,6 +930,21 @@ def __init__(self, hidden_size, filter_size, relu_dropout, bigdl_type="float"):
         super(FeedForwardNetwork, self).__init__(None, bigdl_type,
                                         hidden_size, filter_size, relu_dropout)
 
+class TransformerLayer(Layer):
+
+    '''
+    Implementation TransformerLayer
+    >>> ffn = TransformerLayer(20, 4, 2, 3, 1, 0.1, 0.1, 0.1)
```
wzhongyuan: ffn -> transformer ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/CAddTable.scala
```diff
@@ -38,6 +41,31 @@ class CAddTable[T: ClassTag, D: ClassTag](val inplace: Boolean = false)(
 
   output = Tensor[D]()
 
+  private def canFastBroadcast(tensor: Tensor[D], other: Tensor[D]): Boolean = {
+    if (tensor.nDimension != other.nDimension()) {
+      return false
+    }
+    var d = other.nDimension()
+    while(d > 0) {
+      if (tensor.size(d) != 1 && tensor.size(d) != other.size(d)) {
```
wzhongyuan: why we need to check for non-squeezed tensor ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/CAddTable.scala
```diff
@@ -38,6 +41,31 @@ class CAddTable[T: ClassTag, D: ClassTag](val inplace: Boolean = false)(
...
+        return false
+      }
+      d -= 1
+    }
+    return true
+  }
+
+  private def sumAlongDims(tensor: Tensor[D], other: Tensor[D]): Tensor[D] = {
+    val size = tensor.size()
+    var target: Tensor[D] = other
```
wzhongyuan: why define `target` ?  you can use `other` directly

wzhongyuan:
Split PR into several independent ones ?

zhangxiaoli73:
Have splited this PR into several independent ones

PR #2814, opened 2019-05-14
Use one AllReduceParameter for multi-optim method  training

Menooker:

For multiple layers in DirtriOptimizer, use one shared AllReduceParameter instead of different AllReduceParameters for each layer. Use a map parameterSplit to hold the mapping from layer name to (startIdx, length) of the layer. The optimizer will apply its gradient partition on its weight partition for each layer and if the layer's parameterSplit is out of the partition, the optimizer will truncate or skip the gradient to find an appropriate tensor slice for local gradient partition and weight partition.

## How was this patch tested?

UT

Preliminary changes for PR https://github.com/intel-analytics/BigDL/pull/2810

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/parameters/AllReduceParameter.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/parameters/AllReduceParameter.scala
@@ -18,8 +18,8 @@ package com.intel.analytics.bigdl.parameters
 import com.intel.analytics.bigdl.tensor.Tensor
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
 import com.intel.analytics.bigdl.utils.Engine
-import java.util.concurrent.atomic.AtomicLong
 import java.util.concurrent._
+import java.util.concurrent.atomic.AtomicLong
 import org.apache.commons.lang.exception.ExceptionUtils
 import org.apache.log4j.Logger
 import org.apache.spark.TaskContext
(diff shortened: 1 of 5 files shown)


jason-dai:
Why is there LarsSGD in this PR?

Menooker:
> Why is there LarsSGD in this PR?

The commits of LarsSGD has already been merged to master. It may be an Github issue, since the commit IDs are identical to the merged commits and Github shows this PR has no conflicts. Please ignore the LARS-related changes.

Menooker:
@jason-dai I have manually rebase the commits. The LarsSGD commits should be gone.

spark/dl/src/main/scala/com/intel/analytics/bigdl/parameters/AllReduceParameter.scala
```diff
@@ -135,6 +131,15 @@ class AllReduceParameter[T: ClassTag](
       .getOrElse(throw new IllegalStateException("Please initialize AllReduceParameter first!"))
   }
 
+  def localPartitionRange: (Int, Int) = {
```
wzhongyuan: private[bigdl] ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/parameters/AllReduceParameter.scala
```diff
@@ -135,6 +131,15 @@ class AllReduceParameter[T: ClassTag](
...
+    val taskSize = size / partitionNum
+    val extraSize = size % partitionNum
+    val partitionId = TaskContext.getPartitionId()
```
wzhongyuan: Are you sure the `partitionId` is the same as the `partitionId` of this object ?  use the global one instead

Menooker: This method should only be used in Rdd.mapPartitions(...). The partitionId here should be the id from the `mapPartitions`

Menooker: @wzhongyuan Changed as required. Found that `partitionId` is already valid when calling localPartitionRange

wzhongyuan:
Please run a Jenkins pass after verify the compatibility

Menooker:
jenkins passed: http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/602/

wzhongyuan:
LGTM

Menooker:
Jenkins passed http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/609/

Menooker:
Jenkins fixed: http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/615/

PR #2815, opened 2019-05-14
Dilation in MKL-DNN Convolution

LeicongLi:
Enable Dilation in MKL-DNN Convolution

## How was this patch tested?
1. Create convolution layers both in BLAS and MKLDNN with dilation of one, which is just the regular convolution, and compare their output and weight
2. Create convolution layers both in BLAS and MKLDNN with dilation of two, and compare their output and weight.

Jenkins Result:
http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/581/

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/SpatialConvolution.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/SpatialConvolution.scala
@@ -25,6 +25,49 @@ import com.intel.analytics.bigdl.tensor.{DenseTensorMath, DnnTensor, Tensor}
 
 import scala.collection.mutable.ArrayBuffer
 
+/*
+ * Applies a 2D convolution over an input image composed of several input planes.
+ * The input tensor in forward(input) is expected to be
+ * a 3D tensor (nInputPlane x height x width).
+ *
+ * nInputPlane The number of expected input planes in the image given into forward()
+ * nOutputPlane: The number of output planes the convolution layer will produce.
+ * kernelW: the kernel width of the convolution
+ * kernelH: The kernel height of the convolution
+ * strideW: Int = 1, The step of the convolution in the width dimension.
+ * strideH: Int = 1, The step of the convolution in the height dimension
+ * padW: Int = 0, The additional zeros added per width to the input planes.
+ * padH: Int = 0, The additional zeros added per height to the input planes.
+ * nGroup: Int = 1, Kernel group number
+ * propagateBack: Boolean = true, propagate gradient back
+ * wRegularizer: Regularizer[Float] = null,
+ * bRegularizer: Regularizer[Float] = null,
+ * initWeight: Tensor[Float] = null,
+ * initBias: Tensor[Float] = null,
+ * initGradWeight: Tensor[Float] = null,
+ * initGradBias: Tensor[Float] = null,
+ * withBias: Boolean = true,
+ * format: DataFormat = DataFormat.NCHW,
+ * dilationW: Int = 1,
+ * dilationH: Int = 1
+ *
+ * When padW and padH are both -1, we use a padding algorithm similar to the "SAME"
+ * padding of tensorflow. That is
+ *
+ * outHeight = Math.ceil(inHeight.toFloat/strideH.toFloat)
+ * outWidth = Math.ceil(inWidth.toFloat/strideW.toFloat)
+ *
+ * padAlongHeight = Math.max(0, (outHeight - 1) * strideH + kernelH - inHeight)
+ * padAlongWidth = Math.max(0, (outWidth - 1) * strideW + kernelW - inWidth)
+ *
+ * padTop = padAlongHeight / 2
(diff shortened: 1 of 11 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/SpatialConvolution.scala
```diff
@@ -597,14 +772,51 @@ object SpatialConvolution {
...
+    dilationH: Int = 1): SpatialConvolution = {
     new SpatialConvolution(nInputPlane, nOutputPlane, kW, kH, dW,
       dH, padW, padH, nGroup, propagateBack, wRegularizer, bRegularizer,
-      initWeight, initBias, initGradWeight, initGradBias, withBias, format)
+      initWeight, initBias, initGradWeight,
+      initGradBias, withBias, format, dilationW, dilationH)
   }
 }
 
 object Scale {
```
wzhongyuan: private ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/ModuleSerializable.scala
```diff
@@ -142,6 +143,7 @@ trait ModuleSerializable extends Loadable with Savable{
         clsMirror.reflect(compnInst)
     }
 
+    // Todo: to be replaced with ReflectionUtils.reflect
```
wzhongyuan: Remove this line as we already use the Util  ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/intermediate/ReflectionUtils.scala
```diff
@@ -42,38 +42,61 @@ private[bigdl] object ReflectionUtils {
...
     val numericIter = numerics.iterator
-    var i = 0
-    constructorFullParams.foreach(map => {
-      map.foreach(param => {
-        val name = param.name.decodedName.toString
-        val ptype = param.typeSignature
-        if (ptype <:< universe.typeOf[ClassTag[_]]||
-          ptype.typeSymbol == universe.typeOf[ClassTag[_]].typeSymbol) {
+
+    val clsMirror = universe.runtimeMirror(target.getClassLoader)
```
wzhongyuan: Move this class into parent package ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/SpatialConvolution.scala
LeicongLi: The **Scale** object does not belong to this PR, it is an existing code. 
Will make it to private and move it to a proper place in a separate PR.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/serializer/ModuleSerializable.scala
LeicongLi: No. This piece of code is duplicated with RefelctionUtils.reflect due to minor differences.
Would like to make a separate PR for the refactor.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/SpatialConvolution.scala
```diff
@@ -116,6 +164,22 @@ class SpatialConvolution(
     this
   }
 
+  // get padding type
+  private val paddingType: PaddingType.Value = getPaddingType()
+
+  /*
+  Parameters for Dilated Convolution
+  In most of deep learning framework,
```
wzhongyuan: frameworks

wzhongyuan: and remove `of` ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/SpatialConvolution.scala
```diff
@@ -576,6 +643,114 @@ class SpatialConvolution(
     needQuantize = value
     this
   }
+
+  /**
+   * Todo:
```
wzhongyuan: Todo should be all in upper case for highlighting ?

LeicongLi: it is not case sensitive
this line is a placeholder, i have a plan to continue finishing what left, and it will be removed soon.

wzhongyuan: OK. it's not case sensitive and normally if there is work left to finish, Upper case is the convention

LeicongLi: great, if there is a convention, then lets follow it.

PR #2817, opened 2019-05-15
Optimize backward graph generation and CAddTable

zhangxiaoli73:

. Optimize backward graph generation
. Enhance CAddTable to support tensors with different size

## How was this patch tested?

unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/CAddTable.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/CAddTable.scala
@@ -16,11 +16,14 @@
 
 package com.intel.analytics.bigdl.nn
 
+import javax.print.attribute.standard.MediaSize.Other
+
 import com.intel.analytics.bigdl.nn.abstractnn.AbstractModule
 import com.intel.analytics.bigdl.tensor.Tensor
 import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
-import com.intel.analytics.bigdl.utils.Table
+import com.intel.analytics.bigdl.utils.{T, Table}
 import com.intel.analytics.bigdl.utils.serializer.{DeserializeContext, ModuleSerializable}
+import org.apache.hadoop.yarn.webapp.hamlet.HamletSpec.VAR
 
 import scala.reflect._
 
@@ -37,6 +40,31 @@ class CAddTable[T: ClassTag, D: ClassTag](val inplace: Boolean = false)(
   extends AbstractModule[Table, Tensor[D], T] with MklInt8Convertible {
 
   output = Tensor[D]()
+
+  private def canFastBroadcast(tensor: Tensor[D], other: Tensor[D]): Boolean = {
+    if (tensor.nDimension != other.nDimension()) {
+      return false
+    }
+    var d = other.nDimension()
+    while(d > 0) {
+      if (tensor.size(d) != 1 && tensor.size(d) != other.size(d)) {
+        return false
+      }
+      d -= 1
+    }
+    return true
+  }
+
+  private def sumAlongDims(tensor: Tensor[D], other: Tensor[D]): Tensor[D] = {
+    val size = tensor.size()
+    var target: Tensor[D] = other
+    var i = 0
+    while (i < size.length) {
+      if (size(i) == 1) target = target.sum(i + 1)
+      i += 1
+    }
+    target
+  }
 
   override def updateOutput(input: Table): Tensor[D] = {
     var scalar = ev2.zero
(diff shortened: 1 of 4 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/CAddTable.scala
```diff
@@ -38,6 +41,31 @@ class CAddTable[T: ClassTag, D: ClassTag](val inplace: Boolean = false)(
 
   output = Tensor[D]()
 
+  private def canFastBroadcast(tensor: Tensor[D], other: Tensor[D]): Boolean = {
```
wzhongyuan: what does `canFastBroadcast` mean ?

zhangxiaoli73: We need to handle two tensors with different size. If they have same dims, and size(dim) is same or one of them is 1, that means tensors can be added together.

wzhongyuan: ok, but why we can `canFastBroadcast` ? the naming is confusing

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/CAddTable.scala
```diff
@@ -38,6 +41,31 @@ class CAddTable[T: ClassTag, D: ClassTag](val inplace: Boolean = false)(
...
+        return false
+      }
+      d -= 1
+    }
+    return true
+  }
+
+  private def sumAlongDims(tensor: Tensor[D], other: Tensor[D]): Tensor[D] = {
+    val size = tensor.size()
+    var target: Tensor[D] = other
```
wzhongyuan: why not just use `other` ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/CAddTable.scala
zhangxiaoli73: This name is refered to same operation in tensor add. 
How about `canBroadcast` or `canExpand` ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/CAddTable.scala
zhangxiaoli73: result of tensor sum is another tensor, not itself. Just now I found tensor sum will create new tensors, so add some code refactor.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/CAddTable.scala
wzhongyuan: The method here is we want to add along all dims with 1 and then to see it can be the same size as the reference one right ? need to think how to name it...

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/CAddTable.scala
```diff
@@ -38,6 +41,42 @@ class CAddTable[T: ClassTag, D: ClassTag](val inplace: Boolean = false)(
 
   output = Tensor[D]()
 
+  @transient
+  private var bufferSumInput: Tensor[D] = null
+  @transient
+  private var bufferSumOutput: Tensor[D] = null
+
+  private def canBroadcast(tensor: Tensor[D], other: Tensor[D]): Boolean = {
```
wzhongyuan: from the implementation,  we don't really use `tensor` we just use `size()` instead, probably use `size: Array[Int]` as first parameter and then make it a more meaningful name

wzhongyuan: How about ` private def dimSumResizable(tensor: Tensor[D],  size: Array[Int]): Boolean = {...}`,  or a better one

zhangxiaoli73: OK, how about change to 'dimResizable'?

zhangxiaoli73:
Jenkins pass:
http://mlt-bddl01.sh.intel.com:8080/view/PR/job/BigDL-PRVN-lite/587/

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/CAddTable.scala
wzhongyuan: I think the intention of this method is to sum along specific `dim` and then check if the summed tensor can be resized (identical) to referred tensor ?   that's why I said `dimSum`, we can think a better one

zhangxiaoli73: If two tensors have same dims but different size, size(dim) for small tensor should be 1 or same with another. Then in backward, sum gradOutput along dim that size(dim) is 1.

So firstly we need to check 'dimResizable' (or other name),  and then do `sumAlongDims`.

zhangxiaoli73: change to `expandWithDims`

zhangxiaoli73:
jenkins pass:
http://mlt-bddl01.sh.intel.com:8080/view/PR/job/BigDL-PRVN-lite/591/

Issue #2811, opened 2019-05-13
ResNet50 inference RandomCropper can be replaced by CenterCrop


hkvision:
The accuracy will remain unchanged and the speed will be much faster.
Please confirm this :).

wzhongyuan:
Thanks, will take a look

hkvision:
Also, RandomResize can be replaced by Resize (mainly for avoid ambiguity).

wzhongyuan:
CenterCrop has the same result as the RandomCropper for validation while RandomResize does not with Resize

PR #2818, opened 2019-05-15
[Enhancement] - Replace randomcropper with centercrop for better performance

wzhongyuan:
Both have the same processing result and later one has better performance

## How was this patch tested?

UT and load model validation

fixed https://github.com/intel-analytics/BigDL/issues/2811


wzhongyuan:
Jenkins passed http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/584/

PR #2819, opened 2019-05-15
add layer norm and expand size layers

zhangxiaoli73:

. Implement Layer normalization
. Implement ExpandSize and TableOperationExpand that handling two tensors with different size
. Add corresponding python api

## How was this patch tested?

unit tests

### DIFF
--- a/pyspark/bigdl/nn/layer.py
+++ b/pyspark/bigdl/nn/layer.py
@@ -929,6 +929,39 @@ class FeedForwardNetwork(Layer):
     def __init__(self, hidden_size, filter_size, relu_dropout, bigdl_type="float"):
         super(FeedForwardNetwork, self).__init__(None, bigdl_type,
                                         hidden_size, filter_size, relu_dropout)
+class LayerNormalization(Layer):
+    '''
+    Applies layer normalization.
+
+    >>> norm = LayerNormalization(8)
+    creating: createLayerNormalization
+    '''
+
+    def __init__(self, hidden_size, bigdl_type="float"):
+        super(LayerNormalization, self).__init__(None, bigdl_type, hidden_size)
+
+class createTableOperationExpand(Layer):
+    '''
+    When two tensors have different size, firstly expand small size tensor to large size tensor,
+    and then do table operation.
+
+    >>> norm = TableOperationExpand(CMulTable)
+    creating: createTableOperationExpand
+    '''
+
+    def __init__(self, operation_layer, expand_pos=2, bigdl_type="float"):
+        super(TableOperationExpand, self).__init__(None, bigdl_type, operation_layer, expand_pos)
+
+class createExpandSize(Layer):
+    '''
+    Expand tensor to configured size
+
+    >>> expand = ExpandSize([2, 3, 4])
+    creating: createExpandSize
+    '''
+
+    def __init__(self, sizes, bigdl_type="float"):
+        super(ExpandSize, self).__init__(None, bigdl_type, sizes)
 
 class Linear(Layer):
 
--- a/pyspark/bigdl/nn/layer.py
+++ b/pyspark/bigdl/nn/layer.py
@@ -940,19 +940,20 @@ class LayerNormalization(Layer):
     def __init__(self, hidden_size, bigdl_type="float"):
         super(LayerNormalization, self).__init__(None, bigdl_type, hidden_size)
 
-class createTableOperationExpand(Layer):
+class TableOperation(Layer):
     '''
     When two tensors have different size, firstly expand small size tensor to large size tensor,
     and then do table operation.
 
-    >>> norm = TableOperationExpand(CMulTable)
-    creating: createTableOperationExpand
+    >>> norm = TableOperation(CMulTable())
+    creating: createCMulTable
+    creating: createTableOperation
     '''
 
-    def __init__(self, operation_layer, expand_pos=2, bigdl_type="float"):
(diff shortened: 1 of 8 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ExpandSize.scala
```diff
@@ -0,0 +1,102 @@
...
+
+import scala.collection.mutable.ArrayBuffer
+import scala.reflect.ClassTag
+
+/**
+ * Expand tensor to configured size
+ * @param tgtSizes target tensor sizes, dim whose size is -1 will be ignored
+ * @tparam T Numeric type of parameter(e.g. weight, bias). Only support float/double now.
+ */
+class ExpandSize[T: ClassTag](tgtSizes: Array[Int])
```
wzhongyuan: preferable `targetSizes` ?

zhangxiaoli73: ok, changed

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ExpandSize.scala
```diff
@@ -0,0 +1,102 @@
...
+    }
+    gradInput = _gradOutput
+    gradInput
+  }
+
+  override def toString: String = s"ExpandSize"
+}
+
+object ExpandSize {
+  def apply[@specialized(Float, Double) T: ClassTag](tgtSizes: Array[Int])
```
wzhongyuan: targetSize, please keep consistent with class definition

zhangxiaoli73: ok, changed

zhangxiaoli73:
jenkins pass: http://mlt-bddl01.sh.intel.com:8080/view/PR/job/BigDL-PRVN-lite/593/

pyspark/bigdl/nn/layer.py
```diff
@@ -929,6 +929,40 @@ class FeedForwardNetwork(Layer):
...
+    Applies layer normalization.
+
+    >>> norm = LayerNormalization(8)
+    creating: createLayerNormalization
+    '''
+
+    def __init__(self, hidden_size, bigdl_type="float"):
+        super(LayerNormalization, self).__init__(None, bigdl_type, hidden_size)
+
+class TableOperationExpand(Layer):
```
wzhongyuan: `Expand ` is a verb

wzhongyuan: Probably just remove it

pyspark/bigdl/nn/layer.py
```diff
@@ -929,6 +929,40 @@ class FeedForwardNetwork(Layer):
...
+    '''
+    When two tensors have different size, firstly expand small size tensor to large size tensor,
+    and then do table operation.
+
+    >>> norm = TableOperationExpand(CMulTable())
+    creating: createCMulTable
+    creating: createTableOperationExpand
+    '''
+
+    def __init__(self, operation_layer, expand_pos=2, bigdl_type="float"):
```
wzhongyuan: why we should specify the expand_pos ? can we infer from input ? it's determined by the input stead of layer itself right ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/LayerNormalization.scala
```diff
@@ -0,0 +1,172 @@
...
+}
+
+/**
+ * Implement x * weight vector + bias vector
+ * @param hiddenSize
+ * @param ev$1
+ * @param ev
+ * @tparam T The numeric type in this module parameters
+ */
+private[nn] class LayerLinear[T: ClassTag](val hiddenSize: Int)
```
wzhongyuan: What's the difference with `Linear` ?

pyspark/bigdl/nn/layer.py
zhangxiaoli73: yeah, good idea. Refine code and infer position from input

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/LayerNormalization.scala
zhangxiaoli73: This layer implements `input vector * weight vector + bias vector`. It should be an element-wise multiply operation.
If input is a tensor with two, then we should select tensors along its's first dim to do multiply.

So you see it is different with nn linear. Maybe name is not clear. How about `VectorLinear` or any other?

pyspark/bigdl/nn/layer.py
zhangxiaoli73: ok, changed

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/LayerNormalization.scala
wzhongyuan: `VectorProduct` ?

zhangxiaoli73: ok, changed

zhangxiaoli73:
jenkins pass: http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/597/

PR #2822, opened 2019-05-21
fix: calculate arbitrary mask of scales

i8run:

1. Enable arbitrary mask when do scale generating.  For example, A `[4, 3, 2, 2]` shape of Tensor,
the mask 3 will generate scales based on first two dimension, which will return 3 * 4 = 12 values.
2. Move the `ScaleCalculatorSpec` to `MklInt8ConvertibleSpec`.

## How was this patch tested?

Jenkins and unit tests.


wzhongyuan:
> Enable to calculate arbitrary mask of scales, such as 3 (11

Not clear

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/MklInt8Convertible.scala
```diff
@@ -165,24 +165,7 @@ trait MklInt8Convertible {
...
-        bitMask(binStrLen - i) = binStrMask(binStrLen - i).asDigit
-        if (bitMask(binStrLen - i) == 1) {
-          val dimSize = tensor.size(i)
-          for (j <- 1 to dimSize) {
-            scalesBuffer.append(tensor.select(i, j).clone().abs().max())
-          }
-        }
-      }
-      scalesBuffer.toArray[Float]
+      Utils.calc(tensor, mask)
```
LeicongLi: calc is too general.
please give a meaningful name.
calcScales should be fine.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Utils.scala
```diff
@@ -570,4 +570,57 @@ object Utils {
     times: Array[(AbstractModule[_ <: Activity, _ <: Activity, T], Long, Long)]): (Long, Long) = {
       times.map(t => (t._2, t._3)).reduce((a, b) => (a._1 + b._1, a._2 + b._2))
   }
+
+  private[nn] def calc(tensor: Tensor[Float], mask: Int): Array[Float] = {
```
LeicongLi: I would suggest to add a comment on top of method header.
Whats the representation of the mask, decimal? binary?; if binary, its Big endian or little endian?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Utils.scala
```diff
@@ -570,4 +570,57 @@ object Utils {
...
+    // because the shape of tensor is reverse of mask, for example,
+    // if the mask is 11 (bit representation) and the shape is [4, 3, 2, 2].
+    // the corresponding dimension is [4, 3] which will return 12 values.
+    calc(tensor, mask.toBinaryString.reverse, 0).toArray
+  }
+
+  /**
+   * recursively calculate tensor max values based on the mask and index.
+   *
+   * for a [4, 3, 2, 2] tensor and 11 mask, it will generate 4*3=12 max values.
```
LeicongLi: please refine the explaination, for example, what is 11 mask, it would confuse the readers.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Utils.scala
```diff
@@ -570,4 +570,57 @@ object Utils {
...
+
+    // terminator 1
+    if (index < 0 || index >= mask.length) return Seq()
+
+    // terminator 2, at the end of mask whose bit has been set
+    if (mask(index).asDigit == 1 && index == mask.length - 1) {
+      return (1 to tensor.size(index + 1)).flatMap { idx =>
+        Seq(tensor.narrow(index + 1, idx, 1).clone().abs.max())
+      }
+    }
```
LeicongLi: I would try to find a way to get rid of this code.

LeicongLi: `def calcScales(tensor: Tensor[Float], mask: Int): Seq[Float] = {

    // inner helper function
    def calcScalesHelper(tensor: Tensor[Float], maskStr: String,
                         result: mutable.ListBuffer[Float], index: Int): Unit = {
      if (index < maskStr.length) {
        if (maskStr(index).asDigit == 1) { // mask bit is ON at this dimension
          (1 to tensor.size(index + 1)).foreach(
            i => { // split the tensor based on its size
              calcScalesHelper(tensor.narrow(index + 1, i, 1), maskStr, result, index + 1)
            }
          )
        } else {
          calcScalesHelper(tensor, maskStr, result, index + 1)
        }

      } else { // finished splitting tensor based on its mask bit, aggregate and append the result
        result.append(tensor.clone().abs.max())
      }

    }
    
    val result = mutable.ListBuffer[Float]()
    val maskStr = mask.toBinaryString.reverse

    calcScalesHelper(tensor, maskStr, result, 0)

    return result.toList
  }`

LeicongLi: would you mind to take a look?

i8run: Good. Idea is more clear. I will use it directly. Thanks.

PR #2823, opened 2019-05-22
feat: enable global average pooling

i8run:
This patch enable the global pooling. As the implementation in `SpatialAveragePooling`, use the same parameter style, which means use `var` to handle `kH` and `kW` in the argument list.

## How was this patch tested?
Jenkins and unit tests.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/AvgPooling.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/AvgPooling.scala
@@ -22,12 +22,13 @@ import com.intel.analytics.bigdl.nn.mkldnn.Phase.InferencePhase
 import com.intel.analytics.bigdl.tensor.Tensor
 
 class AvgPooling(
-  kW: Int,
-  kH: Int,
+  var kW: Int,
+  var kH: Int,
   dW: Int = 1,
   dH: Int = 1,
   padW: Int = 0,
-  padH: Int = 0
+  padH: Int = 0,
+  globalPooling: Boolean = false
 ) extends MklDnnLayer {
   @transient private var paddingTL: Array[Int] = _
   @transient private var paddingBR: Array[Int] = _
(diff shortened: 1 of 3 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/AvgPooling.scala
```diff
@@ -22,12 +22,13 @@ import com.intel.analytics.bigdl.nn.mkldnn.Phase.InferencePhase
 import com.intel.analytics.bigdl.tensor.Tensor
 
 class AvgPooling(
-  kW: Int,
-  kH: Int,
+  var kW: Int,
+  var kH: Int,
```
wzhongyuan: Do we need support in nn/AvgPooling as well?

spark/dl/src/test/scala/com/intel/analytics/bigdl/nn/mkldnn/AvgPoolingSpec.scala
```diff
@@ -236,4 +236,33 @@ class AvgPoolingSpec extends BigDLSpecHelper {
 
     Equivalent.nearequals(seq.output.toTensor, seq2.output.toTensor, 1e-2) should be (true)
   }
+
+  "global average pooling" should "work correctly" in {
```
zhangxiaoli73: Add unit tests to check:
1. whether dnn global average pooling equals blas.
2. whether converting global pooling blas to dnn can work correctly

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/AvgPooling.scala
i8run: `nn/AvgPooling` has enabled them.

spark/dl/src/test/scala/com/intel/analytics/bigdl/nn/mkldnn/AvgPoolingSpec.scala
i8run: OK.

i8run:
jenkins passed. http://mlt-bddl01.sh.intel.com:8080/view/PR/job/BigDL-PRVN-lite/599/parameters/

PR #2824, opened 2019-05-22
fix: require consistent between shape and layout of mkldnn

i8run:

The exception information is not clear when the reorder can't work. For example,
a graph with a maxpooling, which will output 4-D tensor. If we convert the 4-D tensor with
a `Memory.Format.nc` format. It will be segment fault.

## How was this patch tested?
Jenkins.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/MemoryData.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/MemoryData.scala
@@ -104,7 +104,8 @@ sealed trait MemoryData extends Serializable {
   private def checkConsistency(shape: Array[Int], layout: Int): Unit = {
     val isConsistency = Memory.Format.any == layout || (shape.length match {
       case 1 => layout == Memory.Format.x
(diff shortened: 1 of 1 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/MemoryData.scala
```diff
@@ -99,6 +100,17 @@ sealed trait MemoryData extends Serializable {
       case _ => throw new UnsupportedOperationException(s"unsupported data type")
     }
   }
+
+  private def checkConsistency(shape: Array[Int], layout: Int): Unit = {
+    // TODO current only handle the 4-D and 2-D shape
```
wzhongyuan: What will be `TODO`

i8run: I want to check the shape is consistent with layout. Currently, it only implements the invalidation check.

i8run:
Jenkins has passed. http://mlt-bddl01.sh.intel.com:8080/view/PR/job/BigDL-PRVN-lite/663/

PR #2827, opened 2019-05-24
fix: support int8 of jointable

i8run:

JoinTable need to specially care. If the previous nodes of `JoinTable` are convolutions, we should make the two conv's output scales to be the same, so JoinTable can do the computing on the same scale factor.

## How was this patch tested?

Unit tests.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnGraph.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnGraph.scala
@@ -327,8 +327,9 @@ class DnnGraph(
 
   private def getInputMemoryData(node: ModuleNode[Float], memoryData: Array[MemoryData])
     : Array[MemoryData] = {
-    if (inputs.length == 1) {
-      require(inputs(0).eq(node), "input node is not in the input list")
+    // the model may contains two inputs and all of them is Input.
+    if (inputs.length == 1 || memoryData.isEmpty) {
+      require(inputs.contains(node), "input node is not in the input list")
       memoryData
     } else {
       val i = inputs.indexOf(node)
@@ -392,13 +393,14 @@ class DnnGraph(
    */
   private def fusion(): Unit = {
     if (!this.train) {
-      for (j <- 0 to 2) {
+      for (j <- 0 to 3) {
         var i = forwardExecution.length - 1
         while (i >= 0) {
           if (j == 0) Fusion.fuseModule(forwardExecution(i))
           // we should do this before sum fusion, because it will change the structure of graph
           if (j == 1) Fusion.setNegativeInputOfConv(forwardExecution(i))
           if (j == 2) Fusion.fuseCAdd(forwardExecution(i))
+          if (j == 3) Fusion.setJoinTableSame(forwardExecution(i))
           i -= 1
         }
       }
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnGraph.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnGraph.scala
@@ -327,9 +327,9 @@ class DnnGraph(
 
   private def getInputMemoryData(node: ModuleNode[Float], memoryData: Array[MemoryData])
     : Array[MemoryData] = {
-    // the model may contains two inputs and all of them is Input.
+    // the model may contain two inputs and all of them is Input.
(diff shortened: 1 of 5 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnGraph.scala
```diff
@@ -327,8 +327,9 @@ class DnnGraph(
 
   private def getInputMemoryData(node: ModuleNode[Float], memoryData: Array[MemoryData])
     : Array[MemoryData] = {
-    if (inputs.length == 1) {
-      require(inputs(0).eq(node), "input node is not in the input list")
+    // the model may contains two inputs and all of them is Input.
```
wzhongyuan: both are input

wzhongyuan: contain

wzhongyuan: both are inputs

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnGraph.scala
```diff
@@ -327,8 +327,9 @@ class DnnGraph(
...
+    if (inputs.length == 1 || memoryData.isEmpty) {
+      require(inputs.contains(node), "input node is not in the input list")
```
wzhongyuan: input node must be in the input list (to keep consistent with `contains` )

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Fusion.scala
```diff
@@ -242,4 +231,46 @@ private[mkldnn] object Fusion {
       node.element.asInstanceOf[SpatialConvolution].negativeInput = false
     }
   }
+
+  def setJoinTableSame(node: Node[AbstractModule[Activity, Activity, Float]]): Unit = {
```
wzhongyuan: what does `same` mean ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Fusion.scala
```diff
@@ -242,4 +231,46 @@ private[mkldnn] object Fusion {
...
+    if (!fuse || !node.element.isInstanceOf[JoinTable]) return
+
+    val preConvs = node.prevNodes.flatMap(x => findAllNonIdentityPrevs(x))
+      .filter(_.element.isInstanceOf[SpatialConvolution])
+      .map(_.element.asInstanceOf[SpatialConvolution])
+
+    var needQuantize = preConvs.exists(_.needQuantize)
```
wzhongyuan: needQuantization ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Fusion.scala
```diff
@@ -242,4 +231,46 @@ private[mkldnn] object Fusion {
...
+    if (!needQuantize) return
+
+    val masks = preConvs.map(_.getOutputDimMask()).toSet
+    require(masks.size == 1, s"all the convolution should be the same mask")
```
wzhongyuan: all preceding convolution layers must have the same mask ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Fusion.scala
```diff
@@ -242,4 +231,46 @@ private[mkldnn] object Fusion {
...
+    }
+
+    preConvs.foreach { conv =>
+      conv.setOutputScales(scales)
+    }
+  }
+
+  private def findAllNonIdentityPrevs(node: Node[AbstractModule[Activity, Activity, Float]])
+  : Seq[Node[AbstractModule[Activity, Activity, Float]]] = {
+    // TODO currently, it will only skip the Identity, MaxPooling, AvgPooling, JoinTable
```
wzhongyuan: TODO what ?

i8run: We need to check whether the input of current node is not negative. So here we skip the previous nodes such as Identity, MaxPooling, AvgPooling And JoinTable. It can work but not an elegant way to implement.

PR #2832, opened 2019-06-04
Add parameter processor for LARS

Menooker:

Add a ParameterProcessor for LARS to calculate the global layer-wise norm

## How was this patch tested?

UT

Replace the PR https://github.com/intel-analytics/BigDL/pull/2810


spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/LarsSGD.scala
```diff
@@ -57,6 +58,11 @@ class LarsSGD[T: ClassTag](
      learningRateSchedule = _learningRateSchedule) {
   @transient
   private var buffer: Tensor[T] = null
+  @transient
+  private[bigdl] var calculatedTrust: Option[T] = None
+
+  require(trust>0.0 && trust<= 1.0, "trust for LARS should be greater than 0 and less than 1")
```
wzhongyuan: and print the actual trust value as well ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/LarsSGD.scala
```diff
@@ -103,6 +95,36 @@ class LarsSGD[T: ClassTag](
     (parameter, Array(fx))
   }
 
+  private[bigdl] def setGradientScale[T](scale: Double): Unit = {
+    calculatedTrust = Some(ev.fromType[Double](scale))
+  }
+
+  private[bigdl] def getGradientScale(dfdx: Tensor[T], parameter: Tensor[T]): T = {
+    val raw_scale_value = if (calculatedTrust.isDefined) {
```
wzhongyuan: use java convention `rawScaleValue`

spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/LarsSGD.scala
```diff
@@ -175,6 +197,33 @@ object LarsSGD {
...
 
+  /**
+   * Check if there is LarsSGD in optimMethods If so, return the weight decay of the first found
+   * LarsSGD. Else, return None
+   * @param optimMethods
+   * @tparam T
+   * @return The weight decay of the first found LarsSGD in the optimMethods.
+   *         Or None if there is not one
+   */
+  def containsLarsSGD[T](optimMethods: Map[String, OptimMethod[T]]): Option[Double] = {
```
wzhongyuan: How about changing to below
``` scala
optimMethods.foreach({ (name, optim) =>
      optim match {
        case d: LarsSGD[T] =>
               return Some(d.weightDecay)
      }
    })
   return None
```

Menooker: The `return` in the lambda has no effect on the outer function...

Menooker: A new commit changes this line using `find` method followed by a `map` operation, which should be more concise. For your other comments, I have modified the code as requested.

wzhongyuan:
Please run a Jenkins pass

Menooker:
Jenkins pass at http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/646/

Issue #2838, opened 2019-06-13
MKL-DNN will modify the model when doing inference


Menooker:
When doing inference using mkl-dnn engine, the weights of the model may be modified by the engine. It seems that the `forward` operation will call `initFwdPrimitives` that will do fusion operations. And the fusion operation will do in-place modifications on the model weights. However,  a model may call "forward" for multiple times, which will corrupt the model.

qiuxin2012:
If the model is corrupted, will the result be Nan?

Menooker:
@qiuxin2012 No, the forward output will be finite, but not correct. In my case, the corrupted model (Resnet-50, 2 classes) will always get "positive" result, but the scores for the both classes are finite. The problem only appears when using a IRGraph model.

qiuxin2012:
@Menooker In my case...I got NaN...
As cherry told me a way to work around this issue. broadcast the model, and call `.value()` to cache the model, then calling the cached model's `forward` multi times won't invoke this issue. Is it right?

Menooker:
@qiuxin2012 In my application, this will corrupt the model. That's how I found this issue (load model from file => toIRGraph() => broadcast the model => MyPredictor.predictImage(localCachedModel) for multiple times => corrupted model)

Menooker:
@qiuxin2012 @zhangxiaoli73 Hi, I have composed a simple program to reproduce the problem.

The program uses Resnet model and prints the sum of the weights to check if the model is modified.

Note: I found that if I remove `.evaluate()` in the line ` val bcast = ModelBroadcast[Float](....` the model will be unchanged.

```scala
import com.intel.analytics.bigdl.mkl.Memory
import com.intel.analytics.bigdl.models.utils.ModelBroadcast
import com.intel.analytics.bigdl.nn.abstractnn.{AbstractModule, Activity}
import com.intel.analytics.bigdl.nn.{Linear, Module, ReLU, Sequential, StaticGraph}
import com.intel.analytics.bigdl.tensor.Tensor
import com.intel.analytics.bigdl.utils.{Engine, LoggerFilter}
import org.apache.spark.SparkContext
import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric

object tester {
  LoggerFilter.redirectSparkInfoLogs()
  def main(argv: Array[String]): Unit ={
    val conf = Engine.createSparkConf().setAppName("Predict ResNet")
      .set("spark.rpc.message.maxSize", "200")
[...]

PR #2839, opened 2019-06-17
Fix weights changed for model fusion

zhangxiaoli73:
    
We will change model structure and weights when doing conv and bn fusion.
 In order to not influence broadcast model and weights, we have to set new storage to weight and bias.

## How was this patch tested?

unit tests

fixed https://github.com/intel-analytics/BigDL/issues/2838


zhangxiaoli73:
jenkins pass: http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/639/

PR #2841, opened 2019-06-24
fix acc bug & init dnn thread

zhangxiaoli73:

1. fix bug in SpatialConvolution initGradWPrimitives
2. Init dnn thread for java local

## How was this patch tested?

unit tests


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -237,6 +237,8 @@ object Engine {
   // For `context` in ThreadPool, it is the called thread when poolSize is 1.
   // So many usages of that thread, we will not change it for now.
   val dnnComputing: ThreadPool = new ThreadPool(1)
+  // We need to init dnn thread in case that users directly call model operation in java local
+  initDnnThread()
```
i8run: I think we should push the =init= at the beginning or the end. It will be much clearer.

zhangxiaoli73:
jenkins pass: http://mlt-bddl01.sh.intel.com:8080/view/PR/job/BigDL-PRVN-lite/667/

PR #2843, opened 2019-06-26
fix: invokeAndWait2 should throw the exception in the tasks

i8run:

The `invokeAndWait2` will not throw exceptions in the tasks to caller thread.

## How was this patch tested?
test cases and Jenkins.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
@@ -23,7 +23,7 @@ import java.util.concurrent.atomic.AtomicBoolean
 import org.apache.log4j.Logger
 import org.apache.spark._
 import com.intel.analytics.bigdl.mkl.MKL
-import com.intel.analytics.bigdl.mkl.hardware.CpuInfo
+import com.intel.analytics.bigdl.mkl.hardware.{Affinity, CpuInfo}
 import org.apache.spark.utils.SparkUtils
 import py4j.GatewayServer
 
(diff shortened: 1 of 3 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -556,7 +556,24 @@ object Engine {
...
+    val affinityCores = Affinity.getAffinity
+    val physicalCoreNum = CpuInfo.getPhysicalProcessorCount
+    val affinityCoreNum = affinityCores.length
+
+    // 1. this library in docker/cgroup env, which sets cpu affinity fist. so we can't use
+    //    resources exceeding limits.
+    // 2. this library is in a hyper thread envs, which enables hyper threading of cpu. so
+    //    we should set the mkl num threads to physical core number for performance
+
+    val default = if (affinityCores.min > 0 && affinityCores.max >= physicalCoreNumber) {
```
zhangxiaoli73: When it happens `affinityCores.max >= physicalCoreNumber`

i8run: hyper threading is enabled.

zhangxiaoli73: ok, why not use physicalCoreNumber to bind affinity in this case.

i8run: this case can't use. for example there're 8 logical cores and 4 physical cores. If the affinity settings is 2-6 (starting from 0), which means this environment only can use core 2, 3, 4. So I can't set the `physicalCoreNumber`.

zhangxiaoli73: ok， I see

PR #2847, opened 2019-07-03
support ntc in dnn layer

zhangxiaoli73:

support ntc in dnn layer

## How was this patch tested?

unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnGraph.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnGraph.scala
@@ -55,7 +55,7 @@ class DnnGraph(
       val out = output.toTensor[Float]
       // for grey image, input should be 3 dims and the first dim should be batch size
       // for non grey image, input should be 4 dims and the first dim should be batch size
-      require(in.nDimension() == 4 || in.nDimension() == 3,
+      require(in.nDimension() == 4 || in.nDimension() == 3 || in.nDimension() == 2,
         s"only support input with 4 dimension or 3 dimension, but get ${in.nDimension()}")
       if (in.size(1) != out.size(1)) out.narrow(1, 1, in.size(1)) else output
     } else output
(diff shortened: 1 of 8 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnGraph.scala
```diff
@@ -55,7 +55,7 @@ class DnnGraph(
       val out = output.toTensor[Float]
       // for grey image, input should be 3 dims and the first dim should be batch size
       // for non grey image, input should be 4 dims and the first dim should be batch size
-      require(in.nDimension() == 4 || in.nDimension() == 3,
+      require(in.nDimension() == 4 || in.nDimension() == 3 || in.nDimension() == 2,
```
i8run: Please add comment about the condition of `in.nDimension() == 2`

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/intermediate/IRGraph.scala
```diff
@@ -141,7 +141,10 @@ private[bigdl] class IRGraph[T: ClassTag](
       if (input.isInstanceOf[Tensor[T]]) {
         // todo: handle for 3 dimensions, expand 3 dims to 4 dims
         val size = input.toTensor[T].size()
-        val sizeNew = if (size.length == 3)  Array(size(0), 1, size(1), size(2)) else size
+        val sizeNew = if (size.length == 3 && inputFormats(0) != Memory.Format.ntc
+          && inputFormats(0) != Memory.Format.tnc) {
+          Array(size(0), 1, size(1), size(2))
+        } else size
```
i8run: Why we should create `sizeNew` ? Could it be handled in `Input` layer/op ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnGraph.scala
zhangxiaoli73: ok, changed

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/intermediate/IRGraph.scala
zhangxiaoli73: input layer needs input shape and format. Dim 3 has no corresponding format, and its actual channel is 1, so here we add one dim and create sizeNew

i8run: OK. Got it.

zhangxiaoli73:
jenkins pass: http://mlt-bddl01.sh.intel.com:8080/view/PR/job/BigDL-PRVN-lite/676/

PR #2853, opened 2019-07-09
fix: use treeset to calc topk to upgrade the performance of DetectionOutputSSD

i8run:

This patch use `TreeSet` to calculate topk, which will have 3x~4x performance speed up for `DetectionOutputSSD`.

## How was this patch tested?

Add 2 tests and Jenkins.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
@@ -1722,11 +1722,6 @@ private[tensor] class DenseTensor[@specialized T: ClassTag](
 
     @inline
     def compare(a: T, b: T): Boolean = ev.isGreater(b, a)  ^ !increase
-    /* if (increase) {
-      ev.isGreater(b, a)
-    } else {
-      ev.isGreater(a, b)
-    } */
 
     DenseTensorDimApply.dimApply3[T](this, resultTensor, indicesTensor, selectDim,
       (tdata, toffset, tstride, tsize, vdata, voffset, vstride, vsize, idata,
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
@@ -1721,38 +1721,36 @@ private[tensor] class DenseTensor[@specialized T: ClassTag](
     indicesTensor.resize(topKSize)
 
     @inline
-    def compare(a: T, b: T): Boolean = if (increase) {
+    def compare(a: T, b: T): Boolean = ev.isGreater(b, a)  ^ !increase
+    /* if (increase) {
       ev.isGreater(b, a)
     } else {
       ev.isGreater(a, b)
-    }
-
-    class LocalComparator(increase: Boolean) extends Comparator[(T, Int)] {
-      override def compare(o1: (T, Int), o2: (T, Int)): Int = {
-        val ret = if (ev.isGreater(o1._1, o2._1)) {
-          1
-        } else if (ev.isGreaterEq(o1._1, o2._1)) {
-          if (o1._2 == o2._2) {
-            0
-          } else {
-            1
-          }
-        } else {
-          -1
-        }
-
-        if (increase) {
-          -ret
-        } else {
-          ret
-        }
-      }
-    }
+    } */
 
     DenseTensorDimApply.dimApply3[T](this, resultTensor, indicesTensor, selectDim,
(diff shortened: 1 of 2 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
```diff
@@ -1717,24 +1720,64 @@ private[tensor] class DenseTensor[@specialized T: ClassTag](
...
+    @inline
+    def compare(a: T, b: T): Boolean = if (increase) {
+      ev.isGreater(b, a)
+    } else {
+      ev.isGreater(a, b)
+    }
+
+    class LocalComparator(increase: Boolean) extends Comparator[(T, Int)] {
+      override def compare(o1: (T, Int), o2: (T, Int)): Int = {
+        val ret = if (ev.isGreater(o1._1, o2._1)) {
```
wzhongyuan: why do we need this `if` as it's covered by below `if`

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
```diff
@@ -1717,24 +1720,64 @@ private[tensor] class DenseTensor[@specialized T: ClassTag](
...
     indicesTensor.resize(topKSize)
 
+    @inline
+    def compare(a: T, b: T): Boolean = if (increase) {
+      ev.isGreater(b, a)
+    } else {
+      ev.isGreater(a, b)
+    }
+
+    class LocalComparator(increase: Boolean) extends Comparator[(T, Int)] {
```
wzhongyuan: `increase` to `asce` ?

wzhongyuan: why name it `Local` ?

wzhongyuan: it's better not define a separate class, just define an anonymous class when declaring the TreeSet

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
i8run: Because there's no `equal` in ev. The next `isGreaterEq` handles the `equal`.

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
i8run: Because it's used only in this method.

i8run: OK.

i8run: Good idea.

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
```diff
@@ -1717,24 +1720,64 @@ private[tensor] class DenseTensor[@specialized T: ClassTag](
     val indicesTensor = if (indices == null) Tensor[T]() else indices
     indicesTensor.resize(topKSize)
 
+    @inline
+    def compare(a: T, b: T): Boolean = if (increase) {
+      ev.isGreater(b, a)
```
wzhongyuan: How about 
```
ev.isGreater(b, a)  ^ !increase
```

i8run: OK.

spark/dl/src/main/scala/com/intel/analytics/bigdl/tensor/DenseTensor.scala
```diff
@@ -1717,24 +1720,62 @@ private[tensor] class DenseTensor[@specialized T: ClassTag](
     val indicesTensor = if (indices == null) Tensor[T]() else indices
     indicesTensor.resize(topKSize)
 
+    @inline
+    def compare(a: T, b: T): Boolean = ev.isGreater(b, a)  ^ !increase
+    /* if (increase) {
```
wzhongyuan: Why comment it instead of deleting it ?

i8run: Much easier to read it.

wzhongyuan: We need to remove it as commented code, probably you may want to add comment on top of the function

PR #2854, opened 2019-07-10
fix: upgrade the performance of normalize

i8run:

Enhance performance of `Normalize` which is the bottleneck of SSD.

## How was this patch tested?

Jenkins and accuracy test.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Normalize.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Normalize.scala
@@ -72,20 +72,24 @@ class Normalize[T: ClassTag](val p: Double, val eps: Double = 1e-10
       // normp.sum(buffer, 2).add(ev.fromType(eps))
       // perf fix: the sum operation of tensor will call Java + element wise
       // perf fix: start
-      normp.resize(Array(buffer.size(1), 1, buffer.size(3), buffer.size(4)))
-      var batchSize = 0
-      while (batchSize < normp.size(1)) {
-        val batchOfNormp = normp.select(1, batchSize + 1).zero
-        val batchOfInput = buffer.narrow(1, batchSize + 1, 1)
+      if (buffer.nDimension() <= 2) {
+         normp.sum(buffer, 2).add(ev.fromType(eps))
+      } else {
+        normp.resize(Array(buffer.size(1), 1, buffer.size(3), buffer.size(4)))
+        var batchSize = 0
+        while (batchSize < normp.size(1)) {
+          val batchOfNormp = normp.select(1, batchSize + 1).zero
+          val batchOfInput = buffer.narrow(1, batchSize + 1, 1)
 
-        var channel = 0
-        while (channel < buffer.size(2)) {
-          batchOfNormp.add(batchOfInput.select(2, channel + 1))
-          channel += 1
+          var channel = 0
+          while (channel < buffer.size(2)) {
+            batchOfNormp.add(batchOfInput.select(2, channel + 1))
+            channel += 1
+          }
+          batchSize += 1
         }
-        batchSize += 1
+        normp.add(ev.fromType(eps))
       }
-      normp.add(ev.fromType(eps))
       // perf fix: end
       norm.resizeAs(normp).pow(normp, ev.fromType(1.0 / p))
     }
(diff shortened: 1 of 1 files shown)


wzhongyuan:
Can you describe more details how you did to improve the performance in description ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Normalize.scala
```diff
@@ -69,13 +69,49 @@ class Normalize[T: ClassTag](val p: Double, val eps: Double = 1e-10
...
+      }
+      normp.add(ev.fromType(eps))
+      // perf fix: end
       norm.resizeAs(normp).pow(normp, ev.fromType(1.0 / p))
     }
     if (norm.dim() <= 2) {
       output.cdiv(inputBuffer, norm.view(norm.nElement(), 1).expandAs(inputBuffer))
     } else if (norm.dim() == 4) {
-      output.cdiv(inputBuffer, norm.view(norm.size()).expandAs(inputBuffer))
+      // output.cdiv(inputBuffer, norm.view(norm.size()).expandAs(inputBuffer))
```
wzhongyuan: remove it ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Normalize.scala
```diff
@@ -69,13 +69,49 @@ class Normalize[T: ClassTag](val p: Double, val eps: Double = 1e-10
       } else {
         buffer.resizeAs(inputBuffer).pow(inputBuffer, ev.fromType(p))
       }
-      normp.sum(buffer, 2).add(ev.fromType(eps))
+      // normp.sum(buffer, 2).add(ev.fromType(eps))
```
wzhongyuan: remove it ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Normalize.scala
i8run: I don't think we should remove it. Because this description is much simpler to understand how does this layer do. But the reason I did this optimization because of the `norm` will be expanded as an uncontiguous Tensor. So I had to compute it channel by channel.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Normalize.scala
i8run: The same as above. This is simple and easy to understand. The while loop below is the optimization version.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Normalize.scala
zhangxiaoli73: Which one is uncontiguous ? If just for contiguous, you can convert uncontiguous to contiguous.

zhangxiaoli73:
What is the performance improvement after optimization?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Normalize.scala
i8run: @zhangxiaoli73 I have tested, the conversion will take too much time and consume a lot of memory.

i8run:
@zhangxiaoli73 for the shape in SSD, this layer's performance has 7x improvement.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Normalize.scala
```diff
@@ -69,13 +69,53 @@ class Normalize[T: ClassTag](val p: Double, val eps: Double = 1e-10
...
       output.cdiv(inputBuffer, norm.view(norm.nElement(), 1).expandAs(inputBuffer))
     } else if (norm.dim() == 4) {
-      output.cdiv(inputBuffer, norm.view(norm.size()).expandAs(inputBuffer))
+      // output.cdiv(inputBuffer, norm.view(norm.size()).expandAs(inputBuffer))
+      // perf fix: after expand, the tensor will be not contiguous.
+      // perf fix: start
+      var batchSize = 0
+      while (batchSize < output.size(1)) {
+
+        val oneBatchOutput = output.narrow(1, batchSize + 1, 1)
```
wzhongyuan: `outputPerBatch` ?  similar to below norm ad input

i8run:
Jenkins passed. http://mlt-bddl01.sh.intel.com:8080/view/PR/job/BigDL-PRVN-lite/708/

PR #2859, opened 2019-07-11
feat: add axis to softmax

i8run:

This patch add `axis` parameter to `SoftMax`. Currently, when user want to softmax on the third dimension for 3-D input, they should use it in `TimeDistributed`. This property can make it simple and high performance with dnn backend.

Note: it only supports inference with dnn backend now.

## How was this patch tested?

Add test case and Jenkins.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/SoftMax.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/SoftMax.scala
@@ -63,24 +63,30 @@ class SoftMax(val axis: Int = -1) extends MklDnnLayer {
         val defaultAxis = inputs(0).shape.length match {
           case 1 => 0
           case 2 => 1
-          case 3 => 0
+          case 3 => 1
           case 4 => 1
           case _ => throw new UnsupportedOperationException("1D, 2D, 3D or 4D tensor expected")
         }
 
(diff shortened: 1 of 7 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/SoftMax.scala
```diff
@@ -58,20 +60,31 @@ class SoftMax() extends MklDnnLayer {
...
+          case 3 => 0
           case 4 => 1
-          case _ => throw new UnsupportedOperationException("1D, 2D, or 4D tensor expected")
+          case _ => throw new UnsupportedOperationException("1D, 2D, 3D or 4D tensor expected")
         }
 
         _inputFormats = singleNativeData(inputs)
+
+        val localInputFormat = if (inputs(0).shape.length == 3) {
+          NativeData(inputs(0).shape, Memory.Format.tnc)
```
zhangxiaoli73: if input is 3 dim, why your format is `tnc` ?

zhangxiaoli73: If support 4 dims input, why not use nchw directly?

i8run:
Jenkins has passed. http://mlt-bddl01.sh.intel.com:8080/view/PR/job/BigDL-PRVN-lite/710/

PR #2873, opened 2019-08-01
fix: Input and ParallelTable connection in Graph for backward

benjamim93:
Fix error which is reported when doing backward of a Graph with a connection between Input and ParallelTable

## How was this patch tested?
Unit test


wzhongyuan:
Add more details for the issue

zhangxiaoli73:
No need to change

PR #2880, opened 2019-08-09
Add set input output format API in Python

jenniew:

Add set input output format API in Python

## How was this patch tested?

add integration test case in Python

### DIFF
--- a/pyspark/bigdl/nn/layer.py
+++ b/pyspark/bigdl/nn/layer.py
@@ -902,6 +902,26 @@ class Model(Container):
         callBigDlFunc(bigdl_type, "saveGraphTopology", self.value, log_path)
         return self
 
+    def set_input_formats(self, input_formats, bigdl_type="float"):
+        """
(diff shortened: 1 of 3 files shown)


pyspark/bigdl/nn/layer.py
```diff
@@ -902,6 +902,26 @@ def save_graph_topology(self, log_path, bigdl_type="float"):
         callBigDlFunc(bigdl_type, "saveGraphTopology", self.value, log_path)
         return self
 
+    def set_input_formats(self, input_formats, bigdl_type="float"):
+        """
+        set input formats for graph.
```
zhangxiaoli73: Only support setting input formats for static graph

jenniew: Add check for the class name

zhangxiaoli73:
jenkins pass:
http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/753/

PR #2881, opened 2019-08-12
fix momentum in branch-0.9

zhangxiaoli73:

convert mklblas momentum in batchnormalization to mkldnn, as they have different meaning.

## How was this patch tested?

training resnet50 with mkldnn

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/resnet/TrainImageNet.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/resnet/TrainImageNet.scala
@@ -68,31 +68,25 @@ object TrainImageNet {
       val model = if (param.modelSnapshot.isDefined) {
         Module.load[Float](param.modelSnapshot.get)
       } else {
-        Engine.getEngineType() match {
-          case MklBlas =>
-            val curModel =
-              ResNet(classNum = param.classes, T("shortcutType" -> shortcut, "depth" -> param.depth,
-                "optnet" -> param.optnet, "dataSet" -> dataSetType))
-            if (param.optnet) {
-              ResNet.shareGradInput(curModel)
-            }
-            ResNet.modelInit(curModel)
-
-            /* Here we set parallism specificall for BatchNormalization and its Sub Layers, this is
-            very useful especially when you want to leverage more computing resources like you want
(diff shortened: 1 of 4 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/models/resnet/TrainImageNet.scala
```diff
@@ -68,31 +68,25 @@ object TrainImageNet {
...
+
+        /* Here we set parallism specificall for BatchNormalization and its Sub Layers, this is
+        very useful especially when you want to leverage more computing resources like you want
+        to use as many cores as possible but you cannot set batch size too big for each core due
+        to the memory limitation, so you can set batch size per core smaller, but the smaller
+        batch size will increase the instability of convergence, the synchronization among BN
+        layers basically do the parameters synchronization among cores and thus will avoid the
+        instability while improves the performance a lot. */
+        val parallisim = Engine.coreNumber
+        setParallism(curModel, parallisim)
```
wzhongyuan: I don't think we need this for MKL-DNN as we use all assigned cores

wzhongyuan:
Do we need to modify the README to sepcify MKL-DNN property ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/resnet/TrainImageNet.scala
zhangxiaoli73: Yeah, when training with MKLDNN, `setParallism` is useless even through it is set, but for MKL-BLAS, it can speed up. So I just keep this line .

zhangxiaoli73:
Yes, we can stress resnet50 training with MKL-DNN. I will refactor right now.

spark/dl/src/main/scala/com/intel/analytics/bigdl/models/resnet/TrainImageNet.scala
wzhongyuan: it's better to have EngineType checking so that only MKL-BLAS mode will be using it

zhangxiaoli73: ok, changed

wzhongyuan:
LGTM

PR #2885, opened 2019-08-16
add Shape layer

LeicongLi:
1. Added Shape layer to support loading ONNX model
https://github.com/onnx/onnx/blob/master/docs/Operators.md#shape

2. Added bridge between python api and onnx package in scala.

## How was this patch tested?


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/onnx/Shape.scala
```diff
@@ -0,0 +1,60 @@
...
+ * @param ev
+ * @tparam T The numeric type in this module parameters
+ */
+class Shape[T: ClassTag](implicit ev: TensorNumeric[T]) extends TensorModule[T] {
+
+  override def updateOutput(input: Tensor[T]): Tensor[T] = {
+    val dimSize = input.nDimension()
+    output = Tensor(dimSize)
+    (1 to dimSize).foreach(i => {
+      output.setValue(i, input.size(i).toFloat.asInstanceOf[T])
```
i8run: Use `ev.fromType` instead the `.toFloat.asInstanceOf[T]`.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/onnx/Shape.scala
```diff
@@ -0,0 +1,60 @@
...
+    })
+    output
+  }
+
+  override def updateGradInput(input: Tensor[T], gradOutput: Tensor[T]): Tensor[T] = {
+    gradInput = gradOutput
```
i8run: So why the `gradInput` is the same as the `gradOutput`. If the input is a `Tensor(Array(4, 3, 224, 224))`, so what's the content and shape of the `gradOutput` ?

LeicongLi: The function of Shape layer is taking an tensor, and output its shape.
e.g. Input Tensor(Array(4, 3, 224, 224)) => Output Tensor(Array(4)), a one dimension array with four elements (4, 3, 224, 224).

i8run: I'm confusing about the shape of gradInput and gradOutput. If the input is the shape of [4, 3, 224, 224], the gradOutput's shape will be [1, 4] or [4]. So the questions is, what's the shape of gradInput ? Is it [4, 3, 224, 224] or [4] ?

LeicongLi: I see what you are pointing out, gradInput should have the same shape as input which is [4, 3, 224, 224].

LeicongLi: It is a constant layer, I dont think the input gradient should be changed.
And currently, I might not enable backforward on this op.

PR #2893, opened 2019-08-21
feat: MKLDNN GRU forward/backward support

benjamim93:
Enable MKLDNN GRU forward/backward

## How was this patch tested?
Unit tests. Compared MKLDNN results with BLAS results.

Jenkins passed: http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/916/

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/RNN.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/RNN.scala
@@ -119,6 +119,11 @@ class RNN(
    * MKLDNN Gate 2 -> nn/LSTM Gate 3 (forget gate)
    * MKLDNN Gate 3 -> nn/LSTM Gate 2 (hidden)
    * MKLDNN Gate 4 -> nn/LSTM Gate 4 (output gate)
+   *
+   * Gate order matching between MKLDNN GRU and nn/GRU:
+   * MKLDNN Gate 1 -> nn/GRU Gate 2
+   * MKLDNN Gate 2 -> nn/GRU Gate 1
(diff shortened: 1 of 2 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/RNN.scala
```diff
@@ -165,14 +171,6 @@ class RNN(
       PropKind.ForwardTraining
     }
 
-    /**
```
zhangxiaoli73: I think those comments should not be deleted.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/RNN.scala
```diff
@@ -181,8 +179,8 @@ class RNN(
         batchSize = inputs(0).shape(0)
         stepSize = inputs(0).shape(1)
       case _ =>
-        throw new UnsupportedOperationException("Not support such input format. " +
-          "The input format is: " + inputs(0).layout)
+        throw new UnsupportedOperationException("Not support such input format: " +
```
wzhongyuan: Unsupported input format

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/RNN.scala
benjamim93: Changed back

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/RNN.scala
benjamim93: Changed

PR #2903, opened 2019-08-30
Add Constant layer

LeicongLi:
Add a Constant layer, which takes no input, and forward the constant tensor it contains to next layer.
Allow forward method in AbstractModule takes no input

## How was this patch tested?


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -252,7 +252,7 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
    * @param input input data
    * @return output data
    */
-  final def forward(input: A): B = {
+  final def forward(input: A = null.asInstanceOf[A]): B = {
```
wzhongyuan: Remove it to avoid confusion

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/onnx/Constant.scala
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * A layer contains a constant tensor.
+ * It forwards the value of constant tensor to next layers.
+ *
+ * @param value constant tensor
+ * @param `classTag$T`
+ * @param ev
+ * @tparam T The numeric type in this module parameters
+ */
+class Constant[T: ClassTag](
```
wzhongyuan: why not reusing reuse `com.intel.analytics.bigdl.nn.tf.Const` ?

spark/dl/src/test/scala/com/intel/analytics/bigdl/nn/onnx/ConstantSpec.scala
```diff
@@ -0,0 +1,42 @@
...
+    val value = Tensor[Float](512, 3, 224, 224).rand()
+    val constant = Constant(value)
+    val out = constant.forward()
+    out should be(value)
+  }
+
+}
+
+
+class AbsSerialTest extends ModuleSerializationTest {
```
wzhongyuan: why AbsSerialTest in this file ?

LeicongLi: Thanks.
Wrong copy & paste, will fix it.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/onnx/Constant.scala
LeicongLi: As metioned earlier, the Constant layer takes **no** input.
com.intel.analytics.bigdl.nn.tf.Const takes input, and it extends Operation, which does not allow backforward.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
LeicongLi: This is the main change of this PR.
Constant layer requires no input, which mean the forward method here needs to take no input, the input needs to be null.

wzhongyuan: This is just default value, it is not mandatory that the input has to be null. Makes no difference here

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/onnx/Constant.scala
wzhongyuan: Do you mean the Const operation in onnx could have preceding layer/ops?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
LeicongLi: The difference is between forward(null) and forward().
First case, forward still takes input, and the input is null.
The second case, forward takes no input.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/onnx/Constant.scala
```diff
@@ -0,0 +1,57 @@
...
+  val value: Tensor[T]
+)(implicit ev: TensorNumeric[T]) extends TensorModule[T] {
+
+  override def updateOutput(input: Tensor[T] = null): Tensor[T] = {
+    output = value
+    output
+  }
+
+  override def updateGradInput(input: Tensor[T], gradOutput: Tensor[T]): Tensor[T] = {
+    gradInput
```
wzhongyuan: `gradInput` is null by default, should be sth like below,
```
  if (gradInput == null || gradInput.nElement() != gradOutput.nElement()) {
     gradInput.resizeAs(gradOutput).zero()
 }
```

LeicongLi:
Change Constant into regular layer.
Add it into another PR, Onnx Support Layers.
Close this one.

PR #2910, opened 2019-09-09
Onnx support: modify unsqueeze function

lingxiao1989:

## How was this patch tested?

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Unsqueeze.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Unsqueeze.scala
@@ -70,6 +70,7 @@ class Unsqueeze[T: ClassTag](
         s"input:dim() is $input, input feature map dim (numInputDims) is $numInputDims.")
       actualPos
     })
+    //increase 1 to the following pos after a previous smaller pos have one dimension inserted.
     for {
       i <- 1 to pos.length
       j <- i + 1 to pos.length
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Unsqueeze.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Unsqueeze.scala
@@ -67,7 +67,7 @@ class Unsqueeze[T: ClassTag](
         s" input feature map dim ${numInputDims}, inputdim ${inputDim}")
       // the actual position; clearer error message for batchMode (if any)
       val actualPos = pos(index) + offsetDim
-      require(actualPos >= 1 && actualPos <= (inputDim + 1), s"Invalid position: $pos. " +
+      require(actualPos >= 1 && actualPos <= (inputDim + 1), s"Invalid position: ${pos(index)}. " +
         s"input:dim() is $input, input feature map dim (numInputDims) is $numInputDims.")
       pos(index) = actualPos
     }
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Unsqueeze.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Unsqueeze.scala
@@ -50,13 +50,7 @@ class Unsqueeze[T: ClassTag](
 
   private def getActualPosition(input: Tensor[_]) : Array[Int] = {
     pos.foreach(pos => {
-      val dim = if (pos <= 0) {
-        require(pos <= 0, s"invalid input: index starts from 1, " +
-          s"creating a singleton dim at the end of the input tensor")
-        input.dim() + pos + 1
-      } else {
-        pos
-      }
+      require(pos > 0, s"invalid input: index starts from 1" )
       // get valid dimension offset for batchMode (if any)
       val inputDim = input.dim() // data batch dim
       numInputDims = if (numInputDims != Int.MinValue) numInputDims else inputDim // feature map dim
@@ -65,7 +59,7 @@ class Unsqueeze[T: ClassTag](
         s" input feature map dim ${numInputDims}, inputdim ${inputDim}")
 
       // the actual position; clearer error message for batchMode (if any)
-      val actualPos = dim + offsetDim
+      val actualPos = pos + offsetDim
       require(actualPos >= 1 && actualPos <= (inputDim + 1), s"Invalid position: $pos. " +
         s"input:dim() is $input, input feature map dim (numInputDims) is $numInputDims.")
       actualPos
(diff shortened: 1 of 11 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Unsqueeze.scala
```diff
@@ -32,46 +32,51 @@ import scala.reflect.ClassTag
...
-
-    // the actual position; clearer error message for batchMode (if any)
-    val actualPos = dim + offsetDim
-    require(actualPos >= 1 && actualPos <= (inputDim + 1), s"Invalid position: $pos. " +
-      s"input:dim() is $input, input feature map dim (numInputDims) is $numInputDims.")
-
-    actualPos
+  private def getActualPosition(input: Tensor[_]) : Array[Int] = {
+    pos.foreach(pos => {
+      val dim = if (pos <= 0) {
```
LeicongLi: Please try to simplify the logic here.
Suggest get rid of If Else statement.
The Require statement enforces pos > 0.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Unsqueeze.scala
```diff
@@ -32,46 +32,51 @@ import scala.reflect.ClassTag
...
+        require(pos <= 0, s"invalid input: index starts from 1, " +
```
LeicongLi: The condition is opposite, we want pos > 0, right?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Unsqueeze.scala
```diff
@@ -32,46 +32,51 @@ import scala.reflect.ClassTag
...
+      // the actual position; clearer error message for batchMode (if any)
+      val actualPos = dim + offsetDim
+      require(actualPos >= 1 && actualPos <= (inputDim + 1), s"Invalid position: $pos. " +
+        s"input:dim() is $input, input feature map dim (numInputDims) is $numInputDims.")
+      actualPos
+    })
+    for {
+      i <- 1 to pos.length
+      j <- i to pos.length
+    } if (pos(j) > pos(i)) { pos(j) = pos(j) + 1}
```
LeicongLi: Please add comment/explanation to this code.
j <- i + 1 to pos.length would make it easier to understand?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Unsqueeze.scala
```diff
@@ -108,6 +113,12 @@ object Unsqueeze {
   def apply[@specialized(Float, Double) T: ClassTag](
     pos: Int,
     numInputDims: Int = Int.MinValue)(implicit ev: TensorNumeric[T]) : Unsqueeze[T] = {
-    new Unsqueeze[T](pos, numInputDims)
+    new Unsqueeze[T](Array(pos), numInputDims)
+  }
+
+  def apply[@specialized(Float, Double) T: ClassTag](
```
LeicongLi: Please add Python API for this.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Unsqueeze.scala
```diff
@@ -32,46 +32,51 @@ import scala.reflect.ClassTag
...
     if (input.getType() != output.getType()) {
       output = input.emptyInstance()
     }
-
-    output
+    actualPos.foreach(output
       .asInstanceOf[Tensor[NumericWildcard]]
-      .addSingletonDimension(input.asInstanceOf[Tensor[NumericWildcard]], actualPos)
-
+      .addSingletonDimension(input.asInstanceOf[Tensor[NumericWildcard]], _))
```
LeicongLi: Can we do the insertions at same time?
The for loop hurts performance.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Unsqueeze.scala
lingxiao1989: Yes, ok

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Unsqueeze.scala
lingxiao1989: the original function takes 0 as input.dim() + pos + 1. Are you suggesting we take this out and only take numbers >0 ?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Unsqueeze.scala
lingxiao1989: ok

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Unsqueeze.scala
lingxiao1989: I think the loop takes more time than one insertion call, but just a linear increase. We will still use a loop to add multiple dimensions to a tensor, right?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Unsqueeze.scala
LeicongLi: How are you going to treat the input if it is 2, do you know it is starting from 0 or 1?

lingxiao1989: the code assume the input start from 1, and there is a require to make sure the pos is no larger than the input size:
require(actualPos >= 1 && actualPos <= (inputDim + 1), s"Invalid position: $pos. " +
        s"input:dim() is $input, input feature map dim (numInputDims) is $numInputDims.")

LeicongLi: I dont understand.
The If statement here guarantee pos <= 0, why there is a require statement after it?

spark/dl/src/main/scala/com/intel/analytics/bigdl/python/api/PythonBigDL.scala
```diff
@@ -1531,6 +1531,13 @@ class PythonBigDL[T: ClassTag](implicit ev: TensorNumeric[T]) extends Serializab
       numInputDims)
   }
 
+  def createMultipleDimUnsqueeze(pos: Array[Int],
```
wzhongyuan: I don't think we need 2 separate bridge method here, please consolidate it into one,  removing above one and use this one works for both.

lingxiao1989: Ok, I have change it into one method.

lingxiao1989:
Jenkins:
http://mlt-bddl01.sh.intel.com:8080/view/PR/job/BigDL-PRVN-lite/915/

lingxiao1989:
http://mlt-bddl01.sh.intel.com:8080/view/PR/job/BigDL-PRVN-lite/918/

PR #2914, opened 2019-09-16
Fix memory leaks on training

Menooker:

Two sources of memory leaks fixed

 * In SpatialBatchNormalization, two different sets of native memory are used for training and inference. Calling initFwdPrimitives will switch to one of the set. No more than 2 sets of native memory are created
 * In preprocessing, OpenCV sub-matrix created by `submat` is not released. According to the OpenCV [source code](https://sourcegraph.com/github.com/opencv/opencv/-/blob/modules/core/src/matrix.cpp#L455), the submats will increase the underlying memory storage's ref-count. Thus we must release the submats. 

## How was this patch tested?

UT, manual testing


Menooker:
@i8run This PR overwrites your commit in SpatialBatchNormalization. Plz check if I miss any thing in your commit.

wzhongyuan:
@i8run  @zhangxiaoli73  please take a look

Menooker:
Jenkins pass http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/930/

PR #2918, opened 2019-09-25
ONNX Support

LeicongLi:
Implement a load function to load onnx model file into BigDL model.

usage example:
```
import numpy as np
from bigdl.contrib.onnx2bigdl import load_onnx

resnet_path = "model_dir/resnet-50.onnx"
resnet = load_onnx(resnet_path)
resnet_tensor = np.random.random([1, 3, 224, 224])
resnet_out = resnet.forward(resnet_tensor)
resnet_out

```

## How was this patch tested?


pyspark/bigdl/contrib/onnx2bigdl/__init__.py
```diff
@@ -0,0 +1,17 @@
+#
+# Copyright 2016 The BigDL Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
```
wzhongyuan: can we just name the package to `onnx`, similar as for other FWKs

pyspark/bigdl/contrib/onnx2bigdl/ops_translation.py
```diff
@@ -0,0 +1,255 @@
+#
+# Copyright 2016 The BigDL Authors.
```
wzhongyuan: translation is a bit confusing

pyspark/bigdl/contrib/onnx2bigdl/import_graph.py
```diff
@@ -0,0 +1,130 @@
...
+#
+
+import numpy as np
+from bigdl.nn.onnx.layer import *
+from bigdl.nn.layer import Identity, Model
+
+from .ops_mapping import _convert_map as convert_map
+
+
+class BigDLGraph(object):
```
wzhongyuan: I think it's a equivalent ONNX graph right ?  Not BigDL graph, loadONNX should eventually return BigDL graph. We should not define a ONNX equivalent class in BigDL, for those states that are different from ONNX, we should hide them in loading and just return BigDL model

LeicongLi: Hi Jerry, thanks for your time!
Those classes are hidden from users, I have updated the PR with some example codes.
If you still think those classes unnecessary, they could be deleted.

wzhongyuan: Not necessarily delete it if required,  it makes more sense to update the name such as `ONNXGraph` instead of 'BigDLGraph' to avoid confusion.

pyspark/bigdl/contrib/onnx2bigdl/import_graph.py
```diff
@@ -0,0 +1,130 @@
...
+		self._inputs = list()
+		self._outputs = list()
+		self._tensors = dict()
+		self._modules = dict()
+		self._root = list()
+		self._dummy_root = Identity()()
+		self._graph = self.load_graph(graph_proto)
+		return
+
+	def load_graph(self, graph_proto):
```
wzhongyuan: why the file name is `import_graph` while here you name it `load_graph` ? please keep consistent.

pyspark/bigdl/contrib/onnx2bigdl/import_model.py
```diff
@@ -0,0 +1,51 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import onnx
+from .import_graph import BigDLGraph
+
+
+class BigDLModel(object):
```
wzhongyuan: I think this class is an utility instead of stateful class, and `BigDLModel` is confusing. `ONNXLoader` makes more sense.

pyspark/bigdl/contrib/onnx2bigdl/ops_translation.py
wzhongyuan: similar to other FWKs support, 'Converter' or 'OpsLoader' is better

pyspark/bigdl/contrib/onnx2bigdl/import_graph.py
LeicongLi: Thanks, changed.

pyspark/bigdl/contrib/onnx2bigdl/import_graph.py
LeicongLi: Thanks, changed.

pyspark/bigdl/contrib/onnx2bigdl/load_graph.py
```diff
@@ -0,0 +1,131 @@
...
+			# If it is a scalar tensor
+			np_array = np.array([to_array(tensor_proto)])
+		return np_array
+
+	def _parse_node_attr(self, node_proto):
+		attrs = {}
+		attr_proto = node_proto.attribute
+
+		for attr in attr_proto:
+			for field in ['f', 'i', 's']:
```
wzhongyuan: what does f i and s mean respectively ?

wzhongyuan: seems for floats, ints and strings ?

pyspark/bigdl/contrib/onnx2bigdl/load_graph.py
```diff
@@ -0,0 +1,131 @@
...
+				if list(getattr(attr, field)):
+					assert attr.name not in attrs, "Only one type of attr is allowed"
+					attrs[attr.name] = tuple(getattr(attr, field))
+
+			for field in ['t', 'g']:
+				if attr.HasField(field):
+					attrs[attr.name] = getattr(attr, field)
+			for field in ['tensors', 'graphs']:
+				if list(getattr(attr, field)):
+					raise NotImplementedError()
```
wzhongyuan: Can we pass a string to the exception ?

pyspark/bigdl/contrib/onnx2bigdl/load_model.py
```diff
@@ -0,0 +1,54 @@
...
+	def load_model(self, model_proto):
+		self._model_proto = model_proto
+		# self._ir_version = model_proto.ir_version
+		# self._opset_import = model_proto.opset_import
+		# self._producer_name = model_proto.producer_name
+		# self._producer_version = model_proto.producer_version
+		# self._domain = model_proto.domain
+		# self._model_version = model_proto.model_version
+		# self._doc_string = model_proto.doc_string
+		bgraph = OnnxGraph()
```
wzhongyuan: bgraph ?

wzhongyuan:
You may want to add an example for it, but please do with another PR

pyspark/bigdl/contrib/onnx2bigdl/load_graph.py
```diff
@@ -0,0 +1,131 @@
...
+		self._tensors = dict()
+		self._modules = dict()
+		self._root = list()
+		self._dummy_root = Identity()()
+		self._graph = self.load_graph(graph_proto)
+		return
+
+	def load_graph(self, graph_proto):
+		if not graph_proto:
+			return None
```
wzhongyuan: Raising an error makes more sense ?

pyspark/bigdl/contrib/onnx2bigdl/load_model.py
```diff
@@ -0,0 +1,54 @@
...
+	def __init__(self, file_path=None):
+		if file_path:
+			self._model_proto = onnx.load_model(file_path)
+		else:
+			self._model_proto = None
+
+	def get_model_proto(self):
+		return self._model_proto
+
+	def load_model(self, model_proto):
```
wzhongyuan: why do we need to create a separate method since per below you just call `bgraph.load_graph(graph_proto)` ?   there is no extra logic inside this method per my understanding

pyspark/bigdl/contrib/onnx2bigdl/load_graph.py
```diff
@@ -0,0 +1,131 @@
...
+
+class OnnxGraph(object):
+	def __init__(self, graph_proto=None):
+		self._inputs = list()
+		self._outputs = list()
+		self._tensors = dict()
+		self._modules = dict()
+		self._root = list()
+		self._dummy_root = Identity()()
+		self._graph = self.load_graph(graph_proto)
```
wzhongyuan: why you need this private member ? I did not see its usage.  And `load_graph` will be called on initialization, why do we still call it outside ? it's duplicated.

pyspark/bigdl/contrib/onnx2bigdl/load_graph.py
```diff
@@ -0,0 +1,131 @@
...
+# limitations under the License.
+#
+
+import numpy as np
+from bigdl.nn.onnx.layer import *
+from bigdl.nn.layer import Identity, Model
+from .ops_mapping import _convert_map as convert_map
+
+
+class OnnxGraph(object):
```
wzhongyuan: it seems to me that this class is not a modeling for `ONNX` , still more like a utility and the class member defined below are just intermediate state for loading model only. We don't even need to model onnx as it's already provide by onnx (below  `onnx.load_model(file_path)`).  Can we combine the two classes together as one like `ONNXLoader` ?

pyspark/bigdl/contrib/onnx2bigdl/load_graph.py
LeicongLi: They are float, int, string.
And there are separate logic dealing with list of them.

pyspark/bigdl/contrib/onnx2bigdl/load_graph.py
LeicongLi: Yes

pyspark/bigdl/contrib/onnx2bigdl/load_model.py
LeicongLi: There are other model meta data besides graph.
Though we just load the graph for now, I think its necessary to load all of them into the model, in case users want to check the loaded onnx model.

pyspark/bigdl/contrib/onnx2bigdl/load_graph.py
LeicongLi: Thanks, will remove it.

pyspark/bigdl/contrib/onnx2bigdl/OnnxLoader.py
```diff
@@ -0,0 +1,111 @@
...
+                raise ValueError("Tensor's name is required")
+            initialized_tensors.add(tensor.name)
+            tensor_data = parse_tensor_data(tensor)
+            tensor_map[tensor.name] = (tensor_data, tensor_data.shape)
+
+        for gin in graph_proto.input:
+            if gin.name not in initialized_tensors:
+                input_nodes.append(gin.name)
+                shape = tuple([dim.dim_value for dim in gin.type.tensor_type.shape.dim])
+                module_map[gin.name] = Identity()(dummy_root)
```
wzhongyuan: why each input is an identity following a dumpy identity ?

wzhongyuan:
Generally LGTM, please see my previous minor comments. Also we need to add an example for a typical model. Doc is needed as well.

wzhongyuan:
LGTM, please add an example and related doc in separate PR

PR #2936, opened 2019-10-17
feat: multi models support with MKL-DNN backend

i8run:

This patch add multi model supports for mkl-dnn backend with sharing weights.

1. Currently, the share weights feature only supports `Predictor` and `Evaluator`. Because for mkl-dnn backend, we can't get the native weight when creating model. So for sharing weights, the dummyInput is needed to generate the weights memory first.

2. There're three modes we can support now.  So add a property named `multiModel` to handle different mkl-dnn modes (single model multi omp thread or multi model with simple omp thread).

## How was this patch tested?

Accuracy tests and performance tests with ResNet-50 by manual. Unit tests will upload later.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
@@ -240,7 +240,8 @@ private[bigdl] class ModelBroadcastImp[T: ClassTag](applyProtoBuffer: Boolean = 
 
   override def broadcast(sc: SparkContext, model: Module[T],
     dummyInput: Activity): this.type = {
-    if (model.isInstanceOf[IRGraph[T]] && Engine.getEngineType() == MklDnn) {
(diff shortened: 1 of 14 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala
```diff
@@ -46,6 +50,11 @@ trait ModelBroadcast[T] extends Serializable {
    */
   def broadcast(sc: SparkContext, model: Module[T]): this.type
 
+  private[bigdl] def broadcast(sc: SparkContext, model: Module[T],
+    dummyInput: Activity): this.type = {
+    throw new UnimplementedException
```
wzhongyuan: Add information such as "broadcasr not supported in ...."

i8run: this method has not string parameters.

i8run:
Jenkins has passed. http://mlt-bddl01.sh.intel.com:8080/view/PR/job/BigDL-PRVN-lite/992/

PR #2941, opened 2019-10-22
feat: add precision recall auc

i8run:

Add a new validation method.

## How was this patch tested?

Unit tests.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/PrecisionRecallAUC.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/PrecisionRecallAUC.scala
@@ -25,8 +25,8 @@ import scala.reflect.ClassTag
  * Precision Recall Area Under Curve will compute the precision-recall pairs and
  * get the area under the curve.
  *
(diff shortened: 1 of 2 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/PrecisionRecallAUC.scala
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+package com.intel.analytics.bigdl.optim
+
+import com.intel.analytics.bigdl.nn.abstractnn.Activity
+import com.intel.analytics.bigdl.tensor.TensorNumericMath.TensorNumeric
+
+import scala.reflect.ClassTag
+
+class PrecisionRecallAUC[T: ClassTag](val ignoreBadMetric: Boolean = false)(
```
wzhongyuan: Add comments ?

i8run: Added.

i8run:
Jenkins has passed.

PR #2947, opened 2019-10-28
fix: takeSample only works for dnn backend and get one batch

i8run:

For multi models of MKL-DNN backend, we should take a batch back to driver first for get the input size, where exists serialization. But for some objects output of transformer can not be serialized. Add this patch to narrow the influence.

## How was this patch tested?

Manual tests and Jenkins.


spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/Evaluator.scala
```diff
@@ -88,7 +89,12 @@ class Evaluator[T: ClassTag] private[optim](model: Module[T])(implicit ev: Tenso
            vMethods: Array[ValidationMethod[T]]
           ): Array[(ValidationResult, ValidationMethod[T])] = {
 
-    val dummyInput = dataset.takeSample(withReplacement = false, num = 1).head.getInput()
+    // TODO this should be fixed next release because of some objects can't be serialized
```
zhangxiaoli73: no need. We are woking on fixing it. Just remind minibatch should be serializable.

zhangxiaoli73:
others LGTM

PR #2958, opened 2019-11-15
fix: skip CAddTable fusion for some special case

benjamim93:

A problem was detected while doing MobileNet V2 inference. MobileNet V2 contains some node connection structure like below (A->B indicates the change from input channel number A to output channel number B)
                     conv1(96 -> 96)
                               /\
                             /    \
        conv2(96 -> 24)   conv3(96 -> 24)
                          /\        |
                        /    \      |
    conv4(24 ->144) CAddTable

The original method would turn forward execution into some graph like:
                     conv1(96 -> 96)
                               /\
                             /    \
                    Identity    conv3(96 -> 24)
                          /\        |
                        /    \      |
    conv4(24 ->144) conv2(96 -> 24)

resulting in a mismatch between the input channel number and output channel number for conv4. 

Skip fusion if the previous node of CAddTable has more than one next nodes.

## How was this patch tested?

Tested MobileNet v2. The inference works.


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/Fusion.scala
```diff
@@ -98,7 +98,10 @@ private[mkldnn] object Fusion {
 
   private def findPrevious(node: Node[AbstractModule[Activity, Activity, Float]])
   : Node[AbstractModule[Activity, Activity, Float]] = {
-    if (node.element.isInstanceOf[Identity] && node.prevNodes.length == 1) {
+    if (node.element.isInstanceOf[Identity]
+      && node.nextNodes.length == 1
+      && node.prevNodes.length == 1
+      && node.prevNodes(0).nextNodes.length == 1) {
```
i8run: Why should we add line 104 here ?

benjamim93: This line is for cases like below (mentioned by @zhangxiaoli73 ):
```
                        conv1(96 -> 96)
                               |
                               |
                        conv2(96 -> 24)
                              /\        
                            /    \      
            conv3(24 ->144)    Identity    conv4(96 -> 24)
                                        \    /
                                          \/
                                        CAddTable
```

benjamim93: I am not sure if there are any possible corner cases not taken into account yet.

i8run: It's OK. But we should add at least two unit tests here for both cases.

benjamim93: Added tests and added a condition for the case shown in the graph at the very beginning. I mean the convs without following relu, bn, etc.

PR #2966, opened 2019-11-19
fix: fuse bn scale and relu to bn.

i8run:

The original fusion only fuses the bn with scale and will skip the bn + identity + relu.

## How was this patch tested?

Unit tests.

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnGraph.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnGraph.scala
@@ -394,14 +394,15 @@ class DnnGraph(
    */
   private def fusion(): Unit = {
     if (!this.train) {
-      for (j <- 0 to 3) {
+      for (j <- 0 to 4) {
         var i = forwardExecution.length - 1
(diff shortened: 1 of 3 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnGraph.scala
```diff
@@ -394,14 +394,15 @@ class DnnGraph(
    */
   private def fusion(): Unit = {
     if (!this.train) {
-      for (j <- 0 to 3) {
+      for (j <- 0 to 4) {
         var i = forwardExecution.length - 1
         while (i >= 0) {
-          if (j == 0) Fusion.fuseModule(forwardExecution(i))
+          if (j == 0) Fusion.fuseScale(forwardExecution(i))
```
zhangxiaoli73: So, `scale` is not fused in our previous release?

i8run: 0.10 only fuse the bn + scale. But has no bn + scale + relu fusion.

zhangxiaoli73:
LGTM

PR #2970, opened 2019-11-26
[WIP] Sequential Keras layers to graph

benjamim93:
Looks like there is no support for converting Sequential model of keras layers into graph. This part is under construction.

## How was this patch tested?

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
@@ -826,7 +826,17 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag, 
    * @return
    */
   def toGraph(startNodes: ModuleNode[T]*): Graph[T] = {
-    val starts = if (startNodes.isEmpty) Array(Input[T]()) else startNodes.toArray
+    val starts = if (startNodes.isEmpty) {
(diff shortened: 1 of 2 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/KerasLayer.scala
```diff
@@ -183,6 +183,15 @@ abstract class KerasLayer[A <: Activity: ClassTag, B <: Activity: ClassTag, T: C
   }
  // scalastyle:on
 
+  // TODO: KerasLayer in Analytics Zoo has a toModel() implementation
+
+  override def parameters(): (Array[Tensor[T]], Array[Tensor[T]]) = {
+    if (this.modules.length > 1) super.parameters()
```
i8run: please add a brace of `if`.

benjamim93: Changed

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/keras/Topology.scala
```diff
@@ -187,6 +187,27 @@ class Model[T: ClassTag](private val _inputs : Seq[ModuleNode[T]],
     checkWithCurrentInputShape(calcInputShape)
     getOutputShape()
   }
+
+  // debug
+  def toGraph(): Graph[T] = {
```
zhangxiaoli73: useless

PR #2975, opened 2019-12-02
support roialign backward

zhangxiaoli73:

support roialign backward

## How was this patch tested?

unit tests


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RoiAlign.scala
```diff
@@ -109,8 +111,180 @@ class RoiAlign[T: ClassTag] (
...
+
+      var roi_width = roi_end_w - roi_start_w
+      var roi_height = roi_end_h - roi_start_h
+
+      if (aligned) {
+        require(roi_width >= 0 && roi_height >= 0,
+          s"ROIs in ROIAlignRotated do not have non-negative size!" +
+            s"But get ${roi_height} ${roi_width}")
+      } else {
+        roi_width = math.max(roi_width, 1.0f)
```
benjamim93: Why this part is skipped if aligned?

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RoiAlign.scala
```diff
@@ -109,8 +111,180 @@ class RoiAlign[T: ClassTag] (
...
+              ev.plus(gradInputArr(gradInputOffset + y_high * width + x_high), g4)
+          }
+        }
+      }
+    }
+  }
+
+  override def updateGradInput(input: Activity, gradOutput: Tensor[T]): Activity = {
+    require(mode == "avg", s"Only support backward for average mode, but get ${mode}")
+    val data = input.toTable[Tensor[Float]](1)
```
benjamim93: Looks like the backward limited to float only?

zhangxiaoli73: no, it should be classtag `T`

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RoiAlign.scala
zhangxiaoli73: for aligned case, we allow roi height/width less than 1.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RoiAlign.scala
benjamim93: Looks good to me.

zhangxiaoli73:
jenkins pass:
http://mlt-bddl01.sh.intel.com:8080/view/PR/job/BigDL-PRVN-lite/1099/

PR #2989, opened 2019-12-24
NHWC support when running with MKL-DNN

zhangxiaoli73:

mainly for NHWC support when running with MKL-DNN, summary changes:
(1) Add `heapFormat` (NHWC or NCHW) for models, so when one layer uses 4 dims, it can get its real format which may be not same with format from previous layer.

(2) Add `DataFormat` for DNN layers according to BLAS layers. E.g. Blas MaxPooling has DataFormat in its construct parameters, so add the same param for DNN MaxPooling.

(3) Refactor mkldnn.Linear to support NHWC input

(4) Deprecate  `DropOut` for MKLDNN, use `BlasWrapper(nn.DropOut)` if needed.

(5) Delete some constraints for NHWC, and now we can run models with NHWC input under MKLDNN backend

## How was this patch tested?

unit tests

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnGraph.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnGraph.scala
@@ -83,7 +83,8 @@ class DnnGraph(
       inputCache(i) = nodeInput
       val output = node.element.forward(nodeInput)
       // resize to heap size
-      if (!skipPrimitiveId(i) && output.isTensor) {
+      if (!skipPrimitiveId(i) && output.isTensor &&
+        !node.element.isInstanceOf[BlasWrapper]) {
         output.toTensor[Float].resize(
           node.element.asInstanceOf[MklDnnLayer].outputFormats()(0).getHeapShape())
       }
@@ -112,7 +113,8 @@ class DnnGraph(
       if (!isStopGradient(curNode.element)) {
         val gradInput = curNode.element.updateGradInput(curInput, curGradOutput)
         // resize to heap size
-        if (!skipPrimitiveId(i) && gradInput.isTensor) {
+        if (!skipPrimitiveId(i) && gradInput.isTensor &&
+          !curNode.element.isInstanceOf[BlasWrapper]) {
           gradInput.toTensor[Float].resize(
             curNode.element.asInstanceOf[MklDnnLayer].gradInputFormats()(0).getHeapShape())
         }
@@ -466,10 +468,10 @@ class DnnGraph(
       if (!skipPrimitiveId(i)) {
         val m = forwardExecution(i)
         lastOutputFormats = findInputFormats(m, inputs)
-        val realInputAndOutputFormats =
+        val (realInputFormats, realOutputFormats) =
           m.element.asInstanceOf[MklDnnModule].initFwdPrimitives(lastOutputFormats, phase)
-        realInputAndOutputFormats._2.foreach(_.setHeapFormat(heapFormat))
-        lastOutputFormats.zip(realInputAndOutputFormats._1).foreach {
+        if (realOutputFormats != null) realOutputFormats.foreach(_.setHeapFormat(heapFormat))
+        lastOutputFormats.zip(realInputFormats).foreach {
           case (o, i) =>
             Utils.copyMaskAndScales(o, i)
             reorderManager.register(o, i)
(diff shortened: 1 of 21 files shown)


zhangxiaoli73:
jenkins pass: http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/1118/

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnGraph.scala
```diff
@@ -409,30 +422,66 @@ class DnnGraph(
...
+    var heapFormat: Int = -1
+    inputs.foreach(m => {
+      if (m.shape.length == 4) {
+        return inputs(0).layout
+      }
+    })
+
+    @inline
+    def transferFormat(format: DataFormat): Int = {
+      if (format == DataFormat.NHWC) Memory.Format.nhwc else Memory.Format.nchw
```
i8run: Does it mean we have no `ntc` and `tnc`.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/intermediate/IRToDnn.scala
```diff
@@ -115,19 +115,54 @@ private[bigdl] class IRToDnn extends ConvertBase[IRElement[Float], Module[Float]
...
+    } else {
+      // special process for NHWC
+      require(t.nGroup == 1, "Only support nGroup is 1 for NHWC")
+      val layer = ReflectionUtils.reflectFromIR(node, Class.forName(prefix + "SpatialConvolution"))
+      val p = layer.parameters()
+      val weight = p._1(0)
+      val gradWeight = p._2(0)
+
+      val weightSize = weight.size() // for nchw
+      val newSize = Array(weightSize(2), weightSize(3), weightSize(1), weightSize(0))// for nhwc
```
i8run: If the data format is `NHWC`,  the weight shape of conv is also `iohw`. Is it right ?

spark/dl/src/test/scala/com/intel/analytics/bigdl/utils/intermediate/IRconvertSpec.scala
```diff
@@ -91,6 +92,144 @@ class IRconvertSpec extends BigDLSpecHelper {
...
+//    val out1 = blas.forward(input)
+//    val out2 = dnn.forward(input)
+//    Equivalent.nearequals(out1.toTensor[Float], out2.toTensor[Float], 1e-5) should be (true)
+//
+//    val gradOutput = Tensor[Float]().resizeAs(out1.toTensor[Float])
+//
+//    val gradInputBlas = blas.backward(input, gradOutput)
+//    val gradInputDnn = dnn.backward(input, gradOutput)
+//    Equivalent.nearequals(gradInputDnn.toTensor, gradInputBlas.toTensor, 1e-4) should be (true)
+//    System.clearProperty("bigdl.engineType")
```
i8run: If it's not used, please delete the comments.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/mkldnn/DnnGraph.scala
zhangxiaoli73: For 3 dims, Blas layer only supports `ntc` , so it's no need to distinguish `ntc` and `tnc`.
But for 4 dims, there are two formats `nchw` and `nhwc` in some layers, so I have to iterate through all layers to find the real format.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/intermediate/IRToDnn.scala
zhangxiaoli73: Yes, weight shape of blas is `hwio`. But for dnn conv, it can still be `oihw` as input and weight format are `any`, and `reorder` will be done.

spark/dl/src/test/scala/com/intel/analytics/bigdl/utils/intermediate/IRconvertSpec.scala
zhangxiaoli73: those test codes will be opened after mengceng's code merged.

zhangxiaoli73:
jenkins pass:
http://mlt-bddl01.sh.intel.com:8080/job/BigDL-PRVN-lite/1124/

PR #2992, opened 2020-01-15
feat: add distri optimizer v2

i8run:

Refactor implementation of DistriOptimizer.

+ Add `Context`. User can set the replica model nums or override it to handle user defined pretrain, train, postTrain.
+ Move constant string values to object attributes. If possible, we should refactor other classes to fit the change.
+ Reuse `AllReduceParameter` if possible.

## How was this patch tested?

Unit tests and Jenkins.


i8run:
With the new spec the unit tests have passed on new Jenkins.

PR #3001, opened 2020-05-19
Remove final for AbstractModule

Le-Zheng:
remove final for cloneModule( ) and save( )
To fix https://github.com/intel-analytics/BigDL/issues/3000

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
@@ -500,7 +500,7 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag, 
    * Clone the model
    * @return
    */
-  final def cloneModule(): this.type = {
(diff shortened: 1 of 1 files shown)


Le-Zheng:
PR validation http://10.239.47.210:18080/job/BigDL-PRVN-lite/1132/

jason-dai:
Do we need to remove final for both `cloneModule` and `save`, or `save` only?

Le-Zheng:
Shall we remove all the `final` better? That will avoid override restrictions in the future.

jason-dai:
LGTM

jason-dai:
> Shall we remove all the `final` better? That will avoid override restrictions in the future.

I don't think you should remove all `final`

Le-Zheng:
> > Shall we remove all the `final` better? That will avoid override restrictions in the future.
> 
> I don't think you should remove all `final`

Ok. Currently, we only need to remove `final` for `save( ) `  to support TFTrainingHelper checkpoint.  Update.

Le-Zheng:
Tests passed on Jenkins http://10.239.47.210:18080/job/BigDL-PRVN-lite/1135/

Issue #2998, opened 2020-05-18
Fix callBigDLFunc 


jason-dai:
Similar to https://github.com/intel-analytics/analytics-zoo/pull/1793

PR #3002, opened 2020-05-19
fix callBigDLFunc

Le-Zheng:
To fix https://github.com/intel-analytics/BigDL/issues/2998


jason-dai:
@qiuxin2012 please take a look.

Le-Zheng:
Passed related tests on Jenkins http://10.239.47.210:18080/job/BigDL-PRVN-lite/1133/

Issue #2999, opened 2020-05-18
Enable DistriOptimizerV2


jason-dai:
- [x] Create `DistriOptimizerV2` in factory method (per user-specified argument)
- [x] Enabled `DistriOptimizerV2` in Scala and Python examples (per user-specified argument)
- [x] Test performance and convergence using `DistriOptimizerV2`

qiuxin2012:
For the factory method, shall we use a DistriOptimizerFactory like this [ModelBroadcastFactory](https://github.com/intel-analytics/BigDL/blob/master/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcastFactory.scala) in [ModelBroadcast](https://github.com/intel-analytics/BigDL/blob/4fe75f5f5f20cfeafde46afd4f1dbafbee4c540f/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala#L75). And set the default value to DistriOptimizerV2. @jason-dai

jason-dai:
> For the factory method, shall we use a DistriOptimizerFactory like this [ModelBroadcastFactory](https://github.com/intel-analytics/BigDL/blob/master/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcastFactory.scala) in [ModelBroadcast](https://github.com/intel-analytics/BigDL/blob/4fe75f5f5f20cfeafde46afd4f1dbafbee4c540f/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala#L75). And set the default value to DistriOptimizerV2. @jason-dai

I was referring to `Optimizer.apply`; we can add an argument (e.g., `V2`) which default to false.

qiuxin2012:
> > For the factory method, shall we use a DistriOptimizerFactory like this [ModelBroadcastFactory](https://github.com/intel-analytics/BigDL/blob/master/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcastFactory.scala) in [ModelBroadcast](https://github.com/intel-analytics/BigDL/blob/4fe75f5f5f20cfeafde46afd4f1dbafbee4c540f/spark/dl/src/main/scala/com/intel/analytics/bigdl/models/utils/ModelBroadcast.scala#L75). And set the default value to DistriOptimizerV2. @jason-dai
> 
[...]

PR #3003, opened 2020-05-20
DistriOptimizerV2 argument

Le-Zheng:

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
@@ -37,6 +37,14 @@ sealed trait EngineType
 case object MklBlas extends EngineType
 case object MklDnn extends EngineType
 
+/**
+ * define optimizer version trait
+ */
+sealed trait OptimizerVersion
+
+case object Optimizer1 extends OptimizerVersion
+case object Optimizer2 extends OptimizerVersion
+
 
 object Engine {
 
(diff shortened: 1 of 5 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -122,6 +122,7 @@ object Engine {
   private val singletonCounter = new AtomicBoolean()
   private var physicalCoreNumber = -1
   private var nodeNum: Int = -1
+  private var optimizerVersion: Boolean = false
```
qiuxin2012, line 125: You can copy some code from `engineType`

Le-Zheng:
To fix https://github.com/intel-analytics/BigDL/issues/2999

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/Engine.scala
```diff
@@ -37,6 +37,14 @@ sealed trait EngineType
 case object MklBlas extends EngineType
 case object MklDnn extends EngineType
 
+/**
+ * define optimizer version trait
+ */
+sealed trait OptimizerVersion
+
+case object Optimizer1 extends OptimizerVersion
+case object Optimizer2 extends OptimizerVersion
```
jason-dai, line 46: change to OptimizerV1 and OptimizerV2

jason-dai:
> To fix #2999

This only fixes the first task in #2999

Le-Zheng:
Jenkins related tests passed http://10.239.47.210:18080/job/BigDL-PRVN-lite/1143/

PR #3007, opened 2020-05-29
test examples by distrioptimizerv2

Le-Zheng:
## changes were proposed in this pull request:

test examples by passing `--optimizerVersion "optimizerv2"`


jason-dai:
We will also need to support python examples, maybe in a separate PR

Le-Zheng:
> We will also need to support python examples, maybe in a separate PR

Yes, python examples are in https://github.com/intel-analytics/BigDL/pull/3008

Issue #3005, opened 2020-05-21
Deprecate nn.keras


jason-dai:

Le-Zheng:
Shall we deprecate all of the nn.keras layers?

jason-dai:
We want to deprecate the `nn.kera` package

PR #3013, opened 2020-06-09
deprecate nn.keras

Le-Zheng:

deprecate nn.keras
1. Python: add deprecate notes for layers
2. Scala: deprecate layers/api which is not extended by ZOO

## How was this patch tested?

will be tested on Jenkins

To fix https://github.com/intel-analytics/BigDL/issues/3005

### DIFF
--- a/pyspark/bigdl/nn/keras/topology.py
+++ b/pyspark/bigdl/nn/keras/topology.py
@@ -211,7 +211,7 @@ class Sequential(KerasModel):
     """
     def __init__(self, jvalue=None, **kwargs):
         warnings.warn("bigdl.nn.keras is deprecated in 0.11. "
-                      "Recommend to use Analytics Zoo's Keras API")
+                      "Recommend to use Analytics Zoo's Keras API.")
         super(Sequential, self).__init__(jvalue, **kwargs)
 
     @staticmethod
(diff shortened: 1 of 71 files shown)


Le-Zheng:
passed Jenkins test: http://10.239.47.210:18080/job/BigDL-PRVN-lite/1149/

Le-Zheng:
There is no cleverer way but to deprecate each class as `nn.kreas` is a namespace.

pyspark/bigdl/nn/keras/topology.py
```diff
@@ -200,6 +204,9 @@ class Sequential(KerasModel):
 
     >>> sequential = Sequential(name="seq1")
     creating: createKerasSequential
+
+    .. note:: `bigdl.nn.keras` is deprecated in 0.11.
+    This will be removed in future releases.
     """
     def __init__(self, jvalue=None, **kwargs):
```
hkvision, line 211: Do we need to add a warning in init?

jason-dai: In general no - Analytics Zoo will also used these classes, yes?

hkvision: No. For the Python side, Analytics Zoo's all keras layers are newly wrapped and have no relationship with bigdl's keras layers. The dependencies only exist on Scala side.

Le-Zheng: Done. On the scala side, only deprecate Sequential, Model, and Input class. On the python side, add a warning for the Sequential and Model class.

PR #3017, opened 2020-06-19
fix throughput

Le-Zheng:

fix throughput time count (+ count stage time)
do batch jobs by multi-thread `cache.localModels`

## How was this patch tested?

manul test + Jenkins test

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizerV2.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizerV2.scala
@@ -35,6 +35,8 @@ import org.apache.spark.rdd.RDD
 import org.apache.spark.{SparkContext, TaskContext}
 
 import scala.collection.mutable
+import scala.collection.mutable.ArrayBuffer
+import scala.concurrent.Future
 import scala.reflect.ClassTag
 
 object DistriOptimizerV2 extends AbstractOptimizer {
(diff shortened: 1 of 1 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/optim/DistriOptimizerV2.scala
```diff
@@ -202,14 +206,13 @@ object DistriOptimizerV2 extends AbstractOptimizer {
...
-
+      }.reduce(_ + _)
+    
     parameterSync(lossSum.value, successModels, cacheOfMaster, models, context)
-
+    val endTime = System.nanoTime() - startTime
+    
     driverStatesUpdate(cacheOfMaster, recordsNum.value,
-      context, trainingTrace, metrics)
+      context, trainingTrace, endTime, metrics)
```
qiuxin2012, line 215: why change the time counting?

Le-Zheng: each iterationTakes should be reduce() + parameterSync()
previous `trainingTrace.traceIteration()` can only hold reduce() time.

qiuxin2012: Can we add a new metric for parameter sync?

Le-Zheng: Done. Put `reduce() + parameterSync()` in `trainingTrace.traceIteration()` to count full time iteration takes.

Le-Zheng:
Jenkins test passed http://10.239.47.210:18080/view/PR/job/BigDL-PRVN-lite/1155/ + http://10.239.47.210:18080/job/BigDL-PRVN-keras-py36/1175/

PR #3022, opened 2020-07-01
fix style check

Le-Zheng:
fix scala style check


Le-Zheng:
Jenkins passed : http://10.239.47.210:18080/job/BigDL-PRVN-StyleCheck/1183/

PR #3034, opened 2020-07-19
fix KerasLayer new parameters()

Le-Zheng:

should call the existed parameters( ) in extended Container

## How was this patch tested?

Jenkins


jason-dai:
Need to add unit tests for the fix

PR #3054, opened 2020-09-04
[WIP] spark 3.0

Le-Zheng:

Support for Spark 3.0
1. Pom: add Spark 3.0 in pom.xml, and update some related dependencies' version
2. Update Accumulator 
3. Fix match case exception, DistriOptimizerV2 serializable error, etc
## How was this patch tested?

Jenkins

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -112,7 +112,7 @@
         <scala.major.version>2.10</scala.major.version>
         <scala.version>2.10.7</scala.version>
         <scala.macros.version>2.1.0</scala.macros.version>
-        <scalatest.version>2.2.4</scalatest.version>
+        <scalatest.version>3.2.0</scalatest.version>
 
         <!-- platform encoding override -->
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
@@ -174,7 +174,7 @@
         <scoverage.plugin.version>1.1.1</scoverage.plugin.version>
         <spark-version.project>1.5-plus</spark-version.project>
         <spark.version>1.5.1</spark.version>
-        <breeze.version>0.11.2</breeze.version>
+        <breeze.version>0.13.2</breeze.version>
         <spark-scope>provided</spark-scope>
 
         <bigdl-core-all-scope>compile</bigdl-core-all-scope>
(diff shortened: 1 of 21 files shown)


pom.xml
```diff
@@ -544,7 +544,7 @@
             <id>spark_2.x</id>
             <properties>
                 <spark-version.project>2.0</spark-version.project>
-                <spark.version>2.0.0</spark.version>
+                <spark.version>2.4.0</spark.version>
```
qiuxin2012, line 547: We can use 2.4.3, the same with analytics-zoo

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RegionProposal.scala
```diff
@@ -143,9 +143,7 @@ class RegionProposal(
       val postNmsTopN = if (this.isTraining()) min(postNmsTopNTrain, bboxNumber)
       else min(postNmsTopNTest, bboxNumber)
 
-      if (output.getOrElse(b, null) == null) {
```
qiuxin2012, line 146: why delete this? output the output this a table, we can reuse the tensors in the output.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Pooler.scala
```diff
@@ -121,7 +121,7 @@ class Pooler[T: ClassTag] (
       val num_rois = rois.size(1)
       totalNum += num_rois
 
-      if (out.getOrElse(i + 1, null) == null) out(i + 1) = Tensor[T]()
```
qiuxin2012, line 124: why delete this? output the output this a table, we can reuse the tensors in the output.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/RegionProposal.scala
Le-Zheng: Here is a warning of "comparing values of types Null and Null using `==' will always yield true "

qiuxin2012: Use output.contains(b) instead

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -1064,6 +1064,7 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
           require(copiedModuleParamTable.get(name) != None, s"cloned module should have for $name")
           setLayerWeightAndBias(params,
             copiedModuleParamTable.get(name).get.asInstanceOf[Table], deepCopy)
+        case _=>
```
qiuxin2012, line 1067: you should throw an exception here, or mismatch will do nothing.

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/abstractnn/AbstractModule.scala
```diff
@@ -1125,6 +1126,7 @@ abstract class AbstractModule[A <: Activity: ClassTag, B <: Activity: ClassTag,
         } else {
           if (matchAll) new Exception(s"module $name cannot find corresponding weight bias")
         }
+      case _=>
```
qiuxin2012, line 1129: you should throw an exception here, or mismatch will do nothing.

spark/dl/src/main/scala/com/intel/analytics/bigdl/utils/caffe/CaffeLoader.scala
```diff
@@ -259,6 +259,7 @@ class CaffeLoader[T: ClassTag](prototxtPath: String, modelPath: String,
     parameterTable.foreach {
       case (name: String, params: Table) =>
         copyParameter(name, params)
+      case _=>
```
qiuxin2012, line 262: you should throw an exception here, or mismatch will do nothing.

spark/dl/src/test/scala/com/intel/analytics/bigdl/integration/torch/LSTMPeepholeSpec.scala
```diff
@@ -668,6 +668,7 @@ class LSTMPeepholeSpec  extends TorchRNNSpec {
         assert(abs(v1 - v2) <= 1e-8)
         v1
       })
+    case _=>
```
qiuxin2012, line 671: maybe you can delete `case ((key: Int, value: Tensor[_])) =>`, just use `asInstanceOf`

spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/Pooler.scala
```diff
@@ -121,7 +121,7 @@ class Pooler[T: ClassTag] (
       val num_rois = rois.size(1)
       totalNum += num_rois
 
-      if (out.contains(i + 1)) out(i + 1) = Tensor[T]()
+      if (!out.contains(i + 1)) out(i + 1) = Tensor[T]()
```
qiuxin2012, line 124: why?

Le-Zheng:
Jenkins passed Spark 2 and Spark 3 http://10.239.47.210:18080/job/BigDL-PRVN-lite/ and Integration test on yarn http://10.239.47.210:18080/view/PR/job/BigDL-PR-Integration-Examples-8/7/

PR #3065, opened 2020-10-27
support jdk11 

Le-Zheng:

Support JDK 11
Split spark 2.x artifact to spark2.x and spark2.4+. Make it as a similar pom structure as ZOO, so that we can use scala 2.12 in Spark2.4

## How was this patch tested?

Jenkins


Le-Zheng:
remind to use oracle jdk11
Currently, openjdk11 has a problem with maven-enforcer-plugin
(maven-enforcer-plugin:1.3.1:enforce failed: An API incompatibility was encountered while executing org.apache.maven.plugins:maven-enforcer-plugin:1.3.1:enforce: java.lang.ExceptionInInitializerError: null)

PR #3068, opened 2020-10-30
Update scala maven plugin

Le-Zheng:

Update scala/maven plugin to support jdk11. 

1.  `./make-dist.sh` is default-compile spark2.4.3 with jdk8. 
2. To compile specified Spark2.x with scala 2.11 and jdk8 is using like: `./make-dist.sh -P spark_2.x -Dspark.version=2.3.1`
3. To compile Spark3.x with scala 2.12 and jdk8: `./make-dist.sh -P spark_3.x`

`mvn` is used similarly as `./make-dist.sh`, e.g. mvn clean package -DskipTests -P spark_2.x

If we would like to compile with jdk 11, it requires to add below in pom.xml
```
<java.version>11</java.version>
<javac.version>11</javac.version>
<maven.compiler.source>1.6</maven.compiler.source>
<maven.compiler.target>1.6</maven.compiler.target>
```

## How was this patch tested?

Jenkins

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -107,8 +107,8 @@
         <findbugs.version>3.0.0</findbugs.version>
 
         <!-- define the Java language version used by the compiler -->
-        <java.version>1.8</java.version>
-        <javac.version>1.8</javac.version>
+        <java.version>1.7</java.version>
+        <javac.version>1.7</javac.version>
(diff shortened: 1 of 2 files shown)


Le-Zheng:
Jenkins passed.

PR #3072, opened 2020-11-13
squeeze target dimension (corner case) in ClassNLLCriterion

Le-Zheng:
## How was this patch tested?

jenkins

### DIFF
--- a/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ClassNLLCriterion.scala
+++ b/spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ClassNLLCriterion.scala
@@ -93,9 +93,9 @@ class ClassNLLCriterion[@specialized(Float, Double) T: ClassTag]
        s"input dim(${input.dim()})")
     val nClasses = input.size(input.dim())
     if (input.dim() == 1) {
-      require(input.dim() == target.dim(),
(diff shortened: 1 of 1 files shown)


spark/dl/src/main/scala/com/intel/analytics/bigdl/nn/ClassNLLCriterion.scala
```diff
@@ -93,6 +93,9 @@ class ClassNLLCriterion[@specialized(Float, Double) T: ClassTag]
        s"input dim(${input.dim()})")
     val nClasses = input.size(input.dim())
     if (input.dim() == 1) {
+      if (target.dim() == 2 && target.size(1) == 1) {
+        target.squeeze()
```
qiuxin2012, line 97: can we copy to a new tensor? squeeze will change the target' size, sometimes it will lead to some error.

Le-Zheng: Is it right?
```
if (target.dim() == 2 && target.size(1) == 1) {
    val newTarget = target
    newTarget.squeeze()
    val curTarget = ev.toType[Int](newTarget.valueAt(dl = 1))
}
else {
    val curTarget = ev.toType[Int](target.valueAt(dl = 1))
}
```

PR #3088, opened 2020-12-01
Fix SerializationUtils clone issue of QuantizedTensor

jenniew:
Fix SerializationUtils clone issue of QuantizedTensor.


jason-dai:
Do we have a unit test for the issue?

jenniew:
OptimPredictorShutdownSpec can test broadcasting of quantized tensor, but we only have local mode test.

jason-dai:
> OptimPredictorShutdownSpec can test broadcasting of quantized tensor, but we only have local mode test.

Shall we add an integration test for "pseudo" cluster testing then? @Le-Zheng

Le-Zheng:
We only have several UTs running on yarn. We may need to add this UT in the [integration test  scripts](https://github.com/intel-analytics/BigDL/blob/master/spark/dl/src/test/integration-test.robot#L93).

jason-dai:
> We only have several UTs running on yarn. We may need to add this UT in the [integration test scripts](https://github.com/intel-analytics/BigDL/blob/master/spark/dl/src/test/integration-test.robot#L93).

Yes, we need to add it.

PR #3111, opened 2021-03-23
fix callBigDLFunc return a Int while the true return value from java is a byte array.

qiuxin2012:

fix callBigDLFunc return a Int while the true return value from java is a byte array.

## How was this patch tested?

unit tests


qiuxin2012:
jenkins passed

Le-Zheng (approved):

PR #3129, opened 2021-05-25
make default DistriOptimizer as V2

Le-Zheng:

Switch default DistriOptimizer version as V2 in Engine.scala.

## How was this patch tested?

Jenkins


Le-Zheng:
jenkins passed: http://10.239.176.111:18080/job/BigDL-PRVN-lite/1211/

PR #3133, opened 2021-06-24
fix dlframe

Le-Zheng:


Le-Zheng:
jenkins passed: http://10.239.176.111:18080/job/BigDL-PRVN-lite/1212/

PR #3153, opened 2021-10-20
Update and combine pip release

hkvision:
TODO:
- Add combined script for mac
- Upload the whls at the end after all builds finish

All pip release scripts are unified in the same format and use only one script/jenkins job to release all packages.


hkvision:
http://10.112.231.51:18888/job/BigDL2.0-NB-Deploy-PYTHON-PyPI/10/

PR #3155, opened 2021-10-20
update dllib readme

Le-Zheng:


jason-dai:
LGTM.

BTW, maybe we should move `NNContext` to `bigdl.dllib` in a separate PR

Le-Zheng:
> LGTM.
> 
> BTW, maybe we should move `NNContext` to `bigdl.dllib` in a separate PR

Shall we move `bigdl.dllib.common.NNContext` to `bigdl.dllib.NNContext`?

PR #3166, opened 2021-10-21
Chronos: fix tsdataset bug when lookback > len(df)

TheaperDeng:
This bug is caused by a possible valid situation in autots.

a migration of PR5040 from AZ.


TheaperDeng:
pass:
http://10.112.231.51:18888/job/BigDL-Chronos-PR-Validation/31/

Issue #3171, opened 2021-10-18
Chronos: Embrace bigdl-nano in Chronos


TheaperDeng:
### Background

Currently `Chronos` rely on `orca.automl` for their pytorch model single node support. This will cause

- Future `chronos-inference` will rely on `orca` (and `dllib`) which make the whole package fat.
- Can not use `nano`'s speed up.
- pytorch-lightning Trainer will bring more feature to Chronos (such as efficient lr/bs tuning)

### Implementation Step

We will implement the whole plan mainly in 5 PRs

- [x] add a new `nano.pytorch.pl_base_model`, this model should add time series specific training step ...

- [x] add `chronos.metric` to replace `orca.automl.metric`. #3608 

- [x] change forecaster(distributed=False) path to bigdl-nano(pl) #3539 

- [x] change TSPipeline path to bigdl-nano(pl) #3609 

- [x] change Automodel (except .fit) to bigdl-nano(pl) #3745

TheaperDeng:
> 1. **distributed=false**: using `nano` (possibly with pseudo-distributed mode)
> 2. **distributed=true**: using `orca`

yes, that is the plan

PR #3181, opened 2021-10-21
Add pytorch-lightning decorator to nano

zhentaocc:
Resolved 1st task in  #3171.

- [x] Add a base class to convert pytorch model to pytorch-lightning.
- [x] Add `Trainer.compile` for user to use.
- [x] Add a unit tests.


python/nano/test/test_models_lightning_support.py
```diff
@@ -0,0 +1,57 @@
...
+
+def loss_creator(config):
+    return nn.CrossEntropyLoss()
+
+
+def optimizer_creator(model, config):
+    return getattr(torch.optim, config.get("optim", "Adam"))(model.parameters(), lr=config.get("lr", 0.001))
+
+
+@lightning_support.lightning_module(loss_creator, optimizer_creator, config)
```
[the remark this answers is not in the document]
TheaperDeng: **It seems to me that the `trainer.compile` has a better user experience** since you can not create a opt instance before the model instance is built, which leads to some config dict. 

Still both of the methods involve a abnormal usage (i.e. multiple-decorator and `trainer.compile`). We need to give thorough and detailed user guide and in-code-warning.

zhentaocc:
> I don't think we should implementation multiple interfaces for the same use case - it just confuses the users; instead, just provide one single interface for one use case.

Sure, we need to decide which usage is much more common and friendly to users. And at the same time onnxruntime support should be aligned if we choose to wrap a torch model instance instead of a torch class.
But in the other way, I implemented another function which can either be a decorator or a usual function on a torch model instance:
```python
@to_lightning(loss, torch.optim.Adam, lr=0.01)
class Net(nn.Module):
    pass
pl_model = Net()
```
or
```python
class Net(nn.Module):
    pass
pl_model = to_lightning(loss, torch.optim.Adam, lr=0.01(Net())
```
I suppose we may have more extensions coming on pytorchlightning model. Maybe we can wrap all these extensions into one so user don't have to decide which function to use:
```python
def composed(*decs):
    def deco(f):
        for dec in reversed(decs):
            f = dec(f)
        return f
    return deco

def preprocess(loss=None, optimizer=None, config=None, onnx=True):
    return composed(
        onnxruntime(onnx),
        to_lightning(loss, optimizer, **config)
    )
pl_model = preprocess(loss, torch.optim.Adam, {"lr": 0.01}, onnx=True)(model)
```
What's your thoughts on this, providing a unified interface or just choose one from it?

python/nano/test/test_models_lightning_support.py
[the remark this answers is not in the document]
jason-dai: > **It seems to me that the `trainer.compile` has a better user experience** since you can not create a opt instance before the model instance is built, which leads to some config dict in the decorator case.
> 
> Still the method involve a abnormal usage (i.e. `trainer.compile`). We need to give thorough and detailed user guide and in-code-warning.
> 
> btw, we don't need to implement `onnx=True` parameter in this PR since I have not merged the onnx PR. I will do it later after this PR has been merged.

What do you mean by abnormal usage? I think the point is
1) If the user simply has a PyTorch model, he or she can directly use it in `nano.pytorch.trainer` methods (if the default behavior of `fit`, `test`, `predict`, etc., works for him or her)

2) If the user needs more complex behavior (e.g., onnxruntime support), he or she needs to explictly convert it to pl_module, and we can provide an API based on either decorator or compile method.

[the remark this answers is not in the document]
TheaperDeng: > > **It seems to me that the `trainer.compile` has a better user experience** since you can not create a opt instance before the model instance is built, which leads to some config dict in the decorator case.
> > Still the method involve a abnormal usage (i.e. `trainer.compile`). We need to give thorough and detailed user guide and in-code-warning.
> > btw, we don't need to implement `onnx=True` parameter in this PR since I have not merged the onnx PR. I will do it later after this PR has been merged.
> 
> What do you mean by abnormal usage? I think the point is
> 
> 1. If the user simply has a PyTorch model, he or she can directly use it in `nano.pytorch.trainer` methods (if the default behavior of `fit`, `test`, `predict`, etc., works for him or her)
> 2. If the user needs more complex behavior (e.g., onnxruntime support), he or she needs to explictly convert it to pl_module, and we can provide an API based on either decorator or compile method.

I think we can just ask all users with a nn.module to do this:
```python
model = Net()
model_pl = trainer.compile(model, loss, opt, onnx=bool)
# then use model_pl (pl.lightningmodule) to do anything else
```
abnormal usage means that `trainer.compile` is not a easy name for users to come up with directly, we need examples, quickstarts and user guides to guide them.

jason-dai:
> > I don't think we should implementation multiple interfaces for the same use case - it just confuses the users; instead, just provide one single interface for one use case.
> 
> Sure, we need to decide which usage is much more common and friendly to users. And at the same time onnxruntime support should be aligned if we choose to wrap a torch model instance instead of a torch class. But in the other way, I implemented another function which can either be a decorator or a usual function on a torch model instance:
> 
> ```python
> @to_lightning(loss, torch.optim.Adam, lr=0.01)
> class Net(nn.Module):
>     pass
> pl_model = Net()
> ```
> 
> or
> 
> ```python
> class Net(nn.Module):
>     pass
> pl_model = to_lightning(loss, torch.optim.Adam, lr=0.01(Net())
> ```
> 
> I suppose we may have more extensions coming on pytorchlightning model. Maybe we can wrap all these extensions into one so user don't have to decide which function to use:
> 
> ```python
> def composed(*decs):
>     def deco(f):
>         for dec in reversed(decs):
>             f = dec(f)
>         return f
>     return deco
> 
> def preprocess(loss=None, optimizer=None, config=None, onnx=True):
>     return composed(
>         onnxruntime(onnx),
>         to_lightning(loss, optimizer, **config)
>     )
> pl_model = preprocess(loss, torch.optim.Adam, {"lr": 0.01}, onnx=True)(model)
> ```
> 
> We can also integrate this function into `trainer.compile()`. What's your thoughts on this, providing a unified interface or just choose one from it?

`pl_model = to_lightning(loss, torch.optim.Adam, lr=0.01(Net())` - is this correct? I think we want to have something similar to `ray.remote`, but support below cases:

* Class definition
  ```python
  Class Net(torch.nn.module):
      ...
  model = Net()
  ```

* Model instance
  ```python
  model = torch.vision.ResNet18()
  ```

* Model creator function
  ```python
  def model_creator():
     model = torch.vision.ResNet18()
     return model
  ```

Need a consistent API for all these use cases.

python/nano/test/test_models_lightning_support.py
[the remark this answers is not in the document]
jason-dai: It does not have to be `trainer.compile`; maybe something similar to `ray.distribtued`, such as `nano.pl_module`

zhentaocc:
Having discussed with @TheaperDeng @yangw1234 , currently we have 3 solutions:

1.  `trainer.compile(model, loss, optimizer)` and return a lightning module
2.  `trainer.compile(model, loss, optimizer)` and return a lightning module. With extra flag to track if `trainer.compile` is called or not, we forced user to use this function before `fit()`. 
3.  Bind model to trainer, convert pytorch model to lightning when trainer is initiated and no need to pass model to `fit()`.

We prefer option 2 to make the conversion as a fixed process and giving proper warning and error to inform users of the correct usage. Finally user need to do:
`create model --> create loss,optim --> pl_model=compile(model, loss, optim) -> fit(pl_model, ...)`
And this can cover all 3 use cases mentioned above.

 @jason-dai What's your thoughts?

jason-dai:
> Having discussed with @TheaperDeng @yangw1234 , currently we have 3 solutions:
> 
> 1. `trainer.compile(model, loss, optimizer)` and return a lightning module
> 2. `trainer.compile(model, loss, optimizer)` and return a lightning module. With extra flag to track if `trainer.compile` is called or not, we forced user to use this function before `fit()`.
> 3. Bind model to trainer, convert pytorch model to lightning when trainer is initiated and no need to pass model to `fit()`.
> 
> We prefer option 2 to make the conversion as a fixed process and giving proper warning and error to inform users of the correct usage. Finally user need to do: `create model --> create loss,optim --> pl_model=compile(model, loss, optim) -> fit(pl_model, ...)` And this can cover all 3 use cases mentioned above.
> 
> The name of this `trainer.compile(model, loss, optimizer)` is to be decided so that no confusion to users.
> 
> @jason-dai What's your thoughts?

I think we only need `compile` for PyTorch model; otherwise the user needs to change his or her PL code.

zhentaocc:
> > Having discussed with @TheaperDeng @yangw1234 , currently we have 3 solutions:
> > 
> > 1. `trainer.compile(model, loss, optimizer)` and return a lightning module
> > 2. `trainer.compile(model, loss, optimizer)` and return a lightning module. With extra flag to track if `trainer.compile` is called or not, we forced user to use this function before `fit()`.
> > 3. Bind model to trainer, convert pytorch model to lightning when trainer is initiated and no need to pass model to `fit()`.
> > 
> > We prefer option 2 to make the conversion as a fixed process and giving proper warning and error to inform users of the correct usage. Finally user need to do: `create model --> create loss,optim --> pl_model=compile(model, loss, optim) -> fit(pl_model, ...)` And this can cover all 3 use cases mentioned above.
> > The name of this `trainer.compile(model, loss, optimizer)` is to be decided so that no confusion to users.
> > @jason-dai What's your thoughts?
> 
> I think we only need `compile` for PyTorch model; otherwise the user needs to change his or her PL code.

The main reason why we want to make `compile` a necessary step is that considering adding other extensions on Lightning (e.g. onnxruntime support, maybe following we'll have other extensions to add), integrating all these to `compile` relieves users from calling many interfaces to use different extensions. Even though user has already created a lightning model, `compile` can still do nothing and return the model. 
```python
pl_model = trainer.compile(torch.vision.ResNet18(),  loss, optimizer, ...)
pl_model = trainer.compile(LightningModule(...)) # This does nothing and returns LightningModule(...)
```
Above are all legal for users to use so they don't have to change their lightning code. By this way, they can be warned to use `complie` if they need any extra extension in nano as well as converting to a pl model.

jason-dai:
> > > Having discussed with @TheaperDeng @yangw1234 , currently we have 3 solutions:
> > > 
> > > 1. `trainer.compile(model, loss, optimizer)` and return a lightning module
> > > 2. `trainer.compile(model, loss, optimizer)` and return a lightning module. With extra flag to track if `trainer.compile` is called or not, we forced user to use this function before `fit()`.
> > > 3. Bind model to trainer, convert pytorch model to lightning when trainer is initiated and no need to pass model to `fit()`.
> > > 
> > > We prefer option 2 to make the conversion as a fixed process and giving proper warning and error to inform users of the correct usage. Finally user need to do: `create model --> create loss,optim --> pl_model=compile(model, loss, optim) -> fit(pl_model, ...)` And this can cover all 3 use cases mentioned above.
> > > The name of this `trainer.compile(model, loss, optimizer)` is to be decided so that no confusion to users.
> > > @jason-dai What's your thoughts?
> > 
> > 
> > I think we only need `compile` for PyTorch model; otherwise the user needs to change his or her PL code.
> 
> The main reason why we want to make `compile` a necessary step is that considering adding other extensions on Lightning (e.g. onnxruntime support, maybe following we'll have other extensions to add), integrating all these to `compile` relieves users from calling many interfaces to use different extensions. Even though user has already created a lightning model, `compile` can still do nothing and return the model.
> 
> ```python
> pl_model = trainer.compile(torch.vision.ResNet18(),  loss, optimizer, ...)
> pl_model = trainer.compile(LightningModule(...)) # This does nothing and returns LightningModule(...)
> ```
> 
> Above are all legal for users to use so they don't have to change their lightning code. By this way, they can be warned to use `complie` if they need any extra extension in nano as well as converting to a pl model.

It's OK to use `compile` to add nano-specific extensions; however, the user should not be required to call `compile` is he or she just has standard PL code and wants to use `nano` for transparent acceleration.

zhentaocc:
As discussed above, make modifications:

1. Refactored `to_lightning` to class `LightningModuleFromTorch`
2. Added unit test to test `LightningModuleFromTorch`
3. Added static method `Trainer.compile(...)` since `self` is not used
4. Added unit test to test `Trainer.compile(...)`

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
```diff
@@ -104,3 +108,23 @@ def __init__(self, num_processes: int = 1,
 
             super().__init__(accelerator=accelerator,
                              plugins=[plugin], *args, **kwargs)
+
+    @staticmethod
+    def compile(model: nn.Module, loss: _Loss = None, optimizer: torch.optim = None):
+        """
+        Compile a pytorch model into a pytorch-lightning model and return it.
+        Args:
```
TheaperDeng, line 116: maybe write this in the format as https://github.com/intel-analytics/BigDL/blob/branch-2.0/python/chronos/src/bigdl/chronos/autots/tspipeline.py#L47-L60

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
```diff
@@ -104,3 +108,23 @@ def __init__(self, num_processes: int = 1,
 
             super().__init__(accelerator=accelerator,
                              plugins=[plugin], *args, **kwargs)
+
+    @staticmethod
```
TheaperDeng, line 112: I think we can make it a member function

python/nano/src/bigdl/nano/pytorch/lightning.py
```diff
@@ -0,0 +1,70 @@
...
+            optimizer: torch optimizer.
+
+        Returns: LightningModule Object
+        """
+        super().__init__()
+        self.copy(model)
+        self.loss = loss
+        self.optimizer = optimizer
+
+    def copy(self, torch_model):
```
TheaperDeng, line 39: make this an internal method `_copy`?

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
zhentaocc: Static method is more reasonable, I think, after all we are not using any Trainer instance in compile. And we can do either:
1.  `trainer.compile(model, loss, optimizer)` 
2.  `Trainer.compile(model, loss, optimizer)`

python/nano/src/bigdl/nano/pytorch/lightning.py
zhentaocc: Cool.

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
```diff
@@ -104,3 +108,24 @@ def __init__(self, num_processes: int = 1,
...
+    def compile(model: nn.Module, loss: _Loss = None, optimizer: torch.optim = None):
+        """
+        Compile a pytorch model into a pytorch-lightning model and return it.
```
jason-dai, line 115: the comment is incorrect

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
```diff
@@ -104,3 +108,24 @@ def __init__(self, num_processes: int = 1,
...
+
+        Returns: A LightningModule converted from model.
+
+        """
+        if isinstance(model, pl.LightningModule):
+            if loss:
+                model.loss = loss
+            if optimizer:
+                model.optimizer = optimizer
+            return model
```
jason-dai, line 129: Why do we want to support this? Can we guarantee this is correct if `configure_optimizers` returns a different optimizer?

zhentaocc: Correct. Added assertion to check if the inputs are correct.

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
zhentaocc: Applied your suggestion. Thanks.

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
zhentaocc: Please review modified docstring. Any further suggestions to properly describe this function?

zhentaocc:
Passed tests. It's ready now.

python/nano/src/bigdl/nano/pytorch/lightning.py
```diff
@@ -0,0 +1,66 @@
...
+        :param model:       Pytorch model to be converted.
+        :param loss:        A torch loss function.
+        :param optimizer:   A torch optimizer.
+        """
+        super().__init__()
+        self._copy(model)
+        self.loss = loss
+        self.optimizer = optimizer
+
+    def _copy(self, torch_model: nn.Module):
```
yangw1234, line 37: Is it safe to do this? 
Looks a little hacking to me. 
Should we stick to the public API pytorch-lightning provides?

jason-dai: Why not make `torch_model` a member variable?

zhentaocc: We have `load_state_dict` in `nn.Module`. So I'm concerned if we make it as a member variable user won't be able to load a pretrained torch model to pytorch-lightning. We can still assume that user has already loaded the parameters before creating this pl model. But I'm not sure if all other functions work fine.

python/nano/src/bigdl/nano/pytorch/lightning.py
```diff
@@ -0,0 +1,66 @@
...
+        for name, child in torch_model._modules.items():
+            setattr(self, name, child)
+        setattr(self, "forward", torch_model.forward)
+
+    def _forward(self, batch):
```
yangw1234, line 42: could you document why you override this method?

python/nano/src/bigdl/nano/pytorch/lightning.py
```diff
@@ -0,0 +1,66 @@
...
+        # Handle different numbers of input for various models
+        nargs = self.forward.__code__.co_argcount
+        return self.forward(*(batch[:nargs - 1]))
+
+    def training_step(self, batch, batch_idx):
+        x, y = batch
+        y_hat = self._forward(batch)
```
yangw1234, line 49: I am sure I understand this method. 
Could we just call `self.forward` in `self._forward` and pass `x` to it?

python/nano/src/bigdl/nano/pytorch/lightning.py
zhentaocc: Having checked Pytorch API doc, it seems only `load_state_dict` is influenced since the module dict will be changed if make it a member variable. I think we have 3 options:

1.  Make pytorch model as a member variable and assume user has already loaded the model.
2.  Make pytorch model as a member variable and override `load_state_dict`.
3.  Move `_copy` into `__init__`. (Not sure if this seems not that hacking.)

python/nano/src/bigdl/nano/pytorch/lightning.py
zhentaocc: We have a few forward  with multiple inputs instead of one. Typically in Seq2Seq, during training the input can be (x, y) to use y as a prior, but input (x) during inference. Another case is that input is (x1,x2,x3,...,y). So in `self._forward` I'm trying to resolve this issue by getting the number of arguments. Another way is call `self.forward(batch)` and we have to handle some extra input like y when defining forward function, for example, using `forward(x, *args)`.

python/nano/src/bigdl/nano/pytorch/lightning.py
jason-dai: We can do #2.

Another option: can we use `__getattr__`, similar to https://github.com/intel-analytics/BigDL/blob/branch-2.0/python/friesian/src/bigdl/friesian/feature/table.py#L663?

python/nano/src/bigdl/nano/pytorch/lightning.py
yangw1234: for teacher forcing, can we let `Seq2Seq` override `train_step`. @TheaperDeng

python/nano/src/bigdl/nano/pytorch/lightning.py
zhentaocc: No idea how `getattr` works in this case. Current implementation will try to load self.model firstly and then try to load self if failed.

```python
try:
    self.model.load_state_dict(state_dict)
except RuntimeError:
    super().load_state_dict(state_dict)
```
And I added two unit tests to make sure they work in both cases.

jason-dai: > No idea how `getattr` works in this case. Current implementation will try to load self.model firstly and then try to load self if failed.
> 
> ```python
> try:
>     self.model.load_state_dict(state_dict)
> except RuntimeError:
>     super().load_state_dict(state_dict)
> ```
> 
> And I added two unit tests to make sure they work in both cases.

Exactly what `Table` does:
``` 
   def __getattr__(self, name):
        return self.model.__getattr__(name)
```

zhentaocc: I understand now. If we use this method, it works when I load state_dict from a torch model checkpoint. 
Let's say keys are "conv.weight", "conv.bias", and keys for lightning module are "model.conv.weight", "model.conv.bias", it works.
Otherwise if I stored a pytorch-lightning model, its keys are actually "model.conv.weight", "model.conv.bias", then it failed. 
If we just want to resolve `load_state_dict`, possibly:

1.  We accept that user can load state dict from either a torch model or pl model.
Simplest way it to try loading into `self.model` and then `self`, otherwise raise error.

2.  We assume user always load state dict from torch model before calling `compile` or `LightningModuleFromTorch`.
Then we don't have to do anything.

zhentaocc: > I understand now. If we use this method, it works when I load state_dict from a torch model checkpoint. Let's say keys are `conv.weight`, `conv.bias`, and keys for lightning module are `model.conv.weight`, `model.conv.bias`, it works. Otherwise if I stored a pytorch-lightning model, its keys are actually `model.conv.weight`, `model.conv.bias`, then it failed. If we just want to resolve `load_state_dict`, possibly:
> 
> 1. We accept that user can load state dict from either a torch model or pl model.
>    Simplest way it to try loading into `self.model` and then `self`, otherwise raise error.
> 2. We assume user always load state dict from torch model before calling `compile` or `LightningModuleFromTorch`.
>    Then we don't have to do anything.

Main difference is keys for a pl model have 'model.' prefix.

jason-dai: > I understand now. If we use this method, it works when I load state_dict from a torch model checkpoint. Let's say keys are `conv.weight`, `conv.bias`, and keys for lightning module are `model.conv.weight`, `model.conv.bias`, it works. Otherwise if I stored a pytorch-lightning model, its keys are actually `model.conv.weight`, `model.conv.bias`, then it failed. If we just want to resolve `load_state_dict`, possibly:
> 
> 1. We accept that user can load state dict from either a torch model or pl model.
>    Simplest way it to try loading into `self.model` and then `self`, otherwise raise error.
> 2. We assume user always load state dict from torch model before calling `compile` or `LightningModuleFromTorch`.
>    Then we don't have to do anything.

I think the point is:
1) We can override whatever methods (such as `load_state_dict`, `forward`, etc.) that we would like to in `LightningModuleFromTorch`
2) For methods defined in the original python module that we do not override, we may use `__getattr__` so that these methods can still work (if needed at all)

yangw1234:
LGTM

python/nano/src/bigdl/nano/pytorch/lightning.py
zhentaocc: Got it. We discussed and preferred 2nd option.
Everything looks good now.

zhentaocc:
Jenkins passed.
http://10.112.231.51:18888/job/ZOO-PR-NanoTests/68/

yangw1234:
http://10.112.231.51:18888/job/ZOO-PR-NanoTests/74/

PR #3224, opened 2021-10-22
Chronos: Fix wrong error message format in chronos

liangs6212:


liangs6212:
http://10.112.231.51:18888/job/BigDL-Chronos-PR-Validation/20/

PR #3346, opened 2021-10-29
Add spark submit to pytorch tf tf2

piaolaidelangman:
Add spark submit to pytorch tf tf2.

### DIFF
--- a/python/orca/example/ray_on_spark/parameter_server/async_parameter_server.py
+++ b/python/orca/example/ray_on_spark/parameter_server/async_parameter_server.py
@@ -24,7 +24,7 @@ import argparse
 import os
 import time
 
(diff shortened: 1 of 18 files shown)


Le-Zheng:
LGTM

python/orca/example/ray_on_spark/parameter_server/async_parameter_server.py
```diff
@@ -24,7 +24,7 @@
 import os
 import time
 
-from python.orca.example.ray_on_spark.parameter_server import model
+from bigdl.example.ray_on_spark.parameter_server import model
```
Le-Zheng, line 27: http://10.112.231.51:18888/job/ZOO-PR-BigDL-Python-ExampleTests-py36-ray/93/console  No module named 'bigdl.example'

PR #3367, opened 2021-11-02
[WIP] Tf2estimator on Pyspark backend big weights support

jenniew:
Tf2estimator on Pyspark backend big weight support


python/orca/src/bigdl/orca/learn/tf2/spark_runner.py
```diff
@@ -344,9 +347,11 @@ def step(self, data_creator, epochs=1, batch_size=32, verbose=1,
         else:
             stats = {k: v[-1] for k, v in history.history.items()}
         if self.rank == 0:
-            # if model_dir is not None:
-            #     model.save_weights(model_dir)
-            return ([weights, stats])
+            if self.model_dir is not None:
```
yangw1234, line 350: how about just using `torch.save(model.state_dict(), PATH)`

jenniew: torch.save can support hdfs and s3?

yangw1234: I meaning inside `save_pkl` and `load_pkl`?

python/orca/src/bigdl/orca/learn/tf2/pyspark_estimator.py
```diff
@@ -151,34 +152,36 @@ def transform_func(iter, init_param, param):
                     param["validation_data_creator"] = make_data_creator(valid_list)
                     return SparkRunner(**init_param).step(**param)
 
-                res = data.zip(validation_data).rdd.repartition(self.num_workers).barrier() \
+                self.states_rdd = data.zip(validation_data).rdd.repartition(self.num_workers).barrier() \
```
yangw1234: I think there is an assumption of this line that `data` and `validataion_data` has the same number of partitions and the same number of elements on each partition.
I think the assumption may not hold if `OrcaContext._shard_size` is not None.

python/orca/src/bigdl/orca/learn/tf2/spark_runner.py
yangw1234: Sorry. This is TF. Orz

jenniew:
Jenkins pass: http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-Orca-PR-Validation/92/

PR #3368, opened 2021-11-02
add all-in-one python zip

Le-Zheng:


Le-Zheng:
compiled zip: scala/target/bigdl-parent-spark_2.4.6-0.14.0-SNAPSHOT-python-api.zip

glorysdj (approved):
LGTM

PR #3381, opened 2021-11-03
[BIGDL NANO] fix pl version to 1.4.2

yangw1234:


yangw1234:
nano pr is still unstable, first merge it

PR #3385, opened 2021-11-03
add mvn parms for spark3 release script

Le-Zheng:


Le-Zheng:
`bash python/dev/release_default_linux_spark312.sh false false -Ddata-store-url=http://10.239.45.10:8081/repository/raw`
pass on local.

PR #3387, opened 2021-11-03
add simpleMLP apps

dding3:


apps/SimpleMlp/README.md
```diff
@@ -0,0 +1,27 @@
+# SimpleMlp
+
+This is an example about how to use [BigDL](https://github.com/intel-analytics/BigDL) to build your own deep learning application. You can use both maven or sbt to create your project and add bigdl-dllib to your dependency, see pom.xml and build.sbt for details.
```
jason-dai, line 3: an Scala example

apps/SimpleMlp/src/main/scala/com/intel/analytics/zoo/tutorial/SimpleMlp.scala
```diff
@@ -0,0 +1,53 @@
...
+      // make up some data
+      val data = sc.range(0, recordSize, 1).map { _ =>
+        val featureTensor = Tensor[Float](dimInput)
+        featureTensor.apply1(_ => scala.util.Random.nextFloat())
+        val labelTensor = Tensor[Float](1)
+        labelTensor(Array(1)) = Math.round(scala.util.Random.nextFloat())
+        Sample[Float](featureTensor, labelTensor)
+      }
+
+      val model = Sequential[Float]()
```
jason-dai, line 38: There is no need to add `[Float]` in each layer; I think you just need to `import com.intel.analytics.bigdl.numeric.NumericFloat`; e.g., see https://github.com/intel-analytics/BigDL/blob/branch-2.0/scala/dllib/src/main/scala/com/intel/analytics/bigdl/dllib/models/lenet/LeNet5.scala

PR #3391, opened 2021-11-04
Chronos: Optimize AppTest operation

liangs6212:
Related to Chronos app test takes too long in https://github.com/intel-analytics/arda-docker/issues/417


yushan111:
Chronos APP test time has been reduced to 21min 
http://10.112.231.51:18888/job/BigDL-PRVN-Chronos-Python-AppTests-py36/17/

PR #3401, opened 2021-11-04
remove [Float] in each layer in SimpleMLP apps

dding3:


jason-dai:
Can we merge this PR?

PR #3418, opened 2021-11-08
fix init_spark_on_k8s not find bigdl jar

Le-Zheng:
fix` Path /opt/bigdl-0.14.0-SNAPSHOT/jars/* specified BIGDL_CLASSPATH does not exist`

### DIFF
--- a/docker/bigdl-k8s/Dockerfile
+++ b/docker/bigdl-k8s/Dockerfile
@@ -98,6 +98,7 @@ ENV RUNTIME_DRIVER_MEMORY               10g
 ENV RUNTIME_PERSISTENT_VOLUME_CLAIM     myvolumeclaim
 ENV SPARK_HOME                          /opt/spark
 ENV JAVA_HOME                           /opt/jdk
+ENV BIGDL_CLASSPATH                     ${BIGDL_HOME}/jars/*
 ENV PYTHONPATH                          ${BIGDL_HOME}/python/bigdl-spark_${SPARK_VERSION}-${BIGDL_VERSION}-python-api.zip:${SPARK_HOME}/python/lib/pyspark.zip:${SPARK_HOME}/python/lib/py4j-*.zip:/opt/models/research/slim
 ENV PATH ${BIGDL_HOME}/bin/cluster-serving:${JAVA_HOME}/bin:${PATH}
 ENV TINI_VERSION                        ${TINI_VERSION}
--- a/python/dllib/src/bigdl/dllib/utils/zoo_engine.py
+++ b/python/dllib/src/bigdl/dllib/utils/zoo_engine.py
@@ -111,9 +111,11 @@ def get_analytics_zoo_classpath():
     Get and return the jar path for analytics-zoo if exists.
     """
     if os.getenv("BIGDL_CLASSPATH"):
-        # for path in os.getenv("BIGDL_CLASSPATH").split(":"):
-        #     if not os.path.exists(path):
-        #         raise ValueError("Path {} specified BIGDL_CLASSPATH does not exist.".format(path))
+        for path in os.getenv("BIGDL_CLASSPATH").split(":"):
+            for pa in path.split("*"):
+                if not os.path.exists(pa):
+                    raise ValueError("Path {} specified BIGDL_CLASSPATH does not exist."
+                                     .format(path))
         return os.environ["BIGDL_CLASSPATH"]
     jar_dir = os.path.abspath(__file__ + "/../../../")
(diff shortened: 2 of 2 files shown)


python/dllib/src/bigdl/dllib/utils/zoo_engine.py
```diff
@@ -112,8 +112,10 @@ def get_analytics_zoo_classpath():
     """
     if os.getenv("BIGDL_CLASSPATH"):
         for path in os.getenv("BIGDL_CLASSPATH").split(":"):
-            if not os.path.exists(path):
-                raise ValueError("Path {} specified BIGDL_CLASSPATH does not exist.".format(path))
+            for pa in path.split("*"):
```
hkvision, line 115: why need to split by *?
change pa to a more meaningful name? Also add some comment for others to understand what this code segment is doing?

Le-Zheng: `export BIGDL_CLASSPATH=/opt/bigdl-0.14.0-SNAPSHOT/jars/*`.  However, `os.path.exists("/opt/bigdl-0.14.0-SNAPSHOT/jars/*") `return false.

hkvision: okay. But init on k8s is for pip install, in this case, uses don't need to set BIGDL_CLASSPATH right? In our code, I think there won't be `*`?

Le-Zheng: Yes. If using pip install, the path is like `/usr/local/envs/pytf1/lib/python3.7/site-packages/bigdl/share/dllib/lib/bigdl-dllib-spark_2.4.6-0.14.0-SNAPSHOT-jar-with-dependencies.jar:/usr/local/envs/pytf1/lib/python3.7/site-packages/bigdl/share/orca/lib/bigdl-orca-spark_2.4.6-0.14.0-SNAPSHOT-jar-with-dependencies.jar:`
However, k8s image is default built on spark312, which may need download whl from Sourceforge when building image. And k8s cluster need spark-submit. In order to manage the image easily for both mode, we prefer to use the BIGDL_CLASSPATH which keeps the same way with original analytics-zoo k8s image?

docker/bigdl-k8s/Dockerfile
```diff
@@ -98,6 +98,7 @@ ENV RUNTIME_DRIVER_MEMORY               10g
 ENV RUNTIME_PERSISTENT_VOLUME_CLAIM     myvolumeclaim
 ENV SPARK_HOME                          /opt/spark
 ENV JAVA_HOME                           /opt/jdk
+ENV BIGDL_CLASSPATH                     ${BIGDL_HOME}/jars/*
```
hkvision, line 101: so can we grep the jars here and set the BIGDL_CLASSPATH with all jars instead of using *? then the code in init k8s can be unchanged.

Le-Zheng: assembly jars/ folder is a lib folder includes all the bigdl artifatcts without dependencies and all the common dependencies. If we do not using *, it requires adding all jars (>100 jars) as a long path.

hkvision: can we filter the jars by name start with bigdl?

glorysdj: all the jars are needed, not only the bigdl jars.

glorysdj (approved):
LGTM

Le-Zheng:
jenkins: http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-PR-Dllib-Validation/53/ 
http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-Orca-PR-Validation/83/

docker/bigdl-k8s/Dockerfile
hkvision: all these jars can be needed in the docker, but why do we need to set all of them to BIGDL_CLASSPATH?

Le-Zheng: K8s image uses non-pip. Set all jars to BIGDL_CLASSPATH can make init_spark_on_k8s get it.

hkvision: but the other non-bigdl jars should be input into jars when spark-submit right? it doesn't make sense to put them in BIGDL_CLASSPATH?

Le-Zheng: Actually, ${BIGDL_HOME}/jars/* is equal to `*-jar-with-dependencies.jar`. need to rename BIGDL_CLASSPATH?

PR #3424, opened 2021-11-08
[PPML] Fix bugs in occlum realtime

Uxito-Ada:
1. Flink prefix was missed in check-status.sh, which led to a logfile-not-found error.
2. Convert argument name from Dlog4j.configuration to Dlog4j.configurationFile, which can avoid a configure-not-found error.


qiyuangong (approved):
LGTM

PR #3429, opened 2021-11-08
Update init_orca_context

sgwhat:
Re-add spark-submit mode for None-activate sc k8s-cluster test case.


hkvision:
Add some detailed description to explain what happened.

sgwhat:
For the spark-submit mainly sends config args to applications, if the applications didn't create spark context, it became a none-activate sc case, which happened during the orca k8s-cluster test. Hence we decided to re-add the "spark-submit" cluster mode for k8s-cluster with none-activate sc for now.

hkvision (approved):

PR #3433, opened 2021-11-09
Update hadoop.md

qiuxin2012:


qiuxin2012:
@jason-dai I think we can put spark-submit-python-with-bigdl.sh to our whl package, then user can use this script directly after he install bigdl using pip, like `pyspark` command in pyspark.

PR #3434, opened 2021-11-09
Fix bugs for graphene realtime

Uxito-Ada:
1. Fix the mis-typed word in check-status.sh.
2. Append jar path to avoid main method missing error.


glorysdj (approved):
LGTM

PR #3439, opened 2021-11-09
add argparser to ncf-dataframe and e2e to support passing parameters

zzti-bsj:
1. add argparser to ncf dataframe
2. modify sql dataframe API spark version from 2.4.6 to 3.1.2
3. add argparser to e2e


glorysdj (approved):
LGTM

PR #3440, opened 2021-11-09
recover shard_size in tests

yangw1234:


yangw1234:
http://10.112.231.51:18888/job/Orca-PR-Validation-Spark-3.1/105/ ut passed

PR #3446, opened 2021-11-10
Orca PytorchEstimator with pyspark backend

yushan111:
### Method implementation 

- [x] constructor
- [x] setup 
- [x] fit
- [x] evaluate
- [x] predict
- [ ] save (Implemented and UT not cover, will add in new PR )
- [ ] load (Implemented and UT not cover, will add in new PR)

### Test

- [x] UT: same UTs as ray backend has all passed 
- [ ] example: will add in new PR


yushan111:
http://10.112.231.51:18888/job/BigDL-Orca-PR-Validation/95/

PR #3459, opened 2021-11-11
remove duplicated tests

yangw1234:


yangw1234:
http://10.112.231.51:18888/job/ZOO-PR-NanoTests/72/

PR #3466, opened 2021-11-12
[ppml-g] fix private repo issue

glorysdj:


jason-dai:
Let's build the jar in a separate job, not as part of image building

glorysdj:
get it, will do that

PR #3480, opened 2021-11-15
VFL PSI algorithm and VFLContext

Litchilitchy:


scala/ppml/src/main/java/com/intel/analytics/bigdl/ppml/psi/test/BenchmarkClient.java
```diff
@@ -83,13 +83,12 @@ public static void main(String[] args) throws Exception {
...
-            String[] arg = {"-c", BenchmarkClient.class.getClassLoader()
-                    .getResource("psi/psi-conf.yaml").getPath()};
-            FLClient flClient = new FLClient(arg);
+
+            FLClient flClient = new FLClient();
             flClient.build();
             
             // Get salt from Server
-            salt = flClient.getSalt();
+            salt = flClient.psiStub().getSalt();
```
jason-dai, line 91: why expose `psiStub` to users here?

Litchilitchy: This is not exposed to user, FLClient is not exposed to user now thus I removed the wrapper in FLClient. User API are `PSI(), LinearRegression`, etc.

So this is simply an internal java test to see if the function works. We could use private package `private [ppml]` if we want to prevent user from using FLClient?

jason-dai:
I think we need to make the API simpler, such as
```scala
import com.intel.analytics.bigdl.ppml.PSI

val psi = new PSI()
val salt = psi.getSalt()
val hashedIds = psi.uploadSet(ids, salt) //handle hashing here

val intersection = psi.downloadIntersection() //handle timeout here
val ids = psi.intersectedIds(hashedIds, intersection)
```

scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/vfl/algorithm/PSI.scala
```diff
@@ -18,38 +18,61 @@ package com.intel.analytics.bigdl.ppml.vfl
...
+    salt
+  }
+  def getSalt(name: String, clientNum: Int, secureCode: String): String = {
+    salt = flClient.psiStub.getSalt(name, clientNum, secureCode)
+    salt
+  }
 
-  def uploadSet(hashedIdArray: util.List[String]): Unit = {
+  def uploadSet(ids: util.List[String]): Unit = {
+    val hashedIdArray = HashingUtils.parallelToSHAHexString(ids, salt)
```
jason-dai, line 56: instead of storing `salt` in the class, pass it in as an argument

scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/vfl/algorithm/PSI.scala
```diff
@@ -18,38 +18,61 @@ package com.intel.analytics.bigdl.ppml.vfl
...
+          logger.info(s"Got empty intersection, retry in $retry ms")
+          Thread.sleep(retry)
+        }
+        else {
+          logger.info("Intersection successful. Intersection's size is " + intersection.size + ".")
+          break
+        }
+      }
+    }
+    intersection
```
jason-dai, line 75: report error if `intersection` is null? shall we make `downloadIntersection` blocking at the server side?

scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/vfl/example/psi/PSIExample.scala
```diff
@@ -19,8 +19,8 @@ package com.intel.analytics.bigdl.ppml.vfl.example.psi
 import org.slf4j.LoggerFactory
 import java.util
 
-import com.intel.analytics.bigdl.ppml.psi.test.TestUtils
-import com.intel.analytics.bigdl.ppml.vfl.PSI
+import com.intel.analytics.bigdl.ppml.psi.HashingUtils
+import com.intel.analytics.bigdl.ppml.vfl.{PSI, VflContext}
```
jason-dai, line 23: Is PSI part of VFL algorithms? I think we should make it `com.intel.analytics.bigdl.ppml.PSI` or `com.intel.analytics.bigdl.ppml.algorithms.PSI` instead.

scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/vfl/algorithm/PSI.scala
Litchilitchy: Current `downloadIntersection` is blocked at server side using 
```
if(null == intersection) {
   this.wait();
}
```
Maybe it could just be replaced `if` by `while` then multiple-try is not necessary?

scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/psi/PSI.scala
```diff
@@ -14,45 +14,30 @@
  * limitations under the License.
  */
 
-package com.intel.analytics.bigdl.ppml.vfl
+package com.intel.analytics.bigdl.ppml.psi
```
jason-dai, line 17: change to `com.intel.analytics.bigdl.ppml.algorithms.PSI`

similarly, `com.intel.analytics.bigdl.ppml.algorithms.vfl.LogisticRegression`, `com.intel.analytics.bigdl.ppml.algorithms.vfl.LinearRegression`, etc.

scala/ppml/src/main/java/com/intel/analytics/bigdl/ppml/psi/PsiIntersection.java
```diff
@@ -131,7 +131,7 @@ public FindIntersection(String[] a,
 
     public List<String> getIntersection() throws InterruptedException{
         synchronized (this) {
-            if(null == intersection) {
+            while (intersection == null) {
```
jason-dai, line 134: timeout or wait indefinitely?

Litchilitchy: Previous code waits indefinitely, I think could change to:
* server does not block, just return
* client have a timeout and retry mechanism, and have a max_retry

scala/ppml/src/test/scala/com/intel/analytics/bigdl/ppml/vfl/PSISpec.scala
```diff
@@ -45,4 +47,21 @@ class PSISpec extends FlatSpec with Matchers with BeforeAndAfter{
     val intersection = pSI1.downloadIntersection()
     require(intersection.size() == 1, "Intersection number is wrong.")
   }
+  "PSI download null intersection" should "work" in {
```
Litchilitchy, line 50: Changed. The behavior could be seen in this UT

when there is no other clients, get null intersection, would return a `TimeoutException`

when the client itself does not upload set, it would throw `Exception`

PR #3482, opened 2021-11-15
Create Nano Windows Support

charlieJ107:
Add windows support for bigdl-nano. 

* Add verbose name for windows version in release script
* Switch cpu schedule logic depend on system platform
* Added logic to download specific lib only on non-windows systems


yangw1234:
LGTM

PR #3487, opened 2021-11-15
Orca Yarn Test

sgwhat:


python/orca/example/learn/openvino/predict.py
```diff
@@ -55,7 +55,7 @@ def crop(img, w, h):
 
     if args.cluster_mode == "local":
         init_orca_context(cores=args.core_num, memory=args.memory)
-    elif args.cluster_mode == "yarn":
+    elif args.cluster_mode.starswith("yarn"):
```
yushan111, line 58: starts

yushan111 (commented):
Shall we add orca integration tests to https://github.com/intel-analytics/BigDL/tree/branch-2.0/python/orca/dev/example

python/orca/example/learn/openvino/predict.py
sgwhat: Sure, it's better to put orca test scripts in orca/dev/example

sgwhat:
> Shall we add orca integration tests to https://github.com/intel-analytics/BigDL/tree/branch-2.0/python/orca/dev/example

Sure, it's better to put orca test scripts in orca/dev/example

PR #3491, opened 2021-11-16
Add spark backend to orca pytorch examples

yushan111:
The evaluation results for "spark" backend are the same with "torch_distributed" backend with the same seed.


yushan111:
http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-Orca-PR-Validation/105/

PR #3493, opened 2021-11-16
replace nn.keras with dllib.keras in examples

dding3:


python/dllib/src/bigdl/dllib/models/lenet/lenet.py
```diff
@@ -43,7 +43,7 @@ def build_model(class_num):
     (X_train, Y_train), (X_test, Y_test) = mnist.load_data(options.dataPath)
 
     model = build_model(10)
-    model.compile(loss='sparse_categorical_crossentropy',
+    model.compile(loss='categorical_crossentropy',
```
jason-dai, line 46: why this change?

scala/dllib/src/main/scala/com/intel/analytics/bigdl/dllib/models/lenet/LeNet5.scala
```diff
@@ -57,9 +56,10 @@ object LeNet5 {
     Graph(input, output)
   }
 
-  def kerasLayer(classNum: Int): keras.Sequential[Float] = {
-    import com.intel.analytics.bigdl.dllib.nn.keras._
+  def kerasLayer(classNum: Int): dllib.keras.models.Sequential[Float] = {
```
jason-dai, line 59: need to make it `dllib.keras.Sequential`

python/dllib/examples/lenet/lenet.py
```diff
@@ -15,8 +15,9 @@
 #
 
 from optparse import OptionParser
-from bigdl.dllib.nn.keras.layers.topology import Sequential
-from bigdl.dllib.nn.keras.layers.layer import *
+from bigdl.dllib.nncontext import *
+from bigdl.dllib.keras.layers import *
+from bigdl.dllib.keras.models import Sequential
```
jason-dai, line 20: need to make it `bigdl.dllib.keras.Sequential`

python/dllib/src/bigdl/dllib/keras/models.py
```diff
@@ -26,53 +26,6 @@
...
-        Create a Python Model base on the given java value
-        :param jvalue: Java object create by Py4j
-        :return: A Python Model
-        """
-        model = Sequential(jvalue=jvalue)
-        model.value = jvalue
-        return model
-
-
 class Model(KerasNet):
```
jason-dai, line 29: Should move `Model` to `dllib.keras.Model` too (similar to `tf.keras.Model`)

scala/dllib/src/test/scala/com/intel/analytics/bigdl/dllib/keras/SequentialSpec.scala
```diff
@@ -1,43 +1,42 @@
...
+//    val tmpFile = ZooSpecHelper.createTmpFile()
+//    model.saveModule(tmpFile.getAbsolutePath, overWrite = true)
+//    val reloadModel = Net.load[Float](tmpFile.getAbsolutePath)
+//    val inputData = Tensor[Float](2, 10).rand()
+//    ZooSpecHelper.compareOutputAndGradInput(
+//      model.asInstanceOf[AbstractModule[Tensor[Float], Tensor[Float], Float]],
+//      reloadModel.asInstanceOf[AbstractModule[Tensor[Float], Tensor[Float], Float]],
+//      inputData)
+//  }
+//}
```
jason-dai, line 42: why disable this test?

dding3: It fails the jekins http://10.112.231.51:18888/job/BigDL-PRVN-scalatest-others-Spark-2.4 which should not run keras related ut. From the log seems like BigDL-PRVN-scalatest-others-Spark-2.4 is trying to run serialzation unit tests for nn.layers, but somehow run keras.sequentialserialtest instead of nn.sequentialserialtest. Have disscussed with @Le-Zheng and she is looking at this.

Also keras.sequential serialization test is covered in other serialization tests(eg. test_save_load_Sequential) . To unblock the work, comment this ut temporary.

jason-dai: You should either fix the issue (if it's a valid test), or remove the test (if it's redundant); it is not reasonable to disable the test.

dding3:
http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-PR-Dllib-Validation/76/ Jekins passed except BigDL-PRVN-Python-ExampleTests as it's using published whl while this pr contains package rename, will rerun BigDL-PRVN-Python-ExampleTests after the published whl is updated.

jason-dai:
This is not quite right; the goal is to have something similar to Keras, in which you can have:

```python
model1 = keras.Sequential(...)
model2 = keras.Model(...)
```
and 
```python
from keras import Sequential
from keras import Model
```

PR #3495, opened 2021-11-16
Chronos: add two version fix for ray[tune] dependencies

TheaperDeng:
This is reported by our internal users.

Cause: 
Ray == 1.2.0 did not claim the version so the latest version of these two lib will be installed
https://github.com/ray-project/ray/blob/ray-1.2.0/python/setup.py#L124-L147

The latest version will cause error (but does not block the search to complete) during the auto tuning.


yushan111 (approved):

TheaperDeng:
http://10.112.231.51:18888/job/BigDL-Chronos-PR-Validation/46/

PR #3504, opened 2021-11-17
Chronos: Add cycle length auto detection in TSDataset and AutoTSEstimator

TheaperDeng:
image

- [x] data.utils
- [x] ut

TSDataset and AutoTSEstimation integration and API design will be left to next PR


TheaperDeng:
http://10.112.231.51:18888/job/BigDL-Chronos-PR-Validation/58/
http://10.112.231.51:18888/job/BigDL-PRVN-chronos-Python-Spark-2.4-py37-ray-part1/

PR #3508, opened 2021-11-17
fix orca libs on mac

yangw1234:


yangw1234:
need to first merge then run jenkins

PR #3513, opened 2021-11-18
[PPML] Seperate device-plugin section from README and rename k8s app to bigdl

Uxito-Ada:
1. Seperate **device-plugin** section from README, which is replaced with a **QuicStart** link.
2. Rename the k8s app name in yaml files from **analytics-*** to **bigdl-***.


glorysdj (approved):
LGTM

PR #3517, opened 2021-11-18
test(test-suits): add test scirpts for spark exampels

zzti-bsj:
1. add spark-examples.sh
2. add spark-sql-examples.sh


glorysdj (approved):
LGTM

PR #3519, opened 2021-11-18
fix data link in App tests for anomaly-detection-hd

Le-Zheng:


Le-Zheng:
http://10.112.231.51:18888/view/BigDL-2.0-NB/job/BigDL-NB-Python-AppTests-part2/6/console

PR #3533, opened 2021-11-22
Tf2estimator on pyspark backend without copy state to driver

jenniew:
TF2Estimator training does not copy model state to driver, evaluation/prediction load stats from shared filesystem.


yangw1234:
There seems to be problems when user does this:

1. `estimator.load` from a checkpoint. Then estimator's `self.model_weights` will be set to some value other than `None`.
2.  `estimator.fit` to resume training and saved new weights in `model_dir`
3. `estimator.evaluate` check the estimator's `self.model_weights` is not None, **it will use `self.model_weights` instead of the latest in `model_dir`**, which I do not think is correct.

I think the problem is we do not have a single source of truth. Both `self.model_dir` and `self.model_weights` could contain the latest weights, and it is hard to decide which one to use.

python/orca/src/bigdl/orca/learn/tf2/spark_runner.py
```diff
@@ -376,8 +377,16 @@ def validate(self, data_creator, batch_size=32, verbose=1, sample_weight=None,
...
             model = self.model_creator(self.config)
-            if self.model_weights:
-                model.set_weights(self.model_weights.value)
+
+        if self.model_weights is not None:
+            print("set weights from broadcast")
+            model.set_weights(self.model_weights.value)
+        elif self.model_dir:
+            start = time.time()
+            states = load_pkl(os.path.join(self.model_dir, self.application_id + "_states.pkl"))
```
yangw1234, line 386: In the case of saving on local file system, it seems only the previous rank 0 node will have the model saved in `self.model_dir`

jenniew:
> There seems to be problems when user does this:
> 
> 1. `estimator.load` from a checkpoint. Then estimator's `self.model_weights` will be set to some value other than `None`.
> 2. `estimator.fit` to resume training and saved new weights in `model_dir`
> 3. `estimator.evaluate` check the estimator's `self.model_weights` is not None, **it will use `self.model_weights` instead of the latest in `model_dir`**, which I do not think is correct.
> 
> I think the problem is we do not have a single source of truth. Both `self.model_dir` and `self.model_weights` could contain the latest weights, and it is hard to decide which one to use.

This is just temporary solution. I will change to just use weights from model_dir later.

python/orca/src/bigdl/orca/learn/tf2/spark_runner.py
jenniew: For cluster mode, we need user to specify the shared fs path not local path, I will add this requirement later.

python/orca/src/bigdl/orca/learn/tf2/pyspark_estimator.py
```diff
@@ -353,35 +339,28 @@ def load_weights(self, filepath, by_name=False):
         """
         model = self.model_creator(self.config)
         model.load_weights(filepath, by_name)
-        self.model_weights = model.get_weights()
+        state = {}
+        state["weights"] = model.get_weights()
+        save_pkl(state, os.path.join(self.model_dir, self.application_id + "_state.pkl"))
```
yushan111, line 344: It seems the logic to formatting the path occur in multiple places, shall we provide a function like `get_state_path`

python/orca/src/bigdl/orca/learn/tf2/spark_runner.py
```diff
@@ -432,10 +443,14 @@ def predict(self, data_creator, batch_size, verbose, steps, callbacks, data_conf
...
-            if self.model_weights:
-                local_model.set_weights(self.model_weights.value)
+            state = load_pkl(os.path.join(self.model_dir, self.application_id + "_state.pkl"))
+            if state:
+                local_model.set_weights(state['weights'])
         else:
             local_model = self.model_creator(self.config)
+            state = load_pkl(os.path.join(self.model_dir, self.application_id + "_state.pkl"))
+            if state:
+                local_model.set_weights(state['weights'])
```
yushan111, line 453: The logic in the two conditions might be the same.

PR #3538, opened 2021-11-22
Vfl NN service implementation

Litchilitchy:
1. add NN service for VFL, including LinearRegression and LogisticRegression, different algorithms are controlled by proto request arg `algorithm`
2. refactor some base class like Aggregator and Storage, to support generic types of data stored
3. move some Dllib depended classes from java to scala, e.g. NNServiceImpl
4. create a FLModel with abstract fit, eval, predict method and all Algorithms inherit from this, remove old VflModel
5. rename and move the class `VflEstimator -> VflNNEstimator`, `VflAggregator -> VflNNAggregator`
6. change the example code to new API usage and remove all example specific wrapped methods, e.g. `getIntersectionKeys` in old example


jason-dai:
Please add descriptions for this PR.

PR #3539, opened 2021-11-22
Chronos: pytorch based forecasters on bigdl-nano

TheaperDeng:
This PR is only to move all pytorch based model (TCN, LSTM, S2S) to rely on bigdl-nano API from bigdl-orca API.

UTs has been passed, since external API is not changed, the UTs are not changed

Currently, TCN/LSTM/S2SForecaster import can be passed with no `bigdl.orca` installed.


python/chronos/src/bigdl/chronos/forecaster/base_forecaster.py
```diff
@@ -15,18 +15,27 @@
...
+    'mae': mean_absolute_error,
+    'mse': mean_squared_error,
+    'rmse': partial(mean_squared_error, squared=False),
+    'mape': mean_absolute_percentage_error,
+    'r2': r2_score,
+    'smape': symmetric_mean_absolute_percentage_error
+}
 
 
 class BasePytorchForecaster(Forecaster):
```
shane-huang, line 41: base_forecaster only contains BasePytorchForcaster. maybe change the name to base_pytorch_forecaster to avoid confusing.

shane-huang: Do you think seperating the models by frameworks make sense? This way we can write some imports in the pytorch/__ini__.py without worrying about importing too many extra modules. Users  can import models like below. 
```
import chronos.forecaster.pytorch.LSTMForecaster
````
instead of
```
import chronos.forecaster.lstm_forecaster.LSTMForecaster
```
Statistical and machine learning models can be put in another folder as forecaster.ml, or just leave it outside the folders.

File: python/chronos/src/bigdl/chronos/model/tcn.py [hunk 2]
```diff
@@ -43,8 +43,7 @@
 import torch
 import torch.nn as nn
```
shane-huang, line 44: To make the models consistant and easier to add new models (tf or pytorhc), rename the file to tcn_pytorch as the other two models, or just move all of them to model/pytorch and remove the  file suffix "pytorch".

yushan111 (approved):

TheaperDeng:
pass
http://10.112.231.51:18888/job/BigDL-Chronos-PR-Validation/92/

PR #3540, opened 2021-11-22
Feature service load user vector & use list

cyita:


jason-dai:
Need to document:
(1) Architecture of the serving pipelines
(2) Schema of the parquet files
(3) Schema of the redis tables and configs
(4) List of the services and their APIs

PR #3551, opened 2021-11-23
Fix pytorch pyspark estimator to keep return value consistent with ray estimator

yushan111:
Changes:

In `PyTorchPySparkEstimator`:

- `fit`: `reduce_result` takes effect
- `evaluate`: return reduced result of `validate_stats` on all workers.

In UT of `PytorchRayEstimator`:

- Add convergence test


yushan111:
http://10.112.231.51:18888/job/BigDL-Orca-PR-Validation/125/

PR #3553, opened 2021-11-23
Ray_on_Spark Parameter_server Update

sgwhat:
This PR is created to solve some problems caused by importing way we did original, just move the parameter_sever model into python/orca/src/bigdl/orca/ray


sgwhat:
Test passed.

python/orca/src/bigdl/orca/ray/model.py
```diff
@@ -0,0 +1,218 @@
+# This file is adapted from https://github.com/ray-project/ray/blob
```
jason-dai, line 1: this should be an example, not part of the main library

sgwhat: Yeah, but the original way to import model in async_parameter_server.py is not working in yarn-test and k8s-test, put model.py in src could solve problems we met before.

jason-dai: "put model.py in src" is not the correct solution; what if the users need to build their own code on yarn or k8s? They cannot put their code in src.

sgwhat: yes, I'll think about other solutions.

sgwhat:
Jenkins passed separately in http://10.112.231.51:18888/view/ZOO-PR/job/ZOO-PR-Python-integration-test/401/ and http://10.112.231.51:18888/view/ZOO-PR/job/ZOO-PR-Python-integration-test/408/

PR #3555, opened 2021-11-24
Chronos: add import changing for chronos.autots

TheaperDeng:
This change is to make TSPipeline import available when user only `pip install bigdl-chronos`


TheaperDeng:
http://10.112.231.51:18888/job/Chronos-PR-Validation-Spark-3.1/68/
http://10.112.231.51:18888/job/Chronos-PR-Validation-Spark-2.4/59/

PR #3558, opened 2021-11-24
support keras.Model and keras.models.Model

dding3:


jason-dai:
LGTM.

dding3:
Most jekin jobs passed. http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-PR-Dllib-Validation/90/. The failed BigDL-PRVN-Python-ExampleTests-py37 because it's using published whl while this pr includes file rename. It should be able to pass after the published whl is updated.

PR #3571, opened 2021-11-24
Nano: onnxruntime support should also be available for multiprocessing

TheaperDeng:


python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
```diff
@@ -28,9 +28,147 @@
                             '_build_ortsess',
                             'update_ortsess',
                             'predict_step',
+                            'on_fit_start',
                             'inference']
 
 
+# internal function to build an ortsess
```
yangw1234, line 35: only changing function source code locations in this file?

python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
```diff
@@ -28,9 +28,147 @@
...
+                        file_path=file_path,
+                        sess_options=sess_options,
+                        **kwargs)
+
+
+# inference (new API to unifying users' inference method)
+def inference(self,
+                input_data,
+                batch_size=None,
+                file_path="model.onnx",
```
yangw1234, line 101: what if a user uses another file_path if ortsess is up to date?

TheaperDeng: good catch! Now, nothing will happen since ortsess will not be built again.

I may remove this param since users may not rely on this inference method to export their onnx file. They always have `.to_onnx()`, the built-in method in pytorch lightning.

python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
TheaperDeng: yep, and I remove `predict_step` since InferenceSession can not be pickled and no matter how we change the code, Trainer.predict() will not support onnxruntime backend when it is a multiprocessing trainer.

TheaperDeng:
http://10.112.231.51:18888/job/ZOO-PR-NanoTests/100/

yangw1234 (approved):

PR #3579, opened 2021-11-25
fix: add spark-example-sql-e2e.jar to Dockerfile

zzti-bsj:
1. fix: add spark-example-sql-e2e.jar to Dockerfile


glorysdj (approved):
LGTM

PR #3582, opened 2021-11-25
[WIP] decouple raycontext from pytorch ray estimators

yangw1234:
This is currently not ready for review

### DIFF
--- a/python/orca/src/bigdl/orca/learn/horovod/horovod_ray_runner.py
+++ b/python/orca/src/bigdl/orca/learn/horovod/horovod_ray_runner.py
@@ -81,9 +81,7 @@ def get_horovod_version():
 class HorovodRayRunner:
 
     # todo check whether horovod is built with gloo
-    def __init__(self, ray_ctx, worker_cls=None, worker_param=None, workers_per_node=1):
-        self.cores_per_node = ray_ctx.ray_node_cpu_cores // workers_per_node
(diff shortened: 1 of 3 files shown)


jason-dai:
I think a preferred approach is to decouple `RayContext` from `SparkContext`, and have our estimators to use `RayContext` and `SparkContext` as needed.

`RayContext` can either wrap native Ray (similar to the current case of local node) or RayOnSpark

yangw1234:
> I think a preferred approach is to decouple `RayContext` from `SparkContext`, and have our estimators to use `RayContext` and `SparkContext` as needed.
> 
> `RayContext` can either wrap native Ray (similar to the current case of local node) or RayOnSpark

Currently, our estimators only need `RayContext` for `num_nodes` and `cores_per_node`. 

In the native Ray case, I do not think RayContext has this information (need user to specify), assuming we are targeting running on an **existing cluster** and will **not** occupy the whole cluster. 

So I think it makes sense to this anyway, even if users need to call RayContext.init() to connect to a Ray cluster.

yangw1234:
> I think a preferred approach is to decouple `RayContext` from `SparkContext`, and have our estimators to use `RayContext` and `SparkContext` as needed.
> 
> `RayContext` can either wrap native Ray (similar to the current case of local node) or RayOnSpark

How about this one #3640

PR #3589, opened 2021-11-25
Fix init_spark_on_k8s

hkvision:
- fix python_env
- Add default values for spark.driver.host/port


Le-Zheng (approved):

hkvision:
http://10.112.231.51:18888/job/BigDL-Orca-PR-Validation/148/ One failure is irrelevant

PR #3607, opened 2021-11-26
update BigDL docker dockerfile and readme

Le-Zheng:


glorysdj (approved):
LGTM

Le-Zheng:
local test all app examples passed.  jupyter test on http://172.25.120.123:12345/tree also passed.

PR #3608, opened 2021-11-29
Chronos: add independent `chronos.metric`

TheaperDeng:
### Why we need `chronos.metric`?
We need to complete `*.evaluate(...)` in chronos with self-contained code without any dependency on `bigdl-orca`.
Convenient for future light HPO.
Convenient for future AD HPO.

### What does this PR do?
first version with basic test and similar API to orca, we may discuss about the details later.


python/chronos/src/bigdl/chronos/metric/forecast_metrics.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        Evaluate a specific metric for y_true and y_pred.
+
+        :param metric: String in ['mae', 'mse', 'rmse', 'r2'
+               , 'mape', 'smape']
+        :param y_true: Array-like of shape = (n_samples, \*).
+               Ground truth (correct) target values.
+        :param y_pred: Array-like of shape = (n_samples, \*).
+               Estimated target values.
+        :param multioutput: String in ['raw_values', 'uniform_average']
```
shane-huang, line 49: "raw_values" and “uniform_average” is too long and does not accurately describe what it means ("raw_values" actually means "element-wise", and "uniform_average" actually means aggregated by mean (comparing againse element-wise). Actually the hyper parameter "multioutput“ itself is confusing as well. 

My suggestion is: If we have only two options here, we can use a bool parameter, e.g. aggrerate=True or element-wise=True. If we have more than two aggregation option plus an element-wise option, then we can use aggregate='None', 'mean', etc.

python/chronos/src/bigdl/chronos/metric/forecast_metrics.py
```diff
@@ -0,0 +1,77 @@
...
+
+        :return: Float or ndarray of floats.
+                 A floating point value, or an
+                 array of floating point values, one for each individual target.
+        """
+        metric = metric.lower()
```
shane-huang, line 55: evaluate does not take a list of metrics?

python/chronos/src/bigdl/chronos/metric/forecast_metrics.py
```diff
@@ -0,0 +1,77 @@
...
+        assert metric in TORCHMETRICS_REGRESSION_MAP.keys(),\
```
shane-huang, line 56: should we use assert for parameter checks?

python/chronos/src/bigdl/chronos/metric/forecast_metrics.py
TheaperDeng: This setting is inherited from `orca.automl` (which inherited from `sklearn`). I think a string type `aggregate` is much better since we do plan to add more settings for our users' convenience.

python/chronos/src/bigdl/chronos/metric/forecast_metrics.py
TheaperDeng: This setting is inherited from `orca.automl`, I think we can support it easily.

python/chronos/test/bigdl/chronos/metric/test_forecast_metircs.py
```diff
@@ -0,0 +1,59 @@
+#
```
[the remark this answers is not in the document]
TheaperDeng: good catch, I will fix it.

python/chronos/src/bigdl/chronos/metric/forecast_metrics.py
TheaperDeng: I think it's OK? Since a false input will definitely give a error. Or should we use a `NotImplementedError` or `RuntimeError` here?

TheaperDeng:
http://10.112.231.51:18888/job/BigDL-Chronos-PR-Validation/59/

TheaperDeng:
I will first merge this PR and add more uts and give some enhancement to `chronos.metric` later with liang

PR #3615, opened 2021-11-29
Enable and fix orca style check

yangw1234:


yangw1234:
http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-Orca-PR-Validation/155/

PR #3626, opened 2021-11-30
Fix pytorch estimator hang due to uneven number of batches

yushan111:
Pytorch Estimator with ray backend will hang when training with uneven data across the workers (different total number of batches)

Fix by adding [`join()`](https://pytorch.org/docs/stable/generated/torch.nn.parallel.DistributedDataParallel.html) in the training loop context.


yushan111:
http://10.112.231.51:18888/job/BigDL-Orca-PR-Validation/158/

PR #3647, opened 2021-12-02
Nano: onnxruntime on nano enhancement

TheaperDeng:
This is a collection change according to the feedback of previews PR. #3240

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
+++ b/python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
@@ -187,6 +187,7 @@ def inference(self,
 # on_fit_start (LightningModule method overwrite)
 def on_fit_start(self):
     self._ortsess_up_to_date = False
+    self._ortsess = None
     self.exit_onnx()
     return self._on_fit_start_old()
 
--- a/python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
+++ b/python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
@@ -260,7 +260,9 @@ def bind_onnxrt_methods(pl_model: LightningModule):
     # check conflicts
     for component in ONNXRT_BINDED_COMPONENTS:
         if component in dir(pl_model):
-            warnings.warn(f"{component} method/property will be replaced.")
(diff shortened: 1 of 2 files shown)


python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
```diff
@@ -174,7 +175,9 @@ def bind_onnxrt_methods(pl_model: LightningModule):
     # check conflicts
     for component in ONNXRT_BINDED_COMPONENTS:
         if component in dir(pl_model):
-            warnings.warn(f"{component} method/property will be replaced.")
+            warnings.warn(f"{component} method/property will be replaced. You may rename your",
+                          " customized attributes or methods and call `Trainer.compile again `",
+                          "to avoid being overwrite.")
```
jason-dai, line 180: do we need to report warning for `on_fit_start`?

TheaperDeng: yes, since `on_fit_start` is now overwritten rather than use a decorator to change it. We need to report the warning. I drop the decorator because that the decorator implementation has no way to support multi-processing training (which needs to pickle the model)

jason-dai: But can the user do anything about it? And we just append something to the original `on_fit_start`, instead of replacing it, yes?

TheaperDeng: we "should" append something, still now we replace it (to support multi-processing). I may find some solution to append and support the multi-processing. (don't have any good idea right now)

TheaperDeng: well they may do some thing in other hook rather than `on_fit_start`.

TheaperDeng: and of course, that's not a good way to do it.

TheaperDeng:
http://10.112.231.51:18888/job/ZOO-PR-NanoTests/172/
http://10.112.231.51:18888/job/BigDL-PRNV-chronos-Python-Spark-2.4-py37-Horovod/240/

PR #3650, opened 2021-12-02
replace data url in estimator openvino UT

Le-Zheng:


Le-Zheng:
Jenkins passed: http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-Orca-PR-Validation/

PR #3678, opened 2021-12-07
Tf2estimator on Pyspark backend log to driver

jenniew:


yangw1234:
Is it possible to wrap the log forwarding logic into a single class and expose a few APIs to the estimator, so that other estimators can easily use that class without copying the code?

jason-dai:
Why not move most of the logic to `LogMonitor`?

jenniew:
> Why not move most of the logic to `LogMonitor`?

Yes, move most of logic to 'LogMonitor' under bigdl.orca.learn to be reused by other estimators.

jenniew:
Jenkins: http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-Orca-PR-Validation/208/

PR #3641, opened 2021-12-01
VFL spark DataFrame API for scala

Litchilitchy:

PR #3685, opened 2021-12-08
PPML add feature column support in API

Litchilitchy:


scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/algorithms/PSI.scala
```diff
@@ -60,5 +65,18 @@ class PSI() extends FLClientClosable {
...
+                                       rowKeyName: String = "IDS",
+                                       maxTry: Int = 5,
+                                       retry: Long = 3000): DataFrame = {
+    val spark = VflContext.getSparkSession()
+    import spark.implicits._
+    val ids = df.select(rowKeyName).as[String].collect().toList
+    uploadSet(ids, salt)
+    val intersections = downloadIntersection(maxTry, retry)
+    val intersectionSet = intersections.toSet
+    val dataSet = df.filter(r => intersectionSet.contains(r.getAs[String](rowKeyName)))
```
jason-dai, line 79: change to 
```scala
val dataSet = psi.getIntersectionSet(df, intersections )
```

jason-dai:
More comments at https://github.com/intel-analytics/BigDL/pull/3641

scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/vfl/FLModel.scala
```diff
@@ -26,18 +26,25 @@ import org.apache.spark.sql.DataFrame
...
+               batchSize: Int = 4,
+               featureColumn: Array[String] = null) = {
     estimator.getEvaluateResults().foreach{r =>
       println(r._1 + ":" + r._2.mkString(","))
     }
   }
-  def predict(data: DataFrame) = {
+  def predict(data: DataFrame,
+              batchSize: Int = 4,
+              featureColumn: Array[String] = null) = {
```
jason-dai, line 48: NotImplementedException

scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/vfl/FLModel.scala
```diff
@@ -26,18 +26,25 @@ import org.apache.spark.sql.DataFrame
...
+          valData: DataFrame = null) = {
     val _trainData = DataFrameUtils.dataFrameToSample(trainData)
     val _valData = DataFrameUtils.dataFrameToSample(valData)
     estimator.train(epoch, _trainData.toLocal(), _valData.toLocal())
   }
-  def evaluate() = {
+  def evaluate(data: DataFrame = null,
+               batchSize: Int = 4,
+               featureColumn: Array[String] = null) = {
     estimator.getEvaluateResults().foreach{r =>
```
jason-dai, line 41: Report error if `data` is not null

scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/vfl/FLModel.scala
```diff
@@ -26,18 +26,25 @@ import org.apache.spark.sql.DataFrame
 class FLModel() {
   val estimator: VflNNEstimator = null
   def fit(trainData: DataFrame,
-          valData: DataFrame,
-          epoch : Int = 1) = {
+          epoch: Int = 1,
+          batchSize: Int = 4,
+          featureColumn: Array[String] = null,
+          labelColumn: Array[String] = null,
```
jason-dai, line 32: the user will always need to provide `featureColumn` for Dataframe; `labelColumn` can be optional

scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/vfl/FLModel.scala
Litchilitchy: Some of these not implemented has been done in #3687

scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/vfl/FLModel.scala
Litchilitchy: In this case the data file without header could not be supported? For example the first lines of data we used in the demo
```
6,148,72,35,0,33.6,0.627,50,1
1,85,66,29,0,26.6,0.351,31,0
8,183,64,0,0,23.3,0.672,32,1
```
Users should add column names of data file before using.

jason-dai: > In this case the data file without header could not be supported? For example the first lines of data we used in the demo
> 
> ```
> 6,148,72,35,0,33.6,0.627,50,1
> 1,85,66,29,0,26.6,0.351,31,0
> 8,183,64,0,0,23.3,0.672,32,1
> ```
> 
> Users should add column names of data file before using.

I think the user can specify column names when reading csv? @hkvision

Litchilitchy: If user just read without header, spark will also create column names such as`c1, c2, ...` but just not very readable. And there are often some data with last column is label and rest are features (both with/without headers).(including our example) So I think this support could just ease the use of this? And if user specifies column name, it could also work in the same way.

hkvision: You mean we do some renaming when reading csv?
https://github.com/intel-analytics/BigDL/blob/branch-2.0/python/friesian/src/bigdl/friesian/feature/table.py#L75
https://github.com/intel-analytics/BigDL/blob/branch-2.0/python/orca/src/bigdl/orca/data/pandas/preprocessing.py#L218

jason-dai: I think it can be simply `spark.cvs(...).read(...).toDF('col1', 'col2', ...)`?

jason-dai: It is not reasonable to assume that there is always a label for VFL; I think the user need to explicitly provide feature list:

```scala
  def fit(trainData: DataFrame,
          epoch: Int = 1,
          batchSize: Int = 4,
          featureColumn: Array[String] = ,
          labelColumn: Array[String] = null,
          valData: DataFrame = null) = {
```
@Litchilitchy

Litchilitchy: Currently https://github.com/intel-analytics/BigDL/pull/3714/files?authenticity_token=OlCGJk1ldS1p6aPsZ1YvkuhiEmMcPj%2B7%2BqEoLkvevZQkD4ejJXlIWHzymBYGSovl6kLwrdjF3l3HWkGLcLMYNw%3D%3D&file-filters%5B%5D=.java&file-filters%5B%5D=.scala&hide-deleted-files=true#r770166975

Would modify in next PRs if necessary.

PR #3689, opened 2021-12-09
add generate-keys.sh && generate-password.sh into BigDL

zzti-bsj:
1. generate-keys.sh 
2. generate-password.sh


glorysdj (approved):
LGTM

PR #3690, opened 2021-12-09
Chronos: Fix forecaster doc

TheaperDeng:
The doc was broke due to preview pr #3539


TheaperDeng:
http://10.112.231.51:18888/job/BigDL-Chronos-PR-Validation/95/
http://10.112.231.51:18888/job/BigDL-PRVN-chronos-Python-Spark-3.1-py36-ray-part2/91/

PR #3707, opened 2021-12-13
remove log4j dependency in dllib pom

Le-Zheng:


Le-Zheng:
http://10.112.231.51:18888/job/BigDL-PR-Dllib-Validation/130/

glorysdj:
maybe we should check with mvn dependency:tree -Dverbose to check all the log4j dependencies

glorysdj (approved):
LGTM

Issue #3667, opened 2021-12-06
A more transparent Onnxruntime API on Nano


TheaperDeng:
### API Design
```python
from bigdl.nano.pytorch.trainer import Trainer

# for nn.Module users
model = Model()
loss = Loss()
optim = Optim()
model = Trainer.compile(model, loss, optim, onnx=True)

# for pl.LightningModule users
model = PL_Model()
model = Trainer.compile(model, onnx=True)

# fit
Trainer.fit(model, data)

# inferencing with `.forward()` (onnx)
model.eval_onnx()
pred = model(x)

# inferencing with `.forward()` (pytorch)
model.eval()
with torch.no_grad():
    pred = model(x)

# incremental fit
Trainer.fit(model, data)  # `on_fit_start` hook will call model.train() & model.exit_onnx()

# inference with `.inference()` (onnx)
pred = model.inference(x, batch_size=32, backend="onnx")

# inference with `.inference()` (pytorch)
pred = model.inference(x, batch_size=32, backend="pytorch")

# predict with `Trainer.predict()` (only pytorch)
Trainer.predict(model, data)

# update onnxruntime session conf (optional)
model.update_ortsess(onnx_conf)

# export onnx file (optional)
model.to_onnx()
```

### Background

In pytorch-lightning, the document clearly states that they expect users to use forward method to predict: https://pytorch-lightning.readthedocs.io/en/latest/starter/new-project.html#predict-or-deploy. 

And many community examples in pytorch-lightning doc do use these methods

1. https://github.com/3springs/attentive-neural-processes/blob/016272a077a19bc51d145d1ad99d910477458876/neural_processes/plot.py#L123-L125
2. https://github.com/ricardorei/lightning-text-classification/blob/master/classifier.py#L146-L166

### Some design memo

1. keep `.inference()` method to support richer feature (e.g. onnx config, batch_size, ...) that are not supported by `.forward()`.

2. `Trainer.predict()` method will never support onnxruntime inference since `onnxruntime.InferenceSession` can not be pickled, which means that it will never support multi-processing.
3. use `model.eval_onnx()` as a flag to illustrate that onnxruntime will be used in `.forward()`.

### Some Limitation and future enhancement

1. We only support the most classical training loop and forwarding procedure for now, which means there is only an input x and the forwarding procedure is `pred = model(x)`.

### How will this be implemented

```python
class LitMNIST(LightningModule):
    def __init__(self):
        super().__init__()

[...]

PR #3716, opened 2021-12-14
Nano: add transparent forward with ort

TheaperDeng:
```python
# inferencing with `.forward()` (onnx)
model.eval_onnx()
pred = model(x)
```
related to issue #3667 

------------------------------------------------------------------------
After benchmarking, we find that this will give us a 10~20% speed up on ResNet18 with Cifar10 inference latency on i7-9700. (median)

This is a smaller speed up since the onnx forward needs to do 2 extra casts (tensor -> numpy array -> tensor) to support the transparent speed up. Still the influence is not that much.

```python
pl_model.eval()
pytorch_latency = []
with torch.no_grad():
    for i, batch in enumerate(train_loader):
        x, y = batch
        start_time = time.time()
        pl_model(x)
        pytorch_latency.append(time.time() - start_time)
    print("pytorch_latency:", np.median(pytorch_latency)*1000, 'ms')


pl_model.eval_onnx()
onnx_latency = []
with torch.no_grad(): # useless but I leave here for transparent test
    for i, batch in enumerate(train_loader):
        x, y = batch
        start_time = time.time()
        pl_model(x)
        onnx_latency.append(time.time() - start_time)
    print("onnx_latency:", np.median(onnx_latency)*1000, 'ms')
```

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
+++ b/python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
@@ -188,6 +188,11 @@ def on_fit_start(self):
 
 
 def _forward_onnx(self, *args):
+    # if self.train() is called without calling self.exit_onnx()
+    if self.training:
+        self.exit_onnx()
+        return self.forward(*args)
+
     ort_inputs = {}
     for i, ort_input_item in enumerate(args):
         ort_inputs[self._forward_args[i]] = ort_input_item.numpy()
@@ -208,6 +213,9 @@ def eval_onnx(self, input_sample=None, file_path="model.onnx", sess_options=None
     :param sess_options: (optional) ortsess options in ort.SessionOptions type.
     :param **kwargs: (optional) will be passed to torch.onnx.export function.
     '''
+    # change to eval mode
+    self.eval()
+
     # get input_sample
     if input_sample is None and self.example_input_array:
         input_sample = self.example_input_array
--- a/python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
+++ b/python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
@@ -15,6 +15,8 @@
 #
 
 from pytorch_lightning import LightningModule
+import torch
+from torch.utils.data import DataLoader
 import onnxruntime as ort
 from functools import partial
 import warnings
@@ -29,6 +31,9 @@ ONNXRT_BINDED_COMPONENTS = ['_ortsess_up_to_date',
                             '_build_ortsess',
                             'update_ortsess',
                             'on_fit_start',
+                            '_forward_onnx',
+                            '_torch_forward',
+                            '_train_mode',
                             'inference']
 
 
@@ -112,10 +117,11 @@ def inference(self,
     or out-of-date.
 
     :param input_data: input data for prediction. If backend is set to "onnx",
-            the data type should be a numpy ndarray, where the first dim should be batch size.
+            the data type should be a numpy ndarray/torch tensor, where the first dim
+            should be batch size.
             If backend is NOT set to "onnx", a torch tensor is needed and the pytorch
-            forwarding method will be called. If there are multiple input, input_data
-            should be a list.
+            forwarding method will be called.
+            If there are multiple input, input_data should be a list.
     :param batch_size: int, inferencing batch_size. This value should not affect the
             final inferencing result but will affect resources cost(e.g. memory and time).
             Default to None, which takes all input_data in one batch.
@@ -125,6 +131,9 @@ def inference(self,
             the pytorch forwarding method.
     :param **kwargs: any other keywords that will be passed to onnx session's building.
     '''
+    # change to eval mode
+    self.eval()
+
     if isinstance(input_data, list):
         input_sample_list = input_data
     else:
@@ -146,28 +155,22 @@ def inference(self,
         # generate ort_inputs
         if batch_size is None:
             # this branch is only to speed up the inferencing when batch_size is set to None.
-            ort_inputs = {}
-            for i, ort_input_item in enumerate(input_sample_list):
-                ort_inputs[self._forward_args[i]] = ort_input_item
-            ort_outs = self._ortsess.run(None, ort_inputs)
-            return ort_outs[0]  # TODO: only support single output
+            return self._forward_onnx(*input_sample_list)
         else:
             yhat_list = []
             sample_num = input_sample_list[0].shape[0]  # the first dim should be sample_num
             batch_num = math.ceil(sample_num / batch_size)
             for batch_id in range(batch_num):
-                ort_inputs = {}
-                for i, ort_input_item in enumerate(input_sample_list):
-                    ort_inputs[self._forward_args[i]] = ort_input_item[batch_id * batch_size:
-                                                                       (batch_id + 1) * batch_size]
-                ort_outs = self._ortsess.run(None, ort_inputs)
-                yhat_list.append(ort_outs[0])  # TODO: only support single output
+                yhat_list.append(self._forward_onnx(*tuple(map(lambda x:x[batch_id * batch_size:
+                                                                          (batch_id + 1) * batch_size],
+                                                               input_sample_list))))
             # this operation may cause performance degradation
             yhat = np.concatenate(yhat_list, axis=0)
             return yhat
     else:
         # inference w/o onnxruntime (fallback to pytorch native forward)
         self.eval()
+        self.exit_onnx()
         with torch.no_grad():
             yhat_list = []
             sample_num = input_sample_list[0].shape[0]  # the first dim should be sample_num
@@ -187,15 +190,17 @@ def on_fit_start(self):
     self.exit_onnx()
 
 
-def _forward_onnx(self, *args):
-    # if self.train() is called without calling self.exit_onnx()
-    if self.training:
-        self.exit_onnx()
-        return self.forward(*args)
+def train(self, mode=True):
+    self.exit_onnx()
+    self._train_mode(mode)
 
+
+def _forward_onnx(self, *args):
     ort_inputs = {}
     for i, ort_input_item in enumerate(args):
-        ort_inputs[self._forward_args[i]] = ort_input_item.numpy()
+        if isinstance(ort_input_item, torch.Tensor):
+            ort_input_item = ort_input_item.numpy()
+        ort_inputs[self._forward_args[i]] = ort_input_item
     ort_outs = self._ortsess.run(None, ort_inputs)
     return torch.from_numpy(ort_outs[0])
 
@@ -208,7 +213,8 @@ def eval_onnx(self, input_sample=None, file_path="model.onnx", sess_options=None
     >>> pred = model(x)  # onnxruntime forwarding
     >>> model.exit_onnx()
 
-    :param input_sample: (optional) torch.Tensor or a list of them for the model tracing.
+    :param input_sample: (optional) a torch dataloader, torch.Tensor or a
+           list of them for the model tracing.
     :param file_path: (optional) The path to save onnx model file.
     :param sess_options: (optional) ortsess options in ort.SessionOptions type.
     :param **kwargs: (optional) will be passed to torch.onnx.export function.
@@ -217,6 +223,8 @@ def eval_onnx(self, input_sample=None, file_path="model.onnx", sess_options=None
     self.eval()
 
     # get input_sample
+    if isinstance(input_sample, DataLoader):
+        input_sample = tuple(next(iter(input_sample))[:-1])
     if input_sample is None and self.example_input_array:
         input_sample = self.example_input_array
     if input_sample is None and self.trainer is None:
@@ -267,5 +275,7 @@ def bind_onnxrt_methods(pl_model: LightningModule):
     pl_model._forward_onnx = partial(_forward_onnx, pl_model)
     pl_model.exit_onnx = partial(exit_onnx, pl_model)
     pl_model._torch_forward = pl_model.forward
+    pl_model._train_mode = pl_model.train
+    pl_model.train = partial(train, pl_model)
 
     return pl_model
--- a/python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
+++ b/python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
@@ -17,8 +17,9 @@
 from pytorch_lightning import LightningModule
 import torch
 from torch.utils.data import DataLoader
+from bigdl.nano.pytorch.lightning import LightningModuleFromTorch
 import onnxruntime as ort
-from functools import partial
+from functools import partial, wraps
 import warnings
 import torch
 import math
@@ -33,7 +34,6 @@ ONNXRT_BINDED_COMPONENTS = ['_ortsess_up_to_date',
                             'on_fit_start',
                             '_forward_onnx',
                             '_torch_forward',
-                            '_train_mode',
                             'inference']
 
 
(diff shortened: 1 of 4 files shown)


python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
```diff
@@ -184,6 +184,53 @@ def inference(self,
 # on_fit_start (LightningModule method overwrite)
 def on_fit_start(self):
     self._ortsess_up_to_date = False
+    self.exit_onnx()
+
+
+def _forward_onnx(self, *args):
```
jason-dai, line 190: What's the difference vs.  `inference`? Should you call `inference` (or have `inference` call  `_forward_onnx`)?

python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
```diff
@@ -184,6 +184,53 @@ def inference(self,
 # on_fit_start (LightningModule method overwrite)
 def on_fit_start(self):
     self._ortsess_up_to_date = False
+    self.exit_onnx()
+
```
jason-dai, line 188: how to handle `nn.Module.train()`?

python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
TheaperDeng: I don't think I should add `batchsize` or other additional parameters here for the forward, that is kind of important for some users (e.g. Chronos users) I think and it's only available in `inference`.

jason-dai: The should `inference` call `_forward_onnx`?

python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
TheaperDeng: good question, but why would a pl user call `nn.Module.train()`? I will go though some example doc and get back to this comment.

python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
TheaperDeng: emmm, but users may directly call `pl_model.inference(x, batch_size=32)` to control the memory usage.

TheaperDeng: This `pl_model(x)` or `pl_model.forward(x)` API is only for those users who want to transform their code to onnxruntime backend with 1 line changes(change `pl_model.eval()` to `pl_model.eval_onnx()`).

TheaperDeng: But may be they can control the batch size them selves by using a data loader like the one I write above. In that case, we may remove the `inference` method.
P.S. I agree `inference` and `forward` should not be provided to users at them same time with such similar functionalities

python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
jason-dai: > good question, but why would a pl user call `nn.Module.train()`? I will go though some example doc and get back to this comment.

It's common practice to call `model.eval` before inference, and call `model.train` before training, no?

python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
jason-dai: Yes, the user can still call `inference`, but internally `inference` should call `_forward_onnx`, no?

python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
TheaperDeng: For inference yes, since `Trainer.predict` (internally call `.eval()`) is not used frequently. Fore training no, `model.train()` is not required and not used in pl docs and users only needs to call `Trainer.fit`.

python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
TheaperDeng: Oh, that's what you mean... I misunderstand that to be using `_forward_onnx` replace `inference`. Yes, I will change it. THX

python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
jason-dai: Not for PL; but it should be used for standard PyTorch code?

TheaperDeng:
```python
pl_model = Trainer.compile(model, loss, optim, onnx=True)
```
Users will get a pl model, and users will use it as a pytorch model?

Emm, it's possible and (maybe) valid. I will check some details later.

jason-dai: > ```python
> pl_model = Trainer.compile(model, loss, optim, onnx=True)
> ```
> 
> Users will get a pl model, and users will use it as a pytorch model?
> 
> Emm, it's possible and (maybe) valid. I will check some details later.

The goal is to transparently accelerate both PL and PyTorch code

TheaperDeng: Surprisingly, it works well... (this example is from https://github.com/pytorch/examples/blob/master/mnist/main.py)
Only 4 lines to change and we can use the training speed up in bigdl-nano-init and inference speed up in onnxruntime
https://github.com/TheaperDeng/git_learning/compare/mnist-example-nano

jason-dai: > Surprisingly, it works well... (this example is from [pytorch/examples@`master`/mnist/main.py](https://github.com/pytorch/examples/blob/master/mnist/main.py?rgh-link-date=2021-12-14T08%3A58%3A34Z)) Only 4 lines to change and we can use the training speed up in bigdl-nano-init and inference speed up in onnxruntime [TheaperDeng/git_learning@`mnist-example-nano` (compare)](https://github.com/TheaperDeng/git_learning/compare/mnist-example-nano)

What is `model.eval_onnx(tuple(next(iter(test_loader))[:-1]))`?

I think the below code might be problematic:

```python
model = Trainer.compile(model, nn.NLLLoss(), optimizer, onnx=True)

model.eval_onnx()
...

#model.exit_onnx() is not called
model.train()
...
```

And if the user just wants to use ONNX, it does not make sense to ask the user to specify loss and optimizer in `compile` (the user may not know what loss or optimizer is at all if he or she does not train the model).

TheaperDeng: 1. If users never call `Trainer.fit(model, data)` they need to put an example input in `eval_onnx`. I can accept a dataloader to be the input later, that is `model.eval_onnx(test_loader)`
2. Yes, I will check and find a way to make it work later...
3. We may allow some thing like this
```python
# model is a nn.module
model = Trainer.compile(model, onnx=True)
# use model like a nn.module
```

python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
```diff
@@ -188,6 +188,11 @@ def on_fit_start(self):
 
 
 def _forward_onnx(self, *args):
+    # if self.train() is called without calling self.exit_onnx()
+    if self.training:
+        self.exit_onnx()
```
jason-dai, line 193: shall we set `self._ortsess_up_to_date = False` here?

TheaperDeng: yes, and since we need to support pytorch users other than pl users, there are many details to be considered. I will switch this PR to WIP for now.

jason-dai: > yes, and since we need to support pytorch users other than pl users, there are many details to be considered. I will switch this PR to WIP for now.

I think you should just override the `train` method?

TheaperDeng:
```python
# Pytorch-lightning users
model = Trainer.compile(model, loss=loss, optimizer=optimizer, onnx=True)

trainer = Trainer()
trainer.fit(model, data)

model.inference(x, batch_size=32)  # ort backend
model.inference(x, batch_size=32, backend=None)  # pytorch backend

model(x)  # pytorch backend
model.eval_onnx()
model(x)  # ort backend

trainer.fit(model, data)  # incremental fitting
```
```python
# pytorch users
model = Trainer.compile(model, onnx=True)

# >>>> start their own training loop >>>>
model.train()
# ...
# <<<< end their own training loop <<<<

model.inference(x, batch_size=32)  # ort backend
model.inference(x, batch_size=32, backend=None)  # pytorch backend

model(x)  # pytorch backend
model.eval_onnx(input_sample)
model(x)  # ort backend

# >>>> start their incremental training loop >>>>
model.train()
# ...
# <<<< end their incremental training loop <<<<
```

python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
```diff
@@ -184,6 +188,66 @@ def inference(self,
 # on_fit_start (LightningModule method overwrite)
 def on_fit_start(self):
     self._ortsess_up_to_date = False
+    self.exit_onnx()
+
+
+def train(self, mode=True):
+    self.exit_onnx()
+    self._train_mode(mode)
```
jason-dai, line 196: why add a new `_train_mode()` instead of calling `super().train()`?

TheaperDeng: `super()` need a class name when it is defined outside a class. And I can't get it here.

python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
```diff
@@ -184,6 +188,66 @@ def inference(self,
...
+    # change to eval mode
+    self.eval()
+
+    # get input_sample
+    if isinstance(input_sample, DataLoader):
+        input_sample = tuple(next(iter(input_sample))[:-1])
+    if input_sample is None and self.example_input_array:
+        input_sample = self.example_input_array
+    if input_sample is None and self.trainer is None:
+        raise RuntimeError("You must state an input_sample or fit on the model to use `eval_onnx`.")
```
jason-dai, line 232: You must specify an input_sample or call Trainer.fit on the model first to use `eval_onnx`

python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
```diff
@@ -184,6 +188,66 @@ def inference(self,
...
+    :param sess_options: (optional) ortsess options in ort.SessionOptions type.
+    :param **kwargs: (optional) will be passed to torch.onnx.export function.
+    '''
+    # change to eval mode
+    self.eval()
+
+    # get input_sample
+    if isinstance(input_sample, DataLoader):
+        input_sample = tuple(next(iter(input_sample))[:-1])
+    if input_sample is None and self.example_input_array:
```
jason-dai, line 229: elif

python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
```diff
@@ -208,5 +272,11 @@ def bind_onnxrt_methods(pl_model: LightningModule):
     pl_model.update_ortsess = partial(update_ortsess, pl_model)
     pl_model.on_fit_start = partial(on_fit_start, pl_model)
     pl_model.inference = partial(inference, pl_model)
+    pl_model.eval_onnx = partial(eval_onnx, pl_model)
+    pl_model._forward_onnx = partial(_forward_onnx, pl_model)
+    pl_model.exit_onnx = partial(exit_onnx, pl_model)
+    pl_model._torch_forward = pl_model.forward
+    pl_model._train_mode = pl_model.train
+    pl_model.train = partial(train, pl_model)
```
jason-dai, line 281: for `_forward_args`, you may explicitly check if the model is `LightningModuleFromTorch`; if not, `pl_model.model` will be incorrect.

python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
jason-dai, line 229: Will `example_input_array` be set somewhere?

python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
jason-dai: `self._ortsess_up_to_date = False`?

python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
jason-dai: So `_forward_onnx` will always return numpy? Is it compactible with `forward`?

python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
```diff
@@ -184,6 +188,66 @@ def inference(self,
 # on_fit_start (LightningModule method overwrite)
 def on_fit_start(self):
     self._ortsess_up_to_date = False
+    self.exit_onnx()
```
jason-dai, line 191: why don't you call the original `on_fit_start` here, just like `train`?

python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
TheaperDeng: yes, it's a property in pytorch lightning. Users may set it themselves

python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
```diff
@@ -29,6 +32,8 @@
                             '_build_ortsess',
                             'update_ortsess',
                             'on_fit_start',
```
jason-dai, line 34: remove 'on_fit_start' from the list?

python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
```diff
@@ -184,6 +188,68 @@ def inference(self,
 # on_fit_start (LightningModule method overwrite)
 def on_fit_start(self):
     self._ortsess_up_to_date = False
+    return self._on_fit_start_old()
```
jason-dai, line 191: also `self.exit_onnx()`?

python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
jason-dai: And does conversion between numpy and tensor impact latency?

jason-dai:
See comments inline; otherwise LGTM

jason-dai:
BTW, you also need support the follow in a separate PR:
1) `loss` and `optimizer` should be optional in `compile` for ONNX and quantization support
2) Handle the interaction between ONNX and quantization

TheaperDeng:
> BTW, you also need support the follow in a separate PR:
> 
> 1. `loss` and `optimizer` should be optional in `compile` for ONNX and quantization support
> 2. Handle the interaction between ONNX and quantization
> 3. Add `export_onnx` support?

for 1,2 I will have some look. For 3, I think `to_onnx` is always there(a built-in method in pl) for the compiled model.

python/nano/src/bigdl/nano/pytorch/onnx/onnxrt_inference.py
TheaperDeng: not really, I did some tests and the impact is minor

TheaperDeng:
http://10.112.231.51:18888/job/ZOO-PR-NanoTests/147/
http://10.112.231.51:18888/job/BigDL-PRVN-chronos-Python-Spark-2.4-py37-ray-part1/173/

Issue #3710, opened 2021-12-13
Chronos: version conflict between `tensorflow>=1.15.0,<2.0.0` and `bigdl-nano[pytorch]`


TheaperDeng:
The conflict is on `pytorch_lightning==1.4.2` and `tensorflow>=1.15.0,<2.0.0` on `tensorboard`.  And this cause problem when users use `pip install bigdl-chronos[all]`, which causes `pytorch_lightning==0.8.5` to be installed

My suggestion is to remove `tensorflow>=1.15.0,<2.0.0` since only mtnet is relied on it. And add clear warning in code and doc let our mtnet users to install `tensorflow>=1.15.0,<2.0.0` themselves.

@yushan111 @shane-huang

jason-dai:
We can remove `tensorflow>=1.15.0,<2.0.0`; but why `pytorch_lightning==0.8.5`?

TheaperDeng:
> We can remove `tensorflow>=1.15.0,<2.0.0`; but why `pytorch_lightning==0.8.5`?

because pip find that `pytorch_lightning==0.8.5` does not conflict with `tensorflow>=1.15.0,<2.0.0` after a long time searching.

TheaperDeng:
merged, to be verified

PR #3720, opened 2021-12-14
Chronos: Remove tf1 from dependencies

TheaperDeng:
related to #3710


TheaperDeng:
http://10.112.231.51:18888/job/BigDL-Chronos-PR-Validation/104/

PR #3731, opened 2021-12-14
fix local-cluster-serving scripts file

zzti-bsj:
1. update scala version from 2.11.8 to 2.12.10 in bash.manifest.template file
2. update start-all.sh
3. update start-http-frontend.sh
4. update start-local-cluster-serving.sh


glorysdj (approved):
LGTM

PR #3733, opened 2021-12-14
Remove XGBoost dependency to make Chronos install depends on Orca Automl

yushan111:


TheaperDeng (approved):
LGTM, the only concern is `h5py`, where is it used in Chronos...

yushan111:
> LGTM, the only concern is `h5py`, where is it used in Chronos...

It is for `tensorflow`.  Since we have removed `tensorflow`, we may not need `h5py` as well.

Related to issue https://github.com/intel-analytics/analytics-zoo/issues/3154

yushan111:
http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-Orca-PR-Validation/211/

PR #3735, opened 2021-12-15
Chronos: TCN model changes (weightnorm -> batchnorm, resort the layers)

### LINKED ISSUE #3736: Chronos TCN model: Local optima issue
## What's the issue?
Training loss for tcn model will stuck at a relatively high value and will not go down any more. 

## What components does the issue affect?
Any component use [tcn model](https://github.com/intel-analytics/analytics-zoo/blob/master/pyzoo/zoo/chronos/model/tcn.py]), such as TCNForecaster, New AutoTSTrainer...

## History
This issue has been reported and found when we work on customer case. For some random seed (10% - 20%), training loss will stuck at a relatively high value for some time (such as 0.5 - 1 epoch). The issue does not raise enough alert since in customer's case, model will "get out of" local optima some time and it seems that it does not affect the final result.

## Why it is raised again?
Still, this issue become much worse when it comes to our user case dataset(nyc_taxi and network traffic), especially when the data is scaled before training. Here are some experiment result.

nyc_taxi: In 8/20 random seed, the issue happens.
network traffic: In 5/20 random seed, the issue happens.

## How to solve the issue?
Sadly, I have not found a way to totally resolve the issue. Still, I found the way to **significantly alleviate** the issue. The initialization of tcn should be replaced by kaiming uniform(default init for pytorch). Here are some experiment result after using kaiming uniform.

nyc_taxi: In 2/20 random seed, the issue happens. (75% down)
network traffic: In 1/20 random seed, the issue happens. (80% down)

## Action Required
1. add an option in TCNForecaster and tcn model to control if to use kaiming uniform or the original init method in paper. (default to kaiming uniform)
2. For those notebook using TCNForecaster, add random seed control to avoid different output.
```python
torch.manual_seed(0)
random.seed(0)
np.random.seed(0)
```
3. give clear warning/note in documentation.

TheaperDeng:
1. weightnorm -> batchnorm, this will help us use multi-processing #3737 and IPEX in the future and possibly fix #3736 
2. resort the layers, this will give us 5% improvement due to layer fusion with no accuracy dropping

TODO:
- [x] accuracy check in use case notebook


yushan111 (approved):
LGTM

TheaperDeng:
http://10.112.231.51:18888/job/BigDL-Chronos-PR-Validation/108/

PR #3744, opened 2021-12-15
Chronos: Fix chronos colab-notebook torchtext version conflict

liangs6212:


TheaperDeng (approved):
LGTM, only needs to make sure it runs well on colab

PR #3745, opened 2021-12-15
Chronos: AutoProphet & AutoARIMA without orca when inference

TheaperDeng:


TheaperDeng:
http://10.112.231.51:18888/job/BigDL-Chronos-PR-Validation/130/
http://10.112.231.51:18888/job/BigDL-PRVN-chronos-Python-Spark-3.1-py36-ray-part2/130/
http://10.112.231.51:18888/job/BigDL-Orca-PR-Validation/229/

PR #3746, opened 2021-12-15
Support TF2Estimator finetune for Keras

hkvision:


jason-dai:
I don't think we need to support this; the users can use any custom logic to handle it, just as they would do for standard TF or PyTorch code.

hkvision:
> 

Okay. Then I will make it an example for the demo later if necessary.

PR #3750, opened 2021-12-16
add Xgb training on RecSys data into friesian

songhappy:


songhappy:
merge this one with PR-3876

PR #3766, opened 2021-12-21
Chronos: fix broken api doc in autots

TheaperDeng:


TheaperDeng:
http://10.112.231.51:18888/job/BigDL-Chronos-PR-Validation/137/

PR #3773, opened 2021-12-22
fix high vulnerabilities in pom dependencies [common/spark/spark-version]

Le-Zheng:


Le-Zheng:
dllib pass: http://10.112.231.51:18888/job/BigDL-PR-Dllib-Validation/160/

Issue #3323, opened 2021-10-28
Remove the spark3 name in whls built on top of spark3?


hkvision:
Whether we need to differ the project names by `-spark3`.

One potential issue of doing this I can think of now is if users wrongly mix the whls built on top of spark2 and spark3, for example using the spark3 dllib and spark2 orca, there will be no way to identify this misusage right? Any suggestions on this? @yangw1234 @jason-dai  @Le-Zheng 
Using the same name is also hard to manage. I would suggest using different names. We can add some doc in python guide to assist users to install whls on spark3.

yizerozhuang:
install step for bigdl on spark 3.1.2
```
pip install https://sourceforge.net/projects/analytics-zoo/files/orca-py/bigdl_orca-0.14.0b20211107-py3-none-manylinux1_x86_64.whl
pip install https://sourceforge.net/projects/analytics-zoo/files/dllib-py/bigdl_dllib-0.14.0b20211107-py3-none-manylinux1_x86_64.whl
pip install https://sourceforge.net/projects/analytics-zoo/files/friesian-py/bigdl_friesian-0.14.0b20211107-py3-none-manylinux1_x86_64.whl
```

PR #3781, opened 2021-12-22
Add version suffix to spark3 release

### LINKED ISSUE #3070: Spark 3 with Scala 2.12
Hi devs!

Not an issue but a question (feel free to add the label):

Given that Spark 3 has been released now a few month ago, do you have any plans to go for it and update to Spark 3 which also needs to move to Scala 2.12?
I'm also wondering if this [Spark 3 deployment script](https://github.com/intel-analytics/BigDL/blob/master/scripts/deploy-spark3.sh) can work with the current version still being on Scala 2.11

Thanks in advance.

hkvision:
Fix #3070 #3323 #3845

### DIFF
--- a/python/chronos/dev/release/release_default_linux_spark312.sh
+++ b/python/chronos/dev/release/release_default_linux_spark312.sh
@@ -43,9 +43,9 @@ if [ "${version}" == "default" ]; then
     version=$(cat $CHRONOS_DIR/../version.txt | head -1)
 fi
 if [[ "${version}" == *"spark3" ]]; then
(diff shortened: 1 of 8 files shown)


hkvision:
Tested locally. Will release after log4j issue gets resolved.

jason-dai:
Need to update user guide too

hkvision:
> Need to update user guide too

Seems spark3 suffix is not valid for pypi version... still trying to seek for alternative solutions...

hkvision:
Since suffix for version is not doable for pypi, closing this PR.

PR #3783, opened 2021-12-22
Adding pyzmq dependency for orca

bendavidsteel:
I encountered a pyzmq dependency in Orca while setting up BigDL in my project, should this be added to reqs?


yushan111:
Thanks for contributing!

PR #3788, opened 2021-12-23
Chronos: add statsmodels to chronos setup

TheaperDeng:
For the auto detection of cycle


TheaperDeng:
http://10.112.231.51:18888/job/BigDL-Chronos-PR-Validation/148/

Issue #3712, opened 2021-12-13
Nano: light weight hpo support in nano tensorflow


shane-huang:
## Overview

Refer to hpo design in nano pytroch in issue #3925 

Enable hyper param tuning in nano tensorflow. Make it as transparent as possible to users. General principals include:
- provide some bkm default settings if no search space is explicitly provided by user. (e.g. learning rate, batch size, etc.) 
- frequently used metrics don't need extra configurations for hpo. e.g. mse -> minimize objective, etc.

### ```model.search```

```python

model = nano.tf.keras.Sequential([...])
...
model.search(sub_x, sub_y, val_data=(subx_val, suby_val), n_trials, epochs=20, ...) # <--- add 1 line code
#model.search(...,resume=True,...)  # <---- tune can be called several time to resume from where it is left.  

model.fit(train_x, train_y, epochs=100, use_tune_id=trial_id, ...) 

```
* ```search``` does not return or save any tuned model, it just collects the statistics of all the trials. So ```fit``` is still needed after ```search```. (Notes: maybe we can provide the option to save checkpoints so that fit just get the best checkpoint w/o training)
* ```search``` automatically search for training related hparams, e.g., batch size, learning rate, etc. The search space is by default inferred and adapted from the input and environment. It can also be explicitly specified in arguments, as the exmaple shown below. 
```python 
model.search(..., n_trials=100, batch_size=automl.space.Categorical([32,64]), learning_rate=automl.space.Real(0.001,0.1), ...) 
```
* ```search``` can be called several times in order to resume the tuning.  
* when tuning finished, user can explicitly call ```end_search``` and specify the parameters in which trial will be used for the following ```fit```. If trial id is not specified in fit, the best trial is selected by default. If ```end_search``` is not explicit called, it will be called in ```fit```. 
* We assume that it is common to use a subset of data in ```search```, while in ```fit``` the user can use the full datast (train+val used in search), or use a larger dataset, or more epochs. 

### ```model.search_summary```

```python
model = nano.tf.keras.Sequential([...])
...
model.search(...) 
hpo_summary_df = model.search_summary()
print(df)

```
* ```search_summary``` returns a data frame containing the hparams used in each trial and the value of target metric (and possibly speed or other attributes). Can serve as some sort of a leaderboard.  
* tune summaries can also be visualized, using tools provided by optuna

### Basic usage W/O search space configurations: 
[...]

PR #3791, opened 2021-12-23
add hpo support for nano tf

shane-huang:
Two ways to specify search space. 
- using closure - i.e. define model creator functions and specify spaces as exmaples shown in this PR. 
- using arguments - e.g. fit(...,lr=hp.float('lr',0.01,0.001),...)

In this PR the closure way is supported for both Sequential/Functional API, so that friesian pipeline use case can be supported. 
arguments way will be supported in follow PR. 

|Data Creation|Model Definition| Model Compile|Model Fit|Support|
| ----------- | ----------- |----------|----------|----------|
|-|closure|closure|-|Yes (w/ example)|
|-|-|closure|-| Yes, needs testing|
|-|closure|-|-| Yes, needs testing|
|-|-|arguments|-| No|
|-|-|-|arguments|No|

Above covers the 2nd and 3nd cases in the design #3712 , where search space is explicitly specified by user. 
For 1nd case where search space is automatically inferred, we will open a new PR to support it.


yangw1234:
Does optuna need to be added to `install_requires`

shane-huang:
> Does optuna need to be added to `install_requires`

Yes. And optuna would require a few dependences as well. Maybe we can discuss if we need to add a new install model, e.g. nano[hpo].  Also The tests are not added yet. Once the interfaces and usage style is agreed on, I will continue to add tests and run through jenkins.

jason-dai:
I'd prefer something simpler, such as
```python
def create_model(...)
    return nano.tf.keras.Model(...)

def compile_model(...)
    ...

model = nano.tf.keras.tuning( \
                   model_creater=create_model, \
                   model_compiler=compile_model)
model.tune(...)
```

python/nano/example/tensorflow/hpo/hpo.py
```diff
@@ -0,0 +1,129 @@
...
+from bigdl.nano.tf.keras import Model
+
+N_TRAIN_EXAMPLES = 3000
+N_VALID_EXAMPLES = 1000
+
+CLASSES = 10
+
+# model_initor takes only a trial argument
+# use lambda to accomodate extra arguments as examples shown in main
+def model_inits(trial, input_shape):
```
jason-dai, line 33: what is `trial` here?

python/nano/src/bigdl/nano/tf/hpo/tunemixin.py
```diff
@@ -0,0 +1,94 @@
...
+            print("  Value: {}".format(best.value))
+            print("  Params: ")
+            for key, value in best.params.items():
+                print("    {}: {}".format(key, value))
+            return self.study.trials_dataframe(attrs=("number", "value", "params", "state"))
+        else:
+            print("Seems you have not done any tuning yet.  \
+                  Call tune and then call tune_summary to get the statistics.")
+
+    def end_tune(self, use_trial_id=-1):
```
jason-dai, line 73: how does the user specify use_trial_id?

shane-huang: user can call model.tune_summary() to get a dataframe that contains all trial results. Each row in the data frame is one trail. Columns include trial_id, value of the hyperparmeters and the evaluation result. If user_trial_id is not set, the trial with the best evaluation results will be used automatically. Or user can select the id he wants to use and pass it in use_trial_id.

python/nano/example/tensorflow/hpo/hpo.py
shane-huang: a trial is a tuning instance where user can get suggested value for hyper parameters.

shane-huang: i'll try revise the API

shane-huang: this example is deprecated

python/nano/src/bigdl/nano/automl/tf/keras/Sequential.py
```diff
@@ -0,0 +1,76 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import tensorflow as tf
+
```
yangw1234, line 18: Should we reuse the `bigdl.nano.tf.keras.sequential` ?

shane-huang: In order not to mess up with the current sequential I just keep it in automl for now. There're are loads of modifications in sequential and model, e.g. autogenerated decorated layers, decide whether to create a normal graph or searchable graph based on search space, lazy computation graph construction and delayed model instantiation (which makes some of the methods not usable until end_search). 

I will merge hpo and tf optimizations into tf.keras.sequential when hpo is relatively stable

shane-huang:
Close this PR as it won't be merged.

PR #3798, opened 2021-12-24
disable nano env check for now

yangw1234:


yangw1234:
http://10.112.231.51:18888/view/BigDL-2.0-PR/job/ZOO-PR-NanoTests/150/

Issue #3797, opened 2021-12-24
org.apache.flink.client.program.ProgramInvocationException: The main method caused an error: null


704572066:
get the error when starting the zoo-cluster-serving:0.9.0 with the  saved bigdl model by calling pretrainedNNModel.model.save()
![微信图片_20211224130844](https://user-images.githubusercontent.com/23404868/147318968-7e01ddee-a127-49d0-b838-1fad9d189c90.png)
![微信截图_20211224130307](https://user-images.githubusercontent.com/23404868/147318629-d09f27c9-c3d0-40e2-bfe0-c8c86a413c04.png)

yangw1234:
@Litchilitchy @dding3 Could you help take a look? It looks like a problem bigdl's ModuleLoader.

Litchilitchy:
seems like bigdl model loading error. If no idea for now, I suggest to use a scala program (would be better in IDE) to directly load InferenceModel for your model first. And see if it works.

usage:
```
val model = new InferenceModel()
model.doLoadBigDL(path/to/model)
```

jason-dai:
@704572066 which version of Analytics Zoo / BigDL are you using?

Litchilitchy:
I think the version is cluster serving docker 0.9.0?

jason-dai:
How about `pretrainedNNModel.model.save()`? How is `pretrainedNNModel` trained? Using Analytics Zoo or BigDL?

704572066:
> @704572066 which version of Analytics Zoo / BigDL are you using?

![image](https://user-images.githubusercontent.com/23404868/147516910-007a9d14-ecf0-49c6-a5b8-cf8e64d98450.png)

704572066:
> How about `pretrainedNNModel.model.save()`? How is `pretrainedNNModel` trained? Using Analytics Zoo or BigDL?

reference the link：https://github.com/mrafayaleem/transfer-learning-bigdl

Litchilitchy:
Could you try `pretrainedNNModel.model.saveModel()` to save?

704572066:
> Could you try `pretrainedNNModel.model.saveModule()` to save?

pretrainedNNModel.model.saveModule() not defined, I tried pretrainedNNModel.model.saveModel() and came across  another error in cluster-serving
![image](https://user-images.githubusercontent.com/23404868/147523993-ffb30238-338c-4369-a817-8f82a85e13d2.png)

704572066:
[...]

PR #3802, opened 2021-12-27
Serving BigDL model from NNModel UT

Litchilitchy:
This UT is based on branch 2.0, it passed. However if I change to zoo branch 0.9, the load would fail.

By the way I am using branch 2.0 bigdl python keras module to export NNModel. The API is `saveModel` instead of `save` and `save` could not be used in branch 2.0. Thus so far I do not know what are the incompatibilities in this case.

* The issue link: https://github.com/intel-analytics/BigDL/issues/3797
* save API link: https://github.com/intel-analytics/BigDL/blob/1725d1156dd992b057a61c0d16ddd4bdfeef1003/python/dllib/src/bigdl/dllib/keras/engine/topology.py#L32

@jason-dai @dding3

### DIFF
--- a/scala/serving/pom.xml
+++ b/scala/serving/pom.xml
@@ -50,6 +50,12 @@
             </exclusions>
         </dependency>
         <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-log4j12</artifactId>
+            <version>1.7.16</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.scala-lang</groupId>
             <artifactId>scala-library</artifactId>
             <version>${scala.version}</version>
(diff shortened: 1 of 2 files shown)


jason-dai:
Form the original issue https://github.com/intel-analytics/BigDL/issues/3797: 
```
get the error when starting the zoo-cluster-serving:0.9.0 with the saved bigdl model by calling pretrainedNNModel.model.save()
```

The user seams to use `model.save` to save the model and then load it in cluster serving?

Litchilitchy:
> Form the original issue #3797:
> 
> ```
> get the error when starting the zoo-cluster-serving:0.9.0 with the saved bigdl model by calling pretrainedNNModel.model.save()
> ```
> 
> The user seams to use `model.save` to save the model and then load it in cluster serving?

Yes, I tried to reproduce on branch 0.9, there is an error but not the exactly same error. So I am not sure about what exact version of NNFrame code (maybe some changes between branch 0.9 and zoo-cluster-serving:0.9.0) accordingly.

Litchilitchy:
> Form the original issue #3797:
> 
> ```
> get the error when starting the zoo-cluster-serving:0.9.0 with the saved bigdl model by calling pretrainedNNModel.model.save()
> ```
> 
> The user seams to use `model.save` to save the model and then load it in cluster serving?

Yes, I tried to reproduce on branch 0.9, there is an error but not the exactly same error. So I am not sure about what exact version of NNFrame code (maybe some changes between branch 0.9 and zoo-cluster-serving:0.9.0) accordingly.

PR #3805, opened 2021-12-27
make deprecated module.save/load not accessiable

dding3:


jason-dai:
private [bigdl]?

dding3:
jekins pass. http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-PR-Dllib-Validation/200/

jason-dai:
Need to add document for:
1) save and load APIs for python Keras 
2) save and load APIs for scala Keras

PR #3807, opened 2021-12-28
Tf2estimator on Pyspark backend Checkpoint support

jenniew:
Support checkpointing with tf.keras.callbacks.ModelCheckpoint() in estimator.fit()
Find the latest checkpoint with Estimator.latest_checkpoint()
And load weights from checkpoint with est.load_weights()


yangw1234:
Is it possible that we make it more general? Other callbacks may also need this. See the tensorboard callback https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/TensorBoard?hl=zh-cn

jenniew:
> Is it possible that we make it more general? Other callbacks may also need this. See the tensorboard callback https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/TensorBoard?hl=zh-cn

OK. Will try to make it general

jenniew:
Jenkins pass: http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-PR-Dllib-Validation/228/

yushan111:
I notice that there are both `save_weights` and `save`, as well as `load_weights` and `load` in pyspark estimator. 
When should user choose to use `save_weights` and `load_weights`, or `save` and `load`?

PR #3810, opened 2021-12-28
fix jekins failure

dding3:
Fix https://github.com/intel-analytics/arda-docker/issues/563

Seems keras softmax doesn't support tensor with dim>3 while tf.keras does.


dding3:
jekins pass. http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-PR-Dllib-Validation/188/

PR #3812, opened 2021-12-28
Upgrade dependency for Ray upgrade to 1.9.0

yushan111:
This PR includes
1. Upgrade dependencies in `setup`
   * In `dllib`, `numpy >=1.7` => `numpy >=1.19.5`
   * In `orca`, update ray dependencies to 1.9.0 in [ray] and [automl]
   * In `chronos`, upgrade pandas to 1.0.5
2. Add python micro version checking in `init_orca_context`. Raise Error if python micro version < .3

### DIFF
--- a/python/chronos/src/setup.py
+++ b/python/chronos/src/setup.py
@@ -54,7 +54,7 @@ def setup_package():
         url='https://github.com/intel-analytics/BigDL',
         packages=get_bigdl_packages(),
         install_requires=['bigdl-orca=='+VERSION, 'bigdl-nano[pytorch]',
-                          'pandas==1.0.5', 'scikit-learn', 'statsmodels'],
+                          'pandas>=1.0.5', 'scikit-learn', 'statsmodels'],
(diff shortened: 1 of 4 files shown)


python/chronos/src/setup.py
```diff
@@ -54,7 +54,7 @@ def setup_package():
         url='https://github.com/intel-analytics/BigDL',
         packages=get_bigdl_packages(),
         install_requires=['bigdl-orca=='+VERSION, 'bigdl-nano[pytorch]',
-                          'pandas==1.0.3', 'scikit-learn', 'statsmodels'],
+                          'pandas==1.0.5', 'scikit-learn', 'statsmodels'],
```
TheaperDeng, line 57: maybe we just relax this version to be `pandas>=1.0.5`?

TheaperDeng (approved):

python/chronos/src/setup.py
yushan111: done

yushan111:
http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-Orca-PR-Validation/259/

PR #3813, opened 2021-12-28
exclude dllib pom Vulnerabilities

Le-Zheng:


Le-Zheng:
Jenkins pass: http://10.112.231.51:18888/job/BigDL-PR-Dllib-Validation/214/

PR #3814, opened 2021-12-28
fix orca Vulnerabilities pom dependency

Le-Zheng:


Le-Zheng:
jenkins pass: http://10.112.231.51:18888/job/BigDL-Orca-PR-Validation/256/

PR #3816, opened 2021-12-29
Add upper version limit for pandas dependency

yushan111:
Pandas 1.3.5 will cause error
http://10.112.231.51:18888/job/BigDL-PRVN-chronos-Python-Spark-3.1-py37-ray-part2/171/console 

pandas 1.0.5 and 1.1.5 works.


TheaperDeng (commented):
@liangs6212 and I had tested that `<1.3.0` should be fine

TheaperDeng:
issue ref: https://stackoverflow.com/questions/58843656/pandas-interpolation-valueerrorinvalid-fill-method-expecting-pad-ffill-or

yushan111:
> @liangs6212 and I had tested that `<1.3.0` should be fine

Changed to `<1.3.0`

PR #3822, opened 2021-12-31
Add Pyspark DataFrame support for AutoXGBoost

yushan111:


python/orca/src/bigdl/orca/automl/search/ray_tune/ray_tune_search_engine.py
```diff
@@ -87,6 +87,8 @@ def compile(self,
                    a tuple in form of (x, y)
                         x: ndarray for training input
                         y: ndarray for training output
+               Spark Dataframe:
```
jason-dai, line 90: dataframe not supported here, no?

jason-dai:
LGTM; please verify on large data size (e.g., 100GBs)

yushan111:
Jenkins: All ray UT and examples passed.
http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-Orca-PR-Validation/277/

python/orca/src/bigdl/orca/automl/search/ray_tune/ray_tune_search_engine.py
yushan111: Yes, will address it in the next PR.

yushan111:
> LGTM; please verify on large data size (e.g., 100GBs)

Working on large data size. And results will be updated in issue #3823

jack1981:
I found inside the API , it still calls convert_df_to_ndarray function and inside the function , it calls toPandas() ! 
I would like to emphasize the original request was to remove toPandas() which still move data to Spark Driver , then what is the difference the user call toPandas() ? 

def convert_df_to_ndarray(data, feature_cols, target_cols):
        df = data.toPandas()
        X = df[feature_cols]
        y = df[target_cols]
        arrays = (X, y)
        return arrays

jack1981:
Would like to share this reference , Uber developed a RayEstimator on a Spark Dataframe , looks like this article 
https://eng.uber.com/elastic-xgboost-ray/

Highlights as below :
When we call Estimator.fit() on a Spark DataFrame, the Ray Estimator hides the following complexities:

- Materializes the Spark DataFrame to storage (if it is already cached, fetch its backing files via the DataFrame metadata)
- Manages any Ray Cluster set up, initialization, and necessary toolings
- Serializes and remotely executes the training function on the Ray Cluster
- Packages the trained binaries and artifacts into servable Transformers

jason-dai:
@jack1981 see https://github.com/intel-analytics/BigDL/issues/3823

PR #3840, opened 2022-01-06
Expose `remote_dir` for AutoXGBoost

yushan111:
This PR includes

- Expose `remote_dir` for `AutoXGBRegressor` and `AutoXGBClassifier`.
- Change permission for the default remote directory after `mkdir`, so that workers could have the write permission of the default remote directory.


yushan111:
http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-Orca-PR-Validation/296/

PR #3844, opened 2022-01-07
Chronos: Use `ray` output instead of `tqdm` for a clearer log in tuning

liangs6212:


liangs6212:
Effect Diagram

image

TODO: The "loc" column contains IP and Port information

python/orca/src/bigdl/orca/automl/search/ray_tune/ray_tune_search_engine.py
```diff
@@ -188,9 +189,10 @@ def run(self):
             search_alg=self._search_alg,
             num_samples=self.num_samples,
             trial_dirname_creator=trial_dirname_creator,
+            callbacks=[CustomProgressCallback()],
```
TheaperDeng, line 192: I think this Callback is kind of useless, we can remove it for now. Please correct me if it's critical

TheaperDeng (commented):
others LGTM

python/orca/src/bigdl/orca/automl/search/ray_tune/ray_tune_search_engine.py
liangs6212: CustomProgresscallback can avoid unnecessary log output.

TheaperDeng (approved):

liangs6212:
http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-Chronos-PR-Validation/211/

PR #3847, opened 2022-01-10
add sparse grad embeding

yangw1234:


yangw1234:
manually tested it, auto tests will be added later.

yangw1234:
http://10.112.231.51:18888/view/BigDL-2.0-PR/job/ZOO-PR-NanoTests/176/

python/nano/src/bigdl/nano/tf/keras/layers/embeddings.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def __init__(self,
+                 input_dim,
+                 output_dim,
+                 embeddings_initializer='uniform',
+                 regularizer=None,
+                 embeddings_constraint=None,
+                 mask_zero=False,
+                 input_length=None,
+                 **kwargs):
```
jason-dai, line 31: I think we should make the interface compatible with `keras.layers.Embedding`; that is, it accepts both `embeddings_regularizer` and `activity_regularizer`. If `embeddings_regularizer` is set, it will report warning, and use it as `activity_regularizer` (if the latter is not set).

yangw1234: Ok, I'll change it in a follow up pr.

PR #3851, opened 2022-01-10
Fix Lauguage Specific Issue when Physical Cores Detection

charlieJ107:
According to discussions in https://github.com/intel-analytics/BigDL/issues/3568：

When the system language is not English, physical cores information cannot be obtained due to the mismatch of the lookup with `grep` in `bigdl-nano-init`. This will cause the setting of some environment variables to fail. 

This PR uses `lscpu -p`  to filter out the parameters to a parsable format, and obtains the physical CPU information according to the parameters. This will make this process unaffected by the system language settings.

### DIFF
--- a/python/nano/script/bigdl-nano-init
+++ b/python/nano/script/bigdl-nano-init
@@ -128,8 +128,8 @@ if [ -f "${LIB_DIR}/libiomp5.so" ]; then
 	cpu_infos=($(lscpu -p=CPU,Socket,Core | grep -P '^(\d*),(\d*),(\d*)$'))
 	max_cpu_info=($(echo ${cpu_infos[-1]} | sed 's/,/\ /g')) 
 	let cpu_=${max_cpu_info[0]}+1
-    let sockets_=${max_cpu_info[1]}+1
-    let core_=${max_cpu_info[2]}+1
+	let sockets_=${max_cpu_info[1]}+1
+	let core_=${max_cpu_info[2]}+1
 	let threads_per_core=$cpu_/$core_
 	let cores_per_socket=$core_/$socket_
 
--- a/python/nano/script/bigdl-nano-init
+++ b/python/nano/script/bigdl-nano-init
@@ -125,10 +125,15 @@ if [ -f "${LIB_DIR}/libiomp5.so" ]; then
 	OPENMP=1
 
 	# detect number of physical cores
-	cores_per_socket=$(lscpu | grep "Core(s) per socket" | sed "s/.*://g" | sed -r "s/\s+//g")
-	threads_per_core=$(lscpu | grep "Thread(s) per core" | sed "s/.*://g" | sed -r "s/\s+//g")
-	sockets_=$(lscpu | grep "Socket(s)" | sed "s/.*://g" | sed -r "s/\s+//g")
-	
+	cpu_infos=($(lscpu -p=CPU,Socket,Core | grep -P '^(\d*),(\d*),(\d*)$'))
+	max_cpu_info=($(echo ${cpu_infos[-1]} | sed 's/,/\ /g')) 
+	let cpu_=${max_cpu_info[0]}+1
+    let sockets_=${max_cpu_info[1]}+1
+    let core_=${max_cpu_info[2]}+1
+	let threads_per_core=$cpu_/$core_
+	let cores_per_socket=$core_/$socket_
+
+
 	# set env variables
 	echo "Setting OMP_NUM_THREADS..."
 


python/nano/script/bigdl-nano-init
```diff
@@ -125,10 +125,15 @@ if [ -f "${LIB_DIR}/libiomp5.so" ]; then
...
 
 	# detect number of physical cores
-	cores_per_socket=$(lscpu | grep "Core(s) per socket" | sed "s/.*://g" | sed -r "s/\s+//g")
-	threads_per_core=$(lscpu | grep "Thread(s) per core" | sed "s/.*://g" | sed -r "s/\s+//g")
-	sockets_=$(lscpu | grep "Socket(s)" | sed "s/.*://g" | sed -r "s/\s+//g")
-	
+	cpu_infos=($(lscpu -p=CPU,Socket,Core | grep -P '^(\d*),(\d*),(\d*)$'))
+	max_cpu_info=($(echo ${cpu_infos[-1]} | sed 's/,/\ /g')) 
+	let cpu_=${max_cpu_info[0]}+1
+    let sockets_=${max_cpu_info[1]}+1
```
yangw1234, line 131: make the indentation consistent?

yangw1234:
@TheaperDeng would mind helping us verify this on your case?

yangw1234:
@charlieJ107 would you mind posting (maybe screenshot?) how do you test the code?

python/nano/script/bigdl-nano-init
charlieJ107: Fixed in eeb5cafe91ae34bef91005b9ce2414785bbb24bf

charlieJ107:
> @charlieJ107 would you mind posting (maybe screenshot?) how do you test the code?

In previous version, we use the following code to detect physical cores:
```shell
cores_per_socket=$(lscpu | grep "Core(s) per socket" | sed "s/.*://g" | sed -r "s/\s+//g")
threads_per_core=$(lscpu | grep "Thread(s) per core" | sed "s/.*://g" | sed -r "s/\s+//g")
sockets_=$(lscpu | grep "Socket(s)" | sed "s/.*://g" | sed -r "s/\s+//g")
```

Maybe we can test whether the values output by the old and new methods are consistent. 
Specifically, I wrote a test script like this:
```shell
#Previous method
cores_per_socket=$(lscpu | grep "Core(s) per socket" | sed "s/.*://g" | sed -r "s/\s+//g")
threads_per_core=$(lscpu | grep "Thread(s) per core" | sed "s/.*://g" | sed -r "s/\s+//g")
sockets_=$(lscpu | grep "Socket(s)" | sed "s/.*://g" | sed -r "s/\s+//g")

echo "cores_per_socket=$cores_per_socket"
echo "threads_per_core=$threads_per_core"
echo "sockets=$sockets_"

# New method
cpu_infos=($(lscpu -p=CPU,Socket,Core | grep -P '^(\d*),(\d*),(\d*)$'))
max_cpu_info=($(echo ${cpu_infos[-1]} | sed 's/,/\ /g')) 
let cpu_=${max_cpu_info[0]}+1
let sockets_=${max_cpu_info[1]}+1
let core_=${max_cpu_info[2]}+1
let threads_per_core=$cpu_/$core_
let cores_per_socket=$core_/$socket_

echo "cores_per_socket=$cores_per_socket"
echo "threads_per_core=$threads_per_core"
echo "sockets=$sockets_"
```

On clx-004, the output of this script like:
```shell
#Previous method
cores_per_socket=24
threads_per_core=2
sockets=2
# New method
cores_per_socket=24
threads_per_core=2
sockets=2
```

On cpx-04, the output of this script like:
```shell
#Previous method
cores_per_socket=28
threads_per_core=2
sockets=4
# New method
cores_per_socket=28
threads_per_core=2
sockets=4
```

yangw1234:
LGTM

TheaperDeng:
> @TheaperDeng would mind helping us verify this on your case?

I will test it on tomorrow's nightly build release

python/nano/script/bigdl-nano-init
```diff
@@ -125,10 +125,15 @@ if [ -f "${LIB_DIR}/libiomp5.so" ]; then
...
-	threads_per_core=$(lscpu | grep "Thread(s) per core" | sed "s/.*://g" | sed -r "s/\s+//g")
-	sockets_=$(lscpu | grep "Socket(s)" | sed "s/.*://g" | sed -r "s/\s+//g")
-	
+	cpu_infos=($(lscpu -p=CPU,Socket,Core | grep -P '^(\d*),(\d*),(\d*)$'))
+	max_cpu_info=($(echo ${cpu_infos[-1]} | sed 's/,/\ /g')) 
+	let cpu_=${max_cpu_info[0]}+1
+	let sockets_=${max_cpu_info[1]}+1
+	let core_=${max_cpu_info[2]}+1
+	let threads_per_core=$cpu_/$core_
+	let cores_per_socket=$core_/$socket_
```
TheaperDeng, line 134: should this be `let cores_per_socket=$core_/$sockets_` ?

PR #3856, opened 2022-01-10
INC tensorflow post-training quantization support

### LINKED ISSUE #3651: Nano: INC tensorflow post-training quantization support
Support tensorflow quantization using INC.
!!!Seems dynamic quantization is not available in tensorflow.

More details in #3600 .  Changes are quite similar to Pull request https://github.com/intel-analytics/BigDL/pull/3602.

zhentaocc:
resolves #3651 
- [x]  Enable INC post-training quantization on tensorflow
- [x]  Add tensorflow tests
- [x]  Fix found issues in pytorch part


zhentaocc:
@jason-dai @TheaperDeng @yangw1234 Please take a look.

zhentaocc:
Added new job http://10.112.231.51:18888/job/ZOO-PR-nano-tensorflow-test/. Passed. Will test again after review.

python/nano/test/run-nano-pytorch-tests.sh
```diff
@@ -2,18 +2,31 @@
...
+python -m pytest -s ${PYTORCH_NANO_TEST_DIR}/tests/ -k 'not ipex'
 
 now=$(date "+%s")
 time=$((now-start))
 
 echo "Bigdl-nano tests finished"
 echo "Time used:$time seconds"
 
+# Install ipex to test IPEX related modules
+pip install torch_ipex==1.9.0 -f https://software.intel.com/ipex-whl-stable
```
yangw1234, line 23: I think adding the installation here would be hard to maintain. I think you can added in the jenkins config.

python/nano/src/bigdl/nano/tf/keras/Model.py
```diff
@@ -13,11 +13,98 @@
...
 import tensorflow as tf
+from neural_compressor.model.model import TensorflowBaseModel
+from tensorflow.keras.metrics import Metric
 
 
 class Model(tf.keras.Model):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+
+    def quantize(self,
```
yangw1234, line 27: How about `Sequential`. I think you can use the MixedIn technique in Shane's pr #3791

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
```diff
@@ -176,9 +175,9 @@ def quantize(self, pl_model: LightningModule,
...
         :param val_dataloader:      A torch.utils.data.dataloader.DataLoader object for evaluation.
         :param metric:              A torchmetrics.metric.Metric object for evaluation.
-        :param backend:             'inc' or 'nncf'('nncf' is not supported yet). Default: 'inc'.
+        :param backend:             Only 'inc' is supported. Default: 'inc'.
         :param conf:        A path to conf yaml file for quantization.
-                            Default: '', use default config.
+                            Default: None, using default config.
         :param framework:   'pytorch', 'pytorch_fx', 'pytorch_ipex'. Default: 'pytorch_fx'.
                             Consistent with Intel Neural Compressor.
         :param approach:    'static' or 'dynamic'.
```
TheaperDeng, line 183: I think there is some typo in `accuracy_criterion`'s document we talked about before

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
```diff
@@ -216,24 +215,14 @@ def quantize(self, pl_model: LightningModule,
...
-            if quantized:
-                return quantized.model
-            else:
-                raise RuntimeError("Found no quantized model satisfying accuracy criterion.")
+                model = pl_model.model
+
+            TorchINCMetric.metric = metric
+            quantizer.post_training_quantize(model, calib_dataloader, val_dataloader,
+                                             TorchINCMetric)
+            return quantizer().model
```
TheaperDeng, line 226: why we need to call `quantizer()` again? I think it has been called in `quantizer.post_training_quantize`?

python/nano/src/bigdl/nano/tf/keras/Model.py
```diff
@@ -13,11 +13,98 @@
...
+
+            KerasINCMetric.metric = metric
+            val_loader = None
+            if val_dataset:
+                val_loader = common.DataLoader(val_dataset, batch)
+            calib_loader = None
+            if calib_dataset:
+                calib_loader = common.DataLoader(calib_dataset, batch)
+            quantizer.post_training_quantize(self, calib_loader, val_loader, KerasINCMetric)
+            return quantizer()
```
TheaperDeng, line 108: same question as above

zhentaocc: thx, fixed.

python/nano/src/bigdl/nano/tf/keras/Model.py
```diff
@@ -13,11 +13,98 @@
...
+                            Combine with max_trials field to decide when to exit.
+         :param max_trials: Max tune times. Default: 1.
+                            Combine with timeout field to decide when to exit.
+                            "timeout=0, max_trials=1" means it will try quantization only once and
+                            return satisfying best model.
+         :param inputs:     A list of input names.
+                            Default: None, automatically get names from graph.
+         :param outputs:    A list of output names.
+                            Default: None, automatically get names from graph.
+         :return:           A TensorflowBaseModel. If there is no model found, return None.
```
yangw1234, line 72: what is a `TensorflowBaseModel`?

python/nano/src/bigdl/nano/quantization/quantization_inc.py
```diff
@@ -60,27 +64,61 @@ def __init__(self,
...
+
+        if metric.metric and not val_dataloader:
+            raise RuntimeError("val_dataloader must be specified when metric is not None.")
 
-class TorchMetricForINC:
-    def __init__(self, metric):
-        self.metric = metric
+
+class INCMetric(ABC):
+    metric: Any = None
```
yangw1234, line 118: Why use a class variable? I think each instance of INCMetric should has its own metric variable.

python/nano/src/bigdl/nano/tf/keras/Model.py
zhentaocc: A class defined by INC wrapping a tensorflow graph inside.

python/nano/src/bigdl/nano/quantization/quantization_inc.py
zhentaocc: I agree, I tried to initiate a `INCMetric` by 
```python
INCMetric(metric=metric)
```
But in tensorflow, this will cause an error when dumping the configuration to yaml. So here I was intended to remove a tf or torch metric as attribute to avoid being dumped...

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
zhentaocc: Fixed.

python/nano/test/run-nano-pytorch-tests.sh
zhentaocc: Seperated ipex/non-ipex into two scripts.

python/nano/src/bigdl/nano/quantization/neural_compressor/quantization.py
```diff
@@ -60,44 +64,63 @@ def __init__(self,
...
-        self.metric = metric
-        self.pred_list = []
-        self.label_list = []
+                class MyMetric(framework_metric):
+                    def __init__(self):
+                        self.metric = metric
+                self.metric = Metric(
+                    MyMetric,
+                    name="%s_%s" % (framework_metric.__name__, type(metric).__name__)
+                )
```
zhentaocc, line 102: Fix class variable in INCMetric by creating a local class inside. Seems better?

python/nano/src/bigdl/nano/tf/keras/Model.py
zhentaocc: Accepted your suggestions in new version. Seems that `Sequential` inherit `Model` according to https://www.tensorflow.org/api_docs/python/tf/keras/Sequential, maybe there is another way to extend original keras layers.

zhentaocc: I tried that method, seems that `Model` will be broken when inc is not installed for quantization. This is not expected since inc is an option.  
So I bind this function to `tf.keras.Model` as below in `__init__.py`.
```python
try:
    from ..quantization import quantize
    tf.keras.Model.quantize = quantize
except ImportError:
    print("Warning: Intel Neural Compressor should be installed if you need quantization.")
```
And `tf.keras.Sequential` is subclass of `tf.keras.Model`. So we don't need to make extra changes on `Sequential`.  
It occurs to me if we should keep these optional extensions like inc, onnx, ipex in their own folder and test both when they are installed and not installed to make sure everything works fine in real cases.

python/nano/src/bigdl/nano/quantization/neural_compressor/quantization.py
```diff
@@ -60,44 +64,63 @@ def __init__(self,
...
+                    framework_metric = METRICS['pytorch']
+                else:
+                    framework_metric = METRICS[framework]
 
-class TorchMetricForINC:
-    def __init__(self, metric):
-        self.metric = metric
-        self.pred_list = []
-        self.label_list = []
+                class MyMetric(framework_metric):
```
yangw1234, line 96: `MyMetric` looks weird. Maybe briefly document why do you need this class here?

zhentaocc: Added document.

python/nano/src/bigdl/nano/tf/keras/__init__.py
```diff
@@ -13,7 +13,12 @@
...
 # limitations under the License.
 #
+import tensorflow as tf
 
-
+try:
+    from ..quantization import quantize
+    tf.keras.Model.quantize = quantize
+except ImportError:
+    print("Warning: Intel Neural Compressor should be installed if you need quantization.")
```
yangw1234, line 22: Maybe adding how to install INC in the message？

yangw1234: If a user does not install INC, it will always get this warning, right? I think a better behavior is raising an error when a user tries to use quantization.

yangw1234:
> Added new job http://10.112.231.51:18888/job/ZOO-PR-nano-tensorflow-test/. Passed. Will test again after review.

This test should be added to http://10.112.231.51:18888/view/BigDL-2.0-PR/job/ZOO-PR-NanoTests/

zhentaocc:
> > Added new job http://10.112.231.51:18888/job/ZOO-PR-nano-tensorflow-test/. Passed. Will test again after review.
> 
> This test should be added to http://10.112.231.51:18888/view/BigDL-2.0-PR/job/ZOO-PR-NanoTests/

Will be added after merged.
Jenkins: 
http://10.112.231.51:18888/job/ZOO-PR-nano-tensorflow-test/ Passed
http://10.112.231.51:18888/job/ZOO-PR-NanoTests/193/ Passed

python/nano/src/bigdl/nano/tf/keras/__init__.py
zhentaocc: Accepted in new commits.

yangw1234:
image

these two configurations should be set automatically

yangw1234:
http://10.112.231.51:18888/job/ZOO-PR-nano-tensorflow-test/18 Passed
http://10.112.231.51:18888/job/ZOO-PR-NanoTests/194/ Passed

yangw1234:
LGTM

PR #3870, opened 2022-01-12
Adding a sparseadam optimizer file

DirkFi:


python/nano/src/bigdl/nano/tf/optimizers/sparse_adam.py
```diff
@@ -0,0 +1,204 @@
+"""SparseAdam optimizer implementation."""
+
```
yangw1234, line 2: add license

python/nano/src/bigdl/nano/tf/optimizers/sparse_adam.py
```diff
@@ -0,0 +1,204 @@
...
+import tensorflow.compat.v2 as tf
+from keras import backend_config
+from keras.optimizer_v2 import optimizer_v2
+from tensorflow.python.util.tf_export import keras_export
+import tensorflow
+
+
+class SparseAdam(tensorflow.keras.optimizers.Adam):
+    r"""Optimizer that implements the SparseAdam algorithm.
```
yangw1234, line 11: let's briefly describe the algorithm here and use the comment format similar to other python in bigdl

python/nano/src/bigdl/nano/tf/optimizers/sparse_adam.py
```diff
@@ -0,0 +1,204 @@
...
+
+  Args:
```
yangw1234, line 13: Let's use the format consistent with other files. See https://github.com/intel-analytics/BigDL/blob/branch-2.0/python/nano/src/bigdl/nano/tf/keras/layers/embeddings.py#L38

python/nano/src/bigdl/nano/tf/optimizers/sparse_adam.py
```diff
@@ -0,0 +1,204 @@
...
+        config = super(SparseAdam, self).get_config()
+        config.update({
+            'learning_rate': self._serialize_hyperparameter('learning_rate'),
+            'decay': self._initial_decay,
+            'beta_1': self._serialize_hyperparameter('beta_1'),
+            'beta_2': self._serialize_hyperparameter('beta_2'),
+            'epsilon': self.epsilon,
+            'amsgrad': self.amsgrad,
+        })
+        return config
```
yangw1234, line 204: you need to add a empty line here

yangw1234:
unit tests should be added

DirkFi:
OK

python/nano/src/bigdl/nano/tf/optimizers/sparse_adam.py
```diff
@@ -0,0 +1,134 @@
...
+    def __init__(self,
+                 learning_rate=0.001,
+                 beta_1=0.9,
+                 beta_2=0.999,
+                 epsilon=1e-7,
+                 amsgrad=False,
+                 name='SparseAdam',
+                 **kwargs):
+        '''
+        This is a slightly modified version of tf.keras.Embedding,
```
yangw1234, line 50: This is not correct, right? This is not a `Embedding`

python/nano/src/bigdl/nano/tf/optimizers/test_optimizers.py
```diff
@@ -0,0 +1,41 @@
...
+    test_array = tf.nn.embedding_lookup(input_array, ids)
+    print(test_array)
+    with tf.GradientTape() as tape:
+        output = model(test_array)
+        labels = tf.zeros_like(output)
+        loss_value = model.compiled_loss(output, labels, regularization_losses=model.losses)
+        grads = tape.gradient(loss_value, model.trainable_variables)
+        for grad in grads:
+            print("grad:", grad)
+            assert isinstance(grad, tf.IndexedSlices)
```
yangw1234, line 41: This test does not check the `SparseAdam` you implemented. This assertion only checks if the grad is sparse. You should check after training one iteration if only some of the slots changed in the embedding.

python/nano/src/bigdl/nano/tf/optimizers/test_optimizers.py
```diff
@@ -0,0 +1,41 @@
...
+    from bigdl.nano.tf.keras import Sequential
+    import tensorflow as tf
+    model = Sequential()
+    model.add(Embedding(1000, 64, input_length=10, regularizer=tf.keras.regularizers.L2()))
+    sparse_optim = SparseAdam()
+    model.compile(sparse_optim, 'mse')
+    input_array = np.random.randint(1000, size=(32, 10))
+
+    ids = tf.constant([0, 10, 20, 30])
+    test_array = tf.nn.embedding_lookup(input_array, ids)
```
yangw1234, line 32: the output of the `Embedding` layer is already the result of a lookup, I do not think you should look up again.
see the documentation of https://www.tensorflow.org/api_docs/python/tf/keras/layers/Embedding

python/nano/src/bigdl/nano/tf/optimizers/test_optimizers.py
```diff
@@ -0,0 +1,41 @@
...
+def test_optimizer_sparseadam():
+
+    from bigdl.nano.tf.keras.layers import Embedding
+    from bigdl.nano.tf.keras import Sequential
+    import tensorflow as tf
+    model = Sequential()
+    model.add(Embedding(1000, 64, input_length=10, regularizer=tf.keras.regularizers.L2()))
+    sparse_optim = SparseAdam()
+    model.compile(sparse_optim, 'mse')
+    input_array = np.random.randint(1000, size=(32, 10))
```
yangw1234, line 29: I recommend you explicitly construct the input array. Let's the input length is 5, and  batch_size is 2. You can construct  you input as the following
```
np.array([[0, 1, 2, 3, 4], [5, 6, 7, 8, 9]])
```
then after training one batch, only the first 10 rows of the embedding matrix should be changed.

You can use `train_on_batch` (see https://www.tensorflow.org/api_docs/python/tf/keras/Model#train_on_batch) to train one batch.
You can use `get_weights` (see https://www.tensorflow.org/api_docs/python/tf/keras/layers/Layer#get_weights) before and after `train_on_batch` and compare the results to see if the weights change as expected.

DirkFi: Alright. I' try it.

python/nano/src/bigdl/nano/tf/optimizers/sparse_adam.py
```diff
@@ -0,0 +1,134 @@
...
+                 learning_rate=0.001,
+                 beta_1=0.9,
+                 beta_2=0.999,
+                 epsilon=1e-7,
+                 amsgrad=False,
+                 name='SparseAdam',
+                 **kwargs):
+        '''
+        This is a slightly modified version of tf.keras.optimizers.Adam,
+        which only apply resource_apply_sparse to the output of the embedding
```
yangw1234, line 51: "which only update moving-average accumulators for sparse variable indices that appear in the current batch"

python/nano/src/bigdl/nano/tf/optimizers/test_optimizers.py
```diff
@@ -0,0 +1,40 @@
+# Copyright 2016 The BigDL Authors.
```
yangw1234, line 1: you should add test file to the test directory.
see https://github.com/intel-analytics/BigDL/tree/branch-2.0/python/nano/test/tf/keras

python/nano/src/bigdl/nano/tf/optimizers/sparse_adam.py
```diff
@@ -0,0 +1,134 @@
...
+
+        with tf.name_scope("update_sparse_mt"):
+            m = self.get_slot(var, 'm')
+            m_scaled_g_values = grad * coefficients['one_minus_beta_1_t']
+            m_sparse = tf.gather(m, indices=indices)
+
+            m_sparse_scaled = m_sparse * (coefficients['beta_1_t'] - 1)
+
+            m_t = self._resource_scatter_add(m, indices, m_scaled_g_values + m_sparse_scaled)
+            m_t_sparse = m_sparse * coefficients['beta_1_t'] + m_sparse_scaled
```
yangw1234, line 101: there is a bug here.

This line should be 
```python
m_t_sparse = m_sparse * coefficients['beta_1_t'] + m_scaled_g_values
```

DirkFi:
PR validation is here [http://10.112.231.51:18888/view/BigDL-2.0-PR/job/ZOO-PR-NanoTests/197/](#197). And this validation is still running though.

DirkFi:
The newest update with PR validation is [here#206](http://10.112.231.51:18888/view/BigDL-2.0-PR/job/ZOO-PR-NanoTests/206/).

PR #3871, opened 2022-01-13
fix shuffle in target codes for skewed data

songhappy:
target_code has a big table join a relatively small table. the big table is always skewed because feature data is always long-tailed. Regular join will pull most of the data into certain nodes and fails; broadcast join will add more data into cluster, but the size is unknown. 
The solution here is, split the data into two parts, broadcast join the top 1m features,  and regular join the left over


python/friesian/src/bigdl/friesian/feature/utils.py
```diff
@@ -160,13 +162,29 @@ def encode_target_(tbl, targets, target_cols=None, drop_cat=True, drop_fold=True
                     new_out_target_mean[out_col] = target_mean
             out_target_mean = new_out_target_mean
 
+        all_size = target_code.size()
+        limit_size = 1000000
```
hkvision, line 166: how this limit size is determined?

hkvision:
Please add some explanations for this PR and how the modification works.

python/friesian/src/bigdl/friesian/feature/utils.py
songhappy: most of the times, size of each categorical feature stay within 1m.

songhappy:
Jenkins passed: 
http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-PR-Friesian-Validation/198/

hkvision (approved):

Issue #3841, opened 2022-01-06
TF2Estimator on Pyspark backend got error on Spark 2.4.0: 'TaskContext' object has no attribute 'getTaskInfos'


jenniew:
When TF2Estimator on Pyspark backend on Spark 2.4.0, got this error: AttributeError: 'TaskContext' object has no attribute 'getTaskInfos'.

Traceback (most recent call last):
  File "/ads_storage/udap/ray/ncf_zoo_tf2.py", line 181, in <module>
    validation_steps=2)
  File "/ads_storage/udap/ray/ray-venv/lib/python3.7/site-packages/bigdl/orca/learn/tf2/pyspark_estimator.py", line 138, in fit
    cluster_info=self._get_cluster_info(sc),
  File "/ads_storage/udap/ray/ray-venv/lib/python3.7/site-packages/bigdl/orca/learn/tf2/pyspark_estimator.py", line 89, in _get_cluster_info
    cluster_info = self.workerRDD.barrier().mapPartitions(find_ip_and_port).collect()
  File "/opt/cloudera/parcels/CDH-6.3.4-1.cdh6.3.4.p4623.11628701/lib/spark/python/lib/pyspark.zip/pyspark/rdd.py", line 816, in collect
  File "/opt/cloudera/parcels/CDH-6.3.4-1.cdh6.3.4.p4623.11628701/lib/spark/python/lib/py4j-0.10.7-src.zip/py4j/java_gateway.py", line 1257, in __call__
  File "/opt/cloudera/parcels/CDH-6.3.4-1.cdh6.3.4.p4623.11628701/lib/spark/python/lib/pyspark.zip/pyspark/sql/utils.py", line 63, in deco
  File "/opt/cloudera/parcels/CDH-6.3.4-1.cdh6.3.4.p4623.11628701/lib/spark/python/lib/py4j-0.10.7-src.zip/py4j/protocol.py", line 328, in get_return_value
py4j.protocol.Py4JJavaError: An error occurred while calling z:org.apache.spark.api.python.PythonRDD.collectAndServe.
: org.apache.spark.SparkException: Job aborted due to stage failure: Could not recover from a failed barrier ResultStage. Most recent failure reason: Stage failed because barrier task ResultTask(102, 9) finished unsuccessfully.
org.apache.spark.api.python.PythonException: Traceback (most recent call last):
[...]

PR #3877, opened 2022-01-14
Fix BarrierTaskcontext issue in TF2Estimator on Spark 2.4.0

jenniew:
Spark 2.4.0 has a bug for BarrierTaskcontext.
Fix TF2Estimator to avoid the bug.
Related issue: https://github.com/intel-analytics/BigDL/issues/3841


jenniew:
Test passed on yarn cluster with Spark 2.4.0, 2.4.3, 3.1.2.

jason-dai:
This should apply to PyTorch as well, yes? @yushan111

python/orca/src/bigdl/orca/learn/tf2/spark_runner.py
```diff
@@ -191,9 +181,7 @@ def _handle_batch_size(self, config):
 
 
 def find_ip_and_port(pre_iter):
-    tc = BarrierTaskContext().get()
```
yangw1234, line 194: If there is no synchronization when finding free ports, I think there is a small chance that different tasks in the same machine will find the same free port. For example,

Task A in machine 1, runs into `find_free_port` and binds port 1234 and releases the port when return.
Then Task B in machine 1, which might be a little slower than Task A, could try to find a free port after Task A's `find_free_port` returns and finds 1234 again. 

This could result in different TensorFlow instances binding the same ip:port.

yushan111:
> This should apply to PyTorch as well, yes? @yushan111

Yes. Will change correspondingly in PyTorch after this PR merge.

python/orca/src/bigdl/orca/learn/tf2/spark_runner.py
jenniew: Yes, I changed back to use tc again.

jenniew:
> > This should apply to PyTorch as well, yes? @yushan111
> 
> Yes. Will change correspondingly in PyTorch after this PR merge.y

@yushan111 I moved these methods to utils so as to avoid duplicated code and modification. You can change to call these utils functions.

yushan111:
> > > This should apply to PyTorch as well, yes? @yushan111
> > 
> > 
> > Yes. Will change correspondingly in PyTorch after this PR merge.y
> 
> @yushan111 I moved these methods to utils so as to avoid duplicated code and modification. You can change to call these utils functions.

Thanks!

jenniew:
Jenkins pass: http://10.112.231.51:18888/job/BigDL-Orca-PR-Validation/334/

PR #3890, opened 2022-01-17
FGBoost functionality and Correctness unit test

Litchilitchy:


jason-dai:
Why are you changing HFL, VFL, NN, etc., in FGBoost PR?

Litchilitchy:
> Why are you changing HFL, VFL, NN, etc., in FGBoost PR?

* The `Table` proto changes so the NN proto based on it is also change
* During this PR, seems some upstream changes from Dllib make implicit type not applicable, so something like `Adam[Float]` is added.

jason-dai:
> > Why are you changing HFL, VFL, NN, etc., in FGBoost PR?
> 
> * The `Table` proto changes so the NN proto based on it is also change
Make this a separate PR

> * During this PR, seems some upstream changes from Dllib make implicit type not applicable, so something like `Adam[Float]` is added.
Do we have unit test for this? Make this a separate PR

> * some DataFrame API generalization makes the intermediate type of NN change.
Make this a separate PR

Litchilitchy:
@jason-dai Noted. This PR will be split to following PRs in sequential.
* proto refactor
* Adam[Float], dllib related changes & fgboost functionality (these two has no dependencies and could create at same time)
* DataFrame API refactor and new functionality (the fgboost correctness test will use not refactored code for now)

PR #3895, opened 2022-01-18
Add Performance Finetune to fit in nano.tf.keras.Model

charlieJ107:


python/nano/src/bigdl/nano/tf/keras/Model.py
```diff
@@ -13,11 +13,54 @@
...
+                                   shuffle, class_weight,
+                                   sample_weight, initial_epoch, steps_per_epoch, validation_steps,
+                                   validation_batch_size, validation_freq,
+                                   max_queue_size, workers, use_multiprocessing)
+            batch_size_12 = int(batch_size / 4)
+            super(Model, self).fit(x, y, batch_size_12, epochs, verbose, callbacks, validation_split, validation_data,
+                                   shuffle, class_weight,
+                                   sample_weight, initial_epoch, steps_per_epoch, validation_steps,
+                                   validation_batch_size, validation_freq,
+                                   max_queue_size, workers, use_multiprocessing)
```
yangw1234, line 60: I think it is better to print a summary line about each trails' performance at the end.

python/nano/src/bigdl/nano/tf/keras/Model.py
```diff
@@ -13,11 +13,54 @@
...
+
+        else:
+            super(Model, self).fit(
```
yangw1234, line 63: I think you should add `return ` here

python/nano/src/bigdl/nano/tf/keras/Model.py
```diff
@@ -13,11 +13,54 @@
...
+            initial_epoch=0,
+            steps_per_epoch=None,
+            validation_steps=None,
+            validation_batch_size=None,
+            validation_freq=1,
+            max_queue_size=10,
+            workers=1,
+            use_multiprocessing=False,
+            perf_tune=None):
+        if perf_tune is not None:
```
yangw1234, line 44: How about making perf_tune a boolean argument, and we will tuning all of `batch_size`, `callbacks` and `dataset`

charlieJ107: Or I add an option "All" to run all of the tunning task.

jason-dai:
@shane-huang who is working on the tuning support in nano

python/nano/src/bigdl/nano/tf/keras/Model.py
```diff
@@ -13,11 +13,92 @@
...
+            if batch_size is None:
+                batch_size = 32
+            # Data pipeline
+            print("Prefetch Datapipeline")
+            assert issubclass(
+                type(x), tf.data.Dataset), "Data object currently only support tf.data.dataset"
+
+            batched_data=[]
+            batched_target = []
+            for i in range(batch_size):
```
yangw1234, line 56: dataset is already assumed to be batched. see https://www.tensorflow.org/api_docs/python/tf/keras/Model#fit

python/nano/src/bigdl/nano/tf/keras/Model.py
yangw1234: 1. `dataset = dataset.take(1).cache().repeat(100)` cache the first elements and repeat to create enough data for 100 iterations, this step is to see the effects of dataset to throughput
2.  `dataset = dataset.unbatch().batch(new_batch_size).cache().repeat(100)`, compare to 1 to see the effects of batch_size on throughput

python/nano/src/bigdl/nano/tf/keras/Model.py
yangw1234: 1. `dataset = dataset.take(1).cache().repeat(100)` cache the first elements and repeat to create enough data for 100 iterations, this step is to see the effects of dataset to throughput
2.  `dataset = dataset.unbatch().batch(new_batch_size).cache().repeat(100)`, compare to 1 to see the effects of batch_size on throughput

python/nano/src/bigdl/nano/tf/keras/Model.py
```diff
@@ -13,11 +13,92 @@
...
+                batched_data_item, batched_target_item = next(iter(x))
+                # batched_data.append((batched_data_item, batched_target_item))
+                batched_data.append(batched_data_item)
+                batched_target.append(batched_target_item)
+            start = time()
+            super(Model, self).fit(
```
yangw1234, line 62: Could you restructure the code to make it easy to understand?
Such as writing the logic in a loop. Inside the loop, you first generate the modified arguments and then call `.fit` using the generated arguments. So that we only see this `super(Model, self).fit` once in the source code. (The DRY principle)

python/nano/src/bigdl/nano/tf/keras/Model.py
```diff
@@ -13,11 +13,71 @@
...
+            ]
+            
+            for x in dataList:
+                start = time()
+                super(Model, self).fit(
+                x, y, batch_size, epochs, verbose, callbacks, validation_split, validation_data, shuffle, class_weight,
+                    sample_weight, initial_epoch, steps_per_epoch, validation_steps, validation_batch_size, validation_freq,
+                    max_queue_size, workers, use_multiprocessing)
+                end = time()
+                print("Throughput: ", len(x)/(end-start))
```
yangw1234, line 65: Maybe you do not need to use `len(x)`, I think `len(x)` is just `100 * batch_size`

charlieJ107: The dataset size become UNKONWN after `unbatch()` called. We cannot get the size of the dataset before the epoch finished: 
image


After the epoch finished, maybe we can get the size information of the dataset through the progress bar displayed:
image

But I still not sure about the relationship between the dataset size and new batch_size. Should I use hard encoded number when calculating throughput?

charlieJ107:
Manually tested on cpx-04:
image

PR #3898, opened 2022-01-18
Nano: add quantized model save&load API

TheaperDeng:
### Save and load
reference to https://pytorch.org/tutorials/beginner/saving_loading_models.html#saving-loading-model-for-inference

#### save
```python
# torch model
torch.save(model.state_dict(), PATH)  # fp32 pytorch model
torch.save(model.quantized_state_dict(), PATH)  # int8 pytorch model
```

#### load

since onnx file is easy to load we only consider torch model.

```python
# torch model
model = TheModelClass(*args, **kargs)
model = Trainer.compile(model)
model.load_state_dict(torch.load(PATH))  # fp32
model.load_quantized_state_dict(torch.load(PATH))  # int8
```

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/runtime_binding/quantization_inference.py
+++ b/python/nano/src/bigdl/nano/pytorch/runtime_binding/quantization_inference.py
@@ -15,7 +15,6 @@
 #
 
 import warnings
-import pickle
 from functools import partial
 from bigdl.nano.pytorch.lightning import LightningModuleFromTorch
 
--- a/python/nano/src/bigdl/nano/pytorch/runtime_binding/quantization_inference.py
+++ b/python/nano/src/bigdl/nano/pytorch/runtime_binding/quantization_inference.py
@@ -47,7 +47,7 @@ def _fx_quantize_eval(self, quantize=False):
             self.forward = self._forward_fx_quantize
         else:
             raise RuntimeError("Please call trainer.quantize again since the quantized model is"
-                               "not up-to-date")
+                               " not up-to-date")
     else:
         self.forward = self._torch_forward
 
@@ -57,7 +57,7 @@ def quantized_state_dict(self):
         return self._quantized_model.state_dict()
     else:
         raise RuntimeError("Please call trainer.quantize again since the quantized model is"
-                           "not up-to-date.")
+                           " not up-to-date.")
 
 
 def load_quantized_state_dict(self, state_dict):
@@ -77,8 +77,7 @@ def load_quantized_state_dict(self, state_dict):
     if back_to_train:
         self.train()
     self._quantized_model = qmodel
-    self._quantized_model_up_to_date = True  # set to true
-
+    self._quantized_model_up_to_date = True  # set to true before training next time
 
 def bind_quantize_methods(pl_model, q_model):
     # check conflicts
--- a/python/nano/src/bigdl/nano/pytorch/runtime_binding/quantization_inference.py
+++ b/python/nano/src/bigdl/nano/pytorch/runtime_binding/quantization_inference.py
@@ -57,13 +57,15 @@ def quantized_state_dict(self):
         return self._quantized_model.state_dict()
     else:
         raise RuntimeError("Please call trainer.quantize again since the quantized model is"
-                           "not up-to-date")
+                           "not up-to-date.")
 
 
 def load_quantized_state_dict(self, state_dict):
     import torch
     from torch.quantization.quantize_fx import prepare_fx, convert_fx
-    self.eval()  # TODO: put back to it's original state
+    back_to_train = self.training
+
+    self.eval()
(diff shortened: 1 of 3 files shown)


TheaperDeng:
http://10.112.231.51:18888/job/ZOO-PR-NanoTests/211/
http://10.112.231.51:18888/job/Chronos-PR-Validation-Spark-3.1/299/

TheaperDeng:
merge this PR and will change according to comments if any in next PR

jason-dai:
- Currently the user will write the following code:
   ```python
   model = trainer.compile(model, loss, optimizer) #no need to set quantize=true
   mode = trainer.quantize(model, ...);
   ```
  and
   ```python
   model = trainer.compile(model, quantize=true) #need to set quantize=true
   mode.model.load_quantized_state_dict(...)
   ```

- I think we need make the APIs a little bit more consistent: either the user needs to set `quantize=true` in both cases, or the user does not need to set it at all.

TheaperDeng:
> * Currently the user will write the following code:
>   ```python
>   model = trainer.compile(model, loss, optimizer) #no need to set quantize=true
>   mode = trainer.quantize(model, ...);
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   and
>   ```python
>   model = trainer.compile(model, quantize=true) #need to set quantize=true
>   mode.model.load_quantized_state_dict(...)
>   ```
> * I think we need make the APIs a little bit more consistent: either the user needs to set `quantize=true` in both cases, or the user does not need to set it at all.

I think I can make quantize methods always bonded to the model. But right now `quantize` parameter is set to `True` by default, users don't need to set it explicitly. Do we really need to remove this paramter?

jason-dai:
> > * Currently the user will write the following code:
> >   ```python
> >   model = trainer.compile(model, loss, optimizer) #no need to set quantize=true
> >   mode = trainer.quantize(model, ...);
> >   ```
> >   
> >   
> >       
> >         
> >       
> >   
> >         
> >       
> >   
> >       
> >     
> >   and
> >   ```python
> >   model = trainer.compile(model, quantize=true) #need to set quantize=true
> >   mode.model.load_quantized_state_dict(...)
> >   ```
> > * I think we need make the APIs a little bit more consistent: either the user needs to set `quantize=true` in both cases, or the user does not need to set it at all.
> 
> I think I can make quantize methods always bonded to the model. But right now `quantize` parameter is set to `True` by default, users don't need to set it explicitly. Do we really need to remove this paramter?

Then why would the user set it to false?

TheaperDeng:
> > > * Currently the user will write the following code:
> > >   ```python
> > >   model = trainer.compile(model, loss, optimizer) #no need to set quantize=true
> > >   mode = trainer.quantize(model, ...);
> > >   ```
> > >   
> > >   
> > >       
> > >         
> > >       
> > >   
> > >         
> > >       
> > >   
> > >       
> > >     
> > >   and
> > >   ```python
> > >   model = trainer.compile(model, quantize=true) #need to set quantize=true
> > >   mode.model.load_quantized_state_dict(...)
> > >   ```
> > > * I think we need make the APIs a little bit more consistent: either the user needs to set `quantize=true` in both cases, or the user does not need to set it at all.
> > 
> > 
> > I think I can make quantize methods always bonded to the model. But right now `quantize` parameter is set to `True` by default, users don't need to set it explicitly. Do we really need to remove this paramter?
> 
> Then why would the user set it to false?

neural-compressor is still not our forced requirement library for Nano. This parameter is for users who want to remove any redundant quantization-related monkey patch on their model since they don't need them. Once this happens, we can ask our users to turn this parameter to false.

jason-dai:
> > > > * Currently the user will write the following code:
> > > >   ```python
> > > >   model = trainer.compile(model, loss, optimizer) #no need to set quantize=true
> > > >   mode = trainer.quantize(model, ...);
> > > >   ```
> > > >   
> > > >   
> > > >       
> > > >         
> > > >       
> > > >   
> > > >         
> > > >       
> > > >   
> > > >       
> > > >     
> > > >   and
> > > >   ```python
> > > >   model = trainer.compile(model, quantize=true) #need to set quantize=true
> > > >   mode.model.load_quantized_state_dict(...)
> > > >   ```
> > > > * I think we need make the APIs a little bit more consistent: either the user needs to set `quantize=true` in both cases, or the user does not need to set it at all.
> > > 
> > > 
> > > I think I can make quantize methods always bonded to the model. But right now `quantize` parameter is set to `True` by default, users don't need to set it explicitly. Do we really need to remove this paramter?
> > 
> > 
> > Then why would the user set it to false?
> 
> neural-compressor is still not our forced requirement library for Nano. This parameter is for users who want to remove any redundant quantization-related monkey patch on their model since they don't need them. Once this happens, we can ask our users to turn this parameter to false.

Then maybe you may bind quantization related methods on demand (i.e., when the user calls `quantize` or `load_quantized_state_dict`).

PR #3906, opened 2022-01-19
PPML rename DataType and package refactor

Litchilitchy:


scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/base/StorageHolder.scala
```diff
@@ -16,45 +16,63 @@
...
+      logger.debug(s"Put TreeEval into client data map, current size: $clientDataSize")
+    } else {
+      throw new IllegalArgumentException("Data is empty, could not uploaded to server.")
     }
   }
-  def getTableStorage() = this.tableStorage
-  def getSplitStorage() = this.splitStorage
-  def getLeafStorage() = this.leafStorage
-  def getBranchStorage() = this.branchStorage
+  def getTableStorage() = this.tensorMapStorage
```
jason-dai, line 74: getTensorMapStorage

scala/ppml/src/main/java/com/intel/analytics/bigdl/ppml/common/FLDataType.java
```diff
@@ -19,6 +19,6 @@
 public enum FLDataType {
     TENSOR_MAP,
     TREE_SPLIT,
-    TREE_LEAF,
+    TREE_LEAVES,
```
jason-dai, line 22: Why TREE_LEAVES instead of TREE_LEAF?

Litchilitchy: This actually store an array of leaves. However this is just a naming and we'd better fix all the names the same.

jason-dai: treeLeavesStorage sounds very awkward

Litchilitchy: How about change the type to `TREE_LEAF` and `getTreeLeafStorage()` and remain the proto `TreeLeaves'

jason-dai: I think we should keep them consistent - either LEAF or LEAVES for all

Litchilitchy: So change to LEAF for all?

PR #3908, opened 2022-01-19
Chronos: change critical test for chronos-lite

TheaperDeng:
remove redundant import in chronos-lite functionalities. so that chronos-lite can be tested on our exact same ut.


TheaperDeng:
http://10.112.231.51:18888/job/BigDL-Chronos-PR-Validation/259/

PR #3914, opened 2022-01-20
Fix BarrierTaskcontext issue in PytorchEstimator on Spark 2.4.0

yushan111:
Related to PR #3877


yushan111:
http://10.112.231.51:18888/job/BigDL-Orca-PR-Validation/347/

PR #3915, opened 2022-01-20
Update bash.manifest.template and start-all.sh

zzti-bsj:
1. update bash.manifest.template
2. update start-all.sh


glorysdj (approved):
LGTM

PR #3917, opened 2022-01-21
[PPML] Modify Files about KMS

Uxito-Ada:
1. Modify kMS to upper KMS.
2. Remove .db transformation in the workflow chart.
4. Rename the dependency jar name.


glorysdj (approved):
LGTM

PR #3921, opened 2022-01-21
[PPML] Append some text to kms-client/README.md

Uxito-Ada:
1. Add parameter SECURE_PASSWORD_PATH, which is needed by Spark job.
2. Add SGX initiation command.


glorysdj (approved):
LGTM

PR #3923, opened 2022-01-21
change parameters order for channelnormalize to BGR

dding3:


dding3:
jekins pass. http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-PR-Dllib-Validation/247/

jason-dai:
OpenCV uses BGR, and Pillow uses RGB; how can we support both cases?

dding3:
We have some `transformer`s support convert BGR to RGB (eg. set `ImageMatToTensor(toRGB=True)`)?

Litchilitchy:
Maybe we can add pillow and opencv-python related UT for this later?

PR #3924, opened 2022-01-21
TF2Estimator on Pyspark add shard size support

jenniew:


jenniew:
Jenkins pass: http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-Orca-PR-Validation/354/

PR #3933, opened 2022-01-24
Chronos: add quantization method to forecaster

TheaperDeng:
- [x] method and ut
- S2S has not support this method. (will be solved later)
- [x] save and load should be changed later.
- [x] Jenkins env should be changed to fit the requirement
- [x] document


TheaperDeng:
http://10.112.231.51:18888/job/BigDL-Chronos-PR-Validation/294/
http://10.112.231.51:18888/job/ZOO-PR-NanoTests/221/

PR #3935, opened 2022-01-24
fix bigdl master docs issue

pinggao187:


pinggao187:
jenkins test：
http://10.112.231.51:18888/job/BigDL-PR-Doc-Generation-Test/4/console

PR #3938, opened 2022-01-24
docker image jupyter 

Le-Zheng:


docker/bigdl/Dockerfile
```diff
@@ -127,7 +127,4 @@ RUN apt-get update --fix-missing && \
 RUN echo "source activate bigdl" > ~/.bashrc
 ENV PATH /usr/local/envs/bigdl/bin:$PATH
 
-ADD ./start-notebook.sh /opt/work
-RUN chmod a+x /opt/work/start-notebook.sh
```
hkvision, line 131: Can we add the jupyter notebook command here to start automatically?

Le-Zheng: Shall we add this jupyter notebook command in start-notebook.sh? and remove the previous jupyter spark local commond?

hkvision: If the other environment variables and settings in the script is not necessary, we can just add the one line `jupyter notebook ....` in the Dockerfile directly?

Le-Zheng: I don't think we can add `jupyter notebook ...`  command in Dockerfile. It will stay at starting jupyter ` [I 23:13:05.897 NotebookApp] http://b61192ea15be:12345/?token=... ` when building image.

hkvision: oh really? Then put into start-notebook.sh and remove the stuff that is not used in that script?

Le-Zheng: ok

hkvision (approved):

PR #3942, opened 2022-01-25
[PPML] Move the Encryption Section Out of The KMS Client Container

Uxito-Ada:
Since it is not necessary for the data-file-encryption process to run in the container, we turn to allow users encrypt files on their local host. README.md is modified accordingly.


glorysdj (approved):
LGTM

PR #3958, opened 2022-01-29
Orca Migration Code

LouieBHLu:
Add Orca Migration Code of brainMRI example


python/orca/example/learn/pytorch/brainMRI/dataset.py
```diff
@@ -0,0 +1,82 @@
...
+    image_files = [file.replace('_mask', '') for file in mask_files]
+    files_df = pd.DataFrame({"image_path": image_files,
+                    "mask_path": mask_files,
+                    "diagnosis": [diagnosis(x) for x in mask_files]})
+
+    train_df, test_df = train_test_split(files_df, stratify=files_df['diagnosis'], test_size=0.15, random_state=0)
+    train_df = train_df.reset_index(drop=True)
+    test_df = test_df.reset_index(drop=True)
+
+    return train_df, test_df
```
hkvision, line 52: Put this in brainMRI.py? Keep the unet file and dataset file unchanged as the original one?

PR #3959, opened 2022-02-04
Tf2estimator on Pyspark support tensorboard

jenniew:
Tf2estimator on Pyspark support tensorboard for fit and evaluate.

### DIFF
--- a/python/orca/src/bigdl/orca/data/file.py
+++ b/python/orca/src/bigdl/orca/data/file.py
@@ -264,7 +264,7 @@ def put_local_dir_tree_to_remote(local_dir, remote_dir):
                                    stderr=subprocess.PIPE)
         out, err = process.communicate()
         if process.returncode != 0:
-            if 'No such file or direcory' in err.decode('utf-8'):
+            if 'No such file or directory' in err.decode('utf-8'):
                 mkdir_cmd = 'hdfs dfs -mkdir {}'.format(remote_dir)
                 mkdir_process = subprocess.Popen(mkdir_cmd, shell=True)
                 mkdir_process.wait()
--- a/python/orca/src/bigdl/orca/data/file.py
+++ b/python/orca/src/bigdl/orca/data/file.py
@@ -276,10 +276,12 @@ def put_local_dir_tree_to_remote(local_dir, remote_dir):
         path_parts = remote_dir.split("://")[1].split('/')
         bucket = path_parts.pop(0)
         prefix = "/".join(path_parts)
-        local_files = [os.path.join(x[0], x[2]) for x in os.walk(local_dir)]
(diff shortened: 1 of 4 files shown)


python/orca/src/bigdl/orca/data/file.py
```diff
@@ -251,3 +251,37 @@ def put_local_dir_to_remote(local_dir, remote_dir):
...
+        import pyarrow as pa
+        host_port = remote_dir.split("://")[1].split("/")[0].split(":")
+        classpath = subprocess.Popen(["hadoop", "classpath", "--glob"],
+                                     stdout=subprocess.PIPE).communicate()[0]
+        os.environ["CLASSPATH"] = classpath.decode("utf-8")
+        fs = pa.hdfs.connect(host=host_port[0], port=int(host_port[1]))
+        if not fs.exists(remote_dir):
+            fs.mkdir(remote_dir)
+        cmd = 'hdfs dfs -put -f {}/* {}/'.format(local_dir, remote_dir)
+        process = subprocess.Popen(cmd, shell=True)
```
yangw1234, line 267: Is it possible to just use pyarrow or hadoop command instead of interleaving them together?

python/orca/src/bigdl/orca/learn/tf2/spark_runner.py
```diff
@@ -294,14 +296,16 @@ def distributed_train_func(self, data_creator, config, epochs=1, verbose=1,
...
-                shutil.rmtree(os.path.dirname(replaced_checkpoint_path))
-
+        if callbacks:
+            if checkpoint:
+                try:
+                    if self.rank == 0:
+                        put_local_dir_to_remote(os.path.dirname(replaced_checkpoint_path),
+                                                original_checkpoint_dir)
+                finally:
+                    shutil.rmtree(os.path.dirname(replaced_checkpoint_path))
```
yangw1234, line 306: If `put_local_dir_to_remote` failed, maybe we should not delete `replaced_checkpoint_path` otherwise user's training result will be lost. 

How about printing a warning stating that there is an error and the checkpoint is located at xxx and then users will have a chance to get them manually.

python/orca/src/bigdl/orca/data/file.py
jenniew: using pyarrow to copy hdfs tree is a little complex. will change to use command all the ways.

python/orca/src/bigdl/orca/learn/tf2/spark_runner.py
jenniew: ok

jenniew:
Jenkins pass: http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-Orca-PR-Validation/400/

PR #3983, opened 2022-02-09
Nano Keras Multiprocessing Training

yangw1234:
TODO:
- [x] add distributed training logic into a mixedin class
- [x] find a way change the state of current model instead of returning a new one (Option 1. use composition instead of inheritance; Option 2 using checkpoint to communicate states back, but not sure how)
- [ ]  unit tests (distributed_train, serialize and deserialize different kind of datasets)


yangw1234:
http://10.112.231.51:18888/view/BigDL-2.0-PR/job/ZOO-PR-NanoTests/246/

unit tests, backend without ray will be added in a follow up pr.

PR #3996, opened 2022-02-11
[Orca]Set raycontext to none when stop it

sgwhat:
- Solved: Sometimes when we call stop_orca_context to stop ray_context, it may return "stopping ray_orca context" twice.


sgwhat:
Jenkins passed: http://10.112.231.51:18888/job/BigDL-Orca-PR-Validation/413/

yangw1234:
LGTM

yangw1234:
http://10.112.231.51:18888/job/BigDL-PRVN-chronos-Python-Spark-3.1-py37-ray-part1/371/console

do you think this error is related to this pr?

PR #3997, opened 2022-02-11
Nano: Add onnxruntime quantization to nano.pytorch

TheaperDeng:
Here is a use-case in pytorch example
https://github.com/TheaperDeng/git_learning/compare/mnist-quantization-onnxrt

For this toy model and mnist dataset, onnxruntime with int8 quantization is slower than onnxruntime with fp32. I don't think we need to worry about it too much since some real model with larger dataset is more realistic.

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/runtime_binding/base_inference.py
+++ b/python/nano/src/bigdl/nano/pytorch/runtime_binding/base_inference.py
@@ -93,34 +93,47 @@ def inference(self,
         input_sample_list = [input_data]
 
     if backend == "onnx":
-        self.eval()
-        assert not quantize,\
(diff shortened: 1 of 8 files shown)


TheaperDeng:
http://10.112.231.51:18888/job/BigDL-PRVN-chronos-Python-Spark-3.1-py37-ray-part2/367/
http://10.112.231.51:18888/job/ZOO-PR-NanoTests/257/

TheaperDeng:
http://10.112.231.51:18888/job/ZOO-PR-NanoTests/266/

TheaperDeng:
This PR will be merged just to keep moving. We can discuss or refactor the code later.

PR #4003, opened 2022-02-11
support keras model fit dataframe of image

dding3:


dding3:
jekins pass. http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-PR-Dllib-Validation/284/

jason-dai:
1) For image preprocessing/transformations, support PyTorch like APIs:
    ```python
    transforms = transforms.Compose([
                            Resize(...),
                            Crop(...)])
    model.fit(df, transform=transforms)
    ``` 

2) Need to add test with image preprocessing/transformations

3) Need to add Python support

PR #4011, opened 2022-02-14
[PPML] Refine kms-client

Uxito-Ada:
1. Modify workflow chart, move the key request stage into spark job according the current design.
2. Move the SGX specification into the script.
3. Re-arrange the input parameter order.


glorysdj (approved):
LGTM

PR #4013, opened 2022-02-14
Chronos: Support `validation_data=None` in `ProphetForecaster`

TheaperDeng:
One of our customers suggest us to have an option to disable validation process to save time.


TheaperDeng:
http://10.112.231.51:18888/job/BigDL-Chronos-PR-Validation/357/
http://10.112.231.51:18888/job/BigDL-PRVN-chronos-Python-Spark-3.1-py36-ray-part1/392/

PR #4014, opened 2022-02-14
[Orca]Update for Stopping Orca Context

sgwhat:
- Set ray context as None which was activated by spark runtime to avoid calling pure ray code when stopping spark orca context.(runtime = "spark")

### DIFF
--- a/python/orca/src/bigdl/orca/common.py
+++ b/python/orca/src/bigdl/orca/common.py
@@ -316,10 +316,10 @@ def init_orca_context(cluster_mode=None, runtime="spark", cores=2, memory="2g", 
                         "system_config"]:
                 if key in kwargs:
                     ray_args[key] = kwargs[key]
-            from bigdl.orca.ray import RayContext
-            ray_ctx = RayContext(runtime="spark", cores=cores, num_nodes=num_nodes,
-                                 sc=sc, **ray_args)
             if init_ray_on_spark:
+                from bigdl.orca.ray import RayContext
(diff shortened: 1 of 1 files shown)


sgwhat:
Jenkins Passed: http://10.112.231.51:18888/job/BigDL-Orca-PR-Validation/419/

yangw1234:
Not sure I understand the logic. Maybe @hkvision know better.

TheaperDeng:
Could you run Chronos PR validation as well?

python/orca/src/bigdl/orca/common.py
```diff
@@ -342,6 +342,8 @@ def stop_orca_context():
         ray_ctx = RayContext.get(initialize=False)
         if ray_ctx.initialized:
             ray_ctx.stop()
+        else:
+            RayContext._active_ray_context = None
```
hkvision, line 346: If changed to this logic, do we need to set active ray context to None when ray context is initialized?

sgwhat: Do you mean set active ray context to None in ray_ctx.stop()?

hkvision: I mean remove the else, and set active to None anyway?

hkvision (approved):

sgwhat:
> Could you run Chronos PR validation as well?

Passed http://10.112.231.51:18888/job/BigDL-Chronos-PR-Validation/

PR #4016, opened 2022-02-14
Add manually save/load checkpoint for pytorch ray estimator

yushan111:
For using the latest checkpoint to continue training
```python
model_dir = "/tmp/test"  # or remote path like hdfs:///tmp/test 
latest_checkpoint_path = Estimator.latest_checkpoint(model_dir)

new_estimator.load_checkpoint(latest_checkpoint_path)
new_estimator.evaluate()
```
For manually save and load checkpoint:
```python
ckpt_file = os.path.join(temp_dir, "manual.ckpt")
estimator.save_checkpoint(ckpt_file)

new_estimator.load_checkpoint(ckpt_file)
new_estimator.predict()
```

### DIFF
--- a/python/orca/src/bigdl/orca/learn/pytorch/pytorch_ray_estimator.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/pytorch_ray_estimator.py
@@ -439,8 +439,8 @@ class PyTorchRayEstimator(OrcaRayEstimator):
         """
         Loads the Estimator state (including model and optimizer) from the provided model_path.
 
-        :param model_path: (str) Path to the existing model. Both local and remote path are supported.
-               e.g. "/tmp/estimator.ckpt" or "hdfs:///tmp/estimator.ckpt"
+        :param model_path: (str) Path to the existing model. Both local and remote path are
+               supported. e.g. "/tmp/estimator.ckpt" or "hdfs:///tmp/estimator.ckpt"
         :return: None
         """
         from bigdl.dllib.utils.file_utils import is_local_path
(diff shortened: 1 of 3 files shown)


yushan111:
http://10.112.231.51:18888/job/BigDL-Orca-PR-Validation/424/
http://10.112.231.51:18888/job/ZOO-PR-BigDL-Python-ExampleTests-py37-ray-horovod-Spark3/424/

yushan111:
test on yarn with remote hdfs path also passed.

python/orca/src/bigdl/orca/learn/pytorch/pytorch_ray_estimator.py
```diff
@@ -416,6 +416,28 @@ def load(self, model_path):
         state_dict = torch.load(model_path)
         self.load_state_dict(state_dict)
 
+    def save_checkpoint(self, model_path):
+        from bigdl.dllib.utils.file_utils import is_local_path
+        if is_local_path(model_path):
+            self.save(model_path)
+        else:
+            results = [
+                worker.save_checkpoint.remote(model_path)
```
yangw1234, line 425: why saving `checkpoint` in every worker?

Does it work if multiple workers in the same node saving to the same location?

yushan111: `worker.save_checkpoint` is responsible for correctly handling the behavior in distributed training, which saves checkpoint only on rank 0. (In that way, `ModelCheckpoint` is transparent for both local and ddp).
In that way, it does work for multiple workers on the same node.

yushan111:
http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-Orca-PR-Validation/426/

yangw1234 (approved):
LGTM

Issue #4002, opened 2022-02-11
Nano quantization report AssertionError on the second time 


TheaperDeng:
Here is a minimal example:
```python
model = ResNet18(10, pretrained=False, include_top=False, freeze=True)
loss = nn.CrossEntropyLoss()
optimizer = torch.optim.Adam(model.parameters(), lr=0.01)
train_loader = create_data_loader(data_dir, batch_size, num_workers, data_transform)
trainer = Trainer(max_epochs=1)
pl_model = Trainer.compile(model, loss, optimizer)
qmodel = trainer.quantize(pl_model, train_loader, train_loader,
                              metric=torchmetrics.F1(10), framework='pytorch_fx',
                              approach='static',
                              tuning_strategy='basic',
                              accuracy_criterion={'relative': 0.99,
                                                  'higher_is_better': True})
qmodel = trainer.quantize(pl_model, train_loader, train_loader,
                              metric=torchmetrics.F1(10), framework='pytorch_fx',
                              approach='static',
                              tuning_strategy='basic',
                              accuracy_criterion={'relative': 0.99,
                                                  'higher_is_better': True})
```
it will report this on the second time.
```bash
Traceback (most recent call last):
  File "test.py", line 68, in <module>
    'higher_is_better': True})
  File "/home/junweid/BigDL/python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py", line 239, in quantize
    metric)
  File "/home/junweid/BigDL/python/nano/src/bigdl/nano/quantization/neural_compressor/quantization.py", line 102, in post_training_quantize
    name="%s_%s" % (framework_metric.__name__, type(metric).__name__)
  File "/home/junweid/miniconda3/envs/chronos/lib/python3.7/site-packages/neural_compressor/experimental/quantization.py", line 287, in metric
    metrics.register(user_metric.name, user_metric.metric_cls)
  File "/home/junweid/miniconda3/envs/chronos/lib/python3.7/site-packages/neural_compressor/experimental/metric/metric.py", line 116, in register
    assert name not in self.metrics.keys(), 'registered metric name already exists.'
AssertionError: registered metric name already exists.
```

I am pretty sure this should be a INC issue, but we may have a look first. @zhentaocc

zhentaocc:
It's not INC's problem. I think each call of `trainer.quantize(..)` will register the metric into INC as `F1`, so this is the reason.   
[...]

PR #4030, opened 2022-02-15
Resolve inc metric duplicate registration

zhentaocc:
resolves https://github.com/intel-analytics/BigDL/issues/4002
In previous implementation, each call of quantization, a metric will be registered.  
This causes error when we call quantization with a same metric in one script twice.

### previous implementation :
Each metric is registered at runtime, so there can be name conflict.
```python
  self.metric = Metric(
      MyMetric,
      name="%s_%s" % (framework_metric.__name__, type(metric).__name__)
  )
```
### After this fix:
Customized metrics firstly are registered once for all statically.
```python
METRICS('pytorch').register('PytorchINCMetric', PytorchINCMetric)
METRICS('tensorflow').register('TensorflowINCMetric', TensorflowINCMetric)
```
Then at runtime, we modify the real metric inside `INCMetric` by:
```python
PytorchINCMetric.set_metric(metric)
```
And fetch the registered metric by class name:
```python
self.cfg.evaluation.accuracy.metric = {PytorchINCMetric.__name__: None}
```

### Modifications

- [x]  Add default criterion to INC, allowing 0.99 acc drop by default
- [x]  Fix metric registered in second time
- [x]  Add test with metric assigned second time

### DIFF
--- a/python/nano/src/bigdl/nano/quantization/neural_compressor/metric/base_metric.py
+++ b/python/nano/src/bigdl/nano/quantization/neural_compressor/metric/base_metric.py
@@ -17,6 +17,8 @@ from abc import ABC, abstractmethod
 
 
 class INCMetric(ABC):
+    METRIC_ID = 0
+
     def __init__(self):
         self.pred_list = []
         self.label_list = []
@@ -55,3 +57,8 @@ class INCMetric(ABC):
     @abstractmethod
     def to_scalar(self, tensor):
         pass
+
+    @staticmethod
+    def get_next_metric_id():
+        INCMetric.METRIC_ID += 1
+        return INCMetric.METRIC_ID
--- a/python/nano/src/bigdl/nano/quantization/neural_compressor/quantization.py
+++ b/python/nano/src/bigdl/nano/quantization/neural_compressor/quantization.py
@@ -96,9 +96,11 @@ class QuantizationINC(Quantization):
                         to the new local class to avoid that.
                         """
                         self.metric = metric
+
                 self.metric = Metric(
                     MyMetric,
-                    name=f"nano_metric_{framework_metric.get_next_metric_id()}"
+                    name=f"{framework}_{type(metric).__name__}_"
+                         f"{framework_metric.get_next_metric_id()}"
                 )
 
         quantized = self()
--- a/python/nano/src/bigdl/nano/quantization/neural_compressor/quantization.py
+++ b/python/nano/src/bigdl/nano/quantization/neural_compressor/quantization.py
@@ -96,10 +96,9 @@ class QuantizationINC(Quantization):
                         to the new local class to avoid that.
                         """
                         self.metric = metric
-
                 self.metric = Metric(
                     MyMetric,
-                    name="%s_%s" % (framework_metric.__name__, type(metric).__name__)
+                    name=f"nano_metric_{framework_metric.get_next_metric_id()}"
                 )
 
         quantized = self()
--- a/python/nano/test/pytorch/tests/test_trainer.py
+++ b/python/nano/test/pytorch/tests/test_trainer.py
@@ -82,11 +82,12 @@ class TestTrainer(TestCase):
         train_loader_iter = iter(self.train_loader)
         trainer = Trainer(max_epochs=1)
         pl_model = Trainer.compile(self.model, self.loss, self.optimizer)
+        x = next(train_loader_iter)[0]
 
         # Case 1: Default
         qmodel = trainer.quantize(pl_model, self.train_loader)
         assert qmodel
-        out = qmodel(next(train_loader_iter)[0])
+        out = qmodel(x)
         assert out.shape == torch.Size([256, 10])
 
         # Case 2: Override by arguments
@@ -98,13 +99,13 @@ class TestTrainer(TestCase):
                                                       'higher_is_better': True})
 
         assert qmodel
-        out = qmodel(next(train_loader_iter)[0])
+        out = qmodel(x)
         assert out.shape == torch.Size([256, 10])
 
         # Case 3: Dynamic quantization
         qmodel = trainer.quantize(pl_model, approach='dynamic')
         assert qmodel
-        out = qmodel(next(train_loader_iter)[0])
+        out = qmodel(x)
         assert out.shape == torch.Size([256, 10])
 
         # Case 4: Invalid approach
@@ -112,6 +113,13 @@ class TestTrainer(TestCase):
         with pytest.raises(ValueError, match="Approach should be 'static' or 'dynamic', "
                                              "{} is invalid.".format(invalid_approach)):
             trainer.quantize(pl_model, approach=invalid_approach)
+
+        # Case 5: Test if registered metric can be fetched successfully
+        qmodel = trainer.quantize(pl_model, self.train_loader, self.train_loader,
+                                  metric=torchmetrics.F1(10))
+        assert qmodel
+        out = qmodel(x)
+        assert out.shape == torch.Size([256, 10])
 
     def test_trainer_quantize_inc_ptq_customized(self):
         # Test if a Lightning Module not compiled by nano works


zhentaocc:
@TheaperDeng Please check if this resolves your issue. Or any other suggestions on the fix.

zhentaocc:
@TheaperDeng I have a new correct version. Take a look.

TheaperDeng (approved):
LGTM

zhentaocc:
Jenkins passed.  
http://10.112.231.51:18888/job/ZOO-PR-NanoTests/253/

python/nano/src/bigdl/nano/quantization/neural_compressor/metric/__init__.py
```diff
@@ -13,10 +13,16 @@
...
+from neural_compressor.metric import METRICS
 
 from .metrics import *
 
-METRICS = {
-    'pytorch': PytorchINCMetric,
-    'tensorflow': TensorflowINCMetric
+METRICS('pytorch').register('PytorchINCMetric', PytorchINCMetric)
+METRICS('tensorflow').register('TensorflowINCMetric', TensorflowINCMetric)
+
```
yangw1234, line 22: Where `PytorchINCMetric` and `TensorflowINCMetric` are  defined?

python/nano/src/bigdl/nano/quantization/neural_compressor/quantization.py
```diff
@@ -81,26 +83,9 @@ def post_training_quantize(self, model, calib_dataloader=None, val_dataloader=No
...
-                        to the new local class to avoid that.
-                        """
-                        self.metric = metric
-
-                self.metric = Metric(
-                    MyMetric,
-                    name="%s_%s" % (framework_metric.__name__, type(metric).__name__)
-                )
+                framework_inc_metric = INCMETRICS[self.cfg.model.framework]
+                framework_inc_metric.set_metric(metric)
```
yangw1234, line 87: Does it make sense to modify a global state in a object? I think it is hard reason about the behavior and maintain the code. E.g. If I have a second `QuantizationINC` object with different metric, will it mess up the state of the first one?

zhentaocc: You can't run them in parallel. Otherwise, this mess would be possible.

zhentaocc: If you don't use multiple threads to do twice quantization in your assumption, that'll be fine. And seems doing that is not a good practice for post-training quantization.

python/nano/src/bigdl/nano/quantization/neural_compressor/metric/__init__.py
zhentaocc: https://github.com/intel-analytics/BigDL/blob/branch-2.0/python/nano/src/bigdl/nano/quantization/neural_compressor/metric/metrics.py#L19-L40

zhentaocc:
Another option is to change the arguments into `quantize`.
Before:
```python
quantize(metric=torchmetrics.F1(10))
```
After:
```python
quantize(metric={torchmetrics.metrics.F1: {num_classes: 10}})
```
which is much more complicated for users to use. @yangw1234

python/nano/src/bigdl/nano/quantization/neural_compressor/quantization.py
yangw1234: I do not think you need multi-threading to mess it up. Does the following code work?
```python
quantA = QuantizationINC(metric-a)
quantB = QuantizationINC(metric-b)

quantA.some_method
quantB.some_method
```

yangw1234: If there is a serialization problem, I think you should try to solve the serialization problem directly.

python/nano/src/bigdl/nano/quantization/neural_compressor/metric/base_metric.py
```diff
@@ -24,13 +24,13 @@ def __init__(self):
...
+            return getattr(INCMetric, '_metric')
         except AttributeError:
             raise RuntimeError("Attribute 'metric' must be set.")
 
-    @metric.setter
-    def metric(self, framework_metric):
-        setattr(self, '_metric', framework_metric)
+    @staticmethod
+    def set_metric(framework_metric):
+        INCMetric._metric = framework_metric
```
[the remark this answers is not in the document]
jason-dai: Why can you dump a class variable but not a member variable? Is it because that a `TensorflowINCMetric` variable cannot be dumped?

python/nano/src/bigdl/nano/quantization/neural_compressor/quantization.py
zhentaocc: > I do not think you need multi-threading to mess it up. Does the following code work?
> 
> ```python
> quantA = QuantizationINC(metric-a)
> quantB = QuantizationINC(metric-b)
> 
> quantA.some_method
> quantB.some_method
> ```
Your use case is not actually correct. It should be
```python
trainer.quantize(model, trainloader, valloader, metric_a)
trainer.quantize(model, trainloader, valloader, metric_b)
```
or
```python
trainer.quantize(model, trainloader, valloader, metric_a)
trainer.quantize(model, trainloader, valloader, metric_b)
```python
quantA = QuantizationINC(**kwargs_a)
quantB = QuantizationINC(**kwargs_b)

quantized_1 = quantA.post_training_quantize(model, calib_dataloader, val_dataloader,
                                             metric_a)
quantized_2 = quantB.post_training_quantize(model, calib_dataloader, val_dataloader,
                                             metric_b)
quantized_3 = quantA.post_training_quantize(model, calib_dataloader, val_dataloader,
                                             metric_b)
quantized_4 = quantB.post_training_quantize(model, calib_dataloader, val_dataloader,
                                             metric_a)
```
With different kwargs and metric, the 4 quantization processes are all different and the metric wouldn't be messed up.

zhentaocc: > If there is a serialization problem, I think you should try to solve the serialization problem directly.

If we want to resolve this directly, then metric pass to INC must be `{cls: kwargs}` and we need to assure that kwargs must be able to be serialized. So the solution is to pass a dict `{torchmetric_class: kwargs}` instead of instance of `torchmetric_class(kwargs)` as stated in comments below. Having discussed with @TheaperDeng , we agreed on passing a torchmetric instance which is more user-friendly.

zhentaocc: > If there is a serialization problem, I think you should try to solve the serialization problem directly.

Or we need to check in some changes into INC directly... But this is not possible for the moment.

jason-dai:
> resolves #4002 In previous implementation, each call of quantization, a metric will be registered. This causes error when we call quantization with a same metric in one script twice.
> 
> ### previous implementation :
> Each metric is registered at runtime, so there can be name conflict.
> 
> ```python
>   self.metric = Metric(
>       MyMetric,
>       name="%s_%s" % (framework_metric.__name__, type(metric).__name__)
>   )
> ```
> 
> ### After this fix:
> Customized metrics firstly are registered once for all statically.
> 
> ```python
> METRICS('pytorch').register('PytorchINCMetric', PytorchINCMetric)
> METRICS('tensorflow').register('TensorflowINCMetric', TensorflowINCMetric)
> ```
> 
> Then at runtime, we modify the real metric inside `INCMetric` by:
> 
> ```python
> PytorchINCMetric.set_metric(metric)
> ```
> 
> And fetch the registered metric by class name:
> 
> ```python
> self.cfg.evaluation.accuracy.metric = {PytorchINCMetric.__name__: None}
> ```
> 
> ### Modifications
> * [x]   Add default criterion to INC, allowing 0.99 acc drop by default
> * [x]   Fix metric registered in second time
> * [x]   Add test with metric assigned second time

I think this is too complex for solving a simple issue; to avoid conflict, you can simple append a unique id to the metrics name when registering it, something like:

```python
    self.metric = Metric(
       MyMetric,
       name="%s_%s_%d" % (framework_metric.__name__, type(metric).__name__, Metric.get_next_id())
   )
 ```

**Comment by zhentaocc** [comment]
> 
> I think this is too complex for solving a simple issue; to avoid conflict, you can simply append a unique id to the metrics name when registering it, something like:
> 
> ```python
>     self.metric = Metric(
>        MyMetric,
>        name="%s_%s_%d" % (framework_metric.__name__, type(metric).__name__, Metric.get_next_id())
>    )
> ```

Yes, this is much simpler.  
I was trying to refactor the code. I will firstly fix the issue, and open new PR to refactor if necessary.

**Comment by zhentaocc** [comment]
@jason-dai @yangw1234 @TheaperDeng How about current version?

File: python/nano/src/bigdl/nano/quantization/neural_compressor/quantization.py [hunk 4]
```diff
@@ -96,10 +96,9 @@ def __init__(self):
                         to the new local class to avoid that.
                         """
                         self.metric = metric
-
                 self.metric = Metric(
                     MyMetric,
-                    name="%s_%s" % (framework_metric.__name__, type(metric).__name__)
+                    name=f"nano_metric_{framework_metric.get_next_metric_id()}"
```
jason-dai, line 101: give it a more meaningful name

TheaperDeng: yeah, maybe a metric name and an unique id.

zhentaocc: OK, let me roll back the name.

python/nano/src/bigdl/nano/quantization/neural_compressor/metric/base_metric.py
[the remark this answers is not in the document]
zhentaocc: It's not about how the metric looks like, it's more about how the metric is created by INC. INC accepts a class `INCMetric` and a set of arguments `kwargs` to create a metric instance by INCMetric(**kwargs). And when we finished quantization in tensorflow, they try to dump the metric by a dict like `{INCMetric.name: kwargs}`. So it seems that we can't pass any argument that is not able to be dumped.  
### The problematic version
```python
self.metric=Metric(INCMetric, metric=metric)
```
Here, `kwargs={'metric': metric}` and `metric` can't be dumped.
### Resolve by class variable
```python
INCMetric.metric=metric
self.metric=Metric(INCMetric)
```
Here, `kwargs={}` , no issue.
### Resolve by local class inside function(current version)
```python
class MyMetric(INCMetric):
    def __init__(self):
        self.metric = metric
self.metric=Metric(MyMetric)
```
Here, `kwargs={}` , no issue.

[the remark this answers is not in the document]
zhentaocc: This issue only appears on tensorflow side. I think maybe pytorch and tensorflow have different postprocessings.

jason-dai:
LGTM

zhentaocc:
Jenkins passed http://10.112.231.51:18888/job/ZOO-PR-NanoTests/263/

yangw1234 (approved):

jason-dai:
Also need to merge it to the main branch?

TheaperDeng:
> Also need to merge it to the main branch?

I think main branch is created after this PR is merged. this commit is in main branch's tree.

PR #4036, opened 2022-02-16
[PPML] Remove maven and old log4j jar

Uxito-Ada:
1. Remove maven installation from stage 4 ppml.
2. remove slf4j-log4j12-1.7.30.jar.


glorysdj (approved):
LGTM

PR #4039, opened 2022-02-16
[PPML] Add log4j2 conf file for spark

Uxito-Ada:
Add a log4j2.xml file to $SPARK_HOME to enable log print on spark driver.


glorysdj (approved):
LGTM

PR #4055, opened 2022-02-17
update Scala doc

qiuxin2012:
1. update Scala doc
2. update pom.xml's profile `all-in-one`


qiuxin2012:
@dding3 do you know why profile `all-in-one` is removed ?

PR #4057, opened 2022-02-17
Test Nano Tensorflow Fit Function (Multiple Processing)

az2181036:


python/nano/test/tf/keras/test_fit.py
```diff
@@ -0,0 +1,93 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+import os
+
+os.system("pip install ray[default]")
```
yangw1234, line 17: you should not add install in the script

python/nano/test/tf/keras/test_fit.py
```diff
@@ -0,0 +1,93 @@
...
+os.system("pip install -U intel_tensorflow")
+
+import pathlib
```
yangw1234, line 20: You generally should not add logic in the global scope. You should wrap them in the test function.

python/nano/test/tf/keras/test_fit.py
```diff
@@ -0,0 +1,93 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+import os
+
+os.system("pip install ray[default]")
+os.system("pip install -U intel_tensorflow")
```
yangw1234, line 18: You can upgrade tensorflow version here https://github.com/intel-analytics/BigDL/blob/branch-2.0/python/nano/src/setup.py#L76-L78, note that these three versions should be the same.

PR #4065, opened 2022-02-21
Tf2estimator on Pyspark backend model checkpoint refactor

jenniew:
Refactor model checkpoint to support same frequency of update of original checkpoint callback.


jenniew:
Jenkins pass: http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-Orca-PR-Validation/449/

PR #4072, opened 2022-02-21
xgboost example

yizerozhuang:
Add xgb example scripts


yizerozhuang:
http://10.112.231.51:18888/job/BigDL-PR-Friesian-Validation/262/

yizerozhuang:
jenkins all passed

hkvision:
Not backported to main and the time is too long

PR #4074, opened 2022-02-22
update occlum sparksql unit test

qiuxin2012:


qiuxin2012:
We need more builded files, like assembly folder. Or Suite like `org.apache.spark.sql.execution.joins.BroadcastJoinSuite` will fail.

qiuxin2012:
> We need more builded files, like assembly folder. Or Suite like `org.apache.spark.sql.execution.joins.BroadcastJoinSuite` will fail.

Just need a new `SPARK_SCALA_VERSION` env.

PR #4088, opened 2022-02-23
PPML FGBoost Multi-party support fix and log refactor

Litchilitchy:
Fix includes following

* In VFL, users hold different features, where the feature IDs would not be in original order. Thus we need to use name to identify them in validation
* Add multi-party sync for different phases, each stage holds a version for sync in next stage
* Log message refactor


Litchilitchy:
test passed.

PR #4089, opened 2022-02-23
Chronos: full support for automodel and tsppl's customized metrics

TheaperDeng:


TheaperDeng:
http://10.112.231.51:18888/job/BigDL-Chronos-PR-Validation/

PR #4092, opened 2022-02-23
Change branch-2.0 -> main in github action style check

yushan111:


yangw1234:
Maybe both main and branch-2.0 needs to be tested.

yushan111:
> Maybe both main and branch-2.0 needs to be tested.

I just changed the style check on main branch from branch-2.0 to main. On branch-2.0, the style check is still for branch-2.0. 

Do you mean that we should both check main and branch-2.0 on the main branch?

PR #4096, opened 2022-02-23
multiprocessing backend using subprocessing.Popen

az2181036:
It works. Thanks for wangyang's idea.  The time cost is almost the same as the ray backend.

### DIFF
--- a/python/nano/src/bigdl/nano/common/multiprocessing/multiprocs_backend.py
+++ b/python/nano/src/bigdl/nano/common/multiprocessing/multiprocs_backend.py
@@ -1,0 +1,103 @@
+#
+# Copyright 2016 The BigDL Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import copy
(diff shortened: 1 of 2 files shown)


python/nano/src/bigdl/nano/common/multiprocessing/multiprocs_backend.py
```diff
@@ -0,0 +1,79 @@
...
+
+        return self.run_subprocess(target, args=args, nprocs=nprocs, envs=envs)
+
+    def run_subprocess(self, target, args=..., nprocs=1, envs=None) -> None:
+        import pickle
+        import subprocess
+        import sys
+
+        temp_dir = args[0]
+        with open(os.path.join(temp_dir, "train_ds_def.pkl"), 'wb') as f:
```
yangw1234, line 51: I think you may dump all args as a list in a single file

python/nano/src/bigdl/nano/common/multiprocessing/multiprocs_backend.py
```diff
@@ -0,0 +1,79 @@
...
+                raise ValueError("envs must be a dict or a list of dict")
+
+        return self.run_subprocess(target, args=args, nprocs=nprocs, envs=envs)
+
+    def run_subprocess(self, target, args=..., nprocs=1, envs=None) -> None:
+        import pickle
+        import subprocess
+        import sys
+
+        temp_dir = args[0]
```
yangw1234, line 50: Here you made a lot of assumptions about the semantics of args, which I do not think is necessary.

az2181036: I will try to dump a list into a single file, this way may avoid using those assumptions.

python/nano/src/bigdl/nano/common/multiprocessing/multiprocs_backend.py
az2181036: done

PR #4105, opened 2022-02-24
fix convert scalar to float64 error

LouieBHLu:
if the input data is of any type of pyspark datatype, it will be convert to numpy.float64, which is conflict with that of torch.float32 if the input data is label.


hkvision (approved):

hkvision:
http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-Orca-PR-Validation/461/
http://10.112.231.51:18888/job/ZOO-PR-BigDL-Python-ExampleTests-py36-ray/526/
http://10.112.231.51:18888/job/ZOO-PR-BigDL-Python-Spark-2.4-py36-ray/496/

hkvision:
Comment on this PR: numpy by default uses float64(double) and int64(long); for float and int, we keep the original datatype.

PR #4127, opened 2022-02-28
Nano: Onnxruntime/pytorch default quantize and accumulated enhancement

TheaperDeng:
This PR is to support `trainer.test` for our examples

1. support(and test `trainer.test` works for quantized model)
2. change `raw_return` to `return_pl` for better understanding
3. Now, quantized model is the default model right after the calling of `trainer.quantize`, here is an example:
```python
# ...
# training process

# defaultly using quantized model
model = trainer.quantize(model, dataloader, ...)
model.eval()  
with torch.no_grad():
    model(x)  # quantized model

# explicitly call quantize=False for fp32 model
model.eval(quantize=False)  
with torch.no_grad():
    model(x)  # fp32 model

# defaultly using fp32 model
trainer.fit(model, data)
model.eval()  
with torch.no_grad():
    model(x)  # fp32 model

```

### DIFF
--- a/python/chronos/src/bigdl/chronos/forecaster/base_forecaster.py
+++ b/python/chronos/src/bigdl/chronos/forecaster/base_forecaster.py
@@ -567,4 +567,4 @@ class BasePytorchForecaster(Forecaster):
                                               accuracy_criterion=accuracy_criterion,
                                               timeout=timeout,
                                               max_trials=max_trials,
-                                              raw_return=False)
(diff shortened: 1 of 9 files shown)


TheaperDeng:
http://10.112.231.51:18888/job/BigDL-PRVN-chronos-Python-Spark-3.1-py37-ray-part2/421/
http://10.112.231.51:18888/job/ZOO-PR-NanoTests/288/

PR #4128, opened 2022-02-28
Update akka.actor.version

Uxito-Ada:
Update akka.actor.version from 2.5.26 to 2.5.31, in order to avoid http-frontend exception.


glorysdj (approved):
LGTM

PR #4131, opened 2022-02-28
fix nano tf init threading config

yangw1234:
It should use the value defined when at `bigdl-nano-init`


yangw1234:
http://10.112.231.51:18888/job/ZOO-PR-NanoTests/281/

PR #4134, opened 2022-02-28
[PPML] Fix yaml parse error in realtime/occlum/k8s

Uxito-Ada:
Line 86 used to not align to the above parameters.


qiyuangong (approved):
LGTM

PR #4138, opened 2022-02-28
Nano: use mobilenet v2 as keras ut for quantization

zhentaocc:
VGG16 is too large as a ut for quantization, which takes very long time.  
Use MobileNetv2 instead.


zhentaocc:
jenkins passed http://10.112.231.51:18888/job/ZOO-PR-NanoTests/283/

yangw1234:
LGTM

PR #4139, opened 2022-02-28
Refactor RayXShards to easily get partition and shard ObjectRefs

yushan111:
Before: 

- The `LocalStore` in `RayXShards` holds all the partition data, but not Ray `ObjectRef`s.

After:

- `LocalStore` holds the `ObjectRef`s of partitions and shards.

- Add `get_partition_ref` which returns a list of `shard_ref`s or a `part_ref`.

### DIFF
--- a/python/orca/src/bigdl/orca/automl/search/ray_tune/ray_tune_search_engine.py
+++ b/python/orca/src/bigdl/orca/automl/search/ray_tune/ray_tune_search_engine.py
@@ -407,17 +407,9 @@ def trial_dirname_creator(trial):
 def get_data_from_part_refs(part_refs):
     from bigdl.orca.data.utils import ray_partitions_get_data_label
 
-    partitions = ray.get(part_refs)
+    partitions = [ray.get(part_ref) for part_ref in part_refs]
 
-    # convert list of dicts to one dict
-    result = {}
-    for part in partitions:
-        result.update(part)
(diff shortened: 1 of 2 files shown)


yushan111:
http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-Orca-PR-Validation/469/

python/orca/src/bigdl/orca/automl/search/ray_tune/ray_tune_search_engine.py
```diff
@@ -407,17 +407,9 @@ def trial_dirname_creator(trial):
 def get_data_from_part_refs(part_refs):
     from bigdl.orca.data.utils import ray_partitions_get_data_label
 
-    partitions = ray.get(part_refs)
+    partitions = [ray.get(part_ref) for part_ref in part_refs]
```
yangw1234, line 410: These two way seems to be equivalent.

yushan111: `part_refs` is a list of `partition_ref`s, where `partition_ref` is a list of `ObjectRef`s for shards. We cannot directly get the data from ray when the data is [[`object_ref`], [`object_ref`, `object_ref`]], therefore only the changed way should work.

PR #4141, opened 2022-02-28
add python keras training with dataframe

dding3:
1. add python keras model fit/evaluate/predict with dataframe
2. add python keras model fit/evaluate/predict with dataframe of image
3. add keras image classification example
4. add transforms.compose


dding3:
jekins pass. http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-PR-Dllib-Validation/336/

PR #4156, opened 2022-03-02
Fix pypi release for source distribution

hkvision:
Manually tested.


hkvision:
http://10.112.231.51:18888/view/BigDL/job/BigDL2.0-NB-Deploy-PYTHON-PyPI-all/44/console

PR #4159, opened 2022-03-03
Nano: add dataloader check for pytorch quantization

### LINKED ISSUE #4163: Nano: Nano: add dataloader check for pytorch quantization to comply with INC calibration format
### Problem
I wrote a unit test to reproduce the error:
```python
from torch import nn
from torch.utils.data import TensorDataset, DataLoader

class MultipleInputs(nn.Module):
    def __init__(self):
        super().__init__()
        self.linear1 = nn.Linear(3, 1)
        self.linear2 = nn.Linear(4, 1)

    def forward(self, x1, x2):
        y1 = self.linear1(x1)
        y2 = self.linear2(x2)
        return y1.squeeze()

model = MultipleInputs()
x1 = torch.ones((10, 3))
x2 = torch.ones((10, 4))
y1 = torch.ones((10))
model(x1, x2)
dataset = TensorDataset(x1, x2, y1)
dataloader = DataLoader(dataset, batch_size=5)

trainer = Trainer()
trainer.quantize(model, dataloader)
```
With more than two tensors in a batch, e.g. `x1, x2, y1`,  INC requires to wrap them as `(x1, x2), y1`.  

### solution
A dataloader should be restricted to return no more than two items. It can be one of them:
- `(tensor_1, ... ,tensor_n), tensor_n1`
- `(tensor_1, ... ,tensor_n), (tensor_n1, ...,  tensor_n2)`
- `tensor_1, (tensor_2, ..., tensor_n)`
- `tensor_1, tensor_2`
- `tensor_1, tensor_2, ...` (no labels attached)

### What to do

- [x] Run sanity check on dataloader and raise error to instruct user to modify their dataloader
- [ ] (Optional) Provide a function to re-collate data from dataloader with index
- [x] Make sure everything remains fine, `trainer.fit(..)` `trainer.quantize(..)`

zhentaocc:
resolves #4163 

- [x] check dataloader before pytorch quantization
- [x] check dataloader by run `model(sample)` 
- [x] unit test for both legal/illegal dataloader


zhentaocc:
@jason-dai @TheaperDeng @yangw1234 any comments?

python/nano/src/bigdl/nano/quantization/neural_compressor/pytorch/utils/dataloader.py
```diff
@@ -0,0 +1,48 @@
...
+    sample = next(iter(loader))
+    try:
+        x, y = sample
+        _check_data_type(x)
+        _check_data_type(y)
+    except ValueError:
+        raise ValueError(
+            "Dataloader for quantization in INC should yield data in format below:\n"
+            "- torch.Tensor, torch.Tensor\n"
+            "- Tuple(torch.Tensor), Tuple(torch.Tensor)\n"
```
TheaperDeng, line 39: `tuple` instead of `Tuple`?

python/nano/src/bigdl/nano/quantization/neural_compressor/pytorch/utils/dataloader.py
```diff
@@ -0,0 +1,48 @@
...
+        for x in data:
+            assert isinstance(x, torch.Tensor), ValueError
+
+
+def _check_loader(loader):
+    if loader is None:
+        return
+    sample = next(iter(loader))
+    try:
+        x, y = sample
```
TheaperDeng, line 32: maybe add a comment here, `# if len(sample)>2, a ValueError will be raised`

python/nano/src/bigdl/nano/quantization/neural_compressor/pytorch/utils/dataloader.py
```diff
@@ -0,0 +1,48 @@
...
+#
+import torch
+
+
+def _check_data_type(data):
+    if isinstance(data, torch.Tensor):
+        return
+    else:
+        for x in data:
+            assert isinstance(x, torch.Tensor), ValueError
```
TheaperDeng, line 24: Oh btw, I am not quite sure about this. Pytorch dataloader should have a default collate func which will try to transform the value to Tensor. Maybe we can loose the checking here?
https://github.com/pytorch/pytorch/blob/master/torch/utils/data/_utils/collate.py#L18-L76

python/nano/src/bigdl/nano/quantization/neural_compressor/pytorch/utils/dataloader.py
```diff
@@ -0,0 +1,62 @@
...
+        else:
+            # If sample is not tuple of length 2, then it should be (x1, x2, x3, ...).
+            # Each one must be of torch.Tensor
+            _check_data_type(sample)
+            # check if datalader yields data complied with what model requires
+            # TypeError will be raised if it fails
+            model(*sample)
+    except (ValueError, TypeError):
+        raise ValueError(
+            "Dataloader for quantization in INC should yield data in format below:\n"
```
jason-dai, line 50: `Dataloader for quantization should yield data in format below:`

python/nano/src/bigdl/nano/quantization/neural_compressor/pytorch/utils/dataloader.py
zhentaocc: This check is after dataloader collate_fn.

python/nano/src/bigdl/nano/quantization/neural_compressor/pytorch/utils/dataloader.py
zhentaocc: accommodate checking on `len(sample) !=2` in new version.

python/nano/src/bigdl/nano/quantization/neural_compressor/pytorch/utils/dataloader.py
zhentaocc: resolved.

zhentaocc:
Jenkins passed http://10.112.231.51:18888/job/ZOO-PR-NanoTests/304/

yangw1234:
LGTM, except you should change the title of this pr

PR #4160, opened 2022-03-03
Add example for spark dataframe to ray dataset with xgboost-ray  

yushan111:
`pip install xgboost_ray`


yushan111:
http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-Orca-PR-Validation/481/

PR #4175, opened 2022-03-08
error log organize

dding3:
1. In our current implementation, if user input invalid inputs, dllib will crash. For better user experience, add `Log4Error.invalidInputError` api to replace some `require` or `assert` errors. If the inputs is not valid, dllib will givie error logs and exit instead of crashes.


jason-dai:
> 1. In our current implementation, if user input invalid inputs, dllib will crash. For better user experience, add `Log4Error.invalidInputError` api to replace some `require` or `assert` errors. If the inputs is not valid, dllib will givie error logs and exit instead of crashes.

Where is `Log4Error` defined?

dding3:
> > 1. In our current implementation, if user input invalid inputs, dllib will crash. For better user experience, add `Log4Error.invalidInputError` api to replace some `require` or `assert` errors. If the inputs is not valid, dllib will givie error logs and exit instead of crashes.
> 
> Where is `Log4Error` defined?

[scala/dllib/src/main/scala/com/intel/analytics/bigdl/dllib/utils/Log4Error.scala](https://github.com/intel-analytics/BigDL/pull/4175/files#diff-759c25c87dc5ab0aa8187539ea6bd69508f8c536d733ad51800d8b9a2ef2e45a)

dding3:
jekins pass. http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-PR-Dllib-Validation/374/

jason-dai:
Please show some sample error messages in the description.

jason-dai:
And split this PR into 2: one for scala, and the other for python.

dding3:
This PR is split to https://github.com/intel-analytics/BigDL/pull/4323 and https://github.com/intel-analytics/BigDL/pull/4324. Close this pr.

PR #4204, opened 2022-03-09
update bigdl-tf bigdl-math version

Le-Zheng:
https://pypi.org/project/bigdl-math/2.0.0/
https://pypi.org/project/bigdl-tf/2.0.0/


yangw1234:
LGTM

PR #4214, opened 2022-03-10
[Orca] Support Ray Dataset in PyTorch Ray Estimator Evaluate

sgwhat:
The ray dataset has already been supported in estimator.fit, so we are going to support the ray dataset in evaluate.


sgwhat:
http://10.112.231.51:18888/job/BigDL-Orca-PR-Validation/

PR #4215, opened 2022-03-10
support horovod in nano keras multiprocessing

yangw1234:


yangw1234:
http://10.112.231.51:18888/job/ZOO-PR-NanoTests/310/

PR #4076, opened 2022-02-22
Nano: Original Pytorch example fro quantization- MNIST ConvNet

zhentaocc:
An example for #3901 
This is completely from https://github.com/pytorch/examples/blob/master/mnist/main.py
A second PR will follow to do some changes to show how nano and quantization are applied on pytorch code.

Pytorch License is here https://github.com/pytorch/pytorch/blob/master/LICENSE

- [x] - Add BigDL header
- [x] -  Add Pytorch header

PR #4236, opened 2022-03-14
Nano: quantization example from pytorch original code

zhentaocc:
Modify #4076 to show how to apply nano quantization on pytorch original code.


yangw1234:
The description of the PR seems not the same as the code.

zhentaocc:
> The description of the PR seems not the same as the code.

Missed one commit. Now it's ready.

yangw1234:
How about separating the framework changes and example changes into two PRs? I think two PRs will make it more clear.

zhentaocc:
> How about separating the framework changes and example changes into two PRs? I think two PRs will make it more clear.

Good idea.

yangw1234:
LGTM

PR #4241, opened 2022-03-14
Chronos: document regular update

TheaperDeng:
- New updates learning from our recent customer feedbacks
- New quantization example
https://bigdl-junweid.readthedocs.io/en/document-update-0311-chronos/


TheaperDeng:
https://bigdl-junweid.readthedocs.io/en/document-update-0311-chronos/doc/Chronos/Overview/chronos.html

Issue #4219, opened 2022-03-11
BigDL PPML kms-client document and code needs to update


jenniew:
The documentation is not fit for current ehsm implementation.
And code needs to update too.

glorysdj:
@Uxito-Ada please update the doc once the EHSMKMS implementation is finished.

qiyuangong:
Issue addressed by #4230 #4234 #4243 #4242

PR #4242, opened 2022-03-14
PPML kms-client python update

jenniew:
PPML kms-client python update.
Remove options for appid and apikey. use environment variable.


jenniew:
Related issue: https://github.com/intel-analytics/BigDL/issues/4219

PR #4256, opened 2022-03-16
Nano: Add openvino support

zhentaocc:
In this PR, users can choose to use opevino as backend inference engine. 
# Usage
```python
        model = mobilenet_v3_small(num_classes=10)
        pl_model = Trainer.compile(model, openvino=True)
        y = torch.ones((10, ), dtype=torch.long)

        pl_model.eval_openvino(x)
        y = pl_model(x)
        pl_model.exit_openvino()
```
or 
```python
        trainer = Trainer(max_epochs=1)
        model = mobilenet_v3_small(num_classes=10)
        pl_model = Trainer.compile(model, loss=torch.nn.CrossEntropyLoss(),
                                   optimizer=torch.optim.SGD(model.parameters(), lr=0.01),
                                   openvino=True)
        y = torch.ones((10, ), dtype=torch.long)
        ds = TensorDataset(x, y)
        # TODO: batch size = 2 will cause error, need to fix this
        dataloader = DataLoader(ds, batch_size=10)
        trainer.fit(pl_model, dataloader)

        # Test if eval_openvino() and exit_openvino() work
        pl_model.eval_openvino()
        y = pl_model(x)
        pl_model.exit_openvino()
```
There are 3 ways to exit openvino mode.
- `model.exit_openvino()`
- `model.train()`
- `model.eval()`

It will automatically fall back to pytorch backend if users run fit, train, validate, test.
# Unresolved issue

- Can't convert a onnx model with dynamic axes to OpenVINO, need to find another function to do that
- With static axes, for example the input is [2, 3, 256, 256], the actual input should be broadcasted. Otherwise it's not accepted.
In this case, batch 1, 2 are acceptable while batch >2 will cause error. Will try to fix this by converting a model with dynamic axes.


jason-dai:
I think we should make sure the APIs for OpenVINO and ONNX RT consistent with each other (including `model.eval_onnx`, `model.eval_openvino`, `trainer.compile`, `inference`), instead of introducing new APIs such as `trainer = Trainer(openvino=True)`

python/nano/src/bigdl/nano/pytorch/lightning.py
```diff
@@ -21,23 +21,27 @@
 from torch import nn, Tensor
 from torch.nn.modules.loss import _Loss
 from torch.optim import Optimizer
+from torch.optim.lr_scheduler import _LRScheduler
 
 
 class LightningModuleFromTorch(LightningModule):
     def __init__(self, model: nn.Module, loss: _Loss = None, optimizer: Optimizer = None,
+                 scheduler: _LRScheduler = None,
```
jason-dai, line 29: Why this change? This PR should focus on OpenCV support; you may submit a separate PR for this change if needed.

python/nano/src/bigdl/nano/pytorch/lightning.py
```diff
@@ -48,6 +52,10 @@ def _forward(self, batch):
         nargs = self.model.forward.__code__.co_argcount
         return self(*(batch[:nargs - 1]))
 
+    def on_train_start(self) -> None:
```
jason-dai, line 55: Why this change? This PR should focus on OpenCV support; you may submit a separate PR for this change if needed.

python/nano/src/bigdl/nano/pytorch/lightning.py
```diff
@@ -56,30 +64,31 @@ def training_step(self, batch, batch_idx):
 
     def validation_step(self, batch, batch_idx):
         y_hat = self._forward(batch)
-        loss = self.loss(y_hat, batch[-1])  # use last output as target
-        self.log("val/loss", loss, on_epoch=True, prog_bar=True, logger=True)
+        if self.loss:
+            loss = self.loss(y_hat, batch[-1])  # use last output as target
+            self.log("val/loss", loss, on_epoch=True,
```
jason-dai, line 69: Why this change? This PR should focus on OpenCV support; you may submit a separate PR for this change if needed.

python/nano/src/bigdl/nano/pytorch/lightning.py
```diff
@@ -56,30 +64,31 @@ def training_step(self, batch, batch_idx):
...
+                     prog_bar=True, logger=True)
         if self.metrics:
-            acc = {"val/" + type(metric).__name__: metric(y_hat, batch[-1])
+            acc = {"val/{}_{}".format(type(metric).__name__, i): metric(y_hat, batch[-1])
```
jason-dai, line 72: Why this change? This PR should focus on OpenCV support; you may submit a separate PR for this change if needed.

python/nano/src/bigdl/nano/pytorch/lightning.py
```diff
@@ -56,30 +64,31 @@ def training_step(self, batch, batch_idx):
...
-        else:
-            acc = None
-        return loss, acc
+
+    def predict_step(self, batch, batch_idx):
+        return self(*batch)
 
     def configure_optimizers(self):
-        return self.optimizer
+        optimizers = [self.optimizer]
```
jason-dai, line 87: Why this change? This PR should focus on OpenCV support; you may submit a separate PR for this change if needed.

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
```diff
@@ -116,6 +117,7 @@ def __init__(self, num_processes: int = 1,
     def compile(model: nn.Module,
                 loss: _Loss = None,
                 optimizer: torch.optim.Optimizer = None,
+                scheduler: _LRScheduler = None,
```
jason-dai, line 120: Why this change? This PR should focus on OpenCV support; you may submit a separate PR for this change if needed.

python/nano/test/pytorch/tests/test_openvino.py
```diff
@@ -0,0 +1,42 @@
...
+        ds = TensorDataset(x, y)
+        dataloader = DataLoader(ds, batch_size=2)
+        trainer.test(pl_model, dataloader)
+        trainer.validate(pl_model, dataloader)
+
+        ds = TensorDataset(x)
+        dataloader = DataLoader(ds, batch_size=2)
+        trainer.predict(pl_model, dataloader)
+
+        trainer.inference(pl_model, x, backend='openvino')
```
jason-dai, line 42: we need to support transparently acceleration using OpenVINO, such as:
```
model.eval_openvino()
out=model(input)
```

zhentaocc:
@jason-dai @TheaperDeng @yangw1234 Please take a look.

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
```diff
@@ -119,7 +119,8 @@ def compile(model: nn.Module,
                 scheduler: _LRScheduler = None,
                 metrics: List[Metric] = None,
                 onnx: bool = False,
-                quantize: bool = True):
+                quantize: bool = True,
```
jason-dai, line 122: why are we making `quantize` to true by default? @TheaperDeng

TheaperDeng: At that time, just to make sure quantized model can be loaded with
```python
model = Trainer.compile(model)
model.load_quantized_state_dict(torch.load(qfile))
```
And it shouldn't make too much of a difference if we ask users to explicitly set `quantize=True`
```python
model = Trainer.compile(model, quantize=True)
model.load_quantized_state_dict(torch.load(qfile))
```
We can change it to `False` by default.

jason-dai: I think we should change it to False; we can add proper checking for all unbinded methods (such as load_quantized_state_dict) later.

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
```diff
@@ -158,7 +159,9 @@ def compile(model: nn.Module,
             except ImportError:
                 raise RuntimeError("You should install onnx and onnxruntime to set `onnx=True`, "
                                    "or just set `onnx=False`.")
-
+        elif openvino:
```
jason-dai, line 162: I think we should check the correct combinations of onnx, openvino and quantize - some of them cannot be true at the same time.

python/nano/src/bigdl/nano/pytorch/runtime_binding/openvino_inference.py
```diff
@@ -0,0 +1,183 @@
...
+
+    if set(OPENVINO_BINDED_COMPONENTS) <= set(dir(pl_model)):
+        return pl_model
+
+    # check conflicts
+    for component in OPENVINO_BINDED_COMPONENTS:
+        if component in dir(pl_model):
+            warnings.warn(f"{component} method/property will be replaced. You may rename your"
+                          " customized attributes or methods and call `Trainer.compile again `"
+                          "to avoid being overwrite.")
```
jason-dai, line 168: overwritten

python/nano/src/bigdl/nano/pytorch/runtime_binding/openvino_inference.py
```diff
@@ -0,0 +1,183 @@
...
+from bigdl.nano.pytorch.lightning import LightningModuleFromTorch
+from functools import partial, wraps
+import warnings
+import torch
+import numpy as np
+import inspect
+from openvino.inference_engine import IECore
+
+
+OPENVINO_BINDED_COMPONENTS = ['ir_up_to_date', 'ir_model']
```
jason-dai, line 29: is there a way for user to export and save the openvino model?

zhentaocc: Users can export the model by `model.eval_openvino(inputs, filepath)`.  
They can find the saved model in `filepath`.

python/nano/src/bigdl/nano/pytorch/runtime_binding/openvino_inference.py
zhentaocc: Removed this part since we just need to bind openvino method only once.

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
zhentaocc: onnx and openvino can't be True at the same time. Add error here in this case.

zhentaocc:
Add new jenkins job here: http://10.112.231.51:18888/job/ZOO-PR-nano-pytorch-openvino/7/console

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
jason-dai: Can openvino and quantize both be ture?

zhentaocc: In the future, I think yes. Right now we don't support this. Add another error for this? Shall I set quantize as False in this PR?  
Otherwise, uses have to use `Trainer.compile(model,  quantize=False, openvino=True)`.

jason-dai: Yes - set quantize to false.

python/nano/src/bigdl/nano/pytorch/runtime_binding/openvino_inference.py
jason-dai: this is not very intuitive; we may add an export method in model or trainer?

zhentaocc: Let's do this in another issue.   
https://github.com/intel-analytics/BigDL/issues/4313

TheaperDeng:
need `quantize=True` here https://github.com/intel-analytics/BigDL/blob/2f22b03a599aed56270d71f65dee5806cf329e2b/python/chronos/src/bigdl/chronos/forecaster/base_forecaster.py#L388

yangw1234:
can we make the code structure look like the one defined in https://github.com/intel-analytics/BigDL/issues/4288

```
bigdl/nano/deps/openvino/openvion_api.py
bigdl/nano/deps/openvino/other_openvino_code_1.py
bigdl/nano/deps/openvino/other_openvino_code_2.py
```

jason-dai:
> can we make the code structure look like the one defined in #4288
> 
> ```
> bigdl/nano/deps/openvino/openvion_api.py
> bigdl/nano/deps/openvino/other_openvino_code_1.py
> bigdl/nano/deps/openvino/other_openvino_code_2.py
> ```

Maybe we can do this in a follow-up PR to deal with the dependency on OV, INC and ORT

yangw1234:
> > can we make the code structure look like the one defined in #4288
> > ```
> > bigdl/nano/deps/openvino/openvion_api.py
> > bigdl/nano/deps/openvino/other_openvino_code_1.py
> > bigdl/nano/deps/openvino/other_openvino_code_2.py
> > ```
> 
> Maybe we can do this in a follow-up PR to deal with the dependency on OV, INC and ORT

Sure

yangw1234:
also adding the new jenkins job to zoo-pr-nano-tests?

PR #4262, opened 2022-03-17
Add subprocess plugin for pytorch distributed training

az2181036:
Since multiprocessing has conflicts with iPython, which means it will return error in jupyter notebook environment, try to add a new distributed strategy using subprocess like what we did in tensorflow.


python/nano/src/bigdl/nano/pytorch/plugins/ddp_subprocess.py
```diff
@@ -0,0 +1,121 @@
...
+        for i in range(self.num_processes):
+            env = {
+                "KMP_AFFINITY": f"granularity=fine,proclist"
+                                f"=[{','.join([str(i) for i in cpu_procs[i]])}],explicit",
+                "OMP_NUM_THREADS": str(len(cpu_procs[i])),
+                "PROCESS_IDX": str(i),
+            }
+            if "PYTHONPATH" in os.environ:
+                env["PYTHONPATH"] = os.environ["PYTHONPATH"]
+            print(env["PYTHONPATH"])
```
yangw1234, line 82: let's remove this print.

python/nano/src/bigdl/nano/pytorch/plugins/ddp_subprocess.py
```diff
@@ -0,0 +1,121 @@
...
+    def execution_loop(self):
+        if self.is_global_zero and not torch.distributed.is_initialized():
+            log.info("-" * 100)
+            log.info(f"distributed_backend={self.distributed_backend}")
+            log.info(f"All DDP processes registered. Starting ddp with {self.world_size} processes")
+            log.info("-" * 100)
+
+        with TemporaryDirectory() as temp_dir:
+            with open(os.path.join(temp_dir, "args.pkl"), 'wb') as f:
+                queue_list = queue_dumper(self.mp_queue)
```
yangw1234, line 110: what elements are in self.mp_queue? 

Is it possible that after you call queue_dumper some code puts other things in self.mp_queue. 

You can write some comments in the code if you think this is safe.

python/nano/src/bigdl/nano/pytorch/plugins/ddp_subprocess.py
az2181036: OK

python/nano/src/bigdl/nano/pytorch/plugins/ddp_subprocess.py
az2181036: Through there is nothing in `self.mp_queue` in my test,  I copy all values in queue to a list for safe, and then pickle the list.

az2181036:
Remove print, and don't dump `self.mp_queue` in `ddp_subprocess.py`, but assign a empty queue  to `plugin.mp_queue` in `worker.py`. Test passed in my local environment.

python/nano/src/bigdl/nano/pytorch/plugins/ddp_subprocess.py
az2181036: Since `queue_dumper` is a function I wrote, I add some comments at where the function defined.

az2181036:
http://10.112.231.51:18888/job/ZOO-PR-NanoTests/355/

python/nano/src/bigdl/nano/pytorch/plugins/ddp_subprocess.py
az2181036: > Is it possible that after you call queue_dumper some code puts other things in self.mp_queue.

It will put something in `self.mp_queue` while running `worker.py`, but I transfer its values when `worker.py` is finished. In my view, this queue is used for passing message in multiprocessing environment, specifically, it is used for transfer `run_stage()` results to the main process. But we use `pickle` to finish the job since Queue isn't supported in `subprocess`.

It's OK if we remove `queue_dumper` which has been tested in my next commit.

az2181036:
http://10.112.231.51:18888/job/ZOO-PR-NanoTests/371/

yangw1234:
LGTM

PR #4263, opened 2022-03-17
Tiny fixes on lightning

zhentaocc:
- Carefully handle loss, metrics in LightningModule
- add `pred_step`
- Add argument `scheduler` to configure scheduler
- Add argument `scheduler` to compile scheduler through `trainer.compile`
- Rename metrics with indexes as suffix to avoid different metrics with same class

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/lightning.py
+++ b/python/nano/src/bigdl/nano/pytorch/lightning.py
@@ -80,6 +80,9 @@ class LightningModuleFromTorch(LightningModule):
                    for i, metric in enumerate(self.metrics)}
             self.log_dict(acc, on_epoch=True, prog_bar=True, logger=True)
 
+    def predict_step(self, batch, batch_idx):
+        return self(*batch)
+
     def configure_optimizers(self):
         return [self.optimizer], [self.scheduler]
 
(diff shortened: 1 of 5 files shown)


python/nano/src/bigdl/nano/pytorch/lightning.py
```diff
@@ -56,30 +64,27 @@ def training_step(self, batch, batch_idx):
...
-        else:
-            acc = None
-        return loss, acc
+
+    def predict_step(self, batch, batch_idx):
+        return self(*batch)
 
     def configure_optimizers(self):
-        return self.optimizer
+        return [self.optimizer], [self.scheduler]
```
yangw1234, line 87: is it ok to pass a scheduler that is `None`

yangw1234:
LGTM, 
except that we need to pay attention whether a None works in `configure_optimizer`

zhentaocc:
http://10.112.231.51:18888/job/ZOO-PR-NanoTests/333/

python/nano/src/bigdl/nano/pytorch/lightning.py
zhentaocc: resolved

yangw1234:
LGTM

PR #4264, opened 2022-03-17
[PPML] Spark k8s: removed Kubeconfig requirements

ColossusChang:
Spark k8s client mode does not need to generate Kubeconfig requirements now.


glorysdj (approved):
LGTM

PR #4266, opened 2022-03-17
Fix dien preprocessing

hkvision:


hkvision:
jenkins passed: http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-PR-Friesian-Validation/295/

PR #4268, opened 2022-03-18
[PPML] Change enclave key to secret

ColossusChang:
- The enclave key is now mounted as a secret. No need to put it in the nfs anymore.
- Enabled argument passing for the job submission script.


glorysdj (approved):
LGTM

PR #4276, opened 2022-03-21
hnsw similarity train

songhappy:


jason-dai:
Is this PR for training or nearline pipelines?

scala/friesian/pom.xml
```diff
@@ -249,6 +249,14 @@
             <artifactId>log4j-1.2-api</artifactId>
             <version>${log4j.version}</version>
         </dependency>
+
+        <dependency>
```
jason-dai, line 253: why do we need this?

scala/friesian/src/main/scala/com/intel/analytics/bigdl/friesian/nearline/utils/NearlineHelper.scala
```diff
@@ -29,6 +29,10 @@ class NearlineHelper extends Serializable {
   @BeanProperty var redisClusterItemSlotType = 0
   @BeanProperty var redisUrl = "localhost:6379"
 
+  // item similarity service atributes
+  @BeanProperty var initialItemNeighborPath: String = _
+  @BeanProperty var itemNeighborColumn: String = _
```
jason-dai, line 34: where are these properties configed?

jason-dai:
Need to add proper python/scala doc and documents to describe the features

scala/friesian/src/main/java/com/intel/analytics/bigdl/friesian/serving/similarity/SimilarityServer.java
```diff
@@ -0,0 +1,134 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.core.config.Configurator;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+public class SimilarityServer extends GrpcServerBase {
```
jason-dai, line 43: I don't think we need to add a similarity service

jason-dai:
I don't think we need hnsw service/examples here; we can extend recall/feature service to support item similarity search in a separate PR.

PR #4291, opened 2022-03-23
Nano: Add 3 original examples for onnxruntime inference (pytorch/pytorch-lightning)

TheaperDeng:
This PR only upload 3 original example adapted from pytorch and pytorch-lightning and their readme(stated the example is WIP)

3 other PRs will be raised for each example. This PR is used to clearly show what is changed from the original code.


TheaperDeng:
http://10.112.231.51:18888/job/ZOO-PR-NanoTests/360/

PR #4299, opened 2022-03-24
[PPML]Occlum JDK change to JDK-8

Forest216:
JDK-11 have some problem on spark 3.1.2. Need to change to JDK-8


qiyuangong:
PR LGTM. Need to pass sparksql unit test

qiyuangong:
SparkSQL Unit Test passed in most cases. But this PR is not necessary now. Closed

PR #4301, opened 2022-03-24
add nano pytorch example cat-vs-dog

Django-Jiang:


python/nano/example/pytorch/pl-cat-vs-dog/pl-cat-vs-dog.py
```diff
@@ -0,0 +1,65 @@
+import os
```
yangw1234, line 1: You should add the license header.

See other files for reference.

python/nano/example/pytorch/nano-cat-vs-dog/README.md
```diff
@@ -0,0 +1,65 @@
+# Nano Pytorch cat-vs-dog Example
```
yangw1234, line 1: let's move this to another standalone directory called `benchmark`

python/nano/example/pytorch/nano-cat-vs-dog/nano-cat-vs-dog.py
```diff
@@ -0,0 +1,94 @@
+#
```
yangw1234, line 1: I think we only need one `.py` file for performance testing

python/nano/example/pytorch/pl-cat-vs-dog/pl-cat-vs-dog.py
Django-Jiang: Added

yangw1234:
LGTM

PR #4305, opened 2022-03-24
[PPML] fixing more scripts

ColossusChang:
- Some fixes that were overlooked in previous PRs. Will update as Shaojie carries on with the tests.


glorysdj (approved):
LGTM

PR #4306, opened 2022-03-24
Test about nano workload performance analysis

az2181036:
It's an early draft, which will continue on next Tuesday.


yangw1234:
we can just record the end to end throughput

yangw1234:
you can refer [here](https://github.com/analytics-zoo/bigdl-nano-benchmarks/pull/21/files#diff-b907ba52349df6bd19e27cd34361cbc8efea45edeb2fe3e374803714a127563aR63-R71) for getting throughput

PR #4309, opened 2022-03-25
Add data creator function support for tf.keras in AutoEstimator

yushan111:
This PR adds data creator function input support for `AutoEstimator.from_keras`, which takes config as input and return a tensorflow Dataset.
For tuple of numpy input, we will also convert to tensorflow Dataset internally and then feed into model.


yushan111:
http://10.112.231.51:18888/job/BigDL-Orca-PR-Validation/515/

PR #4311, opened 2022-03-25
Ray api

DirkFi:
- Move ray part in nano to deps/ray
- Add ray api to make it handy to use ray without importing it from the beginning
- Add init.py


python/nano/src/bigdl/nano/deps/ray/ray_distributed.py
```diff
@@ -45,7 +45,8 @@
 from ray.util.sgd.utils import find_free_port
 from torch.nn import Module
 
-from .environments.ray_envbase import RayEnvironment
+from bigdl.nano.deps.ray.ray_envbase import RayEnvironment
+from ray_api import create_ray_envbase
```
yangw1234, line 49: I think you should not call `ray_api` inside of `deps/ray/` to avoid circular dependency.

DirkFi: OK.

DirkFi:
Newest commit of pr validation is [here](http://10.112.231.51:18888/job/ZOO-PR-nano-ray-test/9/).

yangw1234:
http://10.112.231.51:18888/view/BigDL-2.0-PR/job/ZOO-PR-NanoTests/364/

PR #4312, opened 2022-03-25
[Orca] Support Ray Dataset Input in TF2Estimator on Ray Backend

sgwhat:
This PR supports the ray dataset as a backend in TF2 ray estimator, users just need to read data by ray data read_api to get a ray dataset. The estimator would automatically convert the input ray dataset to a TF dataset and conduct the training pipeline.

```
def transform():
      ...

ray_dataset = ray.data.read_csv(path)
dataset = ray_dataset.map(transform)

orca_estimator = Estimator.from_keras(backend="tf2")
stats = orca_estimator.fit(data=dataset, 
                           label_cols=str, feature_cols=Optional[List[str]],
                           data_config = {"output_signature": Tuple["tf.TypeSpec", "tf.TypeSpec"]})
```

- [x] Fit
- [x] Evaluate


sgwhat:
Pending add ut and update tf_runner

python/orca/src/bigdl/orca/learn/tf2/tf_runner.py
```diff
@@ -196,6 +196,13 @@ def dataset_fn(input_context):
         return dataset
 
     def _handle_sharding(self, dataset):
+        import tensorflow as tf
+        if isinstance(dataset, tf.data.Dataset):
+            options = tf.data.Options()
+            # Disable Tensorflow autosharding since the dataset has already been sharded.
+            options.experimental_distribute.auto_shard_policy = \
+                tf.data.experimental.AutoShardPolicy.OFF
+            dataset = dataset.with_options(options)
```
yushan111, line 205: What if users customize a data creator function?  In that case, the dataset might not been sharded.
Shall we just put this part to the data creator function for ray dataset?

python/orca/test/bigdl/orca/learn/ray/tf/test_tf2estimator_ray_dataset.py
```diff
@@ -0,0 +1,120 @@
...
+        data_config_args = {"output_signature":
+                            (tf.TensorSpec(shape=(None, 1), dtype=tf.float32),
+                             tf.TensorSpec(shape=(None), dtype=tf.float32))
+                            }
+        
+        start_eval_stats = orca_estimator.evaluate(data=validation_dataset,
+                                                   num_steps=2, batch_size=32,
+                                                   label_cols="x", feature_cols=["y"],
+                                                   data_config=data_config_args)
+        print(start_eval_stats)
```
yushan111, line 100: maybe we should check whether the stats are as expected instead of just printing them out, so that the test could fail if anything goes wrong.

python/orca/src/bigdl/orca/learn/tf2/tf_runner.py
sgwhat: Sure. That's a good idea.😄

sgwhat: > What if users customize a data creator function? In that case, the dataset might not been sharded. Shall we just put this part to the data creator function for ray dataset?

Done.

python/orca/src/bigdl/orca/learn/tf2/ray_estimator.py
```diff
@@ -275,6 +305,28 @@ def transform_func(worker, partition_refs):
...
+                tf_dataset = tf_dataset.with_options(options)
+                return tf_dataset
+
+            remote_worker_stats = []
+            for shard, worker in zip(shards, self.remote_workers):
+                params["data_creator"] = data_creator
+                stats = worker.validate.remote(**params)
+                remote_worker_stats.append(stats)
+            worker_stats = ray.get(remote_worker_stats)
+            worker_stats = list(itertools.chain.from_iterable(worker_stats))
```
yushan111, line 329: It seems to be duplicated with the code in `fit`. Maybe we could refactor it.

sgwhat:
Jenkins Passed: http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-Orca-PR-Validation/538/

Issue #4288, opened 2022-03-23
[Nano] Optional dependency management


yangw1234:
# Problem
To maintain the composability of Nano, we make some of the dependencies optional if the corresponding feature is not used. 

However, code related to these optional dependencies is spread in many places in the code base, which makes it **hard to maintain what optional dependencies we currently have and hard to make sure normal code can be executed without meeting "ImportError"**.

# Proposal

## Code Structure
Each optional dependency has its own directory and all code related to this dependency should go into this directory. In addition, each dependency should have a single python file exposing all necessary interfaces to other components. This file should import its dependency lazily. 

E.g.
```
bigdl/nano/deps/inc/inc_api.py # should not import inc in the top-level or lazy import[1]
bigdl/nano/deps/inc/internal_inc_funcs.py # can import inc at the top-level
```
[1] lazy import, https://github.com/intel/neural-compressor/blob/master/neural_compressor/adaptor/tensorflow.py#L31

## Testing
Each dependency should have its own testing environment. 
Normal test cases should be executed in a environment without any of the dependencies.

## Optional Dependencies

- [x] INC
- [x] OPENVINO
- [x] ONNXRT ?
- [x] Horovod
- [x] Ray
- [x] IPEX

yangw1234:
@TheaperDeng @zhentaocc what do you think?

zhentaocc:
Agree on that. I was trying to refactor testing before https://github.com/intel-analytics/BigDL/issues/3919 but little progress on this.  
One note: seems onnxruntime and inc should be tested both separately and together.

TheaperDeng:
Agree on the suggestion and I think Lazy Import might be the sliver bullet. 

BTW, for ut, Chronos recently adopt a pytest feature for different env: https://github.com/intel-analytics/BigDL/blob/main/python/chronos/test/bigdl/chronos/model/tf2/test_Seq2Seq_keras.py#L44

TheaperDeng:
just update: working on onnxrt and onnx

PR #4328, opened 2022-03-30
[Nano]refactor openvino into independent module/directory

zhentaocc:
This is to decouple all openvino usage with basic pytorch and other dependencies(onnx, inc, ...) and gather openvino related code into one directory as a single module.   
Only two APIs are to be exposed to users:

- `bigdl.nano.deps.openvino.torch.export(...)` for users to export pytorch model and save as IR of xml format.
- `bigdl.nano.deps.openvino.torch.bind_openvino_method(...)` for trainer to bind `eval_openvino` and `exit_openvino` into pytorch model for transparently using openvino backend.  

### Inheritance Design
```python
class PytorchBaseInference:
    pass
class BaseOpenVINOInference:
    pass
class PytorchOpenVINOInference(BaseOpenVINOInference, PytorchBaseInference):
    pass
class TensorflowOpenVINOInference(BaseOpenVINOInference, PytorchBaseInference):
    # TODO
    pass
class PytorchONNXInference(BaseOpenVINOInference, PytorchBaseInference):
    pass
class QuantizeInference(PytorchBaseInference):
    pass
class PytorchONNXQuantizeInference(QuantizeInference, PytorchONNXInference):
    pass
```
Currently we can see strong connection between different dependencies. With multiple class inheritance, we can separate the code by their dependencies.


jason-dai:
Does it support lazy import (as proposed in https://github.com/intel-analytics/BigDL/issues/4288)?

And `bigdl.nano.deps.openvino.torch.export` is a pretty bad interface for the user; maybe `model.export`?

zhentaocc:
> Does it support lazy import (as proposed in #4288)?
> 
> And `bigdl.nano.deps.openvino.torch.export` is a pretty bad interface for the user; maybe `model.export`?

LazyImport is supported now.  
`model.export_openvino` is now ready.  But shall we keep similar export usage as `torch.onnx.export()`?

zhentaocc:
Remove lazy import.  
Seems no need to introduce extra lazy import here.

zhentaocc:
http://10.112.231.51:18888/job/ZOO-PR-nano-pytorch-openvino-new/2/
http://10.112.231.51:18888/job/ZOO-PR-NanoTests/400/

yangw1234:
LGTM

Issue #4170, opened 2022-03-07
Nano: pytorch `load_quantized_state_dict` does not work as expected while the quantization config is tuned


TheaperDeng:
Thanks to INC team's feedback, we should move this implementation to using `neural-compressor.utils.pytorch.load` https://github.com/intel/neural-compressor/blob/561a47e8f45b97cdb4eb35fb88b14950e576b461/neural_compressor/utils/pytorch.py#L68

https://github.com/intel-analytics/BigDL/blob/315dd847dfde9a52d4733cc51380104a81e43250/python/nano/src/bigdl/nano/pytorch/runtime_binding/quantization_inference.py#L88-L105

PR #4330, opened 2022-03-31
Nano: bugfix for quantizated model load/save

TheaperDeng:
related to #4170


TheaperDeng:
http://10.112.231.51:18888/job/ZOO-PR-NanoTests/361/

PR #4350, opened 2022-04-05
[PPML] Correctly read device files whose size are always zero

lishen-nt:
In Gramine, the size of /dev/attestation/quote is always zero. Since the size of an actual quote (DCAP) is not a fixed value, BufferedInputStream is used to read the whole quote. Then the quote Array size is validated.

### DIFF
--- a/scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/attestation/GramineQuoteGeneratorImpl.scala
+++ b/scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/attestation/GramineQuoteGeneratorImpl.scala
@@ -29,8 +29,8 @@ import scala.collection.Iterator
 class GramineQuoteGeneratorImpl extends QuoteGenerator {
 
   val logger = LogManager.getLogger(getClass)
-  val UserReportPath = "/dev/attestation/user_report_data"
-  val QuotePath = "/dev/attestation/quote"
+  val USER_REPORT_PATH = "/dev/attestation/user_report_data"
+  val QUOTE_PATH = "/dev/attestation/quote"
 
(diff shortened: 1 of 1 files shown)


scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/attestation/GramineQuoteGeneratorImpl.scala
```diff
@@ -17,20 +17,20 @@
...
 /**
  * QuoteGenerator for Gramine (https://github.com/gramineproject/gramine)
  */
 class GramineQuoteGeneratorImpl extends QuoteGenerator {
 
   val logger = LogManager.getLogger(getClass)
-
-  val userReportPath = "/dev/attestation/user_report_data"
-  val quotePath = "/dev/attestation/quote"
+  val UserReportPath = "/dev/attestation/user_report_data"
```
qiyuangong, line 32: Change to all uppercase .e.g., MAGIC_NUM

qiyuangong:
> In Gramine, the size of /dev/attestation/quote is always zero. Since the size of an actual quote (DCAP) is not a fixed value, BufferedInputStream is used to read the whole quote. Then the quote Array size is validated.

Thank you for providing this fix! :)

scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/attestation/GramineQuoteGeneratorImpl.scala
lishen-nt: Thanks for the feedback. The style of the two constants are changed.

qiyuangong (approved):
LGTM

PR #4353, opened 2022-04-06
[Nano] ipex deps add

Django-Jiang:


Django-Jiang:
http://10.112.231.51:18888/job/ZOO-PR-NanoTests/391/

PR #4354, opened 2022-04-06
[Orca] Update Orca PyTorch Estimator Examples for Testing on Ray Runtime

sgwhat:
This PR is for testing orca pytorch estimator examples on ray runtime.


sgwhat:
Jenkins passed: http://10.112.231.51:18888/job/BigDL-PRVN-ray-ctx/229/

PR #4356, opened 2022-04-06
Nano: support an option to start process by "fork" on UNIX-like OS for multiprocessing training

TheaperDeng:
Since "fork" does not require any serialization, this might be a life-saver for our users who have
1. components that can not be serialized (e.g. `threading.Lock`)
2. closure (e.g. decorator)

I am not sure if current API design is the best.
Current:
```python
trainer = Trainer(distributed_backend="spawn", process_start_method="fork")
```
Is this confusing to our users? we may just create another backend called "fork" and make our users use it like:
```python
trainer = Trainer(distributed_backend="fork")
```
-------------------------
Update: 4/25: discussed with @yangw1234 
Will change this PR's implementation to 
```python
trainer = Trainer(distributed_backend="fork")
```
this design

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/plugins/ddp_spawn.py
+++ b/python/nano/src/bigdl/nano/pytorch/plugins/ddp_spawn.py
@@ -118,6 +118,8 @@ class DDPSpawnPlugin(pl.plugins.DDPSpawnPlugin):
         cpu_for_each_process: Optional[List[List[int]]] = None,
         use_ipex=False,
         enable_bf16=False,
+        process_start_method: str = "spawn",
+        **kwargs: Any,
     ):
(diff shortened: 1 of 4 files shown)


jason-dai:
Does fork work for Jupyter notebook @yangw1234 ?

TheaperDeng:
need more experiment to verify if "fork" works just like "spawn" and does not have any side-effects.

TheaperDeng:
There seems to be a dead lock issue if users don't use intel-openmp (i.e. not using `bigdl-nano-init`)

TheaperDeng:
> Does fork work for Jupyter notebook @yangw1234 ?

Yes, just had a verification run.

TheaperDeng:
This PR is on hold since deadlock problem is unsolved. https://pythonspeed.com/articles/python-multiprocessing/

PR #4358, opened 2022-04-06
[Orca] Support Ray Runtime in TF2Estimator Examples

sgwhat:
This PR is mainly for supporting ray runtime as backend in TF2Estimator Examples.

- [x] Update examples.
- [x] Test Resnet


sgwhat:
Waiting for http://10.239.45.10:8081/repository fixed.

python/orca/example/learn/tf2/yolov3/yoloV3.py
```diff
@@ -575,33 +577,36 @@ def main():
...
+                                    "spark.kubernetes.driver.volumes.persistentVolumeClaim."
+                                    "nfsvolumeclaim.options.claimName": "nfsvolumeclaim",
+                                    "spark.kubernetes.driver.volumes.persistentVolumeClaim."
+                                    "nfsvolumeclaim.mount.path": options.nfs_mount_path})
+        elif options.cluster_mode.startswith("yarn"):
+            init_orca_context(cluster_mode=options.cluster_mode, cores=options.cores,
+                            num_nodes=options.worker_num, memory=options.memory,
+                            init_ray_on_spark=True, enable_numa_binding=options.enable_numa_binding,
+                            object_store_memory=options.object_store_memory)
+        elif options.cluster_mode == "spark-submit":
```
jason-dai, line 608: I think the default behavior of `init_orca_context` should be the same as "spark-submit"?

sgwhat:
YoloV3 cannot support ray runtime, because `read_parquet` requires spark context.

python/orca/example/learn/tf2/yolov3/yoloV3.py
sgwhat: > I think the default behavior of `init_orca_context` should be the same as "spark-submit"?

Yes, when `activate spark context`.

sgwhat: > I think the default behavior of `init_orca_context` should be the same as "spark-submit"?

Only when there is an activate spark context, the default behavior is the same as "spark-submit".

sgwhat: > I think the default behavior of `init_orca_context` should be the same as "spark-submit"?

Yes, when there is an activate spark context.

sgwhat:
Jenkins passed: http://10.112.231.51:18888/job/BigDL-Orca-PR-Validation/555/

Issue #4318, opened 2022-03-29
dllib keras like model summary support returning string instead of printing in console


yangw1234:
from user group

https://groups.google.com/g/bigdl-user-group/c/jhxLZfvG4Ac/m/WtymMK1yAgAJ?utm_medium=email&utm_source=footer

PR #4361, opened 2022-04-06
dllib keras like model summary support returning string

dding3:
Fix https://github.com/intel-analytics/BigDL/issues/4318

### DIFF
--- a/python/dllib/src/bigdl/dllib/keras/engine/topology.py
+++ b/python/dllib/src/bigdl/dllib/keras/engine/topology.py
@@ -500,7 +500,7 @@ class KerasNet(ZooKerasLayer):
                    If the field has a larger length, the remaining part will be trimmed.
                    If the field has a smaller length, the remaining part will be white spaces.
         """
-        callZooFunc(self.bigdl_type, "zooKerasNetSummary",
+        return callZooFunc(self.bigdl_type, "zooKerasNetSummary",
                     self.value,
                     line_length,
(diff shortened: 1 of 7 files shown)


dding3:
jekins pass. http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-PR-Dllib-Validation/398/

python/dllib/src/bigdl/dllib/keras/engine/topology.py
```diff
@@ -500,10 +500,10 @@ def summary(self, line_length=120, positions=[.33, .55, .67, 1.]):
                    If the field has a larger length, the remaining part will be trimmed.
                    If the field has a smaller length, the remaining part will be white spaces.
         """
-        callZooFunc(self.bigdl_type, "zooKerasNetSummary",
-                    self.value,
-                    line_length,
-                    [float(p) for p in positions])
+        return callZooFunc(self.bigdl_type, "zooKerasNetSummary",
```
jason-dai, line 503: I think a more reasonable way to deal with it is:
1) Calling `summary` in Scala should print the message in Scala
2) Calling `summary` in Python should print the message in Python (no print in Scala)
3) `summary` can optionally return a string

yangw1234:
It is ready to update on the user group thread requesting this feature?

dding3:
I am thinking waiting one more day for the nightly build whl?

PR #4362, opened 2022-04-07
[Orca] Ray Dataset Input Example of PyTorch Estimator

sgwhat:
This example is adapted from
 https://github.com/ray-project/ray/blob/releases/1.9.0/doc/examples/datasets_train/datasets_train.py, 
which mainly handles the data ingest with Ray Datasets and trains a PyTorch deep learning model.

### DIFF
--- a/python/orca/example/learn/pytorch/ray_dataset/README.md
+++ b/python/orca/example/learn/pytorch/ray_dataset/README.md
@@ -1,0 +1,49 @@
+# PyTorch with Ray Dataset Input Example
+
+This example is adapted from
+https://github.com/ray-project/ray/blob/releases/1.9.0/doc/examples/datasets_train/datasets_train.py
+
+In this example, you can learn how to use Ray Datasets as a first-class input in Orca PyTorch Estimator to conduct a distributed training pipeline.
+
+## Prepare the environment
+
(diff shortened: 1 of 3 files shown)


jason-dai:
1) This example should only support "ray" runtime, no?

2) We may change rename backends of the estimators to "bigdl", "spark" and "ray"? @yushan111 

3) If it runs on "ray" runtime, the estimators should always use "ray" backend, no?

sgwhat:
> 1. This example should only support "ray" runtime, no?
> 2. We may rename backends of the estimators to "bigdl", "spark" and "ray"? @yushan111
> 3. If it runs on "ray" runtime, the estimators should always use "ray" backend, no?

1. Yes, this example only supports "ray" runtime.
3. Yes, we should always use renamed "ray" backend.

yushan111:
> Maybe "bigdl", "spark", "horovod" and "ray"? @yushan111

Sure. I think both backend names of "torch_distrbuted" and "tf2" could be renamed to "ray"

yushan111:
Maybe we don't need users to parse args of "runtime" and "backend", since the example could only run on "ray" runtime and the current "torch_distributed" backend.

sgwhat:
> Maybe we don't need users to parse args of "runtime" and "backend", since the example could only run on "ray" runtime and the current "torch_distributed" backend.

Sure, will be updated soon.

Issue #4360, opened 2022-04-06
Optimized inference pipeline for Nano


jason-dai:
1. Current status

    <table>
    <tr>
	    <td>
	    <td>FP32
	    <td>BF16
	    <td>IN8
    <tr>
	    <td>PyTorch
	    <td>Y
	    <td>N
	    <td>Y
    <tr>
	    <td>ONNX
	    <td>Y
	    <td>N
	    <td>Y
    <tr>
	    <td>OpenVINO
	    <td>Y
	    <td>N
	    <td>N
    </table>

    - `Trainer.compile(…, onnx=T/F, quantize=T/F, openvino=T/F)` - bind relevant methods/variables
    - `Trainer.quantize(…)` - generate quantized model (PyTorch/ONNX) 
    - `Model.eval(quantize=T/F)` - forward using (quantized) PyTorch model
    - `Model.eval_onnx(quantize=T/F)/eval_openvino()/exit_onnx()/exit_openvino()` - forward using (quantized) ONNX/OpenVINO model

3. Desired status
    - Support all combinations of the above table
    - **Compile**: `Trainer.compile()` – just bind all methods/variables? 
    - **Quantize**: `Trainer.quantize(precision=…, accelerator=…)` 
    - **Forward**: `model.eval(precision=…, accelerator=…)`? – need to call `quantize()` first? 
    - **Export/save**: `Trainer.openvino.export(precision=…)`? – how about onnx/quantized? need to be consistent 
    - **Load**: `model.load()/model.load_quantized_state_dict()`??? - need to have consistent APIs
    - **Status**: `model.eval_status()`? – every model should maintain current/default mode, and report here?
    - What's the interactions of there methods? Any other methods needed?
 
@TheaperDeng @zhentaocc @yangw1234 @shane-huang

TheaperDeng:
### Trainer.compile

Bind all (internal/external) methods to the model

```python
model = Trainer.compile(model)
```

- `model`: could be a nn.module/pl.LightningModule

**All users who want to use Nano inference API should call `Trainer.compile` with no exception**

---
### trainer.quantize

Quantize a model by a specific precision and accelerator.

```python
model = trainer.quantize(model,
                         precision="bf16"/"int8",
                         accelerator="pytorch"/"onnxruntime"/"openvino",
                         method="eager"/"fx"/"ipex"/"qlinear"...,
                         backend="inc"/"pot",
                         **kargs_inc,
                         **kargs_pot)
```

[...]

PR #4381, opened 2022-04-08
[Nano] Nano openvino new api

zhentaocc:
This PR attempts to follow the API design of #4360.
# Usage Diff
### Before
```python
  pl_model = Trainer.compile(model, openvino=True)
  pl_model.eval_openvino(x)
  y = pl_model(x)
```
# After
```python 
openvino_model = trainer.trace(model: nn.Module, x, accelerator='openvino')
y = openvino_model(x)
```
### `model.save`
```python
openvino_model = trainer.trace(model: nn.Module, x, accelerator='openvino')
openvino_model.save('model.xml')
```
### `trainer.save`
save from `PytorchOpenvinoModel`
```python
openvino_model = trainer.trace(model: nn.Module, x, accelerator='openvino')
trainer.save(openvino_model, 'model.xml`)
```
or save from torch model directly:
```python
trainer.save(model: nn.Module, 'model.xml`, input_sample=x, accelerator='openvino', input_sample=x)
```

## Load
```python
openvino_model  = PytorchOpenVINOModel.load('model.xml')
```
```python
openvino_model = Trainer.load('saved_openvino_model.xml', accelerator='openvino')
```


yangw1234:
How about using python context manager to support automatically exiting after coming out of the "with" scope?

e.g.
```python
with pl_model.eval(...):
   pl_model.forward(...)
```

zhentaocc:
> How about using python context manager to support automatically exiting after coming out of the "with" scope?
> 
> e.g.
> 
> ```python
> with pl_model.eval(...):
>    pl_model.forward(...)
> ```

As discussed in #4360 , we renew the usage of openvino.
```python
openvino_model = trainer.trace(model, input_sample, accelerator='openvino')
y= openvino_model(x)
```

zhentaocc:
@yangw1234 @jason-dai any comment?

python/nano/test/openvino/pytorch/test_openvino.py
```diff
@@ -81,3 +82,27 @@ def test_openvino_inputsample_from_trainloader(self):
...
+
+        # trace a torch model 
+        openvino_model = trainer.trace(model, x, 'openvino')
+        y_hat = openvino_model(x)
+        assert y_hat.shape == (10, 10)
+
+        # trace pytorch-lightning model
+        pl_model = Trainer.compile(model, loss=torch.nn.CrossEntropyLoss(),
+                                   optimizer=torch.optim.SGD(model.parameters(), lr=0.01),
+                                   openvino=True)
```
jason-dai, line 101: remove the `openvino` parameter from `Trainer.compile`, and just call `bind_openvino_methods` unconditionally

python/nano/src/bigdl/nano/deps/inference/model.py
```diff
@@ -0,0 +1,123 @@
...
+                input_sample = tuple(next(iter(model.predict_dataloader())))
+        else:
+            raise RuntimeError("You must specify an input_sample or call `Trainer.fit` "
+                               "on the model first to use `eval_openvino`")
+
+    model.example_input_array = input_sample
+    return input_sample
+
+
+def export(model, input_sample=None, onnx_path="model.onnx", dynamic_axes=True):
```
jason-dai, line 98: why onnx specific code in this class?

python/nano/src/bigdl/nano/deps/inference/model.py
```diff
@@ -0,0 +1,123 @@
...
+# limitations under the License.
+#
+from typing import List, Tuple
+from bigdl.nano.pytorch.lightning import LightningModuleFromTorch
+import inspect
+from torch.utils.data import DataLoader
+import torch
+
+
+class AcceleratedLightningModule(LightningModuleFromTorch):
```
jason-dai, line 23: This should be a generic class and not in `deps` module?

python/nano/src/bigdl/nano/deps/inference/model.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def on_init_end(self, *args):
+        pass
+
+    def forward(self, *inputs):
+        inputs = self.on_forward_start(inputs)
+        outputs = self.forward_step(*inputs)
+        return self.on_forward_end(outputs)
+
+    def train(self, mode=True):
```
jason-dai, line 36: what does `mode` mean?

python/nano/test/openvino/pytorch/test_openvino.py
```diff
@@ -81,3 +82,27 @@ def test_openvino_inputsample_from_trainloader(self):
...
+
+    def test_trainer_trace_openvino(self):
+        trainer = Trainer(max_epochs=1)
+        model = mobilenet_v3_small(num_classes=10)
+        
+        x = torch.rand((10, 3, 256, 256))
+        y = torch.ones((10, ), dtype=torch.long)
+
+        # trace a torch model 
+        openvino_model = trainer.trace(model, x, 'openvino')
```
jason-dai, line 94: `trace` without `compile`?

python/nano/test/openvino/pytorch/test_openvino.py
zhentaocc: This ut should be removed after we make sure all our new APIs work.  
I didn't remove the old API using compile to to openvino optimization.  
Shall we remove it in this PR?

python/nano/src/bigdl/nano/deps/inference/model.py
zhentaocc: This is inference-specific code and no onnx-related import here.  
Here `export` converts a pytorch model to onnx.

python/nano/src/bigdl/nano/deps/inference/model.py
zhentaocc: Yes, this class doesn't introduce any new module. But it intends to be a base class of onnx, openvino, inc model.  
So if I put them in pytorch folder, then onnx, openvino, inc will call `bigdl.nano.pytorch.xxx` and be called by `bigdl.nano.pytorch.trainer`, then their relation is a bit confusing in my perspective.

python/nano/src/bigdl/nano/deps/inference/model.py
zhentaocc: `model.eval()` actually calls `train(mode=False)`.  
So with modifications on `train`, we can change the behavior for both functions.

python/nano/test/openvino/pytorch/test_openvino.py
zhentaocc: Yes, `compile` should be a function only for training after we refactored all other modules.   
`trainer.quantize` `trainer.trace` should work on any instance of nn.Module including LightningModule.

zhentaocc: As discussed, `trainer.compile(model, loss, ..., openvino=True, onnx=False, quanti=False)` will finally become `trainer.compile(model, loss, ...)`. Seems that `compile` is then only for converting a pytorch model into PL for training then.

python/nano/test/openvino/pytorch/test_openvino.py
jason-dai: We should switch to new OpenVINO APIs in this PR.

python/nano/src/bigdl/nano/deps/inference/model.py
TheaperDeng: This is simply an overwrite of the original `model.train`, so they should have the same signature.
ref: https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.train

python/nano/src/bigdl/nano/deps/openvino/openvino_api.py
```diff
@@ -27,3 +27,8 @@ def bind_openvino_methods(pl_model):
 def export(model, input_sample=None, xml_path="model.xml"):
     from . import torch_funcs
     torch_funcs.export(model, input_sample, xml_path)
```
TheaperDeng, line 29: should we remove this `export` and `bind_openvino_methods`? Since the usage of `Trainer.compile(openvino=True)` is removed anyway.

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
```diff
@@ -32,10 +32,9 @@
 from bigdl.nano.pytorch.plugins.ddp_spawn import DDPSpawnPlugin
 from bigdl.nano.deps.ray.ray_api import distributed_ray
 from bigdl.nano.deps.ipex.ipex_api import create_IPEXAccelerator, ipex_device
-from bigdl.nano.deps.openvino.openvino_api import bind_openvino_methods
+from bigdl.nano.deps.openvino.openvino_api import bind_openvino_methods, PytorchOpenVINOModel
```
TheaperDeng, line 35: no need to import `bind_openvino_methods`.

jason-dai:
1) Remove code for old APIs (such as `base_openvino.py`, `PytorchOpenVINOInference.py`)
2) `AcceleratedLightningModule.export` should be abstract, and the actual implementation should go to `PytorchOpenVINOModel` (and `PytorchONNXModel` later)

python/nano/src/bigdl/nano/deps/openvino/openvino_model.py
```diff
@@ -0,0 +1,51 @@
...
+        else:
+            assert self.ie_network, "self.ie_network shouldn't be None."
+            self.ie_network.batch_size = input_batch_size
+            self.exec_model = IECore().load_network(network=self.ie_network, device_name='CPU')
+
+    def read_network(self, model: str):
+        self.ie_network = IECore().read_network(model=model)
+
+
+def convert_onnx_to_xml(onnx_file_path, xml_path, batch_size=1):
```
jason-dai, line 47: why put it here? make it a part of `OpenVINOModel` or put it in some utility file

python/nano/src/bigdl/nano/deps/openvino/pytorch/utility.py
```diff
@@ -0,0 +1,30 @@
...
+#
+
+from functools import partial
+from ..core.utility import convert_onnx_to_xml
+from ...inference.pytorch.utility import export as export_to_onnx
+
+
+def export(model, input_sample=None, xml_path="model.xml"):
+    '''
+    Internal function to build a ortsess and bind to the lightningmodule.
```
jason-dai, line 24: This python doc is wrong

zhentaocc: fixed

zhentaocc:
Please check updated description.

zhentaocc:
@jason-dai @TheaperDeng @yangw1234  any further comment?

python/nano/src/bigdl/nano/deps/inference/pytorch/model_utils.py
```diff
@@ -0,0 +1,81 @@
...
+                input_sample = tuple(next(iter(model.predict_dataloader())))
+        else:
+            raise RuntimeError("You must specify an input_sample or call `Trainer.fit` "
+                               "on the model first to use `eval_openvino`")
+
+    model.example_input_array = input_sample
+    return input_sample
+
+
+def export(model, input_sample=None, onnx_path="model.onnx", dynamic_axes=True):
```
jason-dai, line 56: `export_onnx`?

python/nano/src/bigdl/nano/deps/inference/pytorch/model_utils.py
```diff
@@ -0,0 +1,81 @@
...
+    '''
+    Internal function to build a ortsess and bind to the lightningmodule.
```
jason-dai, line 58: Is this comment correct?

python/nano/src/bigdl/nano/deps/openvino/core/openvino_model.py
```diff
@@ -40,17 +36,13 @@ def compile_executable(self, inputs):
...
 
-    def reset(self, model):
-        self.ie_network = None
-        self.exec_model = None
-        model.ov_infer_engine = None
-
     def read_network(self, model: str):
         self.ie_network = IECore().read_network(model=model)
 
+    def save(self, path):
```
jason-dai, line 42: does the path need to be an xml file?

zhentaocc: yes

jason-dai: do we need to check it?

zhentaocc: added check.

jason-dai:
Please add more python doc (especially for non-internal APIs); otherwise LGTM

python/nano/src/bigdl/nano/deps/inference/pytorch/model_utils.py
```diff
@@ -0,0 +1,81 @@
...
+    model.example_input_array = input_sample
+    return input_sample
+
+
+def export_to_onnx(model, input_sample=None, onnx_path="model.onnx", dynamic_axes=True):
+    '''
+    Internal function to export pytorch model as onnx.
+
+    :param input_sample: torch.Tensor or a list for the model tracing.
+    :param file_path: The path to save onnx model file.
```
TheaperDeng, line 61: `:param dynamic_axes: If we set the first dim of each input as a dynamic batch_size`

python/nano/src/bigdl/nano/deps/inference/pytorch/model_utils.py
```diff
@@ -0,0 +1,81 @@
...
+    :param **kwargs: will be passed to torch.onnx.export function.
```
TheaperDeng, line 62: We should actually add this to the function signature, since it is important (e.g. opset_version is important)

python/nano/src/bigdl/nano/deps/inference/pytorch/model_utils.py
```diff
@@ -0,0 +1,81 @@
...
+            dynamic_axes[arg] = {0: 'batch_size'}  # set all dim0 to be dynamic
+    else:
+        dynamic_axes = {}
+    default_onnx_export_args = {'export_params': True,
+                                'opset_version': 11,  # version = 11 by default
+                                'do_constant_folding': True,
+                                'input_names': forward_args,
+                                'dynamic_axes': dynamic_axes,
+                                }
+
```
TheaperDeng, line 80: we should update the `default_onnx_export_args` using the kwargs provided.

python/nano/src/bigdl/nano/deps/openvino/pytorch/pytorch_openvino_model.py
```diff
@@ -0,0 +1,64 @@
...
+            )
+        inputs = self.tensors_to_numpy(inputs)
+        return inputs
+
+    def on_forward_end(self, outputs):
+        outputs = self.numpy_to_tensors(outputs.values())
+        return outputs
+
+    @staticmethod
+    def load(path):
```
TheaperDeng, line 56: Why we still need this function, since there is an exact same method in `OpenVINOModel`

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
```diff
@@ -318,3 +313,67 @@ def quantize(self, pl_model,  # remove the type requirement for type checking
                     quantized_onnx_model)
         else:
             raise NotImplementedError("Backend {} is not implemented.".format(backend))
+
+    @staticmethod
+    def trace(model: nn.Module, input_sample=None, accelerator=None):
+        """
+        Trace a pytorch model and convert it into an accelerated module for inference.
+        For example, this function returns a PytorchOpenVINOModel when accelerator=='opevino'.
```
TheaperDeng, line 321: openvino

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
```diff
@@ -318,3 +313,67 @@ def quantize(self, pl_model,  # remove the type requirement for type checking
...
+
+        :param model: An torch.nn.Module model, including pl.LightningModule.
+        :param input_sample: A set of inputs for trace, defaults to None if you have trace before or
+                             model is a LightningModule with any dataloader attached.
+        :param accelerator: The accelerator to use, defaults to None meaning staying in Pytorch
+                            backend. Only 'openvino' is supported for now.
```
TheaperDeng, line 327: I wonder if we need to provide a `trace_args` for our users to tune the onnx tracing setting. This may enlarge our availability.

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
```diff
@@ -318,3 +313,67 @@ def quantize(self, pl_model,  # remove the type requirement for type checking
...
+        :return: A PytorchOpenVINOModel using openvino for inference if accelerator='openvino'
+        """
+        if accelerator == 'openvino':
+            return PytorchOpenVINOModel(model, input_sample)
+
+    @staticmethod
+    def save(model, path, precision=None, accelerator=None, input_sample=None):
```
TheaperDeng, line 334: Are we going to keep `precision` and `accelerator` here? Since `model` instance already contains these.

python/nano/src/bigdl/nano/deps/inference/pytorch/model.py
```diff
@@ -0,0 +1,65 @@
...
+
+    @staticmethod
+    def numpy_to_tensors(np_arrays):
+        tensors = tuple(map(lambda x: torch.from_numpy(x), np_arrays))
+        if len(tensors) == 1:
+            tensors = tensors[0]
+        return tensors
+
+    def save(self, path):
+        raise NotImplementedError("AcceleratedLightningModule.save should be overridden.")
```
TheaperDeng, line 65: should we add a status property as https://github.com/intel-analytics/BigDL/issues/4360#issuecomment-1091159252

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
zhentaocc: accelerator may be necessary here. I included torch2openvino conversion in `trainer.save` to directly save pytorch model as openvino model in local. So users don't have to run `trainer.trace` and `trainer.save` to export pytorch model. But still we can judge the saving format from the suffix of path.

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
zhentaocc: Let's include this for onnx later. For openvino, no further args are used for now.

python/nano/src/bigdl/nano/deps/openvino/pytorch/pytorch_openvino_model.py
zhentaocc: the return type is different. Actually let me remove that one in `OpenVINOModel`.

python/nano/src/bigdl/nano/deps/inference/pytorch/model.py
zhentaocc: Postpone this after we refreshed quantization and onnx APIs?

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
zhentaocc: We can discuss save and load function later and modify the saved content for these functions.

python/nano/src/bigdl/nano/deps/openvino/pytorch/pytorch_openvino_model.py
zhentaocc: removed load from `OpenVINOModel`.

python/nano/src/bigdl/nano/deps/inference/pytorch/model_utils.py
zhentaocc: Done

python/nano/src/bigdl/nano/deps/inference/pytorch/model_utils.py
zhentaocc: added for this function. But not for `trainer.trace`. We can add this easily for other backend easily.

python/nano/src/bigdl/nano/deps/inference/pytorch/model_utils.py
zhentaocc: added

python/nano/src/bigdl/nano/deps/inference/pytorch/model_utils.py
TheaperDeng: yeah I agree. We may skip this for now.

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
TheaperDeng: Yeah, we may keep this API for now. I think it might be a little confusing if users can have 2 ways to do the same thing.

python/nano/src/bigdl/nano/deps/inference/pytorch/model.py
```diff
@@ -0,0 +1,65 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+from bigdl.nano.pytorch.lightning import LightningModuleFromTorch
+import torch
+
+from .model_utils import get_forward_args
+
+
+class AcceleratedLightningModule(LightningModuleFromTorch):
```
yangw1234, line 22: Let's move this out of deps, since this code does not introduce new dependencies.

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
```diff
@@ -318,3 +313,67 @@ def quantize(self, pl_model,  # remove the type requirement for type checking
                     quantized_onnx_model)
         else:
             raise NotImplementedError("Backend {} is not implemented.".format(backend))
+
+    @staticmethod
+    def trace(model: nn.Module, input_sample=None, accelerator=None):
```
yangw1234, line 318: How about we rename this method to `optimize`? We are also planning to add the new ipex logic here, since in the new ipex version, the usage changes to "model = ipex.optimize(model)".

@jason-dai

yangw1234: > How about we rename this method to `optimize`? We are also planning to add the new ipex logic here, since in the new ipex version, the usage changes to "model = ipex.optimize(model)".
> 
> @jason-dai 

We think it better to first merge this pr and maybe change later if necessary.

yangw1234 (approved):

zhentaocc:
http://10.112.231.51:18888/job/ZOO-PR-NanoTests/438/

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
jason-dai: Let's merge the PR first. 

We plan to use `trace` for inference API

jason-dai:
Is this PR ready for merge?

PR #4386, opened 2022-04-11
Hotfix TF2Estimator Test

sgwhat:


sgwhat:
http://10.112.231.51:18888/job/BigDL-PRVN-ray-ctx/255

PR #4389, opened 2022-04-11
Nano: Move onnnruntime support to nano/deps/onnxruntime

TheaperDeng:


TheaperDeng:
@zhentaocc any comments? (especially those changes regarding to your previous code)

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
```diff
@@ -314,7 +314,7 @@ def quantize(self, pl_model: LightningModule,
                     bind_base_inference_rt_methods
                 from bigdl.nano.pytorch.runtime_binding.quantization_inference import \
                     bind_quantize_methods
-                from bigdl.nano.pytorch.runtime_binding.onnxrt_inference import \
+                from bigdl.nano.deps.onnxruntime.onnxruntime_api import \
```
zhentaocc, line 317: we can import this on top level only once now.

TheaperDeng: exactly, I will change it

PR #4390, opened 2022-04-11
modify epochs in jep example tests

cyita:


cyita:
http://10.112.231.51:18889/job/ZOO-PR-BigDL-Pip-JEP-Examples-Spark2.4/180//console

PR #4402, opened 2022-04-13
disallow use `assert` in python dllib

dding3:
disallow using `assert` in python dllib, will fail the style check if used.


yangw1234:
The nano unit test job has been fixed on main, you can rebase main to trigger a re-run

dding3:
most jekins pass. http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-PR-Dllib-Validation/425/ 

The aborted one is not related to this pr.

PR #4406, opened 2022-04-13
fix pytorch not compatible with setuptools

yangw1234:


TheaperDeng:
should we add it to setup to limit `setuptools==58.0.4`? Since some users have feedback on this issue as well.

yangw1234:
> should we add it to setup to limit `setuptools==58.0.4`? Since some users have feedback on this issue as well.

Maybe we can try it, but I think setuptools is normally installed by conda, I am not sure we can override it in pip.

PR #4415, opened 2022-04-14
fix orca jekins regression

dding3:
Fix https://github.com/intel-analytics/arda-docker/issues/617#issuecomment-1097744728


dding3:
most jekins pass. http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-Orca-PR-Validation/575/. The failed one and aborted one is not related to this change.

PR #4416, opened 2022-04-14
Nano: Change setup and test script reorg for onnx/onnxruntime

TheaperDeng:
I will only add a new test env(job) for onnx/pytorch with inc rather than 2 for onnx-only and onnx+inc.


TheaperDeng:
http://10.112.231.51:18888/job/ZOO-PR-nano-tensorflow-test/
http://10.112.231.51:18888/job/BigDL-NanoTests-PR-Validation/
http://10.112.231.51:18888/job/ZOO-PR-nano-pytorch-onnxruntime/

PR #4418, opened 2022-04-14
[Nano] INC new api and usage

zhentaocc:
Optimize INC usage. 
### Current
```python
trainer.quantize(model, ...)
model.eval(quantize=True)
y=model(x)
```
### Optmized usage
```python
q_model = trainer.quantize(model, ..., return_pl=False)
y = q_model(x)
```
### Following tasks
1. remove support of argument `framework` being a list. 
2. Do the same when framework == `onnx`
3. remove `return_pl`. No need to bind to original pl model any more. 

Will not apply new `trainer.quantize` unless we finished above tasks since it's coupled with chronos.


zhentaocc:
Blocked by #4381 .

python/nano/src/bigdl/nano/deps/neural_compressor/pytorch/quantized_model.py
```diff
@@ -0,0 +1,51 @@
...
+#
+import os
+
+import yaml
+from bigdl.nano.utils.inference.pytorch.model import AcceleratedLightningModule
+from neural_compressor.utils.pytorch import load
+from neural_compressor.model.model import PyTorchModel
+
+
+class QuantizedModel(AcceleratedLightningModule):
```
jason-dai, line 24: PyTorchQuantizedModel

python/nano/test/neural-compressor/pytorch/test_trainer_quantize.py
```diff
@@ -72,7 +72,7 @@ def test_trainer_quantize_inc_ptq_compiled(self):
         x = next(train_loader_iter)[0]
 
         # Case 1: Default
-        qmodel = trainer.quantize(pl_model, self.train_loader)
+        qmodel = trainer.quantize(pl_model, self.train_loader, return_pl=False)
```
jason-dai, line 75: why always `return_pl=False`?

python/nano/src/bigdl/nano/deps/neural_compressor/pytorch/quantized_model.py
```diff
@@ -0,0 +1,51 @@
...
+    def __init__(self, model):
+        super().__init__(model.model)
+        self.quantized = model
+
+    def save(self, path):
+        return self.quantized.save(path)
+
+    def load_state_dict(self, state_dict):
+        load(state_dict, self.quantized._model)
```
jason-dai, line 33: `quantized._model` or `quantized.model`?

python/nano/src/bigdl/nano/deps/neural_compressor/pytorch/quantized_model.py
```diff
@@ -0,0 +1,51 @@
...
+    def state_dict(self):
+        try:
+            stat_dict = self.quantized.model.state_dict()
+            stat_dict['best_configure'] = self.quantized.tune_cfg
+        except IOError as e:
+            raise IOError("Fail to dump configure and weights due to {}.".format(e))
+        return stat_dict
+
+    @staticmethod
+    def load(path, model):
```
jason-dai, line 44: it should be `load(path)` that matches `save(path)`

jason-dai:
(1) The new `quantize` API should be something like:
```python
model=trainer.quantize(model, precision="INT8", accelerator="onnx", ...)
```

(2) `quantize` argument in `Trainer.compile` is no longer needed?

(3) Do we still need `return_pl` in `Trainer.quantize`?

python/nano/src/bigdl/nano/deps/neural_compressor/pytorch/quantized_model.py
zhentaocc: For inc, we need a fp32 model to load quantized model. This calls `load` provided by inc.
https://github.com/intel/neural-compressor/blob/v1.8.1/neural_compressor/utils/pytorch.py#L68-L81

jason-dai: We should handle this automatically, instead of having the user to handle it.

TheaperDeng:
> (1) The new `quantize` API should be something like:
> 
> ```python
> model=trainer.quantize(model, precision="INT8", accelerator="onnx", ...)
> ```
> 
> (2) `quantize` argument in `Trainer.compile` is no longer needed?
> 
> (3) Do we still need `return_pl` in `Trainer.quantize`?

For (3)
`return_pl` is no longer needed. Still, it will removed by me later since there are many src in Chronos use this parameter right now.

python/nano/src/bigdl/nano/deps/neural_compressor/pytorch/quantized_model.py
zhentaocc: 1. same thing, copied from inc src code, will use `quantized.model`
2. will add a test to confirm that.

zhentaocc:
> (1) The new `quantize` API should be something like:
> 
> ```python
> model=trainer.quantize(model, precision="INT8", accelerator="onnx", ...)
> ```
I will add `accelerator`. For `precision`, I don't know if we need it. Usually we quantize a model to "INT8" or "UINT8", but seems that we don't have to specify this. If you intend to have extra option for "FP16" or "BF16". I think we should put it in `trainer.trace` for that this actually is not doing quantization. It just casting data type to 16 bits intead from my understanding. 
> (2) `quantize` argument in `Trainer.compile` is no longer needed?
Yes, will try to remove this. 
> (3) Do we still need `return_pl` in `Trainer.quantize`?
Junwei and I are collaborating on this to remove these arguments with several PRs step by step to avoid long list of changes in on PR.

python/nano/src/bigdl/nano/deps/neural_compressor/pytorch/quantized_model.py
zhentaocc: Yes, my intention for this function was to help uses to quickly deploy saved model on another machine.  
For openvino, users can simply do `PytorchOpenVINOModel.load(path)` to create a model for inference.  
Different from openvino, pytorch saves weights only. So in traditional usage, we usually first construction the model graph and then load weights.  
I need to investigate a bit more about this saving function.

jason-dai:
> > (1) The new `quantize` API should be something like:
> > ```python
> > model=trainer.quantize(model, precision="INT8", accelerator="onnx", ...)
> > ```
> 
> I will add `accelerator`. For `precision`, I don't know if we need it. Usually we quantize a model to "INT8" or "UINT8", but seems that we don't have to specify this. If you intend to have extra option for "FP16" or "BF16". I think we should put it in `trainer.trace` for that this actually is not doing quantization. It just casting data type to 16 bits intead from my understanding.

Conceptually "FP16" or "BF16" is another type of quantization (e.g., see https://www.tensorflow.org/lite/performance/post_training_float16_quant) and should be supported here for inference. Please follow the design in https://github.com/intel-analytics/BigDL/issues/4360

> 
> > (2) `quantize` argument in `Trainer.compile` is no longer needed?
> 
> Yes, will try to remove this.
> 
> > (3) Do we still need `return_pl` in `Trainer.quantize`?
> 
> Junwei and I are collaborating on this to remove these arguments with several PRs step by step to avoid long list of changes in on PR.

zhentaocc:
> > > (1) The new `quantize` API should be something like:
> > > ```python
> > > model=trainer.quantize(model, precision="INT8", accelerator="onnx", ...)
> > > ```
> > 
> > 
> > I will add `accelerator`. For `precision`, I don't know if we need it. Usually we quantize a model to "INT8" or "UINT8", but seems that we don't have to specify this. If you intend to have extra option for "FP16" or "BF16". I think we should put it in `trainer.trace` for that this actually is not doing quantization. It just casting data type to 16 bits intead from my understanding.
> 
> Conceptually "FP16" or "BF16" is another type of quantization (e.g., see https://www.tensorflow.org/lite/performance/post_training_float16_quant) and should be supported here for inference. Please follow the design in #4360
> 
> > > (2) `quantize` argument in `Trainer.compile` is no longer needed?
> > 
> > 
> > Yes, will try to remove this.
> > > (3) Do we still need `return_pl` in `Trainer.quantize`?
> > 
> > 
> > Junwei and I are collaborating on this to remove these arguments with several PRs step by step to avoid long list of changes in on PR.
OK.

python/nano/src/bigdl/nano/deps/neural_compressor/pytorch/quantized_model.py
jason-dai: How doe the save-load workflow look like in these cases?

- For ONNX or OpenVINO???
  ```
  qmodel = trainer.quantize(model, accelerator="ONNXRT", ...)
  qmode.save(path)
  qmodel = model.load(path)
  ```

 - For PyTorch???
   ```
   qmodel = trainer.quantize(model, accelerator="ONNXRT", ...)
   qmode.save(path)
   ???
   ```

python/nano/src/bigdl/nano/deps/neural_compressor/pytorch/quantized_model.py
zhentaocc: Fix this in new commits.

python/nano/src/bigdl/nano/deps/neural_compressor/pytorch/quantized_model.py
```diff
@@ -0,0 +1,45 @@
...
+class PytorchQuantizedModel(AcceleratedLightningModule):
+    def __init__(self, model):
+        super().__init__(model.model)
+        self.quantized = model
+
+    def load_state_dict(self, state_dict):
+        self.quantized.model.load_state_dict(state_dict)
+
+    def state_dict(self):
+        return self.quantized.model.state_dict()
```
TheaperDeng, line 32: Why we need these 2 methods(`load_state_dict`, `state_dict`)?

python/nano/test/neural-compressor/pytorch/test_trainer_quantize.py
```diff
@@ -103,17 +104,35 @@ def test_trainer_quantize_inc_ptq_compiled(self):
...
 
-        qmodel = trainer.quantize(self.user_defined_pl_model, self.train_loader)
+        qmodel = trainer.quantize(self.user_defined_pl_model, self.train_loader, return_pl=False)
         assert qmodel
-        out = qmodel(next(train_loader_iter)[0])
+        out = qmodel(x)
+        assert out.shape == torch.Size([256, 10])
+
+        # save and load
+        trainer.save(qmodel, 'saved_int8')
```
TheaperDeng, line 134: maybe use some `tempfile` here?

python/nano/src/bigdl/nano/deps/neural_compressor/pytorch/quantized_model.py
```diff
@@ -0,0 +1,45 @@
...
+#
+from pathlib import Path
+import yaml
+from bigdl.nano.utils.inference.pytorch.model import AcceleratedLightningModule
+from neural_compressor.utils.pytorch import load
+from neural_compressor.model.model import PyTorchModel
+
+
+class PytorchQuantizedModel(AcceleratedLightningModule):
+    def __init__(self, model):
```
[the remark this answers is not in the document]
zhentaocc: `model.model` is a `nn.Module` and any subclass.

python/nano/src/bigdl/nano/deps/neural_compressor/pytorch/quantized_model.py
zhentaocc: If a pytorch user happen to use these two function, then they will get error.. A `PytorchQuantizedModel` in fact is a `nn.Module` too.

python/nano/test/neural-compressor/pytorch/test_trainer_quantize.py
zhentaocc: OK

python/nano/test/neural-compressor/pytorch/test_trainer_quantize.py
```diff
@@ -72,7 +73,7 @@ def test_trainer_quantize_inc_ptq_compiled(self):
         x = next(train_loader_iter)[0]
 
         # Case 1: Default
-        qmodel = trainer.quantize(pl_model, self.train_loader)
+        qmodel = trainer.quantize(pl_model, self.train_loader, return_pl=False)
```
jason-dai, line 76: The `Trainer.quantize` API is not updated according to the new design in https://github.com/intel-analytics/BigDL/issues/4360?

zhentaocc: Not yet. After this PR, we will have a separated PR to resolve it. All the usage in Chronos will be changed.

zhentaocc:
http://10.112.231.51:18888/job/BigDL-NanoTests-PR-Validation/51/

PR #4419, opened 2022-04-14
[Orca] Ray Runtime Example Test Speed Up

sgwhat:
Speed up for cifar10, set `epochs` to 1 and `batch_size` to 256.


sgwhat:
Jenkins Passed: http://10.112.231.51:18888/job/BigDL-PRVN-ray-ctx-example/18/

Time cost decreasing to 28min.

PR #4424, opened 2022-04-15
Fix Orca Examples Jenkins Test Stuck

sgwhat:
https://github.com/intel-analytics/arda-docker/issues/620


sgwhat:
Not caused by example code, closed.

PR #4444, opened 2022-04-19
Add more datatypes in convert DataFrame to numpy

hkvision:
Currently many datatypes are not correctly converted, and may cause error in subsequent model forward.

### DIFF
--- a/python/dllib/src/bigdl/dllib/utils/utils.py
+++ b/python/dllib/src/bigdl/dllib/utils/utils.py
@@ -23,6 +23,7 @@ import os
 import re
 from bigdl.dllib.utils.log4Error import *
 
+
 if sys.version >= '3':
     long = int
     unicode = str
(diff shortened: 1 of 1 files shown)


hkvision:
failed http://10.112.231.51:18888/job/BigDL-Orca-PR-Validation/590/

yushan111 (approved):
LGTM

hkvision:
In our unit tests, many of the code looks like this:
```
df = rdd.map(lambda x: (np.random.randn(50).astype(np.float).tolist(),
                                [int(np.random.randint(0, 2, size=()))])
                     ).toDF(["feature", "label"])
```
Since python only have float (default is float64), int (default is int64), when converting to DataFrame, the schema will be array<double> and array<bigint>. Then in this case the model will throw error `RuntimeError: expected scalar type Float but found Double`

A few unit tests for example `test.bigdl.orca.learn.ray.pytorch.test_estimator_pytorch_backend.TestPyTorchEstimator testMethod=test_data_parallel_sgd_correctness` throws the opposite error `RuntimeError: expected scalar type Double but found Float`, which is really weird, probably a bug of error report in PyTorch...

@yushan111 Any comments on this?

hkvision:
Comment out the data conversion for array<float> and jenkins passed: http://10.112.231.51:18888/job/BigDL-Orca-PR-Validation/599/

PR #4446, opened 2022-04-19
[Nano] unit test bug: add default acc criterion

zhentaocc:
INC default value for `accuracy_criterion` is `None` which will use internal inc value as ` {'relative': 0.1, 'higher_is_better': True}`.  
This caused error sometimes if uses just want to run without tuning, and the first quantization doesn't satisfy the acc drop of 0.1.  
The default value should be set as maximum 0.99 to make sure when `accuracy_criterion` is not specified, users can still get the correct results.  

Current ut for INC sometimes fails for this reason.


yangw1234:
LGTM

zhentaocc:
http://10.112.231.51:18888/job/ZOO-PR-NanoTests/435/

PR #4458, opened 2022-04-20
[Nano HPO] add hpo space configuration and tests

shane-huang:
add the search space configuration for hyper parameter search.


shane-huang:
jenkins passed.

PR #4465, opened 2022-04-21
Add dienl example on yarn

LouieBHLu:


python/friesian/example/dien/pytorch/yarn_dien_orca.py
```diff
@@ -0,0 +1,213 @@
...
+    driver_memory = "36g"
+    conf = {"spark.executor.memoryOverhead": "130g",
+            "spark.network.timeout": "10000000",
+            "spark.sql.broadcastTimeout": "7200",
+            "spark.sql.shuffle.partitions": "2000",
+            "spark.locality.wait": "0s",
+            "spark.sql.crossJoin.enabled": "true",
+            "spark.task.cpus": "1",
+            "spark.executor.heartbeatInterval": "200s",
+            "spark.driver.maxResultSize": "40G",
```
hkvision, line 102: refer to other examples for this, don't directly copy from notebook version.
OrcaContext.log_output = True  is needed for jupyter only.

PR #4486, opened 2022-04-24
[Nano HPO] global hpo configuration, enable/diable tf hpo

shane-huang:
* add global hpo configuration
* use global hpo configuration to disable/enable hpo for tf and pytorch respectively
* when tf hpo is enabled, decorated layers (e.g. tf.keras.layers), activations(tf.keras.activations), tf functions(i.e. tf.cast), tf.keras functions(i.e. tf.keras.Input) are added to corresonponding nano.tf modules.

### DIFF
--- a/python/nano/src/bigdl/nano/automl/hpo/config.py
+++ b/python/nano/src/bigdl/nano/automl/hpo/config.py
@@ -1,0 +1,191 @@
+#
+# Copyright 2016 The BigDL Authors.
+#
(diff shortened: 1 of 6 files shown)


yangw1234:
hi Shane, would you mind adding a documentation check in here https://github.com/intel-analytics/BigDL/blob/main/python/nano/test/run-nano-codestyle-test.sh#L11.

PR #4494, opened 2022-04-25
Fix Friesian style check for examples

hkvision:
Previously only the files under `src` are checked.


hkvision:
http://10.112.231.51:18888/job/BigDL-Friesian-PR-Validation/16/

PR #4498, opened 2022-04-25
Fix dien train data files

hkvision:
To align with the preprocess output


hkvision:
http://10.112.231.51:18888/job/BigDL-PRVN-Python-Friesian-ExampleTest-Spark-2.4-feature+tf1-pip/370/console

hkvision:
http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-Friesian-PR-Validation/17/ one failure is due to network

PR #4509, opened 2022-04-26
Fix NCF test on jenkins

hkvision:
http://10.112.231.51:18888/job/BigDL-PRVN-Python-Friesian-ExampleTest-Spark-3.1-tf2/325/console
Running Pandas dataframe to spark dataframe will raise `_pickle.PicklingError: Could not serialize object: ValueError: Cell is empty` for Spark3.

Strange issue, after debugging, it is caused by `import tensorflow` at the beginning of the file. Don't know why, but fix this by lazy import tensorflow packages.

Strange that running the following sample code works, but run the entire ncf example fails.
```
from bigdl.orca import init_orca_context, stop_orca_context
from bigdl.friesian.feature import FeatureTable
import pandas as pd
import tensorflow as tf
from bigdl.dllib.feature.dataset import movielens

sc = init_orca_context(cores=8, memory="160g", init_ray_on_spark=True)
movielens_data = movielens.get_id_ratings("/home/kai/Downloads/movielens")
pddf = pd.DataFrame(movielens_data, columns=["user", "item", "label"])
full = FeatureTable.from_pandas(pddf)\
        .apply("label", "label", lambda x: x - 1, 'int')
full.show(5)

stop_orca_context()
```

Thanks @pinggao187 for help the debugging!

### DIFF
--- a/python/friesian/dev/test/run-example-tests-tf2.6.0.sh
+++ b/python/friesian/dev/test/run-example-tests-tf2.6.0.sh
@@ -23,7 +23,6 @@ set -e
 mkdir -p result
 mkdir -p result/stats
 
-
 echo "#1 start example test for two tower train"
 #timer
 start=$(date "+%s")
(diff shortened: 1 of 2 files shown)


hkvision:
jenkins passed: http://10.112.231.51:18888/job/BigDL-Friesian-PR-Validation/30/

PR #4519, opened 2022-04-27
Add deepfm test to jenkins

hkvision:


hkvision:
http://10.112.231.51:18888/job/BigDL-PRVN-Python-Friesian-ExampleTest-Spark-3.1-tf2/335/console spark3.1 has the same error as ncf #4509

PR #4527, opened 2022-04-27
Add deepfm test to jenkins

hkvision:
http://10.112.231.51:18888/job/BigDL-PRVN-Python-Friesian-ExampleTest-Spark-3.1-tf2/335/console spark3.1 has the same error as ncf https://github.com/intel-analytics/BigDL/pull/4509

It is also caused by `import deepctr` at the beginning of the model.

In this example, since DeepFM extends BaseModel, which can't be lazily imported? Thus put model into a separate file and import it in the creator function.

Also, `from deepctr_torch.inputs import SparseFeat, DenseFeat` should import at the very beginning, otherwise seems SparseFeat and DenseFeat will have serialize problems (don't know why...) and thus to make things easy, modified to create them in model_creator.

Since the error only happens on Jenkins with Spark3, not producible on my laptop, this is just a workaround and has no impact to the original implementation.

### DIFF
--- a/python/friesian/dev/test/run-example-tests-tf2.6.0.sh
+++ b/python/friesian/dev/test/run-example-tests-tf2.6.0.sh
@@ -23,77 +23,77 @@ set -e
 mkdir -p result
 mkdir -p result/stats
 
-#echo "#1 start example test for two tower train"
-##timer
-#start=$(date "+%s")
-#if [ -d data/input_2tower ]; then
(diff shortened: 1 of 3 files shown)


hkvision:
jenkins passed: http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-Friesian-PR-Validation/42/

PR #4497, opened 2022-04-25
Nano: add `trainer.trace(accelerator="onnxruntime")`'s support

TheaperDeng:
This PR implement `trainer.trace` only, skip save/load for now waiting for #4487 .
`trainer.quantize`'s support for onnxruntime and the removal of old `eval_onnx()` API will be carried out in following PRs.

PR #4544, opened 2022-04-29
fix test_dataloader

zhentaocc:
this test should be under INC.


python/nano/setup.py
```diff
@@ -79,8 +79,9 @@ def setup_package():
                         "PyTurboJPEG",
                         "opencv-transforms",
                         "onnx",
-                        "onnxruntime"]
-    install_requires = ["intel-openmp"]
+                        "onnxruntime",
```
yangw1234, line 82: @TheaperDeng we do need onnxruntime here anymore, right?

zhentaocc:
http://10.112.231.51:18888/job/BigDL-NanoTests-PR-Validation/40/

python/nano/setup.py
zhentaocc: will remove in https://github.com/intel-analytics/BigDL/pull/4497/

PR #4551, opened 2022-05-05
drop torchvision.transforms into bigdl.nano.pytorch.vision

az2181036:


yangw1234:
maybe we also need a test to check if all transforms in torchvison have been covered. For example, a new transform is added to torchvision and when we upgrade the new torchvision, we will know because the test will fail.

az2181036:
> maybe we also need a test to check if all transforms in torchvison have been covered. For example, a new transform is added to torchvision and when we upgrade the new torchvision, we will know because the test will fail.

Done. By using `__all__` function.

python/nano/src/bigdl/nano/pytorch/vision/transforms/transforms.py
```diff
@@ -199,13 +266,600 @@ def __call__(self, img):
...
+
+    def __repr__(self):
+        return self.__class__.__name__ + '(p={})'.format(self.p)
+
+
+class RandomPerspective(tv_t.RandomPerspective):
+
+    def __call__(self, img):
+        if type(img) == np.ndarray:
+            raise NotImplementedError("Input image must be PIL image or Tensor image for {}"
```
yangw1234, line 782: I think maybe we can first transfer the `np.ndarray` to a PIL image.

az2181036:
http://10.112.231.51:18888/job/BigDL-NanoTests-PR-Validation/93/

yangw1234:
LGTM

PR #4556, opened 2022-05-05
replace `raise` in python dllib with log4error api

dding3:
This PR is used to replace `raise` in python dllib with log4error api, also add rules to disallow use `raise.`


dding3:
jekins pass. http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-PR-Dllib-Validation/451/

PR #4557, opened 2022-05-05
add error handling api guidance

dding3:


docs/docs/ProgrammingGuide/error-log-api.md
```diff
@@ -0,0 +1,95 @@
+BigDL provides error handling api. Please don't use `assert`, `raise`, `throw` to fail the application.
+Use the error handling api instead, it will provide useful message for debugging.
+
+## **Error handling API**
+
+**Scala**
+
```
jason-dai, line 7: what package to import?

docs/docs/ProgrammingGuide/error-log-api.md
```diff
@@ -0,0 +1,95 @@
...
+```
+
+* `condition`: Will throw exception and print errmsg if it's `true`. Otherwise will pass the function
+* `errmsg`: Error message to be print.
+* `fixmsg`: Message about how to fix the error.
+* `cause`: Exception need to throw.
+
+Notes: This API is for future extension, in case we need distinct invalidOperation exception with unKnownException in python
+
+**Python**
```
jason-dai, line 39: what package to import?

jason-dai:
Move to https://github.com/intel-analytics/BigDL/tree/main/docs/readthedocs/source/doc

docs/readthedocs/source/doc/error-log-api.md
```diff
@@ -0,0 +1,109 @@
...
+```python
+from bigdl.dllib.utils.log4Error import *
+```
+
+If user's input is invalid
+```python
+invalidInputError(condition, errMsg, fixMsg=None)
+```
+
+* `condition`: Will throw exception and print errmsg if it's `true`. Otherwise will pass the function
```
cyita, line 57: Actually, the behavior is "Will throw exception and print errmsg if it's `false`"
```
def invalidInputError(condition, errMsg, fixMsg=None):
    if not condition:
        outputUserMessage(errMsg, fixMsg)
        raise RuntimeError(errMsg)
```

dding3: Thanks for pointing out. Fixed

TheaperDeng:
Are we going to follow this API for all components in bigdl? since some components may not rely on bigdl-dllib

PR #4567, opened 2022-05-07
[Nano] Transition to openvino api 2.0

zhentaocc:
Transition to OpenVINO API 2.0
https://docs.openvino.ai/latest/openvino_2_0_transition_guide.html

Use Model Optimizer for model conversion from ONNX to OpenVINO.

### DIFF
--- a/python/nano/src/bigdl/nano/deps/openvino/core/openvino_model.py
+++ b/python/nano/src/bigdl/nano/deps/openvino/core/openvino_model.py
@@ -24,12 +24,7 @@ class OpenVINOModel:
         self.read_network(ie_network)
 
     def forward_step(self, *inputs):
-        try:
-            return self.infer_request.infer(list(inputs))
-        except RuntimeError:
-            for in_tensor, input in zip(self.infer_request.inputs, inputs):
-                in_tensor.shape == input.shape
-            return self.infer_request.infer(list(inputs))
+        return self.infer_request.infer(list(inputs))
 
     def read_network(self, model: str):
         core = Core()
--- a/python/nano/src/bigdl/nano/deps/openvino/core/openvino_model.py
+++ b/python/nano/src/bigdl/nano/deps/openvino/core/openvino_model.py
@@ -14,31 +14,28 @@
 # limitations under the License.
 #
 from pathlib import Path
-from openvino.inference_engine import IECore
+from openvino.runtime import Core
+from openvino.runtime.passes import Manager
 
 
 class OpenVINOModel:
     def __init__(self, ie_network: str):
         self.ie_network = None
         self.read_network(ie_network)
-        self.exec_model = None
 
     def forward_step(self, *inputs):
-        self.compile_executable(inputs)
-        inputs = dict(zip(self.exec_model.input_info, inputs))
-        return self.exec_model.infer(inputs)
-
-    def compile_executable(self, inputs):
-        input_batch_size = inputs[0].shape[0]
-        if self.exec_model and input_batch_size == self.exec_model.input_info[0].shape[0]:
-            return
-        else:
-            assert self.ie_network, "self.ie_network shouldn't be None."
-            self.ie_network.batch_size = input_batch_size
-            self.exec_model = IECore().load_network(network=self.ie_network, device_name='CPU')
+        try:
+            return self.infer_request.infer(list(inputs))
+        except RuntimeError:
+            for in_tensor, input in zip(self.infer_request.inputs, inputs):
+                in_tensor.shape == input.shape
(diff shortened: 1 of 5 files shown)


zhentaocc:
http://10.112.231.51:18888/job/BigDL-NanoTests-PR-Validation/63/

python/nano/src/bigdl/nano/deps/openvino/core/openvino_model.py
```diff
@@ -14,31 +14,28 @@
...
-        else:
-            assert self.ie_network, "self.ie_network shouldn't be None."
-            self.ie_network.batch_size = input_batch_size
-            self.exec_model = IECore().load_network(network=self.ie_network, device_name='CPU')
+        try:
+            return self.infer_request.infer(list(inputs))
+        except RuntimeError:
+            for in_tensor, input in zip(self.infer_request.inputs, inputs):
+                in_tensor.shape == input.shape
+            return self.infer_request.infer(list(inputs))
```
yangw1234, line 32: How about leaving some comments explaining the logic here?

It seems like you are catching the `RuntimeError` to recover from the shape mismatch problem. Are there other types of RuntimeError? Maybe you can also check the error message to get an exact match.

python/nano/src/bigdl/nano/deps/openvino/pytorch/pytorch_openvino_utils.py
```diff
@@ -28,7 +28,7 @@ def export(model, input_sample=None, xml_path="model.xml"):
     :param input_sample: torch.Tensor or a list for the model tracing.
     :param xml_path: The path to save openvino model file.
     '''
-    export_to_onnx(model, input_sample, 'tmp.onnx', dynamic_axes=False)
+    export_to_onnx(model, input_sample, 'tmp.onnx', dynamic_axes=True)
```
yangw1234, line 31: Is there any reason or implications for changing the `dynamic_axes` value? Maybe leaving some comments?

python/nano/src/bigdl/nano/deps/openvino/core/openvino_model.py
zhentaocc: > It seems like you are catching the `RuntimeError` to recover from the shape mismatch problem. Are there other types of RuntimeError? Maybe you can also check the error message to get an exact match.
>

Seems that with `dynamic_axes=True`, we don't have to worry about the shape now. I will remove this try-except.
>
> Also is it safe to change the shape like this? What if the user just passes the wrong input?

No need to change now. If the input is wrong, openvino will raise `RuntimeError: Can't SetBlob with name: x, because model input (shape={?,3,256,256}) and blob (shape=(3.3.512.512)) are incompatible`. I think it's quite explicit for users. Do you think we need to catch and raise another error?

python/nano/src/bigdl/nano/deps/openvino/pytorch/pytorch_openvino_utils.py
zhentaocc: Without using ModelOptimizer to convert onnx to openvino as before, `dynamic_axes=True` is not acceptable. Now I changed the way to convert onnx to openvino by ModelOptimizer, it's more powerful and convenient for inference with different batches or resolutions.

python/nano/src/bigdl/nano/deps/openvino/core/openvino_model.py
yangw1234: > > It seems like you are catching the `RuntimeError` to recover from the shape mismatch problem. Are there other types of RuntimeError? Maybe you can also check the error message to get an exact match.
> 
> Seems that with `dynamic_axes=True`, we don't have to worry about the shape now. I will remove this try-except.
> 
> > Also is it safe to change the shape like this? What if the user just passes the wrong input?
> 
> No need to change now. If the input is wrong, openvino will raise `RuntimeError: Can't SetBlob with name: x, because model input (shape={?,3,256,256}) and blob (shape=(3.3.512.512)) are incompatible`. I think it's quite explicit for users. Do you think we need to catch and raise another error?

I think maybe we can just remove this try-except.

python/nano/src/bigdl/nano/deps/openvino/pytorch/pytorch_openvino_utils.py
yangw1234: ok

yangw1234:
LGTM

zhentaocc:
http://10.112.231.51:18888/job/BigDL-NanoTests-PR-Validation/67

PR #4568, opened 2022-05-09
Chronos: add a new model Autoformer

TheaperDeng:
This PR add a new model Autoformer(https://arxiv.org/abs/2106.13008) to Chronos built-in model.

Detailed description will be listed and here.


TheaperDeng:
http://10.112.231.51:18888/job/BigDL-Chronos-PR-Validation/756/

PR #4573, opened 2022-05-09
add invalidOperationError in python log4error

dding3:
This PR adds invalidOperationError api to make the python log4error as close as scala


dding3:
jekins pass. http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-PR-Dllib-Validation/456/

PR #4579, opened 2022-05-10
add ut for ppml context

qiuxin2012:


jason-dai:
Also need to ensure `PPMLContext.read` returns the same results as `SparkSession.read`, and `PPMLContext.read` can correctly read the result of `PPMLContext.write`

PR #4594, opened 2022-05-12
[PPML] Add DummyAttestation unit test

Forest216:


scala/ppml/src/test/scala/com/intel/analytics/bigdl/ppml/attestation/DummyAttestationServiceSpec.scala
```diff
@@ -0,0 +1,17 @@
+package com.intel.analytics.bigdl.ppml.attestation
+
+import org.scalatest.{FlatSpec, Matchers}
+
+class DummyAttestationServiceSpec extends FlatSpec with Matchers {
+    "Dummy Attestation " should "work" in {
+        val dummyAttestationService = new DummyAttestationService()
+        val logger = dummyAttestationService.logger
+        val quote = dummyAttestationService.getQuoteFromServer()
```
qiyuangong, line 9: Need to split getQuote and attestWith Server

scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/attestation/DummyAttestationService.scala
```diff
@@ -46,15 +46,10 @@ class DummyAttestationService extends AttestationService {
             }
             val nonce: String = "test"
             val response: JSONObject = new JSONObject()
-            val number = new BigInteger(quote)
-            val zero = new BigInteger("0")
-            var verifyQuoteResult = true
+            val verifyQuoteResult = Random.nextBoolean()
```
qiyuangong, line 49: Random true of false is not good design. I think we can get "true" from quote string, then return true. Otherwise return false

Forest216: update in https://github.com/intel-analytics/BigDL/pull/4594/files#diff-83601ee9b506b76581cdfd8d7811c08da5e7b0fde78efb91870e57e1fe4e5cb5R52

scala/ppml/src/test/scala/com/intel/analytics/bigdl/ppml/attestation/DummyAttestationServiceSpec.scala
Forest216: update in https://github.com/intel-analytics/BigDL/pull/4594/files#diff-35e71054734927b078bd0d50242dc3b43f8173e511ccb59b05e4f6e9062e684bR12

scala/ppml/src/test/scala/com/intel/analytics/bigdl/ppml/attestation/DummyAttestationServiceSpec.scala
```diff
@@ -0,0 +1,27 @@
+package com.intel.analytics.bigdl.ppml.attestation
+
+import org.apache.logging.log4j.Logger
+import org.scalatest.{FlatSpec, Matchers}
+
+class DummyAttestationServiceSpec extends FlatSpec with Matchers {
+
+    val dummyAttestationService = new DummyAttestationService()
+    val logger: Logger = dummyAttestationService.logger
+    var quote: String = ""
```
qiyuangong, line 10: Don't change var in unit test. because unit test running order may change

scala/ppml/src/test/scala/com/intel/analytics/bigdl/ppml/attestation/DummyAttestationServiceSpec.scala
```diff
@@ -0,0 +1,27 @@
...
+
+    "Get Quote " should "work" in {
+        quote = dummyAttestationService.getQuoteFromServer()
+        quote shouldNot equal ("")
+        logger.info(quote)
+    }
+
+    "Attest With Server " should "work" in {
+        var result = dummyAttestationService.attestWithServer(quote)._1
```
qiyuangong, line 19: So, this test is not necessary

qiyuangong (approved):
LGTM

PR #4596, opened 2022-05-12
Chronos/Nano: Remove redundant APIs and apply new APIs in Chronos

TheaperDeng:
- [x] Forecaster uts
- [x] TSPipeline uts
- [x] Nano uts


python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
```diff
@@ -245,8 +225,7 @@ def quantize(self, model,  # remove the type requirement for type checking
                  accuracy_criterion: dict = {'relative': 0.99, 'higher_is_better': True},
                  timeout=0,
                  max_trials=1,
-                 input_sample=None,
-                 return_pl=False
+                 input_sample=None
                  ):
```
jason-dai, line 229: I think you should move the parameters that matters to the the user most to the front of the list, such as
```python
    def quantize(self, model,  # remove the type requirement for type checking
                 precision='int8',
                 accelerator=None,
                 calib_dataloader: DataLoader = None,
                 val_dataloader: DataLoader = None,
                 metric: Optional[Metric] = None,
                 accuracy_criterion: dict = {'relative': 0.99, 'higher_is_better': True},
                 method='fx',
                 backend='inc',
                 conf: Optional[str] = None,
                 approach='static',
                 tuning_strategy='bayesian',
                 timeout=0,
                 max_trials=1,
                 input_sample=None
                 ):
```

TheaperDeng: sure, I will reorganize the order

TheaperDeng:
http://10.112.231.51:18888/job/BigDL-NanoTests-PR-Validation/96/

TheaperDeng:
@zhentaocc please also have a review, especially the nano part

zhentaocc:
LGTM

PR #4615, opened 2022-05-17
Chronos: add built-in dataset: uci_electricity_wide

Rebekk:


python/chronos/src/bigdl/chronos/data/utils/public_dataset.py
```diff
@@ -208,6 +211,26 @@ def preprocess_uci_electricity(self):
...
         return self
 
+    def preprocess_uci_electricity_wide(self):
+        '''
+        return data that meets the minimum requirements of tsdata.
+        '''
+        if not os.path.exists(self.final_file_path):
+            zip_file = zipfile.ZipFile(os.path.join(
+                                       os.path.expanduser(self.dir_path),
+                                       DATASET_NAME[self.name[:-5]][0]))
```
TheaperDeng, line 221: why we need this `[:-5]`

TheaperDeng (commented):
This dataset is quite large so it might be unreasonable to have it in our automatic test. Please manually test it.

python/chronos/src/bigdl/chronos/data/utils/public_dataset.py
Rebekk: Oh, I forgot I already add 'uci_electricity_wide' in DATASET_NAME. "[:-5]" was added for if we don't modify DATASET_NAME.

TheaperDeng (approved):
run jenkins

Rebekk:
http://10.112.231.51:18888/job/BigDL-Chronos-PR-Validation/807/

PR #4624, opened 2022-05-17
replace `assert` `raise` with new error api in nano module

dding3:
1. replace `assert` `raise` with new error api in nano module
2. disallow using `assert` `raise` in nano
3. add new error api in nano


yangw1234:
Maybe you should also add the check to https://github.com/intel-analytics/BigDL/blob/main/python/nano/test/run-nano-codestyle-test.sh#L9

dding3:
most jekins pass. http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-NanoTests-PR-Validation/124/

Failed one is not related to this pr.
ssl.SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate in certificate chain (_ssl.c:1091)

PR #4626, opened 2022-05-18
[PPML] Attestation fix code style 

qiyuangong:
* Remove require from attestation
* Add comments

### DIFF
--- a/scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/attestation/EHSMAttestationService.scala
+++ b/scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/attestation/EHSMAttestationService.scala
@@ -19,11 +19,21 @@ package com.intel.analytics.bigdl.ppml.attestation
 
 import com.intel.analytics.bigdl.ppml.utils.EHSMParams
 import com.intel.analytics.bigdl.ppml.utils.HTTPUtil.postRequest
+import org.apache.logging.log4j.LogManager
 import org.json.JSONObject
 
+/**
+ * Attestation Service provided by ehsm
+ * @param kmsServerIP ehsm IP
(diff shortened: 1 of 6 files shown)


scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/attestation/EHSMAttestationService.scala
```diff
@@ -41,13 +51,17 @@ class EHSMAttestationService(kmsServerIP: String, kmsServerPort: String,
...
 
   def getQuoteFromServer(): String = {
+    // TODO Get qutoe from ehsm
     "test"
   }
 
   override def attestWithServer(quote: String): (Boolean, String) = {
     // TODO nonce
     val nonce: String = "test"
-    require(quote != null, "quote should be specified")
```
qiuxin2012, line 50: `Log4Error.invalidInputError(` you can use this one.

qiyuangong: Addressed

scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/attestation/EHSMAttestationService.scala
```diff
@@ -59,8 +60,8 @@ class EHSMAttestationService(kmsServerIP: String, kmsServerPort: String,
     // TODO nonce
     val nonce: String = "test"
     if (quote == null) {
```
qiuxin2012, line 62: why not `Log4Error.invalidInputError(quote != null, "Quote should be specified")`?
Log4Error.invalidInputError is as the same as `require`

PR #4628, opened 2022-05-18
Change multiprocessing backend to subprocess

yangw1234:


yangw1234:
more test cases will be added in follow up PR.

yangw1234:
http://10.112.231.51:18888/job/BigDL-NanoTests-PR-Validation/108/

PR #4635, opened 2022-05-19
Add tensorflow keras notebooks

y199387:


python/nano/notebooks/tensorflow/cifar10/nano_tensorflow_fit_example.ipynb
```diff
@@ -0,0 +1,1054 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "id": "659f2f72",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "class BasicBlock(Model):\n",
```
yangw1234, line 79: I think you can just use the existing reset implementation in `tf.keras.applications` instead of implementing it your self.

PR #4638, opened 2022-05-19
Fix RayOnSpark when there is activate sparkcontext

hkvision:
Previously the rayonspark init is only in the branch where there is no activate sc...


sgwhat (approved):
LGTM.

hkvision:
http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-Orca-PR-Validation/652/ Some failures are strange and unrelated. Merge it first to catch up with today's nightly build.

PR #4643, opened 2022-05-20
add model check for keras inc quantization

zhentaocc:
`keras.Model` has two usage according to https://keras.io/api/models/model/.  
Raise error if the model for inc to quantize misses the Input layer.


zhentaocc:
http://10.112.231.51:18888/job/BigDL-NanoTests-PR-Validation/136/

PR #4651, opened 2022-05-23
Nano: add customized route support for pytorch quantization/trace

TheaperDeng:
This PR helps users who write their inference route in another method rather than `forward` in their lightning module.

```python
# quantization
int8_model = trainer.quantize(..., inference_method_name="predict")

# trace
traced_accelerated_model = trainer.trace(..., inference_method_name="predict")
```


python/nano/test/neural-compressor/pytorch/test_trainer_quantize.py
```diff
@@ -142,3 +158,26 @@ def test_trainer_quantize_inc_ptq_customized(self):
...
             out = loaded_qmodel(x)
             assert out.shape == torch.Size([256, 10])
+
+    def test_trainer_quantize_inc_ptq_customized_no_forward(self):
+        # Test if a Lightning Module not compiled by nano works
+        train_loader_iter = iter(self.train_loader)
+        x = next(train_loader_iter)[0]
+        trainer = Trainer(max_epochs=1)
+
+        qmodel = trainer.quantize(self.user_defined_pl_model_no_forward,
```
jason-dai, line 168: `Trainer.quantize`

python/nano/test/neural-compressor/pytorch/test_trainer_quantize.py
```diff
@@ -142,3 +158,26 @@ def test_trainer_quantize_inc_ptq_customized(self):
             assert loaded_qmodel
             out = loaded_qmodel(x)
             assert out.shape == torch.Size([256, 10])
+
+    def test_trainer_quantize_inc_ptq_customized_no_forward(self):
+        # Test if a Lightning Module not compiled by nano works
+        train_loader_iter = iter(self.train_loader)
+        x = next(train_loader_iter)[0]
+        trainer = Trainer(max_epochs=1)
```
jason-dai, line 166: remove `trainer`

python/nano/test/neural-compressor/pytorch/test_trainer_quantize.py
```diff
@@ -142,3 +158,26 @@ def test_trainer_quantize_inc_ptq_customized(self):
...
+        assert qmodel
+        out = qmodel(x)
+        assert out.shape == torch.Size([256, 10])
+
+        # save and load
+        with tempfile.TemporaryDirectory() as tmp_dir_name:
+            trainer.save(qmodel, tmp_dir_name)
+            loaded_qmodel = trainer.load(tmp_dir_name,
+                                         self.user_defined_pl_model_no_forward,
+                                         inference_method_name="predict")
```
jason-dai, line 180: Can we save `inference_method_name` in the meta file?

python/nano/src/bigdl/nano/utils/inference/pytorch/model_utils.py
```diff
@@ -82,3 +84,37 @@ def export_to_onnx(model, input_sample=None, onnx_path="model.onnx", dynamic_axe
...
+
+
+class FP32InferenceModelWrapped(pl.LightningModule):
+    def __init__(self, model, inference_method_name="forward"):
+        super().__init__()
+        self.model = model
+        self.inference_method_name = inference_method_name
+
+    def forward(self, *args):
+        method_to_call = getattr(self.model, self.inference_method_name)
```
jason-dai, line 96: does it work if `self.inference_method_name="forward"`?

python/nano/src/bigdl/nano/utils/inference/pytorch/model_utils.py
```diff
@@ -82,3 +84,37 @@ def export_to_onnx(model, input_sample=None, onnx_path="model.onnx", dynamic_axe
                                 }
     default_onnx_export_args.update(kwargs)
     torch.onnx.export(model, input_sample, onnx_path, **default_onnx_export_args)
+
+
+class FP32InferenceModelWrapped(pl.LightningModule):
```
jason-dai, line 89: Shall we change the `inference_method_name` method (e.g., "predict" or "inference") to call `forward`?

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
```diff
@@ -219,6 +220,7 @@ def quantize(self, model,  # remove the type requirement for type checking
                  precision='int8',
                  accelerator=None,
                  calib_dataloader: DataLoader = None,
+                 inference_method_name = "forward",
```
jason-dai, line 223: move this argument down in the list, say after `val_dataloader` or after `accuracy_criterion`

python/nano/src/bigdl/nano/utils/inference/pytorch/model_utils.py
```diff
@@ -18,12 +18,14 @@
...
 import torch
+import pytorch_lightning as pl
 from bigdl.nano.utils.log4Error import invalidInputError
 
 
 def get_forward_args(model):
     forward_args = inspect.getfullargspec(model.forward).args[1:]
-    if isinstance(model, LightningModuleFromTorch):
+    if isinstance(model, LightningModuleFromTorch)\
+         or isinstance(model, FP32InferenceModelWrapped):
```
jason-dai, line 28: to make it simple, we can have `LightningModuleFromTorch` and `FP32InferenceModelWrapped` to extend a common `nano_pytorch_wrapper` base?

python/nano/test/neural-compressor/pytorch/test_trainer_quantize.py
```diff
@@ -142,3 +158,26 @@ def test_trainer_quantize_inc_ptq_customized(self):
...
+        # Test if a Lightning Module not compiled by nano works
+        train_loader_iter = iter(self.train_loader)
+        x = next(train_loader_iter)[0]
+        trainer = Trainer(max_epochs=1)
+
+        qmodel = trainer.quantize(self.user_defined_pl_model_no_forward,
+                                  calib_dataloader=self.train_loader,
+                                  inference_method_name="predict")
+        assert qmodel
+        out = qmodel(x)
```
jason-dai, line 172: better to support `out = qmodel.predict(x)`

python/nano/src/bigdl/nano/utils/inference/pytorch/model_utils.py
jason-dai, line 89: we should also explicitly make this class inference only (e.g., overriding `train`)

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
```diff
@@ -275,6 +279,7 @@ def quantize(self, model,  # remove the type requirement for type checking
         :return:            A accelerated Pytorch-Lightning Model if quantization is sucessful.
         """
         if backend == 'inc':
+            model = fp32_inference_model_wrapper(model, inference_method_name)
```
jason-dai, line 282: what if the model is an ONNX model?

python/nano/src/bigdl/nano/utils/inference/pytorch/model_utils.py
TheaperDeng: yeah it works, still to avoid any new problems involved by this wrapper, we will not wrap users' model if users use "forward" as their inference route.

python/nano/test/neural-compressor/pytorch/test_trainer_quantize.py
TheaperDeng: yeah, might be a good idea. I will first have a try and see if their is any block there.

python/nano/src/bigdl/nano/utils/inference/pytorch/model_utils.py
TheaperDeng: We don't need to do this in `FP32InferenceModelWrapped` since it is totally an internal class (i.e. users won't know this class, there won't be any external API return this class or any class inherited from this class.)

To support `model_int8.inference_method_name()`, I will try to change some code in  `AcceleratedLightningModule`.

python/nano/test/neural-compressor/pytorch/test_trainer_quantize.py
TheaperDeng: #4668, will raise another PR for this

python/nano/test/neural-compressor/pytorch/test_trainer_quantize.py
TheaperDeng: #4668, will raise another PR for this

python/nano/test/neural-compressor/pytorch/test_trainer_quantize.py
TheaperDeng: yes, it's critical for the usage of `model_int8.inference_method_name`

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
TheaperDeng: Done

python/nano/src/bigdl/nano/utils/inference/pytorch/model_utils.py
TheaperDeng: Since `FP32InferenceModelWrapped` is only an internal class while `LightningModuleFromTorch` is an external class, I think I will keep them separate.

python/nano/src/bigdl/nano/utils/inference/pytorch/model_utils.py
TheaperDeng: Since it's only internal. I will not override `train`.

python/nano/test/neural-compressor/pytorch/test_trainer_quantize.py
jason-dai: then there is no need to have the user the specify `inference_method_name` when loading the model, yes?

python/nano/src/bigdl/nano/utils/inference/pytorch/model_utils.py
jason-dai: It doesn't matter if `FP32InferenceModelWrapped` is internal or external; the point is that if we introduce additional "model wrapper" classes later, the list can become longer; or we may need to check the condition in other places later. This can be hard to maintain.

To make it easy to maintain, we can have all these "model wrapper" classes to extend a common base class (`NanoPytorchWrapper`) that wraps a PyTorch model, and then just check `if isinstance(model, NanoPytorchWrapper)` in the code.

PR #4655, opened 2022-05-23
[WIP] Chronos: add autoformer_forecaster.py

Rebekk:


python/chronos/src/bigdl/chronos/forecaster/autoformer_forecaster.py
```diff
@@ -0,0 +1,230 @@
...
+                 moving_avg=25,
+                 d_model=512,
+                 embed='timeF',
+                 dropout=0.05,
+                 factor=3,
+                 n_head=8,
+                 d_ff=2048,
+                 activation='gelu',
+                 e_layer=2,
+                 d_layers=1):
```
TheaperDeng, line 46: I think we need parameter document for these parameters like https://github.com/intel-analytics/BigDL/blob/main/python/chronos/src/bigdl/chronos/forecaster/tcn_forecaster.py#L55-L95

You may find some explanation in https://github.com/zhouhaoyi/Informer2020#usage

Some parameters can be renamed to align with other forecasters (e.g. `seq_len` = `past_seq_len`, `pred_len` = `future_seq_len`...)

python/chronos/src/bigdl/chronos/forecaster/autoformer_forecaster.py
```diff
@@ -0,0 +1,230 @@
...
+        current_num_threads = torch.get_num_threads()
+        self.num_processes = max(1, current_num_threads//8)
+        self.use_ipex = False
+        self.onnx_available = True
+        self.quantize_available = True
+        self.use_amp = False
+
+        self.model_creator = model_creator
+        self.internal = model_creator(self.config)
+
```
TheaperDeng, line 81: it there any possibility we move these changes to base_forecaster, which means making base forecaster support a more generalized input?

PR #4657, opened 2022-05-23
Fix Nano subprocess distributed backend performance

yangw1234:


yangw1234:
http://10.112.231.51:18888/job/BigDL-NanoTests-PR-Validation/137/

PR #4663, opened 2022-05-24
replace all assert/raise with new api in friesian

dding3:
1. replace all assert/raise with new api in friesian
2. disallow use `assert`, `raise` in friesian


dding3:
jekins pass. http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-Friesian-PR-Validation/79/

PR #4676, opened 2022-05-25
replace `raise`, `assert` with new api in orca

dding3:
1. replace `raise`, `assert` with new api in orca
2. disallow use `raise`, `assert` in orca


dding3:
most jekins pass. http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-Orca-PR-Validation/695/.
The failed on should not related to this pr, as nightly jekins also has the same failure.

python/orca/src/bigdl/orca/common.py
```diff
@@ -308,24 +316,26 @@ def init_orca_context(cluster_mode=None, runtime="spark", cores=2, memory="2g",
...
+                                  "but got: %s".format(cluster_mode))
+            ray_args = {}
+            for key in ["redis_port", "password", "object_store_memory", "verbose", "env",
+                        "extra_params", "num_ray_nodes", "ray_node_cpu_cores", "include_webui",
+                        "system_config"]:
+                if key in kwargs:
+                    ray_args[key] = kwargs[key]
+            from bigdl.orca.ray import OrcaRayContext
+            ray_ctx = OrcaRayContext(runtime="spark", cores=cores, num_nodes=num_nodes,
+                                     sc=sc, **ray_args)
```
hkvision, line 331: Shouldn't remove the indent for these code blocks...
See this PR: https://github.com/intel-analytics/BigDL/pull/4638
@cyita will help revert this in a new PR

PR #4699, opened 2022-05-26
PPML VFL Pytorch change interactive layer to customizable

Litchilitchy:
user has to define custom layer in server model to decide how the input(list of tensors) to aggregate


python/ppml/src/bigdl/ppml/fl/nn/pytorch/aggregator.py
```diff
@@ -102,27 +102,38 @@ def put_client_data(self, client_id, data):
...
-        input, target = [], None        
-        for ndarray_map in self.client_data.values():
+        input, target = [], None
+        # to record the order of tensors with client ID
+        client_ids = []
+        for cid, ndarray_map in self.client_data.items():
             for k, v in ndarray_map.items():
                 if k == 'input':
                     input.append(torch.from_numpy(v))
+                    client_ids.append(cid)
```
jason-dai, line 112: this only works if the order of client inputs are not important (which is true for your current examples), which is generally not the case for an arbitrary interaction layer.

Litchilitchy: should we add a parameter in user request (specifying the order of this user in Tensor list), or use some pre-defined way like alphabetical order, or some other approaches to solve the order requirement?

jason-dai: Maybe each client can have a client ID that determines their order

Litchilitchy: An order support is added in [new commit](https://github.com/intel-analytics/BigDL/pull/4699/commits/74101b2bdbf68129d9edd328f1b8629f6d4cc802)

python/ppml/src/bigdl/ppml/fl/nn/pytorch/aggregator.py
```diff
@@ -102,27 +102,44 @@ def put_client_data(self, client_id, data):
...
     def aggregate(self):
-        input, target = [], None        
-        for ndarray_map in self.client_data.values():
+        input, target = [], None
+        # to record the order of tensors with client ID
+        for cid, ndarray_map in self.client_data.items():
             for k, v in ndarray_map.items():
                 if k == 'input':
-                    input.append(torch.from_numpy(v))
+                    input.append((cid, torch.from_numpy(v)))
```
jason-dai, line 110: how's the client ID determined?

Litchilitchy: I think I can make this be set in `ppml-conf.yaml` (if not set it would be random uuid). And the order will be default the the string comparison order. So just set `1, 2, ..` if this make sense.

jason-dai: Maybe set it in `init_fl_context`?

Litchilitchy: [new commit](https://github.com/intel-analytics/BigDL/pull/4699/commits/ecab9fe823d9e910e872cefb85064c89f33adf19) supports pass client_id in `FLContext.init_fl_context()` and changes the FLClient to singleton by default, the same as we previously implemented in Scala and Py4J version.

python/ppml/src/bigdl/ppml/fl/nn/fl_context.py
```diff
@@ -0,0 +1,33 @@
...
+from bigdl.ppml.fl.nn.fl_client import FLClient
+
+
+# NN FLContext
+# TODO: merge FGBoost
+class FLContext():
+    fl_client = None
+    _lock = threading.Lock()
+    @staticmethod
+    def init_fl_context(client_id=None):
```
jason-dai, line 27: Need to make sure `client_id` is not None for PyTorch support; maybe we can just require user to specify a valid `client_id`

Litchilitchy:
Some changes do lost during rebase... I updated them and commented in new commits.

python/ppml/src/bigdl/ppml/fl/estimator.py
```diff
@@ -30,26 +30,34 @@
...
                    optimizer_cls,
-                   optimizer_args=None,
+                   optimizer_args={},                   
                    target="localhost:8980",
                    server_model=None):
-        optimizer = optimizer_cls(client_model.parameters(), **optimizer_args)
-        estimator = PytorchEstimator(client_model, loss_fn, optimizer, target=target)
-        if server_model is not None:
-            estimator.add_server_model(server_model, loss_fn, optimizer_cls, optimizer_args)
+        estimator = PytorchEstimator(model=client_model,
```
Litchilitchy, line 39: Estimator refactor and make `add_server_model` private

python/ppml/test/bigdl/ppml/fl/nn/pytorch/test_logistic_regression.py
```diff
@@ -49,7 +49,8 @@ def mock_process(data_train, target, client_id):
     set_one_like_parameter(model)
     loss_fn = nn.BCELoss()
     server_model = LogisticRegressionNetwork2()
-    ppl = Estimator.from_torch(client_model=model, 
+    ppl = Estimator.from_torch(client_model=model,
+                               client_id=client_id,
```
Litchilitchy, line 53: The related client_id in another test party has already been set [here](https://github.com/intel-analytics/BigDL/blob/e1fae35cf75b3a4878a4306dedef0bc4976a675b/python/ppml/test/bigdl/ppml/fl/nn/pytorch/test_logistic_regression.py#L114)

python/ppml/src/bigdl/ppml/fl/nn/pytorch/aggregator.py
Litchilitchy: This is changed into `Estimator.from_torch(client_id=...)`

PR #4700, opened 2022-05-26
Fix Apps Test

sgwhat:
Apps test failed because RayContext has been renamed to OrcaRayContext.


sgwhat:
Pass http://10.112.231.51:18888/job/BigDL-PR-Python-AppTests-part2/2/

PR #4707, opened 2022-05-26
replace raise, assert with new api in ppml

dding3:
1. replace raise, assert with new api in ppml
2. disallow use raise, assert in ppml


dding3:
most jekins pass. http://10.112.231.51:18889/view/BigDL-PR-Validation/job/BigDL-PR-scala-ppml-unittests-Validation/144/. Failed one is not related to this pr

PR #4708, opened 2022-05-27
Chronos: Fixed printing unexpected warnings when import forecaster

liangs6212:


python/chronos/src/bigdl/chronos/forecaster/__init__.py
```diff
@@ -35,8 +35,10 @@
 except:
     warnings.warn("Please install `tensorflow>2.0.0` to use MTNetForecaster.")
 try:
-    import prophet
-    prophet_available = True
+    dependency_list = subprocess.getoutput(["pip freeze | grep prophet"])
```
TheaperDeng, line 38: `pip` could be not available in some users' environment (e.g. they use `pip3` or something like it).

TheaperDeng:
https://github.com/intel/neural-compressor/blob/master/neural_compressor/adaptor/tensorflow.py#L31
https://github.com/intel/neural-compressor/blob/a89afcf97cb76dc30251078a7788efbef31b636e/neural_compressor/utils/utility.py#L46

TheaperDeng (approved):

liangs6212:
http://10.112.231.51:18889/view/BigDL-PR-Validation/job/BigDL-Chronos-PR-Validation/573/

PR #4717, opened 2022-05-30
Move xgboost example from zoo to bigdl

Forest216:


Forest216:
In BigDL, setFeaturesCol requires a String to select the feature to use  https://github.com/intel-analytics/BigDL/blob/main/scala/dllib/src/main/scala/com/intel/analytics/bigdl/dllib/nnframes/NNClassifier.scala#L414
So we need to do a VectorAssembler before setFeaturesCol https://github.com/intel-analytics/BigDL/pull/4717/files#diff-4aa257635730db5cda00b4650100078cf47d0031d1aa40804d8032fbf61ee902R70

qiyuangong (approved):
LGTM

PR #4719, opened 2022-05-30
Refine ppml crypto 

qiuxin2012:


qiuxin2012:
jenkins passed

scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/crypto/Crypto.scala
```diff
@@ -16,25 +16,108 @@
...
+  def encryptFile(binaryFilePath: String, savePath: String): Unit
+}
+
+object Crypto {
+  def apply(cryptoMode: CryptoMode): Crypto = {
+    cryptoMode match {
+      case AES_CBC_PKCS5PADDING =>
+        new BigDLEncrypt()
+      case default =>
+        throw new EncryptRuntimeException("No such crypto mode!")
```
jason-dai, line 58: do not throw exception here; use the new error API (https://github.com/intel-analytics/BigDL/blob/main/docs/readthedocs/source/doc/error-log-api.md). @dding3 - have we enforced the style check for scala?

jason-dai:
@qiuxin2012 
1) Please add scala doc
2) Do we have test for `PPMLContext.write`?

qiuxin2012:
> @qiuxin2012
> 
> 1. Please add scala doc
> 2. Do we have test for `PPMLContext.write`?

1. Sure, I will add scala doc later.
2. PPMLContext.write should be changed in next PR.

scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/crypto/Crypto.scala
qiuxin2012: > do not throw exception here; use the new error API (https://github.com/intel-analytics/BigDL/blob/main/docs/readthedocs/source/doc/error-log-api.md). @dding3 - have we enforced the style check for scala?

Type mismatch, if I just change the this line to `Log4Error.invalidInputError(true, "Unsupported crypto mode!")`
I need some work around.

dding3: > I just change the this line to `Log4Error.invalidInputError(true, "Unsupported crypto mode!")`

We have enforced no assert, no require check in scala, disallow `throw exception` is on the way.

dding3: > > do not throw exception here; use the new error API (https://github.com/intel-analytics/BigDL/blob/main/docs/readthedocs/source/doc/error-log-api.md). @dding3 - have we enforced the style check for scala?
> 
> Type mismatch, if I just change the this line to `Log4Error.invalidInputError(true, "Unsupported crypto mode!")` I need some work around.

I think after call `Log4Error.invalidInputError(true, "Unsupported crypto mode!")`, you can add `null` to fix the compile issue

PR #4721, opened 2022-05-30
[Nano] new tensorflow quantize api

zhentaocc:
Still WIP.  
Make tensorflow api consistent with pytorch.

### DIFF
--- a/python/nano/src/bigdl/nano/tf/quantization.py
+++ b/python/nano/src/bigdl/nano/tf/quantization.py
@@ -21,17 +21,17 @@ from bigdl.nano.utils.log4Error import invalidInputError
 
 
 def quantize(self,
-             precision='int8',
-             accelerator=None,
+             precision: str = 'int8',
+             accelerator: str = None,
              calib_dataset: tf.data.Dataset = None,
-             metric: Optional[Metric] = None,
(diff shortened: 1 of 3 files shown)


python/nano/src/bigdl/nano/tf/quantization.py
```diff
@@ -21,6 +21,7 @@
 
 
 def quantize(self,
+             precision='int8',
```
jason-dai, line 24: will it be possible to support other accelerators in future (such as openvino)?

zhentaocc: Yes, sure for openvino. As for onnx, maybe we can follow https://onnxruntime.ai/docs/tutorials/tf-get-started.html. But before that, we should figure out how to wrap a quantized tf model.

jason-dai: The we may need an `accelerator` parameter?

zhentaocc:
http://10.112.231.51:18888/job/BigDL-PRNV-chronos-Python-Spark-2.4-py37-Horovod/1085/  
http://10.112.231.51:18888/job/BigDL-NanoTests-PR-Validation/207/

Issue #4689, opened 2022-05-25
bigdl-nano-init slows down tcn forecaster with multiple processing


yushan111:
### To produce: 
1. `source bigdl-nano-init`
2. run `fit` in https://github.com/intel-analytics/BigDL/blob/main/python/chronos/use-case/network_traffic/network_traffic_multivariate_multistep_tcnforecaster.ipynb

### More info

18 physical cores, run `num_processes=2` with nano pytorch trainer  
Intel(R) Core(TM) i9-9980XE CPU @ 3.00GHz

TheaperDeng:
please have a try @liangs6212 to reproduce this. thx

liangs6212:
The problem still exists when `forecaster.numprocess` is not 1. I tried two solutions,
* unset `KMP_INIT_AT_FORK`
* Make sure `TCNForecaster` is initialized before `sklearn`.

I ensure correctness by restarting the kernel multiple times when testing in the notebook.

TestInfo
* Intel(R) Core(TM) i9-7900X CPU @ 3.30GHz
* 10 physical cores (set magic_num to 4)
* forecaster.num_processes is 2
* network_traffic_multivariate_multistep_tcnforecaster

TheaperDeng:
same issue to #4590

Issue #4590, opened 2022-05-11
Nano: Sklearn has some compatibility issue with nano.pytorch's ddp training


TheaperDeng:
```python
from sklearn.preprocessing import StandardScaler
# any nano application
# (e.g. https://github.com/intel-analytics/BigDL/blob/main/python/nano/example/pytorch/pl-finetune/pl-finetune.py)
```

Then we will have 
```bash
OMP: Warning #124: Ignoring invalid OS proc ID 1.
OMP: Warning #124: Ignoring invalid OS proc ID 2.
OMP: Warning #124: Ignoring invalid OS proc ID 3.
...
```
And only core 0 will be used.

We @liangs6212 found that it could be caused by the setting of `KMP_INIT_AT_FORK`
https://github.com/scikit-learn/scikit-learn/blob/9b033758ec681e8fd7433a8bb35d9777acd4f8ba/sklearn/__init__.py#L57

PR #4731, opened 2022-05-31
Chronos/Nano: unset the KMP_INIT_AT_FORK for pytorch training

TheaperDeng:
related to
https://github.com/intel-analytics/BigDL/issues/4689
https://github.com/intel-analytics/BigDL/issues/4590

### DIFF
--- a/python/chronos/src/bigdl/chronos/forecaster/__init__.py
+++ b/python/chronos/src/bigdl/chronos/forecaster/__init__.py
@@ -13,7 +13,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+
 import warnings
+# unset the KMP_INIT_AT_FORK
+# which will cause significant slow down in multiprocessing training
+import os
+os.unsetenv('KMP_INIT_AT_FORK')
 
 
 # dependencies check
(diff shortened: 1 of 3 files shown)


python/chronos/src/bigdl/chronos/forecaster/__init__.py
```diff
@@ -13,7 +13,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+
 import warnings
+# unset the KMP_INIT_AT_FORK
+# which will cause significant slow down in multiprocessing training
+import os
+os.unsetenv('KMP_INIT_AT_FORK')
```
jason-dai, line 21: why do you need to unset it in both nano and chronos?

TheaperDeng: We should avoid any `import torch` or with this env variable still exist.
In Chronos, nano trainer is imported really late (during `fit`), torch is imported before nano trainer's import, so I need to do it here.

TheaperDeng:
@yangw1234 any comments on this PR?

yangw1234:
LGTM

TheaperDeng:
http://10.112.231.51:18888/job/BigDL-NanoTests-PR-Validation/225/
http://10.112.231.51:18888/job/BigDL-PRVN-chronos-Python-Spark-3.1-py37-Horovod/1106/

PR #4733, opened 2022-05-31
change nano hpo tf resume searching behavior

guoyaol:


python/nano/src/bigdl/nano/automl/tf/mixin.py
```diff
@@ -229,6 +230,19 @@ def search(
             and search backend arguments (e.g. n_trials, pruner, etc.)
             are allowed in kwargs.
         """
+        do_create = True
+        if resume:
+            if 'storage' not in kwargs.keys() or kwargs['storage'].strip() == "":
+                if self.study is None:
+                    warnings.warn(
+                        'You did not specify storage path and your study is None. \
```
shane-huang, line 238: Change the warning message to "A new study is created since there's no existing study to resume from. "

PR #4745, opened 2022-06-01
Nano: refactor nano test code

y199387:


python/nano/test/ray/test_ray_trainer.py
```diff
@@ -21,8 +21,8 @@
 import pytest
 import torch
 from pytorch_lightning import LightningModule
-from _train_torch_lightning import create_data_loader, data_transform
-from _train_torch_lightning import train_with_linear_top_layer
+from python.nano.test.pytorch.utils._train_torch_lightning import create_data_loader, data_transform
```
yangw1234, line 24: I think the import should be `test.pytorch.utils` as the directory of `python` and `nano` are not python packages.

The reason why `python.nano.test.` works is because the `python -m pytest` will add current directory in `sys.path`

yangw1234:
Some directories do not have `__init__.py` file and maybe this is the reason why importing `test.xxx` does not work.

python/nano/test/run-nano-pytorch-inc-tests.sh
```diff
@@ -2,7 +2,7 @@
 
 export ANALYTICS_ZOO_ROOT=${ANALYTICS_ZOO_ROOT}
 export NANO_HOME=${ANALYTICS_ZOO_ROOT}/python/nano/src
-export INC_PYTORCH_NANO_TEST_DIR=${ANALYTICS_ZOO_ROOT}/python/nano/test/neural-compressor/pytorch
+export INC_PYTORCH_NANO_TEST_DIR=${ANALYTICS_ZOO_ROOT}/python/nano/test/neuralCompressor/pytorch
```
yangw1234, line 5: why change the directory name?

yangw1234: I think underscore is more appropriate

y199387: But the name neural-compressor will conflicts with other modules

yangw1234:
LGTM, we can merge it after jenkins pass

y199387:
Jenkins pass
http://10.112.231.51:18888/job/BigDL-NanoTests-PR-Validation/216/

PR #4750, opened 2022-06-02
[PPML] Migrate SplitAndEncrypt From E2E To BigDL

piaolaidelangman:


jenniew:
For csv, Split and Encrypt may use PPMLContext.read.csv(), then repartition and PPMLContext.write.csv() to save to CSV with encryption.
For other textfile, Split and Encrypt may use PPMLContext.textFile(), then repartition and PPMLContext.saveAsTextFile() with encryption to save. @qiuxin2012 , will we add this api?
That may be faster for huge file.

jason-dai:
> For csv, Split and Encrypt may use PPMLContext.read.csv(), then repartition and PPMLContext.write.csv() to save to CSV with encryption. For other textfile, Split and Encrypt may use PPMLContext.textFile(), then repartition and PPMLContext.saveAsTextFile() with encryption to save. @qiuxin2012 , will we add this api? That may be faster for huge file.

Yes, the plan is to call PPMLContext read/write in the CLI tools

piaolaidelangman:
> > For csv, Split and Encrypt may use PPMLContext.read.csv(), then repartition and PPMLContext.write.csv() to save to CSV with encryption. For other textfile, Split and Encrypt may use PPMLContext.textFile(), then repartition and PPMLContext.saveAsTextFile() with encryption to save. @qiuxin2012 , will we add this api? That may be faster for huge file.
> 
> Yes, the plan is to call PPMLContext read/write in the CLI tools

I will close this pull request.

PR #4758, opened 2022-06-06
add yolo test back

Le-Zheng:


Le-Zheng:
jenkins passed: http://10.112.231.51:18888/job/ZOO-PR-BigDL-Python-ExampleTests-py37-ray-horovod-Spark3/707/

PR #4761, opened 2022-06-06
[Nano]Enable quantized keras model

zhentaocc:
Follow what we did for pytorch to enable transparent INC acceleration for keras.  
### Before
```python
q_model = model.quantize(calib_dataset=train_dataset)
with q_model.sess as sess:
    output = sess.run(q_model.output_tensor,
                      feed_dict={q_model.input_tensor[0]: train_examples[0:10]})
assert output[0].shape == (10, 10)
```
### After
```python
q_model = model.quantize(calib_dataset=train_dataset)
output = q_model(train_examples[0:10])
assert output.shape == (10, 10)
```


zhentaocc:
http://10.112.231.51:18888/job/BigDL-NanoTests-PR-Validation/194/

PR #4762, opened 2022-06-06
import visualization from nano hpo

guoyaol:


shane-huang:
just import * from optuna.visualization.

python/nano/test/automl/common/test_visualization.py
```diff
@@ -0,0 +1,68 @@
...
+            from bigdl.nano.automl.hpo.visualization import plot_contour
+        except ImportError:
+            self.fail("cannot import plot_contour from nano.aotoml.hpo.visualization.")
+        try:
+            from bigdl.nano.automl.hpo.visualization import plot_param_importances
+        except ImportError:
+            self.fail("cannot import plot_param_importances from nano.aotoml.hpo.visualization.")
+
+    def test_multi_enable_disable(self):
+        nano_automl.hpo_config.disable_hpo_tf()
```
shane-huang, line 48: no need to enable and disable tf. This is common util and have nothing todo with tf and pytorch.

PR #4765, opened 2022-06-06
Nano: relax dataloader check in quantization

TheaperDeng:
This is a quick fix for demo.
1. change the check from strictly pytorch dataloader to an Iterable
2. remove all the check under non-tunning quantization


TheaperDeng:
http://10.112.231.51:18888/job/BigDL-NanoTests-PR-Validation/197/

Issue #4769, opened 2022-06-07
Nano: A utils function to generate the environment variables to be set by `bigdl-nano-init`


TheaperDeng:
This is useful when we need to set it in some python API.

```python
res = gen_nano_env_var()
print(res)

# {"KMP_BLOCKTIME": 1.
# ...}
```

PR #4776, opened 2022-06-07
Nano: A utils function to generate the environment variables for nano

TheaperDeng:
related to #4769 

TODO: align this function with `bigdl-nano-init` script.


TheaperDeng:
http://10.112.231.51:18888/job/BigDL-NanoTests-PR-Validation/208/

PR #4778, opened 2022-06-07
fix: use -f as file test operator for file secured-argvs

liu-shaojun:
In Linux Shell Script:
**if [ -c filename ]**: checks if file is a character special file
**if [ -f filename ]**: checks if file is an ordinary file as opposed to a directory or special file

In ppml/trusted-big-data-ml/python/docker-graphene/init.sh, /ppml/trusted-big-data-ml/secured-argvs is an ordinary file, so we should use operator **-f**.


glorysdj (approved):
LGTM

PR #4783, opened 2022-06-07
[PPML] Add TPC-H example

jenniew:
Add TPC-H example to BigDL PPML


jason-dai:
Need to add reference and credit for where the code is adapted from

scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/examples/tpch/Q01.scala
```diff
@@ -0,0 +1,63 @@
+/*
```
jason-dai, line 1: Did we make any changes to the files?

jenniew: no change for Q01-Q22.
modified TPCHQuery and TPCHSchemaProvider

jason-dai: > no change for Q01-Q22. modified TPCHQuery and TPCHSchemaProvider

If there no change, you need to add "This file is copied from ..." at the beginning

If the file is changed, you may refer to an example at https://github.com/intel-analytics/BigDL/blob/main/python/orca/example/learn/tf2/yolov3/__init__.py

jenniew: Our scala style check only allow BigDL license in the header.  So maybe we split the original license to second part? Or we change our scala style check not contain ending "*/" in the header?

jason-dai: just exclude these files in style check

jenniew: ok

jenniew:
jenkins pass: http://10.112.231.51:18889/job/BigDL-PR-scala-ppml-unittests-Validation/201/

PR #4787, opened 2022-06-08
Chronos: Extend the predict method

liangs6212:
About 4X speedup when `batch_size` is set to None, not recommended for large datasets.

Removed `training=False`, because normally the user does not update the weight matrix when predicting.


TheaperDeng (approved):
LGTM

liangs6212:
http://10.112.231.51:18888/job/BigDL-Chronos-PR-Validation/1062/

PR #4789, opened 2022-06-08
Refactor NN Pytorch to a Estimator-style API

Litchilitchy:


python/ppml/src/bigdl/ppml/fl/nn/pytorch/estimator.py
```diff
@@ -36,7 +35,13 @@ def __init__(self, model: nn.Module, loss_fn, optimizer: torch.optim.Optimizer,
         self.fl_client = FLClient(target)
         self.loss_history = []
 
-    
+    @staticmethod
+    def from_torch(model: nn.Module,
```
jason-dai, line 39: make it `ppml.vfl.Estimator.from_torch` instead

python/ppml/src/bigdl/ppml/fl/nn/pytorch/estimator.py
```diff
@@ -36,7 +35,13 @@ def __init__(self, model: nn.Module, loss_fn, optimizer: torch.optim.Optimizer,
...
+                   loss_fn,
+                   optimizer: torch.optim.Optimizer,
+                   target="localhost:8980"):
+        return PytorchEstimator(model, loss_fn, optimizer, target=target)
+
     
     def add_server_model(self, model: nn.Module, loss_fn=None, optimizer_cls=None, optimizer_args={}):
```
jason-dai, line 46: instead of `add_server_model`, make it 
```python
Estimator.from_torch(client_model, 
                     client_optimizer, 
                     server_model=None, #only set when client_id==1
                     server_optimizer=None, 
                     ...)
```

python/ppml/test/bigdl/ppml/fl/pytorch/test_logistic_regression.py
```diff
@@ -48,7 +48,7 @@ def mock_process(data_train, target):
     loss_fn = nn.BCELoss()
     optimizer = torch.optim.SGD(model.parameters(), lr=1e-3)
     server_model = LogisticRegressionNetwork2()
-    ppl = PytorchPipeline(model, loss_fn, optimizer, target=target)
+    ppl = PytorchEstimator(model, loss_fn, optimizer, target=target)
```
jason-dai, line 51: use `Estimator.from_torch`

python/ppml/test/bigdl/ppml/fl/pytorch/test_mnist.py
```diff
@@ -111,7 +111,7 @@ def train(dataloader, model, loss_fn, optimizer):
         vfl_model_1 = NeuralNetworkPart1()
         optimizer1 = torch.optim.SGD(vfl_model_1.parameters(), lr=1e-3)
         set_one_like_parameter(vfl_model_1)
-        vfl_client_ppl = PytorchPipeline(vfl_model_1, loss_fn, optimizer1, target=self.target)
+        vfl_client_ppl = PytorchEstimator(vfl_model_1, loss_fn, optimizer1, target=self.target)
```
jason-dai, line 114: `Estimator.from_torch`

python/ppml/src/bigdl/ppml/fl/nn/pytorch/estimator.py
Litchilitchy: do we need to make `ppml.fl....` and `ppml.vfl` co-exist or `ppml.fl.vfl...`

jason-dai: Maybe `ppml.fl.Estimator` (if we can extend to HFL later)?

Litchilitchy: @jason-dai `fl.Estimator` is available now. And note that `PytorchEstimator` could not extend `fl.Estimator` but have to extend another abstract class because of the circular dependency import issue. This is the same with the implementation in Orca

python/ppml/src/bigdl/ppml/fl/nn/pytorch/estimator.py
jason-dai: I don't think we need `add_server_model` here

Litchilitchy: This is just a wrapper method and with `fl.Estimator` I guess this `pytorch.estimator` could be no longer exposed to users?

jason-dai: The `add_server_model` is redundant and can be removed.

Litchilitchy: Currently the style is `fl.Estimator.from_torch` creates an instance of `PytorchEstimator` and pass `server_model` to the args of `add_server_model` so this method is not redundant I guess?

jason-dai: then make it internal

Litchilitchy: `fl.Estimator` and `pytorch.Estimator` has no inheriting relation so private/protected method could not be called across class. 

Current API is `fl.Estimator.from_torch` and `fl.Estimator.from_keras`, should we change to `pytorch.Estimator` and `tensorflow.Estimator` respectively?

PR #4792, opened 2022-06-08
Chronos: add lightweight hpo to pytorch forecaster

shane-huang:
`Forecaster.tune` can be used to tune the model. After tune, Forecaster.fit will automatically fit on the best model. 

There's several issues left to do in following PRs:

1. The model built in `Forecaster.tune` has train_dataloader and val_dataloader attached to it (so that batch size can be tuned).  
`Forecaster.fit` will use the new train_dataloader instead of the train_dataloader attached to model, which is the correct behavior. But since no validation data is provided in fit, validation data is still there.  
1. the best batch size needs to be applied to train data loader construction in fit.  
1. tune only support numpy array input at present, so that batch size can be tuned. We could also support dataloaders as input just don't provide batch size tuning option. 
1. more thorough tests needs to be done for tcnforecaster.tune. Also a demo is needed.


shane-huang:
for val_dataloader, since tuning does not affect that, I can remove the method val_data_loader from the LightningModule and make it as a fixed input. So model won't have attached val_dataloader.

PR #4798, opened 2022-06-08
Chronos: support customized loss and metric in forecaster

TheaperDeng:


TheaperDeng:
http://10.112.231.51:18888/job/BigDL-Chronos-PR-Validation/982/

PR #4806, opened 2022-06-10
Update python.md

qiuxin2012:


docs/readthedocs/source/doc/UserGuide/python.md
```diff
@@ -84,9 +84,14 @@ You may test if the installation is successful using the interactive Python shel
 * Try to run the example code below to verify the installation:
 
   ```python
-  from bigdl.orca import init_orca_context
+from bigdl.dllib.nncontext import init_nncontext
```
hkvision, line 87: The layout is broken: https://github.com/intel-analytics/BigDL/blob/705c1b24deec64651580fba22cd7a18c952aa2f2/docs/readthedocs/source/doc/UserGuide/python.md

python/dllib/src/bigdl/dllib/nn/keras/layers/topology.py
```diff
@@ -209,13 +209,13 @@ class Sequential(KerasModel):
...
-    .. note:: `bigdl.dllib.keras` is deprecated in 0.11.
+    .. note:: `bigdl.dllib.nn.keras` is deprecated in 0.11.
     This will be removed in future releases.
     """
 
     def __init__(self, jvalue=None, **kwargs):
-        warnings.warn("bigdl.dllib.keras is deprecated in 0.11. "
-                      "Recommend to use Analytics Zoo's Keras API.")
+        warnings.warn("bigdl.dllib.nn.keras is deprecated in 0.11. "
+                      "Recommend to use bigdl.dllib.keras Keras API.")
```
hkvision, line 218: remove `Keras API`?

docs/readthedocs/source/doc/UserGuide/python.md
qiuxin2012: fixed

docs/readthedocs/source/doc/UserGuide/python.md
```diff
@@ -84,9 +84,13 @@ You may test if the installation is successful using the interactive Python shel
 * Try to run the example code below to verify the installation:
 
   ```python
-  from bigdl.orca import init_orca_context
+  from bigdl.dllib.nncontext import init_nncontext
```
jason-dai, line 87: why change from orca to dllib?

qiuxin2012: Kai said it's a little confused init orca context, but call dllib's library.

jason-dai: But still why do we use dllib instead of orca to verify?

hkvision: What Orca code do you suggest to verify? Seems it is not easy to find some simplest code in Orca?

jason-dai: `init_orca_context` on local? if this works, the installation should be successful?

hkvision: So do not modify this? Already there: https://bigdl.readthedocs.io/en/latest/doc/UserGuide/python.html#interactive-shell

jason-dai: Maybe show what the example output should look like if the installation is successful

PR #4829, opened 2022-06-14
[Nano] Fix on Pytorch Inference Doc

zhentaocc:
- fix a few format issues
- restrict inc to 1.11. We don't need to use multiple versions


yangw1234:
LGTM

PR #4830, opened 2022-06-14
update jupyter to jupyterLab

zzti-bsj:
1. update Dockerfile for jupyterLab installation
2. update start-notebook.sh to support jupyterLab


glorysdj (approved):
LGTM

PR #4838, opened 2022-06-14
Fix stackoverflowerror for transformer example

dding3:
Fix https://github.com/intel-analytics/arda-docker/issues/657


dding3:
most jekins pass. http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-Orca-PR-Validation/720/

The failed one http://10.112.231.51:18888/job/ZOO-PR-BigDL-Python-ExampleTests/753//console is using nightly build whl, which doesn't include this fix. It should be able to pass after this pr merged and new whl is generated.

PR #4840, opened 2022-06-14
add avx512 check

y199387:
add avx512 check in nano pytorch trainer


yangw1234:
let's also add the ipex 1.9 test in this pr.

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
```diff
@@ -91,6 +92,13 @@ def __init__(self, num_processes: int = 1,
             self.hposearcher = None
 
         accelerator = None
+
+        if TORCH_VERSION_LESS_1_10 and use_ipex and not check_avx512():
+            warning("Enable ipex1.9 in a cpu instruction set"
+                    " without avx512 may cause some random error."
```
yangw1234, line 98: "will crash" not "may cause some random error"

yangw1234:
LGTM

PR #4842, opened 2022-06-15
Chronos: make multi-process training more robust

TheaperDeng:
1. automatically use multi-process training when user has an environment larger or equal to 24 cores.
2. change distributed_backend to spawn.


TheaperDeng:
http://10.112.231.51:18888/job/BigDL-Chronos-PR-Validation/1002/

PR #4857, opened 2022-06-16
Chronos: use tstrainer in ae_detector

Rebekk:


TheaperDeng:
try this use-case https://github.com/intel-analytics/BigDL/blob/main/python/chronos/use-case/AIOps/AIOps_anomaly_detect_unsupervised.ipynb

and test some performance data

python/chronos/src/bigdl/chronos/detector/anomaly/ae_detector.py
```diff
@@ -160,19 +161,17 @@ def fit(self, y):
             ae_model = create_torch_model(self.compress_rate, len(y[0]))
             optimizer = optim.Adadelta(ae_model.parameters(), lr=self.lr)
             criterion = nn.BCELoss()
+            internal_model = Trainer.compile(model=ae_model, loss=criterion,
+                                             optimizer=optimizer)
+            trainer = Trainer(logger=False, max_epochs=self.epochs)
```
TheaperDeng, line 166: try if num_processes>1 helps the training performance. if helps, please follow this policy: https://github.com/intel-analytics/BigDL/blob/main/python/chronos/src/bigdl/chronos/forecaster/lstm_forecaster.py#L132-L135

TheaperDeng: checkpoint to false

TheaperDeng: if checkpoint is false, distributed_backend should be spawn.

python/chronos/src/bigdl/chronos/detector/anomaly/ae_detector.py
```diff
@@ -160,19 +161,17 @@ def fit(self, y):
...
-                for x_batch, y_batch in train_loader:
-                    optimizer.zero_grad()
-                    yhat = ae_model(x_batch)
-                    loss = criterion(yhat, y_batch)
-                    loss.backward()
-                    optimizer.step()
+            trainer.fit(internal_model, train_loader)
             y_pred_list = []
             for x_batch, y_batch in train_loader:
                 y_pred_list.append(ae_model(x_batch).detach().numpy())
```
TheaperDeng, line 177: use `internal_model` here and wrap it under a 
```python
internal_model.eval()
with torch.no_grad():
    ...
```

PR #4858, opened 2022-06-16
[Orca] Support Pytorch Pyspark Estimator on K8s

sgwhat:
This PR is mainly to support tcp server store creation when running pytorch ddp on k8s and fix https://github.com/intel-analytics/arda-docker/issues/647.

To Do:

- [x] Test on k8s client
- [x] Test on k8s cluster
- [x] Test on yarn client/cluster
- [x] Test on standalone


python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
```diff
@@ -158,15 +158,25 @@ def __init__(
...
+    def create_tcpstore_server(self):
+        import torch.distributed as dist
+        from torch._C._distributed_c10d import _DEFAULT_PG_TIMEOUT
+        server_store = dist.TCPStore(self.ip, self.tcp_port, -1, True, _DEFAULT_PG_TIMEOUT)
+        return server_store
+
+    def _get_cluster_info(self):
+        cluster_info = []
+        for i in range(self.num_executors):
+            cluster_info.append(f"{self.ip}:{self.tcp_port}")
```
yushan111, line 179: It seems all workers get the same `ip:port` in `cluster_info`. Since we infer the world rank for each worker through `cluster_info`, that may lead to different workers sharing the same rank. I wonder whether PyTorch ddp can work correctly in that case...

yushan111 (commented):
It would be helpful if you elaborate on the test case/examples you used for testing.

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
```diff
@@ -158,15 +158,25 @@ def __init__(
 
         local_init_params = self.worker_init_params.copy()
         local_init_params["log_to_driver"] = False
+        self.tcp_port = find_free_port()
+        self.num_executors = int(sc.getConf().get("spark.executor.instances"))
```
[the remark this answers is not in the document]
sgwhat: > What's the difference between `num_executors` and `num_workers`

seems they are the same, maybe I should use `num_workers` directly.

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
sgwhat: But workers should get the same `ip:port`, I will add a new function to deal with rank.

jason-dai: Why are we using the driver `ip:port` in the `cluster_info`? I would assume `cluster_info` contains the information of all the PyTorch workers?

sgwhat: > Why are we using the driver `ip:port` in the `cluster_info`? I would assume `cluster_info` contains the information of all the PyTorch workers?

TCPStore is a server-client, when the server store is created on driver, we need to set other client stores' URL and Port the same as the server store. Or it will cause a connection timeout error.

sgwhat: > Why are we using the driver `ip:port` in the `cluster_info`? I would assume `cluster_info` contains the information of all the PyTorch workers?

We cannot use each worker's `ip:port` to create tcpstore, that's not a work solution.

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
```diff
@@ -222,6 +232,7 @@ def fit(self,
                                                            num_workers=self.num_workers)
 
         sc = OrcaContext.get_spark_context()
+        server_store = self.create_tcpstore_server()
```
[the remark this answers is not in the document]
sgwhat: > where it is used?

on driver, to create a server store before creating other client stores.

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
yushan111: @sgwhat Shall we keep the original `cluster_info`, and maybe directly pass the `driver_ip` and `tcp_port` to workers?

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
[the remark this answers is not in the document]
yushan111: Maybe omit the return value if it isn't used.

yushan111:
It would be helpful if you elaborate on how you test this PR. 

A thorough test should include train/evaluate/predict. You should also test whether the logging to driver works.

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
sgwhat: > @sgwhat Shall we keep the original `cluster_info`, and maybe directly pass the `driver_ip` and `tcp_port` to workers?

Replacing the `cluster_info` is one of the PR's purpose, since the original `cluster_info` is not working on k8s.

jason-dai:
> @yushan111 How do I test the PR:
> 
> Demo: We deploy two K8s Pods, first create a server TCPStore on driver, then active client store on each pod to connect with the server store.
> 
> BigDL K8s:
> 
> * example: a simple pytorch pyspark estimator example on K8s.
> * executors: we deploy a modified image on executors to log the client TCPStore information.
> * driver: we collect the new `cluster_info` and the whole example works well without any connection error.

There is no code in BigDL that connects to the driver TCP store; does it mean the user has to depend on our docker image?

sgwhat:
> 

I add a 

> > @yushan111 How do I test the PR:
> > Demo: We deploy two K8s Pods, first create a server TCPStore on driver, then active client store on each pod to connect with the server store.
> > BigDL K8s:
> > 
> > * example: a simple pytorch pyspark estimator example on K8s.
> > * executors: we deploy a modified image on executors to log the client TCPStore information.
> > * driver: we collect the new `cluster_info` and the whole example works well without any connection error.
> 
> There is no code in BigDL that connects to the driver TCP store; does it mean the user has to depend on our docker image?

That's all done by PyTorch, we just need to create a server store, then set the `url` param in `dist.init_process_group` as the driver's address, Pytorch will handle the connection stuff.

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
sgwhat: > 

We could no longer use `cluster_info`, the key objective is to make sure each executor could reach the `ip:port` that we used to create server store on the driver.

sgwhat: > > > @sgwhat Shall we keep the original `cluster_info`, and maybe directly pass the `driver_ip` and `tcp_port` to workers?
> > 
> > 
> > Replacing the `cluster_info` is one of the PR's purpose, since the original `cluster_info` is not working on k8s.
> 
> How is `cluster_info` used after the changes? It does not make sense to have a `cluster_info` with all the same `ip:port`. Maybe you no longer need a `cluster_info` after the changes.

We could no longer use `cluster_info`, the key objective is to make sure each executor could reach the `ip:port` that we used to create server store on the driver.

yushan111: `cluster_info` is also used to infer **rank** for each worker. You need to fix it as we mentioned before, the identical `ip:port` results in the same rank for different workers. Also, testing with more than 2 pods should uncover the issue. 

Replacing `cluster_info` might be a way but not the purpose of this PR. Before the change, we used `cluster_info[0]` as an identifier URL for pytorch DDP (`dist.init_processs_group(init_method=cluster_info[0])`). You could just change the URL to `{driver_ip}:{port}` to achieve your purpose, instead of directly changing the semantics of `cluster_info`.





https://github.com/intel-analytics/BigDL/blob/0d6d445eed986379222a55c7f0e84b0c29631a99/python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_worker.py#L114-L118

yushan111:
We might need tests with more than two pods. It would be great if you could also monitor the training process to make sure it works as expected. (e.g the loss, accuracy)

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
sgwhat: > `cluster_info` is also used to infer **rank** for each worker. You need to fix it as we mentioned before, the identical `ip:port` results in the same rank for different workers. Also, testing with more than 2 pods should uncover the issue.
> 
> Replacing `cluster_info` might be a way but not the purpose of this PR. Before the change, we used `cluster_info[0]` as an identifier URL for pytorch DDP (`dist.init_processs_group(init_method=cluster_info[0])`). You could just change the URL to `{driver_ip}:{port}` to achieve your purpose, instead of directly changing the semantics of `cluster_info`.
> 
> https://github.com/intel-analytics/BigDL/blob/0d6d445eed986379222a55c7f0e84b0c29631a99/python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_worker.py#L114-L118

It's a better solution than replacing `cluster_info` directly.

sgwhat:
> We might need tests with more than two pods. It would be great if you could also monitor the training process to make sure it works as expected. (e.g the loss, accuracy)

The number of pods does not affect the testing result, it's just used to connect the client store to the server store manually. We will compare the loss and accuracy ion k8s with the original code running on yarn, which could work as well.

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
yushan111: > It's a better solution than replacing cluster_info directly.

Great!

sgwhat:
> > The number of pods does not affect the testing result, it's just used to connect the client store to the server store manually.
> 
> Still for the concern about the correctness of the `rank` value of each worker...

Oh I see, you mean the number of executors when submitting application on k8s. I always set it to 4.

yushan111:
> Oh I see, you mean the number of executors when submitting application on k8s.

Exactly. The number of executor pods automatically started by Spark while using Pytorch Pyspark Estimator.

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
```diff
@@ -165,6 +167,12 @@ def __init__(
 
         self.state_dict = self.driver_runner.get_state_dict()
 
+    def create_tcpstore_server(self):
+        import torch.distributed as dist
+        from torch._C._distributed_c10d import _DEFAULT_PG_TIMEOUT
```
yushan111, line 172: We should avoid directly using torch private constants, since it may change in different torch versions. 

Firstly, ` _DEFAULT_PG_TIMEOUT` is only introduced after `1.9.0`. Being compatible with some lower versions might be good for user experience. 
Secondly,  it could introduce bugs, which might be very hard to locate, as torch evolves.

yushan111: Shall we just change it to 
```python
from dist.constants import default_pg_timeout
```

sgwhat: > Shall we just change it to
> 
> ```python
> from dist.constants import default_pg_timeout
> ```

Maybe just use the default value `timeout: datetime.timedelta = datetime.timedelta(seconds=300)` ? Since the 30 minutes of `dafault_pg_timeout` may too many for creating server store on client node.

yushan111: > Maybe just use the default value timeout: datetime.timedelta = datetime.timedelta(seconds=300) ? Since the 30 minutes of dafault_pg_timeout may too many for creating server store on client node.

Pytorch DDP is built on the c10d library. And pytorch DDP uses a default timeout of `dafault_pg_timeout` to create TCPStore in `dist.init_process_group` ([code](https://github.com/pytorch/pytorch/blob/b4ed13ea0ff091328c6d0dfdf5d751d4280fb67f/torch/distributed/distributed_c10d.py#L488-L491)).

So I guess the default timeout of Pytorch DDP, instead of the default timeout of TCPStore in c10d library, should be more appropriate for our distributed training based on DDP.

yushan111:
Since we already created a TCP store on driver, shall we just pass the store to init_process_group, instead of passing a URL as an `init_method`?
```python
dist.init_process_group(store=..., rank=..., world_size=...)
```
Otherwise, DDP will automatically create a TCPStore sever again on worker with rank0 with your URL in `init_method`.

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
```diff
@@ -154,7 +155,8 @@ def __init__(
             model_dir=self.model_dir,
             log_to_driver=self.need_to_log_to_driver,
             driver_ip=self.ip,
-            driver_port=self.port)
+            driver_port=self.port,
```
yushan111, line 158: Shall we rename `driver_port` to `driver_log_port`, `tcp_port` to `driver_tcp_store_port`, or other names you prefer?

python/orca/src/bigdl/orca/learn/pytorch/pytorch_ray_worker.py
```diff
@@ -81,10 +81,20 @@ def setup_address(self):
...
         """Returns the IP address of the current node."""
         return ray._private.services.get_node_ip_address()
 
+    def setup_distribute(self, rank, size, driver_ip, tcp_port):
+        import torch.distributed as dist
+        client_store = dist.TCPStore(driver_ip, tcp_port, rank, False)
+        super().setup_torch_distribute(tcp_store=client_store,
+                                       world_rank=rank,
+                                       world_size=size)
+
```
yushan111, line 97: Maybe move to torch_runner, so that it could be shared between ray and pyspark?

python/orca/src/bigdl/orca/learn/pytorch/pytorch_ray_estimator.py
```diff
@@ -167,11 +168,15 @@ def __init__(
 
             head_worker = self.remote_workers[0]
             address = ray.get(head_worker.setup_address.remote())
+            driver_ip = ray.get(head_worker.get_node_ip.remote())
+            tcp_port = ray.get(head_worker.get_free_port.remote())
```
yushan111, line 172: In worker,  since we may not need `setup_address` any more, we could change it to `get_ip_port`, which directly returns `ip` and `port`.
Then we could use `driver_ip, tcp_port = ray.get(head_worker.get_ip_port.remote())`

python/orca/src/bigdl/orca/learn/pytorch/pytorch_ray_worker.py
```diff
@@ -81,10 +81,20 @@ def setup_address(self):
         port = find_free_port()
         return f"tcp://{ip}:{port}"
 
+    def get_free_port(self):
+        return find_free_port()
+
```
yushan111, line 86: Maybe remove this?

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
```diff
@@ -154,7 +155,8 @@ def __init__(
             model_dir=self.model_dir,
             log_to_driver=self.need_to_log_to_driver,
             driver_ip=self.ip,
-            driver_port=self.port)
+            driver_log_port=self.port,
```
yushan111, line 158: `self.port` -> `self.log_port`

python/orca/src/bigdl/orca/learn/pytorch/pytorch_ray_estimator.py
```diff
@@ -166,12 +167,14 @@ def __init__(
             ])
 
             head_worker = self.remote_workers[0]
-            address = ray.get(head_worker.setup_address.remote())
+            driver_ip, driver_tcp_store_port = ray.get(head_worker.get_node_ip_port.remote())
 
-            logger.info(f"initializing pytorch process group on {address}")
+            _ = dist.TCPStore(driver_ip, driver_tcp_store_port, -1, True,
```
yushan111, line 172: maybe remove `_ =`?

sgwhat: It's needed.

sgwhat: > maybe remove `_ =`?

It has to exist.

yushan111: Sure

sgwhat:
Jenkins Test: http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-Orca-PR-Validation/786/

One job failed but was not related to this PR, also re-run pass.

PR #4859, opened 2022-06-16
pytorch_lightning 1.6 upgrade

MeouSker77:
now it works well in all tests


python/nano/src/bigdl/nano/deps/ipex/ipex_accelerator.py
```diff
@@ -15,55 +15,37 @@
...
-        model, optimizer = ipex.optimize(model, optimizer=self.optimizers[0],
-                                         inplace=True, dtype=dtype)
-        self.optimizers = [optimizer]
+    def __init__(self) -> None:
+        super().__init__()
+
+    @staticmethod
+    def is_available() -> bool:
+        """Detect if IPEX accelerator is available"""
+        if not check_avx512():
```
yangw1234, line 30: for ipex 1.11, they support running on machines without avx-512

MeouSker77: does ipex 1.9 support machines without avx-512 ?

yangw1234: > does ipex 1.9 support machines without avx-512 ?

No ipex 1.9 and 1.10 need avx-512 to run

MeouSker77: ok, fixed it

python/nano/src/bigdl/nano/deps/ray/ray_distributed.py
```diff
@@ -72,49 +79,103 @@ def get_env_vars(self, key: str):
...
+                (i, trainer, function, args, kwargs)
+            )
+            for i in range(strategy.num_workers)
+        ]
+
+        not_ready = futures
+        while not_ready:
+            ready, not_ready = ray.wait(not_ready, timeout=0)
+            ray.get(ready)
+        ray.get(ready)
```
yangw1234, line 133: I think these lines are equivalent to `ray.get(futures)`, right?

python/nano/src/bigdl/nano/pytorch/plugins/ddp_spawn.py
```diff
@@ -33,188 +33,241 @@
...
+               trainer: Optional["pl.Trainer"] = None, **kwargs: Any) -> Any:
+        # pytorch_lightning 1.6 uses this method to create child processes
+        os.environ["MASTER_PORT"] = str(self._strategy.cluster_environment.main_port)
+
+        if self._strategy.cpu_for_each_process is None:
+            cpu_procs = schedule_workers(self._strategy.num_processes)
+        else:
+            cpu_procs = self._strategy.cpu_for_each_process
+
+        # fix bug
```
yangw1234, line 83: maybe more explanation here? Fix which bug? And how do you fix this bug?

python/nano/src/bigdl/nano/deps/ray/ray_distributed.py
MeouSker77: these lines are copied from the original `execution_loop` method, but it seems they are redundant, I'll delete them

python/nano/src/bigdl/nano/pytorch/plugins/ddp_spawn.py
MeouSker77: ok, I'll add more explanation

yangw1234:
maybe adding a few unit tests for the bug you fixed

MeouSker77:
> maybe adding a few unit tests for the bug you fixed

ok

PR #4860, opened 2022-06-16
fix import error in IPEX Check

y199387:


python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
```diff
@@ -99,6 +99,8 @@ def __init__(self, num_processes: int = 1,
                     "Fall back to regular pytorch.")
             use_ipex = False
 
+        self.use_ipex = use_ipex
+
```
yangw1234, line 103: let's use `self.use_ipex` for rest of this function

python/nano/test/pytorch/utils/_train_ipex_callback.py
```diff
@@ -25,11 +25,10 @@
 
 class CheckIPEXCallback(Callback):
     def on_train_start(self, trainer, pl_module):
+        if trainer.use_ipex == False:
+           warnings.warn("CheckIPEXCallback is used, but ipex fail. ")
```
yangw1234, line 29: "but ipex is disabled"?

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
y199387: Done

python/nano/test/pytorch/utils/_train_ipex_callback.py
y199387: Done

y199387:
http://10.112.231.51:18888/job/BigDL-NanoTests-PR-Validation/238/

yangw1234:
LGTM

PR #4885, opened 2022-06-17
Nano : expose onnx trace setting to user

rnwang04:
In some cases, user may want to change the  opset_version to enable some complex operator in pytorch, we may expose the setting to users.
Also in quantize, we should add these parameters(https://github.com/intel-analytics/BigDL/issues/4893).


python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
```diff
@@ -367,6 +368,8 @@ def trace(model: nn.Module,
                             backend. 'openvino' and 'onnxruntime' are supported for now.
         :param onnxruntime_session_options: The session option for onnxruntime, only valid when
                                             accelerator='onnxruntime', otherwise will be ignored.
+        :param opset_version: opset_version for export_to_onnx, only valid when
+                                            accelerator='onnxruntime', otherwise will be ignored.
```
TheaperDeng, line 372: I think `accelerator="openvino"` will also be affected by this opset_version? no? @zhentaocc

zhentaocc: I suppose highest opset version should work.

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
```diff
@@ -354,7 +354,8 @@ def quantize(model,  # remove the type requirement for type checking
     def trace(model: nn.Module,
               input_sample=None,
               accelerator=None,
-              onnxruntime_session_options=None):
+              onnxruntime_session_options=None,
+              opset_version=11):
```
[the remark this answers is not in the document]
zhentaocc: why not use latest opset as  default?

python/nano/src/bigdl/nano/deps/onnxruntime/onnxruntime_api.py
```diff
@@ -15,7 +15,8 @@
 #
 
 
-def PytorchONNXRuntimeModel(model, input_sample=None, onnxruntime_session_options=None):
+def PytorchONNXRuntimeModel(model, input_sample=None,
+                            onnxruntime_session_options=None, **kwargs):
     """
         Create a ONNX Runtime model from pytorch.
```
TheaperDeng, line 22: add some comment to tell future developers what may be contained in kwargs?

TheaperDeng: same things to some of other changes as well

rnwang04: fix

rnwang04:
http://10.112.231.51:18888/job/BigDL-NanoTests-PR-Validation/274/
http://10.112.231.51:18888/job/BigDL-PRVN-chronos-Python-Spark-2.4-py37-ray-part1/1167/

PR #4889, opened 2022-06-20
Chronos: add Evaluator doc

Rebekk:


Rebekk:
https://evaluator-doc.readthedocs.io/en/latest/doc/PythonAPI/Chronos/evaluator.html

PR #4898, opened 2022-06-21
Add e2e dien inference

hkvision:
Will rename the file name in a following PR since if rename it here, the file changes won't be detected easily.


hkvision:
Some remarks:
- noclk sequence input is not used in the later inference stage and thus it is not included in the data processing.
- Padding is conducted for each batch using the longest sequence within this batch. FeatureTable can only do global padding and thus here padding is performed in prepare_data before the inference.
- In Jack's team's code, the records are sorted by length for every group of 10*batch_size; here I sort within each partition. This doesn't impact the inference result.
- For the model, given a batch of data, if one record in the batch is different, the inference result of other records would be slightly different as well. Thus the result acc and auc can't be exactly the same (can only be almost the same) with Jack's team unless the order of all records are the same. But the correctness is verfied to be OK.

hkvision:
http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-Friesian-PR-Validation/128/ one failure is irrelevant.

PR #4905, opened 2022-06-22
[nano] fix distributed training backend

Django-Jiang:
fix spawn model issue and add args check in trainer

### DIFF
--- a/python/nano/benchmark/pytorch/pytorch-cat-dog.py
+++ b/python/nano/benchmark/pytorch/pytorch-cat-dog.py
@@ -62,6 +62,7 @@ def create_data_loader(root_dir, batch_size, nproc):
     catdogs = ImageFolder(dir_path, data_transform)
     dataset_size = len(catdogs)
 
+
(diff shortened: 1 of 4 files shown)


yangw1234:
let's also enable this test for spawn https://github.com/intel-analytics/BigDL/blob/main/python/nano/test/pytorch/tests/test_plugin.py#L78

Django-Jiang:
> let's also enable this test for spawn https://github.com/intel-analytics/BigDL/blob/main/python/nano/test/pytorch/tests/test_plugin.py#L78

Test added

yangw1234:
@TheaperDeng can we enforce the checkpoint path to be true for now?

TheaperDeng:
> @TheaperDeng can we enforce the checkpoint path to be true for now?

yes we can, we are doing it in #4925

TheaperDeng:
> > @TheaperDeng can we enforce the checkpoint path to be true for now?
> 
> yes we can, we are doing it in #4925

done

yangw1234:
http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-NanoTests-PR-Validation/262/

Issue #4587, opened 2022-05-11
Fix `model` attribute in Orca Pytorch Callback


yushan111:
Currently we haven't called `set_model` in Orca Pytorch Estimator, leaving the value to be None.

PR #4907, opened 2022-06-22
Fix model attribute in Orca Pytorch Callback

yushan111:
This PR is to fix issue #4587


yushan111:
http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-Orca-PR-Validation/735/

Issue #4586, opened 2022-05-11
Enhance Pytorch Estimator Callbacks interface with `logs` input


yushan111:
Orca Pytorch Estimator Callbacks refers to [tf.keras callback](https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/Callback).

In tf.keras callback, the logs dictionary that callback methods take as argument will contain keys for quantities relevant to the current batch or epoch. For example, 
* In `on_epoch_end`, `logs` is a dictionary containing metric results for this training epoch, and for the validation epoch if validation is performed. Validation result keys are prefixed with val_. For training epoch, the values of the Model's metrics are returned. Example : {'loss': 0.2, 'accuracy': 0.7}.
* In `on_train_batch_ends`, `logs` is a dictionary containing aggregated metric results up until this batch.

We should also enable users using the `logs` argument in Orca Pytorch Estimator Callbacks to retrieve the training metric results within their own customized callbacks. 

Tasks:

- [ ] `logs` in `on_epoch_end`
- [ ] `logs` in  `on_train_batch_ends`

PR #4908, opened 2022-06-22
Enhance Pytorch Estimator Callbacks interface with logs input 

yushan111:
Issue #4586

### Interface change
This PR introduce `logs` in Pytorch Callbacks, as Tensorflow keras callback does, so that user could easily access the training status and evaluation results during the training process.

Specifically, this PR add `logs` support only for `on_epoch_end` and `on_train_end`.
For `on_xx_begin`, `logs` are not used. 
For `on_batch_end`, we may not support getting evaluation results via `logs` now,  since currently we only compute aggregated metrics after each training **epoch** ends, computing aggregated metrics on each **batch** ends would harm the overall performance. We could add support if requested by customers. 

```python

class CustomCallback(Callback):

    def on_train_end(self, logs=None):
        keys = list(logs.keys())
        model = self.model
        print("Stop training; got log keys: {}".format(keys))

    def on_epoch_end(self, epoch, logs=None):
        keys = list(logs.keys())
        print("End epoch {} of training; got log keys: {}".format(epoch, keys))
```

### DIFF
--- a/python/orca/src/bigdl/orca/learn/pytorch/callbacks/base.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/callbacks/base.py
@@ -60,7 +60,10 @@ class Callback(object):
         Subclasses should override for any actions to run. This function should only
         be called during TRAIN mode.
         @param epoch:  Integer, index of epoch.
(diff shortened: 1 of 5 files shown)


yushan111:
http://10.112.231.51:18888/job/BigDL-Orca-PR-Validation/736/

PR #4910, opened 2022-06-22
Chronos: add electricity_tcn use-case

Rebekk:


python/chronos/use-case/electricity/tcn.py
```diff
@@ -0,0 +1,80 @@
...
+from bigdl.chronos.data import TSDataset
+from sklearn.preprocessing import StandardScaler
+from bigdl.orca.automl.metrics import Evaluator
+from torch.utils.data.dataloader import DataLoader
+import torch
+import time
+import numpy as np
+
+if __name__ == '__main__':
+    torch.set_num_threads(1)
```
TheaperDeng, line 28: we can leave this setting just before performance(latency) test.

python/chronos/use-case/electricity/tcn.py
```diff
@@ -0,0 +1,80 @@
...
+                            input_feature_num = 321,
+                            output_feature_num = 321,
+                            num_channels = [30] * 7,
+                            repo_initialization = False,
+                            kernel_size = 3, 
+                            dropout = 0.1, 
+                            lr = 0.001,
+                            seed = 1)
+    forecaster.fit(train_loader, epochs=1, batch_size=32)
+
```
TheaperDeng, line 65: evaluate the forecaster, (train 30 epochs)

PR #4915, opened 2022-06-22
avoid installation of python 3.7.0 and typo

onurbarut:
python 3.7.0 and 3.7.1 is problematic with Ray (happens while orca context being init.)

### DIFF
--- a/docs/readthedocs/source/doc/Chronos/Overview/chronos.md
+++ b/docs/readthedocs/source/doc/Chronos/Overview/chronos.md
@@ -16,7 +16,7 @@ You can use _Chronos_ to do:
 
 Install `bigdl-chronos` from PyPI. We recommened to install with a conda virtual environment. To install Conda, please refer to https://docs.conda.io/en/latest/miniconda.html#.
 ```bash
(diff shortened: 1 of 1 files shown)


TheaperDeng:
Thank you for the contribution!

we are working on validating on what circumstances will  `conda create -n my_env python=3.7 setuptools=58.0.4` will install 3.7.0. It seems that typically, it will install latest 3.7.x (e.g. on my dev server). Installing latest 3.7.x is the ideal behavior for our installation cmd.

We will give some feedback once we have some finding.

TheaperDeng:
Thank you for the contribution, we have update our python supported version in user guide and we still want to maintain the conda env create cmd as before since we want to support more python versions.

Could you remove the changes to conda env create cmd and only leave the pyarrow installation typo? so that we can merge the PR.

TheaperDeng:
The reason why `conda create -n my_env python=3.7 setuptools=58.0.4` uses 3.7.0/3.7.1 is that users may have 3.7.0/3.7.1 downloaded in other env(e.g. base).

PR #4927, opened 2022-06-23
[WIP] Friesian online serving tutorial

cyita:


scala/friesian/scripts/setup_redis_master_replicas.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+set -e
+
+master_server='172.168.3.116'
```
qiuxin2012, line 4: It's better reading these IPs from parameters and providing a simple example to user.

PR #4932, opened 2022-06-23
add labelencoder, minmaxscaler for shard

dding3:
This PR adds label_encoder, minmax_scaler operation for sparkshard.

Usage will be like:
    data_shard.label_encoder(inputColumn="target", outputColumn="indexedTarget")
    
scaled_shard =\
        trans_data_shard.minmax_scaler(inputColumn=["hours", "awake"], outputColumn="x_scaled")


TheaperDeng:
this could be useful or a good reference for the implementation of `XShardsTSDataset.scale` @smurf-1119

python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -463,6 +466,63 @@ def zip(self, other):
...
+            return f
+
+        pd_rdd = df.rdd.mapPartitions(to_pdf(df.columns))
+        return pd_rdd
+
+
+    def label_encoder(self, inputColumn, outputColumn):
+        df = self.to_sparkdf()
+        from pyspark.ml.feature import StringIndexer
+        indexer = StringIndexer(inputCol=inputColumn, outputCol=outputColumn)
```
hkvision, line 501: Then in this case can it be merge with Friesian's feature_encode?
https://github.com/intel-analytics/BigDL/blob/main/python/friesian/src/bigdl/friesian/feature/table.py#L1000
Since StringIndexer seems to have poor performance when the dataset is large

hkvision:
@jason-dai Potentially if we add operations in this way, many operations will overlap with FeatureTable, any idea how we handle this?

yexinyinancy:
@dding3 It seems that for minmaxencoder, it only has fit_transform function, and it does not support transform function. Usually we need fit_transform for train set and transform for valid_set.

dding3:
> @dding3 It seems that for minmaxencoder, it only has fit_transform function, and it does not support transform function. Usually we need fit_transform for train set and transform for valid_set.

Nice catch. Will add one.

yexinyinancy:
@dding3 The return type of fit_transform function is pyspark.ml.linalg.densevector. However, for torch module, feature column requires type to be (tensor, dtype=torch.float32) and label columns to be (tensor, dtype=torch.long). Is it necessary to change the type in the transform function before output?

python/orca/example/data/spark_pandas.py
```diff
@@ -51,4 +51,13 @@ def process_feature(df, awake_begin=6, awake_end=23):
     trans_data_shard = data_shard.transform_shard(process_feature)
     data2 = trans_data_shard.collect()
 
+    from bigdl.orca.data.transform import *
+    scale = MinMaxScaler(inputCol=["hours", "awake"], outputCol="x_scaled")
+    data_shard = scale.fit_transform(trans_data_shard)
+    data_shard.rdd.count()
```
jason-dai, line 57: (1) change to `data_shard.compute()`
(2) why do you need to compute it here anyway?

python/orca/example/data/spark_pandas.py
```diff
@@ -51,4 +51,13 @@ def process_feature(df, awake_begin=6, awake_end=23):
...
+
+    encode = LabelEncode(inputCol="value", outputCol="encoded_value")
+    encoded_data_shard = encode.fit_transform(trans_data_shard)
+    encoded_data_shard.rdd.count()
```
jason-dai, line 61: (1) change to `data_shard.compute()`
(2) instead of `compute`, maybe better to show or save the results?

python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -463,6 +463,45 @@ def zip(self, other):
                               "The two SparkXShards should have the same number of elements "
                               "in each partition")
 
+    def to_sparkdf(self):
```
jason-dai, line 466: need to check if this is an XShards of Pandas Dataframe; otherwise report error

python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -463,6 +463,45 @@ def zip(self, other):
...
+            for pdf in iter:
+                return [pdf.columns.values]
+
+        rdd = self.rdd.mapPartitions(f)
+        column = self.rdd.mapPartitions(getSchema).first()
+        df = rdd.toDF(list(column))
+        return df
+
+    @staticmethod
+    def from_sparkdf(df):
```
jason-dai, line 482: `from_spark_df`

yexinyinancy:
When I used the latest codes, I still met some errors. Torch model askes the feature column to be `array` type while it is `list` type. The numbers in the feature column should be `np.float32` while numbers in the label column should be `np.long`. It is not supported now.

python/orca/example/data/spark_pandas.py
dding3: I call `count` here as I want to ensure the backend rdd is actually computed to ensure the new added function can succesfully run. I think by default `compute` is called during `__init__` in eager fashion, after we apply some transformation to the shards, we need actively call `compute` to trigger the action?

jason-dai: > I call `count` here as I want to ensure the backend rdd is actually computed to ensure the new added function can succesfully run. I think by default `compute` is called during `__init__` in eager fashion, after we apply some transformation to the shards, we need actively call `compute` to trigger the action?

There is no need to `compute` if you are creating a new `SparkXShards`

dding3: Make sense, will remove the `compute`

python/orca/src/bigdl/orca/data/transform.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def setInputOutputCol(self, inputCol, outputCol):
+        self.inputCol = inputCol
+        self.outputCol = outputCol
+        self.__createIndexer__()
+
+    def __createIndexer__(self):
+        invalidInputError(self.inputCol, "inputColumn cannot be empty")
+        invalidInputError(self.outputCol, "outputColumn cannot be empty")
+        self.indexer = SparkStringIndexer(inputCol=self.inputCol, outputCol=self.outputCol)
```
jason-dai, line 85: do not use StringIndexer

python/orca/src/bigdl/orca/data/transform.py
```diff
@@ -0,0 +1,99 @@
...
+            self.scaler = SparkMinMaxScaler(min=self.min, max=self.max,
+                                            inputCol=self.inputCol, outputCol=self.outputCol)
+
+    def setInputOutputCol(self, inputCol, outputCol):
+        self.inputCol = inputCol
+        self.outputCol = outputCol
+        self.__createScaler__()
+
+    def fit_transform(self, shard):
+        df = shard.to_spark_df()
```
jason-dai, line 55: instead of converting to/from Spark Dataframes, why cannot we simply use `rdd.map` to perform the scaling?

dding3:
> For `MinMaxScaler.fix_transform`, if the input of `inputCol` is a list of one string, it will lead to the following errors. It only works for a list of several strings. image

Fixed. Please try latest code

python/orca/src/bigdl/orca/data/transform.py
dding3: Yes, we can use `rdd.map`, and we need first calculate X.max(axis=0) and X.min(axis=0) which involve another two global operations, not sure if we can have better performance compared to spark `MinMaxScaler`

python/orca/src/bigdl/orca/data/shard.py
dding3: Fixed

python/orca/src/bigdl/orca/data/shard.py
dding3: Fixed

python/orca/src/bigdl/orca/data/transform.py
dding3: > When I used the latest codes, I still met some errors. Torch model askes the feature column to be `array` type while it is `list` type. The numbers in the feature column should be `np.float32` while numbers in the label column should be `np.long`. It is not supported now.

Changed `array` to `list` type. 
Could you elaborate more about "The numbers in the feature column should be `np.float32` while numbers in the label column should be `np.long`. It is not supported now.". Does that mean we need to convert feature column to be np.float32 and label to be np.long to make training success?

python/orca/src/bigdl/orca/data/transform.py
jason-dai: I think we already have the operations related to `stat`?

python/orca/src/bigdl/orca/data/transform.py
jason-dai: > do not use StringIndexer

We should use the string indexer implementation in Friesian

python/orca/src/bigdl/orca/data/transform.py
dding3: didn't find `min`, `max` supported in shards, but found seems friesian also use spark `minmaxscaler` internally?

hkvision:
Is this PR ready for merge?

dding3:
close this pr since it's duplicated to 
https://github.com/intel-analytics/BigDL/pull/5112
https://github.com/intel-analytics/BigDL/pull/5111
https://github.com/intel-analytics/BigDL/pull/5085

PR #4938, opened 2022-06-24
Rename Orca Tensorflow Estimator backend name from "tf2" to "ray"

yushan111:
Old PR #4393 will be closed


yushan111:
http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-Orca-PR-Validation/745/

PR #4939, opened 2022-06-24
[PPML] Fix bash error in occlum SGX_LOG_LEVEL

00-Luoshu:
* Modify the condition statement about SGX_LOG_LEVEL


qiyuangong (approved):
LGTM

PR #4945, opened 2022-06-27
fix: construct Url to retrieve primary key

liu-shaojun:
This submission is to construct the correct url to retrieve primary key.


glorysdj (approved):
LGTM

PR #4980, opened 2022-06-30
update visualization docs

shane-huang:


TheaperDeng:
suspect (did not find clear evidence), this PR will break list's render.
image

shane-huang:
> suspect (did not find clear evidence), this PR will break list's render. image

fixed the bullets in #4992

PR #4981, opened 2022-06-30
delete servlet-api in dllib's pom

qiuxin2012:
BigDL's init commit by yiheng. We can delete the direct dependency in our pom.
Then it will be a dependency of spark-core.


qiuxin2012:
jenkins passed

PR #4982, opened 2022-06-30
fix: parsing jsonObject error

liu-shaojun:
This submission is to fix the error `JSONObject["result"] is not a string.`
Annotation 2022-06-30 090027


scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/utils/HTTPUtil.scala
```diff
@@ -27,7 +27,7 @@ object HTTPUtil {
   def postRequest(url: String, postString: String): JSONObject = {
```
glorysdj, line 27: please add unit tests for these methods in HTTPUtil

glorysdj (approved):
LGTM

PR #4985, opened 2022-06-30
Serving and Friesian pom changes related to GPL license

Litchilitchy:


cyita:
Friesian jenkins http://10.112.231.51:18888/job/BigDL-Friesian-PR-Validation/136/

Issue #4984, opened 2022-06-30
[ORCA] Torch version for ray backend


yexinyinancy:
I meet some errors when I use torch.distributed in Orca following the [tutorial](https://github.com/intel-analytics/BigDL/blob/main/docs/readthedocs/source/doc/Orca/QuickStart/orca-pytorch-distributed-quickstart.md). The version of torch should be 1.8.1 to avoid such error.
![MicrosoftTeams-image](https://user-images.githubusercontent.com/51090416/176689568-457208a2-02a5-47b1-bcee-4358e64140bf.png)
![MicrosoftTeams-image (2)](https://user-images.githubusercontent.com/51090416/176689614-b0afc70c-52d7-4dd6-8abc-b41a0084473e.png)

hkvision:
@yushan111 @yangw1234 Do we have any torch version constraint? At least for 1.7.1 it doesn't work. Maybe we can check torch version when creating the Estimator?

Also BigDL backend requires torch 1.7.1 in the document, when users first try bigdl backend and fails, want to try other backends, they need to upgrade torch as well. Do you have any ideas on how to better handle this?

yushan111:
> At least for 1.7.1 it doesn't work. 

I failed to reproduce the issue with `torch=1.7.1`

hkvision:
> Fix for `torch=1.7.1` has been merged through PR (#4996).
> 
> Could you help verify whether the issue has been solved with the latest nightly built `bigdl-orca`?
> 
> If that works, `torch=1.7.1` remains a workable version of orca. We may not need a version check, or upgrade for PyTorch when changing backends.

Okay, but what about lower versions than 1.7.1? I think specifying a lower bound for torch version is still necessary?

yushan111:
> Okay, but what about lower versions than 1.7.1? I think specifying a lower bound for torch version is still necessary?

We don't automatically install torch when installing `bigdl-orca`. In our use-case/quick start documents, we direct users to install torch with a specified version (`pip install bigdl-orca[ray] torch==1.7.1 torchvision==0.8.2`).

yexinyinancy:
It now works for `torch=1.7.1`.

PR #4996, opened 2022-07-01
fix all_reduce for pytorch estimator while fitting with validation_data

yushan111:
Related issue: #4984


yushan111:
http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-Orca-PR-Validation/755/

PR #4997, opened 2022-07-01
Fix invalidInputError in Friesian examples and tests

hkvision:


hkvision:
http://10.112.231.51:18888/job/BigDL-PRVN-Python-Friesian-ExampleTest-Spark-2.4-feature+tf1-pip/500/console

PR #5022, opened 2022-07-05
Fix return type of hash encode

hkvision:
Should be int not string


hkvision:
http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-Friesian-PR-Validation/140/ jenkins passed

PR #5026, opened 2022-07-05
wrap BigDLEncrypt into a codec for en/decrypting csv/json data

qiuxin2012:
wrap BigDLEncrypt to codec for en/decrypting csv/json data.
encrypted data file's extension name is `cbc`. (temporary name)


qiuxin2012:
jenkins passed

PR #5029, opened 2022-07-06
Chronos: fix tsdataset.roll/to_torch_data_loader(time_enc=True) freq check bug

TheaperDeng:
**1. What is the change?**
This pr changes the policy to generate time encoding features when users call `time_enc=True` in `tsdataset.roll` and `tsdataset.to_torch_data_loader`

**2. Why the change?**
Bug report from customer. We found that the original policy (generate time features "smaller or equal" to the frequency) has some issue when it comes to some specific frequencies.
e.g. when freq is 2h, the original policy will not generate "hour of day" feature, which is wrong.

**3. Summary of the change (interface and design)**
interface is not changed in this bug fix, it remains
```python
# tsdata is a TSDataset
dataloader = tsdata.to_torch_data_loader(..., time_enc=True, ...)
data = tsdata.roll(..., time_enc=True, ...)
```

**4. Link for the design**
Not Applied

**5. How to test?**
- [x] use data with different freq to test the time enc dim

### DIFF
--- a/python/chronos/src/bigdl/chronos/data/utils/time_feature.py
+++ b/python/chronos/src/bigdl/chronos/data/utils/time_feature.py
@@ -121,30 +121,30 @@ def time_features_from_frequency_str(offset) -> List[TimeFeature]:
     """
 
     features_by_offsets = (
-        (Timedelta(seconds=1), [
+        (Timedelta(seconds=60), [
             SecondOfMinute,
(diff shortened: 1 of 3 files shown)


TheaperDeng:
http://10.112.231.51:18889/job/BigDL-Chronos-PR-Validation/527/

PR #5063, opened 2022-07-11
fix PPML scala ut failed due to parallel test

qiuxin2012:
Giving random a fixed seed can fix the failed test.


qiuxin2012:
jenkins scala test passed

PR #5065, opened 2022-07-11
[Nano] implement default LightningModule.validate and LightningModule.test

zhentaocc:
## Description
Current Nano inference model doesn't support lightning usages. Taking openvino as example, users can run inference in a Pytorch style like:
```python
ov_model = trainer.trace(model, accelerator='openvino')
y = ov_model(x)
```
But for lightning users, they may call:
```python
ov_model = trainer.trace(model, accelerator='openvino')
trainer.validate(ov_model, dataloader)
```
which will raise error since these functions are not implemented.
## Motivation and Context
Nano Pytorch is based on Pytorch-Lightning, it would be better to support all APIs provided by PL.  
As far as I know, we have four kinds of models requiring this feature:

- LightningModuleFromTorch
- PytorchOpenVINOModel
- PytorchONNXRuntimeModel
- PytorchQuantizedModel

In this PR, we provided default function implementations of train/validate/test/predict so that users don't have to implement these functions.
For `trainer.trace`, 
```python
ov_model = trainer.trace(model, accelerator='openvino')
trainer.validate(ov_model, dataloader)
trainer.test(ov_model, dataloader)
trainer.predict(ov_model, dataloader)
```
For `trainer.quantize`, 
```python
ov_model = trainer.quantize(model, accelerator='openvino')
trainer.validate(ov_model, dataloader)
trainer.test(ov_model, dataloader)
trainer.predict(ov_model, dataloader)
```
### How to infer number of inputs
For OpenVINO and ONNXRuntime, it's easy to get how many inputs in the graph. So we can run each forward step as:
```python
    def validation_step(self, batch, batch_idx):
        """Define a single validation step."""
        y_hat = self(*batch[:self._nargs])
```  
`self._nargs` is a property method to get how many inputs.

While for INC quantized model with Pytorh FX, we can't infer the number of inputs, so we skip the this by assigning `self._nargs` as -1.

## How was this PR tested?

Unit tests:

-  `trainer.validate`
-  `trainer.test`
-  `trainer.predict`

- [ ] jenkins passed (please provide link):


zhentaocc:
http://10.112.231.51:18889/job/BigDL-Chronos-PR-Validation/585/

PR #5070, opened 2022-07-11
upload Friesian example recall codes

devWangBin:


hkvision:
Thanks to the code contribution of @nyamashi :)

PR #5074, opened 2022-07-12
add ipex and ddp support for lightning-lite

MeouSker77:
## Description

This PR adds ipex, ddp and ray backend support for `LightningLite`.

## Motivation and Context

We want to support accelerating `LightningLite` with nano, so we should add ipex, ddp_spawn, ddp_subprocess and ray strategy support for `LightningLite`

## API Usage or Code Design

Add a nano's `LightningLite` to replace pytorch lightning's `LightningLite`.

See below for details.

## Related Link or Issue

analytics-zoo/nano#176

## How was this PR tested?

- [x] jenkins passed (please provide link): <http://10.112.231.51:18888/job/BigDL-NanoTests-PR-Validation/298/>

## License & Dependency
N/A

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/strategies/ddp_spawn.py
+++ b/python/nano/src/bigdl/nano/pytorch/strategies/ddp_spawn.py
@@ -38,6 +38,7 @@ import multiprocessing
 from typing import Any, List, Optional, Callable, Union, Dict
 
 import torch
+from torch import nn
 from torch import Tensor
 from torch.multiprocessing.spawn import _wrap, ProcessContext
 from torch.nn.parallel.distributed import DistributedDataParallel
@@ -75,8 +76,8 @@ class _DDPSpawnLauncher(_SpawnLauncher):
     def launch(self, function: Callable, *args: Any,
                trainer: Optional["pl.Trainer"] = None, **kwargs: Any) -> Any:
         # pytorch_lightning 1.6 uses this method to create child processes
-        invalidInputError(trainer is not None and self._strategy.cluster_environment is not None,
-                          'strategy.cluster_environment and trainer cannot be None')
+        invalidInputError(self._strategy.cluster_environment is not None,
+                          'strategy.cluster_environment cannot be None')
 
         os.environ["MASTER_PORT"] = str(self._strategy.cluster_environment.main_port)
 
@@ -97,7 +98,7 @@ class _DDPSpawnLauncher(_SpawnLauncher):
         # pytorch lightning 1.4 resets datamodule automatically before creating child processes,
         # so we do not need to do this, but 1.6 resets datamodule after creating child processes,
         # so we must reset datamodule here.
-        if self._strategy.use_ipex and TORCH_VERSION_LESS_1_10:
+        if self._strategy.use_ipex and TORCH_VERSION_LESS_1_10 and trainer is not None:
             # args[1] is dataloader, args[3] is datamodule when training,
             # and args[4] is datamodule when testing
             if isinstance(args[1], LightningDataModule):
@@ -141,6 +142,10 @@ class _DDPSpawnLauncher(_SpawnLauncher):
 
         # restore the state of child process
         spawn_output = return_queue.get()
+
+        if trainer is None:
+            return spawn_output
+
         self._recover_results_in_main_process(spawn_output, trainer)
         return spawn_output.trainer_results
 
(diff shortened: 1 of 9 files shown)


MeouSker77:
# Usage

## Non-distributed mode

### Using pytorch lightning's `LightningLite`

### Using nano's `LightningLite`

## Distributed mode

### Using pytorch lightning's `LightningLite`

### Using nano's `LightningLite`

yangw1234:
I think we also need
1. API doc
2. add a section in overview doc
3. a tutorial page and notebook

MeouSker77:
> I think we also need
> 
> 1. API doc
> 2. add a section in overview doc
> 3. a tutorial page and notebook


maybe we can add these in a follow up PR? this PR is already large enough

yangw1234:
sure

> > I think we also need
> > 
> > 1. API doc
> > 2. add a section in overview doc
> > 3. a tutorial page and notebook
> 
> maybe we can add these in a follow up PR? this PR is already large enough

sure

yangw1234:
Does this PR depend on PyTorch-Lightning 1.6.4? Or it works on both 1.4 and 1.6?

python/nano/src/bigdl/nano/pytorch/strategies/ddp_spawn.py
```diff
@@ -75,8 +74,8 @@ def __init__(self, strategy: 'DDPSpawnStrategy') -> None:   # type: ignore[overr
     def launch(self, function: Callable, *args: Any,
                trainer: Optional["pl.Trainer"] = None, **kwargs: Any) -> Any:
         # pytorch_lightning 1.6 uses this method to create child processes
-        invalidInputError(trainer is not None and self._strategy.cluster_environment is not None,
-                          'strategy.cluster_environment and trainer cannot be None')
+        invalidInputError(self._strategy.cluster_environment is not None,
+                          'strategy.cluster_environment cannot be None')
```
yangw1234, line 78: Maybe adding in comments in which case the `strategy.cluster_environment` will be None? 

And also telling what the users should do if he/she meets this error in the error message (if he should not see this error, maybe we can say "you should raise an issue in BigDL.").

python/nano/src/bigdl/nano/pytorch/strategies/ddp_spawn.py
```diff
@@ -141,6 +140,10 @@ def launch(self, function: Callable, *args: Any,
 
         # restore the state of child process
         spawn_output = return_queue.get()
+
+        if trainer is None:
```
yangw1234, line 144: Maybe adding in comments when the `trainer` will be `None`

python/nano/src/bigdl/nano/pytorch/strategies/ddp_spawn.py
```diff
@@ -230,18 +233,10 @@ def setup(self, trainer: "pl.Trainer") -> None:
         # which is used by ipex==1.9, so we move this line here
         self.model_to_device()
 
-    def configure_ddp(self):
```
yangw1234, line 233: Could you explain a little bit why changing here?

python/nano/src/bigdl/nano/pytorch/strategies/ipex/ipex_strategy.py
```diff
@@ -57,12 +57,12 @@ def setup(self, trainer: pl.Trainer) -> None:
 
         :param trainer: the trainer instance
         """
-        self.setup_precision_plugin()
```
yangw1234, line 60: Could you explain a little bit why changing here?

python/nano/src/bigdl/nano/pytorch/strategies/ray/ray_distributed.py
```diff
@@ -127,25 +122,27 @@ def launch(self, function: Callable, *args: Any,
...
 
         results = ray.get(futures)  # type: ignore
 
+        if trainer is None:
+            return results[0]
+
         # Get the results, checkpoint path, and model weights from worker 0.
         results, best_path, state_dict = results[0]  # type: ignore
-        strategy._model.load_state_dict(state_dict)
+        strategy.model.load_state_dict(state_dict)
```
yangw1234, line 137: is this a bug in the original code?

python/nano/src/bigdl/nano/pytorch/strategies/ddp_spawn.py
MeouSker77: in fact, the cluster_environment cannot be None in any situations

python/nano/src/bigdl/nano/pytorch/strategies/ddp_spawn.py
MeouSker77: We need tto change the creation of `DistributedDataParallel`

in the default implementation of pl 1.6, the trainer will call `configure_ddp`, and `configure_ddp` will call `_setup_model` to create a DistributedDataParallel, but LightningLite will call `_setup_model` directly to create a DistributedDataParallel.

we choose to override the `configure_ddp` and create DistributedDataParallel directly in it before, but this method makes no effect when using LightningLite, so we choose to override the `_setup_model`, which will be used in both cases.

python/nano/src/bigdl/nano/pytorch/strategies/ipex/ipex_strategy.py
MeouSker77: there are some strange things here: when using ddp_spawn or ddp_subprocess, if we call `trainer.test()`, then the `len(self.optimizers)` will be 0, so we need to handle this case.

here we don't take this case into consideration before, so I changed it to handle this case, but the strange thing is that when using IPEX strategy, the `len(self.optimizers)` is still 1 when we call `trainer.test()`.

so it seems this change is not necessary.

in addition, use `super().setup()` to replace `self.setup_precision_plugin()` and `self.setup_optimizers()` for reliability

python/nano/src/bigdl/nano/pytorch/strategies/ray/ray_distributed.py
MeouSker77: not a bug, the `model` refer to `_model`, they are the same, change it for consistency

python/nano/src/bigdl/nano/pytorch/strategies/ddp_spawn.py
MeouSker77: added it

yangw1234:
we have refactored the github action workflow, maybe you should rebase on main branch

yangw1234:
LGTM

PR #5076, opened 2022-07-12
add shard keras tutorial

dding3:


dding3:
Most jekins pass. http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-Orca-PR-Validation/874/

jason-dai:
1) Need to add description about this example; why do we need it?

2) Need to describe new user API in the description too

PR #5077, opened 2022-07-12
add pytorch shard tutorials

dding3:


python/orca/example/shard_pytorch_regression.py
```diff
@@ -0,0 +1,93 @@
...
+init_orca_context(memory="4g")
+
+path = 'new_ionosphere.csv'
+data_shard = bigdl.orca.data.pandas.read_csv(path)
+
+def getSchema(iter):
+    for pdf in iter:
+        return [pdf.columns.values]
+
+column = data_shard.rdd.mapPartitions(getSchema).first()
```
jason-dai, line 74: do not user rdd operations

dding3:
Fixed in https://github.com/intel-analytics/BigDL/pull/5258

PR #5088, opened 2022-07-13
Chronos: reduce mem allocation in TCN example on electricity dataset

plusbang:
Modified for test_loader and MSE calculation process


TheaperDeng (approved):
LGTM

PR #5095, opened 2022-07-13
[PPML] Helm to submit ppml job

liu-shaojun:
### 1. Why the change?

In BigDL PPML, there are two ways to launch application: ppml cli and helm chart. This submission is to update helm chart to invoke ppml cli to launch application. Before this change, users needed to prepare a script used to submit Spark job (./submit-spark-k8s.sh) and put it in the NFS directory. Now we directly invoke ppml cli in bigdl-ppml-helm/template/submit-job.yaml, no need to prepare submit-spark-k8s.sh, users only need to update configs in bigdl-ppml-helm/values.yml, then


ppml/trusted-big-data-ml/python/docker-graphene/kubernetes/bigdl-ppml-helm/values.yaml
```diff
@@ -1,5 +1,7 @@
 image: your_ppml_image
 k8sMaster: k8s://https://master_ip:master_port
+jar: your_jar
+class: your_main_class
```
ColossusChang, line 4: Thanks for adding these two. I'm pretty surprised I forgot to add them.

ColossusChang (approved):
LGTM

glorysdj (approved):
LGTM

Issue #5096, opened 2022-07-14
Chronos : can't reproduce paper's result as lack of scheduler


rnwang04:
I can't reproduce paper's result on the setting of Autoformer 97-720 forecast.
I find the reason is current implementation of Autoformer lack of a lr scheduler comparing with the source code.

PR #5097, opened 2022-07-14
Chronos : add scheduler for Autoformer

rnwang04:
## Description

add scheduler for Autoformer to fix https://github.com/intel-analytics/BigDL/issues/5096

### DIFF
--- a/python/chronos/src/bigdl/chronos/model/autoformer/Autoformer.py
+++ b/python/chronos/src/bigdl/chronos/model/autoformer/Autoformer.py
@@ -179,7 +179,9 @@ class AutoFormer(pl.LightningModule):
         return outputs
 
     def configure_optimizers(self):
-        return getattr(optim, self.optim)(self.parameters(), lr=self.lr)
+        optimizer = getattr(optim, self.optim)(self.parameters(), lr=self.lr)
+        scheduler = torch.optim.lr_scheduler.MultiStepLR(optimizer, milestones=[3,4,5,6,7,8,9,10], gamma=0.5, verbose=True)
+        return [optimizer], [scheduler]
 
 
 def model_creator(config):


python/chronos/src/bigdl/chronos/model/autoformer/Autoformer.py
```diff
@@ -179,7 +179,9 @@ def predict_step(self, batch, batch_idx):
         return outputs
 
     def configure_optimizers(self):
-        return getattr(optim, self.optim)(self.parameters(), lr=self.lr)
+        optimizer = getattr(optim, self.optim)(self.parameters(), lr=self.lr)
+        scheduler = torch.optim.lr_scheduler.MultiStepLR(optimizer, milestones=[3,4,5,6,7,8,9,10], gamma=0.5, verbose=True)
```
TheaperDeng, line 183: I think we should allow users to set scheduler / milestones (I think this might be a better choice). Furthermore ,they may be allowed to disable scheduler.

rnwang04: done

python/chronos/src/bigdl/chronos/forecaster/autoformer_forecaster.py
```diff
@@ -49,6 +49,7 @@ def __init__(self,
                  optimizer="Adam",
                  loss="mse",
                  lr=0.0001,
+                 lr_scheduler_milestones=[3, 4, 5, 6, 7, 8, 9, 10],
                  metrics=["mse"],
                  seed=None,
                  distributed=False,
```
TheaperDeng, line 55: add api doc (i.e. `:param lr_scheduler_milestones: xxxx`) for this parameter

TheaperDeng (commented):
others LGTM

python/chronos/src/bigdl/chronos/forecaster/autoformer_forecaster.py
rnwang04: done

PR #5101, opened 2022-07-14
fix PPMLContext python api can't read encrypted file

PatrickkZ:
## Description
the way of encrypt/decrypt file changes, this cause PPMLContext python api can't decrypt files. this pr fix the problem.

### 1. Why the change?
to fix PPMLContext python api can't read encrypted file

### 2. Summary of the change
1.change the way to create a PPMLContext in python, init a SparkContext in python code, not in scala code anymore.
2.update UT code

### 3. How to test?
- [ ] N/A
- [x] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/python/ppml/src/bigdl/ppml/ppml_context.py
+++ b/python/ppml/src/bigdl/ppml/ppml_context.py
@@ -30,7 +30,7 @@ def check(ppml_args, arg_name):
 
 
 class PPMLContext(JavaValue):
-    def __init__(self, app_name, ppml_args=None):
+    def __init__(self, app_name, ppml_args=None, master="local", deploy_mode="client", **kwargs):
         self.bigdl_type = "float"
         conf = {"spark.app.name": app_name,
                 "spark.hadoop.io.compression.codecs": "com.intel.analytics.bigdl.ppml.crypto.CryptoCodec"}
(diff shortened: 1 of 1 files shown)


scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/PPMLContext.scala
```diff
@@ -193,6 +193,46 @@ object PPMLContext{
...
+   * init ppml context with an existed SparkSession
+   * @param sparkSession a SparkSession
+   * @return a PPMLContext
+   */
+  def initPPMLContext(sparkSession: SparkSession): PPMLContext = {
+    val conf = sparkSession.sparkContext.getConf
+    val kmsType = conf.get("spark.bigdl.kms.type", defaultValue = "SimpleKeyManagementService")
+    val kms = kmsType match {
+      case KMS_CONVENTION.MODE_EHSM_KMS =>
+        val ip = conf.get("spark.bigdl.kms.ehs.ip", defaultValue = "0.0.0.0")
```
qiuxin2012, line 206: Throw an error if this ip is not in conf. And also below port, key, id, etc.  This should be set in python side.

PR #5103, opened 2022-07-14
Nano: Add support for channels_last

y199387:
## Description
This pr Add support for automatically converting model and data's memory_format to channels_last.
If user instantiate `nano.pytorch.Tainer` with `channels_last=True`，the model will be converted to channels_last and its hook function `on_before_batch_transfer` will be replaced at `setup` stage. The replaced `on_before_batch_transfer` function will convert every batch to data with `memory_format=channels_last`.

### 1. Why the change?
Channels Last improves the throughput of convolution operations in networks for computer vision. Using channels_last and ipex at the same time can get better acceleration effect. It's 
necessary to provide user a simple way to apply `torch.channels` to both of the model object and data.
https://github.com/analytics-zoo/nano/issues/170

### 3. How to test?
- [ ] Unit test


y199387:
related docs and source
https://pytorch-lightning.readthedocs.io/en/stable/api/pytorch_lightning.core.hooks.DataHooks.html#pytorch_lightning.core.hooks.DataHooks.on_before_batch_transfer
https://github.com/Lightning-AI/lightning/blob/1.6.4/pytorch_lightning/core/hooks.py#L715

python/nano/test/pytorch/tests/test_channels_last.py
```diff
@@ -0,0 +1,225 @@
...
+        x = torch.flatten(x, 1)
+        x = self.fc(x)
+        x = self.head(x)
+
+        return x
+
+class ConvModel(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.conv1 = torch.nn.Conv2d(2, 1, (1, 2), bias=False)
```
y199387, line 63: For Convolutional layer with only one channel, the layer.param.is_contiguous() & layer.param.is_contiguous(memory_format=torch.channels_last) is True after the model is converted to channels_last.

yangw1234:
you may also rebase on main since we updated the github actions workflows

yangw1234:
http://10.112.231.51:18889/view/BigDL-PR-Validation/job/ZOO-PR-NanoTests/323/

Issue #5102, opened 2022-07-14
Windows bigdl-nano[pytorch] issues


cyita:
- [ ] pip install bigdl-nano[pytorch] will install 0.14.0xxxx instead of 2.1.0xxxx
- [ ] unsetenv is not supported on windows, https://github.com/intel-analytics/BigDL/blob/main/python/nano/src/bigdl/nano/pytorch/__init__.py#L22
```bash
if 'KMP_INIT_AT_FORK' in os.environ:
    del os.environ['KMP_INIT_AT_FORK']
```
- [ ] intel_extension_for_pytorch is not available on windows.

torch issues on windows
- [ ] may encounter dll load issues in jupyter notebook, https://github.com/pytorch/pytorch/issues/4518

PR #5106, opened 2022-07-15
Fix nano pytorch import issue on windows

cyita:
## Description

Fix nano pytorch import issue on windows since unsetenv is not supported on windows.

Related issue: #5102


TheaperDeng (approved):
lgtm, please fix some conflict and check the github action before merging

PR #5112, opened 2022-07-15
support Stringindexer in sparkxshards operation

dding3:
### 1. Why the change?

support Stringindexer in sparkxshards operation

### 2. Summary of the change

1. add Stringindexer, the usage is like:
    encode = StringIndexer(inputCol="value")
    encoded_data_shard = encode.fit_transform(trans_data_shard)
2. the implementation is convert shards to spark df, then reuse the encode_category api in friesian to do the indexing, then convert spark df back to shards
3. since friesian depends on orca, to avoid circle dependency, move the related encode_category code from friesian to orca

### 3. How to test?
Application test
UT


dding3:
most jekins pass. http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-Orca-PR-Validation/816/

PR #5115, opened 2022-07-18
[Nano]Make OpenVINOModel as public API

zhentaocc:
## Description

Allow users to use `deps/openvino` directly without Pytorch/Tensorflow. And add extra `device` option for users to specify the device.  If users want to load an openvino model without Pytorch/TF installed, current usage is like
```python
from bigdl.nano.deps.openvino.core.model import OpenVINOModel
ov_model = OpenVINOModel("model.xml")
```
From our previous design, `deps/xxx_api` is the only thing in the dependencies we expose to users. So we need to provide one 
`OpenVINOModel` api.

### 1. Why the change?

Mainly for the AI for workforce course.  
And after this change, users can load openvino model by `OpenVINOModel` without Pytorch/TF installed.

### 2. Summary of the change

-  make `OpenVINOModel` accessible to users
### Before
If users want to load an openvino model without Pytorch/TF installed, they may do this:
```python
from bigdl.nano.deps.openvino.core.model import OpenVINOModel
ov_model = OpenVINOModel("model.xml")
```
### After
```python
ov_model = OpenVINOModel("model.xml", device='cpu')
```
- add `device` option to `OpenVINOModel` in case the model is run on other type of devices

### 3. How to test?
- [x] N/A

### DIFF
--- a/python/nano/src/bigdl/nano/deps/openvino/core/model.py
+++ b/python/nano/src/bigdl/nano/deps/openvino/core/model.py
@@ -26,8 +26,9 @@ from .utils import save
 
 
 class OpenVINOModel:
-    def __init__(self, ie_network: str):
+    def __init__(self, ie_network: str, device='CPU'):
         self._ie = Core()
+        self._device = device
         self.ie_network = ie_network
 
     def forward_step(self, *inputs):
@@ -43,7 +44,8 @@ class OpenVINOModel:
             self._ie_network = self._ie.read_model(model=str(model))
         else:
             self._ie_network = model
-        self._compiled_model = self._ie.compile_model(model=self.ie_network, device_name='CPU')
+        self._compiled_model = self._ie.compile_model(model=self.ie_network,
+                                                      device_name=self._device)
         self._infer_request = self._compiled_model.create_infer_request()
 
     def _save_model(self, path):
--- a/python/nano/src/bigdl/nano/deps/openvino/openvino_api.py
+++ b/python/nano/src/bigdl/nano/deps/openvino/openvino_api.py
@@ -48,3 +48,9 @@ def KerasOpenVINOModel(model, input_sample=None):
     """
     from .tf.model import KerasOpenVINOModel
     return KerasOpenVINOModel(model)
+
+
+def OpenVINOModel(model, device='CPU'):
+    from .core.model import OpenVINOModel
+    return OpenVINOModel(model, device)
+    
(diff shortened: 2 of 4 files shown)


jason-dai:
Is this change visible to the user? It's not clear from the description.

zhentaocc:
> Is this change visible to the user? It's not clear from the description.

I updated the description. `OpenVINOModel` class is always accessible to users actually. But from our design, everything open to users should be in `openvino_api`.

jason-dai:
I don't think we actually expose `nano.deps.xxx` to the users before; the user always use `nano.tf` or `nano.pytorch` which in turn calls `nano.deps.xxx_api`; we may add `nano.openvino` which then calls `nano.deps.openvino_api`

zhentaocc:
> I don't think we actually expose `nano.deps.xxx` to the users before; the user always use `nano.tf` or `nano.pytorch` which in turn calls `nano.deps.xxx_api`; we may add `nano.openvino` which then calls `nano.deps.openvino_api`

Good idea, I will try this.

zhentaocc:
http://10.112.231.51:18889/view/BigDL-PR-Validation/job/ZOO-PR-NanoTests/321/  
http://10.112.231.51:18888/job/ZOO-PR-nano-pytorch-ipex-test/

jason-dai:
> > I don't think we actually expose `nano.deps.xxx` to the users before; the user always use `nano.tf` or `nano.pytorch` which in turn calls `nano.deps.xxx_api`; we may add `nano.openvino` which then calls `nano.deps.openvino_api`
> 
> Good idea, I will try this.

Please update the description on how the user should use this feature

zhentaocc:
updated the description. @jason-dai

jason-dai:
> updated the description. @jason-dai

Then why do export other interface such as PyTorchOpenVINOModel, etc. in `nano.openvino`?

python/nano/src/bigdl/nano/deps/openvino/core/model.py
```diff
@@ -26,8 +26,9 @@
 
 
 class OpenVINOModel:
-    def __init__(self, ie_network: str):
+    def __init__(self, ie_network: str, device='CPU'):
         self._ie = Core()
+        self._device = device
```
qiuxin2012, line 31: how many devices do we support? You should make sure that, this device is valid.

zhentaocc: OpenVINO itself will make sure of that. Is that necessary for us to check this again? 
Supported devices https://docs.openvino.ai/latest/openvino_docs_OV_UG_supported_plugins_Supported_Devices.html

zhentaocc:
> > updated the description. @jason-dai
> 
> Then why do we export other interface such as PyTorchOpenVINOModel, etc. in `nano.openvino`?

resolve in https://github.com/intel-analytics/BigDL/pull/5170

PR #5127, opened 2022-07-19
fix tf inc metric format

y199387:
## Description
  Fix Nano TensorFlow INC metric bug.

related issue
https://github.com/intel-analytics/BigDL/issues/5305

### 3. How to test?
- [x] Unit test
- [x] Jenkins
http://10.112.231.51:18889/job/BigDL-Chronos-PR-Validation/653/

### DIFF
--- a/python/nano/src/bigdl/nano/deps/neural_compressor/tensorflow/metric.py
+++ b/python/nano/src/bigdl/nano/deps/neural_compressor/tensorflow/metric.py
@@ -23,7 +23,7 @@ class TensorflowINCMetric(BaseINCMetric):
         # calculate accuracy
         preds = tf.stack(preds)
         labels = tf.stack(labels)
-        return preds, labels
+        return labels, preds
 
     def to_scalar(self, tensor):
         return tensor.numpy()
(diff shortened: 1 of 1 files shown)


yangw1234:
@zhentaocc could you help review this?

zhentaocc:
What's the bug?

y199387:
> What's the bug?

According to Keras API doc, y_true should be first and y_pred after.
The current code may make the accuracy calculation wrong.
https://keras.io/api/metrics/accuracy_metrics/

python/nano/src/bigdl/nano/deps/neural_compressor/tensorflow/metric.py
```diff
@@ -21,9 +21,9 @@ class TensorflowINCMetric(BaseINCMetric):
     def stack(self, preds, labels):
 
         # calculate accuracy
-        preds = tf.stack(preds)
         labels = tf.stack(labels)
-        return preds, labels
+        preds = tf.stack(preds)
+        return labels, preds
```
zhentaocc, line 26: I think you should override the `result` function instead.
```suggestion
        preds = tf.stack(preds)
        labels = tf.stack(labels)
        return preds, labels

    def result(self):
        # calculate accuracy
        preds, labels = self.stack(self.pred_list, self.label_list)
        accuracy = self.metric(labels, preds)
        return self.to_scalar(accuracy)
```

python/nano/src/bigdl/nano/deps/neural_compressor/tensorflow/metric.py
```diff
@@ -21,9 +21,16 @@ class TensorflowINCMetric(BaseINCMetric):
     def stack(self, preds, labels):
 
         # calculate accuracy
+        labels = tf.stack(labels)
```
zhentaocc, line 24: This line is redundant.

PR #5128, opened 2022-07-19
fix ray test in nano

MeouSker77:
## Description

the default value of `num_processes` argument of `Trainer` is 1, and the trainer won't use distributed backend when `num_processes==1`, so we need to set it

in addition, fix the bug that the test of ray backend doesn't set the `PYTHONPATH` environment variable and ray backend doesn't pass this environment variable to subprocesses


yangw1234:
LGTM

PR #5129, opened 2022-07-19
Chronos: support unscale_xshards in XShardsTSDataset

smurf-1119:
## Description

add `xshards.unscale_xshards` and test function `tests.test_xshardstsdataset_unscale_numpy`.
initialize scaler_index and update scaler_index in xshards.roll.

### 1. Why the change?

help user unscale the xshards data containing numpy array.

### 2. Summary of the change

add `xshards.unscale_xshards` and test function `tests.test_xshardstsdataset_unscale_numpy`.
initialize scaler_index and update scaler_index in `xshards.roll`.
change the order of [feature_col, target_col] in `xshards.scale` and `xshards.unscale`.

### 3. How to test?
- [x] Unit test http://10.112.231.51:18889/job/BigDL-Chronos-PR-Validation/581/

### DIFF
--- a/python/chronos/src/bigdl/chronos/data/experimental/xshards_tsdataset.py
+++ b/python/chronos/src/bigdl/chronos/data/experimental/xshards_tsdataset.py
@@ -22,6 +22,7 @@ from bigdl.chronos.data.utils.roll import roll_timeseries_dataframe
 from bigdl.chronos.data.utils.impute import impute_timeseries_dataframe
 from bigdl.chronos.data.utils.split import split_timeseries_dataframe
 from bigdl.chronos.data.experimental.utils import add_row, transform_to_dict
+from bigdl.chronos.data.utils.scale import unscale_timeseries_numpy
 
 
 _DEFAULT_ID_COL_NAME = "id"
--- a/python/chronos/src/bigdl/chronos/data/experimental/xshards_tsdataset.py
+++ b/python/chronos/src/bigdl/chronos/data/experimental/xshards_tsdataset.py
@@ -41,6 +41,7 @@ class XShardsTSDataset:
         self.dt_col = schema["dt_col"]
         self.feature_col = schema["feature_col"].copy()
         self.target_col = schema["target_col"].copy()
+        self.scaler_index = [i for i in range(len(self.target_col))]
 
         self.numpy_shards = None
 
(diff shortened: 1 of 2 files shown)


python/chronos/src/bigdl/chronos/data/experimental/xshards_tsdataset.py
```diff
@@ -364,6 +364,33 @@ def _inverse_transform(df, id_col, scaler, feature_col, target_col):
...
+            from sklearn.utils.validation import check_is_fitted
+            from bigdl.nano.utils.log4Error import invalidInputError
+
+            id = data['id'][0, 0]
+            scaler_for_this_id = scaler[id]
+            invalidInputError(not check_is_fitted(scaler_for_this_id),
+                              "scaler is not fitted. When calling scale for the first time, "
+                              "you need to set fit=True.")
+
+            return unscale_timeseries_numpy(data['y'], scaler_for_this_id, scaler_index)
```
TheaperDeng, line 385: I think the key need to be set by users, since there are 2 possibilities: 'prediction' and 'y'

python/chronos/src/bigdl/chronos/data/experimental/xshards_tsdataset.py
```diff
@@ -364,6 +364,33 @@ def _inverse_transform(df, id_col, scaler, feature_col, target_col):
...
+
+        def _get_features(df):
+            return df.columns
+
+        cols = self.shards.transform_shard(_get_features).collect()[0]
+        scaler_index = [cols.get_loc(col) + 1 for col in self.target_col]
```
TheaperDeng, line 391: I think +1 is kind of like a ad-hoc patch for your specific test data. The real problem is:
1. You can never know if target_col(e.g. "value") is ahead of feature_col(e.g. "extra_features") or not
2. The order in your scaler is [feature_col, target_col]
3. That's why the order is wrong!


Solution is:
1. change the order in your scaler to [target_col, feature_col] (just like what we do in TSDataset)
2. get the scaler_index in roll rather than unscale (just like what we do in TSDataset)
3. scaler_index can be set to `[self.target_col.index(feature_col[i]) for i in range(len(feature_col))] ` (just like what we do in TSDataset)

python/chronos/src/bigdl/chronos/data/experimental/xshards_tsdataset.py
```diff
@@ -364,6 +364,33 @@ def _inverse_transform(df, id_col, scaler, feature_col, target_col):
                                                   self.target_col)
         return self
 
+    def unscale_xshards(self, data):
+        '''
```
TheaperDeng, line 368: add a parameter "key" here, default to None, in that case:
1. priority is "prediction" > "y"
2. if no "prediction" or "y" return an error and require our users to input a key

TheaperDeng:
add some description to this PR as well

TheaperDeng (approved):
LGTM, please run jenkins

PR #5137, opened 2022-07-20
Abort using dllib's method to init SparkContext

PatrickkZ:
## Description
change the way to initialize a SparkContext in PPMLContext python api, instead of using dllib's "init_spark_on_xxx" method to initialize a SparkContext, now directly creating a SparkContext with --conf parameters.

### 1. Why the change?

create a SparkContext directly with --conf parameters is more flexible.

### 2. User API changes

N/A

### 3. Summary of the change 

change the way to initialize a SparkContext in PPMLContext python api, instead of using dllib's "init_spark_on_xxx" method to initialize a SparkContext, now directly creating a SparkContext with --conf parameters.

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...


jason-dai:
Why the change? It's not clear from the description.

Issue #5090, opened 2022-07-13
Make RayContext init lighter


hkvision:
Reported by our customer.
https://github.com/intel-analytics/BigDL/blob/main/python/orca/src/bigdl/orca/ray/ray_on_spark_context.py#L449
In RayContext init, it is a bit heavy. Since RayContext instance will always be created when init_orca_context, we'd better make the init lighter, so that for users that don't use ray, there wouldn't be so much contraint.

As the customer reports, if the spark environment uses dynamicAllocation, there's no spark.exectutor.cores, even if ray is not used, it will report the error `Exception: spark.executor.cores not detected in the SparkContext, you need to manually specify num_ray_nodes and ray_node_cpu_cores for RayOnSparkContext to start ray services` and need to specify ray_node_cpu_cores as a workaround, which may not be reasonable.

hkvision:
Also need to add dynamic allocation to known issues or FAQ.

PR #5148, opened 2022-07-20
[Orca] Make RayContext init lighter

lalalapotter:
## Description
This PR is to solve #5090. In order to have a lighter RayContext initialization, we do not set up the RayOnSparkContext Instance when init OrcaRayContext in init_orca_context, instead we set up it when set up ray cluster. Thus, the init process of RayContext in init_orca_context become lighter for users who do not use the ray cluster.

### DIFF
--- a/python/orca/src/bigdl/orca/ray/ray_on_spark_context.py
+++ b/python/orca/src/bigdl/orca/ray/ray_on_spark_context.py
@@ -541,7 +541,7 @@ class RayOnSparkContext(object):
         if self.initialized:
             print("The Ray cluster has been launched.")
         else:
-            self.activateRayOnSparkContext()
+            self.setupRayOnSparkContext()
             if self.is_local:
                 if self.env:
                     os.environ.update(self.env)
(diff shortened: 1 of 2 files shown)


python/orca/src/bigdl/orca/ray/ray_on_spark_context.py
```diff
@@ -397,11 +397,17 @@ def __init__(self, sc, redis_port=None, password="123456", object_store_memory=N
                 self.system_config = self.extra_params.pop("_system_config")
         self.include_webui = include_webui
         self._address_info = None
+        self.redis_port = random.randint(10000, 65535) if not redis_port else int(redis_port)
+        self.ray_node_cpu_cores = ray_node_cpu_cores
+        self.num_ray_nodes = num_ray_nodes
+        RayOnSparkContext._active_ray_context = self
+
+    def setUpRayOnSparkContext(self):
```
hkvision, line 405: rename to "setup"?

hkvision (approved):

lalalapotter:
PR Validation Results:

- Orca-PR-Validation-Spark-3.1
- Orca-PR-Validation-Spark-2.4
- Orca-PRVN-Others

PR #5150, opened 2022-07-21
fix bugs of ray backend and subprocess backend

MeouSker77:
## Description

- change the step of restoring subprocess' states in ray backend: we restore states manually before, which causes some states cannot be restored correctly, such as the `trainer.callback_metrics`, now we change it to pl's method.
- fix the bug of ddp_subprocess backend, see comments in codes for detail


MeouSker77:
http://10.112.231.51:18888/job/BigDL-NanoTests-PR-Validation/312/

yangw1234:
LGTM

PR #5170, opened 2022-07-22
[Nano] Remove redundant exposed apis and add basic openvino unit tests

zhentaocc:
## Description

This PR is to fix some unresolved issues in https://github.com/intel-analytics/BigDL/pull/5115

1. As mentioned in https://github.com/intel-analytics/BigDL/pull/5115, we should expose `bigdl.nano.openvino.OpenVINOModel` only for basic openvino usage.
2. Add basic unit tests w\o pytorch/tf installed.

### 1. Why the change?

1. expose Pytorch/TF APIs in `bigdl.nano.openvino` is not that reasonable.
2. We should have UTs to make sure openvino work fine w\o pytorch/tf.

### 2. User API changes

N/A

### 3. Summary of the change 

1. Only expose `OpenVINOModel` in `bigdl.nano.openvino`
2. Add UTs for `OpenVINOModel` usage

### 4. How to test?
- [x] Unit test

### DIFF
--- a/.github/workflows/nano_unit_tests_basic.yml
+++ b/.github/workflows/nano_unit_tests_basic.yml
@@ -1,4 +1,4 @@
-name: Nano Unit Tests PyTorch
+name: Nano Unit Tests Basic
 
 # Controls when the action will run. 
 on:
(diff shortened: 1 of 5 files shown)


zhentaocc:
http://10.112.231.51:18888/job/BigDL-NanoTests-PR-Validation/324/  
http://10.112.231.51:18888/job/BigDL-PRVN-chronos-Python-Spark-2.4-py37-ray-part2/1195/

PR #5171, opened 2022-07-22
fix the issue that child process can't inherit parent process's `sys.path`

MeouSker77:
## Description

- fix the issue that child process can't inherit parent process's `sys.path`
- add a test covering this bug


yangw1234:
LGTM

PR #5173, opened 2022-07-22
change the API of LightningLite

MeouSker77:
### 1. Why the change?

We need our interfaces to accelerate the custom training loop with as few changes as possible

### 2. User API changes

the API before:
```python
from bigdl.nano.pytorch.lite import LightningLite
...
class Lite(LightningLite):
    def run(...):
        ...
        # add these lines
        model, optimizer = self.setup(model, optimizer)
        dataloaders = self.setup_dataloaders(dataloaders)
        ...
        self.backward(loss)
        ...

Lite(...).run(...)
```
the API now:
```python
from bigdl.nano.pytorch import TorchNano
...
class MyNano(TorchNano):
   def train(...):
      ... 
      # add this line
      model, optimizer, train_dataloader, loss_func = self.setup(model, optimizer, train_dataloader, loss_func)
      ...
      self.backward(loss)
      ...

MyNano(...).train(...)
```

### 3. Summary of the change 

Change the user API of LightningLite support

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...


jason-dai:
Need to add description, especially the new API

MeouSker77:
custom training loop:
```python
class MyModel(nn.Module):
    ...

model = ...
optimizer = ...
train_dataloader = ...
loss_func = ...
model.train()

for epoch in range(num_epochs):
    ...
```

code using our NanoLite:
```python
class MyModel(nn.Module):
    ...

from bigdl.nano.pytorch.lite import NanoLite

class MyLite(NanoLite):
    def train(...):
        model = ...
        optimizer = ...
        train_dataloader = ...
        loss_func = ...

        # add this line
        model, optimizer, train_dataloader, loss_func = self.setup_all(model, optimizer, train_dataloader, loss_func)

        model.train()

        for epoch in range(num_epochs):
            ...

MyLite(use_ipex=True, strategy="spawn").train(...)
```

only two steps are needed:
1. derive from our NanoLite class and override its train method
2. copy all training codes before and add one line

jason-dai:
> custom training loop:
> 
> ```python
> class MyModel(nn.Module):
>     ...
> 
> model = ...
> optimizer = ...
> train_dataloader = ...
> loss_func = ...
> model.train()
> 
> for epoch in range(num_epochs):
>     ...
> ```
> 
> code using our NanoLite:
> 
> ```python
> class MyModel(nn.Module):
>     ...
> 
> from bigdl.nano.pytorch.lite import NanoLite
> 
> class MyLite(NanoLite):
>     def train(...):
>         model = ...
>         optimizer = ...
>         train_dataloader = ...
>         loss_func = ...
> 
>         # add this line
>         model, optimizer, train_dataloader, loss_func = self.setup_all(model, optimizer, train_dataloader, loss_func)
> 
>         model.train()
> 
>         for epoch in range(num_epochs):
>             ...
> 
> MyLite(use_ipex=True, strategy="spawn").train(...)
> ```
> 
> only two steps are needed:
> 
> 1. derive from our NanoLite class and override its train method
> 2. copy all training codes before and add one line
> 
> how about changing to this API? @jason-dai @yangw1234

1. Please update the "Description" of this PR at the top
2. How about the API below?

```
from bigdl.nano.pytorch import TorchNano

class MyNano(TorchNano):
   def train(...):
      ... 
      # add this line
      model, optimizer, train_dataloader, loss_func = self.setup(model, optimizer, train_dataloader, loss_func)
      ....

MyNano(use_ipex=True, strategy="spawn").train(...)
```

MeouSker77:
> 2\. How about the API below?

I think it's ok

MeouSker77:
http://10.112.231.51:18888/job/BigDL-NanoTests-PR-Validation/326/

yangw1234:
What is the type requirement of the `loss_func` ?

Is it possible that the user does not have a `loss_func` in their code?

MeouSker77:
> What is the type requirement of the `loss_func` ?

the `loss_func` needs to be `_Loss`, which is from pytorch and used as all loss function's base class, but in fact, we only require it can be called directly by `()`

> Is it possible that the user does not have a `loss_func` in their code?

in the current code, user must have a `loss_func`.

we can make some changes to remove this requirement, but if user does not hava a `loss_func`, he needs to replace `loss.backward()` with `self.backward(loss)` mannually.

MeouSker77:
our current API requires a `loss_func` and modifies it, so the API can be:
```python
from bigdl.nano.pytorch import TorchNano

class MyNano(TorchNano):
   def train(...):
      ... 
      # add this line
      model, optimizer, dataloaders, loss_func = self.setup(model, optimizer, dataloaders, loss_func)
      ....
      loss = loss_func(...)
      loss.backward()
      ...

MyNano(use_ipex=True, strategy="spawn").train(...)
```

but if user does not have a `loss_func`, he must replace `loss.backward()` with `self.backward(loss)` manually.

considering this situation, then the API will be:

```python
from bigdl.nano.pytorch import TorchNano

class MyNano(TorchNano):
   def train(...):
      ... 
      # add this line
      model, optimizer, dataloaders = self.setup(model, optimizer, dataloaders)
      ....
      loss = ...
      self.backward(loss)    # modify this line
      ...

MyNano(use_ipex=True, strategy="spawn").train(...)
```

@jason-dai

python/nano/src/bigdl/nano/pytorch/lite/lite.py
```diff
@@ -46,7 +49,7 @@ def __init__(self, num_processes: int = 1,
                  strategy: Union[str, Strategy] = "subprocess",
```
jason-dai, line 49: need to specify what the valid values are for `strategy` in the API doc

jason-dai:
> our current API requires a `loss_func` and modifies it, so the API can be:
> 
> ```python
> from bigdl.nano.pytorch import TorchNano
> 
> class MyNano(TorchNano):
>    def train(...):
>       ... 
>       # add this line
>       model, optimizer, dataloaders, loss_func = self.setup(model, optimizer, dataloaders, loss_func)
>       ....
>       loss = loss_func(...)
>       loss.backward()
>       ...
> 
> MyNano(use_ipex=True, strategy="spawn").train(...)
> ```
> 
> but if user does not have a `loss_func`, he must replace `loss.backward()` with `self.backward(loss)` manually.
> 
> considering this situation, then the API will be:
> 
> ```python
> from bigdl.nano.pytorch import TorchNano
> 
> class MyNano(TorchNano):
>    def train(...):
>       ... 
>       # add this line
>       model, optimizer, dataloaders = self.setup(model, optimizer, dataloaders)
>       ....
>       loss = ...
>       self.backward(loss)    # modify this line
>       ...
> 
> MyNano(use_ipex=True, strategy="spawn").train(...)
> ```
> 
> @jason-dai

I don't think there is an easy way to enforce `self.backward(loss)` in the user code, yes?

What is the type of `loss_func`? Is it easy for the user to create the `loss_func` from `loss`?

python/nano/src/bigdl/nano/pytorch/lite/lite.py
```diff
@@ -16,11 +16,15 @@
```
jason-dai, line 16: change the file path to `pytorch/torch_nano.py`

MeouSker77:
> I don't think there is an easy way to enforce `self.backward(loss)` in the user code, yes?
> 
> What is the type of `loss_func`? Is it easy for the user to create the `loss_func` from `loss`?

the `loss_func` should be a callable object, it takes a `input` Tensor and a `target` Tensor, then compute the `loss` Tensor, it should return the `loss` Tensor, then user call `loss.backward()`  to do backpropagation

but this process has a lot limits, user may not compute loss in this way, or they may use the `loss` Tensor to get a new loss Tensor and call `backward()` in the new Tensor.

current API cannot handle these situations, so we decide to remove the requirement of `loss_func`, and require user to replace `loss.backward()` in the raw pytorch code with `self.backward(loss)`

jason-dai:
> require user to replace `loss.backward()` in the raw pytorch code with `self.backward(loss)`
How do you enforce that? What if the user does not call `self.backward(loss)`?

MeouSker77:
> How do you enforce that? What if the user does not call `self.backward(loss)`?

It seems that we can only add this requirement in the document and tutorial.

If user does not call `self.backward(loss)`, then in most situations the training loop can still run normally, only in some special situations it will meet some errors, these situations can be found in LightingLite's `backward` method

jason-dai:
> > How do you enforce that? What if the user does not call `self.backward(loss)`?
> 
> It seems that we can only add this requirement in the document and tutorial.
> 
> If user does not call `self.backward(loss)`, then in most situations the training loop can still run normally, only in some special situations it will meet some errors, these situations can be found in LightingLite's `backward` method

Is there a way to automatically check if `setup` and `backward` are called in the `train` method?

MeouSker77:
> Is there a way to automatically check if `setup` and `backward` are called in the `train` method?

I suddenly found if user only use our strategies ('spawn', 'subprocess', 'ray'), he will not meet the special situations that requires `self.backward(loss)`, i.e. if user only use our strategies, he has no need to replace `loss.backward()` with `self.backward(loss)`.

The only method maybe checking source code in the runtime, but it is not reliable.

Our API only changes the running process before and after the `train` method, if user does not call our `setup` method, once the `train` method starts running, we cannot do anything until it ends, because the codes of `train` method are all user's codes. 
So we can only require user to call `setup` in the `train` method, so that we can do something, but we have no method to enforce users to do this or check whether they called `setup`.

The `self.backward` call is no need if user only uses our strategies.

yangw1234 (approved):

PR #5175, opened 2022-07-22
Chronos: add more installation options

TheaperDeng:
## Description
A quick way to understand new installation options is https://bigdl-junweid.readthedocs.io/en/installation-option-chronos/doc/Chronos/Overview/chronos.html

Please refer to detailed design and discussions in: https://github.com/intel-analytics/BigDL/issues/4791

### 4. How to test?
- [x] Manual test
- [x] Unit test http://10.112.231.51:18889/job/BigDL-Chronos-PR-Validation/662/
- Application test will be done after this PR is merged

PR #5183, opened 2022-07-23
Chronos: An MVP docker for single node with build-args

zpeng1898:
### 1. Why the change?

To help users who pursue a quick use of Chronos build a docker image with dependencies installed and unitest examples passed. Users can build with `build-args` so as to customize their installation mode.

### 2. Summary of the change

Add a directory `BigDL/docker/chronos-args` with a `Dockerfile`, `install-python-env.sh`, and `README.md`.

### 3. How to test?

Go into the directory `BigDL/docker/chronos-args` and follow `README.md` in it.


TheaperDeng:
Will hold this PR until #5175 is merged

PR #5187, opened 2022-07-25
Chronos: error message is added for tsdataset with less than 1 rolling sample

plusbang:
## Description

A error message is added to .to_torch_data_loader()

(http://10.112.231.51:18889/view/BigDL-PR-Validation/job/BigDL-Chronos-PR-Validation/616/)

### DIFF
--- a/python/chronos/src/bigdl/chronos/data/tsdataset.py
+++ b/python/chronos/src/bigdl/chronos/data/tsdataset.py
@@ -781,8 +781,9 @@ class TSDataset:
                 need_dflen = self.lookback + max(self.horizon)
             if len(self.df) < need_dflen:
                 invalidInputError(False,
-                                  "The length of the dataset must be larger than "
-                                  "'lookback' plus 'horizon'! ")
+                                  "The length of the dataset must be larger than the sum "
+                                  "of lookback and horizon, while get lookback+horizon="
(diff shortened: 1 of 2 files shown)


python/chronos/src/bigdl/chronos/data/tsdataset.py
```diff
@@ -774,6 +774,16 @@ def to_torch_data_loader(self,
...
                               "'gen_global_feature' and 'gen_rolling_feature' methods.")
+
+            if isinstance(self.horizon, int):
+                need_dflen = self.lookback + self.horizon
+            else:
+                need_dflen = self.lookback + max(self.horizon)
+            if len(self.df) < need_dflen:
+                invalidInputError(False,
+                                  "The length of the dataset must be larger than "
+                                  "'lookback' plus 'horizon'! ")
```
TheaperDeng, line 785: f"The length of the dataset must be larger than the sum of lookback and horizon, while get lookback+horizon={lookback+horizon} and length of dataset={len(self.df)}"

python/chronos/src/bigdl/chronos/data/tsdataset.py
```diff
@@ -774,6 +774,16 @@ def to_torch_data_loader(self,
...
+
```
TheaperDeng, line 786: add a ut for this error

TheaperDeng (approved):

PR #5193, opened 2022-07-25
[PPML] Update occlum SparkPi example config

hzjane:
## Description

The config of occlum example PI is out of date. Need to update.

### 1. Why the change?
To update the right config.

### 2. User API changes
No.

### 3. Summary of the change 
Change the config and use proper value.

### 4. How to test?

- [x] Application test


qiyuangong (approved):
LGTM

Issue #5192, opened 2022-07-25
Orca: Training process is stuck for a long time then executors connect timed out


lalalapotter:
**Describe the bug**
The training process stuck and executors show connect timed out when I run [example](https://github.com/intel-analytics/BigDL/tree/main/python/orca/example/learn/pytorch/fashion_mnist) on yarn cluster.

```
[E socket.cpp:793] [c10d] The client socket has timed out after 1800s while trying to connect to (172.16.0.152, 34771).
Traceback (most recent call last):
  File "fashion_mnist.py", line 207, in <module>
2022-07-25 15:59:11,173 WARNING worker.py:1245 -- (ip=172.16.0.107) The agent on node Almaren-Node-107 failed to be restarted 5 times. There are 3 possible problems if you see this error.
  1. The dashboard might not display correct information on this node.
  2. Metrics on this node won't be reported.
  3. runtime_env APIs won't work.
Check out the `dashboard_agent.log` to see the detailed failure messages.
    main()
  File "fashion_mnist.py", line 190, in main
    backend=args.backend)
  File "/home/kai/anaconda3/envs/orca-zcg/lib/python3.7/site-packages/bigdl/orca/learn/pytorch/estimator.py", line 95, in from_torch
    log_level=log_level)
  File "/home/kai/anaconda3/envs/orca-zcg/lib/python3.7/site-packages/bigdl/orca/learn/pytorch/pytorch_ray_estimator.py", line 173, in __init__
    dist.constants.default_pg_timeout)
TimeoutError: The client socket has timed out after 1800s while trying to connect to (172.16.0.152, 34771).
```

**To reproduce**

- Environment
  - bigdl-orca[ray]>=2.1.0b20220707
- Scripts

```
export HADOOP_CONF_DIR=the directory of the hadoop and yarn configurations

python fashion_mnist.py --cluster_mode yarn --backend ray
```

lalalapotter:
@sgwhat Maybe you can refer to this PR #4858

yushan111:
@sgwhat 
I have also reproduced the issue. Could you try testing on yarn-client mode again?

sgwhat:
> 

Sure.

sgwhat:
> @sgwhat I have also reproduced the issue. Could you try testing on yarn-client mode again?

Sure.

yushan111:
[...]

PR #5195, opened 2022-07-26
[Orca] Replace Head Worker in PyTorch Ray Estimator 

sgwhat:
## Description

Replace the head worker with the current client node.

### 1. Why the change?

To fix https://github.com/intel-analytics/BigDL/issues/5192. Users may get errors when there are existing ray clusters in Yarn cluster.

### 2. User API changes

N/A

### 3. Summary of the change 
1. Remove head worker processing in `pytorch_ray_worker`.
2. Use the `node_ip_address` and `free_port` in the client node (instead of the original head worker) to create TCPStore (server).  

### 4. How to test?
- [x] Unit test
- [x] Application test on Yarn Client / Cluster
- [x] Application test on K8s Client / Cluster.

### DIFF
--- a/python/orca/src/bigdl/orca/learn/pytorch/pytorch_ray_estimator.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/pytorch_ray_estimator.py
@@ -90,6 +90,7 @@ def partition_refs_to_creator(partition_refs):
 
     return data_creator
 
+
 def get_node_ip():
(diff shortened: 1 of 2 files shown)


yushan111 (approved):
LGTM

sgwhat:
Jenkins test: http://10.112.231.51:18888/job/BigDL-Orca-PR-Validation/856/, one failed but re-run pass.

PR #5197, opened 2022-07-26
Chronos: add scale and unscale to spark df example

TheaperDeng:
## Description
add scale and unscale to the spark df example with xshardstsdataset.
The rmse and mae is the same as non-distributed forecaster with tsdataset.'
fix some bug in xshardstsdataset.

### 4. How to test?
- [ ] Application test


TheaperDeng:
http://10.112.231.51:18889/view/BigDL-PR-Validation/job/BigDL-Chronos-PR-Validation/613/

PR #5202, opened 2022-07-26
[DLlib] XGBoost Criteo example update

CharleneHu94:
## Description

1. Update timing in example 
2. Update XGBoost example README

### 1. Why the change?

1. For more accurate time breakdown between phases
2. `nthread` should be no larger than `spark.task.cpus` according to https://github.com/dmlc/xgboost/blob/34408a7fdcebc0e32142ed2f52156ea65d813400/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala#L138

### 2. User API changes

None

### 3. Summary of the change 

Modify data read time and add model save time, modify README

### 4. How to test?
- [ ] Application test
- [ ] Document test
- [ ] ...


qiyuangong (approved):
LGTM

PR #5203, opened 2022-07-26
PPML FGBoost server model save support

Litchilitchy:
### 1. Why the change?

To support FLServer to save Aggregator model so that server could load and use trained model

### 2. User API changes

user has to specify `fgBoostServerModelPath` in `ppml-conf.yaml`

### 3. Summary of the change 

* a `FLConfig` is created for each aggregator, it could store `modelPath` so far, and other config needed by aggregator could be added in future
* at the end of each boosted round, the server will save the treeleaves info, aka. server model, to the path specified in the config

### 4. How to test?
added a unittest to save and load on server side


jason-dai:
Why nn tutorial is also included in the PR?

Litchilitchy:
> Why nn tutorial is also included in the PR?

some mistake... will fix

Litchilitchy:
fixed

PR #5207, opened 2022-07-26
Revise github landing page

shane-huang:
## Description

Revise github landing page
preview at https://github.com/shane-huang/BigDL/blob/github-landing/README.md


shane-huang:
@jason-dai added flow chart for how to choose lib in BigDL.

PR #5211, opened 2022-07-26
Orca: set conf to None when using spark-submit

lalalapotter:
## Description
This PR is to fix the issues report by customers when using spark-submit to submit tasks. If we use `conf` in `init_orca_context`, the `conf` will not be convert to `SparkConf` when using spark-submit and lead to exception. In this PR, we check the `conf`  before using it as `SparkConf` to avoid `no attribute 'getAll()'` exception.

### DIFF
--- a/python/orca/src/bigdl/orca/common.py
+++ b/python/orca/src/bigdl/orca/common.py
@@ -252,7 +252,7 @@ def init_orca_context(cluster_mode=None, runtime="spark", cores=2, memory="2g", 
                 from bigdl.dllib.nncontext import init_nncontext
                 if "conf" in spark_args and spark_args["conf"] is not None:
                     warnings.warn("For spark-submit cluster_mode, all conf should be "
-                                  + "specified in the spark-submit command , but got "
+                                  + "specified in the spark-submit command, but got "
                                   + repr(spark_args["conf"]) + ", ignored", Warning)
                     spark_args["conf"] = None
                 sc = init_nncontext(**spark_args)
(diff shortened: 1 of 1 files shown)


python/orca/src/bigdl/orca/common.py
```diff
@@ -250,6 +250,10 @@ def init_orca_context(cluster_mode=None, runtime="spark", cores=2, memory="2g",
                 sc = init_spark_on_local(cores, **spark_args)
             elif cluster_mode == "spark-submit":
                 from bigdl.dllib.nncontext import init_nncontext
+                if "conf" in spark_args and spark_args["conf"] is not None:
+                    warnings.warn("For Spark-submit mode, conf should be None, but got "
```
hkvision, line 254: Spark-submit mode -> spark-submit cluster_mode

hkvision: change to `all conf should be specified in the spark-submit command`

lalalapotter:
PR Validation Results:

[Passed](http://10.112.231.51:18888/job/BigDL-Orca-PR-Validation/868/), several test failed at first validation and passed when rerunning the test.

python/orca/src/bigdl/orca/common.py
lalalapotter: fixed in new commit

hkvision (approved):

python/orca/src/bigdl/orca/common.py
```diff
@@ -250,6 +250,11 @@ def init_orca_context(cluster_mode=None, runtime="spark", cores=2, memory="2g",
                 sc = init_spark_on_local(cores, **spark_args)
             elif cluster_mode == "spark-submit":
                 from bigdl.dllib.nncontext import init_nncontext
+                if "conf" in spark_args and spark_args["conf"] is not None:
+                    warnings.warn("For spark-submit cluster_mode, all conf should be "
+                                  + "specified in the spark-submit command , but got "
```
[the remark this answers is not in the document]
lalalapotter: fixed

PR #5224, opened 2022-07-27
Chronos: How-to-Guides (How to train forecaster)

Chaselzxy:
## Description

Add How-to-Guides, and help users train forecaster.

### 1. Why the change?

Make our doc more friendly.

### 2. User API changes

No changed.

### 3. Summary of the change 
https://howtodoc.readthedocs.io/en/how_to/doc/Chronos/Howto/how_to_train_forecaster_on_one_node.html

### 4. How to test?
- [ ] Document test

### 5. New dependencies


TheaperDeng:
We can follow such overview for this page, currently this draft is kind of like an API doc :)

Maybe use a jupyter notebook and transform to html could be a good choice, since we may need to show some output to our users.

# Train Forecaster on one node

A paragraph quickly illustrate

1. Why we need to train a forecaster? (to learn the pattern (like the period, scale...) in history data)

2. Chronos only support CPU training (and acceleration on Intel CPU) for training,. Since time series data is not large, CPU training is enough for most cases.

## Data preparation

We support 3 data representaitons, briefly illustrate what they are and a short code-snip how to generate them.

1. TSDataset

2. Numpy Ndarray

3. pytorch dataloader / tf dataset

## Training

show them how to use `fit`. A short code-snip would do.

### validation

show them a code snip how to use "best_epoch" and use a paragraph to illustrate why validation is important.

### automatically acceleration

leave this blank to me

PR #5227, opened 2022-07-27
Chronos: Lazy load Forecaster

liangs6212:
## Description

LazyImport does not load the required module immediately, it only loads the module when it is called, makes module loading more friendly.

### 1. Why the change?

The coupling between modules is too high, the first load time is too long, and all modules will be loaded.
Not all modules need to be loaded when using Installation Options.

### 2. User API changes

No changed

### 3. Summary of the change 

Optimized forecaster startup time, Runs 100% faster.
Increase flexibility in writing code and reduce coupling.

### 4. How to test?
- [ ] Unit test
- [ ] Application test

### 5. API usage

```python
# 1. will use __call__
LSTMForecaster = LazyImport('bigdl.chronos.forecaster.LSTMForecaster')
forecaster = LSTMForecaster(input_feature_num=1,
                                               output_feature_num=1,
                                               past_seq_len=48)


# 2. will use __getattr__
TCNForecaster = LazyImport('bigdl.chronos.forecaster.TCNForecaster')
tcn = TCNForecaster.from_tsdataset(tsdataset. ...)
```

### 6. Known issues
- [x] LazyImport imports may import multiple times
- [x] LazyImport does not throw errors and warnings
- [x] LazyImport may import all Forecasters, making the message unfriendly.
- [x] other possible changes: remove all warnings.
- [x] The form bigdl.chronos.forecaster.lstm_forecaster.LSTMForecaster is not accepted because LSTMForecaster is a class
- [x] Migration Comments and remove warnings.warn.

### DIFF
--- a/python/chronos/src/bigdl/chronos/forecaster/__init__.py
+++ b/python/chronos/src/bigdl/chronos/forecaster/__init__.py
@@ -61,7 +61,6 @@ if not orca_available:
     warnings.warn("Please install `bigdl-orca` to use full collection of forecasters.")
 
 # import forecasters
-bigdl = LazyImport('bigdl')
 PREFIXNAME = 'bigdl.chronos.forecaster.'
 if torch_available:
     LSTMForecaster = LazyImport(PREFIXNAME+'lstm_forecaster.LSTMForecaster')
--- a/python/chronos/src/bigdl/chronos/forecaster/__init__.py
+++ b/python/chronos/src/bigdl/chronos/forecaster/__init__.py
@@ -16,6 +16,7 @@
 
 import warnings
 import logging
+import importlib
 from bigdl.chronos.utils import LazyImport
 # unset the KMP_INIT_AT_FORK
 # which will cause significant slow down in multiprocessing training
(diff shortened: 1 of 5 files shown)


python/chronos/src/bigdl/chronos/utils.py
```diff
@@ -30,3 +31,33 @@ def wrapped(*args, **kwargs):
...
+    def __init__(self, module_name, pkg=None):
+        self.module_name = module_name
+        self.pkg = pkg
+
+    def __getattr__(self, name):
+        try:
+            module = importlib.import_module(self.module_name, package=self.pkg)
+            mod = getattr(module, name)
+        except:
+            spec = importlib.util.find_spec(str(self.module_name + '.' + name))
```
TheaperDeng, line 53: On what circumstances this `except` will be effective, I don't think we may find this "..LSTMForecaster.predict" or some thing like that

python/chronos/src/bigdl/chronos/forecaster/__init__.py
```diff
@@ -64,16 +65,17 @@ def __exit__(self, *args, **opts):
     warnings.warn("Please install `bigdl-orca` to use full collection of forecasters.")
 
 # import forecasters
+PREFIX_PATH = 'bigdl.chronos.forecaster'
 if torch_available:
-    from .lstm_forecaster import LSTMForecaster
-    from .tcn_forecaster import TCNForecaster
-    from .seq2seq_forecaster import Seq2SeqForecaster
-    from .nbeats_forecaster import NBeatsForecaster
+    LSTMForecaster = LazyImport('..LSTMForecaster', pkg=PREFIX_PATH+'.lstm_forecaster')
```
TheaperDeng, line 70: I think these 2 dots is used to avoid error in `LazyImport`'s `__call__`, I think we can further enhance the implementation of `LazyImport` rather than use some walkaroud when we call it.

python/chronos/src/bigdl/chronos/forecaster/__init__.py
```diff
@@ -64,16 +65,17 @@ def __exit__(self, *args, **opts):
...
+    NBeatsForecaster = LazyImport('..NBeatsForecaster', pkg=PREFIX_PATH+'.nbeats_forecaster')
     if orca_available:
-        from .tcmf_forecaster import TCMFForecaster
+        TCMFForecaster = LazyImport('..TCMFForecaster', pkg=PREFIX_PATH+'.tcmf_forecaster')
 if tf_available:
-    from .tf.mtnet_forecaster import MTNetForecaster
+    MTNetForecaster = LazyImport('..MTNetForecaster', pkg=PREFIX_PATH+'.tf.mtnet_forecaster')
 if prophet_available:
-    from .prophet_forecaster import ProphetForecaster
+    ProphetForecaster = LazyImport('..ProphetForecaster', pkg=PREFIX_PATH+'.ProphetForecaster')
```
TheaperDeng, line 79: `+'.prophet_forecaster'`

python/chronos/src/bigdl/chronos/forecaster/__init__.py
```diff
@@ -64,16 +65,17 @@ def __exit__(self, *args, **opts):
...
 if arima_available:
-    from .arima_forecaster import ARIMAForecaster
+    ARIMAForecaster = LazyImport('..ARIMAForecaster', pkg=PREFIX_PATH+'.ARIMAForecaster')
```
TheaperDeng, line 81: `+'.arima_forecaster'`

python/chronos/src/bigdl/chronos/utils.py
```diff
@@ -30,3 +31,33 @@ def wrapped(*args, **kwargs):
...
+            mod = importlib.util.module_from_spec(spec)
+            spec.loader.exec_module(mod)
+        return mod
+
+    def __call__(self, *args, **kwargs):
+        function_name = self.module_name.split('.')[-1]
+        module_name = self.module_name.split(f'.{function_name}')[0]
```
TheaperDeng, line 60: I think you may just want to first check if there is a function name (by checking if there is a `.` in module_name) and then devide into 2 paths.
1. function_name is valid, we will follow the same code here.
2. function_name is not valid, we will just `return module(*args, **kwargs)`

python/chronos/src/bigdl/chronos/forecaster/__init__.py
```diff
@@ -64,16 +65,17 @@ def __exit__(self, *args, **opts):
     warnings.warn("Please install `bigdl-orca` to use full collection of forecasters.")
 
 # import forecasters
+PREFIX_PATH = 'bigdl.chronos.forecaster'
 if torch_available:
```
TheaperDeng, line 69: check `torch_available` and many other variables by `importlib.util.find_spec`

python/chronos/src/bigdl/chronos/utils.py
liangs6212: you are right. `..LSTMForecaster` was originally designed to solve the problem of relative path import, but will cause the above problem, and now it has all been imported with absolute path.

python/chronos/src/bigdl/chronos/forecaster/__init__.py
liangs6212: In the previous answer, the pkg parameter is also preserved in `LazyImport`, but will not be used.

python/chronos/src/bigdl/chronos/forecaster/__init__.py
liangs6212: fixed

python/chronos/src/bigdl/chronos/forecaster/__init__.py
liangs6212: fixed

python/chronos/src/bigdl/chronos/forecaster/__init__.py
liangs6212: Except for `tensorflow` and `prophet`, other modules have been replaced.
Reason
* We will use `tf.__version__` to check if the user has the correct version installed.
* `Prophet` will print redundant information after using `importlib.util.find_spec`, like this (import `plotly` failure!), We don't need this package.

python/chronos/src/bigdl/chronos/utils.py
liangs6212: I don't check the function name judgment here, but I can add it later.
The purpose of splitting is to import the parent module. If the function name is invalid, then `__getattr__` will report an error.

liangs6212:
Can we delete some `xxx_available` because their functions have been replaced. @TheaperDeng

python/chronos/src/bigdl/chronos/utils.py
```diff
@@ -30,3 +32,67 @@ def wrapped(*args, **kwargs):
...
     return deprecated_decorator
+
+
+# code adaptted from https://github.com/intel/neural-compressor/
+#                    blob/master/neural_compressor/utils/utility.py#L62
+class LazyImport:
+    """
+    Example:
+        >>> from bigdl.chronos.utils import LazyImport
+        >>> forecaster = LazyImport('bigdl.chronos.forecaster.lstm_forecaster.LSTMForecaster')
```
TheaperDeng, line 43: just to confirm, so the `pkg` parameter won't be used right?

liangs6212: Yep.

python/chronos/src/bigdl/chronos/forecaster/__init__.py
```diff
@@ -29,51 +31,48 @@ def __exit__(self, *args, **opts):
...
-arima_available = False
-orca_available = False
-try:
-    import torch
-    torch_available = True
-except:
+torch_available = bool(importlib.util.find_spec('torch'))
+if not torch_available:
     warnings.warn("Please install `torch` to use forecasters, including TCMFForecaster, "
                   "TCNForecaster, LSTMForecaster, Seq2SeqForecaster.")
```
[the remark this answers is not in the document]
liangs6212: Will be done with migration comments.

python/chronos/src/bigdl/chronos/utils.py
```diff
@@ -30,3 +32,58 @@ def wrapped(*args, **kwargs):
...
+
+        try:
+            module = importlib.import_module(parent_name)
+            module = getattr(module, child_name) if child_name else module
+        except AttributeError:
+            # concatenate possible class names
+            full_module_name = parent_name+'.'+child_name if child_name else parent_name
+            spec = importlib.util.find_spec(full_module_name)
+            module = importlib.util.module_from_spec(spec)
+            spec.loader.exec_module(module)
```
TheaperDeng, line 77: on what circumstances this branch (line 72 - 77) is needed?

TheaperDeng (commented):
others LGTM

python/chronos/src/bigdl/chronos/utils.py
TheaperDeng: had an offline discussion, this is used for some staticmethod such as `from_tsdataset`. Please add some comment for line 70-71 and line 72-77

TheaperDeng (approved):

liangs6212:
http://10.112.231.51:18889/view/BigDL-PR-Validation/job/BigDL-Chronos-PR-Validation/693/

Issue #5229, opened 2022-07-27
quantization using openvino cannot handle when target in dataloader is not a tensor


yangw1234:
The following code will fail without adding `target_transform=transforms.Lambda(lambda label: torch.tensor(label, dtype=torch.long))`

```python

import torch
from torchvision import transforms
from torchvision.datasets import OxfordIIITPet
from torch.utils.data.dataloader import DataLoader
import torch
from torchvision.models import resnet18
from bigdl.nano.pytorch import Trainer
from torchmetrics import Accuracy


def finetune_pet_dataset(model_ft):

    train_transform = transforms.Compose([transforms.Resize(256),
                                          transforms.RandomCrop(224),
                                          transforms.RandomHorizontalFlip(),
                                          transforms.ColorJitter(brightness=.5, hue=.3),
                                          transforms.ToTensor(),
                                          transforms.Normalize([0.485, 0.456, 0.406],
                                                               [0.229, 0.224, 0.225])])
    val_transform = transforms.Compose([transforms.Resize([224, 224]),
                                        transforms.ToTensor(),
                                        transforms.Normalize([0.485, 0.456, 0.406],
                                                             [0.229, 0.224, 0.225])])

    # Apply data augmentation to the tarin_dataset
    train_dataset = OxfordIIITPet(root="/tmp/data",
                                  transform=train_transform,
                                  # target_transform=transforms.Lambda(
                                  #    lambda label: torch.tensor(label, dtype=torch.long)),
                                  download=True)
    val_dataset = OxfordIIITPet(root="/tmp/data",
                                transform=val_transform,
                                # target_transform=transforms.Lambda(
                                #     lambda label: torch.tensor(label, dtype=torch.long)),
                                )

    # obtain training indices that will be used for validation
    indices = torch.randperm(len(train_dataset))
    val_size = len(train_dataset) // 4
    train_dataset = torch.utils.data.Subset(train_dataset, indices[:-val_size])
    val_dataset = torch.utils.data.Subset(val_dataset, indices[-val_size:])

    # prepare data loaders
    train_dataloader = DataLoader(train_dataset, batch_size=32)
    val_dataloader = DataLoader(val_dataset, batch_size=32)

    num_ftrs = model_ft.fc.in_features

    # Here the size of each output sample is set to 37.
    model_ft.fc = torch.nn.Linear(num_ftrs, 37)
    loss_ft = torch.nn.CrossEntropyLoss()
[...]

PR #5237, opened 2022-07-28
[Nano] Fix python data type cannot be converted in Pytorch Dataloader

zhentaocc:
## Description

Fix quantization using openvino cannot handle when target in dataloader is of python int instead of torch tensor.

### 1. Why the change?

#5229 

### 2. User API changes

N/A

### 3. Summary of the change 

Only when the data is of `torch.Tensor`, the data will be converted to numpy.

### 4. How to test?
- [ ] #5229 script


zhentaocc:
http://10.112.231.51:18888/job/ZOO-PR-nano-pytorch-ipex-test/299/  
http://10.112.231.51:18889/view/BigDL-PR-Validation/job/BigDL-Chronos-PR-Validation/628/

yangw1234:
LGTM

PR #5238, opened 2022-07-28
Update NCF example

hkvision:
## Description

Make the default values smaller and update some minor things.


hkvision:
jenkins passed: http://10.112.231.51:18888/job/BigDL-Friesian-PR-Validation/174/

PR #5244, opened 2022-07-29
PPML: Add Hadoop filesystem support when accessing keys in KeyManagementService

jenniew:
### 1. Why the change?
Current we can only read/write primary key/datakey from/to local file system. But for cluster mode, the primary key/datakey may not on the local file system. So add hadoop fs support for reading/writing primarykey/datakey in PPMLContext and KeyManagement.

### 2. User API changes
Add hadoop configuration option in KeyManagementService:
def retrievePrimaryKey(primaryKeySavePath: String, config: Configuration = null)
def retrieveDataKey(primaryKeyPath: String, dataKeySavePath: String, config: Configuration = null)
def retrieveDataKeyPlainText(primaryKeyPath: String, dataKeyPath: String,
                               config: Configuration = null): String

For PPMLContext APIs, no change. User uses same PPMLContext APIs.

### 3. Summary of the change 
Add hadoop configuration options for KeyManagementService.
In KeyManagementService, call keyReaderWriter.writeKeyToFile and keyReaderWriter.readKeyFromFile with hadoop configuration option.

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [x] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/scala/ppml/pom.xml
+++ b/scala/ppml/pom.xml
@@ -188,11 +188,11 @@
                 </exclusion>
             </exclusions>
         </dependency>
-        <dependency>
-            <groupId>commons-cli</groupId>
-            <artifactId>commons-cli</artifactId>
-            <version>1.4</version>
-        </dependency>
+        <!--<dependency>-->
+            <!--<groupId>commons-cli</groupId>-->
+            <!--<artifactId>commons-cli</artifactId>-->
+            <!--<version>1.4</version>-->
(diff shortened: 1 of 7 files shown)


jenniew:
Jenkins pass: http://10.112.231.51:18889/job/BigDL-PR-scala-ppml-unittests-Validation/383/

Issue #5246, opened 2022-07-29
[readthedocs] Scala User Guide error


Le-Zheng:
run spark shell `${BIGDL_HOME}/bin/spark-shell-with-dllib.sh`, get this error:
```
object intel is not a member of package com
       import com.intel.analytics.bigdl.dllib.NNContext
```

![image](https://user-images.githubusercontent.com/30695225/181667139-e6fec8ec-b7e5-4bb6-b2ab-d263461c0641.png)

PR #5251, opened 2022-07-29
fix spark-shell-with-dllib.sh

Le-Zheng:
## Description
To fix: https://github.com/intel-analytics/BigDL/issues/5246
```
object intel is not a member of package com
       import com.intel.analytics.bigdl.dllib.NNContext
```

### Summary of the change 

bigdl jars path is `${BIGDL_HOME}/jars/*`

###  How to test?
- [x] manual


glorysdj (approved):
LGTM

PR #5253, opened 2022-07-29
[Nano] enable BFloat16 in IPEX inference model

zhentaocc:
## Description

Enable BFloat16 precision in IPEX. This is a quick implementation for inference only.

### 1. Why the change?

IPEX with BFloat16 is not supported in Nano.

### 2. User API changes
With this PR, users are able to use IPEX with BFloat16 acceleration as follow:
```python
bf16_model = trainer.quantize(model, precision='bf16')
y_hat = bf16_model(x)
```

### 3. Summary of the change 

- Merge `IPEXJITModel` to `PytorchIPEXJITModel`. IPEX is for Pytorch only, so we don't need to have two classes.
- Add `PytorchIPEXJITBF16Model` to enable precision autocasting.
- Add unit tests to verify ipex bf16 is working

### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/deps/ipex/ipex_api.py
+++ b/python/nano/src/bigdl/nano/deps/ipex/ipex_api.py
@@ -63,8 +63,7 @@ def PytorchIPEXJITBF16Model(model, input_sample=None, use_ipex=False,
     '''
     from .ipex_inference_bf16_model import PytorchIPEXJITBF16Model
     return PytorchIPEXJITBF16Model(model, input_sample=input_sample, use_ipex=use_ipex,
-                                   dtype=torch.bfloat16, use_jit=use_jit,
-                                   channels_last=channels_last)
+                                   use_jit=use_jit, channels_last=channels_last)
 
 
 def load_ipexjit_model(path, model):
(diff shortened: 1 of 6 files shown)


python/nano/src/bigdl/nano/deps/ipex/ipex_api.py
```diff
@@ -50,6 +50,22 @@ def PytorchIPEXJITModel(model, input_sample=None, use_ipex=False,
                                use_jit=use_jit, channels_last=channels_last)
 
 
+def PytorchIPEXJITBF16Model(model, input_sample=None, use_ipex=False,
```
yangw1234, line 53: Maybe it is better to use lower case separated with an underscore for function names.

zhentaocc: Maybe we should open another PR to resolve the function name issue? I think many of them should be modified in `deps`.

yangw1234 (approved):
LGTM

zhentaocc:
http://10.112.231.51:18889/view/BigDL-PR-Validation/job/BigDL-Chronos-PR-Validation/656/  
http://10.112.231.51:18888/job/ZOO-PR-nano-pytorch-ipex-test/312/  
@yangw1234  ready for merge.

PR #5259, opened 2022-07-29
Chronos: Update ReadTheDocs 'Windows User Guide'

zpeng1898:
### 1. Why the change?

To help windows users install `chronos` based on `wsl`. 

### 2. Summary of the change

Changed the file `BigDL/docs/readthedocs/_toc.yml`, add a page `doc/Chronos/Overview/windows_guide`, and remade html files.

### 3. How to test?

Rebuild bigdl's readthedocs and open `https://bigdl.readthedocs.io/en/latest/doc/Chronos/Overview/windows_guide.html` to see contents.


TheaperDeng:
There seems to be too many files being changed accidentally. We only need a page in document.

PR #5264, opened 2022-08-01
Change the structure of benchmark directory for auto performance regression test

MeouSker77:
### 1. Why the change?

We need to refactor the benchmark directory to support running more workloads using our auto performance regression test

### 3. Summary of the change 

Change the structure of benchmark directory.

<https://github.com/intel-analytics/BigDL/blob/57204d33dc1c2fa95ca3ff044379eb53495bff47/python/nano/benchmark/README.md>

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...


python/nano/benchmark/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+This directory contains workloads used for bigdl-nano preformance regression test, this document shows how to add new workloads:
+
+## Directory Structure
+
+Every subdirectory under the current directory is a workload, the name of subdirectory is the name of workload.
+
+Every subdirectory should contain a `run.sh`, which is used for running workload.
+
+No other rules to follow.
```
yangw1234, line 11: how about specifying the output format of performance numbers?

MeouSker77: have not decided yet

MeouSker77: maybe we also need a way to specify the argument config? i.e. use_ipex, num_processes, etc.

yangw1234:
LGTM

Issue #5236, opened 2022-07-28
[PPML] write encrypted JSON file will raise "Bad arguments" Exception


PatrickkZ:
This Exception only happens when DataFrame's record number is less than the partition number. 
for example, DataFrame `df` only contain 3 Row, when we config --master 'local[n]' or use `df.repartition(n)` where `n > 3`, some partitions will be empty. so when try to encrypt the empty partition will cause the Exception.

trace:
```
java.lang.IllegalArgumentException: Bad arguments
        at javax.crypto.Cipher.doFinal(Cipher.java:2222) ~[?:1.8.0_292]
        at com.intel.analytics.bigdl.ppml.crypto.BigDLEncrypt.doFinal(BigDLEncrypt.scala:165) ~[classes/:?]
        at com.intel.analytics.bigdl.ppml.crypto.BigDLEncryptCompressor.compress(BigDLEncryptCompressor.scala:72) ~[classes/:?]
        at org.apache.hadoop.io.compress.CompressorStream.compress(CompressorStream.java:81) ~[hadoop-common-2.7.7.jar:?]
        at org.apache.hadoop.io.compress.CompressorStream.finish(CompressorStream.java:92) ~[hadoop-common-2.7.7.jar:?]
        at org.apache.hadoop.io.compress.CompressionOutputStream.close(CompressionOutputStream.java:60) ~[hadoop-common-2.7.7.jar:?]
        at org.apache.hadoop.io.compress.CompressorStream.close(CompressorStream.java:106) ~[hadoop-common-2.7.7.jar:?]
        at sun.nio.cs.StreamEncoder.implClose(StreamEncoder.java:320) ~[?:1.8.0_292]
        at sun.nio.cs.StreamEncoder.close(StreamEncoder.java:149) ~[?:1.8.0_292]
        at java.io.OutputStreamWriter.close(OutputStreamWriter.java:233) ~[?:1.8.0_292]
        at com.fasterxml.jackson.core.json.WriterBasedJsonGenerator.close(WriterBasedJsonGenerator.java:999) ~[jackson-core-2.11.4.jar:2.11.4]
        at org.apache.spark.sql.catalyst.json.JacksonGenerator.close(JacksonGenerator.scala:239) ~[spark-catalyst_2.12-3.1.2.jar:3.1.2]
[...]

PR #5268, opened 2022-08-01
Fix issue #5236, fix the bug when try to encrypt empty partition

PatrickkZ:
## Description

when the partition number is more than the actual data record number(maybe because we config --master 'local[n]' or use df.repartition(n) where n > data record number). some partitions may be empty. when try to encrypt the empty partition, there will be a "Bad Argument" error in our custom `codec`.
this error may happen when the DataFrame is really small. so this pr fix this problem. 
related issue: https://github.com/intel-analytics/BigDL/issues/5236


glorysdj (approved):
LGTM

PR #5270, opened 2022-08-01
[Nano] OpenVINO model enahancement

### LINKED ISSUE #5199: [Nano] improve OpenVINO user experience
## Description
`bigdl.nano.openvino.OpenVINOModel` has now provided limited features for users and few notifications about current status. We can only simply run inference like：
```python
openvino_model = OpenVINOModel("resnet18-xnor-binary-onnx-0001.xml")
preds = openvnino_model.forward_step(x)
```
This is the only thing we can do right now. Here we can discuss about more features we can provide to improve user experience

## Device setting
Similar to Pytorch, we can provide a function for users to simply copy models from one device to another.
```python
openvino_cpu_model = OpenVINOModel("resnet18-xnor-binary-onnx-0001.xml", device='CPU')
openvino_gpu_model = openvino_cpu_model.to('GPU')
preds = openvnino_model.forward_step(x)
```
## Notification/Log
Notify users of model information.
```python
openvino_model = OpenVINOModel("resnet18-xnor-binary-onnx-0001.xml")
```
will output:
```shell
[INFO] File Path : resnet18-xnor-binary-onnx-0001.xml
[INFO] Precision: FP32
[INFO] Graph has 1 input and 2 outputs.
[INFO] Device: CPU
```
## Extensions
Add extensions to OpenVINO Core. Seems if we want to support more devices, we need to load extra extensions. For example, NCS needs extra extension.  
Needs further investigation.

zhentaocc:
## Description

Enhance OpenVINO model with more features and logs.

### 1. Why the change?
resolves #5199 
We get inspired by AI course notebooks that there are some features that users may need.  
Reference: https://github.com/simpledevelopments/OpenVINO-Python-Utils

#### Before
```python
class OpenVINOModel:
    def __init__(self, ie_network: str, device='CPU'):
```
#### After
```python
class OpenVINOModel:
    def __init__(self, ie_network: str, device='CPU', *extensions):
        ...
    def add_extensions(self, *extensions):
        """
        Add extensions to Core. 
        """
        ...

    def to(self, device, inplace=False):
        ...
```

#### `OpenVINOModel.to`
Users are able to copy the model from one device to another by:
```python
ov_cpu_model = OpenVINOModel("model.xml', device="CPU")
ov_gpu_model = ov_cpu_model.to('GPU') # create a new model on GPU
ov_gpu_model = ov_cpu_model.to('GPU', inplace=True) # move the model to GPU
```

#### `OpenVINOModel.add_extensions` 
similar to https://docs.openvino.ai/latest/api/ie_python_api/_autosummary/openvino.runtime.Core.html#openvino.runtime.Core.add_extension

### 3. Summary of the change 

- Add new API `OpenVINOModel.to`, `OpenVINOModel.add_extensions` 
- Add Log Info
When users create a OpenVINO model, it will notify users like:
```python
ov_cpu_model = OpenVINOModel("model.xml', device="CPU")
```
```text
XML File : ./intel/resnet18-xnor-binary-onnx-0001/FP16-INT1/resnet18-xnor-binary-onnx-0001.xml
Model is loaded on Device: CPU
```

### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/deps/openvino/core/model.py
+++ b/python/nano/src/bigdl/nano/deps/openvino/core/model.py
@@ -71,8 +71,10 @@ class OpenVINOModel:
     def ie_network(self, model):
         if isinstance(model, (str, Path)):
             self._ie_network = self._ie.read_model(model=str(model))
+            logging.info("XML File : {}".format(str(model)))
         else:
             self._ie_network = model
+            logging.info("Model is loaded from existing instance.")
         self.to(self._device, inplace=True)
 
     def _save_model(self, path):
@@ -87,6 +89,7 @@ class OpenVINOModel:
                           "self.ie_network shouldn't be None.")
         xml_path = path / self.status['xml_path']
         save(self.ie_network, xml_path)
+        logging.info("Model is saved in {}".format(xml_path))
 
     def pot(self,
             dataloader,
--- a/python/nano/src/bigdl/nano/deps/openvino/core/model.py
+++ b/python/nano/src/bigdl/nano/deps/openvino/core/model.py
@@ -19,6 +19,7 @@ from openvino.runtime import Core
 from bigdl.nano.utils.log4Error import invalidInputError
 from openvino.runtime import Model
 from .utils import save
+import logging
 
 
 class OpenVINOModel:
@@ -27,6 +28,8 @@ class OpenVINOModel:
         self.extensions = []
         self.add_extensions(*extensions)
         self._device = device
+        self._compiled_model = None
+        self._infer_request = None
         self.ie_network = ie_network
 
     def forward_step(self, *inputs):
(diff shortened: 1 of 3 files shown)


jason-dai:
Why do we need to support these two APIs? Do you have some 3rd-party examples of how these two APIs can be used?

zhentaocc:
#### `OpenVINOModel.to`
This is a similar implementation to set/get a compiled model on specific machine which avoid the same model being compiled on one device.
https://github.com/simpledevelopments/OpenVINO-Python-Utils/blob/master/utils/opv.py#L31-L36

#### `OpenVINOModel.add_extensions` 
With a more deep investigation, extensions are loaded for some customized operators in most cases. So I think maybe this is too advanced for our users? I would remove this now in case we need this feature.
similar to https://docs.openvino.ai/latest/api/ie_python_api/_autosummary/openvino.runtime.Core.html#openvino.runtime.Core.add_extension

jason-dai:
How will these APIs be used? Provide some external examples.

PR #5271, opened 2022-08-01
Fix duplicate repartition in tf2 estimator spark backend

hkvision:
## Description

In fit, if input is dataframe, we will do repartition if partition < num workers when converting to SparkXShards.
After converting to SparkXShards, the data should have num_partitions >= num_workers for training (if there is no empty partition in original user's data, which is assumed).
Since in SparkXShards, each partition is combined to a single dict, if you call repartition once more on SparkXShards, it will result in empty partitions again.

```
>>> rdd = sc.parallelize([1, 2, 3, 4], numSlices=4)
>>> rdd.glom().collect()
[[1], [2], [3], [4]]
>>> rdd2 = rdd.repartition(4)
>>> rdd2.glom().collect()
[[4, 3], [2], [], [1]]

# 1, 2, 3, 4 correspond to the entire dict of data in each SparkXShards partition
```

### DIFF
--- a/python/orca/test/bigdl/orca/learn/ray/tf/test_tf_spark_estimator.py
+++ b/python/orca/test/bigdl/orca/learn/ray/tf/test_tf_spark_estimator.py
@@ -13,20 +13,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+
+import os
+import pytest
 import shutil
 import tempfile
 from unittest import TestCase
-import time
 
-import time
 import numpy as np
-import pytest
 import tensorflow as tf
 
 from bigdl.orca.learn.tf2 import Estimator
 from bigdl.orca import OrcaContext
 
-import os
 
 resource_path = os.path.join(
     os.path.realpath(os.path.dirname(__file__)), "../../../../resources")
@@ -126,7 +125,7 @@ class TestTFEstimator(TestCase):
                 model_creator=model_creator,
                 verbose=True,
                 config=config,
-                workers_per_node=2,
+                workers_per_node=3,
                 backend="spark",
                 model_dir=temp_dir)
 
--- a/python/orca/src/bigdl/orca/learn/tf2/pyspark_estimator.py
+++ b/python/orca/src/bigdl/orca/learn/tf2/pyspark_estimator.py
@@ -288,7 +288,7 @@ class SparkTFEstimator():
             data_config=data_config,
         )
 
-        if isinstance(data, SparkXShards):
+        if isinstance(data, SparkXShards):  # In this case, data partitions must be >= num_workers
             # set train/validation data
             if data._get_class_name() == 'pandas.core.frame.DataFrame':
                 data = process_xshards_of_pandas_dataframe(data, feature_cols, label_cols)
@@ -300,7 +300,7 @@ class SparkTFEstimator():
 
             rdd = data.rdd
             if rdd.getNumPartitions() > self.num_workers:
-                rdd = rdd.coalesce(self.num_workers)
+                rdd = rdd.coalesce(self.num_workers)  # Use coalesce to avoid repartition
             res = rdd.barrier().mapPartitions(
                 lambda iter: transform_func(iter, init_params, params)).collect()
         else:
--- a/python/orca/src/bigdl/orca/learn/tf2/pyspark_estimator.py
+++ b/python/orca/src/bigdl/orca/learn/tf2/pyspark_estimator.py
@@ -125,8 +125,14 @@ class SparkTFEstimator():
         """
         sc = OrcaContext.get_spark_context()
 
-        # After converting Spark DataFrame to SparkXShards,
-        # we must have num_partition >= num_workers
+        # Data partition should be equal to num workers.
+        # Repartition Spark DataFrame before converting to SparkXShards.
+        # Repartition on SparkXShards will result in empty partitions.
+        if isinstance(data, DataFrame) or isinstance(data, SparkXShards):
+            if data.rdd.getNumPartitions() != self.num_workers:
+                data = data.repartition(self.num_workers)
+            if validation_data and validation_data.rdd.getNumPartitions() != self.num_workers:
+                validation_data = validation_data.repartition(self.num_workers)
         data, validation_data = maybe_dataframe_to_xshards(data, validation_data,
                                                            feature_cols, label_cols,
                                                            mode="fit",
@@ -182,10 +188,7 @@ class SparkTFEstimator():
                     param["data_creator"] = make_data_creator(partition_data)
                     return SparkRunner(**init_param).step(**param)
 
-                rdd = data.rdd
-                if rdd.getNumPartitions() > self.num_workers:
-                    rdd = rdd.repartition(self.num_workers)  # coalesce can be used with barrier
-                res = rdd.barrier().mapPartitions(
+                res = data.rdd.barrier().mapPartitions(
                     lambda iter: transform_func(iter, init_params, params)).collect()
             else:
                 def transform_func(iter, init_param, param):
@@ -198,10 +201,7 @@ class SparkTFEstimator():
 
                 train_rdd = data.rdd.mapPartitions(lambda iter: [list(iter)])
                 val_rdd = validation_data.rdd.mapPartitions(lambda iter: [list(iter)])
-                zip_rdd = train_rdd.zip(val_rdd)
-                if zip_rdd.getNumPartitions() > self.num_workers:
-                    zip_rdd = zip_rdd.repartition(self.num_workers)
-                res = zip_rdd.barrier().mapPartitions(
+                res = train_rdd.zip(val_rdd).barrier().mapPartitions(
                     lambda iter: transform_func(iter, init_params, params)).collect()
         else:
             params["data_creator"] = data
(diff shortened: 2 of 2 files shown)


jason-dai:
Why do we repartition previously? @jenniew

jenniew:
> Why do we repartition previously? @jenniew

The previous repartition is in `maybe_dataframe_to_xshards()` which was originally created for ray estimator. It avoids empty partition for worker when partitions < num of workers.  TF2Estimator on Pyspark reused this function to change dataframe to xshards. 
I think we may not remove the repartition when training/evaluation.
We do repartition to let the training/evaluation only on number of workers. 
If there are thousands of partitions, we don't want to train on thousands of workers. 
So I think for the fix of situation that partition < num workers may like:  
```
if data.rdd.getNumPartitions() > self.num_workers:
    res = data.rdd.repartition(self.num_workers).barrier() \
                    .mapPartitions(
                    lambda iter: transform_func(iter, init_params, params)).collect()
else:
   res = data.rdd.barrier() \
                    .mapPartitions(
                    lambda iter: transform_func(iter, init_params, params)).collect()
```

hkvision:
Why do we must have the number of partitions equal to the number of workers?
Repartition is expensive, if the number of partitions is already larger than the number of workers, why can't each worker just take more than 1 partition like the ray estimator.

jason-dai:
> Why do we must have the number of partitions equal to the number of workers? Repartition is expensive, if the number of partitions is already larger than the number of workers, why can't each worker just take more than 1 partition like the ray estimator, in mapPartitions one worker finish one partition and takes another partition?

I think we can do repartitioning without shuffling?

hkvision:
> > Why do we must have the number of partitions equal to the number of workers? Repartition is expensive, if the number of partitions is already larger than the number of workers, why can't each worker just take more than 1 partition like the ray estimator, in mapPartitions one worker finish one partition and takes another partition?
> 
> I think we can do repartitioning without shuffling?

After converting to XShards, the number of partitions must be >= the number of workers. So we can directly call coalesce when num partitions > num workers? If the original data is relatively evenly partitioned, each worker will get around the same amount of data?

jenniew:
> > > Why do we must have the number of partitions equal to the number of workers? Repartition is expensive, if the number of partitions is already larger than the number of workers, why can't each worker just take more than 1 partition like the ray estimator, in mapPartitions one worker finish one partition and takes another partition?
> > 
> > 
> > I think we can do repartitioning without shuffling?
> 
> After converting to XShards, the number of partitions must be >= the number of workers. So we can directly call coalesce when num partitions > num workers? If the original data is relatively evenly partitioned, each worker will get around the same amount of data?

Using coalesce is a better way.

jenniew (approved):
LGTM

hkvision:
Will coalesce result in unbalanced partitions?
e.g. node1 has 9 partitions and node2 has 1 partition, after coalecse to 2 partitions, will each new partition has 5 smaller partitions or one has 9 and one has 1?
@jason-dai @jenniew

yushan111:
The issue seems to be caused by a duplicated repartition when input data partition < num_workers, shall we also add test case for that? 
Can the issue detected by a similar test case in ray: [test_partition_num_less_than_workers](https://github.com/intel-analytics/BigDL/blob/21edc298f7fdd299a2d0d8669f9662deeb64c092/python/orca/test/bigdl/orca/learn/ray/tf/test_tf_ray_estimator.py#L440) with `workers_per_node=4`?

hkvision:
After changing to coalesce will throw the following error:
```
>                       format(target_id, ".", name), value)
E                   py4j.protocol.Py4JJavaError: An error occurred while calling z:org.apache.spark.api.python.PythonRDD.collectAndServe.
E                   : org.apache.spark.scheduler.BarrierJobUnsupportedRDDChainException: [SPARK-24820][SPARK-24821]: Barrier execution mode does not allow the following pattern of RDD chain within a barrier stage:
E                   1. Ancestor RDDs that have different number of partitions from the resulting RDD (eg. union()/coalesce()/first()/take()/PartitionPruningRDD). A workaround for first()/take() can be barrierRdd.collect().head (scala) or barrierRdd.collect()[0] (python).
E                   2. An RDD that depends on multiple barrier RDDs (eg. barrierRdd1.zip(barrierRdd2)).
E                   	at org.apache.spark.scheduler.DAGScheduler.checkBarrierStageWithRDDChainPattern(DAGScheduler.scala:372)
E                   	at org.apache.spark.scheduler.DAGScheduler.createResultStage(DAGScheduler.scala:448)
E                   	at org.apache.spark.scheduler.DAGScheduler.handleJobSubmitted(DAGScheduler.scala:963)
E                   	at org.apache.spark.scheduler.DAGSchedulerEventProcessLoop.doOnReceive(DAGScheduler.scala:2069)
E                   	at org.apache.spark.scheduler.DAGSchedulerEventProcessLoop.onReceive(DAGScheduler.scala:2061)
E                   	at org.apache.spark.scheduler.DAGSchedulerEventProcessLoop.onReceive(DAGScheduler.scala:2050)
E                   	at org.apache.spark.util.EventLoop$$anon$1.run(EventLoop.scala:49)
E                   	at org.apache.spark.scheduler.DAGScheduler.runJob(DAGScheduler.scala:738)
E                   	at org.apache.spark.SparkContext.runJob(SparkContext.scala:2061)
E                   	at org.apache.spark.SparkContext.runJob(SparkContext.scala:2082)
E                   	at org.apache.spark.SparkContext.runJob(SparkContext.scala:2101)
E                   	at org.apache.spark.SparkContext.runJob(SparkContext.scala:2126)
E                   	at org.apache.spark.rdd.RDD$$anonfun$collect$1.apply(RDD.scala:990)
E                   	at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:151)
E                   	at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:112)
E                   	at org.apache.spark.rdd.RDD.withScope(RDD.scala:385)
E                   	at org.apache.spark.rdd.RDD.collect(RDD.scala:989)
E                   	at org.apache.spark.api.python.PythonRDD$.collectAndServe(PythonRDD.scala:166)
E                   	at org.apache.spark.api.python.PythonRDD.collectAndServe(PythonRDD.scala)
E                   	at sun.reflect.GeneratedMethodAccessor80.invoke(Unknown Source)
E                   	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
E                   	at java.lang.reflect.Method.invoke(Method.java:498)
E                   	at py4j.reflection.MethodInvoker.invoke(MethodInvoker.java:244)
E                   	at py4j.reflection.ReflectionEngine.invoke(ReflectionEngine.java:357)
E                   	at py4j.Gateway.invoke(Gateway.java:282)
E                   	at py4j.commands.AbstractCommand.invokeMethod(AbstractCommand.java:132)
E                   	at py4j.commands.CallCommand.execute(CallCommand.java:79)
E                   	at py4j.GatewayConnection.run(GatewayConnection.java:238)
E                   	at java.lang.Thread.run(Thread.java:748)
```

hkvision:
@jason-dai @jenniew 
barrier can't be performed on the rdd that comes from coalesce...
So we still keep the repartition if rdd.getNumPartitions > num workers?

python/orca/test/bigdl/orca/learn/ray/tf/test_tf_spark_estimator.py
```diff
@@ -126,7 +125,7 @@ def test_dataframe_with_empty_partition(self):
                 model_creator=model_creator,
                 verbose=True,
                 config=config,
-                workers_per_node=2,
+                workers_per_node=3,
```
hkvision, line 128: After changing num_workers to 3, the unit test will fail. 4 partitions with an empty partition repartition to 3 will still have an empty partition.
```
>>> rdd = sc.parallelize([1, 2, 3], numSlices=4)
>>> rdd.glom().collect()
[[], [1], [2], [3]]                                                        	 
>>> rdd2 = rdd.repartition(3)
>>> rdd2.glom().collect()
[[], [2], [1, 3]]
```
Please take a look @jenniew

hkvision: We'd better do repartition before converting to SparkXShards to avoid such issue.

jason-dai:
> @jason-dai @jenniew barrier can't be performed on the rdd that comes from coalesce... So we still keep the repartition if rdd.getNumPartitions > num workers?

Can we do repartitioning without shuffle?

hkvision:
> take

Seems no? To reduce num partitions without shuffle, use coalesce, which can't be combined with barrier. Unsupported to increase partitions without shuffle: https://stackoverflow.com/questions/71070709/increase-the-number-of-partitions-without-repartition-on-hadoop

hkvision:
> Will coalesce result in unbalanced partitions? e.g. node1 has 9 partitions and node2 has 1 partition, after coalecse to 2 partitions, will each new partition has 5 smaller partitions or one has 9 and one has 1? @jason-dai @jenniew

Also, even if coalesce can work with barrier, it will result in unbalanced partitions, which will have issues in the training.

I suppose currently under this implementation we still need to do repartition.

hkvision:
Issue conclusion:
- The implementation of spark backend requires the number of data partitions equal to the number of workers
- After converting to SparkXShards, each partition will have only one element (if shard_size is not specified), and if we do repartition on SparkXShards, very likely will result in empty partitions.
- Move the repartition to Spark DataFrame before converting to SparkXShards, which is the best solution at the moment.

hkvision:
http://10.112.231.51:18888/job/ZOO-PR-BigDL-Python-Spark-3.1-py37-Horovod/1330/

sgwhat (approved):

PR #5277, opened 2022-08-02
PPML add assembly package for production env

Litchilitchy:
## Description
* Add PPML assembly package
* start FLServer is supported from scripts in this package
* To use it, 1. get assembly folder 2. `source setup-env.sh` 3. run `python start-fl-server.py`


Litchilitchy:
The contents and related changes are made and tested by #5396 , thus this is closed

PR #5294, opened 2022-08-03
[Orca] Support saving/loading trained TF model with h5/saveModel format

sgwhat:
### 1. Why the change?

* Current `save` and `load` do not support `h5` and `savemodel` format, since they mainly retrieve model states.
* Users need the API to save model as `h5` or `savemodel` format.

### 2. User API changes
For new added API:
* save model
  ```python
  estimator.save(filepath, overwrite, include_optimizer, save_format, signatures, options)
  ```
* load model
   ```python
  estimator.load(filepath, custom_objects, compile)
  ```

For original save/load API:
* save -> save_checkpoint
   ```python
   estimator.save_checkpoint(checkpoint)
   ```
* load -> load_checkpoint
   ```python
   estimator.load_checkpoint(checkpoint)
   ```

### 3. Summary of the change 

Provide new `save` and `load` API to support `h5` and `savemodel` format (both load and hdfs).

### 4. How to test?
- [x] Unit test
- [x] Test saving/loading on HDFS
- [x] Test on Yarn


python/orca/src/bigdl/orca/learn/tf2/ray_estimator.py
```diff
@@ -517,6 +522,99 @@ def load(self, checkpoint, **kwargs):
...
+             compile=True):
+        """
+        Loads a model saved via `estimator.save()
+
+        :param filepath: (str) Path of saved model.
+        :param custom_objects: Optional dictionary mapping names
+          (strings) to custom classes or functions to be
+          considered during deserialization.
+        :param compile: Boolean, whether to compile the model after loading.
+        :param options: Optional `tf.saved_model.LoadOptions` object that specifies
```
hkvision, line 583: what's this? not in the above arguments?

python/orca/src/bigdl/orca/learn/tf2/ray_estimator.py
```diff
@@ -517,6 +522,99 @@ def load(self, checkpoint, **kwargs):
...
+        if is_local_path(filepath):
+            params = dict(
+                filepath=filepath,
+                custom_objects=custom_objects,
+                compile=compile
+            )
+            model_refs = [
+                worker.load_model.remote(**params) for worker in self.remote_workers
+            ]
+            ray.get(model_refs)
```
hkvision, line 596: is this logic correct? Will the filepath be available to all workers?
Should we first load to the driver and send the model to all workers?

sgwhat: > is this logic correct? Will the filepath be available to all workers? Should we first load to the driver and send the model to all workers?

Yes, it's available. Load to driver maybe not a good solution, since the model will be created in `TFRunner`.

python/orca/src/bigdl/orca/learn/tf2/ray_estimator.py
sgwhat: added.

hkvision (approved):

PR #5295, opened 2022-08-03
Solve the problem of missing tag when schedule dispatch

Romanticoseu:
## Description

update docker-publish-bigdl for set env, solve the problem of missing tag when schedule


glorysdj (approved):
LGTM

PR #5298, opened 2022-08-03
add performance regression test workflow

MeouSker77:
## Description

- add performance regression test workflow
- remove the benchmark test in nano unit test


yangw1234:
LGTM

PR #5303, opened 2022-08-03
support lightning model

yexinyinancy:
### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

### 5. New dependencies

- [ ] New Python dependencies
       - Dependency1 
       - Dependency2
       - ...
       - Dependency1 and license1
       - Dependency2 and license2
       - ...

### DIFF
--- a/python/orca/src/bigdl/orca/learn/pytorch/estimator.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/estimator.py
@@ -22,7 +22,7 @@ class Estimator(object):
     @staticmethod
     def from_torch(*,
                    model,
-                   optimizer,
+                   optimizer=None,
                    loss=None,
                    metrics=None,
                    scheduler_creator=None,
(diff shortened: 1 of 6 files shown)


yexinyinancy:
Currently support:
1. pytorch-lightning model with function ``forward``, ``cross_entropy_loss``, ``training_step``, ``loss``, ``validation_step`` and ``configure_optimizers``.
2. The return type of ``training_step`` can be loss or dictionary.
3. The return type of ``configure_optimizers`` can be optimizer, scheduler and dictionary.
4. pytorch_lightning callbacks.
5. The input data type should be xshards.

yexinyinancy:
Current issue:
1. If the input feature has only one column, x in ``training_step`` is a list, we need to use x[0] to get the actual data and put it into the model.
2. We cannot support all the function in pytorch_lightning callbacks.

PR #5307, opened 2022-08-04
Chronos : add multi-objective HPO with latency tutorial

rnwang04:
## Description
Add a tutorial of how to tune a Time Series Forecasting Model with multi-objective hyperparameter optimization.

### 4. How to test?
- [x] Document test
https://ruonantestdoc.readthedocs.io/en/mohpo_tutorial/


TheaperDeng:
1. please teach our users how to download the electricity dataset. the link you give is the raw data while the `electricity.csv` you use is a preprocessed version.
  
2. [Some thing we may change in future], use `forecaster.tune_trainer.hposearcher.study` directly might not be that graceful, we may add some external API for this.
  
3. [Some thing we may change in future] The figure legend is quiet hard to understand, I think it's just a simple ranking... we may remove it or give some explanation in future.
  
4. How could users find the lower left one is trail_6?
  
5. So, at last we did not meet the 1ms requirement we proposed in the backgroud, we may just write our requirement is 1.5ms...
  
6. remove In[3] and In[1]'s output
  
7. I understand the latency in the figure is calculated by the batch_size in validation data, but we may want to explain this in text or just set validation data's batch size to 1.
  
8. [Some thing we may change in future], we need to add automatically test for this notebook.
  
9. In In[16], we don't need `forecaster.build_onnx(thread_num=1)`

rnwang04:
> 1. please teach our users how to download the electricity dataset. the link you give is the raw data while the `electricity.csv` you use is a preprocessed version.
> 2. [Some thing we may change in future], use `forecaster.tune_trainer.hposearcher.study` directly might not be that graceful, we may add some external API for this.
> 3. [Some thing we may change in future] The figure legend is quiet hard to understand, I think it's just a simple ranking... we may remove it or give some explanation in future.
> 4. How could users find the lower left one is trail_6?
> 5. So, at last we did not meet the 1ms requirement we proposed in the backgroud, we may just write our requirement is 1.5ms...
> 6. remove In[3] and In[1]'s output
> 7. I understand the latency in the figure is calculated by the batch_size in validation data, but we may want to explain this in text or just set validation data's batch size to 1.
> 8. [Some thing we may change in future], we need to add automatically test for this notebook.
> 9. In In[16], we don't need `forecaster.build_onnx(thread_num=1)`

1. Modified the link of electricity dataset to the preprocessed version, thanks.
4. Actually if user run In[5] himself, each point will display its trial number and corresponding metric number, but in notebook the original figure can't display. So I just put a static image here.
5. Good catch, thanks. Done.
6. Done.
7. Good catch, thanks. Done.
9. Done.

TheaperDeng:
In In[1], it seems that `test_data` is never used.

rnwang04:
> In In[1], it seems that `test_data` is never used.
> 
> others lgtm.

Done. Thanks.

PR #5310, opened 2022-08-04
Orca: add pyarrow dependency

lalalapotter:
## Description
This PR is to add pyarrow dependency in setup.py for environment intallation.


python/orca/src/setup.py
```diff
@@ -97,7 +97,8 @@ def setup_package():
         packages=get_bigdl_packages(),
         install_requires=['packaging', 'filelock',
                           'bigdl-tf==0.14.0.dev1', 'bigdl-math==0.14.0.dev1',
-                          'bigdl-dllib=='+VERSION, 'pyzmq'],
+                          'bigdl-dllib=='+VERSION, 'pyzmq',
+                          'pyarrow'],
```
hkvision, line 101: put in the same line?

lalalapotter: fixed

PR #5313, opened 2022-08-04
fix jackson-databind Vulnerability

Le-Zheng:
## Description

Upgrade jackson in dllib

### DIFF
--- a/scala/dllib/pom.xml
+++ b/scala/dllib/pom.xml
@@ -112,13 +112,13 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
-            <version>2.11.4</version>
+            <version>2.12.2</version>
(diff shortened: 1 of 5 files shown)


Le-Zheng:
http://10.112.231.51:18888/job/BigDL-PR-Dllib-Validation/579/ pass

jenniew:
@glorysdj should we put all jackson dependency version definition in parent pom, and only inference it in children pom? Or we specify all jackson dependency version individually in children pom?

Le-Zheng:
Agreed with putting the Jackson dependency version definition in parent pom.

cyita (approved):

Le-Zheng:
Jenkins pass: http://10.112.231.51:18888/job/BigDL-PR-Dllib-Validation/581/  http://10.112.231.51:18888/job/BigDL-Friesian-PR-Validation/196/

PR #5319, opened 2022-08-04
Resolve jackson and reactor dependency for Azure SDK

jenniew:
## Description
Resolve jackson and reactor-core dependency for Azure SDK


Le-Zheng:
LGTM

scala/pom.xml
```diff
@@ -261,6 +261,11 @@
                 <version>${hadoop.version}</version>
                 <scope>${spark-scope}</scope>
             </dependency>
+            <dependency>
+                <groupId>io.projectreactor</groupId>
+                <artifactId>reactor-core</artifactId>
+                <version>3.4.14</version>
```
cyita, line 267: Define reactor-core version in parent pom?

PR #5320, opened 2022-08-05
Nano: fix ipex_strategy when set enable_bf16=True

y199387:
### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

### 5. New dependencies

- [ ] New Python dependencies
       - Dependency1 
       - Dependency2
       - ...
       - Dependency1 and license1
       - Dependency2 and license2
       - ...


yangw1234:
Can we try to use precision plugin (https://pytorch-lightning.readthedocs.io/en/stable/extensions/plugins.html?highlight=precision#precision-plugins) to do this?

PR #5325, opened 2022-08-05
Orca: refactor HDFS operation in estimator

lalalapotter:
## Description
Refer to the report by customers, the program reported `list index out of range` error if they do not specify the HDFS port. Because that the port is always set to default value(8020/9000) in real deployments, users are used to ignore the setting of HDFS port. This PR is to improve user experience that they do not have to specify the HDFS port for the connection to HDFS.

Refactor HDFS operations in estimator.

### DIFF
--- a/python/orca/src/bigdl/orca/data/file.py
+++ b/python/orca/src/bigdl/orca/data/file.py
@@ -155,16 +155,13 @@ def exists(path):
             invalidOperationError(False, str(ex), cause=ex)
         return True
     elif path.startswith("hdfs://"):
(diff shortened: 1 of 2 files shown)


sgwhat:
Does the usage api looks like:
```python
model_dir = "hdfs://xxx.xxx.xx.xx/path/to/remote_dir"
```

lalalapotter:
Yes, you can just specify the host IP or specify both host IP and port.

hkvision (approved):

PR #5328, opened 2022-08-05
Friesian serving http add shutdown hook

cyita:
## Description

`System.in.read();` will be killed in k8s deployment, use `shutdown hook` to fix this issue.


cyita:
Jenkins passed http://10.112.231.51:18888/job/BigDL-Friesian-PR-Validation/204/

PR #5340, opened 2022-08-08
Save benchmark result to database

MeouSker77:
## Description

* change it can be triggered by comments 'APRT'
* save benchmark result to database


yangw1234:
Maybe we should run the performance test on every PR request. I think it is best if we can leave a specific comment on the PR conversion to trigger the tests.

MeouSker77:
test

MeouSker77:
test

MeouSker77:
test

python/nano/benchmark/run.py
```diff
@@ -43,11 +46,41 @@ def run(self):
     @staticmethod
     def _extract_logs(output: str):
         logs = [match.group(1) for match in re.finditer(">>>(.*?)<<<", output)]
+        logs = [json.loads(log) for log in logs]
         return logs
 
     def _save_logs(self, logs: list):
+        is_pr = True if os.environ.get('IS_PR') is not None else False
+        timestamp = datetime.now()
+        conn = psycopg2.connect(
```
yangw1234, line 55: @glorysdj we are using k8s secrets to connect to a database deployed on k8s, do you think it is safe to do this in github action?

yangw1234: Or should we use the secrets in github action https://github.com/Azure/actions-workflow-samples/blob/master/assets/create-secrets-for-GitHub-workflows.md

yangw1234:
LGTM

MeouSker77:
APRT

PR #5342, opened 2022-08-08
PPML add psiSalt argument in config

Litchilitchy:
## Description

To support a user-defined `psiSalt` parameter in config file, and also `FLContext.setPsiSalt`

In the past, we only supported the `getSalt()` method which returns a random salt.


qiyuangong (approved):
LGTM

jason-dai:
Need to describe the user API in the PR

PR #5343, opened 2022-08-09
[Actions] Fix the java unittest scope for mvn test

hoshibara:
Because the unittest newly introduced for friesian serving dependend on redis and mkl, and the previous mvn test did not specify the scope of java unittest. As a result, both ppml and dllib ran these unrelated tests, and reported errors.

Issue: [FeatureInitializerTest.testFeatureInit:103 » RedisConnection Unable to connect](https://github.com/intel-analytics/arda-docker/issues/690)


hoshibara:
* add test scope for each mvn test
* add friesian unittest when pr related code
* disable friesian NB runs on pr or commit

PR #5351, opened 2022-08-09
DLlib: add enable_hdfs_load decorator

lalalapotter:
## Description
According to the customer feedback, there is no `SparkContext` error while loading HDFS file in ray workers, which is because users use `enable_multi_fs_load` decorator to load HDFS file and need `SparkContext` to get gateway of Scala service in each worker. This PR is to add `enable_hdfs_load` decorator, which is totally implemented in Python, to address the customer issues.

### DIFF
--- a/python/dllib/src/bigdl/dllib/utils/file_utils.py
+++ b/python/dllib/src/bigdl/dllib/utils/file_utils.py
@@ -13,16 +13,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-from bigdl.dllib.utils.common import Sample as BSample, JTensor as BJTensor,\
-    JavaCreator, _get_gateway, _py2java, _java2py
-import numpy as np
-import os
-import tempfile
-import uuid
+
 import functools
 import glob
-
+import os
+import subprocess
+import tempfile
 from urllib.parse import urlparse
+import uuid
+
+import numpy as np
+import pyarrow as pa
+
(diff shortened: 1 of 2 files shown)


python/dllib/src/bigdl/dllib/utils/file_utils.py
```diff
@@ -13,16 +13,20 @@
...
-    JavaCreator, _get_gateway, _py2java, _java2py
-import numpy as np
+
+import functools
+import glob
 import os
+import subprocess
 import tempfile
+from urllib.parse import urlparse
 import uuid
```
hkvision, line 23: exchange these two lines?

python/dllib/src/bigdl/dllib/utils/file_utils.py
```diff
@@ -13,16 +13,20 @@
...
-import functools
-import glob
 
-from urllib.parse import urlparse
+import numpy as np
+import pyarrow as pa
```
hkvision, line 26: put this import locally in the function? as dllib is not necessarily depends on pyarrow

hkvision:
Need to add a unit test

python/dllib/src/bigdl/dllib/utils/file_utils.py
lalalapotter: fixed

hkvision (approved):

PR #5361, opened 2022-08-10
PPML FGBoost add support for server/client save

Litchilitchy:
## Description

add incremental training for FGBoost and add according tutorial section


jason-dai:
Need detailed description of the API and design for this PR. @Litchilitchy

jason-dai:
> ### API
> In `ppml-conf.yaml`, specify `fgBoostServerModelPath` to save/load

Use python code instead of configuration file

Litchilitchy:
> > ### API
> > In `ppml-conf.yaml`, specify `fgBoostServerModelPath` to save/load
> 
> Use python code instead of configuration file

Put it in 1. Server code like `./start-fl-server --fgBoostServerModelPath` or 2. in client code `config.set(attr, value)` and upload this config to server?

PR #5365, opened 2022-08-11
Delete schedule trigger

Romanticoseu:
## Description

There is a timeout problem in the process of manual triggering. Now only the manual selection of build products is reserved


.github/workflows/docker-publish-ppml.yaml
```diff
@@ -1,8 +1,6 @@
-name: Nightly Build Docker Publish BigDL-PPML
+name: Manually Build Docker Publish BigDL-PPML
```
glorysdj, line 1: keep the old name please

glorysdj (approved):
LGTM

PR #5374, opened 2022-08-12
[Chronos] fix save and load of tuned built-in forecaster model

rnwang04:
## Description

fix save and load of tuned built-in forecaster model.

### 1. Why the change?

Customer report Autoformer can't load model after saving a tuned model. We find all Chronos forecaster models have the same problem.

### 2. User API changes
For save : 
```
forecatser = AutoformerForecaster(...) # parameters with space
forecatser.tune()
forecatser.fit()
forecatser.save(ckpt_name)
```
When load : 
```
forecatser = AutoformerForecaster(...) # parameters with or without space
forecatser.load(ckpt_name)
```

### 3. Summary of the change 

fix save and load of tuned forecaster model.

### 4. How to test?
- [x] Unit test
- [x] Jenkins
http://10.112.231.51:18889/view/BigDL-PR-Validation/job/BigDL-Chronos-PR-Validation/706/
http://10.112.231.51:18889/job/BigDL-PRNV-chronos-Python-Spark-2.4-py37-Horovod/727/


python/chronos/src/bigdl/chronos/forecaster/autoformer_forecaster.py
```diff
@@ -420,12 +420,14 @@ def load(self, checkpoint_file):
...
         self.internal = AutoFormer.load_from_checkpoint(checkpoint_file, configs=args)
 
-    def save(self, checkpoint_file):
+    def save(self, checkpoint_file, tune=False):
         """
         save the forecaster.
 
         :param checkpoint_file: The checkpoint file location you want to load the forecaster.
         """
+        if tune:
```
TheaperDeng, line 429: I think there is a `self.use_hpo`, could we just use that member so that we don't need our users to explicitly set `tune`?

rnwang04: Good catch! Thanks, done.

TheaperDeng (approved):
LGTM

TheaperDeng (approved):
LGTM

PR #5379, opened 2022-08-12
PPML add pytorch incremental training and tutorial

Litchilitchy:
### 1. Why the change?
To support pytorch incremental training and add it in tutorial.
### 2. User API changes
client should call `estimator.save_server_model(path)` to trigger the server saving model to `path`, and call `estimator.load_server_model(path)` to trigger the server loading model from `path`

Clients have to save/load client models using `torch.save` and `torch.load`

### 3. Summary of the change 
Some related logics are added. 
* During first uploading from client, the meta (loss & optimizer) would be saved to disk
* When loading model, the meta would be loaded afterwards to bind the optimizer to the model
* The args between server and aggregator is changed to a `conf` which is a Python dict

The usage is added in tutorial doc. In tutorial code, client 1 would call `save_server_model` after training and call `load_server_model` before the second time training.
### 4. How to test?
Use original unittests, and example tests would be added later


jason-dai:
> ### 2. User API changes
> When start FL Server, need to specify `pytorchServerModelPath` in `ppml-conf.yaml`, and server would save the serve model automatically (100 aggregations per save by default)
> 
> Clients have to save/load client models using `torch.save` and `torch.load`

Instead of using config file, specify it in actual code. We should only use configure file to set the configurations for server, not the code logic.

jason-dai:
Need to update the description for the new APIs

Litchilitchy:
Tutorial doc and tutorial code updated

jason-dai:
> Tutorial doc and tutorial code updated

Add to PR description too

jason-dai:
> ### 2. User API changes
> client should call `estimator.save_server_model(path)` to trigger the server saving model to `path`, and call `estimator.load_server_model(path)` to trigger the server loading model from `path`
> 
> Clients have to save/load client models using `torch.save` and `torch.load`
> 

How to you regularly checkpoint (save) the model during training?

Litchilitchy:
> > ### 2. User API changes
> > client should call `estimator.save_server_model(path)` to trigger the server saving model to `path`, and call `estimator.load_server_model(path)` to trigger the server loading model from `path`
> > Clients have to save/load client models using `torch.save` and `torch.load`
> 
> How to you regularly checkpoint (save) the model during training?

For client model or server model or both?

Litchilitchy:
updated in [latest commit](https://github.com/intel-analytics/BigDL/pull/5379/files#diff-e86b581ea2608ba603b0591e65aad05110d37a5364cf6a7cfc06ec276e2e3689)

jason-dai:
It is not clear to me how the user will use this feature; you need to include example codes in the PR description for others to review the APIs.

jason-dai:
> 
> ```
> estimator = Estimator.from_torch(...,
>                         server_model_path='/path/to/server_model',
>                         client_model_path='/path/to/client_model')
> ```
>
>
> Client model and server model would be saved automatically after each epoch as long as user specifies arguments

What if the `server_model_path` in different parties are different? I think one and only one party is allowed to specify server model related arguments.

> When these paths are specified, the application would also try to load the respective model from path (client and server model) when this estimator is created
> 

I don't think this behavior is consistent with Orca or other estimator; users need to explicitly load the model instead.

> client could also explicitly call `estimator.save_server_model(path)` to trigger the server saving model to `path`, and call `estimator.load_server_model(path)` to trigger the server loading model from `path`
> 
> For client model, user can also explicitly save/load client models using native Pytorch method `torch.save(estimator.model, path)` and `torch.load(path)`
> 

How exactly does users do that (to resume training or predict)? Please show code examples.

jason-dai:
> * server would only take the `server_model_path` at the first time, and the ones after that would be ignored

Not only `server_model_path`, but also `server_model`; you can make it simple by accepting the first one, or party with id1, etc.

> #### save/load model
> user should call `load_server_model` or `load_model` explicitly to load model, and pass a `client_model_path` and a `server_model_path` when creating estimator
> 

It's not clear to what exactly you mean here; show a code example on how to load a model for prediction or resuming training.

PR #5380, opened 2022-08-12
fix bugs of regression test

MeouSker77:
## Description

fix bugs of regression test

* now the `IS_PR` environment can be passed into workload
* now it can checkout to correct branch when triggered by `APRT`


MeouSker77:
APRT

MeouSker77:
APRT

PR #5382, opened 2022-08-12
[PPML] add occlum driver and excutor JVM config

hzjane:
## Description
It will cost double memory if the jvm config is not set.

### 1. Why the change?
To enables reduce the native memory use.

### 2. User API changes
Add two new config in *.yaml .

### 3. Summary of the change 
Add two new config.

### 4. How to test?
- [ ] Application test

### DIFF
--- a/ppml/trusted-big-data-ml/scala/docker-occlum/entrypoint.sh
+++ b/ppml/trusted-big-data-ml/scala/docker-occlum/entrypoint.sh
@@ -133,7 +133,7 @@ case "$SPARK_K8S_CMD" in
             --hostname $SPARK_EXECUTOR_POD_IP
             )
     else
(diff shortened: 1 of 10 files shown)


ppml/trusted-big-data-ml/scala/docker-occlum/entrypoint.sh
```diff
@@ -55,6 +55,11 @@ fi
 
 case "$SPARK_K8S_CMD" in
   driver)
+    echo "SGX_LOG_LEVEL $SGX_LOG_LEVEL" && \
```
qiyuangong, line 58: Move echo before case. That will print all

qiyuangong (approved):
LGTM

PR #5385, opened 2022-08-12
support convert rdd of pandas df to spark df with arrow

dding3:
## Description
support convert shards to spark df with arrow optim.

### 1. Why the change?
It's used to boost performance of convert shards to spark df.
perf result details: 20 cores, 80g mem, csv file with 1.1g(5 million rows with 35 cols).
data_shard.to_spark_df_without_arrow: 5.12 s
data_shard.to_spark_df_with_arrow: 1.26 s

### 2. User API changes
N/A

### 3. Summary of the change 
1. wrapper several Spark private sql api into orca
2. support convert shards to spark df with arrow. The workflow is:
    a. convert pandas df to arrow stream then dump to file
    b. jvm read the file and generate a java data frame
    c. python wrapped it into a python spark df

### 4. How to test?
Application test

Test code can be found at:
root@clx001:/home/ding/pandas_perf/shards_to_spark_df.py


jason-dai:
Please collect comparable perf data through panda udf in spark

python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -13,18 +13,25 @@
...
 from bigdl.orca import OrcaContext
 from bigdl.dllib.nncontext import init_nncontext, ZooContext
-from bigdl.dllib.utils.common import get_node_and_core_number
+from bigdl.dllib.utils.common import *
 from bigdl.dllib.utils import nest
 from bigdl.dllib.utils.log4Error import *
 
 import numpy as np
 
+import pyarrow as pa
```
jason-dai, line 27: import arrow inside `try` below?

python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -542,18 +549,64 @@ def f(iter):
...
+                        arrow_schema = pa.Schema.from_pandas(pdf, preserve_index=False)
+                        struct = StructType()
+                        for name, field in zip(schema, arrow_schema):
+                            struct.add(
+                                name, from_arrow_type(field.type), nullable=field.nullable
+                            )
+                        schema = struct
+                        return [schema]
+
+            schema = self.rdd.mapPartitions(getSchemaStructType).first()
```
jason-dai, line 576: van you use `self.get_schema()` to get the pandas schema instead of `mapPartitions`?

python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -542,18 +549,64 @@ def f(iter):
...
+                    finally:
+                        tempFile.close()
+                    return [tempFile.name]
+
+            jiter = self.rdd.mapPartitions(f)
+            from bigdl.dllib.utils.file_utils import callZooFunc
+            df = callZooFunc("float", "orcaToDataFrame", jiter, schema.json(), sqlContext)
+            return df
+        except Exception as e:
+            print("Failed to convert shards to spark dataframe with shards with arrow optim,"
```
jason-dai, line 606: This message is not very clear

python/orca/src/bigdl/orca/data/shard.py
dding3: `mapPartitions` will return pyspark schema(`structtype`) which is needed to create df

jason-dai: > `mapPartitions` will return pyspark schema(`structtype`) which is needed to create spark df

You may create pyspark schema `structtype` from the pandas schema returned by `self.get_schema()`, yes?

dding3: 'self.get_schema()' takes more time than create `structtype` by `mapPartitions`, 0.37s VS 0.2 s, if we create `structtype` from 'self.get_schema()', it will take longer.

jason-dai: > 'self.get_schema()' takes more time than create `structtype` by `mapPartitions`, 0.37s VS 0.2 s, if we create `structtype` from 'self.get_schema()', it will take longer.

How's 'self.get_schema()' implemented? You can fix that method.

python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -170,7 +173,8 @@ def transform(iter, func, *args):
                 yield func(x, *args)
 
         transformed_shard = SparkXShards(self.rdd.mapPartitions(lambda iter:
-                                                                transform(iter, func, *args)))
+                                                                transform(iter, func, *args)),
+                                         class_name = self.class_name)
```
jason-dai: this is not correct - the transformed class maybe different from the original class

python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -155,6 +155,9 @@ def __init__(self, rdd, transient=False):
         if self.eager:
             self.compute()
         self.type = {}
+        if class_name:
+            self.type['class_name'] = class_name
```
jason-dai, line 159: where is it used?

jason-dai:
Please also update the description

python/orca/src/bigdl/orca/data/shard.py
dding3: in many shards operations, we will check if its class_name is `pandas.core.frame.DataFrame`. eg, https://github.com/intel-analytics/BigDL/blob/main/python/orca/src/bigdl/orca/data/shard.py#L536

jason-dai: OK; you may want to optimize `_get_class_name` and `_get_schema` (e.g., using one `rdd.map` instead of two).

python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -532,7 +540,7 @@ def zip(self, other):
                               "The two SparkXShards should have the same number of elements "
                               "in each partition")
 
-    def to_spark_df(self):
+    def to_spark_df_without_arrow(self):
```
jason-dai, line 543: change to `_to_spark_df_without_arrow`

python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -532,7 +540,7 @@ def zip(self, other):
...
         if self._get_class_name() != 'pandas.core.frame.DataFrame':
```
jason-dai, line 544: no need to check here as it is already checked in `to_spark_df`

dding3:
most jekins pass. http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-Orca-PR-Validation/1015/

The failed one should not be related to this pr.

The failed orca tutorials shards is caused by unexpected tensorflow version `2.7.0`, not related to this pr.

PR #5392, opened 2022-08-15
Automatically unwrap attributes access of _LiteModule

MeouSker77:
## Description

Automatically unwrap attributes access of `_LiteModule` and remove redundant documents

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/torch_nano.py
+++ b/python/nano/src/bigdl/nano/pytorch/torch_nano.py
@@ -34,6 +34,15 @@ from bigdl.nano.pytorch.strategies import create_IPEXStrategy, DDPSpawnStrategy,
     DDPSubprocessStrategy, create_RayStrategy
 
 
+class _TorchNanoModule(_LiteModule):
+    def __getattr__(self, name: str):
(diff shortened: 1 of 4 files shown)


python/nano/src/bigdl/nano/pytorch/torch_nano.py
```diff
@@ -34,6 +34,15 @@
     DDPSubprocessStrategy, create_RayStrategy
 
 
+class _TorchNanoModule(_LiteModule):
+    def __getattr__(self, name: str):
+        # automatically unwrap attributes access of _LiteModule
+        try:
+            return super().__getattr__(name)
```
jason-dai, line 41: will `__getattr__` be invoked if the attribute is found in `super()`? 
https://docs.python.org/3.7/reference/datamodel.html#customizing-attribute-access

MeouSker77: no, it won't be invoked

the `__getattr__` method of `_TorchNanoModule` will be invoked only when the attribute is not in it and all its super classes (i.e. `_LiteModule`, `nn.Module`)

PR #5402, opened 2022-08-15
Chronos : add regression functions for forecast metrics

Chaselzxy:
## Description

Rewrites the regression functions whose input and output are Numpy. 
### 1. Why the change?

To facilitate customers to use metrics without torchmetrics. Only Numpy is required to complete relevant operations.

### 3. Summary of the change 
Rewrites the regression functions whose input and output are Numpy according to the calculation formula of relevant regression functions.

### 4. How to test?
- [x] Unit test

http://10.112.231.51:18889/job/BigDL-PRNV-chronos-Python-Spark-2.4-py37-Horovod/814/
http://10.112.231.51:18889/job/BigDL-Chronos-PR-Validation/784/

### DIFF
--- a/python/chronos/src/bigdl/chronos/metric/forecast_metrics.py
+++ b/python/chronos/src/bigdl/chronos/metric/forecast_metrics.py
@@ -14,13 +14,8 @@
 # limitations under the License.
 #
 
-import torch
-import numpy as np
-from torch import Tensor
 from numpy import ndarray
-from functools import partial
-from torchmetrics.functional import mean_squared_error, mean_absolute_error,\
-    mean_absolute_percentage_error, r2_score
+import numpy as np
 from bigdl.nano.utils.log4Error import invalidInputError
 from timeit import repeat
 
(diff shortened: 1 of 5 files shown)


TheaperDeng:
@liangs6212 please also have a look

python/chronos/src/bigdl/chronos/metric/forecast_metrics.py
```diff
@@ -109,26 +186,15 @@ def evaluate(metrics, y_true, y_pred, aggregate='mean'):
...
-                metric_func = TORCHMETRICS_REGRESSION_MAP[metric]
             if len(original_shape) in [2, 3] and aggregate is None:
-                res = torch.zeros(y_true.shape[-1])
+                res = np.zeros(y_true.shape[-1])
                 for i in range(y_true.shape[-1]):
-                    if callable(metric):
-                        res[i] = torch.from_numpy(metric_func(y_true[..., i], y_pred[..., i]))
-                    else:
-                        res[i] = metric_func(y_pred[..., i], y_true[..., i])
+                    res[i] = eval(metric)(y_true[..., i], y_pred[..., i])
```
TheaperDeng, line 192: we should never use `eval` function since it might be dangerous and vulnerable. just make `REGRESSION_MAP` a dict e.g. `{'mae': mae, 'mse': mse, ...}`

TheaperDeng (commented):
others LGTM

python/chronos/src/bigdl/chronos/metric/forecast_metrics.py
```diff
@@ -14,36 +14,123 @@
...
+        \\text{MAE} = \\frac{1}{n}\\sum_{t=1}^n |y_t-\\hat{y_t}|
+    :param y_label: Array-like of shape = (n_samples, \*).
+           Ground truth (correct) target values.
+    :param y_predict: Array-like of shape = (n_samples, \*).
+           Estimated target values.
+    :return: Ndarray of floats.
+             An array of non-negative floating point values (the best value is 0.0).
+    """
+    y_label = np.array(y_label)
+    y_predict = np.array(y_predict)
```
liangs6212, line 39: We can delete all np.array, because ln89 will check the data type.
If you need to convert the input data type, implement it with a function.

TheaperDeng (approved):
LGTM, pls run jenkins

PR #5404, opened 2022-08-16
add new ways to create PPMLContext in python

PatrickkZ:
## Description

provide more ways to create PPMLContext in Python API

### 1. Why the change?

In Scala API, there are many ways to create PPMLContext, but Python API provide only one way to create. so we add more ways in Python API to keep consistent with Scala API.

### 2. User API changes

Before, the only way to create a PPMLContext in Python is to provide 2 parameters(app_name and ppml_args)
```python
sc = PPMLContext("MyApp", args)
```

Now, there are more ways
- only provide a app_name
  ```python
  sc = PPMLContext("MyApp")
  ```
  When you don't need to read/write encrypted files, you can use this way

- provide app_name and ppml_args(as before)

- provide app_name, ppml_args and a sparkConf
  If you need to set Spark configurations, you can provide a `SparkConf`  to create a `PPMLContext`.
  ```python
  conf = SparkConf()
  conf.setMaster("local[4]")
  sc = PPMLContext("MyApp", ppml_args, conf)
  ```

### 3. Summary of the change 

provide more ways to create PPMLContext in Python API
### 4. How to test?
- [ ] N/A
- [x] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/python/ppml/src/bigdl/ppml/README.md
+++ b/python/ppml/src/bigdl/ppml/README.md
@@ -8,7 +8,19 @@ This is a tutorial about how to use `PPMLContext` in python to read/write files 
 
 So before you read/write files, you need to create a PPMLConext first. 
 
-#### 1.1 create with app_name & ppml_args
+#### 1.1 create with app_name
+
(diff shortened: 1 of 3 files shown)


glorysdj (approved):
LGTM

PR #5406, opened 2022-08-16
notify user after running regression test

MeouSker77:
## Description

- now it will automatically notify user after running regression test triggered by pr comment
- fix the bug that it cannot checkout to the correct branch when trigged by a comment in a closed pr


MeouSker77:
APRT

MeouSker77:
APRT

glorysdj:
@MeouSker77 Your benchmark <https://github.com/intel-analytics/BigDL/pull/5406#issuecomment-1216266272> has finished,             see <https://github.com/intel-analytics/BigDL/actions/runs/2866160190> for details

Use `uuid=8a83e382-1d39-11ed-b6dd-ea16527ea10a` to query the result

PR #5409, opened 2022-08-16
[PPML] Remove private key generation in gramine Dockerfile

Uxito-Ada:
## Description

Remove the automated enclave-key generation when building the image, which should use the key mounted by users.

### 1. Why the change?

The `enclave-key.pem` refers to the mounted one, rather than the demo key generated by `gramine-sgx-gen-private-key` API.

### 2. User API changes

None.

### 3. Summary of the change 

Remove private key generation in gramine Dockerfile.

### 4. How to test?

Unit test.


glorysdj (approved):
LGTM

PR #5418, opened 2022-08-16
Chronos: the ut test_init_doppelganer() in simulator is fixed

plusbang:
## Description

Considering the environment variable FTP_URI is only available in Jenkins, the unit test test_init_doppelganer() in simulator can be directly skipped when not in Jenkins.


TheaperDeng (approved):
LGTM

PR #5420, opened 2022-08-16
uncache shards if it's converted to spark df

dding3:
## Description

This PR is used to reduce memory usage. If shards is converted to spark df, most likely we will create a new shards, we can uncache the original shards to save memory.

### 1. How to test?
- Application test


jason-dai:
Does this solve the conversion perf issue?

dding3:
> Does this solve the conversion perf issue?

No, it's for memory usage perf issue.

dding3:
most jekins pass. http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-Orca-PR-Validation/945/. The failed one is not related to this pr.

PR #5421, opened 2022-08-17
Chronos: Test Installation options(autots)

liangs6212:
## Description

Add later

### 1. Why the change?

Add later

### 2. User API changes

No changed

### 3. Summary of the change 

Add later

### 4. How to test?
- [ ] Unit test


python/chronos/src/bigdl/chronos/autots/__init__.py
```diff
@@ -24,8 +24,7 @@
                   "in your bash terminal")
 
 try:
-    # TODO: make this a LazyImport
-    from .autotsestimator import AutoTSEstimator
     from .tspipeline import TSPipeline
-except ImportError:
+    from .autotsestimator import AutoTSEstimator
+except:
```
TheaperDeng, line 29: why we remove the `ImportError` limit here?

TheaperDeng (commented):
others LGTM

python/chronos/src/bigdl/chronos/autots/__init__.py
liangs6212: undo this change.

PR #5427, opened 2022-08-17
[Nano] add JIT, channel_last to LightningModule

zhentaocc:
## Description

This is to enable JIT, channel_last in Nano.

### 1. Why the change?

JIT, channel_last can optimize performance for training and inference.

### 2. User API changes

```python
model:LightningModule
model.channel_last()
model.jit()
```

### 3. Summary of the change 

enable JIT, channel_last in Nano

### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/lightning.py
+++ b/python/nano/src/bigdl/nano/pytorch/lightning.py
@@ -36,9 +36,14 @@ class LightningModule(pl.LightningModule):
     a LightningModule.
     """
 
-    def __init__(self, model: nn.Module, loss: _Loss = None, optimizer: Optimizer = None,
+    def __init__(self, model: nn.Module,
+                 loss: _Loss = None,
+                 optimizer: Optimizer = None,
                  scheduler: _LRScheduler = None,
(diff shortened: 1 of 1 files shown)


jason-dai:
We also have such support in Trainer; what's the difference?

zhentaocc:
> We also have such support in Trainer; what's the difference?

The implementation in Trainer is in  https://github.com/intel-analytics/BigDL/blob/main/python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py.
 JIT and channel_last do not actually depend on IPEX.

1.  From code structure, JIT and channel_last should be binded to pytorch/pytorch-lightning instead of IPEX.
2. When users are not intended to use ipex, to use JIT and channel_last, users are still required to install ipex. (But seems that ipex is installed for nano by default right now.) It's not that good to keep this dependency.
3. As tested before, even without ipex, Pytorch+channel_last can obtain great perf improvement for cv workloads. So I think these two features should be enabled for all pytorch backend instead of ipex only.

zhentaocc:
Still figuring out a best way to enable JIT and channel_last.

PR #5428, opened 2022-08-17
PPML: add k8s to Gramine Docker

ShanSimu:
## Description

Add k8s related files to Gramine dockerfile and test the function of k8s.

### 1. Why the change?

We need Gramine Docker to support k8s and make sure k8s works well.

### 2. User API changes

Same as Graphene.

### 3. Summary of the change 

Add k8s related files to Gramine Docker, including support and start files. Test various modes of k8s in the sgx environment, including local, client and cluster modes. Test different startup methods of k8s, including native command line, script bigdl-ppml-submit.sh written for users, and helm chart.

### 4. How to test?

Test methods are written in the README.


glorysdj (approved):
LGTM

PR #5431, opened 2022-08-17
hotfix akka-actor scala version

Le-Zheng:
## Description

support akka-actor for both Scala 2.11 and 2.12


Le-Zheng:
http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-PR-Dllib-Validation/597/

PR #5433, opened 2022-08-17
[PPML] Fix spark jar url missing in build-docker-image

Uxito-Ada:
## Description

Append SPARK_JAR_URL in build-docker-image API of both graphene and gramine.

### 1. Why the change?

SPARK_JAR_URL is missed in build-docker-image.sh script, which is necessary in Dockerfile and this will cause a build failure when users build image through this script.

### 2. User API changes

Users should specify SPARK_JAR_URL in build-docker-image.sh

### 3. Summary of the change 

Fix spark jar url missing in build-docker-image

### 4. How to test?

Unit test.


glorysdj (approved):
LGTM

PR #5437, opened 2022-08-17
import so lib from nano not download it

zhanglirong1999:
- Modify some comment.
- Nano already has installed libturbojpeg.so.0.2.0 lib , so remove the additional installation.

### DIFF
--- a/python/nano/tutorial/training/pytorch-lightning/lightning_cv_data_pipeline.py
+++ b/python/nano/tutorial/training/pytorch-lightning/lightning_cv_data_pipeline.py
@@ -26,13 +26,9 @@ import pytorch_lightning as pl
 from os.path import split, join, realpath
 import cv2
 import os
-from logging import warning
 from PIL import Image
-import urllib.request
 import os
-import stat
 
-LIB_URL = "https://github.com/leonardozcm/libjpeg-turbo/releases/download/2.1.1/libturbojpeg.so.0.2.0"
 # These images in the pet dataset that don't have a proper format. 
 # Some of them are actually .png files instead .jpg, 
 # even though they are in .jpg extension.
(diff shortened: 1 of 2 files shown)


jason-dai:
Add description

python/nano/tutorial/training/pytorch-lightning/lightning_cv_data_pipeline.py
```diff
@@ -192,6 +192,8 @@ def create_dataloaders():
 if __name__ == "__main__":
     # get dataset
     model = MyLightningModule()
+    # BigDL-Nano can accelerate computer vision data pipelines
+    # by providing a drop-in replacement of torch_vision’s datasets and transforms
     train_loader, val_loader = create_dataloaders()
     # CV Data Pipelines
```
jason-dai, line 198: move the description below to above and combine with the description above

python/nano/tutorial/training/pytorch-lightning/lightning_cv_data_pipeline.py
```diff
@@ -15,26 +15,20 @@
 
 import torch
 import torchvision
-from turbojpeg import TurboJPEG, TJPF_GRAY
+from turbojpeg import TurboJPEG
```
jason-dai, line 18: shall we add built-in support for JPEP decode support in nano? @MeouSker77

MeouSker77: We already have `ImageFolder`, which has built-in support. But this Oxford Pet dataset has some gif pictures, so it cannot be used with `ImageFolder`. We may need to implement our Oxford Pet dataset, and warp these codes into our Oxford Pet dataset, then this example can use our Oxford Pet dataset directly.

MeouSker77:
This PR will be closed due to a new version #5448

PR #5441, opened 2022-08-17
update the download url of .so files

MeouSker77:
## Description

- Update the download url of the .so files, now these .so files are all downloaded from repos under analytics-zoo
- Rebuild these .so files with latest source codes


MeouSker77:
http://10.112.231.51:18889/view/BigDL-PR-Validation/job/ZOO-PR-NanoTests/357/

yangw1234:
LGTM

PR #5451, opened 2022-08-18
fix the trigger condition of checkout steps

MeouSker77:
## Description

Fix the trigger condition of checkout steps


MeouSker77:
APRT

glorysdj:
@MeouSker77 Your benchmark <https://github.com/intel-analytics/BigDL/pull/5451#issuecomment-1218914460> has finished,             see <https://github.com/intel-analytics/BigDL/actions/runs/2879531471> for details

Use `uuid=45443d6a-1e9e-11ed-bb45-ea16527ea10a` to query the result

PR #5459, opened 2022-08-18
Change default cores of init_orca_context

zhanglirong1999:
- Set core to None by default
- If in local mode, set core="*", in other mode, set core=2

Issue fix:
init_orca_context default cores=2, and users may not be aware of this setting and may just call it with default parameters. And in this case if user compare the local performance between the original TF/PyTorch script with the Orca version, they will find much performance drop since Orca is only using two cores while the original script will use all cores in most cases.
Changing to "*" for local mode and add print/warnings.

### DIFF
--- a/python/orca/src/bigdl/orca/common.py
+++ b/python/orca/src/bigdl/orca/common.py
@@ -172,7 +172,7 @@ def _check_python_micro_version():
                           f"with micro version >= 10 (e.g. 3.{sys.version_info[1]}.10)")
 
 
-def init_orca_context(cluster_mode=None, runtime="spark", cores=2, memory="2g", num_nodes=1,
+def init_orca_context(cluster_mode=None, runtime="spark", cores=None, memory="2g", num_nodes=1,
                       init_ray_on_spark=False, **kwargs):
     """
     Creates or gets a SparkContext for different Spark cluster modes (and launch Ray services
@@ -210,6 +210,11 @@ def init_orca_context(cluster_mode=None, runtime="spark", cores=2, memory="2g", 
     print("Initializing orca context")
     import atexit
     atexit.register(stop_orca_context)
+    if not cores:
+        if cluster_mode == "local":
+            cores = "*"
+        else:
+            cores = 2
     if runtime == "ray":
         invalidInputError(cluster_mode is None,
                           "Currently, cluster_mode is not supported for ray runtime and"
--- a/python/orca/src/bigdl/orca/common.py
+++ b/python/orca/src/bigdl/orca/common.py
@@ -222,7 +222,7 @@ def init_orca_context(cluster_mode=None, runtime="spark", cores=None, memory="2g
             cores = 2
             import warnings
             warnings.warn("Cores is not specified, using 2 cores per node by default.", Warning)
-            
+
     if runtime == "ray":
         invalidInputError(cluster_mode is None,
                           "Currently, cluster_mode is not supported for ray runtime and"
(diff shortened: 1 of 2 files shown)


python/orca/src/bigdl/orca/common.py
```diff
@@ -210,6 +214,15 @@ def init_orca_context(cluster_mode=None, runtime="spark", cores=2, memory="2g",
...
     import atexit
     atexit.register(stop_orca_context)
+    if not cores:
+        if cluster_mode == "local" and runtime == "spark":
+            cores = "*"
+            print("Cluster_mode is local and runtime is spark, default to be '*'.")
+        else:
+            cores = 2
+            import warnings
+            warnings.warn("No cores are specified, default to be 2.", Warning)
```
hkvision, line 224: cores is not specified, using 2 cores per node by default

python/orca/src/bigdl/orca/common.py
```diff
@@ -195,7 +196,10 @@ def init_orca_context(cluster_mode=None, runtime="spark", cores=2, memory="2g",
            For "k8s-client" and "k8s-cluster", you are supposed to additionally specify the
            arguments master and container_image.
     :param runtime: The runtime for backend. One of "ray" and "spark". Default to be "spark".
-    :param cores: The number of cores to be used on each node. Default to be 2.
+    :param cores: The number of cores to be used on each node.
+           For spark local mode, default to use all the cores on the node.
```
hkvision, line 200: spark-> Spark

python/orca/src/bigdl/orca/common.py
```diff
@@ -195,7 +196,10 @@ def init_orca_context(cluster_mode=None, runtime="spark", cores=2, memory="2g",
...
+           For other cluster_mode, default to use 2 cores per node.
+           You are highly recommended to set this value by yourself.
```
hkvision, line 202: add `instead of using the default one`

python/orca/src/bigdl/orca/common.py
```diff
@@ -210,6 +214,15 @@ def init_orca_context(cluster_mode=None, runtime="spark", cores=2, memory="2g",
     print("Initializing orca context")
     import atexit
     atexit.register(stop_orca_context)
+    if not cores:
+        if cluster_mode == "local" and runtime == "spark":
+            cores = "*"
+            print("For spark local mode, default to use all the cores on the node.")
```
hkvision, line 220: spark -> Spark

python/orca/src/bigdl/orca/common.py
```diff
@@ -210,6 +214,15 @@ def init_orca_context(cluster_mode=None, runtime="spark", cores=2, memory="2g",
...
+        else:
+            cores = 2
+            import warnings
+            warnings.warn("Cores is not specified, using 2 cores per node by default.", Warning)
```
hkvision, line 224: remove `is`

hkvision:
@jason-dai Is the change OK?

python/orca/src/bigdl/orca/common.py
```diff
@@ -195,7 +196,11 @@ def init_orca_context(cluster_mode=None, runtime="spark", cores=2, memory="2g",
            For "k8s-client" and "k8s-cluster", you are supposed to additionally specify the
            arguments master and container_image.
     :param runtime: The runtime for backend. One of "ray" and "spark". Default to be "spark".
-    :param cores: The number of cores to be used on each node. Default to be 2.
+    :param cores: The number of cores to be used on each node.
+           For Spark local mode, default to use all the cores on the node.
```
hkvision, line 200: remove `mode` to align with Line 206?

hkvision (approved):

python/orca/src/bigdl/orca/common.py
```diff
@@ -195,7 +196,11 @@ def init_orca_context(cluster_mode=None, runtime="spark", cores=2, memory="2g",
...
+           For other cluster_mode, default to use 2 cores per node.
+           You are highly recommended to set this value by yourself,
+           instead of using the default one.
     :param memory: The memory allocated for each node. Default to be '2g'.
     :param num_nodes: The number of nodes to be used in the cluster. Default to be 1.
            For Spark local, num_nodes should always be 1 and you don't need to change it.
```
hkvision, line 206: add mode to Spark local?

hkvision:
jenkins http://10.112.231.51:18888/job/BigDL-Orca-PR-Validation/940/
some failures are irrelevant

PR #5470, opened 2022-08-19
Nano: update ipex_bf16_inference_model

y199387:
## Description

IPEX BF16 optimization needs cpu support AVX512 at least. Update `PytorchIPEXJITBF16Model` to check if bfloat16 instructions are available.

### 1. Why this change
IPEX BF16 weight prepack needs the cpu support avx512_core.

### 2. User API changes
N/A

### 3. Summary of the change
Add cpu instructions set check in `PytorchIPEXJITBF16Model`.

### 4. How to test
- [x] Unit test
- [x] Jenkins
http://10.112.231.51:18889/job/Chronos-PR-Validation-Spark-2.4/749/
http://10.112.231.51:18889/job/BigDL-PRVN-chronos-Python-Spark-2.4-py37-onnxrt1.6/705/
http://10.112.231.51:18889/job/BigDL-PRNV-chronos-Python-Spark-2.4-py37-Horovod/765/


yangw1234:
@zhentaocc could you help review this pr?

zhentaocc:
What if users try to use ipex bf16 on a non-avx512 machine? Will they get a specific error? I think we should catch that error ahead of time to tell users how to fix.

python/nano/src/bigdl/nano/deps/ipex/ipex_inference_bf16_model.py
```diff
@@ -38,13 +48,90 @@ def __init__(self, model, input_sample=None, use_ipex=False,
...
+
+        :return: The bool value of ALLOW_NON_BF16_ISA
+        """
+        return os.environ.get("ALLOW_NON_BF16_ISA", None) == '1'
+
+    @property
+    def _has_bf16_isa(self):
+        """Indicator to verify if bf16 instructions are available."""
+        msg = subprocess.check_output(["lscpu"]).decode("utf-8")
+        return 'avx512' in msg or 'amx' in msg
```
zhentaocc, line 74: seems that this function is not checking bf16?

python/nano/src/bigdl/nano/deps/ipex/ipex_inference_bf16_model.py
```diff
@@ -38,13 +48,90 @@ def __init__(self, model, input_sample=None, use_ipex=False,
...
-    @autocast()
+    @property
+    def _allow_non_bf16(self):
+        """
+        ALLOW_NON_BF16_ISA indicates if we restrict bf16 instructions support to be available.
+        ALLOW_NON_BF16_ISA='1' sometimes helps debug and test cases without AVX512 or AMX
+
+        :return: The bool value of ALLOW_NON_BF16_ISA
+        """
+        return os.environ.get("ALLOW_NON_BF16_ISA", None) == '1'
```
zhentaocc, line 68: I think this is not necessary? If running on non-avx512 machine, it's better to raise error.

python/nano/src/bigdl/nano/deps/ipex/ipex_inference_bf16_model.py
```diff
@@ -38,13 +48,90 @@ def __init__(self, model, input_sample=None, use_ipex=False,
...
+        max_bf16_isa = None
+        # IPEX 1.11 BF16 support AVX512
+        # IPEX 1.12 BF16 support AMX, AVX512_BF16, AVX512_VNNI, AVX512
+        if 'amx_bf16' in dnnl_log:
+            max_bf16_isa = 'AMX'
+        elif 'avx512_core_bf16' in dnnl_log:
+            max_bf16_isa = 'AVX512_BF16'
+        elif 'avx512_core' in dnnl_log:
+            max_bf16_isa = 'AVX512'
+        return max_bf16_isa
```
zhentaocc, line 95: IPEX doen't require BF16 instructions, right? I think we don't need this check.

y199387: Sure, removed redundant method and related unit test.

python/nano/test/pytorch/tests/test_bf16_ipex.py
```diff
@@ -35,30 +36,43 @@ def test_bf16_pytorch_less_1_10(self):
             trainer.quantize(model, precision='bf16', use_ipex=True)
 
 
-class Pytorch1_12:
-    def test_bf16_common(self):
-        """
-        Debug mode. Allow run bf16 forward without bf16 instruction support.
-        """
+class CaseWithoutAVX512:
+    @patch('bigdl.nano.deps.ipex.ipex_inference_bf16_model.PytorchIPEXJITBF16Model._check_cpu_isa', new_callable=PropertyMock)
```
zhentaocc, line 40: `CaseWithoutAVX512` will be run only when `check_avx512() == False` which means `_check_cpu_isa=False` by default. Is this mock necessary?

zhentaocc (commented):
Some cases need to be covered:

1. pytorch 1.9, raise `pytorch version` error
2. pytorch >= 1.10(1.10, 1.11, 1.12) with non-avx512, raise `avx512 is required `
3. pytorch >= 1.10(1.10, 1.11, 1.12) with avx512, successful

python/nano/src/bigdl/nano/pytorch/amp/bfloat16.py
```diff
@@ -46,6 +48,40 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         return super().__exit__(exc_type, exc_val, exc_tb)
 
 
+class RedirectStream(object):
```
jason-dai, line 51: where is this used?

y199387: The contextmanager was used to capture the output of dnnl verbose output , because python `contextlib.redirect_stdout` only works for python package. I removed the BF16 check related method in `PytorchIPEXJITBF16Model` because it is not necessary for IPEX BF16. But this contextmanager is still useful for [`bigdl.nano.pytorch.amp.bfloat16.BF16Model`](https://github.com/intel-analytics/BigDL/blob/main/python/nano/src/bigdl/nano/pytorch/amp/bfloat16.py#L115) to verify if bf16 instructions are used, so I keep it in this pr.

PR #5471, opened 2022-08-19
Orca OpenVINO estimator add input order param & dataframe input support multi-output

cyita:
## Description

1. Add input order param in predict function.
2. Fix multi-output issue when input is a spark dataframe.
3. Return result shard instead of zipping (input, result) to reduce memory consumption.
4. Add & modify testcase.

### 1. Why the change?

1. Customer issue (Yidu cloud).
2 & 3. User group issue.

### 2. User API changes

`def predict(self, data, feature_cols=None, batch_size=4)` -> `def predict(self, data, feature_cols=None, batch_size=4, inputs=None)`

### 3. Summary of the change 

1. Add input order param in predict function.
2. Fix multi-output issue when input is a spark dataframe.
3. Return result shard instead of zipping (input, result) to reduce memory consumption.

### 4. How to test?
- [ ] Unit test


python/orca/src/bigdl/orca/learn/openvino/estimator.py
```diff
@@ -53,7 +53,7 @@ def fit(self, data, epochs, batch_size=32, feature_cols=None, label_cols=None,
         """
         invalidInputError(False, "not implemented")
 
-    def predict(self, data, feature_cols=None, batch_size=4):
+    def predict(self, data, feature_cols=None, batch_size=4, inputs=None):
```
jason-dai, line 56: Add Python Doc for `inputs`

python/orca/src/bigdl/orca/learn/openvino/estimator.py
```diff
@@ -63,6 +63,9 @@ def predict(self, data, feature_cols=None, batch_size=4):
                feature}, where feature(label) is a numpy array or a list of numpy arrays.
         :param feature_cols: Feature column name(s) of data. Only used when data is a Spark
                DataFrame. Default: None.
+        :param inputs: Str or List of str. The model input list(order related). Users can specify
```
jason-dai, line 66: maybe change to `input_cols`?

cyita:
Jenkins: http://10.112.231.51:18888/job/BigDL-Orca-PR-Validation/1058/

Issue #5466, opened 2022-08-18
[Nano] torch.save will raise FileNotFound if parent directory is not created 


zhentaocc:
### script to reproduce
```python
from bigdl.nano.pytorch import Trainer
from torchvision.models.mobilenetv3 import mobilenet_v3_small
trainer = Trainer(max_epochs=1)
model = mobilenet_v3_small(num_classes=10)
x = torch.rand((10, 3, 256, 256))

# save and load pytorch model
openvino_model = trainer.trace(model, accelerator='openvino', input_sample=x)
trainer.save(openvino_model, "parent/model")
```
If `parent` is not created, this will raise `FileNotFound`.

PR #5473, opened 2022-08-19
Nano: Fix torch.save will raise FileNotFound if parent directory is not created

hjzin:
## Description

Fix torch.save will raise FileNotFound if parent directory is not created.

### 1. Why the change?

releated issue: [issue-5466](https://github.com/intel-analytics/BigDL/issues/5466)

### 2. How to Test?
- [x] Jenkins
http://10.112.231.51:18889/view/BigDL-PR-Validation/job/BigDL-Chronos-PR-Validation/728/


TheaperDeng (approved):
LGTM, to be merged

PR #5474, opened 2022-08-19
[PPML] Fix graphene image nightly build failure

Uxito-Ada:
## Description

Move the proxy ARG declaration to the specific stage where it is used.

### 1. Why the change?

As the ARG scope is different among different docker versions, the pre-declaration at the beginning of Dockerfile caused a nightly-build failure.

Move the ARG to be under the specific stage can enable the right build-argvs pass.

### 2. User API changes

None.

### 3. Summary of the change 

Fix graphene image nightly build failure

### 4. How to test?

Github action job.

Has been tested with @Romanticoseu .


glorysdj (approved):
LGTM

PR #5477, opened 2022-08-19
Update chronos tutorial: Add filter.

AndyLuo1029:
## Description

Add the filter function in chronos tutorial page.

Preview page: https://andytestdoc.readthedocs.io/en/latest/doc/Chronos/QuickStart/index.html


.gitignore
```diff
@@ -1,50 +1,50 @@
...
+*/.cache-tests
+*/.classpath
+*/.project
+*/.settings/
+*.so
+__pycache__
+*.egg-info
+target
+build
+dist
```
TheaperDeng, line 50: I am not sure what's happening to this file. There seems to be no changes but only whitespace(invisible ones). To make sure nothing is broken, we should fix this problem before merging the PR.

AndyLuo1029:
Since the problem about .gitignore file can't be solved, I will open another pr and close this one to avoid unexpected changes.

PR #5486, opened 2022-08-19
add orca ray python UT  to github action

pinggao187:
## Description

add orca raypython UT  to github action


.github/workflows/nb-ut-orca-python-ray-py37-spark3.yml
```diff
@@ -0,0 +1,93 @@
...
+                  }
+                }
+              }
+          }]
+        mirrors: '[{"id": "ardaNexus", "name": "ardaNexus", "mirrorOf": "*", "url": "${NEXUS_URL}" }]'
+
+    - name: Setup Env
+      run: |
+        apt-get update
+        apt-get install wget
```
liu-shaojun, line 69: why is wget used?

PR #5489, opened 2022-08-21
Chronos: remove checking in autots import

TheaperDeng:
## Description

remove checking for autots since it will break colab running.


TheaperDeng:
http://10.112.231.51:18889/job/BigDL-Chronos-PR-Validation/733/

Issue #5335, opened 2022-08-08
Model optimization pipeline APIs


TheaperDeng:
## Background

Nano currently has a great collection of pytorch/tf inference acceleration methods, while our users might need an automatical (intellectural guided) pipeline to find which one is the best.

## Methodology

**Currently** we can have 3 workflows for our users who cares about the inference performance.

| Method             | Accuracy Drop      | Expected Acceleration Ratio | Retrain | Success Ratio |
| ------------------ | ------------------ | --------------------------- | ------- | ------------- |
| `Trainer.quantize` | True (except bf16) | 1~4X                        | False   | low           |
| `Trainer.trace`    | False              | 1~2X                        | False   | high          |
| `Trainer.search`   | True               | 1~20X                       | True    | medium        |

For this pipeline design, we will classify users to 2 catagories

1. Users bring a trained model (maybe loaded from checkpoint files), and would like to optimize this specific model in a short time.
   
   We will have a new API to find the best accelearted model for our users automatically, detailed API design is illustrated below. This new API will conver the original `Trainer.quantize` and `Trainer.trace`.

2. Users have a model definition and would like to find the best hyperparameter configuration to balance the accuracy and latency.
   
   `Trainer.search` will handle this case very easily, **we will not cover this part carefully in this issue**.

## API Design

Please find a prototype implementation in: 

This API is designed to be 

1. really easy to use without extra parameters

2. detailed acceleration strategy is completely hidden to our users.

```python
# bigdl.nano.pytorch.common_inference

def accelerate_inference(model,
                         training_data,
                         validation_data=None,
                         metric=None,
                         tollerated_accuracy_drop=None):
'''
:param model: A nn.module to be optimized
:param training_data: A pytorch dataloader for training dataset.
       Users should be careful with this parameter since this dataloader
       might be exposed to the model, which causing data leak. The
       batch_size of this dataloader is important as well, users may
       want to set it to the same batch size you may want to use the model
       in real deploy environment. E.g. batch size should be set to 1
       if you would like to use the accelerated model in an online service.
:param validation_data: (optional) A pytorch dataloader for accuracy evaluation
       This is only needed when users care about the possible accuracy drop.
:param metric: (optional) A callable object takes prediction and target
       and returns a accuracy value in this calling method `metric(pred, target)`
:param tollerated_accuracy_drop: (optional) a float represents the accuracy
       drop ratio that can be tollerated, such as 0.05 means we can accept a
[...]

PR #5490, opened 2022-08-22
Nano: Enhance optimizer pipeline

rnwang04:
### 1. Why the change?

https://github.com/intel-analytics/BigDL/issues/5335

### 2. User API changes

Obtain all optimization combinations
```
                       training_data=train_loader,
                       validation_data=test_loader, 
                       metric=torchmetrics.Accuracy(num_classes=10, top_k=1),
                       direction="max",
                       cpu_num=1,
                       logging=False)
for key, value in inference_opt.optimized_model_dict.items():
    print(key, value["latency"], value["accuracy"])
```
sample output:
```
original 1.6805797815322876 0.1682
fp32_ipex 1.7428427934646606 0.1682
bf16_ipex 1.857537031173706 0.1688
int8 4.562163352966309 0.1692
jit_fp32 1.113685965538025 0.1682
jit_fp32_ipex 1.1103034019470215 0.1682
jit_fp32_ipex_clast 0.8933722972869873 0.1682
openvino_fp32 0.798383355140686 0.1688
openvino_int8 0.860518217086792 0.162
onnxruntime_fp32 1.720300316810608 0.1682
onnxruntime_int8_qlinear 1.4863044023513794 0.169
```

Obtain the model with minimum latency
`acc_model, option = inference_opt.get_best_model()`

Obtain the model optimized by onnxruntime with minimum latency
`acc_model, option = inference_opt.get_best_model(accelerator="onnxruntime")`

Obtain the model optimized to int8 with minimum latency
`acc_model, option = inference_opt.get_best_model(precision="int8")`

Obtain the model with minimum latency and a metric drop of no more than 5%
`acc_model, option = inference_opt.get_best_model(accuracy_criterion=0.05)`

### 3. Summary of the change 
- [x] refactor code
- [x] support jit and channels_last
- [x] implement cpu_num and direction in `InferenceOptimizer.optimize`
- [x] implement `InferenceOptimizer.get_best_model`
- [x] clean output of openvino(fp32, int8)  `logging`
- [x] clean output of inc by set environment variable
- [x] example to show the full ppl (unittest)

### 4. How to test?
- [x] Unit test
- [x] Jenkins test
http://10.112.231.51:18889/view/BigDL-PR-Validation/job/ZOO-PR-NanoTests/369/

### 5. May need to do in next PR
- [ ] setting cpu num for openvino
- [ ] accuracy control in optimize()
- [ ] migrate optimizer.trace
- [ ] migrate optimizer.quantize


python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -73,8 +98,8 @@ def optimize(self, model,
                  training_data,
                  validation_data=None,
                  metric=None,
-                 direction=None,
-                 cpu_num=None):
+                 cpu_num: int = None,
+                 trials: int = 100):
```
TheaperDeng, line 102: `trails` may confuse our users with the "trial" in automl searching. Since this param is not used every time, we may make it longer and easier to understand such as `latency_sample_num`

TheaperDeng: `direction` can be moved here and only for record, since this is a "feature" according to the metric, and no users would like to filter a model that is "worse" than xxx metric value.

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -92,105 +117,202 @@ def optimize(self, model,
...
+                                                                  accelerator=accelerator,
+                                                                  use_ipex=use_ipex,
+                                                                  channels_last=use_channels_last,
+                                                                  input_sample=input_sample)
+                            else:
+                                accelerated_model = Trainer.trace(model=model,
+                                                                  accelerator=accelerator,
+                                                                  input_sample=input_sample,
+                                                                  # remove output of openvino
+                                                                  logging=False)
```
TheaperDeng, line 161: why are we split the calling signature here for "jit" and None vs Onnx and openvino

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -92,105 +117,202 @@ def optimize(self, model,
...
                     except Exception as e:
                         print(e)
+                        continue
 
                 # if precision is int8 or bf16, then we will use quantize method
-                if split_method[1] == "int8" or split_method[1] == "bf16":
-                    precision = split_method[1]
-                    q_method = None if len(split_method) < 4 else split_method[3]
+                elif precision in ("int8", "bf16"):
+                    ort_method = _detect_ort_method(method)
```
TheaperDeng, line 168: `_detect_quantization_method`? since this output will be assigned to `method` in `Trainer.quantize`

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -92,105 +117,202 @@ def optimize(self, model,
...
                     try:
-                        # TODO: remove the logging of quantization
-                        accelerated_model = Trainer.quantize(model=model,
+                        accelerated_model = Trainer.quantize(model=deepcopy(model),
```
TheaperDeng, line 170: Use `Optimizer.quantize`

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -92,105 +117,202 @@ def optimize(self, model,
...
                         print(e)
+                        continue
+
+                result_map[method] = {}
+
+                def func_test(model, input_sample):
+                    # with torch.cpu.amp.autocast():
+                    model(*input_sample)
+
+                torch.set_num_threads(cpu_num)
```
TheaperDeng, line 188: This setting is only useful for `accelerator=None`

TheaperDeng: For onnxruntime, you need to set onnxrt session opt during the trianing.
For openvino, we don't have this setting right now, we need to find a way to set this.

python/nano/test/pytorch/tests/test_inference_pipeline.py
```diff
@@ -0,0 +1,132 @@
...
+    '''
+    train_set = CIFAR10(root=dir, train=False,
+                        download=True, transform=transform)
+    # `subset` is the number of subsets. The larger the number, the smaller the training set.
+    mask = list(range(0, len(train_set), subset))
+    train_subset = torch.utils.data.Subset(train_set, mask)
+
+    data_loader = DataLoader(train_subset, batch_size=batch_size, shuffle=False,
+                             num_workers=num_workers)
+    return data_loader
```
TheaperDeng, line 73: I think you may sinmply import these utils functions here: https://github.com/intel-analytics/BigDL/blob/8f75766d0e9381e25721be9eacccd19585cbec15/python/nano/test/pytorch/utils/_train_torch_lightning.py

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -261,7 +390,9 @@ def _throughput_calculate_helper(iterrun, func, *args):
         st = time.time()
         func(*args)
         time_list.append(time.time() - st)
-    # TODO: remove first and last 10 percent
+    time_list.sort()
```
TheaperDeng, line 393: I think we don't need to sort here. if we sort this time list, then removing the top/least 10% data is meaningless, we will always find the median of the whole time list anyway.

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -220,7 +342,7 @@ def _openvino_checker():
     '''
     check if openvino-dev is installed
     '''
-    return not find_spec("openvino-dev") is None
+    return not find_spec("openvino") is None
```
TheaperDeng, line 345: I think we still need `openvino-dev` for pot?

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: done

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: done

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: Done

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: because channels_last for openvino will cause erorr in torch.export.

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: done

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: changed to np.mean. thanks.

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: done

python/nano/test/pytorch/tests/test_inference_pipeline.py
rnwang04: Done. thanks!

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
```diff
@@ -265,6 +265,7 @@ def quantize(model,  # remove the type requirement for type checking
                  max_trials: int = None,
                  input_sample=None,
                  onnxruntime_session_options=None,
+                 logging=True,
```
jason-dai, line 268: add type hint

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
```diff
@@ -419,6 +423,7 @@ def trace(model: nn.Module,
               accelerator=None,
               use_ipex=False,
               onnxruntime_session_options=None,
+              logging=True,
```
jason-dai, line 426: add type hint

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -14,48 +14,76 @@
...
+
+    def __init__(self, *args, **kwargs):
+        '''
+        initialize optimization option
+        '''
+        for option in _whole_acceleration_options:
+            setattr(self, option, kwargs.get(option, False))
+
+    def get_precision(self):
+        if self.inc:
```
jason-dai, line 53: how about pot?

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -14,48 +14,76 @@
...
-        "onnxruntime_int8_noipex_qlinear": {"inc": True, "ipex": False, "onnxruntime": True,
-                                            "openvino": False, "pot": False, "bf16": False},
-        "onnxruntime_int8_noipex_integer": {"inc": True, "ipex": False, "onnxruntime": True,
-                                            "openvino": False, "pot": False, "bf16": False},
-        "openvino_fp32_noipex": {"inc": False, "ipex": False, "onnxruntime": False,
-                                 "openvino": True, "pot": False, "bf16": False},
-        "openvino_int8_noipex": {"inc": False, "ipex": False, "onnxruntime": False,
-                                 "openvino": True, "pot": True, "bf16": False},
+        "original": AccelerationOption(),
+        "None_fp32_ipex": AccelerationOption(ipex=True),
```
jason-dai, line 74: use more user friendly name, such as `ipex_fp32`

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -14,48 +14,76 @@
...
+        "None_int8": AccelerationOption(inc=True),
+        "jit_fp32": AccelerationOption(jit=True),
+        "jit_fp32_ipex": AccelerationOption(jit=True, ipex=True),
+        "jit_fp32_ipex_clast": AccelerationOption(jit=True, ipex=True,
+                                                  channels_last=True),
+        "openvino_fp32": AccelerationOption(openvino=True),
+        "openvino_int8": AccelerationOption(openvino=True, inc=True),
+        "onnxruntime_fp32": AccelerationOption(onnxtunrime=True),
+        "onnxruntime_int8_qlinear": AccelerationOption(onnxruntime=True, inc=True),
+        "onnxruntime_int8_integer": AccelerationOption(onnxruntime=True, inc=True),
```
jason-dai, line 86: what's the difference between the above two?

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -94,103 +123,234 @@ def optimize(self, model,
...
+
+            # After the above conditions are met, the latency comparison is performed
+            if result["latency"] < best_metric.latency:
+                best_model = result["model"]
+                best_metric = CompareMetric(method, result["latency"], result["accuracy"])
+
+        return best_model, _format_acceleration_info(best_metric.method_name)
+
+    @staticmethod
+    def trace(model: nn.Module,
```
jason-dai, line 304: move the implementation of `Trainer.trace` and `Trainer.quantize` to `Optimizer`, and have `Trainer` to call `Optimizer`

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -231,6 +391,13 @@ def _bf16_checker():
     return "avx512_bf16" in msg or "amx_bf16" in msg
 
 
+def _detect_quantization_method(method_name):
+    method_name = method_name.split("_")[-1]
```
jason-dai, line 395: make this an attribute of option instead of parsing the name

jason-dai:
1) change to `nano.pytorch.InferenceOptimizer`?

3) Add type hint to new APIs in `nano.pytorch.InferenceOptimizer`

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
TheaperDeng: These are 2 common method we use for onnxruntime quantization. It is hard to decide which one is better since it varies from case to case. the method name (i.e. qlinear, integer) will be parsed in later codes, this dictionary only states the dependencies they need ( they have same dependencies requirement)

jason-dai: As mentioned in the comments below, this can be confusing and hard to maintain in the long run (by parsing the string name); add another attribute (such as `method`) instead

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
rnwang04: done

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
rnwang04: done

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: done

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: done

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: add pot here.

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
TheaperDeng: We may split this one to next PR to make the PR smaller and easier to control. The `quantize` and `trace` API will be migrate to InfereceOptimizer in 4 steps:

0. This PR: InferenceOptimizer call Trainer.
1. PR + 1: Trainer call InferenceOptimizer.
2. PR + 2: Revise our example, use-case.
3. PR + 3: deprecate Trainer's quantize and trace API.

rnwang04:
> 1. change to `nano.pytorch.InferenceOptimizer`?
> 2. Add type hint to new APIs in `nano.pytorch.InferenceOptimizer`

Has changed to InferenceOptimizer, and add type hints.

python/nano/test/pytorch/tests/test_inference_pipeline.py
```diff
@@ -24,7 +24,7 @@
 from torch.utils.data import DataLoader
 from torch.utils.data.sampler import SequentialSampler
 from bigdl.nano.pytorch import Trainer
-from bigdl.nano.pytorch.inference import Optimizer
+from bigdl.nano.pytorch.inference import InferenceOptimizer
```
jason-dai, line 27: change to from bigdl.nano.pytorch import InferenceOptimizer

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -73,8 +101,10 @@ def optimize(self, model,
                  training_data,
```
jason-dai, line 101: add type hint for this API

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -14,52 +14,80 @@
...
+        "onnxruntime_fp32": AccelerationOption(onnxtunrime=True),
+        "onnxruntime_int8_qlinear": AccelerationOption(onnxruntime=True, inc=True,
+                                                       method="qlinear"),
+        "onnxruntime_int8_integer": AccelerationOption(onnxruntime=True, inc=True,
+                                                       method="integer"),
     }
 
 
-class Optimizer:
+class InferenceOptimizer:
```
jason-dai, line 90: add type hint for APIs in this class

python/nano/test/pytorch/tests/test_inference_pipeline.py
rnwang04: done

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: done

jason-dai:
LGTM

PR #5491, opened 2022-08-22
[Nano] Openvino quantization notebooks with nano

zhentaocc:
## Description

modify openvino quantization notebook with nano.

### 1. Why the change?

modify openvino quantization notebook with nano.

### 2. User API changes

N/A

### 3. Summary of the change 

modify openvino quantization notebook with nano.

### 4. How to test?

- [x] Unit test


zhentaocc:
To check modifications, please check following two files:
### before
https://github.com/intel-analytics/BigDL/blob/main/python/nano/notebooks/pytorch/openvino/quantization-simplified-mode/quantization-simplified-mode.ipynb
### after
https://github.com/zhentaocc/BigDL/blob/openvino_quantization_notebooks_with_nano/python/nano/notebooks/pytorch/openvino/quantization-simplified-mode/quantization-simplified-mode.ipynb

jason-dai:
@yangw1234 shall we make this a notebook, or a python example?

yangw1234:
Since the original one is a notebook and we already finish a notebook. How about keeping this notebook and also add a simple python example in https://github.com/intel-analytics/BigDL/tree/main/python/nano/tutorial/inference/pytorch

zhentaocc:
To do:

- Add Async inference API
- modify this notebook with Async inference API
- Make this as an example

@hjzin

PR #5497, opened 2022-08-22
Fix bug that string_embed function Only replace one columns when replace=True.

ForJadeForest:
## Description

Fix the bug that string_embed function Only replace one columns when `replace=True`. 
Add the for loop to replace all columns.


hkvision (approved):

hkvision:
http://10.112.231.51:18888/job/BigDL-PRVN-Python-Friesian-UnitTest-Spark-3.1/565/consoleFull

PR #5499, opened 2022-08-22
Orca: add shutdown API to tf2 pyspark estimator.

lalalapotter:
## Description
This PR is to add shutdown API to tf2 PySpark estimator aligning with the other estimators.


python/orca/src/bigdl/orca/learn/tf2/pyspark_estimator.py
```diff
@@ -528,3 +528,32 @@ def get_model(self):
     @property
     def _model_saved_path(self):
         return os.path.join(self.model_dir, "{}_model.h5".format(self.application_id))
+
+    def shutdown(self):
+        """
+        Shutdown estimator and release resources.
+        """
+        if self.log_server_thread.is_alive():
+            import inspect
```
sgwhat, line 537: How about moving this part of logic into https://github.com/intel-analytics/BigDL/blob/main/python/orca/src/bigdl/orca/learn/log_monitor.py as a function named like `stop_log_server`? 🤣

lalalapotter: Sure.

sgwhat (approved):
LGTM

PR #5505, opened 2022-08-23
Orca: update known-issues.

lalalapotter:
## Description
This PR is to update Orca Known-issues which is related to PyArrow package.

### DIFF
--- a/docs/readthedocs/source/doc/Orca/Overview/known_issues.md
+++ b/docs/readthedocs/source/doc/Orca/Overview/known_issues.md
@@ -67,7 +67,7 @@ You could follow below steps to workaround:
 
 ### **OSError: Unable to load libhdfs: ./libhdfs.so: cannot open shared object file: No such file or directory**
 
-This error is because that PyArrow fails to locate `libhdfs.so` in default path of `$HADOOP_HOME/lib/native`.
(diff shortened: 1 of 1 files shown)


docs/readthedocs/source/doc/Orca/Overview/known_issues.md
```diff
@@ -84,3 +62,26 @@ You could follow below steps to workaround:
       ```
 
 2. If you really need to use ray on spark, please install bigdl-orca under a conda environment. Detailed information please refer to [here](./orca.html).
+
+## **Other Issues**
+
+### **OSError: Unable to load libhdfs: ./libhdfs.so: cannot open shared object file: No such file or directory**
+
+This error is because that PyArrow fails to locate `libhdfs.so` in default path of `$HADOOP_HOME/lib/native`.
```
hkvision, line 70: remove `that`
Add `when you run with YARN on Cloudera`

hkvision (approved):

docs/readthedocs/source/doc/Orca/Overview/known_issues.md
lalalapotter: fixed.

PR #5508, opened 2022-08-23
Orca: add get remote file to local decorator in orca file utils.

lalalapotter:
## Description
This PR is to add python implemented decorator for multiple FS loading without PyArrow.

### DIFF
--- a/python/orca/src/bigdl/orca/data/file.py
+++ b/python/orca/src/bigdl/orca/data/file.py
@@ -544,8 +544,8 @@ def multi_fs_load(load_func):
 
     Enable loading file or directory in multiple file systems.
     It supports local, hdfs, s3 file systems.
-    Note: this decorator is different from dllib decorator @enable_multi_fs_load. 
-          This decorator can load on each worker while @enable_multi_fs_load can only load on driver.
+    Note: this decorator is different from dllib decorator @enable_multi_fs_load.
+    This decorator can load on each worker while @enable_multi_fs_load can only load on driver.
 
     :param load_func: load file or directory function
     :return: load file or directory funtion for specific file system
--- a/python/orca/src/bigdl/orca/data/file.py
+++ b/python/orca/src/bigdl/orca/data/file.py
@@ -18,6 +18,7 @@ import os
 import subprocess
 import logging
 import shutil
+import functools
 import glob
 from distutils.dir_util import copy_tree
 from bigdl.dllib.utils.log4Error import *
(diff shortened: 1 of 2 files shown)


jason-dai:
Do you have a unit test?

lalalapotter:
Sure, I will add an unit test.

python/orca/src/bigdl/orca/data/file.py
```diff
@@ -537,3 +537,31 @@ def get_remote_files_with_prefix_to_local(remote_path_prefix, local_dir):
         except Exception as e:
             invalidOperationError(False, str(e), cause=e)
     return os.path.join(local_dir, prefix)
+
+
+def enable_multi_fs_load(load_func):
```
hkvision, line 542: rename to multi_fs_load

python/orca/src/bigdl/orca/data/file.py
```diff
@@ -537,3 +537,31 @@ def get_remote_files_with_prefix_to_local(remote_path_prefix, local_dir):
         except Exception as e:
             invalidOperationError(False, str(e), cause=e)
     return os.path.join(local_dir, prefix)
+
+
```
hkvision, line 541: Add some comment to tell the difference between the one in dllib and the one here.
For example, this decorator can load on each worker while that can only load on driver

python/orca/src/bigdl/orca/data/file.py
lalalapotter: fixed.

python/orca/src/bigdl/orca/data/file.py
lalalapotter: Added.

python/orca/src/bigdl/orca/data/file.py
```diff
@@ -537,3 +537,43 @@ def get_remote_files_with_prefix_to_local(remote_path_prefix, local_dir):
...
+def multi_fs_load(load_func):
+    """
+
+    Enable loading file or directory in multiple file systems.
+    It supports local, hdfs, s3 file systems.
+    Note: this decorator is different from dllib decorator @enable_multi_fs_load.
+    This decorator can load on each worker while @enable_multi_fs_load can only load on driver.
+
+    :param load_func: load file or directory function
+    :return: load file or directory funtion for specific file system
```
hkvision, line 551: typo `function`
for `the`

lalalapotter: fixed.

hkvision (approved):

PR #5509, opened 2022-08-23
Correct the path of start-fl-server.py in assembly.xml

Hanyu-Jin:
## Description

  The old path in assembly.xml is wrong.


glorysdj (approved):
LGTM

PR #5513, opened 2022-08-23
[PPML] Use tracker.py to build xgboost4j_2.12-1.1.2.jar

gc-fu:
## Description
Instead of using tracker.py to build `${BIGDL_HOME}/jars/bigdl-dllib-spark_${SPARK_VERSION}-${BIGDL_VERSION}.jar`, we now use it to build xgboost4j_2.12-1.1.2.jar.
### 1. Why the change?
The original tracker.py in `${BIGDL_HOME}/jars/bigdl-dllib-spark_${SPARK_VERSION}-${BIGDL_VERSION}.jar` cannot be reached when running xgboost code.

### 2. User API changes
None.

### 3. Summary of the change 
Change to use tracker.py to update the `xgboost4j_2.12-1.1.2.jar`

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [ ] Document test
- [x] Github Action


glorysdj (approved):
LGTM

PR #5518, opened 2022-08-24
Fix the problem of cannot find the directory.

Hanyu-Jin:
## Description

  The directory isn't created resulting in the error.


glorysdj (approved):
LGTM

PR #5528, opened 2022-08-25
[PPML] Fix tracker used for xgboost examples

gc-fu:
## Description
The original code which is shown below
```python
host_ip = os.environ['RABIT_TRACKER_IP']
```
will lead to keyError if RABIT_TRACKER_IP is not set in the environment.

The ideal result should be return None.

### 1. Why the change?
To fix the error mentioned in description.

### 2. User API changes
None

### 3. Summary of the change 
Fix a potential error that would lead to unexpected keyError.

### 4. How to test?
- [x] N/A
- [ ] Unit test
- [ ] Application test
- [ ] Document test


qiyuangong (approved):
LGTM

Issue #4886, opened 2022-06-19
Chronos: chronos benchmark tool


TheaperDeng:
## motivation
This is motivated by https://docs.openvino.ai/latest/openvino_inference_engine_tools_benchmark_tool_README.html, we need a benchmark tool for developers and users to benchmark the training and inferencing performance on their own platform.

Our users need an easy command to know what their platform can do. Some of our external developers need to user this benchmark tools to get the performance benchmark result. Currently it's only designed for performance rather than accuracy and we can keep iterating on the tool.

## usage
```bash
./benchmark-chronos -model/m <model_name>\ # model name
                    -stage/s <train/latency/throuput>\ # train, inference latency, inference throughput
                    -dataset/d <nyc_taxi/electricy>\ # dataset name, currently for built-in model
                    -core/c 4\ # core number
                    -cpuset/u 0,1,2,3\ # bind to some cpu id(optional), default to no pinning
                    -lookback/l <lookback>\ # lookback(optional), default to lookback lookback
                    -horizon/h <horizon> # horizon(optional), default to dataset horizon
```

PR #5533, opened 2022-08-25
Chronos: test-run is added to the chronos benchmark tool

plusbang:
## Description

We need the benchmark tool for developers and users to benchmark the training and inferencing performance on their own platform. Therefore, besides the hardware and environment diagnosis, test-run part is added.
 Users can choose one model type and find out the predict latency of the corresponding forecaster.

### 1. Why the change?

https://github.com/intel-analytics/BigDL/issues/4886

### 2. User API changes

```python
python benchmark-chronos.py --model/-m <model_name> --stage/-s latency
```
The model_name can be one of ["LSTM", "Seq2Seq", "TCN", "Autoformer"]. And an example can be:
```python
python benchmark-chronos.py -m TCN -s latency
```

### 3. Summary of the change 

Users can use the tool to test out the predict latency of choosen forecaster on their own platform.

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/chronos/script/benchmark-chronos.py
+++ b/python/chronos/script/benchmark-chronos.py
@@ -14,154 +14,47 @@
 # limitations under the License.
 #
 
-
-import psutil
-import platform
-import subprocess
-import os
 import sys
-import logging
-import warnings
-from typing import Union, Dict, List, Optional
+import getopt
 import numpy as np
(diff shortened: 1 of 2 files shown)


TheaperDeng:
close this since we will have a new PR for this

PR #5539, opened 2022-08-25
cicd: fix minor issues in github actions workflows

liu-shaojun:
## Description

This submission
- fix format error of yaml file
- fix typo
- add `shell: bash` in composite actions. [runs.steps[*].shell](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runsstepsshell) is required if run is used in steps.


glorysdj (approved):
LGTM

PR #5540, opened 2022-08-25
[Nano] add __call__ method for OpenVINOModel

hjzin:
### 1. Why the change?

In OpenVINOModel, we can only call `OpenVINOModel.forward_step()` to inference:
```python
openvino_model = OpenVINOModel(model_path)
result = openvino_model.forward_step(x)
```
To simplify the usage and make it consistent with pytorch, we can add `__call__` method for OpenVINOModel, then users can directly call the OpenVINOModel object to inference.

### 2. User API changes

**Before**
```python
openvino_model = OpenVINOModel(model_path)
result = openvino_model.forward_step(x)
```
**After**
```python
openvino_model = OpenVINOModel(model_path)
result = openvino_model(x)
```
### 3. Summary of the change 

1. Add `__call__` method for OpenVINOModel class.
2. Now all base classes of PytorchOpenVINOModel have `__call__` method, so  `__call__` method is added for PytorchOpenVINOModel to specify which method it should call.

### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/deps/openvino/core/model.py
+++ b/python/nano/src/bigdl/nano/deps/openvino/core/model.py
@@ -53,18 +53,17 @@ class OpenVINOModel:
         input_names = [t.any_name for t in self._ie_network.inputs]
         self._forward_args = input_names
 
-    def _save_model(self, path):
+    def _save(self, path):
         """
         Save OpenVINOModel to local as xml and bin file
 
         :param path: Directory to save the model.
         """
-        path = Path(path)
-        path.mkdir(exist_ok=True)
-        invalidInputError(self.ie_network,
-                          "self.ie_network shouldn't be None.")
-        xml_path = path / self.status['xml_path']
-        save(self.ie_network, xml_path)
+        if self._model_exists_or_err("self.ie_network shouldn't be None."):
+            path = Path(path)
+            path.mkdir(exist_ok=True)
+            xml_path = path / 'ov_saved_model.xml'
+            save(self.ie_network, xml_path)
 
     def pot(self,
             dataloader,
@@ -151,3 +150,8 @@ class OpenVINOModel:
             model = Core().read_model(model_path)
             model.reshape(orig_shape)
         return model
+
+    def _model_exists_or_err(self, err_msg):
+        if self.ie_network is None:
+            invalidInputError(False, err_msg)
+        return True
--- a/python/nano/src/bigdl/nano/deps/openvino/core/model.py
+++ b/python/nano/src/bigdl/nano/deps/openvino/core/model.py
@@ -53,18 +53,17 @@ class OpenVINOModel:
         input_names = [t.any_name for t in self._ie_network.inputs]
         self._forward_args = input_names
 
-    def _save_model(self, path, model):
+    def _save_model(self, path):
         """
         Save OpenVINOModel to local as xml and bin file
 
         :param path: Directory to save the model.
-        :param model: Model object to get status property
         """
         path = Path(path)
         path.mkdir(exist_ok=True)
         invalidInputError(self.ie_network,
                           "self.ie_network shouldn't be None.")
-        xml_path = path / model.status['xml_path']
+        xml_path = path / self.status['xml_path']
         save(self.ie_network, xml_path)
 
     def pot(self,
--- a/python/nano/src/bigdl/nano/deps/openvino/core/model.py
+++ b/python/nano/src/bigdl/nano/deps/openvino/core/model.py
@@ -53,17 +53,18 @@ class OpenVINOModel:
         input_names = [t.any_name for t in self._ie_network.inputs]
         self._forward_args = input_names
 
-    def _save_model(self, path):
+    def _save_model(self, path, model):
         """
         Save PytorchOpenVINOModel to local as xml and bin file
 
         :param path: Directory to save the model.
+        :param model: Model object to get status property
         """
         path = Path(path)
         path.mkdir(exist_ok=True)
         invalidInputError(self.ie_network,
                           "self.ie_network shouldn't be None.")
-        xml_path = path / self.status['xml_path']
+        xml_path = path / model.status['xml_path']
         save(self.ie_network, xml_path)
 
     def pot(self,
--- a/python/nano/src/bigdl/nano/deps/openvino/core/model.py
+++ b/python/nano/src/bigdl/nano/deps/openvino/core/model.py
@@ -30,8 +30,8 @@ class OpenVINOModel:
     def forward_step(self, *inputs):
         return self._infer_request.infer(list(inputs))
 
-    def __call__(self, *inputs, **kwargs):
-        return self.forward_step(*inputs, **kwargs)
+    def __call__(self, *inputs):
+        return self.forward_step(*inputs)
 
     @property
     def forward_args(self):
--- a/python/nano/src/bigdl/nano/deps/openvino/core/model.py
+++ b/python/nano/src/bigdl/nano/deps/openvino/core/model.py
@@ -30,6 +30,9 @@ class OpenVINOModel:
     def forward_step(self, *inputs):
         return self._infer_request.infer(list(inputs))
 
+    def __call__(self, *inputs, **kwargs):
+        return self.forward_step(*inputs, **kwargs)
+
     @property
     def forward_args(self):
         return self._forward_args
--- a/python/nano/src/bigdl/nano/deps/openvino/core/model.py
+++ b/python/nano/src/bigdl/nano/deps/openvino/core/model.py
@@ -55,7 +55,7 @@ class OpenVINOModel:
 
     def _save_model(self, path, model):
         """
-        Save PytorchOpenVINOModel to local as xml and bin file
+        Save OpenVINOModel to local as xml and bin file
 
         :param path: Directory to save the model.
         :param model: Model object to get status property
--- a/python/nano/src/bigdl/nano/deps/openvino/core/model.py
+++ b/python/nano/src/bigdl/nano/deps/openvino/core/model.py
@@ -59,11 +59,11 @@ class OpenVINOModel:
 
         :param path: Directory to save the model.
         """
-        if self._model_exists_or_err("self.ie_network shouldn't be None."):
-            path = Path(path)
-            path.mkdir(exist_ok=True)
-            xml_path = path / 'ov_saved_model.xml'
-            save(self.ie_network, xml_path)
+        self._model_exists_or_err()
+        path = Path(path)
+        path.mkdir(exist_ok=True)
+        xml_path = path / 'ov_saved_model.xml'
+        save(self.ie_network, xml_path)
 
     def pot(self,
             dataloader,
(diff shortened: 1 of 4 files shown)


python/nano/src/bigdl/nano/deps/openvino/pytorch/model.py
```diff
@@ -45,6 +45,9 @@ def __init__(self, model, input_sample=None, **export_kwargs):
             OpenVINOModel.__init__(self, ov_model_path)
             AcceleratedLightningModule.__init__(self, None)
 
+    def __call__(self, *inputs, **kwargs):
+        return AcceleratedLightningModule.__call__(self, *inputs, **kwargs)
```
jason-dai, line 49: does this call go to `LightningModule`?

hjzin: `AcceleratedLightningModule.__call__` will call `__call__` method of `LightningModule`

jason-dai: Does it make sense? Should it call `AcceleratedLightningModule.forward`

hjzin: `LightningModule.__call__` will call `AcceleratedLightningModule.forward` because `AcceleratedLightningModule` inherits from `LightningModule` and overrides the `forward` function.

zhentaocc, line 49:
```suggestion
    def forward(self, *inputs):
        return AcceleratedLightningModule.forward(self, *inputs)
```
How about this?

zhentaocc:
Need to modify KerasOpenVINOModel.

hjzin:
> 

Yes, I'm working on it.

zhentaocc:
Class definition like this may cause conflict of `__call__` function since both `OpenVINOModel `and `LightningModule` have `__call__`:
```python
class PytorchOpenVINOModel(OpenVINOModel, AcceleratedLightningModule):
    def __init__(self, model, input_sample=None, **export_kwargs):
        ov_model_path = model
        with TemporaryDirectory() as dir:
            dir = Path(dir)
            if isinstance(model, torch.nn.Module):
                export(model, input_sample, str(dir / 'tmp.xml'), **export_kwargs)
                ov_model_path = dir / 'tmp.xml'
            OpenVINOModel.__init__(self, ov_model_path)
            AcceleratedLightningModule.__init__(self, None)
```
can we change the class:
```python
class PytorchOpenVINOModel(AcceleratedLightningModule):
    def __init__(self, model, input_sample=None, **export_kwargs):
        ov_model_path = model
        with TemporaryDirectory() as dir:
            dir = Path(dir)
            if isinstance(model, torch.nn.Module):
                export(model, input_sample, str(dir / 'tmp.xml'), **export_kwargs)
                ov_model_path = dir / 'tmp.xml'
            ov_model = OpenVINOModel(ov_model_path)
            super().__init__(ov_model)
```

hjzin:
> Class definition like this may cause conflict of `__call__` function since both `OpenVINOModel `and `LightningModule` have `__call__`:
> 
> ```python
> class PytorchOpenVINOModel(OpenVINOModel, AcceleratedLightningModule):
>     def __init__(self, model, input_sample=None, **export_kwargs):
>         ov_model_path = model
>         with TemporaryDirectory() as dir:
>             dir = Path(dir)
>             if isinstance(model, torch.nn.Module):
>                 export(model, input_sample, str(dir / 'tmp.xml'), **export_kwargs)
>                 ov_model_path = dir / 'tmp.xml'
>             OpenVINOModel.__init__(self, ov_model_path)
>             AcceleratedLightningModule.__init__(self, None)
> ```
> 
> can we change the class:
> 
> ```python
> class PytorchOpenVINOModel(AcceleratedLightningModule):
>     def __init__(self, model, input_sample=None, **export_kwargs):
>         ov_model_path = model
>         with TemporaryDirectory() as dir:
>             dir = Path(dir)
>             if isinstance(model, torch.nn.Module):
>                 export(model, input_sample, str(dir / 'tmp.xml'), **export_kwargs)
>                 ov_model_path = dir / 'tmp.xml'
>             ov_model = OpenVINOModel(ov_model_path)
>             super().__init__(ov_model)
> ```

Ok, I'll try it. And `KerasOpenVINOModel` may also be changed.

zhentaocc:
Shall we make it consistent between TF and Pytorch?

python/nano/src/bigdl/nano/deps/openvino/tf/model.py
```diff
@@ -79,3 +83,6 @@ def _load(path):
             invalidInputError(False, "nano_model_meta.yml must specify 'xml_path' for loading.")
         xml_path = Path(path) / status['xml_path']
         return KerasOpenVINOModel(xml_path)
+
+    def _save_model(self, path):
+        self.ov_model._save_model(path, model=self)
```
zhentaocc, line 88: what is this change for?

python/nano/src/bigdl/nano/deps/openvino/tf/model.py
```diff
@@ -39,11 +40,14 @@ def __init__(self, model):
...
-            OpenVINOModel.__init__(self, ov_model_path)
+            self.ov_model = OpenVINOModel(ov_model_path)
             AcceleratedKerasModel.__init__(self, None)
 
+    def forward_step(self, *inputs):
+        return self.ov_model.forward_step(*inputs)
+
     def on_forward_start(self, inputs):
-        if self.ie_network is None:
+        if self.ov_model.ie_network is None:
```
zhentaocc, line 50: can you move this check to `OpenVINOModel`? I think this check should be the same for Pytorch/TF/OpenVINO.

python/nano/src/bigdl/nano/deps/openvino/tf/model.py
```diff
@@ -39,11 +40,14 @@ def __init__(self, model):
             if isinstance(model, tf.keras.Model):
                 export(model, str(dir / 'tmp.xml'))
                 ov_model_path = dir / 'tmp.xml'
-            OpenVINOModel.__init__(self, ov_model_path)
+            self.ov_model = OpenVINOModel(ov_model_path)
             AcceleratedKerasModel.__init__(self, None)
 
+    def forward_step(self, *inputs):
+        return self.ov_model.forward_step(*inputs)
+
```
zhentaocc, line 48:
```suggestion
            model = OpenVINOModel(ov_model_path)
            super().__init__(model)

```

zhentaocc: I think using `super` is better.

zhentaocc: And you don't have to override `forward_step`.

python/nano/src/bigdl/nano/deps/openvino/tf/model.py
```diff
@@ -14,6 +14,7 @@
 # limitations under the License.
 #
 from pathlib import Path
+from statistics import mode
```
zhentaocc, line 17: why?

zhentaocc (changes requested):

python/nano/src/bigdl/nano/deps/openvino/tf/model.py
hjzin: `_save_model` in OpenVINOModel needs model's status property, but it can't be retrieved when OpenVINOModel is not the base class of PytorchOpenVINOModel and KerasOpenVINOModel, so I add a `model` parameter to `_save_model` in OpenVINOModel and pass the model object to `model` parameter.

Besides this change, we can also simply copy the implementation of `_save_model` in OpenVINOModel to PytorchOpenVINOModel and KerasOpenVINOModel, but I think it will be a little redundant.

python/nano/src/bigdl/nano/deps/openvino/tf/model.py
hjzin: import by mistake, I will remove it

python/nano/src/bigdl/nano/deps/openvino/tf/model.py
hjzin: Yes, `forward_step` in PyTorchOpenVINOModel can be stripped, but it seems that KerasOpenVINOModel need it because `AcceleratedKerasModel` don't have `forward_step` method.

hjzin: Or should we add `forward_step` in `AcceleratedKerasModel`?

zhentaocc: https://github.com/intel-analytics/BigDL/blob/main/python/nano/src/bigdl/nano/utils/inference/tf/model.py#L39-L42 I think it has?

python/nano/src/bigdl/nano/deps/openvino/tf/model.py
hjzin: should the error message be different? or just keep the same.

python/nano/src/bigdl/nano/deps/openvino/tf/model.py
zhentaocc: This is an issue we need to resolve. Is there any way to fix this without modifying the signature? We should have the same usage:
```python
OpenVINOModel.save()
PytorchOpenVINOModel.save()
KerasOpenVINOModel.save()
```
If you must retrieve status, just pass `self.status`.

python/nano/src/bigdl/nano/deps/openvino/tf/model.py
hjzin: seems it will call `forward_step` in `AcceleratedModel`, which directly raise an error.
https://github.com/intel-analytics/BigDL/blob/main/python/nano/src/bigdl/nano/utils/inference/model.py#L31-L32

python/nano/src/bigdl/nano/deps/openvino/tf/model.py
hjzin: ok, I will change it.

python/nano/src/bigdl/nano/deps/openvino/core/model.py
```diff
@@ -148,3 +150,8 @@ def pot(self,
             model = Core().read_model(model_path)
             model.reshape(orig_shape)
         return model
+
+    def _model_exists_or_err(self, err_msg):
+        if self.ie_network is None:
+            invalidInputError(False, err_msg)
```
zhentaocc, line 156:
```suggestion
        invalidInputError(self.ie_network is not None, err_msg)
```

python/nano/src/bigdl/nano/deps/openvino/core/model.py
```diff
@@ -50,18 +53,17 @@ def ie_network(self, model):
...
 
         :param path: Directory to save the model.
         """
-        path = Path(path)
-        path.mkdir(exist_ok=True)
-        invalidInputError(self.ie_network,
-                          "self.ie_network shouldn't be None.")
-        xml_path = path / self.status['xml_path']
-        save(self.ie_network, xml_path)
+        if self._model_exists_or_err("self.ie_network shouldn't be None."):
```
zhentaocc, line 62: remove if.

zhentaocc (commented):
I think implementation of `_save_model` is a bit complicated. Maybe we can try to have a better one after this PR.

hjzin:
@TheaperDeng can you help review this pr?

zhentaocc:
LGTM, require Jenkins tests.

hjzin:
http://10.112.231.51:18889/view/BigDL-PR-Validation/job/BigDL-Chronos-PR-Validation/810/

PR #5544, opened 2022-08-26
[PPML] refine eHSM-KMS according to user experience

Uxito-Ada:
## Description

Refine the existing bigdl-pccs-ehsm-kms according to user experience.


glorysdj (approved):
LGTM

PR #5545, opened 2022-08-26
PPML FL Pytorch NN tutorial refine

Litchilitchy:
## Description

Some small logic changes and document refinements
* server would only take the first model uploaded or called server_load. (when model exists, it would not do anything to change it)
* add a few more details in 2.7 section about resuming training and predicting


python/ppml/src/bigdl/ppml/fl/nn/pytorch/aggregator.py
```diff
@@ -127,6 +127,13 @@ def sort_by_key(kv_tuple):
             invalidInputError(False,
                               f'Invalid phase: {phase}, should be train/eval/pred')
 
+    def load_uploaded_model(self, client_id, model_path):
+        if self.model is not None:
+            logging.warn(f"Model exists, model uploading from {client_id} ignored.")
```
jason-dai, line 132: Only one client can specify server model and server model path; you need to check that, and report error (instead of warning)

python/ppml/src/bigdl/ppml/fl/nn/tensorflow/aggregator.py
```diff
@@ -119,3 +120,9 @@ def sort_by_key(kv_tuple):
             invalidInputError(False,
                               f'Invalid phase: {phase}, should be train/eval/pred')
 
+    def load_uploaded_model(self, client_id, model_path):
+        if self.model is not None:
+            logging.warn(f"Model exists, model uploading from {client_id} ignored.")
```
jason-dai, line 125: Only one client can specify server model and server model path; you need to check that, and report error (instead of warning)

python/ppml/src/bigdl/ppml/fl/nn/pytorch/aggregator.py
Litchilitchy: Should I use log ERROR in client or just raise ERROR and terminate the client program?

jason-dai: Report error and exist on client side.

PR #5546, opened 2022-08-26
cicd: add a new workflow called PR_Validation

liu-shaojun:
## Description

This submission
- add a new workflow called PR_Validation. 
- three jobs are added into this workflow: path filter job, friessian ut job and dllib ut job, friessian and dllib ut are dependent on path filter job, if the friessian or dllib related files are changed then friessian or dllib ut will be triggered otherwise it will be skipped.
- other test job can be added in the same way as friessian or dllib ut.

image

### DIFF
--- a/.github/workflows/PR_validation.yml
+++ b/.github/workflows/PR_validation.yml
@@ -28,7 +28,7 @@ jobs:
         with:
           filters: |
             src:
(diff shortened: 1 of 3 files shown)


glorysdj (approved):
LGTM

PR #5552, opened 2022-08-26
upgrade bigdl-tf/math with fixed vulnerabilities

Le-Zheng:
## Description

Upgrade bigdl-math and bigdl-tf to new version because some fix in zoo-core-tfnet are merged for solving high Vulnerabilities.


yangw1234:
LGTM

PR #5561, opened 2022-08-29
Fix GitHub action precision

y199387:
## Description

Remove dtype check in precision unit test
https://github.com/intel-analytics/BigDL/issues/5565

### How to test?
- [x] Unit test


y199387:
Close as sloved

PR #5567, opened 2022-08-29
Orca: OMP env default value reset.

lalalapotter:
## Description

This PR is to reset OpenMP default value for Orca.

According to the users' feedback, the setting of environment variables in dllib `init_nncontext()` will affect the efficiency of other applications. Therefore, we propose the following two solutions to this issue and implement solution A.

- Solution A
  Since users are more likely to initialize through `init_orca_context`, we will reset these environment variables in `init_orca_context` to avoid the impact of these environment variables on orca and other applications.
**Pros**: Easy to implement.
**Cons**: May have trouble using estimator with bigdl backend.
- Solution B
  Do not set the default value of environment variables in dllib, only set the corresponding default value when dllib uses these environment variables
**Pros**: Can solve the problem without concerns.
**Cons**: Need to go through dlllib code to add the corresponding settings.

### DIFF
--- a/python/dllib/src/bigdl/dllib/nncontext.py
+++ b/python/dllib/src/bigdl/dllib/nncontext.py
@@ -533,7 +533,7 @@ def init_nncontext(conf=None, cluster_mode="spark-submit", spark_log_level="WARN
             num_nodes = conf.get("spark.executor.instances")
         spark_args.update(conf.getAll())
     if cluster_mode == "spark-submit":
-        sc = init_internal_nncontext(conf, spark_log_level, redirect_spark_log)
+        sc = init_internal_nncontext(conf, spark_log_level, redirect_spark_log, **kwargs)
     elif cluster_mode == "local":
         if conf:
             os.environ["SPARK_DRIVER_MEMORY"] = conf.get("spark.driver.memory")
@@ -619,7 +619,7 @@ def init_nncontext(conf=None, cluster_mode="spark-submit", spark_log_level="WARN
     return sc
 
 
-def init_internal_nncontext(conf=None, spark_log_level="WARN", redirect_spark_log=True):
+def init_internal_nncontext(conf=None, spark_log_level="WARN", redirect_spark_log=True, **kwargs):
     """
     Creates or gets a SparkContext with optimized configurations for BigDL performance.
     This method will also initialize the BigDL engine.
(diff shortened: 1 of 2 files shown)


jason-dai:
I don't think this fix works, especially for distributed environment (where you need to set the proper environment variable before creating Spark context). Basically, you need to:
1) Set OMP THREADS to one when using BigDL backend
2) Set OMP THREADS to `cores` for other backends

lalalapotter:
Yes, I will update the implementation soon.

python/dllib/src/bigdl/dllib/nncontext.py
```diff
@@ -743,11 +742,12 @@ def init_env(conf):
...
 
 
-def init_spark_conf(conf=None):
+def init_spark_conf(conf=None, **kwargs):
     spark_conf = SparkConf()
     if conf:
         spark_conf.setAll(conf.items())
-    init_env(spark_conf)
+    if "is_orca" is not in kwargs or kwargs["is_orca"] is False:
+        init_env(spark_conf)
```
jason-dai, line 750: We need to properly set the ENV variables for Orca too

lalalapotter: We already have corresponding settings in the runner for TensorFlow and PyTorch.
- TensorFlow 1.x: [link](https://github.com/intel-analytics/BigDL/blob/fa0741d53cf7cd4179e2f852d63416229a6928a2/python/orca/src/bigdl/orca/learn/tf/tf_runner.py#L86)
- TensorFlow 2.x: [link](https://github.com/intel-analytics/BigDL/blob/fa0741d53cf7cd4179e2f852d63416229a6928a2/python/orca/src/bigdl/orca/learn/tf2/tf_runner.py#L256)
- PyTorch: [link](https://github.com/intel-analytics/BigDL/blob/main/python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py#L166)

jason-dai: > We already have corresponding settings in the runner for TensorFlow and PyTorch.
> 
> * TensorFlow 1.x: [link](https://github.com/intel-analytics/BigDL/blob/fa0741d53cf7cd4179e2f852d63416229a6928a2/python/orca/src/bigdl/orca/learn/tf/tf_runner.py#L86)
> * TensorFlow 2.x: [link](https://github.com/intel-analytics/BigDL/blob/fa0741d53cf7cd4179e2f852d63416229a6928a2/python/orca/src/bigdl/orca/learn/tf2/tf_runner.py#L256)
> * PyTorch: [link](https://github.com/intel-analytics/BigDL/blob/main/python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py#L166)
> 
> Do we need to explore the possibility of unifying these settings？

1) These only set the config inside respective library, not the global environment variables
2) Environment variables should be set before starting the JVM or Python process, not inside hte process
3) And yes, we need to refactor all the settings of configs and environment variables to make it more robust

hkvision: Why do we want to set the global environment variables? Is it more reasonable to set in each library so that if we run two workloads in one context, they won't impact each other?

PR #5572, opened 2022-08-29
Fixprocessing

dding3:
## Description
Improve several processings in tutorials so we can use the original dataset download from website.

### 1. How to test?
Application test


dding3:
All checks passed.

PR #5558, opened 2022-08-27
support convert spark df to shards with arrow

dding3:
## Description
support convert spark df to shards with arrow optim. This pr includes https://github.com/intel-analytics/BigDL/pull/5385 as need to do e2e perf test.

### 1. Why the change?

It's used to boost performance of convert spark df to shards.
perf result details: 20 cores, 80g mem, csv file with 1.1g(5 million rows with 35 cols).
spark df to shards without_arrow: 7 s
spark df to shards with arrow: 0.7 s

### 2. User API changes

N/A

### 3. Summary of the change 

support convert spark df to shards with arrow. The workflow is:
a. jvm convert spark df to arrow stream and dump to file
b. python read the file and generate a rdd of pandas dataframe
c. python create shards with the rdd

### 4. How to test?
- [x] Unit test
- [x] Application test

PR #5573, opened 2022-08-29
optim get schema and get class name in shards into 1 rdd operation

dding3:
## Description

optimize  `_get_class_name` and `_get_schema` into one rdd.map operation.

### 1. Why the change?

In current implementation, `_get_class_name` and `_get_schema` are get with 2 rdd operations which will bring extra overhead, put them into one to save cost.

### 2. User API changes

N/A

### 3. How to test?
- [x] Unit test
- [x] Application test


python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -454,20 +454,29 @@ def _get_schema(self):
         if 'schema' in self.type:
             return self.type['schema']
         else:
-            if self._get_class_name() == 'pandas.core.frame.DataFrame':
-                import pandas as pd
-                columns, dtypes = self.rdd.map(lambda x: (x.columns, x.dtypes)).first()
-                self.type['schema'] = {'columns': columns, 'dtypes': dtypes}
-                return self.type['schema']
-            return None
+            class_name, schema = self._get_schema_class_name()
```
jason-dai, line 457: only call `_get_schema_class_name` if `class_name` does not exist, or is pandas dataframe

dding3:
most jekins passed. http://10.112.231.51:18888/view/ZOO-PR/job/BigDL-Orca-PR-Validation/1011/.  http://10.112.231.51:18888/job/Orca-PRVN-Others/1015/console

python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -598,21 +598,32 @@ def utility_func(x, func, *args, **kwargs):
...
-                self.type['schema'] = {'columns': columns, 'dtypes': dtypes}
-                return self.type['schema']
-            return None
+
+        if 'class_name' not in self.type\
+                or self.type['class_name'] == 'pandas.core.frame.DataFrame':
+            class_name, schema = self._get_schema_class_name()
+            self.type['class_name'] = class_name
+            self.type['schema'] = schema
+            return self.type['schema']
```
jason-dai, line 607: Need to return None otherwise

dding3: Fixed in https://github.com/intel-analytics/BigDL/pull/5558

PR #5574, opened 2022-08-29
lightGBM python example

songhappy:
1. Why the change?

to support python LightGBM on Friesian

2. Summary of the change

add  train_lightGBM.py
based on https://microsoft.github.io/SynapseML/

3. How to test?
run example 
python/friesian/example/lightGBM/train_lightGBM.py


songhappy:
refactor into PR-5696

PR #5575, opened 2022-08-29
fix dllib pom scope and typo in examples

songhappy:
### 1. Why the change?
1. examples => example, it is typo when migrating the code
2. currently, dependencies of slf4j and log4f only cover scope of test, it fails when running scala examples in IDE, so remove the scope of the test only.

### DIFF
--- a/scala/dllib/pom.xml
+++ b/scala/dllib/pom.xml
@@ -324,11 +324,13 @@
             <groupId>log4j</groupId>
             <artifactId>log4j</artifactId>
             <version>1.2.17</version>
+            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-log4j12</artifactId>
             <version>1.7.16</version>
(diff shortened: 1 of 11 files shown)


scala/dllib/pom.xml
```diff
@@ -324,13 +324,11 @@
             <groupId>log4j</groupId>
             <artifactId>log4j</artifactId>
             <version>1.2.17</version>
```
glorysdj, line 326: 1.2.17 log4j has high vulnerabilities, do NOT package it with compile scope.

scala/dllib/pom.xml
```diff
@@ -324,11 +324,13 @@
...
+            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-log4j12</artifactId>
             <version>1.7.16</version>
+            <scope>test</scope>
```
jason-dai, line 333: need to update the PR description

PR #5576, opened 2022-08-30
add action for gramine local spark example test

zzti-bsj:
## Description

add `action` for gramine local spark example test


liu-shaojun:
could you put **Example Tests PPML Spark Local on Gramine** into Nightly_Test.yml?

PR #5577, opened 2022-08-30
PPML FL FGBoost save/load server model

Litchilitchy:
## Description
This is the functionality change. Updates of the tutorial would be done after this is merged.
### 1. Why the change?

to refine save/load server model logic 

### 2. User API changes
* User could pass an argument `serverModelPath` when create `FGBoostRegression/FGBoostClassification` instance
* User could call `model.loadServerModel` to ask server to load the model on aggregator
Scala
```scala
//train and configure saving model
model = FGBoostRegression(serverModelPath = path)
model.fit(...)
//load and continue training
model = FGBoostRegression()
model.loadServerModel(path)
model.fit(...)
```
Python
```python
model = FGBoostRegression(server_model_path = path)
model.fit(...)
//load and continue training
model = FGBoostRegression()
model.load_server_model(path)
model.fit(...)
```
### 3. Summary of the change 
Refine FGBoost save/load server model to the one consistent with the NN. User could pass an argument `serverModelPath` when create `FGBoostRegression/FGBoostClassification` instance and server would save it after every boost round. This model is on disk and could be loaded later.

This is a refinement based on #5361 

### 4. How to test?
Use existed unittests

### DIFF
--- a/python/ppml/src/bigdl/ppml/fl/algorithms/fgboost_regression.py
+++ b/python/ppml/src/bigdl/ppml/fl/algorithms/fgboost_regression.py
@@ -25,9 +25,11 @@ import logging
 
 
 class FGBoostRegression(FLClientClosable):
-    def __init__(self, jvalue=None, learning_rate:float=0.1, max_depth=7, min_child_size=1):
+    def __init__(self, jvalue=None, learning_rate:float=0.1,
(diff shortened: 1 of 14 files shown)


jason-dai:
Show code example in `2. User API changes` of the description

PR #5580, opened 2022-08-30
Nano : ResNet Demo for InferenceOptmizer

rnwang04:
## Description

Use Finetuned ResNet18 as an example to illustrate how to apply InferenceOptimizer to quickly find acceleration method with the minimum inference latency.

### 1. Why the change?

To provide a simple example illustrating how to use InferenceOptimizer.

### 2. User API changes

no changes.

### 3. Summary of the change

Basic example for usage of InferenceOptimizer.

### DIFF
--- a/python/nano/example/pytorch/inference_pipeline/inference_pipeline.py
+++ b/python/nano/example/pytorch/inference_pipeline/inference_pipeline.py
@@ -1,0 +1,319 @@
+#
+# Copyright 2016 The BigDL Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This file is adapted from PyTorch Lightning.
+# https://github.com/Lightning-AI/lightning/blob/master/examples/
+# pl_domain_templates/computer_vision_fine_tuning.py
+# Copyright The PyTorch Lightning team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Computer vision example on Transfer Learning. This computer vision example illustrates how one could fine-tune a
+pre-trained network (by default, a ResNet50 is used) using pytorch-lightning. For the sake of this example, the
+'cats and dogs dataset' (~60MB, see `DATA_URL` below) and the proposed network (denoted by `TransferLearningModel`,
+see below) is trained for 15 epochs.
+
+The training consists of three stages.
+
+From epoch 0 to 4, the feature extractor (the pre-trained network) is frozen except
+maybe for the BatchNorm layers (depending on whether `train_bn = True`). The BatchNorm
+layers (if `train_bn = True`) and the parameters of the classifier are trained as a
+single parameters group with lr = 1e-2.
+
+From epoch 5 to 9, the last two layer groups of the pre-trained network are unfrozen
+and added to the optimizer as a new parameter group with lr = 1e-4 (while lr = 1e-3
+for the first parameter group in the optimizer).
+
+Eventually, from epoch 10, all the remaining layer groups of the pre-trained network
+are unfrozen and added to the optimizer as a third parameter group. From epoch 10,
+the parameters of the pre-trained network are trained with lr = 1e-5 while those of
+the classifier is trained with lr = 1e-4.
+
+Note:
+    See: https://pytorch.org/tutorials/beginner/transfer_learning_tutorial.html
+"""
+
+
+import logging
+from pathlib import Path
+from typing import Union
+
+import torch
+import torch.nn.functional as F
+from torch import nn, optim
+from torch.optim.lr_scheduler import MultiStepLR
+from torch.optim.optimizer import Optimizer
+from torch.utils.data import DataLoader
+from torchmetrics import Accuracy
+from torchvision import models, transforms
+from torchvision.datasets import ImageFolder
+from torchvision.datasets.utils import download_and_extract_archive
+
+from pytorch_lightning import LightningDataModule, LightningModule
+from pytorch_lightning.callbacks.finetuning import BaseFinetuning
+from pytorch_lightning.utilities.rank_zero import rank_zero_info
+
+from bigdl.nano.pytorch.trainer import Trainer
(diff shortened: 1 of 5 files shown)


python/nano/example/pytorch/inference_pipeline/inference_pipeline.py
```diff
@@ -0,0 +1,319 @@
...
+def accuracy(pred, target):
+    pred = torch.sigmoid(pred)
+    target = target.view((-1, 1)).type_as(pred).int()
+    return Accuracy()(pred, target)
+
+# accelaration inference using InferenceOptimizer
+model.eval()
+optimizer = InferenceOptimizer()
+optimizer.optimize(model=model,
+                   training_data=datamodule.predict_dataloader(),
```
TheaperDeng, line 305: we need to use training dataloader here to avoid data leak as we stated here https://github.com/intel-analytics/BigDL/blob/1b45bf03830cd5c47edaf23bd78705edd5143313/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py#L115-L121

python/nano/example/pytorch/inference_pipeline/inference_pipeline.py
```diff
@@ -0,0 +1,319 @@
...
+                   validation_data=datamodule.val_dataloader(),
+                   metric=accuracy,
+                   direction="max",
+                   cpu_num=1,
+                   latency_sample_num=30)
```
TheaperDeng, line 310: how long does it take?

python/nano/example/pytorch/inference_pipeline/inference_pipeline.py
```diff
@@ -0,0 +1,319 @@
...
+
+for key, value in optimizer.optimized_model_dict.items():
+    print("accleration option: {}, latency: {:.4f}ms, accuracy: {:.4f}".format(key, value["latency"], value["accuracy"]))
+
+acc_model, option = optimizer.get_best_model(accelerator="onnxruntime")
+print(option)
+
+acc_model, option = optimizer.get_best_model(accuracy_criterion=0.05)
+print(option)
```
TheaperDeng, line 319: show how to use `acc_model` to predict.

jason-dai:
We want to focus on the demo of inference pipeline optimization, but 90% of the code in this demo is not related to that. We need to make sure out focus area stands out from the code.

rnwang04:
> We want to focus on the demo of inference pipeline optimization, but 90% of the code in this demo is not related to that. We need to make sure out focus area stands out from the code.

How about moving these code(model definition, dataset and loader definition, finetuning) to another _help.py file and only keep last lines about inference in this pipeline.py file ? @jason-dai @TheaperDeng

TheaperDeng:
> > We want to focus on the demo of inference pipeline optimization, but 90% of the code in this demo is not related to that. We need to make sure out focus area stands out from the code.
> 
> How about moving these code(model definition, dataset and loader definition, finetuning) to another file(_help.py or _finetune.py) and only keep last lines about inference in this pipeline.py file ? @jason-dai @TheaperDeng

I think that would be fine.

python/nano/example/pytorch/inference_pipeline/resnet/README.md
```diff
@@ -0,0 +1,75 @@
+# Bigdl-nano InferenceOptimizer example on Cat vs. Dog dataset
+
+This example illustrates how to apply InferenceOptimizer to quickly find acceleration method with the minimum inference latency under specific restrictions or without restrictions for a trained model. 
+For the sake of this example, we first train the proposed network(by default, a ResNet50 is used) on the [cats and dogs dataset](https://storage.googleapis.com/mledu-datasets/cats_and_dogs_filtered.zip), which consists both [frozen and unfrozen stages](https://github.com/PyTorchLightning/pytorch-lightning/blob/495812878dfe2e31ec2143c071127990afbb082b/pl_examples/domain_templates/computer_vision_fine_tuning.py#L21-L35). Then, by calling `optimize()`, we can obtain all avaliable accelaration combinations provided by BigDL-Nano for inference. By calling `get_best_mdoel()` , we could get an accelerated model whose inference is 7.8x times faster.
```
TheaperDeng, line 4: avaliable -> available

python/nano/example/pytorch/inference_pipeline/resnet/README.md
```diff
@@ -0,0 +1,75 @@
...
+## Prepare the environment
+We recommend you to use [Anaconda](https://www.anaconda.com/distribution/#linux) to prepare the environment.
+```
+conda create -n nano python=3.7  # "nano" is conda environment name, you can use any name you like.
+conda activate nano
+pip install jsonargparse[signatures]
+pip install --pre --upgrade bigdl-nano[pytorch]
+
+# bf16 is avaliable only on torch1.12
+pip install torch==1.12.0 torchvision torchaudio --extra-index-url https://download.pytorch.org/whl/cpu
```
TheaperDeng, line 16: we don't need `torchaudio`?

python/nano/example/pytorch/inference_pipeline/resnet/README.md
```diff
@@ -0,0 +1,75 @@
...
+pip install --pre --upgrade bigdl-nano[pytorch]
+
+# bf16 is avaliable only on torch1.12
+pip install torch==1.12.0 torchvision torchaudio --extra-index-url https://download.pytorch.org/whl/cpu 
+# Necessary packages for inference accelaration
+pip install --upgrade intel-extension-for-pytorch
+pip install onnx onnxruntime onnxruntime-extensions
+pip install openvino-dev
+pip install --upgrade neural-compressor
+```
```
TheaperDeng, line 22: may be we should tell our users to ignore the warnings might pop out here? (about numpy version)

TheaperDeng (commented):
others LGTM

python/nano/example/pytorch/inference_pipeline/resnet/README.md
rnwang04: Yes, good catch. thanks : ) Done.

python/nano/example/pytorch/inference_pipeline/resnet/README.md
rnwang04: Done

python/nano/example/pytorch/inference_pipeline/resnet/README.md
rnwang04: done

TheaperDeng (approved):

rnwang04:
# Update 9/1
To make this demo run faster, there are some modifications:
- replace Resnet50 with Resnet18
- Only finetune 2 epochs
- Use subset of validation dataset(80 samples) to calculate accuracy
 
Now, fitune cost about 2 min and inference cost about 4 min (on cpx1 which has 112 cores). 
@jason-dai @TheaperDeng

python/nano/example/pytorch/inference_pipeline/resnet/inference_pipeline.py
```diff
@@ -0,0 +1,67 @@
...
+    optimizer = InferenceOptimizer()
+    # optimize may take about 2 minutes to run all possible accelaration combinations
+    optimizer.optimize(model=model,
+                       # To obtain the latency of single sample, set batch_size=1
+                       training_data=datamodule.train_dataloader(batch_size=1),
+                       # here we only take part samples to calculate a rough accuracy
+                       validation_data=datamodule.val_dataloader(limit_num_samples=160),
+                       metric=accuracy,
+                       direction="max",
+                       cpu_num=1,
```
TheaperDeng, line 49: I am not sure if set this `cpu_num` to 1 is proper at least for now. Since we know there are many limitation and know issues in our resource control for onnxrt and openvino. 

Would it be more reasonable if we just leave this `cpu_num` to be None which means all cores can be used without restrictions.

TheaperDeng:
will first merge this PR and fix any issues come up later

PR #5593, opened 2022-08-31
[PPML] Update writing TPC-H example time output to hdfs path

jenniew:
Update writing TPC-H example time output to hdfs path.


jenniew:
Jenkins pass: http://10.112.231.51:18889/job/BigDL-PR-scala-ppml-unittests-Validation/452/

PR #5599, opened 2022-08-31
Nano : vision transformers demo for InferenceOptimizer

rnwang04:
## Description

Use several vision transformers as an example to illustrate how to apply InferenceOptimizer to quickly find acceleration method with the minimum inference latency.

### 1. Why the change?

To provide a simple example illustrating how to use InferenceOptimizer.

### 2. User API changes

no changes.

### 3. Summary of the change

Basic example for usage of InferenceOptimizer.
Maybe as a third-party example.

### 4. How to test?
- [x] Unit test
- [x] local test


python/nano/example/pytorch/inference_pipeline/vision_transformers/vision_transformer.ipynb
```diff
@@ -0,0 +1,952 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "> ⚠️ **Warning**\n",
+    ">\n",
+    "> Swin don't support dynamic batch, so the batch_size of faked_train_dataloader must be the same with val_dataloader.\n",
+    ">\n",
+    "> Otherwise the accuracy will be really slow."
```
jason-dai, line 755: what do you mean by "accuracy is slow"?

rnwang04: sorry, here is a typo. Change to "accuracy will be very low."

TheaperDeng:
- Typically we guide our users to create a conda environment as this
  
  `conda create -n nano python=3.7 setuptools=58.0.4`
  
  `setuptools` will not affect the inference optimization, but we do want to avoid any possiblity to cause strange error.
  
- Do we really need this one? `pip install jsonargparse[signatures]`
  
- "then unset environment variable `KMP_AFFINITY`." we don't need to unse it anymore.
  
- what does `val_dataloader.dataset.dataset.transform = val_dataloader.dataset.transform` stands for
  
- maybe we can clean In[4] output as well?
  
- Should we show how to get the model using `get_best_model` and how to save / load them?

rnwang04:
> * Typically we guide our users to create a conda environment as this
>   `conda create -n nano python=3.7 setuptools=58.0.4`
>   `setuptools` will not affect the inference optimization, but we do want to avoid any possiblity to cause strange error.
> * Do we really need this one? `pip install jsonargparse[signatures]`
> * "then unset environment variable `KMP_AFFINITY`." we don't need to unse it anymore.
> * what does `val_dataloader.dataset.dataset.transform = val_dataloader.dataset.transform` stands for
> * maybe we can clean In[4] output as well?
> * Should we show how to get the model using `get_best_model` and how to save / load them?

Good catch, thanks ! Have fixed all these problems, could you please take a look again ?

rnwang04:
> * I think a better we is to recommend [Miniconda — conda documentation](https://docs.conda.io/en/latest/miniconda.html) rather than anaconda
> * Today, vision Transformer is becoming more and more popular. On one hand, people are constantly searching for larger pre-training corpus and pre-training model, on the other hand, how to deploy the vision transformer in the industrial scene is also a very sought-after issue.
> * Here we take several popular vision Transformer architectures as examples to demonstrate how to use InferenceOptimizer in BigDL-Nano to accelerate model's inference speed.
> * Some thoughts(no need to change this in this PR): "Swin don't support dynamic batch", should we warning users in output?
> * we should use inferenceoptimizer's save and load?

Thanks! Changes have been made according to the above suggestions, except that 4. 4 will be considered in the future.

python/nano/example/pytorch/inference_pipeline/vision_transformers/vision_transformer.ipynb
```diff
@@ -0,0 +1,795 @@
...
+   "outputs": [],
+   "source": [
+    "InferenceOptimizer.save(acc_model, path=\"ckpt\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Then load it by `Trainer.load`"
```
jason-dai, line 749: Then load it by `InferenceOptimizer.load`

We need to show case how the user can export the optimized model in standard format (PyTorch, ONNX, OpenVINO, etc.)

rnwang04: ` InferenceOptimizer.load` -> will fix this bug in next PR.

`how the user can export the optimized model in standard format (PyTorch, ONNX, OpenVINO)`  this case should be put in this example or another independent example ?

jason-dai: Maybe another example?

PR #5601, opened 2022-08-31
add orca horovod tf2 test to github actions

pinggao187:
## Description

add orca horovod tf2 test to github actions


.github/actions/orca/orca-python-horovod-tf2-py37-spark3-action/action.yml
```diff
@@ -0,0 +1,57 @@
+name: 'Run Orca Python Horovod Tf2 Py37 Spark3'
```
liu-shaojun, line 1: will this file be removed later?

.github/workflows/nightly_test.yml
```diff
@@ -95,6 +101,21 @@ jobs:
...
+    - uses: actions/checkout@v3
+    - name: Set up JDK8
+      uses: ./.github/actions/jdk-setup-action
+    - name: Set up maven
+      uses: ./.github/actions/maven-setup-action 
+    - name: Setup env
+      uses: ./.github/actions/orca/setup-env/setup-horovod-tf2-py37
+    - name: Run Test
+      uses: ./.github/actions/orca/orca-python-horovod-tf2-py37-spark3-action/nightly-test
+
```
liu-shaojun, line 118: Will a remove-env action be added?

.github/actions/orca/orca-python-horovod-tf2-py37-spark3-action/action.yml
pinggao187: It will be removed after the test is successful

.github/workflows/nightly_test.yml
pinggao187: removing the env will cause each test to take too long, will no longer add remove-env.

liu-shaojun (approved):

Issue #5444, opened 2022-08-17
Chronos: make `roll=True` the default value for `to_torch_data_loader`


TheaperDeng:
This parameter is quite difficult to be understood by our users while they nearly always need to set this one to `True` (current default value is `False`).

plusbang:
The source code of `to_torch_data_loader` and scripts calling `to_torch_data_loader` have been modified. (#5605)

PR #5605, opened 2022-08-31
Chronos: make roll=True the default value for to_torch_data_loader()

plusbang:
## Description

The roll parameter in to_torch_data_loader() is currently default to be false. However, our users always need to set this one to True.

### 1. Why the change?

https://github.com/intel-analytics/BigDL/issues/5444

### 3. Summary of the change 

Change the source code of to_torch_data_loader() and scripts calling to_torch_data_loader().

### 4. How to test?
- [x] Unit test
(jenkins: http://10.112.231.51:18889/view/BigDL-PR-Validation/job/BigDL-Chronos-PR-Validation/779/)
- [ ] Document test


TheaperDeng (approved):
LGTM, but this PR definitely teach us a lesson, it's really difficult to make sure if all the places needed to be changed is fixed. Pls run pr validation.

PR #5607, opened 2022-08-31
Pytorch VFL NN refactor, refine and example test

Litchilitchy:
### 1. Why the change?
To refactor NN logics, and add example test

### 2. User API changes
Users should now set client ID and target URL by
```python
from bigdl.ppml.fl.nn.fl_context import init_fl_context
init_fl_context(client_id, target)
```
The params in `Estimator.from_torch(client_id=..., target=...)` are removed and not available now.

### 3. Summary of the change 
client ID and channel target URL can now only be set by `init_fl_context(client_id, target)`

client_id and target become singleton in `FLClient`, and separate the logic of NN to another class `NNClient` (the same level with `PSI`) which calls `FLClient`

removed all these params settings in old tutorial, examples and tests, and added `init_fl_context(client_id, target)` instead

### 4. How to test?
Added a new example test


python/ppml/example/pytorch_nn_lr/pytorch_nn_lr_1.py
```diff
@@ -50,8 +54,10 @@ def forward(self, x: List[Tensor]):
 
 @click.command()
 @click.option('--load_model', default=False)
-def run_client(load_model):
-    df_train = pd.read_csv('./data/diabetes-vfl-1.csv')    
+@click.option('--data_path', default="./data/diabetes-vfl-1.csv")
+def run_client(load_model, data_path):
+    init_fl_context('1')
```
jason-dai, line 59: should `id` will be an integer instead?

python/ppml/src/bigdl/ppml/fl/nn/fl_context.py
```diff
@@ -17,6 +17,8 @@
 from ..nn.fl_client import FLClient
 
 def init_fl_context(client_id, target="localhost:8980"):
```
jason-dai, line 19: add python doc

python/ppml/example/pytorch_nn_lr/pytorch_nn_lr_1.py
Litchilitchy: We can do this. And user must pass an integer when calling `init_fl_context` and the client ID could only be set in this method.

All the protobuf types of ID are String now, I would open another PR to modify all of them.

python/ppml/example/pytorch_nn_lr/pytorch_nn_lr_1.py
```diff
@@ -83,11 +87,9 @@ def run_client(load_model):
...
         ppl = Estimator.from_torch(client_model=model,
-                                   client_id='1',
                                    loss_fn=loss_fn,
                                    optimizer_cls=torch.optim.SGD,
-                                   optimizer_args={'lr':1e-5},
-                                   target='localhost:8980',
+                                   optimizer_args={'lr':1e-4},
                                    server_model=server_model,
                                    server_model_path='/tmp/pytorch_server_model',
                                    client_model_path='/tmp/pytorch_client_model_1.pt')
```
jason-dai, line 95: save after `fit`?

Litchilitchy: Estimator would automatically save after each epoch of `fit` like Orca Estimator.

Issue #5610, opened 2022-09-01
Nano: Enhancement and bug fix of InferenceOptimizer


rnwang04:
1. when user don't specify the metric and validation_data, there will be some bug when call "get_best_model"
![image](https://user-images.githubusercontent.com/105281011/187812677-7c732edf-792f-44b0-bb68-644f45d5f8c0.png)

2. dynamic control the sample_num for calculating latency to avoid the overall time of inference pipeline is too long

3. We should automaticly print these information at the end of optimize()
```
accleration option: original, latency: 133.7490ms, accuracy: 0.9890
accleration option: fp32_ipex, latency: 102.1587ms, accuracy: 0.9890
accleration option: bf16_ipex, latency: 628.2516ms, accuracy: 0.9890
accleration option: int8, latency: 34.6394ms, accuracy: 0.9880
accleration option: openvino_fp32, latency: 31.9775ms, accuracy: 0.9890
accleration option: openvino_int8, latency: 17.0428ms, accuracy: 0.9870
accleration option: onnxruntime_fp32, latency: 121.4465ms, accuracy: 0.9890
accleration option: onnxruntime_int8_qlinear, latency: 19.3709ms, accuracy: 0.9870
accleration option: onnxruntime_int8_integer, latency: 43.3232ms, accuracy: 0.9860
accleration option: jit_fp32, latency: 103.0402ms, accuracy: 0.9890
accleration option: jit_fp32_ipex, latency: 102.9166ms, accuracy: 0.9890
accleration option: jit_fp32_ipex_clast, latency: 55.8827ms, accuracy: 0.9890
```
Rather than let user print these information by following commands
```
for key, value in optimizer.optimized_model_dict.items():
        print("accleration option: {}, latency: {:.4f}ms, accuracy: {:.4f}".format(key, value["latency"], value["accuracy"]))
```

Issue #5589, opened 2022-08-30
Nano: inference pipeline metric evaluation sometimes may bring some wrong answers


TheaperDeng:
As said in this TODO, this implementation is only valid for dataloader with all batchsizes are the same

https://github.com/intel-analytics/BigDL/blob/1b45bf03830cd5c47edaf23bd78705edd5143313/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py#L436-L444

jason-dai:
Maybe we should check and report error (instead of failing silently)

TheaperDeng:
> Maybe we should check and report error (instead of failing silently)

I think we need to fix it directly since 1. this bug will only affect minorly to the final metric value 2. it happens very frequently since sample_num/batch_size is not often to be an integer.

Some possible implementation can be:
```python
def _accuracy_calculate_helper(model, metric, data): 
     sample_num_count = 0
     # ...
     for i, (data_input, target) in enumerate(data): 
        # ...
        sample_num_count += data_input.shape[0]
    return np.sum(metric_list) / sample_num_count
```

PR #5611, opened 2022-09-01
Nano : fix and enhancement of InferenceOptimizer

rnwang04:
## Description

Fix https://github.com/intel-analytics/BigDL/issues/5609, https://github.com/intel-analytics/BigDL/issues/5610, https://github.com/intel-analytics/BigDL/issues/5589

### 1. Why the change?

Fix some bugs and provide some inhancements for InferenceOptimizer

### 2. User API changes

No change.

### 3. Summary of the change 

- [x] fix accuracy KeyError when no metric and validation in optimize
- [x] add unittest for no metric
- [x] add model.eval
- [x] automatically print accelaration results at the end of optimize
- [x] dynamic control of latency_sample_num
- [x] add unittest for get_best_model before optimize
- [x] fix accuracy calculating

### 4. How to test?
- [x] Unit test
- [x] Jenkins
http://10.112.231.51:18889/job/ZOO-PR-NanoTests/384/

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
+++ b/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
@@ -440,15 +440,22 @@ def _available_acceleration_combination():
 
 def _throughput_calculate_helper(iterrun, func, *args):
     '''
-    A simple helper to calculate median latency
+    A simple helper to calculate average latency
     '''
+    start_time = time.perf_counter()
     time_list = []
-    for _ in range(iterrun):
-        st = time.time()
+    for i in range(iterrun):
+        st = time.perf_counter()
         func(*args)
-        time_list.append(time.time() - st)
+        end = time.perf_counter()
+        time_list.append(end - st)
+        # at least need 10 iters and try to control calculation
+        # time less than 2 min
+        if i+1 >= min(iterrun, 10) and (end - start_time) > 2:
+            iterrun = i+1
+            break
     time_list.sort()
-    # remove top and least 10% data.ni
+    # remove top and least 10% data
     time_list = time_list[int(0.1 * iterrun): int(0.9 * iterrun)]
     return np.mean(time_list) * 1000
 
--- a/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
+++ b/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
@@ -251,11 +251,11 @@ class InferenceOptimizer:
         if self._calculate_accuracy:
             for key, value in self.optimized_model_dict.items():
                 print("accleration option: {}, latency: {:.4f}ms, accuracy : {:.4f}"
-                        .format(key, value["latency"], value["accuracy"]))
+                      .format(key, value["latency"], value["accuracy"]))
         else:
             for key, value in self.optimized_model_dict.items():
-                print("accleration option: {}, latency: {:.4f}ms : {:.4f}"
-                        .format(key, value["latency"]))
+                print("accleration option: {}, latency: {:.4f}ms :"
+                      .format(key, value["latency"]))
 
     def get_best_model(self,
                        accelerator: str = None,
--- a/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
+++ b/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
@@ -242,8 +242,8 @@ class InferenceOptimizer:
                                                    metric, validation_data)
                 else:
                     result_map[method]["accuracy"] = None
-                result_map[method]["model"] = acce_model
 
+                result_map[method]["model"] = acce_model
             else:
                 pass
 
--- a/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
+++ b/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
@@ -109,7 +109,7 @@ class InferenceOptimizer:
         '''
         This function will give all available inference acceleration methods a try
         and record the latency, accuracy and model instance inside the Optimizer for
-        future usage.
+        future usage. All model instance is setting to eval mode.
 
         :param model: A nn.module to be optimized
         :param training_data: A pytorch dataloader for training dataset.
@@ -165,7 +165,6 @@ class InferenceOptimizer:
 
         result_map: Dict[str, Dict] = {}
 
-        training_state = model.training
         model.eval()  # change model to eval state
 
         for method, available in available_dict.items():
@@ -246,10 +245,6 @@ class InferenceOptimizer:
                 result_map[method]["model"] = acce_model
             else:
                 pass
-
-        # restore model's state
-        if training_state:
-            model.train()
 
         self.optimized_model_dict: Dict = result_map
         print("==========================Optimization Results==========================")
--- a/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
+++ b/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
@@ -180,11 +180,11 @@ class InferenceOptimizer:
                     input_sample = tuple(next(iter(training_data))[:-1])
                     try:
                         if accelerator is None and use_ipex is False:
-                            acce_model = model
+                            acce_model = deepcopy(model)
                         else:
                             if accelerator in ("jit", None):
                                 acce_model = \
-                                    InferenceOptimizer.trace(model=model,
+                                    InferenceOptimizer.trace(model=deepcopy(model),
                                                              accelerator=accelerator,
                                                              use_ipex=use_ipex,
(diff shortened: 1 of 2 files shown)


python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -160,6 +165,8 @@ def optimize(self, model: nn.Module,
 
         result_map: Dict[str, Dict] = {}
 
+        model.eval()  # change model to eval state
```
jason-dai, line 168: Can you first record model's status, and restore it after we are done?

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -420,15 +440,22 @@ def _available_acceleration_combination():
...
     '''
+    start_time = time.perf_counter()
     time_list = []
-    for _ in range(iterrun):
-        st = time.time()
+    for i in range(iterrun):
+        st = time.perf_counter()
         func(*args)
-        time_list.append(time.time() - st)
+        end = time.perf_counter()
```
TheaperDeng, line 450: This perf counter might be a good bkm for future time measurement

python/nano/test/pytorch/tests/test_inference_pipeline_ipex.py
```diff
@@ -69,25 +69,54 @@ class TestInferencePipeline(TestCase):
...
+        acc_model(next(iter(self.train_loader))[0])
+
+    def test_pipeline_without_metric(self):
+        inference_opt = InferenceOptimizer()
+        inference_opt.optimize(model=self.model,
+                               training_data=self.train_loader,
+                               cpu_num=1)
+
+        acc_model, option = inference_opt.get_best_model()
+        print(option)
```
TheaperDeng, line 106: no need to print

python/nano/test/pytorch/tests/test_inference_pipeline_ipex.py
```diff
@@ -69,25 +69,54 @@ class TestInferencePipeline(TestCase):
...
+        acc_model, option = inference_opt.get_best_model(accelerator="onnxruntime")
+        assert option == "" or "onnxruntime" in option
+        acc_model, option = inference_opt.get_best_model(precision="int8")
+        assert option == "" or "inc" in option
```
TheaperDeng, line 110: `assert option == "" or "inc" in option or "pot" in option`?

python/nano/test/pytorch/tests/test_inference_pipeline_ipex.py
```diff
@@ -69,25 +69,54 @@ class TestInferencePipeline(TestCase):
...
+        assert error_msg == "If you want to specify accuracy_criterion, you need "\
+                            "to set metric and validation_data when call 'optimize'."
+
 
 if __name__ == "__main__":
     test = TestInferencePipeline()
-    test.test_pipeline()
+    test.test_get_model_without_optimize()
+    test.test_pipeline_with_metric()
+    test.test_pipeline_without_metric()
```
TheaperDeng, line 122: no need to put this main program here, pytest will capture all the legal unittest automatically.

python/nano/test/pytorch/tests/test_inference_pipeline_ipex.py
rnwang04: done

python/nano/test/pytorch/tests/test_inference_pipeline_ipex.py
rnwang04: done

python/nano/test/pytorch/tests/test_inference_pipeline_ipex.py
rnwang04: Good catch! done

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: Yes. After update, now I record model's status by `model.training` at the beginning of `optimize`, and restore it at the end of `optimize`.
So if a model is in training status, after `optimize`, it is still in training status, but all accelarated models will be in eval status always.

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -228,17 +236,31 @@ def func_test(model, input_sample):
...
+                else:
+                    result_map[method]["accuracy"] = None
                 result_map[method]["model"] = acce_model
 
             else:
                 pass
 
+        # restore model's state
+        if training_state:
+            model.train()
```
TheaperDeng, line 252: should we consider if `self.optimized_model_dict["original"]["model"]` be set to training model? Since we directedly set model without deepcopy https://github.com/intel-analytics/BigDL/blob/b455568a73ca92683bd402c2a617bd15a40ab775/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py#L183 that's not good for users.

rnwang04: Good catch! Add deepcopy. Now only the original model will be set to training mode(if it is in training mode before optimize) and all accelerated model(include "original") will be set in eval mode.

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -179,19 +180,19 @@ def optimize(self, model: nn.Module,
                     input_sample = tuple(next(iter(training_data))[:-1])
                     try:
                         if accelerator is None and use_ipex is False:
-                            acce_model = model
+                            acce_model = deepcopy(model)
```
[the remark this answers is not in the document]
rnwang04: Here I use deepcopy to make `self.optimized_model_dict["original"]["model"]` always in eval mode no matter whether original model in training mode.
If remove deepcopy, when user pass a model in training mode,  `self.optimized_model_dict["original"]["model"]` will also be training mode. So It cannot be guaranteed that the models saved by optimize are all eval model. Is this behavior reasonable?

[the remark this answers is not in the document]
jason-dai: Will this be an issue if you are optimizing, say, BERT-Large? Will making a deep copy be a problem?

It is probably reasonable that the "original" model is the same as the original one?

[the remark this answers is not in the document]
TheaperDeng: Users may get accelerated model from `optimizer.get_best_model(...)`, but if all accelerated method is not valid (either failed or does not meet user's accuracy requirement), we will let users get the original model.

Users may familiar this usage:
```python
acc_model = optimizer.get_best_model()
acc_model(x)
```

So, that's why we want to make the returned model in eval mode.

Well, it's just a minor point and it may does cause some mem issue. maybe we will just note our user the original model is the same reference as the model before optimization (i.e. they have the same training/eval mode)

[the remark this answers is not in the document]
rnwang04: > Will this be an issue if you are optimizing, say, BERT-Large? Will making a deep copy be a problem in this case?
> 
> It is probably reasonable that the "original" model is the same as the original one?

Yeah, I agree. I will remove this deepcopy, and we will maybe note our user the original model is the same reference as the model before optimization as junwei comment.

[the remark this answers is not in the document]
jason-dai: > Users may get accelerated model from `optimizer.get_best_model(...)`, but if all accelerated method is not valid (either failed or does not meet user's accuracy requirement), we will let users get the original model.
> 
> Users may familiar this usage:
> 
> ```python
> acc_model = optimizer.get_best_model()
> acc_model(x)
> ```
> 
> So, that's why we want to make the returned model in eval mode.
> 
> Well, it's just a minor point and it may does cause some mem issue. maybe we will just note our user the original model is the same reference as the model before optimization (i.e. they have the same training/eval mode)

Maybe a simpler semantics is to let the user know that `InferenceOptimizer` will always set all models (including the original one) to `eval`?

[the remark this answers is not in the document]
rnwang04: So we don't restore model's status at the end of `optimize`, and just remind user that all models will be set to `eval`？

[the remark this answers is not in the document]
jason-dai: Yes, I think it's better than deepcopy; to the user, `optimize` will always set `eval`, which is easier to understand.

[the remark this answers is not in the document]
rnwang04: Done.

Issue #5507, opened 2022-08-23
[Nano] OpenVINO API for Async Inference 


zhentaocc:
### Problem 
Nano has enabled `OpenVINOModel.forward` with sync mode. For example:
```python
from bigdl.nano.openvino import OpenVINOModel
ov_model = OpenVINOModel("model.xml")
for img in dataset:
    out = ov_model(img)
```
This will get the results batch by batch. As for async mode in https://docs.openvino.ai/latest/openvino_docs_OV_UG_Python_API_exclusives.html#asyncinferqueue, we didn't enable this for Nano. With async inference, we can obtain better performance.

### API Design
For raw OpenVINO:
```python
from bigdl.nano.openvino import OpenVINOModel
ov_model = OpenVINOModel("model.xml")
model.predict(dataset: np_array)
```
`OpenVINOModel.precit(...)` will be the function to enable async inference loop.

For Pytorch OpenVINO"
```python
from bigdl.nano.pytorch.trainer import Trainer
trainer = Trainer()
ov_model = Trainer.trace(pytorch_model, accelerator='openvino')
trainer.predict(ov_model, dataloader)
```
`Trainer.predict()` will be overridden to run async inference loop.

### Tasks
assigned to @hjzin 
1. enable async API for raw OpenVINO
2. enable async API for Pytorch OpenVINO
3. modify https://github.com/intel-analytics/BigDL/pull/5491

jason-dai:
Why do we need two different sets of APIs? Make it consistent for this specific feature.

zhentaocc:
> Why do we need two different sets of APIs? Make it consistent for this specific feature.

Because they are for different frameworks actually.
There is no Trainer for raw OpenVINO since we don't require Pytorch-Lightning installation in this case.  
For pytorch-lightning:
```python
trainer.predict(ov_model, dataloader)
```
For Keras:
```python
ov_model.predict(dataloader)
```
For OpenVINO without pytorch/tf:
```python
ov_model.predict(data)
```
They seem to be kind of consistent.

jason-dai:
I think we can then have all cases to use `model.predict`, no? And `model.predict` can support both sync and async modes.

zhentaocc:
> I think we can then have all cases to use `model.predict`, no? 

Sure we can do this. This will introduce a new API for Pytorch users, `model.predict(...)`. Shall we make `trainer.predict` same as `model.predict(...)`? Users might be confused with these two API with the same name.

[...]

PR #5615, opened 2022-09-01
Add Nano openvino async api

hjzin:
## Description

Add OpenVINO API for Async Inference

### 1. Why the change?

Releated issue: [issue#5507](https://github.com/intel-analytics/BigDL/issues/5507)

### 2. User API changes

**For OpenVINOModel**
```python
# input data should be List of numpy array
x = [np.random.randn(1, 3, 224, 224) for i in range(5)]

# do async inference
# async inference api usage: openvino_model.async_predict(input_data, num_requests)
# param input_data: Input data to be inferenced.
#                  Users can put multiple input data in a list to infer them together.
#                  Can be List[numpy.ndarray] or
#                  List[List[numpy.ndarray]] if the model has multiple inputs.
# param num_requests: Number of requests in the asynchronous infer requests pool.
#                     Defaults to 0.
#                     If 0, it will be set automatically to the optimal number.
# return value: a List containing the infer result of each input array in x
# return value type: List[numpy.array]
result = openvino_model.async_predict(input_data=x, num_requests=5)
```
**For PytorchOpenVINOModel**
```python
from torch.utils.data.dataloader import DataLoader

trainer = Trainer()

y = torch.ones((10, ), dtype=torch.long)

pytorch_openvino_model = trainer.trace(model, input_sample=x, accelerator='openvino')

# async inference api usage: pytorch_openvino_model.async_predict(input_data, num_requests)
# param input_data: Input data to be inferenced.
#                   Users can put multiple input data in a list or
#                   put all data in a DataLoader to infer and get results of all input data
#                   Can be torch.utils.data.dataloader.DataLoader and
#                   List[torch.Tensor] or
#                   List[List[torch.Tensor]] if the model has multiple inputs.
#                   If input_data is a DataLoader object,the format in DataLoader should be
#                   (x1, x2, ..., xn, y).
# param jobs: same as OpenVINOModel
# return value: List of output Tensor of each input data (same as sync inference)
# return value type: List[torch.Tensor]

# do async_predict use dataloader as input

# wrap data as a torch.utils.data.dataloader.DataLoader object
ds = TensorDataset(x, y)
dataloader = DataLoader(ds, batch_size=2)

# will infer all data in dataloader batch by batch
result = pytorch_openvino_model.async_predict(dataloader,num_requests=3)

# do async_predict use List of torch.Tensor as input
x = [torch.rand((10, 3, 256, 256)) for i in range(3)]
result = openvino_model.async_predict(x, num_requests=3)


```
### 3. Summary of the change 

- use openvino `AsyncInferQueue` api to implement async inference.
- user can put multiple input data in a list or a dataloader to infer and get results of all input data.
- make the output format same as `PytorchOpenVINOModel` when doing inference with `OpenVINOModel` 

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/deps/openvino/core/model.py
+++ b/python/nano/src/bigdl/nano/deps/openvino/core/model.py
@@ -15,6 +15,7 @@
 #
 from pathlib import Path
 from tempfile import TemporaryDirectory
+from typing import List, Union  # for typehint
 from openvino.runtime import Core
 from bigdl.nano.utils.log4Error import invalidInputError
 from openvino.runtime import Model
@@ -177,7 +178,9 @@ class OpenVINOModel:
     def _model_exists_or_err(self):
         invalidInputError(self.ie_network is not None, "self.ie_network shouldn't be None.")
 
-    def async_predict(self, input_data, num_infer_requests=0):
+    def async_predict(self,
+                      input_data: Union[List[np.ndarray], List[List[np.ndarray]]],
+                      num_requests: int = 0) -> List[np.ndarray]:
         """
         Perfrom model inference using async mode.
 
@@ -185,9 +188,11 @@ class OpenVINOModel:
                            Users can put multiple input data in a list to infer them together.
                            Can be List[numpy.ndarray] or
                            List[List[numpy.ndarray]] if the model has multiple inputs.
-        :param num_infer_requests: Number of infer requests in the AsyncInferQueue, default to 0.
-                                   If 0, it will be set automatically in
-                                   constructor of AsyncInferQueue.
+        :param num_requests: Number of requests in the asynchronous infer requests pool.
+                             Each element in input_data will be bound to an idle async
+                             infer request in the pool to do inference.
+                             Defaults to 0.
+                             If 0, it will be set automatically to the optimal number.
 
         :return: A List containing result of each inference. Type: List[numpy.ndarray]
         """
@@ -197,7 +202,7 @@ class OpenVINOModel:
         def call_back(requests, idx):
             results[idx] = self.on_forward_end(requests.results)
 
-        infer_queue = AsyncInferQueue(self._compiled_model, jobs=num_infer_requests)
+        infer_queue = AsyncInferQueue(self._compiled_model, jobs=num_requests)
         infer_queue.set_callback(call_back)
 
         for id, model_input in enumerate(input_data):
--- a/python/nano/src/bigdl/nano/deps/openvino/core/model.py
+++ b/python/nano/src/bigdl/nano/deps/openvino/core/model.py
@@ -157,10 +157,9 @@ class OpenVINOModel:
             model.reshape(orig_shape)
         return model
 
-<<<<<<< HEAD
     def _model_exists_or_err(self):
         invalidInputError(self.ie_network is not None, "self.ie_network shouldn't be None.")
-=======
+
     def async_predict(self, inputs, jobs):
         """
         Perfrom model inference using async mode.
@@ -186,6 +185,5 @@ class OpenVINOModel:
             infer_queue.start_async([model_input], userdata=id)
 
         infer_queue.wait_all()
- 
+
         return results
->>>>>>> add openvino async_predict api
--- a/python/nano/src/bigdl/nano/deps/openvino/core/model.py
+++ b/python/nano/src/bigdl/nano/deps/openvino/core/model.py
@@ -20,6 +20,7 @@ from bigdl.nano.utils.log4Error import invalidInputError
 from openvino.runtime import Model
 from .utils import save
 from openvino.runtime import AsyncInferQueue
+import numpy as np
 
 class OpenVINOModel:
     def __init__(self, ie_network: str, device='CPU', thread_num=None):
@@ -28,11 +29,27 @@ class OpenVINOModel:
         self.thread_num = thread_num
         self.ie_network = ie_network
 
+    def on_forward_start(self, inputs):
+        invalidInputError(self.ie_network,
+                          "self.ie_network shouldn't be None.")
+        return inputs
+
     def forward_step(self, *inputs):
         return self._infer_request.infer(list(inputs))
 
+    def on_forward_end(self, outputs):
+        arrays = tuple(map(lambda x: x, outputs.values()))
+        if len(arrays) == 1:
+            arrays = arrays[0]
+        return arrays
+
+    def forward(self, *inputs):
+        inputs = self.on_forward_start(inputs)
+        outputs = self.forward_step(*inputs)
+        return self.on_forward_end(outputs)
+
     def __call__(self, *inputs):
-        return self.forward_step(*inputs)
+        return self.forward(*inputs)
 
     @property
     def forward_args(self):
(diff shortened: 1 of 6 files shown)


python/nano/src/bigdl/nano/deps/openvino/core/model.py
```diff
@@ -148,3 +149,31 @@ def pot(self,
...
+        Perfrom model inference using async mode.
+
+        :param inputs: List of input data
+        :type List[numpy.array]
+        :param jobs: numer of infer requests in the AsyncInferQueue
+        :type int
+
+        :return A List containing result of each input
+        :rtype List[Dict[openvino.runtime.ConstOutput, numpy.array]]
+        """
```
zhentaocc, line 164: please use consistent format of docstring.

zhentaocc, line 164: return np.array or a list of np.array

python/nano/src/bigdl/nano/deps/openvino/core/model.py
```diff
@@ -148,3 +149,31 @@ def pot(self,
...
+
+        infer_queue = AsyncInferQueue(self._compiled_model, jobs=jobs)
+        infer_queue.set_callback(call_back)
+
+        for id, model_input in enumerate(inputs):
+            infer_queue.start_async([model_input], userdata=id)
+
+        infer_queue.wait_all()
+
+        return results
```
zhentaocc, line 179: concat all data?

python/nano/src/bigdl/nano/deps/openvino/pytorch/model.py
```diff
@@ -99,3 +100,31 @@ def pot(self,
...
+        :param jobs: numer of infer requests in the AsyncInferQueue
+        :type int
+
+        :return A List containing result of each input
+        :rtype List[torch.Tensor]
+        """
+        if isinstance(inputs, DataLoader):
+            input_data = []
+            for data in inputs:
+                input_data.append(data[0])
```
zhentaocc, line 120: How about a model with multiple inputs?

python/nano/test/openvino/basic/test_openvino.py
```diff
@@ -26,3 +26,11 @@ def test_openvino_model(self):
         x = np.random.randn(1, 3, 224, 224)
         y_hat = openvino_model.forward_step(x)
         assert tuple(next(iter(y_hat)).shape) == (1, 1000)
+
+    def test_openvino_model_async_predict(self):
+        openvino_model = OpenVINOModel("./intel/resnet18-xnor-binary-onnx-0001/FP16-INT1/resnet18-xnor-binary-onnx-0001.xml")
+        x = [np.random.randn(1, 3, 224, 224) for i in range(5)]
```
zhentaocc, line 32: I think a more common case is that `x =np.random.randn(5, 3, 224, 224) ` ?

python/nano/test/openvino/basic/test_openvino.py
```diff
@@ -26,3 +26,11 @@ def test_openvino_model(self):
...
+
+        result = openvino_model.async_predict(x,jobs=5)
+        for res in result:
+            assert tuple(next(iter(res)).shape) == (1, 1000)
```
zhentaocc, line 36: How about making `result.shape==(5,1000)`

python/nano/test/openvino/pytorch/test_openvino.py
```diff
@@ -69,3 +69,25 @@ def test_trainer_save_openvino(self):
...
+        ds = TensorDataset(x, y)
+        dataloader = DataLoader(ds, batch_size=2)
+
+        # do async_predict use dataloader as input
+        result = openvino_model.async_predict(dataloader, jobs=3)
+        for res in result:
+            assert res.shape == (2, 10)
+
+        # do async_predict use List of Tensor as input
+        x = [torch.rand((10, 3, 256, 256)) for i in range(3)]
```
zhentaocc, line 90: same as above

python/nano/test/openvino/pytorch/test_openvino.py
```diff
@@ -69,3 +69,25 @@ def test_trainer_save_openvino(self):
...
+        result = openvino_model.async_predict(x, jobs=3)
+        for res in result:
+            assert res.shape == (10, 10)
```
zhentaocc, line 93: same as above

python/nano/src/bigdl/nano/deps/openvino/core/model.py
```diff
@@ -148,3 +149,31 @@ def pot(self,
             model = Core().read_model(model_path)
             model.reshape(orig_shape)
         return model
+
+    def async_predict(self, inputs, jobs):
```
zhentaocc, line 153: any BKM for `job`? Can we set the job number according to OS environment(n_cores?) automatically? Otherwise, please give a default value to `jobs`.

zhentaocc (changes requested):
I think if the model has only one input, it's better to have input type as one np.array or torch.Tensor instead of a List?

python/nano/test/openvino/basic/test_openvino.py
hjzin: It's not a batch of data, line #32 means there are 5 inputs and we are going to do inference 5 times. So I think it's better to use a list of arrays than concatenate them together?

python/nano/src/bigdl/nano/deps/openvino/core/model.py
jason-dai: change `jobs` to `concurrency` or `parallelism?

hjzin: > change `jobs` to `concurrency` or `parallelism`?

The documentation of OpenVINO  doesn't mention that async infer requests in the `AsyncInferQueue` are performed in parallel, it just says `start_async` function will do async inference and return immediately without blocking until the inference finishes, so I think change to `concurrency` or `parallelism` may be a little inaccurate?
Since `jobs` means number of infer requests will be created in the `AsyncInferQueue`, how about change to something like `num_infer_requests` ?

Here are the releated docs:
https://docs.openvino.ai/latest/api/ie_python_api/_autosummary/openvino.runtime.AsyncInferQueue.html#openvino.runtime.AsyncInferQueue.start_async
https://docs.openvino.ai/latest/api/ie_python_api/_autosummary/openvino.runtime.InferRequest.html#openvino.runtime.InferRequest.start_async

hjzin: > any BKM for `job`? Can we set the job number according to OS environment(n_cores?) automatically? Otherwise, please give a default value to `jobs`.

OK. 
Since `jobs` defaults to 0 in definition of `AsyncInferQueue`, and if it's set to 0, the value will be automatically set by OpenVINO, so I think we can keep this default value.
https://docs.openvino.ai/latest/api/ie_python_api/_autosummary/openvino.runtime.AsyncInferQueue.html#openvino.runtime.AsyncInferQueue.__init__

jason-dai: > `num_infer_requests`

just `num_requests`

python/nano/src/bigdl/nano/deps/openvino/core/model.py
```diff
@@ -148,3 +166,37 @@ def pot(self,
...
+
+    def async_predict(self, input_data, num_infer_requests=0):
+        """
+        Perfrom model inference using async mode.
+
+        :param input_data: Input data to be inferenced.
+                           Users can put multiple input data in a list to infer them together.
+                           Can be List[numpy.ndarray] or
+                           List[List[numpy.ndarray]] if the model has multiple inputs.
+        :param num_infer_requests: Number of infer requests in the AsyncInferQueue, default to 0.
```
jason-dai, line 178: I don't think the user will understand `AsyncInferQueue`; instead, explain it's a pool of asynchronous  interference requests.

python/nano/src/bigdl/nano/deps/openvino/core/model.py
```diff
@@ -148,3 +166,37 @@ def pot(self,
             model = Core().read_model(model_path)
             model.reshape(orig_shape)
         return model
+
+    def async_predict(self, input_data, num_infer_requests=0):
```
jason-dai, line 170: add type hint

python/nano/src/bigdl/nano/deps/openvino/pytorch/model.py
```diff
@@ -99,3 +100,50 @@ def pot(self,
                             maximal_drop=maximal_drop, max_iter_num=max_iter_num,
                             n_requests=n_requests, sample_size=sample_size)
         return PytorchOpenVINOModel(model)
+
+    def async_predict(self, input_data, num_infer_requests=0):
```
jason-dai, line 104: add type hint

python/nano/src/bigdl/nano/deps/openvino/core/model.py
```diff
@@ -148,3 +166,37 @@ def pot(self,
...
+        """
+        Perfrom model inference using async mode.
+
+        :param input_data: Input data to be inferenced.
+                           Users can put multiple input data in a list to infer them together.
+                           Can be List[numpy.ndarray] or
+                           List[List[numpy.ndarray]] if the model has multiple inputs.
+        :param num_infer_requests: Number of infer requests in the AsyncInferQueue, default to 0.
+                                   If 0, it will be set automatically in
```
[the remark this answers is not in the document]
hjzin: I will change these docstrings to make it more clearly to users.

python/nano/src/bigdl/nano/deps/openvino/core/model.py
```diff
@@ -153,3 +172,42 @@ def pot(self,
...
+
+        :param input_data: Input data to be inferenced.
+                           Users can put multiple input data in a list to infer them together.
+                           Can be List[numpy.ndarray] or
+                           List[List[numpy.ndarray]] if the model has multiple inputs.
+        :param num_requests: Number of requests in the asynchronous infer requests pool.
+                             Each element in input_data will be bound to an idle async
+                             infer request in the pool to do inference.
+                             Defaults to 0.
+                             If 0, it will be set automatically to the optimal number.
```
TheaperDeng, line 190: I understand what will happen if num_requests is set > 1. Could you have some test on the compatibility between nano's env variables(open by `source bigdl-nano-init`) and this feature?

TheaperDeng: I am worrying about the core binding of intel openmp might affect this function

hjzin: > I understand what will happen if num_requests is set > 1. Could you have some test on the compatibility between nano's env variables(open by `source bigdl-nano-init`) and this feature?

You mean run test of async api when the nano's env variables are set by `source bigdl-nano-init`?

hjzin:
jenkins test
http://10.112.231.51:18889/view/BigDL-PR-Validation/job/BigDL-Chronos-PR-Validation/844/

PR #5616, opened 2022-09-01
fix cmd injection

zhentaocc:
## Description
Fix command injection during model conversion:
Details:
the system passes user-supplied input directly to exec(), eval(), and os.system() methods without validation, which could result in command injection.

### 1. Why the change?

Fix command injection during model conversion

### 2. User API changes

N/A

### 3. Summary of the change 

use `subprocess.checkoutput` instead of `os.system`.

### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/deps/openvino/core/utils.py
+++ b/python/nano/src/bigdl/nano/deps/openvino/core/utils.py
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-import os
+import subprocess
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


TheaperDeng (approved):
LGTM

TheaperDeng:
just found this PR......, we definitely need to merge this. @zhentaocc do you have any changes need to do on this PR? I will merge it tomorrow.

TheaperDeng:
@hjzin any comments about this PR? I will merge it today if you are ok with it.

hjzin:
> @hjzin any comments about this PR? I will merge it today if you are ok with it.

nothing to update, you can merge it.

PR #5619, opened 2022-09-01
Nano: Add tensorflow quantization example.

mycprotein:
### 1. Why the change?

Add an example which shows the quantization of model during inference.


python/nano/tutorial/inference/tensorflow/mnist.py
```diff
@@ -0,0 +1,121 @@
...
+# pip install 'neural-compressor'
+
+# calib_dataset only accept tf.data.Dataset object
+tune_dataset = tf.data.Dataset.from_tensor_slices((x_train, y_train))
+
+# Execute quantization
+q_model = model.quantize(precision='int8',
+                         accelerator=None,
+                         calib_dataset=tune_dataset,
+                         accuracy_criterion={
```
TheaperDeng, line 106: we would like to present the easiest API calling to our users in fundamental example, so we may leave parameters after `accuracy_criterion` to be default values.

python/nano/tutorial/inference/tensorflow/mnist.py
```diff
@@ -0,0 +1,121 @@
...
+                         )
+
+y_test_hat = q_model(x_test)
+loss = float(tf.reduce_mean(
+             losses.categorical_crossentropy(y_test, y_test_hat)))
+categorical_accuracy = metrics.CategoricalAccuracy()
+categorical_accuracy.update_state(y_test, y_test_hat)
+accuracy = categorical_accuracy.result().numpy()
+print("Quantization test loss:", loss)
+print("Quantization test accuracy:", accuracy)
```
TheaperDeng, line 121: how is the accuracy drop?

python/nano/tutorial/inference/tensorflow/mnist.py
mycprotein: OK, I will try to simplify it.

python/nano/tutorial/inference/tensorflow/mnist.py
mycprotein: I will add it to a comment.

python/nano/tutorial/inference/tensorflow/mnist.py
```diff
@@ -0,0 +1,117 @@
...
+# Raw model test loss: 0.024767747148871422
+# Raw model test accuracy: 0.9918000102043152
+# Quantized model test loss: 0.02494174614548683
+# Quantized model test accuracy: 0.9917
+# Accuracy loss: 0.01%
```
TheaperDeng, line 117: that's reasonable accuracy drop. But we may just use some qualitative way to describe this in comment (e.g. accuracy drop is with on 0.1%). That's because the quantization may bring different result every time you run it.

python/nano/tutorial/inference/tensorflow/mnist.py
```diff
@@ -0,0 +1,117 @@
...
+
+# Quantization using Intel Neural Compressor
+# pip install 'neural-compressor'
+
+# calib_dataset only accept tf.data.Dataset object
+tune_dataset = tf.data.Dataset.from_tensor_slices((x_train, y_train))
+
+# Execute quantization
+q_model = model.quantize(calib_dataset=tune_dataset)
+
```
TheaperDeng, line 104: add comment "Inference using quantized model"

python/nano/tutorial/inference/tensorflow/mnist.py
```diff
@@ -0,0 +1,117 @@
...
+y_test_hat = q_model(x_test)
+loss = float(tf.reduce_mean(
```
TheaperDeng, line 106: add comment "evaluate the quantized model" and make some empty lines here

TheaperDeng (commented):
others LGTM

python/nano/tutorial/inference/tensorflow/mnist.py
```diff
@@ -0,0 +1,125 @@
...
+categorical_accuracy = metrics.CategoricalAccuracy()
+categorical_accuracy.update_state(y_test, y_test_hat)
+accuracy = categorical_accuracy.result().numpy()
+
+print("Quantization test loss:", loss)
+print("Quantization test accuracy:", accuracy)
+# Raw model test loss: 0.024767747148871422
+# Raw model test accuracy: 0.9918000102043152
+# Quantized model test loss: 0.02494174614548683
+# Quantized model test accuracy: 0.9917
```
TheaperDeng, line 122: remove these specific evaluation number

jason-dai:
MNIST is too trivial.

We should used some other (simple but more meaningful) model/data.

mycprotein:
I will change the dataset to CIFAR10, and use ResNet as model. Maybe they are more meaningful and modern.

mycprotein:
Dataset is CIFAR10 now, but the model is still a toy model. I have tried some relatively large models and mobilenet v2, they all take too long to train, so I think a toy model is better here.

python/nano/tutorial/inference/tensorflow/cifar10.py
```diff
@@ -0,0 +1,134 @@
...
+model.add(layers.Conv2D(128, (3,3), padding='same', activation='relu'))
+model.add(layers.BatchNormalization())
+model.add(layers.MaxPooling2D(pool_size=(2,2)))
+model.add(layers.Dropout(0.5))
+
+model.add(layers.Flatten())
+model.add(layers.Dense(128, activation='relu'))
+model.add(layers.BatchNormalization())
+model.add(layers.Dropout(0.5))
+model.add(layers.Dense(num_classes, activation='softmax'))
```
TheaperDeng, line 90: put these in a `create_model` function

python/nano/tutorial/inference/tensorflow/cifar10.py
```diff
@@ -0,0 +1,134 @@
...
+
+model.add(layers.Flatten())
+model.add(layers.Dense(128, activation='relu'))
+model.add(layers.BatchNormalization())
+model.add(layers.Dropout(0.5))
+model.add(layers.Dense(num_classes, activation='softmax'))  
+
+# this line is optional
+# model = Model(inputs=model.inputs, outputs=model.outputs)
+
```
TheaperDeng, line 94: put following part under `if __name__ == "__main__":`

python/nano/tutorial/inference/tensorflow/cifar10.py
```diff
@@ -0,0 +1,134 @@
...
+             losses.categorical_crossentropy(y_test, y_test_hat)))
+categorical_accuracy = metrics.CategoricalAccuracy()
+categorical_accuracy.update_state(y_test, y_test_hat)
+accuracy = categorical_accuracy.result().numpy()
+
+print("Quantization test loss:", loss)
+print("Quantization test accuracy:", accuracy)
+# Note: accuracy decrease varies from different tasks and situations,
+# and sometimes the accuracy is even higher.
+# You can set a quantization threshold when making a quantization model.
```
TheaperDeng, line 134: how long does this example run?

mycprotein: About half an hour.

TheaperDeng: a little bit long for a fundamental example yah? 

@MeouSker77 how much time does it take to run through https://github.com/intel-analytics/BigDL/blob/main/python/nano/tutorial/training/tensorflow/tensorflow_train_multi_instance.py

I wonder if we can use this model and data in this example.

No comment on other parts of the code

MeouSker77: > @MeouSker77 how much time does it take to run through https://github.com/intel-analytics/BigDL/blob/main/python/nano/tutorial/training/tensorflow/tensorflow_train_multi_instance.py

about 7 minutes

MeouSker77: > I wonder if we can use this model and data in this example.

I remember using the model and data in that example may encounter the conflict of tcmalloc and tf quantization

MeouSker77: > > @MeouSker77 how much time does it take to run through https://github.com/intel-analytics/BigDL/blob/main/python/nano/tutorial/training/tensorflow/tensorflow_train_multi_instance.py
> 
> about 7 minutes

I uses `num_epochs = int(os.environ.get('NUM_EPOCHS', 10))` to get the epoch num, then `export NUM_EPOCHS=1` in bash script before running that example to accelerate it.

mycprotein: In my machine, this example takes about an hour to finish. (10 epochs)
The cifar10 example takes 1.5 minutes for an epoch, and half an hour in total. (20 epochs)

python/nano/tutorial/inference/tensorflow/cifar10.py
```diff
@@ -0,0 +1,142 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This file is adapted from Keras Tutorial.
+# https://github.com/keras-team/keras-io/blob/
+# master/examples/vision/mnist_convnet.py
```
TheaperDeng, line 18: remove this

PR #5620, opened 2022-09-01
Change the environment of container and fix the problem of cannot compare two variables.

Hanyu-Jin:
## Description

  First, fix the environment settings to run FL Server.
  Second, when run FL Server in SGX, it cannot compare two variables with different types. Fix this problem by cast string to int.

### 1. Why the change?

  FL needs protobuf==3.19.0 and torch==1.5.0.
  Code running in SGX seems not allowed to cast automatically while compare two variables.

### 2. User API changes

  None.

### 3. Summary of the change 

  Set Dockerfile to build docker image with protobuf==3.19.0 and torch==1.5.0.
  Cast string to int in nn_service and psi_intersection.

### 4. How to test?
  None

### 5. New dependencies

- [x] New Python dependencies
       - protobuf==3.19.0 
       - torch==1.5.0
       - ...


Litchilitchy (approved):

glorysdj (approved):
LGTM

PR #5622, opened 2022-09-01
Chronos: modify examples of forecaster.evaluate() and Evaluator.evaluate()

plusbang:
## Description

In Chronos API doc, examples of forecaster.evaluate() and forecaster.evaluate_with_onnx() are incorrect. And example of Evaluator.evaluate() is added.


TheaperDeng (approved):
LGTM

PR #5635, opened 2022-09-05
Add nano torch support for channels last

y199387:
## Description

Add support for automatically converting model and data's memory_format to channels_last.

### 1. Why the change?

Channels Last improves the throughput of convolution operations in networks for computer vision.  It's
necessary to provide user a simple way to apply torch.channels to both of the model object and data.

### 2. User API changes
```python
MyNano(TorchNano):
    def train(self):
        ...

MyNano(channels_last=True).train()
```

### 3. Summary of the change
N/A

### 4. How to test?
- [x] Unit test
- [x] Jenkins
http://10.112.231.51:18889/job/BigDL-Chronos-PR-Validation/801/

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/torch_nano.py
+++ b/python/nano/src/bigdl/nano/pytorch/torch_nano.py
@@ -162,13 +162,13 @@ class TorchNano(LightningLite):
 
     def _setup(
         self,
-        model: nn.Module,  # type: ignore[override]
+        model: nn.Module,
         optimizers: List[Optimizer],
         move_to_device: bool = True,
     ) -> Any:
         """Used to replace LightningLite's setup method."""
         if self.channels_last:
-            model = model.to(memory_format=torch.channels_last)
+            model = model.to(memory_format=torch.channels_last)  # type: ignore
         # LightningLite won't call `Strategy.setup()` method,
         # in which we add IPEX's optimization when using `trainer`.
 
--- a/python/nano/src/bigdl/nano/pytorch/torch_nano.py
+++ b/python/nano/src/bigdl/nano/pytorch/torch_nano.py
@@ -75,7 +75,7 @@ class _TorchNanoModule(_LiteModule):
             return getattr(self.module, name)
 
     def forward(self, *args: Any, **kwargs: Any) -> Any:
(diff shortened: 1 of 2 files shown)


python/nano/src/bigdl/nano/pytorch/torch_nano.py
```diff
@@ -35,13 +36,28 @@
...
+    def forward(self, *args: Any, **kwargs: Any) -> Any:
+        """Casts all inputs to the right memory format."""
+        if self.channels_last:
+            def _convert_to_channels_last(t: torch.Tensor) -> torch.Tensor:
+                if t.dim() == 4:
+                    return t.to(memory_format=torch.channels_last)  # type: ignore
+                return t
+            args, kwargs = apply_to_collection([args, kwargs], function=_convert_to_channels_last,
+                                               dtype=torch.Tensor)
+        return super().forward(*args, **kwargs)
```
TheaperDeng, line 59: I could not remember if this(i.e. `forward(*args, **kwargs)`) may cause a problem for quantization/trace, could please have a quick check?

The structure for test code could be
```python
# define a nanotorch model
t_model = Trainer.trace(model, ...)
t_model(x)
q_model = Trainer.quantize(model, ...)
q_model(x)
```

TheaperDeng (commented):
others LGTM, @MeouSker77 please also have a look

python/nano/src/bigdl/nano/pytorch/torch_nano.py
MeouSker77: > I could not remember if this(i.e. `forward(*args, **kwargs)`) may cause a problem for quantization/trace, could please have a quick check?

yes, it reported an error in my test, but will we use `Trainer` and `_TorchNanoModule` together ?

TheaperDeng: Yeah I think. if a user have a pytorch nnmodule and he/she change it to a torchnano model for training acceleration, then, he/she wants to accelerate the model for inference.

y199387: Should we add the example in this PR or a follow up PR?

PR #5642, opened 2022-09-06
PPML FL api Python estimator doc

Litchilitchy:
## Description

Add FL  API doc, continuing updating.

add API doc using the style of [Orca estimator](https://github.com/intel-analytics/BigDL/blob/main/python/orca/src/bigdl/orca/tfpark/estimator.py)

for all the API level methods, including following @jason-dai please directly edit the list if anything else

- [ ] [NN estimator](https://github.com/intel-analytics/BigDL/blob/main/python/ppml/src/bigdl/ppml/fl/estimator.py)
- [ ] [PSI](https://github.com/intel-analytics/BigDL/blob/main/python/ppml/src/bigdl/ppml/fl/psi/psi_client.py)
- [ ] [FGBoost regression](https://github.com/intel-analytics/BigDL/blob/main/python/ppml/src/bigdl/ppml/fl/algorithms/fgboost_regression.py)


python/ppml/src/bigdl/ppml/fl/nn/fl_context.py
```diff
@@ -17,6 +17,11 @@
 from ..nn.fl_client import FLClient
 
 def init_fl_context(client_id, target="localhost:8980"):
+    """Initialize FL Context. Need to be called before calling any FL Client algorithms.
+    
+    :param client_id: A number, should be in range of [1, total_party_number].
+    :param target: FL Server address.
```
jason-dai, line 23: then why call it target? just `server_addr`

python/ppml/src/bigdl/ppml/fl/nn/fl_context.py
```diff
@@ -17,6 +17,11 @@
 from ..nn.fl_client import FLClient
 
 def init_fl_context(client_id, target="localhost:8980"):
+    """Initialize FL Context. Need to be called before calling any FL Client algorithms.
+    
+    :param client_id: A number, should be in range of [1, total_party_number].
```
jason-dai, line 22: an integer

PR #5651, opened 2022-09-06
support init_orca_context with spark.master=local[*, F]/local[N, F]

PatrickkZ:
## Description

Currently, BigDL can't parse format like spark.master=local[\*, F]/local[N, F], so when call `init_orca_context` with conf spark.master=local[\*, F]/local[N, F] will fail. this PR add support for this format.

### DIFF
--- a/scala/dllib/src/main/scala/com/intel/analytics/bigdl/dllib/utils/Engine.scala
+++ b/scala/dllib/src/main/scala/com/intel/analytics/bigdl/dllib/utils/Engine.scala
@@ -531,9 +531,13 @@ object Engine {
     val master = conf.get("spark.master", null)
     if (master.toLowerCase.startsWith("local")) {
       // Spark local mode
+      // patternLocalN example: local[4]
       val patternLocalN = "local\\[(\\d+)\\]".r
+      // patternLocalNF example: local[4,2]
       val patternLocalNF = "local\\[(\\d+),\\s*(\\d+)\\]".r
+      // patternLocalStar example: local[*]
       val patternLocalStar = "local\\[\\*\\]".r
+      // patternLocalStarF example: local[*,4]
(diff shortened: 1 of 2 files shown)


scala/dllib/src/main/scala/com/intel/analytics/bigdl/dllib/utils/Engine.scala
```diff
@@ -532,10 +532,14 @@ object Engine {
     if (master.toLowerCase.startsWith("local")) {
       // Spark local mode
       val patternLocalN = "local\\[(\\d+)\\]".r
+      val patternLocalNF = "local\\[(\\d+),\\s*(\\d+)\\]".r
       val patternLocalStar = "local\\[\\*\\]".r
+      val patternLocalStarF = "local\\[\\*,\\s*(\\d+)\\]".r
```
hkvision, line 537: Add some comments for each case.
e.g. local[*, 4]

scala/dllib/src/main/scala/com/intel/analytics/bigdl/dllib/utils/Engine.scala
```diff
@@ -532,10 +532,14 @@ object Engine {
...
       master match {
         case patternLocalN(n) => Some(1, n.toInt)
+        case patternLocalNF(n, f) => Some(1, n.toInt)
         case patternLocalStar(_*) => Some(1, getNumMachineCores)
+        case patternLocalStarF(_*) => Some(1, getNumMachineCores)
         case _ =>
           Log4Error.invalidOperationError(false, s"Can't parser master $master")
```
hkvision, line 544: parser -> parse

dding3:
Can we have some unit tests for this pr?

hkvision (approved):

PR #5656, opened 2022-09-07
Chronos:  Make `TSDataset` more friendly

liangs6212:
## Description

Before that, since the horizon is not 0, the length of the predicted data will be less than the actual demand data.
Now if `is_predict` set to True, then horizon will be set to 0, which avoids the prediction data being split.

### Usage
```python

test = TSDataset.from_pandas(...)
test.is_predict = True
yhat = forecaster.predict(test)
```

### DIFF
--- a/python/chronos/src/bigdl/chronos/data/tsdataset.py
+++ b/python/chronos/src/bigdl/chronos/data/tsdataset.py
@@ -28,7 +28,8 @@ from bigdl.chronos.data.utils.resample import resample_timeseries_dataframe
 from bigdl.chronos.data.utils.split import split_timeseries_dataframe
 from bigdl.chronos.data.utils.cycle_detection import cycle_length_est
 from bigdl.chronos.data.utils.utils import _to_list, _check_type,\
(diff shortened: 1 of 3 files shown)


TheaperDeng:
Should we use a setter for this member? since we want to use it as a external API

liangs6212:
> Should we use a setter for this member? since we want to use it as a external API

Like this?
```python
yhat = forecaster.predict(test, is_predict=True)
```

It looks a little weird, but it's friendly to use.

TheaperDeng:
> > Should we use a setter for this member? since we want to use it as a external API
> 
> Like this?
> 
> ```python
> yhat = forecaster.predict(test, is_predict=True)
> ```
> 
> It looks a little weird, but it's friendly to use.

no I mean make this member a https://docs.python.org/3/library/functions.html#property .

We may try some thing new in this PR, since we might have such syntax supported more and more in future.

liangs6212:
Diff:
 1. Removed redundant parameters(is_predict)
 2. Some changes based on property(experiment)
 3. Descriptor applied(experiment)

PR #5664, opened 2022-09-07
[PPML]correct the command and add explanation in kms-utils README

ShanSimu:
## Description

Correct the parameters of commands and explain file structure in kms-utils README.


Uxito-Ada (approved):
LGTM

PR #5667, opened 2022-09-07
Nano : migrate trace and quantize from Trainer to InferenceOptimizer

rnwang04:
## Description

move the implementation of Trainer.trace and Trainer.quantize to InferenceOptimizer, and have Trainer to call InferenceOptimizer

### 1. Why the change?

All subsequent code related to inference optimizations will be put into InferenceOptimizer, this is the first step. Later, we will delete the trace and quantify interfaces in Trainer.

### 2. User API changes

No change.

### 3. Summary of the change 

move the implementation of Trainer.trace and Trainer.quantize to InferenceOptimizer, and have Trainer to call InferenceOptimizer

### 4. How to test?
- [x] Unit test
- [x] Jekins test
http://10.112.231.51:18889/view/BigDL-PR-Validation/job/ZOO-PR-NanoTests/390/

### DIFF
--- a/python/nano/example/pytorch/semantic_segmentation/semantic_segmentation.py
+++ b/python/nano/example/pytorch/semantic_segmentation/semantic_segmentation.py
@@ -303,7 +303,7 @@ def main(hparams: Namespace):
     # ------------------------
     # 2 INIT TRAINER
     # ------------------------
-    trainer = Trainer.from_argparse_args(hparams)
+    trainer = Trainer.from_argparse_args(hparams)  # type: ignore
 
     # ------------------------
     # 3 START TRAINING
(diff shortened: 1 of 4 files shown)


jason-dai:
Also need to deprecate old APIs in `Trainer`

rnwang04:
> Also need to deprecate old APIs in `Trainer`

Shall we deprecate old APIs currently or after 2.1 release?

jason-dai:
> > Also need to deprecate old APIs in `Trainer`
> 
> Shall we deprecate old APIs currently or after 2.1 release?

We should deprecate old APIs now

TheaperDeng:
https://github.com/intel-analytics/BigDL/blob/main/python/chronos/src/bigdl/chronos/utils.py#L23-L34 might help

PR #5673, opened 2022-09-07
Chronos: fix input feature num != output feature num case for autoformer

TheaperDeng:
## Description
Fix autoformer to support input feature num != output feature num case.

### 4. How to test?
Unit test to be added
- [ ] Unit test


rnwang04:
LGTM

Issue #4642, opened 2022-05-20
Spark backend does not work on Databricks


HaraldVanWoerkom:
When I create an estimator with a 'spark' backend on Databricks, I have to provide a path to a shared folder. On Databricks that is on /dbfs/...:
est = Estimator.from_keras(model_creator=my_model_creator, backend='spark', model_dir = '/dbfs/tmp/bigdl')

Unfortunately, the fit crashes with an "operation not supported" exception. This is (apparently) because the model writing code uses random access, which is not supported by the file system.
BigDL has a solution for this, it can write to local storage and then copy to the shared folder. This is triggered by setting model_dir to 'dbfs:/tmp/bigdl'. Unfortunately, this causes an exception in bigdl/orca/learn/utils.py, the save_pkl function, because 'open' does not support the dbfs: format (yes, in Databricks some APIs require dbfs:, while others do not support it).

I patched utils.py to support dbfs (line 380):
    else:
        if path.startswith("file://"):
            path = path[len("file://"):]
        elif path.startswith("dbfs:/"):  # NEW
            path = "/dbfs/" + path[len("dbfs:/"):] # NEW
        with open(path, 'wb') as f:
            pickle.dump(data, f)

This seems to work.

jason-dai:
@HaraldVanWoerkom Thanks for reporting the issue; we'll take a look.

Le-Zheng:
@HaraldVanWoerkom many thanks for your posting!  I have verified the solution `path = "/dbfs/" + path[len("dbfs:/"):] ` works on the Databricks. I will create a fix in the source code.

HaraldVanWoerkom:
@Le-Zheng Thanks for picking this up.

Le-Zheng:
> @Le-Zheng Thanks for picking this up.

sure. `Estimator.from_keras API` does not call `save_pkl` function.

Le-Zheng:
This issue will be fixed in the patch https://github.com/intel-analytics/BigDL/pull/4674

PR #5679, opened 2022-09-08
fix issue 4642, fix DBFS file path problem on Dataricks

PatrickkZ:
## Description

when use the following code to create an estimator on Databricks
```python
est = Estimator.from_keras(model_creator=my_model_creator, backend='spark', model_dir = '/dbfs/model')
```
the `model_dir` should be a shared path to save model, which is a path starts with `/dbfs` or `dbfs:/` on Databricks. but currently, no matter which format(`/dbfs` or `dbfs:/`) you use, the process of saving model won't succeed because of the incorrect save path.

the error comes from [save_model](https://github.com/PatrickkZ/BigDL/blob/main/python/orca/src/bigdl/orca/learn/tf2/spark_runner.py#L345) and [save_pkl](https://github.com/PatrickkZ/BigDL/blob/main/python/orca/src/bigdl/orca/learn/tf2/spark_runner.py#L350). 

when use `/dbfs` format, save_model method will consider the path as a local path, trying to write to this path will cause a  "operation not supported" exception, because model save use random access, which is not supported by the file system. the way to fix this problem is save the model to a local path and then copy it to the shared path(`/dbfs`).

when use `dbfs:/` format, save_pkl will fail because it calls open(), which don't support `dbfs:/` format. the way to fix this problem is if the input model_dir starts with `dbfs:/`, then change it to `/dbfs/`

### 1. Why the change?

fix related issue: https://github.com/intel-analytics/BigDL/issues/4642
### 2. User API changes

the model_dir parameter on Databrick now support two formats, `/dbfs` and `dbfs:/`
### 3. Summary of the change 

fix the model save problem on Databricks.
### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [x] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/python/dllib/src/bigdl/dllib/utils/file_utils.py
+++ b/python/dllib/src/bigdl/dllib/utils/file_utils.py
@@ -70,6 +70,8 @@ def mkdirs(path):
 
 
 def is_local_path(path):
+    if path.startswith("/dbfs"):
+        return False
     parse_result = urlparse(path)
     return len(parse_result.scheme.lower()) == 0 or parse_result.scheme.lower() == "file"
 
--- a/python/orca/src/bigdl/orca/learn/tf2/estimator.py
+++ b/python/orca/src/bigdl/orca/learn/tf2/estimator.py
@@ -73,6 +73,8 @@ class Estimator(object):
                 invalidInputError(False,
                                   "cpu_binding should not be True when using spark backend")
             from bigdl.orca.learn.tf2.pyspark_estimator import SparkTFEstimator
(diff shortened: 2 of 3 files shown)


hkvision (approved):

hkvision:
Merge it first. Will verify it on databricks tomorrow using tonight's build.

PatrickkZ:
Verified on Databricks. format `/dbfs` and `dbfs:` both work.

hkvision:
> Verified on Databricks. format `/dbfs` and `dbfs:/` both work.

Great!

PR #5680, opened 2022-09-08
Add more data types in convert DataFrame to numpy

PatrickkZ:
## Description

support more data types when convert DataFrame element type to numpy type


hkvision:
http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-Orca-PR-Validation/1041/

PR #5684, opened 2022-09-08
Chronos: forecaster save/load how-to guide

smurf-1119:
## Description

forecaster user guide

how to save and load forecaster


TheaperDeng:
we need more words to make sure our users to understand how to save/load. But I think the save/load might be refactored recently.

rnwang04:
doc test : https://ruonantetdoc.readthedocs.io/en/forecaster_save_doc/doc/Chronos/Howto/index.html

TheaperDeng:
If you are in a new session, then you should define a forecaster first, then load it **ny** -> **by** filename.

others LGTM

PR #5687, opened 2022-09-08
Support load state dict in TorchNano

MeouSker77:
### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

### 5. New dependencies

- [ ] New Python dependencies
       - Dependency1 
       - Dependency2
       - ...
       - Dependency1 and license1
       - Dependency2 and license2
       - ...

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/torch_nano.py
+++ b/python/nano/src/bigdl/nano/pytorch/torch_nano.py
@@ -14,13 +14,15 @@
 # limitations under the License.
 #
 
-from typing import Any, Union, List
+import copy
+from typing import Any, Union, List, Dict
 from logging import warning
 from functools import partial
 from abc import abstractmethod
+from collections import OrderedDict
 
 import torch
-from torch import nn
+from torch import nn, Tensor
 from torch.optim import Optimizer
 from torch.utils.data import DataLoader
 from pytorch_lightning.lite import LightningLite
@@ -32,6 +34,44 @@ from bigdl.nano.pytorch.utils import TORCH_VERSION_LESS_1_10, TORCH_VERSION_LESS
 from bigdl.nano.pytorch.strategies.ipex.ipex_api import ipex_optimize
 from bigdl.nano.pytorch.strategies import create_IPEXStrategy, DDPSpawnStrategy, \
     DDPSubprocessStrategy, create_RayStrategy
+
+
+class _TorchNanoModuleWrapper():
+    def __init__(self, model: "_TorchNanoModule", torch_nano: "TorchNano", backup_model: nn.Module,
+                 backup_optimizers: List[Optimizer], move_to_device: bool):
+        self.model = model
+        self.torch_nano = torch_nano
+        self.backup_model = backup_model
+        self.backup_optimizers = backup_optimizers
+        self.move_to_device = move_to_device
+
+    def state_dict(self, *args, **kwargs):
+        return self.model._module.state_dict(*args, **kwargs)
+
+    def load_state_dict(self, state_dict: 'OrderedDict[str, Tensor]', strict: bool = True,
+                        optimizer_state_dict: 'Union[Dict, List[Dict]]' = None):
+        # load model and optimizer's state dict
+        self.backup_model.load_state_dict(state_dict, strict)
+        if optimizer_state_dict is not None:
+            optimizer_state_dict_list = optimizer_state_dict \
+                if isinstance(optimizer_state_dict, list) else [optimizer_state_dict]
+            for optimizer, state in zip(self.backup_optimizers, optimizer_state_dict_list):
+                optimizer.load_state_dict(state)
+
+        # re-setup loaded model and optimizer
+        model, optimizers = self.torch_nano._setup(self.backup_model, self.backup_optimizers,
+                                                   self.move_to_device)
+        self.model = model
+
+        if optimizer_state_dict is not None:
+            optimizer = optimizers if isinstance(optimizer_state_dict, list) else optimizers[0]
+            return optimizer
+
+    def __call__(self, *args, **kwargs):
+        return self.model(*args, **kwargs)
+
+    def __getattr__(self, name: str):
+        return getattr(self.model, name)
 
 
 class _TorchNanoModule(_LiteModule):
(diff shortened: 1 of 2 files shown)


python/nano/src/bigdl/nano/pytorch/torch_nano.py
```diff
@@ -180,12 +243,21 @@ def setup(self, model: nn.Module,    # type: ignore[override]
...
-        model, optimizers = self._setup(model, optimizers, move_to_device=move_to_device)
+        if self.use_ipex and not TORCH_VERSION_LESS_1_10:
+            (backup_model, backup_optimizers) = copy.deepcopy((model, optimizers))
+            model, optimizers = self._setup(model, optimizers, move_to_device=move_to_device)
+            model_wrapper = _TorchNanoModuleWrapper(model, self, backup_model,  # type: ignore
+                                                    backup_optimizers, move_to_device)
+        else:
+            model, optimizers = self._setup(model, optimizers, move_to_device=move_to_device)
+            model_wrapper = _TorchNanoModuleWrapper(model, self, model,  # type: ignore
+                                                    optimizers, move_to_device)
```
jason-dai, line 254: 1) Why not handle this in `_setup` instead?
2) I think you can extend `_TorchNanoModule` instead of introducing a new class `_TorchNanoModuleWrapper`? `_TorchNanoModule` is already a wrapper class and also deals with `__getattr__`.

jason-dai:
> ### 2. User API changes
> user can wirte the following code now:
> 
> ```python
> class MyNano(TorchNano):
>     def train(self):
>         ...
>         model, optimizer = self.setup(model, optimizer)
>         ...
>         model_state_dict = model.state_dict()
>         optimizer_state_dict = optimizer.state_dict()
>         ...
>         # no more training
>         model.load_state_dict(model_state_dict)
>         # still want to continue training
>         optimizer = model.load_state_dict(model_state_dict, optimizer_state_dict=optimizer_state_dict)
>         ...
> ```
> 
> * note: if user want to continue training after calling `load_state_dict`, he must pass the optimizer's state dict together with the model's state dict, and use the returned optimizer for the following training (the returned optimizer is same as the optimizer passed to `self.setup` method)

I don't think this API aligns with standard PyTorch interfaces for saving & loading the model (https://pytorch.org/tutorials/beginner/saving_loading_models.html)

python/nano/src/bigdl/nano/pytorch/torch_nano.py
```diff
@@ -34,6 +36,63 @@
...
+        :return: return optimizers if passed `optimizer_state_dict`, otherwise return nothing
+        """
+        # load optimizer's state dict using backup optimizers
+        if optimizer_state_dict is not None:
+            optimizer_state_dict_list = optimizer_state_dict \
+                if isinstance(optimizer_state_dict, list) else [optimizer_state_dict]
+            for optimizer, state in zip(self.backup_optimizers, optimizer_state_dict_list):
+                optimizer.load_state_dict(state)
+
+        if self.torch_nano.use_ipex and not TORCH_VERSION_LESS_1_10:
```
jason-dai, line 72: use a flag such as `has_backup`

python/nano/src/bigdl/nano/pytorch/torch_nano.py
MeouSker77: if we extend `_TorchNanoModule` instead of using `_TorchNanoModuleWrapper`, then the `load_state_dict` method will be:
```python
class TorchNanoModule(_LiteModule):
    def load_state_dict(self, state_dict, xxx):
        backup_model.load_state_dict(state_dict)
        new_model = xxx._setup(backup_model)
        self = new_model
```

however, `self = new_model` is invalid in python, we cann't replace `self` in the `load_state_dict` method

MeouSker77:
> I don't think this API aligns with standard PyTorch interfaces for saving & loading the model (https://pytorch.org/tutorials/beginner/saving_loading_models.html)

yes, it doesn't align with standard PyTorch, but we may have no other way.

If we replaced a model, then we have to replace the optimizer too. Standard Pytorch model loads the state dict in place, so it doesn't need a new optimizer. But we replaced the old model, and the old optimizer is stilled bound to the old model, so we need a new optimizer, which is bound to the new model.

python/nano/src/bigdl/nano/pytorch/torch_nano.py
jason-dai: > if we extend `_TorchNanoModule` instead of using `_TorchNanoModuleWrapper`, then the `load_state_dict` method will be:
> 
> ```python
> class TorchNanoModule(_LiteModule):
>     def load_state_dict(self, state_dict, ...):
>         backup_model.load_state_dict(state_dict)
>         new_model = xxx._setup(backup_model)
>         self = new_model
> ```
> 
> however, `self = new_model` is invalid in python, we can't replace `self` in the `load_state_dict` method

We can have `_setup`  to return the normal model (possibly optimized by ipex), and wrap it with `_TorchNanoModule` in `setup`; then we can have something like:

```python
class TorchNano(LightningLite):
  def setup(...)
        ...
        model, optimizers = self._setup(model, optimizers, ...)
        model = _TorchNanoModule(model, self._precision_plugin)
        ...

class TorchNanoModule(_LiteModule):
  def load_state_dict(self, state_dict, ...):
        backup.load_state_dict(state_dict)
        module = xxx._setup(backup_model)
        ...
```

jason-dai:
> > I don't think this API aligns with standard PyTorch interfaces for saving & loading the model (https://pytorch.org/tutorials/beginner/saving_loading_models.html)
> 
> yes, it doesn't align with standard PyTorch, but we may have no other way.
> 
> If we replaced a model, then we have to replace the optimizer too. Standard Pytorch model loads the state dict in place, so it doesn't need a new optimizer. But we replaced the old model, and the old optimizer is stilled bound to the old model, so we need a new optimizer, which is bound to the new model.



> > I don't think this API aligns with standard PyTorch interfaces for saving & loading the model (https://pytorch.org/tutorials/beginner/saving_loading_models.html)
> 
> yes, it doesn't align with standard PyTorch, but we may have no other way.
> 
> If we replaced a model, then we have to replace the optimizer too. Standard Pytorch model loads the state dict in place, so it doesn't need a new optimizer. But we replaced the old model, and the old optimizer is stilled bound to the old model, so we need a new optimizer, which is bound to the new model.

What type of code pattern do we want to support? Maybe we should save the non-ipex version of the model, and do not allow ipex-optimized model to load; and we will always load non-ipex version of the model, and need to call `setup` again for retraining?

```python
torch.save({
            'model_state_dict': model.state_dict(),
            'optimizer_state_dict': optimizer.state_dict(),
             ...
            }, PATH)
```
```python
model = TheModelClass(...)
optimizer = TheOptimizerClass(...)

checkpoint = torch.load(PATH)
model.load_state_dict(checkpoint['model_state_dict'])
optimizer.load_state_dict(checkpoint['optimizer_state_dict'])
...

model, optimizer = self.setup(model, optimizer, ...)
```

TheaperDeng (approved):
lgtm, we definitly need a user guide for this "trick"/"workaround" about state dict

jason-dai:
> lgtm, we definitly need a user guide for this "trick"/"workaround" about state dict

Need to add basic examples/how-to on model save & load (for both training and inference).

PR #5692, opened 2022-09-08
Fix Orca Example styles

hkvision:


hkvision:
http://10.112.231.51:18888/job/BigDL-Orca-PR-Validation/1032/ some failures are irrelevant

PR #5698, opened 2022-09-09
cicd: scan dependency tree to help debug if fossa scan failed

liu-shaojun:
## Description

This submission
- add a step in license-scan workflow to filter the dependency tree, this step will be executed only if the previous fossa scan and fossa test failed. This is to help debug when license scan failed.


glorysdj (approved):
LGTM

PR #5701, opened 2022-09-09
Add selfHostedRunner input

Romanticoseu:
## Description

When use self-hosted-runner , proxy must be set. But in github-runner, set proxy will cause network erros. So add a input called "selfHostedRunner" , and default value is false.


glorysdj (approved):
LGTM

PR #5703, opened 2022-09-09
bigdl all in one jar

Le-Zheng:
## Description

Build one jar `bigdl-assembly-spark_2.4.6-2.1.0-SNAPSHOT-jar-with-dependencies.jar`
1. change bigdl-assembly packaging as jar
2. add plugin to append all dependencies and build one jar
3. remove single *jar-with-dependency in fat-assembly, and add all in one jar in fat-assembly
image

### DIFF
--- a/scala/assembly/pom.xml
+++ b/scala/assembly/pom.xml
@@ -151,42 +151,10 @@
                 </executions>
 	    </plugin>
             <plugin>
-                <!--Please don't merge this shade plugin with spark-version's shade plugin to spark/pom.xml,
-                 or shade plugin will be executed after assembly plugin. -->
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-shade-plugin</artifactId>
                 <version>3.2.1</version>
-                <configuration>
-                    <filters>
-                        <filter>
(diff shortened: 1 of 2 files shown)


Le-Zheng:
It looks the size is equal
image

hkvision:
Will there be any place we can download this jar only instead of the fat zip? (e.g. for databricks we only need the jar)

glorysdj (approved):
LGTM

Le-Zheng:
> Will there be any place we can download this jar only instead of the fat zip? (e.g. for databricks we only need the jar)

Yes, it will have an individual link for this jar.

Le-Zheng:
> > Will there be any place we can download this jar only instead of the fat zip? (e.g. for databricks we only need the jar)
> 
> Yes, it will have an individual link for this jar.

jar link: https://oss.sonatype.org/content/repositories/snapshots/com/intel/analytics/bigdl/bigdl-assembly-spark_3.1.2/2.1.0-SNAPSHOT/bigdl-assembly-spark_3.1.2-2.1.0-20220912.135611-231-jar-with-dependencies.jar

Issue #5608, opened 2022-08-31
Nano: trace to openvino inference engine need a thread(core) num control option


TheaperDeng:
For onnxruntime, users could control how many cores their accelerated model could use, while we don't have such option for openvino right now.

rnwang04:
Currently, no matter what the number of intra_op_num_threads and inter_op_num_threads is, only one core is used. 
And more threads, inference is slower.

jason-dai:
> Currently, no matter what the number of intra_op_num_threads and inter_op_num_threads is, only one core is used. And more threads, inference is slower.
> 
> ![image](https://user-images.githubusercontent.com/105281011/187829459-91776040-e5e7-4d2f-b86d-71147fd31f7d.png) ![image](https://user-images.githubusercontent.com/105281011/187829470-1a9aa5b7-339e-4a29-8efd-9414d4bac584.png)

@qiyuangong Is this behavior expected?

TheaperDeng:
> > Currently, no matter what the number of intra_op_num_threads and inter_op_num_threads is, only one core is used. And more threads, inference is slower.
> > ![image](https://user-images.githubusercontent.com/105281011/187829459-91776040-e5e7-4d2f-b86d-71147fd31f7d.png) ![image](https://user-images.githubusercontent.com/105281011/187829470-1a9aa5b7-339e-4a29-8efd-9414d4bac584.png)
> 
> @qiyuangong Is this behavior expected?

After huge amount of experiment and feedback from team members and customers I found

**ONNXRuntime's core num control is completely fragile and useless, it may not respect user's input randomly**

For fp32:
- some hardware (e.g. i7-9700), the control is stable and follows the setting
- some hardware (e.g. cpx), the control is useless, and the core usage might change randomly even on same SKU.

For int8
- some hardware (e.g. i7-9700), no matter how many cores you set, it will only use 1 core
- some hardware (e.g. icx), it will use 16 cores if you set < 16 cores, and respect your input if you set > 16 cores

We are contacting onnxruntime support team (MS and Intel) to find some solution.

rnwang04:
After `unset KMP_AFFINITY`, onnxruntimie works fine.

PR #5705, opened 2022-09-09
Nano : Enhancement for output format of InferenceOptimizer

rnwang04:
## Description

To provide our user with better experience when using InferenceOptimizer, this PR will make the output of `optimize` more clear and detailed.
Also fix https://github.com/intel-analytics/BigDL/issues/5608

### 1. Why the change?

To make the output of `optimize` more clear and detailed.

### 2. User API changes

New API : `summary()` to print format string representation for optimization result
```
optimizer = InferenceOptimizer()
optimizer.optimize(model=model, 
                       training_data=train_loader,
                       validation_data=test_loader, 
                       metric=torchmetrics.Accuracy(num_classes=10, top_k=1),
                       direction="max",
                       thread_num=1)
optimizer.summary()
```

### 3. Summary of the change 

- make the output of `optimize` more clear and detailed.
- add `summary` API
- support thread num control in openvino

### 4. How to test?
- [x] Unit test
- [x] Application test
http://10.112.231.51:18889/job/BigDL-Chronos-PR-Validation/833/
http://10.112.231.51:18889/job/BigDL-PRVN-chronos-Python-Spark-3.1-py37-Horovod/888/
http://10.112.231.51:18889/job/BigDL-PRVN-chronos-Python-Spark-2.4-py37-ray-part1/869/

### demo output
image
image
image

### DIFF
--- a/python/nano/example/pytorch/inference_pipeline/resnet/inference_pipeline.py
+++ b/python/nano/example/pytorch/inference_pipeline/resnet/inference_pipeline.py
@@ -45,7 +45,7 @@ if __name__ == "__main__":
                        validation_data=datamodule.val_dataloader(limit_num_samples=160),
                        metric=accuracy,
                        direction="max",
-                       cpu_num=1,
+                       thread_num=1,
                        latency_sample_num=30)
 
     # 4. Get the best model under specific restrictions or without restrictions
--- a/python/nano/src/bigdl/nano/deps/openvino/core/model.py
+++ b/python/nano/src/bigdl/nano/deps/openvino/core/model.py
@@ -22,9 +22,10 @@ from .utils import save
 
 
 class OpenVINOModel:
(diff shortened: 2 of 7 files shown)


jason-dai:
1) Use a different character than "*" (which is used by the error) to mark the start and ending of each experiment
2) Showing the progress of the experiments using, say, "Start testing ipex_fp32 model (2/13)"
3) De we need to show "trace" or "quantize" to the user?

rnwang04:
> 1. Use a different character than "*" (which is used by the error) to mark the start and ending of each experiment
> 2. Showing the progress of the experiments using, say, "Start testing ipex_fp32 model (2/13)"
> 3. De we need to show "trace" or "quantize" to the user?

Thanks for these suggestions!
1. replace  "*" by "-"
2. add a progress bar at the start and end of an experiment
3. remove the "method type" column.

PR #5712, opened 2022-09-10
support param map for lightGBM

songhappy:
## Description

support to build LightGBMClassier and LightGMRegressor with param map

### 1. Why the change?
follow BigDL XGBClassifier to handle parameters setting in an easier way. 
class XGBClassifier (val xgboostParams: Map[String, Any] = Map()) 
class LightGBMClassifier (val lgbmParams: Map[String, Any] = Map()) 
class LightGBMRegressor (val lgbmParams: Map[String, Any] = Map())

### 2. User API changes
old way: 
new LightGBMClassifier()
new LightGBMRegressor()
new way:
new LightGBMClassifier()
new LightGBMRegressor()
new LightGBMClassifier(params)
new LightGBMRegressor(params)

### 3. Summary of the change 
1. LightGBMClassifier and LightGBMRegressor can take param map like XGBClassier
2. parameter names follow the names of the original lightGBM, with python style to scala style.
https://github.com/Microsoft/LightGBM/blob/master/docs/Parameters.rst#boosting
3. Current parameters from the original LightGBM constructor and tuning guide are supported except the ones that are not supported by synapseML. random_state, importance_type, max_bin_by_feature, max_cat_threshold, extra_trees, path_smooth are not supported here. 
https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.LGBMClassifier.html#lightgbm.LGBMClassifier, https://github.com/microsoft/LightGBM/blob/master/docs/Parameters-Tuning.rst 

### 4. How to test?
- [ ] Unit test
com.intel.analytics.bigdl.dllib.nnframes.LightGBMTrainSpec


songhappy:
PR-validation 
http://10.112.231.51:18888/view/PR-Validation/job/BigDL-PR-Dllib-Validation/652/

scala/dllib/src/main/scala/com/intel/analytics/bigdl/dllib/nnframes/TreeModelUtils.scala
```diff
@@ -0,0 +1,66 @@
...
+
+object TreeModelUtils {
+
+  def convert2CamelCase(pythonStyle: String): String = {
+    val data = pythonStyle.split("_")
+    data(0) + data.slice(1, data.size)
+      .map(x => x.substring(0, 1).toUpperCase() + x.substring(1)).mkString("")
+  }
+
+  def setParams(lgbmEstimator: Any, lgbmParams: Map[String, Any]): Unit = {
```
dding3, line 31: maybe we can limit lgbmEstimator to LGBMModel type?

songhappy: > maybe we can limit lgbmEstimator to LGBMModel type?

there is a LightGBMBase, but it is Trait

songhappy: throwing error message if not supported in  https://github.com/intel-analytics/BigDL/pull/5789

PR #5719, opened 2022-09-13
Chronos: install openvino in Chronos-PRVN-Spark3.1-ray check

plusbang:
## Description

Some uts are added, install openvino-dev in 'Chronos PRVN Python Spark3.1 py37 ray' check correspondingly.


TheaperDeng (approved):
LGTM

PR #5726, opened 2022-09-13
[PPML]Add log file argument to bigdl-ppml-submit.sh

ShanSimu:
## Description

Add log file argument to bigdl-ppml-submit.sh and change README accordingly

### 1. Why the change?

The original log file name is fixed, which makes it impossible to record the results of multiple runs.


liu-shaojun (approved):

liu-shaojun:
Have you tested this code change?

ShanSimu:
> Have you tested this code change?

test finished

PR #5732, opened 2022-09-13
[PPML] Fix format error of quote from attestation service

xiangyuT:
## Description

Fix format error of quote from attestation service.

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/attestation/AttestationCLI.scala
+++ b/scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/attestation/AttestationCLI.scala
@@ -21,8 +21,6 @@ import org.apache.logging.log4j.LogManager
 import scopt.OptionParser
 
 import java.util.Base64
-import java.io._
-import java.nio.charset.StandardCharsets
 
 /**
  * Simple Attestation Command Line tool for attestation service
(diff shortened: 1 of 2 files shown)


qiyuangong (approved):
LGTM

qiyuangong (approved):
LGTM

scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/attestation/AttestationCLI.scala
```diff
@@ -76,10 +76,16 @@ object AttestationCLI {
 
         val challengeString = params.challenge
         if (challengeString.length() > 0) {
-            val asQuote = as.getQuoteFromServer(challengeString)
-            // System.out.print(asQuote)
+            val asQuote = params.asType match {
+              case ATTESTATION_CONVENTION.MODE_EHSM_KMS =>
+                Base64.getDecoder().decode(as.getQuoteFromServer(challengeString))
+              case ATTESTATION_CONVENTION.MODE_DUMMY =>
```
qiyuangong, line 82: By default, Dummy server return random string for bytes. The format is not correct from SGX's view. I think we can skip for dummy

PR #5739, opened 2022-09-14
cicd: add debug info to license-scan workflow

liu-shaojun:
## Description

This submission add debug steps to license-scan.yml, if fossa scan failed, the debug steps will be executed and filter the dependency tree and list jars under assembly zip to help debug.
image

image


glorysdj (approved):
LGTM

PR #5740, opened 2022-09-14
Nano : reduce time cost for InferenceOptimizer and update demo

rnwang04:
## Description

Update resnet18 demo for  InferenceOptimizer and reduce time cost for InferenceOptimizer.

### 1. Why the change?

Make the optimization pipeline faster.

### 2. User API changes

No change.

### 3. Summary of the change 

- modify the example of resnet18 according to recent changes for InferenceOptimizer
- remove accuracy calculation of traced model
- add model pruning for some really slow methods like bf16_ipex
- remove some warnings
- add ut for summary and invaliad input

### 4. How to test?
- [x] Unit test
- [x] Application test
http://10.112.231.51:18889/view/BigDL-PR-Validation/job/ZOO-PR-NanoTests/396/
http://10.112.231.51:18889/job/BigDL-PRVN-chronos-Python-Spark-3.1-py37-ray-part2/818/

### demo output
image

### DIFF
--- a/python/nano/example/pytorch/inference_pipeline/resnet/README.md
+++ b/python/nano/example/pytorch/inference_pipeline/resnet/README.md
@@ -29,18 +29,23 @@ unset KMP_AFFINITY
 ``` 
 You may find environment variables set like follows:
 ```
+OpenMP library found...
 Setting OMP_NUM_THREADS...
 Setting OMP_NUM_THREADS specified for pytorch...
 Setting KMP_AFFINITY...
 Setting KMP_BLOCKTIME...
 Setting MALLOC_CONF...
+Setting LD_PRELOAD...
+nano_vars.sh already exists
 +++++ Env Variables +++++
-LD_PRELOAD=./../lib/libjemalloc.so
-MALLOC_CONF=oversize_threshold:1,background_thread:true,metadata_thp:auto,dirty_decay_ms:-1,muzzy_decay_ms:-1
+LD_PRELOAD=/opt/anaconda3/envs/nano/bin/../lib/libiomp5.so /opt/anaconda3/envs/nano/lib/python3.7/site-packages/bigdl/nano//libs/libtcmalloc.so
+MALLOC_CONF=
 OMP_NUM_THREADS=112
 KMP_AFFINITY=granularity=fine,compact,1,0
 KMP_BLOCKTIME=1
-TF_ENABLE_ONEDNN_OPTS=
+TF_ENABLE_ONEDNN_OPTS=1
+ENABLE_TF_OPTS=1
+NANO_TF_INTER_OP=1
 +++++++++++++++++++++++++
 Complete.
 ```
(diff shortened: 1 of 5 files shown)


python/nano/example/pytorch/inference_pipeline/resnet/README.md
```diff
@@ -25,6 +25,7 @@ pip install --upgrade numpy==1.21.6
 Initialize environment variables with script `bigdl-nano-init` installed with bigdl-nano.
 ```
 source bigdl-nano-init
+unset KMP_AFFINITY
```
jason-dai, line 28: we should not have the user to manually manage this @MeouSker77 @TheaperDeng

python/nano/example/pytorch/inference_pipeline/resnet/README.md
```diff
@@ -56,23 +57,30 @@ python inference_pipeline.py
...
-accleration option: onnxruntime_fp32, latency: 44.3872ms, accuracy: 0.9937
-accleration option: onnxruntime_int8_qlinear, latency: 10.1866ms, accuracy: 0.9937
-accleration option: onnxruntime_int8_integer, latency: 18.8731ms, accuracy: 0.9875
+ -------------------------------- ---------------------- -------------- ------------
+|             method             |        status        | latency(ms)  |  accuracy  |
+ -------------------------------- ---------------------- -------------- ------------
+|            original            |      successful      |    43.447    |   0.994    |
+|           fp32_ipex            |      successful      |    32.827    |   0.994    |
+|              bf16              |   fail to forward    |     None     |    None    |
+|           bf16_ipex            |        pruned        |   201.702    |    None    |
```
jason-dai, line 69: pruned -> early stopped?

python/nano/example/pytorch/inference_pipeline/resnet/README.md
```diff
@@ -56,23 +57,30 @@ python inference_pipeline.py
...
+|         openvino_fp32          |      successful      |    10.51     |   0.994    |
+|         openvino_int8          |      successful      |    6.637     |   0.994    |
+|        onnxruntime_fp32        |      successful      |    20.55     |   0.994    |
+|    onnxruntime_int8_qlinear    |      successful      |     8.15     |   0.994    |
+|    onnxruntime_int8_integer    |   fail to convert    |     None     |    None    |
+ -------------------------------- ---------------------- -------------- ------------
+
+Optimization cost 64.3s at all.
+===========================Stop Optimization===========================
 When accelerator is onnxruntime, the model with minimal latency is:  inc + onnxruntime + qlinear
```
jason-dai, line 83: why output this result?

python/nano/example/pytorch/inference_pipeline/resnet/README.md
```diff
@@ -56,23 +57,30 @@ python inference_pipeline.py
...
+|        onnxruntime_fp32        |      successful      |    20.55     |   0.994    |
+|    onnxruntime_int8_qlinear    |      successful      |     8.15     |   0.994    |
+|    onnxruntime_int8_integer    |   fail to convert    |     None     |    None    |
+ -------------------------------- ---------------------- -------------- ------------
+
+Optimization cost 64.3s at all.
+===========================Stop Optimization===========================
 When accelerator is onnxruntime, the model with minimal latency is:  inc + onnxruntime + qlinear 
 When accuracy drop less than 5%, the model with minimal latency is:  openvino + pot 
 The model with minimal latency is:  openvino + pot
```
jason-dai, line 85: openvino + int8

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -238,19 +254,28 @@ def func_test(model, input_sample):
...
 
                 torch.set_num_threads(default_threads)
                 if self._calculate_accuracy:
-                    result_map[method]["accuracy"] =\
-                        _accuracy_calculate_helper(acce_model,
-                                                   metric, validation_data)
+                    # TODO: here we suppose trace don't change accuracy,
+                    # so we jump it to reduce time cost of optimize
+                    if precision == "fp32" and method != "original":
+                        result_map[method]["accuracy"] = result_map["original"]["accuracy"]
```
jason-dai, line 274: does this make sense? shall we say "not recomputed"?

python/nano/example/pytorch/inference_pipeline/resnet/README.md
MeouSker77: It seems the default value of `KMP_AFFINITY` in `bigdl-nano-init` is wrong, now its default value will cause program use only one core , so here unset it to use more cores. I'll fix its default value to use all cores by default

rnwang04: discuss with @MeouSker77 and he will fix this bug so that this line can be removed.

python/nano/example/pytorch/inference_pipeline/resnet/README.md
rnwang04: done

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: yeah,  "not recomputed" is more reasonable. I will change it.

python/nano/example/pytorch/inference_pipeline/resnet/README.md
rnwang04: This line was supposed to show `get_ best_model` has filter function for accelerator. Or just delete this line and only leave the result by `optimizer.get_best_model(accuracy_criterion=0.05)`

python/nano/example/pytorch/inference_pipeline/resnet/README.md
rnwang04: done

python/nano/example/pytorch/inference_pipeline/resnet/README.md
TheaperDeng: This could be a really tricky one. KMP_AFFINITY default value has some conflict with onnxruntime's core resource control (which makes it behave strangely). We cannot unset or reset this sys variable once the user start their python script. Some solution we have used in training is creating a new process(with different KMP_AFFINITY) to handle the work, while it is not reasonable as well since creating a process will cost 1-10ms and very unfriendly to low latency requirement. 

One possible solution is:
1. make `KMP_AFFINITY` default value be None(that is not setting this value in `bigdl-nano-init`)
2. Only set this value during multi-process training by creating new processes (since this var does not give much postive affect for inferencing or single process training)

What do you think? @jason-dai @MeouSker77

MeouSker77: Now the default value of `KMP_AFFINITY` is `granularity=fine,compact,1,0`. In my test, if remove the `compact` option, i.e., set `KMP_AFFINITY="fine,1,0"` can avoid the conflict with onnxruntime, so it's also a possible solution.

Here is the explanation of the `compact` option:
image

Shall we take this solution? @jason-dai @TheaperDeng 

By the way, the multi-processes training will set `KMP_AFFINITY` for sub-processes automatically, it won't be affected by this default value.

TheaperDeng: > Now the default value of `KMP_AFFINITY` is `granularity=fine,compact,1,0`. In my test, if remove the `compact` option, i.e., set `KMP_AFFINITY` to `granularity=fine,1,0` can avoid the conflict with onnxruntime, so it's also a possible solution.
> 
> Here is the explanation of the `compact` option: image
> 
> Shall we take this solution? @jason-dai @TheaperDeng
> 
> By the way, the multi-processes training will set `KMP_AFFINITY` for sub-processes automatically, it won't be affected by this default value.

I think that should be OK.

jason-dai: Looks good; need to test with our performance suit.

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -167,6 +173,17 @@ def optimize(self, model: nn.Module,
...
 
+        input_sample = tuple(next(iter(training_data))[:-1])
+        st = time.perf_counter()
+        try:
+            with torch.no_grad():
+                model(*input_sample)
+        except Exception:
+            invalidInputError(False,
+                              "training_data is incompatible with your model input.")
+            exit(1)
```
TheaperDeng, line 184: why we need to exit the process manually?

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -167,6 +173,17 @@ def optimize(self, model: nn.Module,
 
         model.eval()  # change model to eval mode
 
+        input_sample = tuple(next(iter(training_data))[:-1])
```
TheaperDeng, line 176: This basically assume users have only one output and one/multiple input for the model's forward, maybe we can inspect the model and set a more accurate param num

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: will delete this line.

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: good catch, will consider this.

PR #5741, opened 2022-09-14
Chronos: More autoformer bug fix and enhancement

TheaperDeng:
## Description
This PR brings many enhancement and bug fix to Autoformer 

### 1. Why the change?

- [x] fix `TSDataset.roll` when `id_col` is not `None`, sample num is not aligned between x,y and x_time,y_time
- [x] add a new API `AutoformerForecaster.from_tsdataset()`
- [x] support `from bigdl.chronos.forecaster import AutoformerForecaster`

### 2. User API changes
Currently, users can create a forecaster by:
```python
tsdata = TSDataset.from_pandas(...)
train_loader = tsdata.to_torch_data_loader(lookback=24, horizon=5, time_enc=True)
forecaster = AutoformerForecaster.from_tsdataset(tsdata)
```
### 3. Summary of the change 
TSDataset, AutoformerForecaster and their uts. nothing special

### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/chronos/src/bigdl/chronos/forecaster/autoformer_forecaster.py
+++ b/python/chronos/src/bigdl/chronos/forecaster/autoformer_forecaster.py
@@ -24,6 +24,7 @@ from bigdl.chronos.model.autoformer.Autoformer import AutoFormer, _transform_con
 from bigdl.nano.utils.log4Error import invalidInputError, invalidOperationError
 from bigdl.chronos.forecaster.utils import check_transformer_data
 from bigdl.chronos.pytorch import TSTrainer as Trainer
+from bigdl.chronos.data import TSDataset
 from bigdl.nano.automl.hpo.space import Space
 from bigdl.chronos.forecaster.utils_hpo import GenericTSTransformerLightningModule, \
     _config_has_search_space
@@ -38,8 +39,8 @@ class AutoformerForecaster(Forecaster):
                  future_seq_len,
                  input_feature_num,
                  output_feature_num,
-                 label_len,
                  freq,
+                 label_len=None,
                  output_attention=False,
                  moving_avg=25,
                  d_model=128,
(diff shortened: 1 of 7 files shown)


python/chronos/src/bigdl/chronos/forecaster/autoformer_forecaster.py
```diff
@@ -448,3 +531,18 @@ def metric(y_label, y_predict):
...
+
+
+def _timedelta_to_delta_str(offset):
+    features_by_offsets = (
+        (Timedelta(seconds=60), 's'),  # 6 for second - minutes
+        (Timedelta(minutes=60), 't'),  # 5 for minutes - hour
+        (Timedelta(hours=24), 'h'),  # 4 for hour - day
+        (Timedelta(days=7), 'd'),  # 3 for day - week
+        (Timedelta(days=30), 'w'),  # 2 for week - month
+        (Timedelta(days=365), 'm'),  # 1 for month - year
```
rnwang04, line 543: what does these number comment( "1/2/3/4/5/6") mean ? Don't find where use these info.

python/chronos/src/bigdl/chronos/forecaster/autoformer_forecaster.py
```diff
@@ -434,6 +436,87 @@ def save(self, checkpoint_file):
...
+            future_seq_len = future_seq_len if isinstance(future_seq_len, int) \
+                else max(future_seq_len)
+            output_feature_num = len(tsdataset.target_col)
+            input_feature_num = len(tsdataset.feature_col) + output_feature_num
+        else:
+            invalidInputError(False,
+                              "Forecaster requires 'past_seq_len' and 'future_seq_len' to specify "
+                              "the history time step and output time step.")
+        
+        if label_len is None:
```
rnwang04, line 496: what if tsdataset.label_len is not None？

python/chronos/src/bigdl/chronos/forecaster/autoformer_forecaster.py
TheaperDeng: I will delete these, the number indicates how many features will be generated in x_time and y_time under this freq

python/chronos/src/bigdl/chronos/forecaster/autoformer_forecaster.py
TheaperDeng: good catch, will consider it

rnwang04:
LGTM

PR #5744, opened 2022-09-14
Chronos: add conda env to Chronos PRVN Actions

plusbang:
## Description

Setup virtual environment in Chronos PRVN Actions.

### 1. Why the change?

Currently, the test environment is not separated which often results errors.


TheaperDeng (approved):
LGTM

PR #5745, opened 2022-09-14
[Nano] Openvino model inference notebook example with Nano

hjzin:
### 1. Why the change?
Add fundamental notebook examples to show how to run sync and async inference on OpenVINO model using Nano.

### 2. User API changes

N/A

### 3. Summary of the change 

Add fundamental notebook examples to show how to run sync and async inference on OpenVINO model using Nano.
Related issue: #5773 
### 4. How to test?
- [ ] Unit test


hjzin:
need some change after #5615 is merged, because the output format of `model(x)` is changed in that PR.

hjzin:
Example: https://zhaojie-doc.readthedocs.io/en/nano-notebook-example-openvinoir/doc/Nano/Howto/index.html#openvino

TheaperDeng:
1. The code snip (i.e. `code snip`) does not work in a bold text, may find a way to fix it.
2. Should we suggest users what this API doing is basically "predict in parallel"?
> Each asynchronous infer request will run inference independently in asynchronous mode, and 5 inference results will be returned together in a list.

hjzin:
> 2. Should we suggest users what this API doing is basically "predict in parallel"?
> 
> > Each asynchronous infer request will run inference independently in asynchronous mode, and 5 inference results will be returned together in a list.

Good idea, I will simplify the explanation.

TheaperDeng (approved):
LGTM

PR #5753, opened 2022-09-14
refactor assmbly deploy

Le-Zheng:
## Description

Refactor deploy assembly zip

Currently, we have two assembly zip. One keeps the same as previous assembly zip. The other has a all-in-one-assembly-jar in jars.
The issue is that only one assembly is released in ossh nightly. To fix this issue, this pr refactor the part and add a new execution for our second assembly zip.


Le-Zheng:
local tested by maven install:
image

PR #5754, opened 2022-09-14
fix the issue of setting LD_PRELOAD in `bigdl-nano-init` script

MeouSker77:
## Description

When setting `LD_PRELOAD`, the `bigdl-nano-init` script prefers to look for `libtcmalloc.so` and `libjemalloc.so` in `$LIB_DIR` directory, which is `${CONDA_PREFIX}/lib`. However, these .so files should be in `${NANO_DIR}/libs` directory. Fix this and check whether these .so files exist before adding them to `LD_PRELOAD`


yangw1234 (approved):
LGTM

PR #5766, opened 2022-09-15
[PPML] Fix tracker bind issue happens in xgboost examples with SGX enabled

gc-fu:
### 1. Why the change?
In the original setting, the tracker.py file cannot acquire the correct IP address to bind with SGX enabled.  

Therefore, the environment variable **RABIT_TRACKER_IP** needs to be set to tell the tracker the host address to bind.

For a detailed illustration, see the following two figures:
With SGX disabled, the tracker.py can acquire the correct hostname and then bind to the correct IP address.

native

However, with SGX enabled, the hostname and IP address it acquires are incorrect:

sgx_without_ip

This pull request will fix this problem so that the user do not need to set the **RABIT_TRACKER_IP** before trying to run xgboost examples.

### 2. User API changes
Now the user no longer needs to set **RABIT_TRACKER_IP** before running xgboost examples with SGX enabled.

### 3. Summary of the change 
Fix the tracker bind issue which causes it cannot bind to the correct IP address with SGX enabled.

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [ ] Document test
- [X] Test by running examples

The xgboost_example.py can now works correctly under all the modes including SGX cluster mode.

Besides, check the following two figures to see its correct behavior.

This figure shows its behavior w/o SGX enabled.
after_change_native

The figure shows its behavior with SGX enabled.

after_change_sgx

Now, their behaviors are identical.


qiyuangong (approved):
LGTM

qiyuangong:
Need to test it on K8S. To check if hostname is equal to pod name.

qiyuangong:
We just move XGBoost into deprecated for Occlum because Rabit's network (contains gradient, split and env) is not protect.

gc-fu:
@qiyuangong Thanks for your review :smiley:

I have tested it on k8s(icx cluster) with cluster mode and it works correctly.  But it does require that the pod have a correct setting in **/etc/hostname** and **/etc/hosts**.

For the second comment, @glorysdj shall we also set it to deprecated for Graphene or Gramine?

glorysdj:
> @qiyuangong Thanks for your review 😃
> 
> I have tested it on k8s(icx cluster) with cluster mode and it works correctly. But it does require that the pod have a correct setting in **/etc/hostname** and **/etc/hosts**.
> 
> For the second comment, @glorysdj shall we also set it to deprecated for Graphene or Gramine?

Sure.

ppml/trusted-big-data-ml/python/docker-graphene/README.md
```diff
@@ -422,9 +422,11 @@ The result should be similar to
 
 ##### Example 5: XGBoost Regressor
```
qiyuangong, line 423: Add deprecated to title

qiyuangong:
https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/

ppml/trusted-big-data-ml/python/docker-graphene/README.md
gc-fu: Thanks, will do

qiyuangong (approved):
LGTM

PR #5793, opened 2022-09-16
add bigdl-ppml jar with denpendency jar in tdx docker

Le-Zheng:
## Description

add bigdl-ppml jar with denpendency jar in tdx docker


Le-Zheng:
manual test pass: 1. docker build pass 2. spark-submit using jar-with-dependency pass on local mode.

PR #5796, opened 2022-09-16
Atari test

ForJadeForest:
## Description
Because the license of atari package is GPL and  an orca example  `rl_pong`  need this package. 
After deleting the package the jenkins test failed. So we delete the `rl_pong` example test in `run-example-test-ray.sh`


ForJadeForest:
http://10.112.231.51:18888/job/ZOO-PR-BigDL-Python-ExampleTests/1123/ all example success.

PR #5809, opened 2022-09-16
Chronos/Nano: update nano and chronos action trigger domain

TheaperDeng:
## Description
1. Nano's action should only be triggered if a nano-related PR is merged to main branch
2. Chronos' pr validation action should be triggered for `bigdl.chronos`, `bigdl.nano` or `bigdl.orca.automl/data/learn` developers

### DIFF
--- a/.github/workflows/nano_notebooks_tests.yml
+++ b/.github/workflows/nano_notebooks_tests.yml
@@ -5,6 +5,8 @@ on:
   # Triggers the workflow on push or pull request events but only for the main branch
   push:
     branches: [ main ]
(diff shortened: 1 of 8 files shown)


.github/workflows/nano_notebooks_tests.yml
```diff
@@ -5,6 +5,9 @@ on:
   # Triggers the workflow on push or pull request events but only for the main branch
   push:
     branches: [ main ]
+    paths:
```
plusbang, line 8: need push events ?

PR #5814, opened 2022-09-18
nano test on branch-2.1

pinggao187:


TheaperDeng:
This PR is only used for temp test without merging intention right?

pinggao187:
> This PR is only used for temp test without merging intention right?

yes, test completion will close.

PR #5830, opened 2022-09-19
Support tf2.1 load model for ray backend

hkvision:
tf 2.1 doesn't have options argument


sgwhat (approved):

sgwhat (commented):
maybe we should also modify this in `load_remote_model` function.

sgwhat (approved):

hkvision:
http://10.112.231.51:18888/job/BigDL-Orca-PR-Validation/1077/ some failures are irrelevant

Issue #5773, opened 2022-09-15
Nano: openvino related fundamental examples


TheaperDeng:
- [ ] need fundamental examples to `python/nano/tutorial/inference/openvino` and show how to directly load an openvino model and use it as `model(x)`. 2 examples for `async_predict`, `model(x)` respectively.

- [ ] need save/load examples to `/python/nano/tutorial/notebook/inference/pytorch` and show how to save(export)/load an optimized model.

hjzin:
I think we may also need a fundamental example to `python/nano/tutorial/inference/pytorch` to show the usage of `async_predict` on optimized pytorch model.

TheaperDeng:
> I think we may also need a fundamental example to `python/nano/tutorial/inference/pytorch` to show the usage of `async_predict` on optimized pytorch model.

Yeah I think we do, but maybe we can start a thread to first add the save(export) fundamental example first.

hjzin:
> > I think we may also need a fundamental example to `python/nano/tutorial/inference/pytorch` to show the usage of `async_predict` on optimized pytorch model.
> 
> Yeah I think we do, but maybe we can start a thread to first add the save(export) fundamental example first.

OK, I wil work on it.

PR #5832, opened 2022-09-19
Add fundamental example scripts for sync and async inference of OpenVINO model

hjzin:
## Description

Add fundamental example scripts for sync and async inference of OpenVINO model
### 1. Why the change?

related issue #5773
### 2. User API changes

N/A
### 3. Summary of the change 

- Add fundamental example scripts for image classfication task on OpenVINO model using sync and async inference API of Nano and OpenVINO
### 4. How to test?
- [ ] Unit test


python/nano/tutorial/inference/openvino/openvino_inference_async.py
```diff
@@ -0,0 +1,63 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# Required Dependecies
+
+# ```bash
+# pip install openvino-dev
+# ```
+
```
TheaperDeng, line 22: add methods to download data and checkpoints here rather than host them in bigdl repo

python/nano/tutorial/inference/openvino/openvino_inference_async.py
```diff
@@ -0,0 +1,63 @@
...
+    for id, model_input in enumerate(x):
+        infer_queue.start_async([model_input], userdata=id)
+
+    infer_queue.wait_all()
+
+    for res in async_results:
+        output_array = list(res.values())[0]
+        print(output_array.argmax(axis=1))
+
+    print("\n")
```
TheaperDeng, line 55: We don't need to show how to use openvino api here

python/nano/tutorial/inference/openvino/openvino_inference_sync.py
```diff
@@ -0,0 +1,49 @@
...
+    ie = Core()
+
+    model = ie.read_model(model=model_path)
+    compiled_model = ie.compile_model(model, device_name="CPU")
+
+    output_layer_ir = compiled_model.output(0)
+    y_hat = compiled_model([x])[output_layer_ir]
+    predictions = y_hat.argmax(axis=1)
+    print(predictions)
+
```
TheaperDeng, line 43: We don't need to show how to use openvino api here

python/nano/tutorial/inference/openvino/openvino_inference_async.py
TheaperDeng: could we find some meaningful case here (not necessary)

hjzin: > could we find some meaningful case here (not necessary)
The model in this script can be used for image classification task, maybe we could use a real image as input ? Or find some other tasks?

TheaperDeng: if that's the case, it should be OK, no need to use real image as input

TheaperDeng (approved):
LGTM

TheaperDeng:
btw, we need a test script and related script added to github workflow

hjzin:
> btw, we need a test script and related script added to github workflow

I will add them in another PR

PR #5838, opened 2022-09-19
Chronos: fix issues in unit tests of tensorflow 

plusbang:
## Description

To make the [tensorflow] installation option independently, corresponding unit tests also need to be separately run.
But there exists import errors in unit tests of tensorflow under '/autots' folder.

### 1. Why the change?

To separately run unit tests of tensorflow.

### 2. Summary of the change 

1. use LazyImport, such as change `import torch` to `torch=LazyImport('torch')`
2. add `gen_RandomDataset()` and `gen_CustomizedNet()` rather than use `RandomDataset` and `CustomizedNet` directly

### 3. How to test?
- [ ] Unit test


python/chronos/src/bigdl/chronos/autots/model/utils.py
```diff
@@ -0,0 +1,96 @@
...
+
+def model_creator_pytorch(config):
+    '''
+    Pytorch customized model creator
+    '''
+    return CustomizedNet(dropout=config["dropout"],
+                         input_size=config["past_seq_len"],
+                         input_feature_num=config["input_feature_num"],
+                         hidden_dim=config["hidden_dim"],
+                         output_size=config["output_feature_num"])
```
TheaperDeng, line 96: so this file is for testing right? should we put it here rather than under `test/`?

plusbang: done

TheaperDeng (approved):
LGTM

Issue #5606, opened 2022-08-31
Chronos: Change TCNForecaster's default parameter `num_channels` to [16]*3


TheaperDeng:
With more and more experience on this model, we found that [16]*3 is enough for most data. It has a better latency and easier to be trained.

TheaperDeng:
@devWangBin is working on this.

PR #5840, opened 2022-09-19
Chronos: Change TCNForecaster's default parameter num_channels to [16]*3

devWangBin:
## Description

Chronos: Change TCNForecaster's default parameter of num_channels from [30]*7  to [16]*3

### 1. Why the change?
Based on experimental results on real datasets, we find that the parameter setting of [16]*3 generally have more stable and superior performance. 

Github issue link: https://github.com/intel-analytics/BigDL/issues/5606


TheaperDeng:
I am pretty sure this should work, but we should be careful about this. @liangs6212 do you have any comments about this?

liangs6212:
> I am pretty sure this should work, but we should be careful about this. @liangs6212 do you have any comments about this?

Can the same modification be used for TF2 `TCNForecaster`?

TheaperDeng:
> > I am pretty sure this should work, but we should be careful about this. @liangs6212 do you have any comments about this?
> 
> Can the same modification be used for TF2 `TCNForecaster`?

Yeah, good catch. @devWangBin Please change `bigdl.chronos.forecaster.tf.tcnforecaster` as well

devWangBin:
> Yeah, good catch. @devWangBin Please change `bigdl.chronos.forecaster.tf.tcnforecaster` as well
Okay, no problem.

TheaperDeng (approved):
LGTM

TheaperDeng:
@plusbang @ruobing-wang Since this default value is used everywhere so I am asking more stockholders to have a review

rnwang04:
LGTM

plusbang:
LGTM

PR #5849, opened 2022-09-20
[PPML] Remove redundant value-assignment for JDK_URL

gc-fu:
### 1. Why the change?
Remove the redundant value-assignment for JDK_URL in the build script
It will cause error because the value of SPARK_JAR_URL will be assigned to JDK_URL

### 2. User API changes
None.

### 3. Summary of the change 
Remove the redundant value-assignment for JDK_URL in the build script

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [ ] Document test
- [x] Docker build test


ShanSimu (approved):
I'm surprised I ever wrote this mistake. Thx for fix.

PR #5856, opened 2022-09-20
fix vlun in ppml

Le-Zheng:
## Description

https://github.com/intel-analytics/arda-docker/issues/710


Le-Zheng:
http://10.112.231.51:18888/job/BigDL-NB-scala-PPML-UnitTests-Spark-3.1/350/consoleFull

PR #5860, opened 2022-09-20
fix hadoop-common 2.7.7 in serving

Le-Zheng:
## Description
org.apache.hadoop:hadoop-common@2.7.7
image

### 4. How to test?
- [x] Unit test

### New dependencies
No, upgrade version


Le-Zheng:
http://10.112.231.51:18888/job/BigDL-Serving-PR-Validation/165/

PR #5864, opened 2022-09-20
update chronos test configure on github actions

pinggao187:
## Description

delete configure
~~~
sed -i "s/'bigdl-core=='+VERSION/'bigdl-core==2.1.0b20220811'/g" python/dllib/src/setup.py
~~~

### 1. Why the change?
use bigdl-core==2.2.0.dev0
https://github.com/intel-analytics/BigDL/blob/main/python/dllib/src/setup.py#L110


plusbang (approved):
LGTM

Issue #5834, opened 2022-09-19
Chronos : List the input data type of forecaster and locate some dismatch


rnwang04:
Nowdays,  input data types for different methods in  BaseForecaster and AutoformerForecaster is different and is confusing.
The following table lists the support data types of each method:
| Method  | numpy input | loader input |  tsdataset instance | xshard input |  
| ------------- | ------------- |  ------------- |  ------------- |  ------------- |
| BaseForecaster.fit  | ✔  |  ✔ | ✔ | ✔ |
| AutoformerForecaster.fit  | ✔  | ✔ | ❌ | ❌ |
| BaseForecaster.tune  | ✔  | ❌ | ❌ | ❌ |
| AutoformerForecaster.tune  | ✔  | ❌ | ❌ | ❌ |
| BaseForecaster.predict  | ✔  |  ✔ | ✔ | ✔ |
| AutoformerForecaster.predict(*)  | ✔  | ✔ | ❌ | ❌ |
| BaseForecaster.predict_with_onnx  | ✔  |  ✔ | ✔ | ❌ |
| AutoformerForecaster.predict_with_onnx(×)  | -  | - | - | - |
| BaseForecaster.predict_with_openvino  | ✔  | ❌ | ❌ | ❌ |
| AutoformerForecaster.predict_with_onnx(×)  | -  | - | - | - |
| BaseForecaster.evaluate  | ✔  |  ✔ | ✔ | ✔ |
| AutoformerForecaster.evaluate(*)  | ✔  | ✔ | ❌ | ❌ |
| BaseForecaster.predict_interval  | ✔  |  ✔ | ✔ | ❌ |
| AutoformerForecaster.predict_interval  | ✔  | ✔ | ❌ | ❌ |

Note: × means don't have this method, * means there maybe some dismatch in metrics.
For `fit` and `predict_interval`, there are two parameters : data and validation_data(or val_data)
| Method  | numpy input | loader input |  tsdataset instance | xshard input |  
| ------------- | ------------- |  ------------- |  ------------- |  ------------- |
| data for BaseForecaster.fit  | ✔  |  ✔ | ✔ | ✔ |
| validation_data for BaseForecaster.fit  | ✔  |  ✔ | ❌ | ❌ |
| data for AutoformerForecaster.fit  | ✔  | ✔ | ❌ | ❌ |
| validation_data for AutoformerForecaster.fit(x)  | -  |  - | - | - |
[...]

PR #5865, opened 2022-09-20
Chronos : Support val data in autoformer fit

rnwang04:
## Description

We need to align function in BaseForecaster and AutoformerForecaster synchronously.

### 1. Why the change?

https://github.com/intel-analytics/BigDL/issues/5834

### 2. User API changes

```
from bigdl.chronos.forecaster import AutoformerForecaster
forecaster = AutoformerForecaster(...)
forecaster.fit(data, validation_data, validation_mode='output', earlystop_patience=1)
```

### 3. Summary of the change 

- [x] support validation_data, validation_mode and earlystop_patience for AutoformerForecaster.fit
- [x] support get_model for AutoformerForecaster
- [x] code refactor

### 4. How to test?
- [ ] Unit test
- [ ] Application test

### DIFF
--- a/python/chronos/src/bigdl/chronos/forecaster/autoformer_forecaster.py
+++ b/python/chronos/src/bigdl/chronos/forecaster/autoformer_forecaster.py
@@ -432,7 +432,7 @@ class AutoformerForecaster(Forecaster):
                     shuffle=False)
             # transform a TSDataset instance to dataloader
             if isinstance(validation_data, TSDataset):
-                _rolled = data.numpy_x is None
+                _rolled = validation_data.numpy_x is None
                 validation_data = validation_data.to_torch_data_loader(
                     batch_size=batch_size,
                     roll=_rolled,
@@ -443,6 +443,7 @@ class AutoformerForecaster(Forecaster):
                     feature_col=validation_data.roll_feature,
                     target_col=validation_data.roll_target,
                     shuffle=False)
+
             self.trainer.fit(self.internal, data, validation_data)
             self.fitted = True
             fit_out = read_csv('./forecaster_tmp_log/version_0/metrics.csv', loss_name='val_loss')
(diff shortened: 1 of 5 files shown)


python/chronos/src/bigdl/chronos/forecaster/autoformer_forecaster.py
```diff
@@ -18,6 +18,7 @@
 import numpy as np
 from pandas import Timedelta
 from bigdl.chronos.forecaster.abstract import Forecaster
+from bigdl.chronos.forecaster.utils import *
```
TheaperDeng, line 21: please use `from bigdl.chronos.forecaster.utils import xxx, xxx, xxx`

python/chronos/test/bigdl/chronos/forecaster/test_autoformer_forecaster.py
```diff
@@ -270,8 +271,7 @@ def test_autoformer_forecaster_confidence_interval_with_loader(self):
                                           output_feature_num=2,
                                           label_len=12,
                                           freq='s',
-                                          seed=0,
-                                          moving_avg=20) # even
+                                          seed=0)
```
TheaperDeng, line 274: why are we removing this even moving_avg test?

rnwang04: even moving_avg is test in test_autoformer_forecaster_even_kernel. So this test is probably not needed here.

python/chronos/src/bigdl/chronos/forecaster/autoformer_forecaster.py
rnwang04: done

TheaperDeng (approved):
LGTM

Some comment: we might need an issue states we need to remove the magic number/magic string in our code. we may use a utils file such as format.py in future to store all the default paths and some thing

PR #5870, opened 2022-09-20
Delete PYTHONHOME when start ray process

PatrickkZ:
## Description

when call `init_orca_context(cluster_mode="spark-submit")`, `PYTHONHOME` will be set inside this method. This environment variable will cause ray process start failed and throw `ModuleNotFoundError: No module named encodings`, delete PYTHONHOME when start ray process will fix this problem.
This is tested on Databricks.

### DIFF
--- a/python/orca/src/bigdl/orca/ray/ray_on_spark_context.py
+++ b/python/orca/src/bigdl/orca/ray/ray_on_spark_context.py
@@ -98,6 +98,7 @@ class RayServiceFuncGenerator(object):
         modified_env.pop("KMP_BLOCKTIME", None)
         modified_env.pop("KMP_AFFINITY", None)
         modified_env.pop("KMP_SETTINGS", None)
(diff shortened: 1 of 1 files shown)


hkvision (approved):

jason-dai:
Does this work on non-Databricks environment?

hkvision:
> Does this work on non-Databricks environment?

Should not impact. PYTHONHOME should only be used for jep...

But seems when deleting this k8s will start time out... Need more investigation.

hkvision:
Tested on k8s @hkvision  and yarn @lalalapotter and PYTHONHOME doesn't affect starting ray.
Actually the PYTHONHOME is set according to the driver environment and should not be used by executor for launching ray.

Sometimes running rayonspark will have timeout when launching ray, recorded in RayOnSpark issues.
Merging this PR first.

PR #5882, opened 2022-09-21
[PPML]add k8s pyspark example test action

ShanSimu:
## Description

add k8s pyspark example test action

### 1. Why the change?

Automatically test support for k8s pyspark

### 2. How to test?
- [x] Github Action


glorysdj (approved):
LGTM

PR #5884, opened 2022-09-21
Chronos: Run action when only test files are modified.

TheaperDeng:
## Description

Make sure action is triggerred when only test files are modified.


plusbang (approved):
LGTM

Issue #5885, opened 2022-09-21
Orca tf2 tf_runner import copy conflict


cyita:
Need to fix in both main branch and branch-2.1

PR [#5731](https://github.com/intel-analytics/BigDL/pull/5731) `from copy import copy` conflict with `import copy`

http://10.112.231.51:18888/job/BigDL-PRVN-tf2-rayonspark/365/console

leonardozcm:
branch 2.1: https://github.com/intel-analytics/BigDL/pull/5886
main: https://github.com/intel-analytics/BigDL/pull/5887

PR #5886, opened 2022-09-21
Orca: `copy` name duplicated

leonardozcm:
## Description

`from copy import copy`  will duplicate the global name `copy` in python runtime.
https://github.com/intel-analytics/BigDL/issues/5885

### 1. Why the change?

`from copy import copy`  will duplicate the global name `copy` in python runtime.

### 2. User API changes

None

### 3. Summary of the change 

`from copy import copy` -> `import copy; copy.copy(config)`

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...


leonardozcm:
http://10.112.231.51:18888/job/BigDL-PRVN-tf2-rayonspark/368/

PR #5889, opened 2022-09-21
[Nano] Add save/load fundamental examples for optimized Pytorch model

hjzin:
## Description

Add save/load fundamental examples for optimized Pytorch model.
### 1. Why the change?

related issue: #5773 

### 2. User API changes

N/A
### 3. Summary of the change 

Add save/load fundamental examples for optimized Pytorch model using openvino and onnxruntime as accelerator.
### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/nano/tutorial/inference/pytorch/pytorch_save_and_load_onnx.py
+++ b/python/nano/tutorial/inference/pytorch/pytorch_save_and_load_onnx.py
@@ -39,6 +39,10 @@ if __name__ == "__main__":
     print(predictions)
 
     # Save Optimized Model
+    # The saved model files will be saved at "./optimized_model_ort" directory
+    # There are 2 files in optimized_model_ort:
+    #   nano_model_meta.yml: meta information of the saved model checkpoint
+    #   onnx_saved_model.onnx: model checkpoint for general use, describes model structure
     InferenceOptimizer.save(ort_model, "./optimized_model_ort")
 
     # Load the Optimized Model
--- a/python/nano/tutorial/inference/pytorch/pytorch_save_and_load_onnx.py
+++ b/python/nano/tutorial/inference/pytorch/pytorch_save_and_load_onnx.py
@@ -30,16 +30,17 @@ if __name__ == "__main__":
     model_ft.eval()
     
     # Accelerated Inference Using ONNX Runtime
-    from bigdl.nano.pytorch import Trainer
-    ort_model = Trainer.trace(model_ft,
-                              accelerator="onnxruntime",
-                              input_sample=torch.rand(1, 3, 224, 224))
+    from bigdl.nano.pytorch import InferenceOptimizer
+    ort_model = InferenceOptimizer.trace(model_ft,
+                                         accelerator="onnxruntime",
+                                         input_sample=torch.rand(1, 3, 224, 224))
 
     y_hat = ort_model(x)
     predictions = y_hat.argmax(dim=1)
     print(predictions)
 
     # Save Optimized Model
+    from bigdl.nano.pytorch import Trainer
     Trainer.save(ort_model, "./optimized_model_ort")
 
     # Load the Optimized Model
--- a/python/nano/tutorial/inference/pytorch/pytorch_save_and_load_onnx.py
+++ b/python/nano/tutorial/inference/pytorch/pytorch_save_and_load_onnx.py
@@ -1,0 +1,51 @@
+#
+# Copyright 2016 The BigDL Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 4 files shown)


rnwang04:
Maybe use InferenceOptimizer instead of Trainer here.

python/nano/tutorial/inference/pytorch/pytorch_save_and_load_onnx.py
```diff
@@ -0,0 +1,52 @@
...
+    ort_model = InferenceOptimizer.trace(model_ft,
+                                         accelerator="onnxruntime",
+                                         input_sample=torch.rand(1, 3, 224, 224))
+
+    y_hat = ort_model(x)
+    predictions = y_hat.argmax(dim=1)
+    print(predictions)
+
+    # Save Optimized Model
+    from bigdl.nano.pytorch import Trainer
```
rnwang04, line 43: No need for `import Trainer`.
Now InferenceOptimizer also provide save / load function. You can just import InferenceOptimizer and use InferenceOptimizer.save / InferenceOptimizer.load here.

jason-dai: Explain how the user can also retrieve the standard model (ONNX or OpenVINO) files

TheaperDeng: Need explaination about the files in saving dir (one is a onnx/openvino ckpt for general use, the other one is a meta information illustrates the precision and other information about the general ckpt.)

python/nano/tutorial/inference/pytorch/pytorch_save_and_load_onnx.py
```diff
@@ -0,0 +1,52 @@
...
+# ```
+
+import torch
+from torchvision.models import resnet18
+
+if __name__ == "__main__":
+    model_ft = resnet18(pretrained=True)
+
+    x = torch.rand(2, 3, 224, 224)
+    model_ft.eval()
```
TheaperDeng, line 30: no need to switch to eval mode

rnwang04: At present， it seems that we don't set `model.eval()` in openvino and onnxruntime.

python/nano/tutorial/inference/pytorch/pytorch_save_and_load_onnx.py
hjzin: Added some explanation for saved model files

python/nano/tutorial/inference/pytorch/pytorch_save_and_load_onnx.py
hjzin: > At present, it seems that we don't set `model.eval()` in openvino and onnxruntime.

So you think maybe we could add `model.eval` here?

rnwang04: Yeah, i think could add here.  Maybe we should add `model.eval()` in `trace` and `quantize` later.

python/nano/tutorial/inference/pytorch/pytorch_save_and_load_onnx.py
```diff
@@ -0,0 +1,55 @@
...
+
+    y_hat = ort_model(x)
+    predictions = y_hat.argmax(dim=1)
+    print(predictions)
+
+    # Save Optimized Model
+    # The saved model files will be saved at "./optimized_model_ort" directory
+    # There are 2 files in optimized_model_ort:
+    #   nano_model_meta.yml: meta information of the saved model checkpoint
+    #   onnx_saved_model.onnx: model checkpoint for general use, describes model structure
```
TheaperDeng, line 46: maybe add "Users only need to take this file for further usage."

python/nano/tutorial/inference/pytorch/pytorch_save_and_load_openvino.py
```diff
@@ -0,0 +1,56 @@
...
+    y_hat = ov_model(x)
+    predictions = y_hat.argmax(dim=1)
+    print(predictions)
+
+    # Save Optimized Model
+    # The model files will be saved at "./optimized_model_ov" directory
+    # There are 3 files in optimized_model_ov:
+    #   nano_model_meta.yml: meta information of the saved model checkpoint
+    #   ov_saved_model.bin: contains the weights and biases binary data of model
+    #   ov_saved_model.xml: model checkpoint for general use, describes model structure
```
TheaperDeng, line 47: maybe add "Users only need to take last 2 files for further usage."

TheaperDeng (commented):
others LGTM

python/nano/tutorial/inference/pytorch/pytorch_save_and_load_openvino.py
hjzin: done

TheaperDeng (approved):
LGTM

PR #5892, opened 2022-09-21
Chronos: support thread ctrl in openvino's prediction

TheaperDeng:
## Description

Since we have now supported thread_num control for openvino, we could add this method in forecaster as well


rnwang04:
LGTM

PR #5901, opened 2022-09-22
Chronos: add description about thread_num control for openvino in how-to guide

plusbang:
## Description

Since we have now supported thread_num control for openvino in forecaster, we should add this description in how-to guide.

### 4. How to test?
- [x] Document test
https://binbin-5901.readthedocs.io/en/latest/doc/Chronos/Howto/how_to_speedup_inference_of_forecaster_through_OpenVINO.html


TheaperDeng (approved):
LGTM

PR #5903, opened 2022-09-22
Add attribute sys.stdout.fileno when it is missed on Databricks

PatrickkZ:
## Description

In [Databricks RayDemo](https://www.databricks.com/notebooks/raydemo.html) mentioned:
```python
# Ray will crash unless sys.stdout is modified.
import sys
 
sys.stdout.fileno = lambda: 0
```
This is because in Databricks notebook, sys.stdout is redirect to a `ConsoleBuffer` object, this object has no attribute fileno, and cause ray init crash. Normally, sys.stdout should have attribute fileno and is set to 1. So set sys.stdout.fileno to 1 when this attribute is missed, this should fix the bug when run BigDL ray backend on Databricks, and should not affect non-Databricks env.

This is verified on Databricks.

### DIFF
--- a/python/orca/src/bigdl/orca/ray/raycontext.py
+++ b/python/orca/src/bigdl/orca/ray/raycontext.py
@@ -30,6 +30,9 @@ class OrcaRayContext(object):
                  cores=2,
                  num_nodes=1,
                  **kwargs):
+        import sys
+        if not hasattr(sys.stdout, 'fileno'):
+            sys.stdout.fileno = lambda: 1
 
         self.runtime = runtime
         self.initialized = False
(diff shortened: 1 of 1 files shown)


hkvision (approved):

PatrickkZ:
Test on Jenkins，http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-Orca-PR-Validation/1117/
some fail is irrelevant

python/orca/src/bigdl/orca/ray/raycontext.py
```diff
@@ -30,6 +30,9 @@ def __init__(self,
                  cores=2,
                  num_nodes=1,
                  **kwargs):
+        # Add sys.stdout.fileno for Databricks. In Databricks notebook, sys.stdout is redirect to a ConsoleBuffer
+        # object, this object has no attribute fileno, and cause ray init crash. Normally, sys.stdout should have
+        # attribute fileno and is set to 1. So set sys.stdout.fileno to 1 when this attribute is missed
```
hkvision, line 35: redirect -> redirected
missed -> missing
. at the very end

PR #5917, opened 2022-09-22
Fix the access of custom attributes when using `TorchNano` and multi-instance training

MeouSker77:
## Description

- Fix the access of custom attributes when using `TorchNano` and multi-instance training.
- Fix the state dict saved by multi-instance training cannot be loaded in single-instance training.

### 1. Why the change?

Before this PR, when user runs multi-instance training(i.e. set `num_processes=n`, where n>1), he cannot access the custom properties of the model. For example:

```python
class MyNano(TorchNano):
    def train():
        origin_model = ModelClass(...)
        model = self.setup(origin_model)
        model.xxx
``` 

Here `model` is an instance of `_TorchNanoModule`.
If `xxx` attribute doesn't exist in `_TorchNanoModule`, `_TorchNanoModule` will find `xxx` attribute in `model.module`, which is the original model in non-multi-instance training. 
However, when using multi-instance training, `model.module` will be DDP(DistributedDataParallel), instead, `model.module.module` will be the original model.

Before this PR, user cannot load the state dict saved by multi-instance training.

### 2. User API changes

Before this PR, user can only use `model.custom_attribute` to access custom attributes in single-instance training, he must use `model.module.custom_attribute` in multi-instance training.

Now user can use `model.custom_attribute` in both single-instance and multi-instance training.

Now user can load the state dict saved by multi-instance training.

### 3. Summary of the change 

- Fix the access of custom attributes when using `TorchNano` and multi-instance training
- Fix the state dict saved by multi-instance training cannot be loaded in single-instance training
- Add and fix UT

### 4. How to test?
- [ ] N/A
- [X] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/torch_nano.py
+++ b/python/nano/src/bigdl/nano/pytorch/torch_nano.py
@@ -55,7 +55,12 @@ class _TorchNanoModule(_LiteModule):
         except AttributeError:
             pass
 
+        # When using multi-instance training, self.module will be DistributedDataParallel(DDP),
+        # otherwise, `self.module` will be original module.
         if isinstance(self.module, DistributedDataParallel):
+            # just in case that users try to access an attribute of DDP
+            # or an attribute of both DDP and original model,
+            # we should first try to find it in DDP
             try:
                 return getattr(self.module, name)
             except AttributeError:
--- a/python/nano/src/bigdl/nano/pytorch/torch_nano.py
+++ b/python/nano/src/bigdl/nano/pytorch/torch_nano.py
@@ -23,6 +23,7 @@ import torch
 from torch import nn
 from torch.optim import Optimizer
 from torch.utils.data import DataLoader
+from torch.nn.parallel.distributed import DistributedDataParallel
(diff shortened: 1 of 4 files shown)


python/nano/src/bigdl/nano/pytorch/torch_nano.py
```diff
@@ -36,17 +37,31 @@
...
+        # automatically unwrap attributes access of _LiteModule,
+        # always throw a single-level exception when the attribute doesn't exist
+        # for a more user-friendly exception message
         try:
             return super().__getattr__(name)
         except AttributeError:
+            pass
+
+        if isinstance(self.module, DistributedDataParallel):
+            try:
```
yangw1234, line 59: maybe adding some comments on why trying to access `self.module.custome_attr` first?

yangw1234:
LGTM

python/nano/src/bigdl/nano/pytorch/torch_nano.py
MeouSker77: done

Issue #5877, opened 2022-09-21
Orca: call spark_df_to_pd_sparkxshards() failed after joining two spark dataframes


zpeng1898:
When I want to call `spark_df_to_pd_sparkxshards()` to transform the `spark dataframe` to `sparkxshards` after the `join()` operation, an error occurs in the function. It shows `Must pass schema or at least one RecordBatch`. But I can print the schema of the spark dataframe successfully.  Maybe there is something wrong in the function.

**Source Code**
![image](https://user-images.githubusercontent.com/42887453/191401815-0a61c48a-8552-4194-bf58-9dd8609a2be3.png)
![image](https://user-images.githubusercontent.com/42887453/191401824-16cb6a47-af26-40cb-81a9-93518044cf0e.png)

**Error**
![image](https://user-images.githubusercontent.com/42887453/191401851-e18c9742-d138-4f5b-86cf-34597cea4c20.png)
![image](https://user-images.githubusercontent.com/42887453/191401858-a6a904ce-8aa3-4d3c-86f7-dce59d6eada0.png)
![image](https://user-images.githubusercontent.com/42887453/191401872-aaf15309-0104-4365-866c-4ba7a0eeeee6.png)

**The `schema` is printed successfully**
![image](https://user-images.githubusercontent.com/42887453/191402017-2803f442-15e6-49a8-ae3b-2dd0b3baf593.png)

**But when I call the spark.dataframe.limit() function, it runs without error**
![image](https://user-images.githubusercontent.com/42887453/191402749-0011eabc-c2b6-4eed-b6d5-537e4ec993c7.png)

hkvision:
@dding3 Please take a look.

zpeng1898:
[Executor task launch worker for task 0.0 in stage 6.0 (TID 6)] WARN  org.apache.spark.storage.BlockManager  - Putting block rdd_49_0 failed due to exception org.apache.spark.api.python.PythonException: Traceback (most recent call last):
  File "/root/miniconda3/envs/py372/lib/python3.7/site-packages/pyspark/python/lib/pyspark.zip/pyspark/worker.py", line 604, in main
    process()
  File "/root/miniconda3/envs/py372/lib/python3.7/site-packages/pyspark/python/lib/pyspark.zip/pyspark/worker.py", line 596, in process
    serializer.dump_stream(out_iter, outfile)
[...]

PR #5922, opened 2022-09-22
process empty parition

dding3:
## Description
Fix https://github.com/intel-analytics/BigDL/issues/5877. Add logic to check if there is empty partition of spark dataframe, if there is and if `spark_df_to_pd_sparkxshards` is called, will throw exception.


jason-dai:
Why not create an empty spark dataframe partition?

dding3:
> Why not create an empty spark dataframe partition?

You mean create empty pandas df for empty spark dataframe partition? I think it may cause potential problems in the further processing(eg. `filter`) and the error message may be confusing? Also I think it may bring extra overheads for many empty partitions?

PR #5929, opened 2022-09-23
Nano : add sample size for quantize

rnwang04:
## Description

During demo writing, we find in some cases, openvino pot process cost too much time (more than 10 minutes).
So plan to expose `sample_size` paramter of pot to quantize, and manully change it in optimize().

### 1. Why the change?

Reduce time cost of openvino pot process.

### 2. User API changes

new parameter of quantize : 
`InferenceOptimizer.quantize(..., sample_size=15)`

### 3. Summary of the change 

expose `sample_size` paramter of pot to quantize, and manully change it in optimize()

### 4. How to test?
- [ ] Unit test
- [ ] Application test

case : binary_classification_for_defect_detection
After this change: 
image


TheaperDeng:
I wonder if this will affect the accuracy drop in general case

PR #5935, opened 2022-09-23
fix DBFS file path problem on Dataricks

PatrickkZ:
## Description

when using pytorch on Databricks, there are some problems.
- driver don't  support loading `state.pkl` from DBFS 
- don't support saving pytorch model to DBFS
- don't support loading pytorch model from DBFS

when call
```python
orca_estimator = Estimator.from_torch(model=model_creator,
                                      optimizer=optimizer_creator,
                                      loss=criterion,
                                      metrics=[Accuracy()],
                                      model_dir=model_dir',
                                      use_tqdm=True,
                                      backend=backend)
```
on Databricks cluster, model_dir should be a shared path to save files, which is a path starts with /dbfs or dbfs:/ on Databricks. but currently, the DBFS format path is not support yet, the same problem happens when call `orca_estimator.save()` or `orca_estimator.load()`

### 1. Why the change?

To fix the mentioned problems above.
-  support loading `state.pkl` from DBFS 
- support saving pytorch model to DBFS
- support loading pytorch model from DBFS
### 2. User API changes

make sure the model save/load path starts with `/dbfs` on databricks.
### 3. Summary of the change 

Fix pytorch model save/load problems on Databricks.
### 4. How to test?
- [ ] N/A
- [x] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

Verified on Databricks.

### DIFF
--- a/python/orca/src/bigdl/orca/data/file.py
+++ b/python/orca/src/bigdl/orca/data/file.py
@@ -221,7 +221,7 @@ def write_text(path, text):
             result = f.write(text.encode('utf-8'))
             f.close()
             return result
-    elif path.startswith("s3"):   # s3://bucket/file_path
+    elif path.startswith("s3"):  # s3://bucket/file_path
         access_key_id = os.environ["AWS_ACCESS_KEY_ID"]
         secret_access_key = os.environ["AWS_SECRET_ACCESS_KEY"]
         import boto3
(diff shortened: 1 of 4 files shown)


hkvision (approved):

hkvision (approved):

python/orca/src/bigdl/orca/data/file.py
```diff
@@ -578,3 +577,51 @@ def fs_load(path, *args, **kwargs):
...
+            import uuid
+            import tempfile
+            from bigdl.dllib.utils.file_utils import append_suffix
+            file_name = str(uuid.uuid1())
+            file_name = append_suffix(file_name, path)
+            temp_path = os.path.join(tempfile.gettempdir(), file_name)
+
+            try:
+                result = save_func(obj, temp_path, *args, **kwargs)
+                put_local_file_to_remote(temp_path, path)
```
lalalapotter, line 599: Do we need to consider whether we can overwrite the original content?

PatrickkZ:
Test on Jenkins
http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-Orca-PR-Validation/1134/,
one test failure is irrelevant

python/orca/src/bigdl/orca/data/file.py
hkvision: the function only supports filemode, no overwrite option, pending this for now.

PR #5942, opened 2022-09-26
[Nano] Add an option for `framework` parameter in `build_and_install.sh` to install nano without any framework

### LINKED ISSUE #5919: Nano: install nano without frameworks like TF or Pytorch could be supported in `build_and_install.sh`
We can use `OpenVINOModel` API to load an OpenVINO model and do inference in Nano, so in this case, there is no need to install nano with frameworks like TF and Pytorch, and we can simply run `pip install bigdl-nano` to install without framework options.

In `python/nano/dev/build_and_install.sh`, if `framework` parameter is not specified, it will raise an error.
https://github.com/intel-analytics/BigDL/blob/main/python/nano/dev/build_and_install.sh#L23-L28

Maybe we can make `framework` parameter optional to support this circumstance.

hjzin:
## Description

This PR wants to simplify the usage of building and installing Nano when we need to install Nano without any framework.

### 1. Why the change?

resolve #5919 
When inferencing with OpenVINO model using `OpenVINOModel`, we don't need to install Nano with TF or Pytorch. 
When we add test scripts to github workflow, we often use `build_and_install.sh` to build and install Nano. But the `build_and_install.sh` will raise an error when we try to install Nano without any framework, so we have to use more commands to install in this circumstance. For example, in `.github/workflows/nano_unit_tests_basic.yml`:
```bash
bash python/nano/dev/release_default_linux.sh default false
whl_name=`ls python/nano/dist`
pip install python/nano/dist/${whl_name}
```
Since these commands are already in `build_and_install.sh`, so we can add an option for the  `framework` parameter in the script to simplify the usage.
This PR adds an option "basic" to `framework` parameter, if `framework` is specified to "basic", Nano will be installed without any framework.
After this change, the install commands in `.github/workflows/nano_unit_tests_basic.yml` can be changed to:
```bash
bash python/nano/dev/build_and_install.sh linux default false basic
```
### 2. User API changes

When using `build_and_install.sh` to install Nano, if `framework` is specified to "basic", Nano will be installed without any framework.
### 3. Summary of the change 

- Add an option "basic" for `framework` parameter in `build_and_install.sh` to install Nano without any framework
- Simplify the install command in `.github/workflows/nano_unit_tests_basic.yml` 
### 4. How to test?
- [ ] Unit test

### DIFF
--- a/.github/workflows/nano_unit_tests_basic.yml
+++ b/.github/workflows/nano_unit_tests_basic.yml
@@ -41,9 +41,7 @@ jobs:
           $CONDA/bin/conda create -n bigdl-init -y python==3.7.10 setuptools==58.0.4
           source $CONDA/bin/activate bigdl-init
           $CONDA/bin/conda info
-          bash python/nano/dev/release_default_linux.sh default false
-          whl_name=`ls python/nano/dist`
-          pip install python/nano/dist/${whl_name}
+          bash python/nano/dev/build_and_install.sh linux default false basic
           source bigdl-nano-init
           if [ 0"$LD_PRELOAD" = "0" ]; then
             exit 1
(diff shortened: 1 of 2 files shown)


jason-dai:
Does it change the behavior of pip install?

hjzin:
> Does it change the behavior of pip install?

When installing Nano without framework, the install command will change, but this PR doesn't affect the current usage when we need to install Nano with TF or Pytorch.
When `framework` is specified to "tensorflow" or "pytorch", it will run `pip install bidgl-nano[tensorflow]` or `pip install bidgl-nano[pytorch]`.
When `framework` is specified to "basic", it will run `pip install bidgl-nano`.

hjzin:
@glorysdj @liu-shaojun could you please take a look at it?

glorysdj (approved):
LGTM

Issue #5943, opened 2022-09-26
One-line patch API: feature and prospective


TheaperDeng:
## Introduction

Inspired by `scikit-learn-intelex`, we explore the possibility to use a patching API for  bigdl-nano to easy the usage of current functions and enable more functions(e.g. GPU->CPU tool).

Here is a simple usage demo

```python
from bigdl.nano import patch_something
# patch many modules(e.g.torch.Tensor.cuda) to customized ones
patch_something()

from patched_module import patched_class
patched_class  # it is changed to customized_class
```

## Methodology

Nothing special is used here other than monkey-patchining. A typical implementation is here:

```python
import original_module
from replace_module import REPLACE_MODULE

def patch():
    setattr(original_module, "ORIGINAL_MODULE", REPLACE_MODULE)
```

## Features

To make it clear, the patching API should be called **before importing the original Class**. E.g.

```python
from patched_module import patched_class
from bigdl.nano import patch_something

patch_something()
patched_class  # it is still patched_class
```

But it does not matter if you only import the module. E.g.

```python
import patched_module as pm
from bigdl.nano import patch_something

patch_something()
pm.patched_class  # it is changed to customized_class
```

Monkey patch is still very tricky implementation, we should be very very carefull about this.

### Change GPU only code to CPU friendly ones

Will be illustrated more detailedly in https://github.com/analytics-zoo/nano-thirdparty-examples/pull/11

### Enable more friendly API in existing APIs

- Tensorflow: to raise a PR later

jason-dai:
Shall we have two APIs, one for `patch_tensorflow`, and the other for `patch_torch`? Patching GPU should be one of the features in these two APIs?

TheaperDeng:
> Shall we have two APIs, one for `patch_tensorflow`, and the other for `patch_torch`? Patching GPU should be one of the features in these two APIs?

so we will have 2 APIs
1. pytorch
```python
from bigdl.nano.pytorch import patch_torch
patch_torch(gpu_to_cpu=True/False(default))  # stay default to False until the gpu-cpu patch is stable.
```
2. tensorflow
```python
from bigdl.nano.tf import patch_tensorflow
patch_tensorflow(gpu_to_cpu=True/False(default))
```

@yangw1234 any thoughts about this?

yangw1234:
> > Shall we have two APIs, one for `patch_tensorflow`, and the other for `patch_torch`? Patching GPU should be one of the features in these two APIs?
> 
[...]

PR #5946, opened 2022-09-26
Nano: add a patch_torch/patch_tensorflow API for friendly usage

TheaperDeng:
## Description
This PR add a new patching API which can patch all the optimized class
refer to https://github.com/intel-analytics/BigDL/issues/5943 for details

### Simple usage
```python
from bigdl.nano.pytorch import patch_torch
patch_torch()

# all code is optimized
# 1. pytorch_lightning.Trainer -> bigdl.nano.pytorch.Trainer
# 2. torchvision.transforms -> bigdl.nano.pytorch.vision.transforms
# 3. torchvision.datasets -> bigdl.nano.pytorch.vision.datasets

unpatch_torch()
# code is not optimized (it won't change the instances' type)

from bigdl.nano.tf import patch_tensorflow
patch_tensorflow()

# all code is optimized
# 1. tf.keras.Model/keras.Model -> bigdl.nano.tf.keras.Model
# 2. tf.keras.Model/keras.Sequential -> bigdl.nano.tf.keras.Sequential
# 3. tf.keras.layers.Embedding/keras.layers.Embedding -> bigdl.nano.tf.keras.layers.Embedding
# 4. tf.optimizers.Adam -> bigdl.nano.tf.optimizers.SparseAdam

unpatch_tensorflow()
# code is not optimized (it won't change the instances' type)
```

- [x] More validation work will be done (tutorial of pytorch lightning and tensorflow training passed)
- [x] `unpatch_torch`/`unpatch_tensorflow` is needed

**gpu-to-cpu will be added in later PR to make the PR controllable**


yangw1234:
looks a little dangerous to me

TheaperDeng:
> looks a little dangerous to me

Yeah, exactly that's a concern. Still, `scikit-learn-intelex` use same method for quite a few years and it seems to work fine https://github.com/intel/scikit-learn-intelex/blob/db12454b49a2fb3097567a2ee2b68852b221e4b5/daal4py/sklearn/monkeypatch/dispatcher.py#L102-L111

python/nano/src/bigdl/nano/__init__.py
```diff
@@ -41,6 +43,47 @@ def _check_nano_envs():
 # _check_nano_envs()
 
 
+@lru_cache(maxsize=None)
```
yangw1234: why adding a cache here? Will `get_patch_map` be called many times?

TheaperDeng: It doesn't matter if we only need `patch_nano`. This cache here is to make this function return the same reference every time. This is really important if I want the "original" class back during `unpatch_nano`.

Inspired by https://github.com/intel/scikit-learn-intelex/blob/develop/daal4py/sklearn/monkeypatch/dispatcher.py#L57

yangw1234: Does the following code work? Seems more easy to understand.

```python

_tf_mapping = None
_pytorch_mapping = None

def get_mapping():
    global _tf_mapping
    global _pytorch_mapping

    if _tf_mapping is None:
        _tf_mapping = xxx

    if _pytorch_mapping is None:
        _pytorch_mapping = xxx
    return _tf_mapping, _pytorch_mapping
```

`unpatch` then just use `_tf_mapping` and `_pytorch_mapping`

yangw1234: Or just adding some comments. I am fine with either way.

TheaperDeng: > Does the following code work? Seems more easy to understand.
> 
> ```python
> _tf_mapping = None
> _pytorch_mapping = None
> 
> def get_mapping():
>     global _tf_mapping
>     global _pytorch_mapping
> 
>     if _tf_mapping is None:
>         _tf_mapping = xxx
> 
>     if _pytorch_mapping is None:
>         _pytorch_mapping = xxx
>     return _tf_mapping, _pytorch_mapping
> ```
> 
> `unpatch` then just use `_tf_mapping` and `_pytorch_mapping`

Yeah, that should be much clearer (if works), I will try it later. Thx!

TheaperDeng: it works, changed to global var

python/nano/src/bigdl/nano/dispatcher.py
```diff
@@ -0,0 +1,130 @@
...
+           will unpatch defaultly if tensorflow is installed
+    :param unpatch_torch: bool, if unpatch pytorch related classes,
+           will unpatch defaultly if pytorch is installed
+    """
+    if unpatch_tf is None:
+        unpatch_tf = find_spec("tensorflow") is not None
+    if unpatch_torch is None:
+        unpatch_torch = find_spec("pytorch_lightning") is not None
+
+    mapping_tf, mapping_torch = _get_patch_map()
```
yangw1234, line 122: maybe just directly use global variable `_mapping_tf ` and `_mapping_torch `?

Say in jupyter notebook:
1. user installs tensorflow
2. user calls patch nano
3. user decide to try pytorch and install pytorch
4. user calls unpatch nano

Is the behavior correct in this scenario?

TheaperDeng: will check and give feedbacks here.

TheaperDeng: well it won't happen under current API design. But I tried this:
1. users install pytorch-lightning
2. users call patch torch
3. users uninstall pytorch-lightning
4. users call unpatch torch
Although I don't think this is a common use-case, but at least the api calling does not cause errors.

python/nano/src/bigdl/nano/pytorch/dispatcher.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+_mapping_torch = None
+
+
+def _get_patch_map():
+    global _mapping_torch
+
+    # decide if generate
+    patch_torch = find_spec("pytorch_lightning") is not None
```
jason-dai, line 27: I think you can still patch `torchvision` even if `pytorch_lightning` is not available?

TheaperDeng: yeah, I can do that for the patching logic.

btw, I don't think a user who install `torchvision` but not `pytorch_lightning` can use nano smoothly (due to some `__init__.py`'s importing logic), we may fix those importing in later PR if necessary.

jason-dai: > yeah, I can do that for the patching logic.
> 
> btw, I don't think a user who install `torchvision` but not `pytorch_lightning` can use nano smoothly (due to some `__init__.py`'s importing logic), we may fix those importing in later PR if necessary.

The goal is to support pytorch-only users using `TorchNano`; we should not depend on `pytorch_lightning` in this case.

TheaperDeng: yeah, but `TorchNano` is inherited from `LightningLite` which is depended on `pytorch_lightning`. So maybe the only goal for this is to support those who only want to use `torchvision`'s acceleration.

yangw1234 (approved):
LGTM

PR #5954, opened 2022-09-26
Change BIGDL_VERSION in Dockerfile.

Hanyu-Jin:
## Description

  Change BIGDL_VERSION in Dockerfile.

### 1. Why the change?

  The BIGDL_VERSION of the newest assembly is 2.2.0-SNAPSHOT, while the BIGDL_VERSION in Dockerfile is still 2.1.0-SNAPSHOT

### 2. User API changes

NONE.

### 3. Summary of the change 

  Change BIGDL_VERSION in Dockerfile.
  Modify several start-scripts from fixed bigdl ersion and spark version to versions obtained from environment variables.


glorysdj (approved):
LGTM

Issue #5715, opened 2022-09-12
how to use forecaster to predict future data


TheaperDeng:
We need such a how to guide to help users to understand how to predict future data by using a trained forecaster

Critical APIs includes
`forecaster.predict`
`tsdataset.unscale_numpy()` (and why it is important)

AndyLuo1029:
I've finished the first version of this how-to-guide. Here is a preview link: https://andytestdoc.readthedocs.io/en/jupyter/doc/Chronos/Howto/how_to_use_forecaster_to_predict_future_data.html

PR #5959, opened 2022-09-26
Chronos: upload prediction how-to-guide.

AndyLuo1029:
Finished the second version of how-to-guide notebook file as required in #5715 

#### But there is a Problem:

During testing the execution of the notebook both locally and on colab, @Oscilloscope98  and I found that when the `exit( )` function exist, the kernel will crash and the execution will fail. So I deleted this function in my how-to-guide.

I also tried to run other notebooks which contain `exit( )` locally and on colab (by clicking the colab logo on their how-to-guide page), they all failed :(

We think this is quite a serious problem so I mention it here.

### DIFF
--- a/docs/readthedocs/source/_toc.yml
+++ b/docs/readthedocs/source/_toc.yml
@@ -185,6 +185,7 @@ subtrees:
                 - file: doc/Chronos/Howto/how_to_speedup_inference_of_forecaster_through_ONNXRuntime
                 - file: doc/Chronos/Howto/how_to_speedup_inference_of_forecaster_through_OpenVINO
                 - file: doc/Chronos/Howto/how_to_evaluate_a_forecaster
+                - file: doc/Chronos/Howto/how_to_use_forecaster_to_predict_future_data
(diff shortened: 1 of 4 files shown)


AndyLuo1029:
The [preview of doc is here](https://andytestdoc.readthedocs.io/en/jupyter/doc/Chronos/Howto/how_to_use_forecaster_to_predict_future_data.html) and the [executable colab page is here](https://colab.research.google.com/github/AndyLuo1029/BigDL/blob/jupyter/python/chronos/colab-notebook/howto/how_to_use_forecaster_to_predict_future_data.ipynb).

Since the link I added in my guide will be valid only after this pr is merged, you can't redirect to a useable page by clicking the colab logo on the doc page.

TheaperDeng:
don't think we need this sentence

> When forecaster is non-distributed, aggregating of multiple output values can be defined through multioutput. 

We may need more explanation about 

> To make a pure prediction

add something like "This will enable you to predict target features of unobserved timestamps"

PR #4737, opened 2022-05-31
[Orca] Tutorials for Submit and Run Programs on Yarn

sgwhat:
Add tutorials for submitting and running programs on yarn.

- [x] rename `spark-submit-with-bigdl` to `bigdl-submit`
- [x] add a multiple python files demo
- [x] built-in
- [x] bigdl-submit
- [x] spark-submit

PR #5960, opened 2022-09-26
[Orca] Add bigdl-submit mode support

sgwhat:
## Description
Separated from **Yarn Tutorial PR** #4737 .
In this PR, we add `bigdl-submit` script (rename from `spark-submit-with-bigdl`) and `bigdl-submit` cluster mode in Orca for yarn users.   

### 1. Why the change?
As a new cluster mode, `bigdl-submit` will not only includes `spark-submit`, but also  `ray job submit` etc. TThis PR could be thought of as a preparation for adding more `submit` supports.

### 2. User API changes

Users need to specify the cluster mode as `bigdl-submit`.

```python
# In train.py
from bigdl.orca import init_orca_context

init_orca_context(cluster_mode="bigdl-submit")
```

Then submit the application as below:
```bash
${BIGDL_HOME}/scripts/bigdl-submit \
    --master yarn \
    --deploy-mode client \
    --archives /path/to/environment.tar.gz#environment \
    --conf spark.pyspark.driver.python=/path/to/python \
    --conf spark.pyspark.python=environment/bin/python \
    train.py
```

### 3. How to test?
- [x] Application test


hkvision:
Need to modify all the places that uses spark-submit-with-bigdl, at least the following two:
https://github.com/intel-analytics/BigDL/blob/main/python/dllib/src/setup.py#L91
https://github.com/intel-analytics/BigDL/blob/main/python/orca/src/bigdl/orca/common.py#L191

sgwhat:
> Need to modify all the places that uses spark-submit-with-bigdl, at least the following two: https://github.com/intel-analytics/BigDL/blob/main/python/dllib/src/setup.py#L91 https://github.com/intel-analytics/BigDL/blob/main/python/orca/src/bigdl/orca/common.py#L191

Done.

hkvision (approved):

Issue #5975, opened 2022-09-27
Nano : Performance test of ONNX Simplifier on some models and design for integrating with Nano


rnwang04:
# Background

[ONNX Simplifier](https://github.com/daquexian/onnx-simplifier) is presented to simplify the ONNX model. It infers the whole computation graph and then replaces the redundant operators with their constant outputs (a.k.a. constant folding).
Some projects like YOLOv5 have adpoted ONNX Simplifier as a supplement to ONNX Runtime.

We are considering integrating ONNX Simplifier into Nano's InferenceOptimizer. First and foremost , have to do some performance test of ONNX Simplifier.

# Performance
| -     | ResNet50  |  TCNForecaster  |  LSTMForecaster  |  efficientnet_b7 | swin |
| ----------- | ----------- |  ----------- |  ----------- |  ----------- |  ----------- | 
| onnx latency  |  9.2ms   |   0.42ms   |  0.215ms |  121.6ms |  111.2ms |  95.8ms | 
|  onnxsim latency |  9.2ms   |     **0.35ms**   | **0.117ms** |  **101.3ms** |   95.5ms |
| onnx size |  97.4M  |   151.7K  |  2.5K |  252.4M |  339M |
| onnxsim size |   97.4M   |    149.9k  | 1.4K | 252.4M |  339M |

#  Conclusion and Design
ONNX Simplifier does improve some models,  especially for LSTM-style model.
What's more, it's easy to integrate ONNX Simplifier.  We can make it as a default choice.

Only problem is this will introduce additional dependency `onnxsim` and modify corresponding workflow. `onnxsim` depends on `onnx` and `onnxruntime`.
So it should be considered if we should support onnxsim in openvino as it will increase three additional dependencies : `onnx`, `onnxruntime`, `onnxsim`
For onnxruntime, it will increase only one additional dependency `onnxsim`.

## Design 1
Apply onnxsim on both onnxruntime and openvino compulsively
```
from bigdl.nano.deps.onnxsim.onnxsim_api import onnx_simplify
onnx_simplify(onnx_path)
```
will be put at the end of https://github.com/intel-analytics/BigDL/blob/13a08a1eb103c9fe21224c646b1920155cc16494/python/nano/src/bigdl/nano/utils/inference/pytorch/model_utils.py#L98

## Design 2
Only apply onnxsim on onnxruntime compulsively
```
from bigdl.nano.deps.onnxsim.onnxsim_api import onnx_simplify
[...]

PR #5977, opened 2022-09-27
Nano : support ONNX Simplifier in Nano's inference

rnwang04:
## Description

We are considering integrating ONNX Simplifier into Nano's InferenceOptimizer.

### 1. Why the change?

https://github.com/intel-analytics/BigDL/issues/5975

### 2. User API changes

No change.

### 3. Summary of the change 

support onnxsim in Nano

### 4. How to test?
- [ ] Unit test
- [ ] Application test

### 5. New dependencies

- [x] New Python dependencies
       - onnxsim

### DIFF
--- a/python/nano/src/bigdl/nano/deps/onnxruntime/pytorch/pytorch_onnxruntime_model.py
+++ b/python/nano/src/bigdl/nano/deps/onnxruntime/pytorch/pytorch_onnxruntime_model.py
@@ -51,6 +51,9 @@ class PytorchONNXRuntimeModel(ONNXRuntimeModel, AcceleratedLightningModule):
             export_to_onnx(model, input_sample=input_sample, onnx_path='tmp.onnx',
                            **export_kwargs)
             onnx_path = 'tmp.onnx'
+            # simplify model
+            from bigdl.nano.deps.onnxsim.onnxsim_api import onnx_simplify
+            onnx_simplify(onnx_path)
         AcceleratedLightningModule.__init__(self, None)
         ONNXRuntimeModel.__init__(self, onnx_path, session_options=onnxruntime_session_options)
         if os.path.exists('tmp.onnx'):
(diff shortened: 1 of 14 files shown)


python/nano/src/bigdl/nano/deps/onnxruntime/pytorch/pytorch_onnxruntime_model.py
```diff
@@ -48,10 +48,12 @@ def __init__(self, model, input_sample=None, onnxruntime_session_options=None,
         onnx_path = model
         if isinstance(model, torch.nn.Module):
             # Typically, when model is fp32, we use this path
-            # TODO: expose ONNX export parameters to users
             export_to_onnx(model, input_sample=input_sample, onnx_path='tmp.onnx',
                            **export_kwargs)
             onnx_path = 'tmp.onnx'
+            # simplify model
+            from bigdl.nano.deps.onnxsim.onnxsim_api import onnx_simplify
+            onnx_simplify(onnx_path)
```
yangw1234, line 56: does it always work? 

How about adding a flag to disable it?

rnwang04: Yeah, if the conversion of onnx_simplify fails, onnx_path still save the original model.

I will add a flag to disable this.

TheaperDeng (approved):
LGTM

python/nano/src/bigdl/nano/deps/onnxruntime/pytorch/pytorch_onnxruntime_model.py
```diff
@@ -41,17 +41,24 @@ def __init__(self, model, input_sample=None, onnxruntime_session_options=None,
...
         if isinstance(model, torch.nn.Module):
             # Typically, when model is fp32, we use this path
-            # TODO: expose ONNX export parameters to users
             export_to_onnx(model, input_sample=input_sample, onnx_path='tmp.onnx',
                            **export_kwargs)
             onnx_path = 'tmp.onnx'
+            if simplification is True:
+                # simplify model
+                from bigdl.nano.deps.onnxsim.onnxsim_api import onnx_simplify
+                onnx_simplify(onnx_path)
```
TheaperDeng, line 61: add a try here and maybe catch all Exceptions (since both import and simplify error might happens)

Issue #5934, opened 2022-09-23
Chronos: add gen_synthetic_data API to provide users with offline dataset


plusbang:
## Background
Built-in dataset can be downloaded and preprocessed by `get_public_dataset`, but considering that users need dataset to benchmark, we may need to add a new API to help users generate dataset even with poor Internet connection.

## API design
```python
def gen_synthetic_data(len=10000, amplitude=10.0, angular_freq=0.01, scale=1.0,
                       time_freq="D", **kwargs):
# Generate dataset according to sine function with a Gaussian noise.
# Datetime is generated according to `time_freq` with the current time as endtime.
```
parameter:
 - len: int, the number indicates the dataset size. Default to 10000.
 - amplitude: float, the number indicates amplitude of the sine function. Default to 10.0.
 - angular_freq: float, the number indicates angular frequency of the sine function. Default to 0.01.
- scale: float, the number indicates the standard deviation of the Gaussian noise while the mean is set to 0. Default to 1.0.
- time_freq: str, the frequency of the generated dataframe, default to 'D'(calendar day frequency). The frequency can be anything from the pandas list of frequency strings here:
           https://pandas.pydata.org/docs/user_guide/timeseries.html#timeseries-offset-aliases
- kwargs: extra arguments passed to initialize the tsdataset, including with_split, val_ratio and test_ratio.

example:
```python
from bigdl.chronos.data import gen_synthetic_data
tsdata_gen = gen_synthetic_data()
```

PR #5983, opened 2022-09-27
Chronos: add `gen_synthetic_data` to provide users with offline dataset

plusbang:
## Description

Add the new API `gen_synthetic_data` to help users generate dataset to do some tests even with poor Internet connection.
The dataset is generated according to sine function with a Gaussian noise. And the datetime is generated according to `time_freq` with the current time as endtime.

### 1. Why the change?

https://github.com/intel-analytics/BigDL/issues/5934

### 2. User API changes

2.1. if users want to use `get_public_dataset`
```python
from bigdl.chronos.data import get_public_dataset
```
2.2. if users want to use `gen_synthetic_data` to generate TSDataset
```python
def gen_synthetic_data(len=10000, sine_amplitude=10.0, angular_freq=0.01,
                       noise_amplitude=0.01, noise_scale=1.0, seed=1,
                       time_freq="D", **kwargs)
```

```python
from bigdl.chronos.data import gen_synthetic_data
tsdata_gen = gen_synthetic_data()
tsdata_train_gen, tsdata_val_gen, tsdata_test_gen = gen_synthetic_data(with_split=True, val_ratio=0.1, test_ratio=0.1)
```

### 3. Summary of the change 

3.1. Add import in _init_.py
3.2. Add src and ut of `gen_synthetic_data`

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/chronos/src/bigdl/chronos/data/__init__.py
+++ b/python/chronos/src/bigdl/chronos/data/__init__.py
@@ -15,3 +15,4 @@
 #
 
 from .tsdataset import TSDataset
+from .repo_dataset import get_public_dataset, gen_synthetic_data
--- a/python/chronos/src/bigdl/chronos/data/repo_dataset.py
+++ b/python/chronos/src/bigdl/chronos/data/repo_dataset.py
@@ -13,14 +13,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+import numpy as np
+import pandas as pd
+import time
 from bigdl.chronos.data.utils.public_dataset import PublicDataset
+from bigdl.chronos.data.tsdataset import TSDataset
 
 
 def get_public_dataset(name, path='~/.chronos/dataset', redownload=False, **kwargs):
     """
     Get public dataset.
 
-    >>> from bigdl.chronos.data.repo_dataset import get_public_dataset
+    >>> from bigdl.chronos.data import get_public_dataset
     >>> tsdata_network_traffic = get_public_dataset(name="network_traffic")
 
     :param name: str, public dataset name, e.g. "network_traffic".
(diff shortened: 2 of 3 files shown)


python/chronos/src/bigdl/chronos/data/__init__.py
```diff
@@ -15,3 +15,4 @@
 #
 
 from .tsdataset import TSDataset
+from .repo_dataset import get_public_dataset, gen_synthetic_data
```
TheaperDeng, line 18: change the document as well
https://bigdl.readthedocs.io/en/latest/doc/Chronos/Overview/quick-tour.html
https://bigdl.readthedocs.io/en/latest/doc/Chronos/Overview/data_processing_feature_engineering.html#built-in-dataset

python/chronos/src/bigdl/chronos/data/repo_dataset.py
```diff
@@ -92,3 +96,36 @@ def get_public_dataset(name, path='~/.chronos/dataset', redownload=False, **kwar
                           "Only network_traffic, AIOps, fsi, nyc_taxi, uci_electricity"
                           " uci_electricity_wide"
                           f"are supported in Chronos built-in dataset, while get {name}.")
+
+
+def gen_synthetic_data(len=10000, **kwargs):
+    """
+    Generate dataset according to sine function y=10*sin(PI*x/1000).
```
TheaperDeng, line 103: try give more flexibility to the synthetic data, e.g. 10 and 1000 could be 2 parameters, a Gaussian noise can be added to the data as well (with another parameter indicates the intensity)

python/chronos/src/bigdl/chronos/data/repo_dataset.py
```diff
@@ -92,3 +96,36 @@ def get_public_dataset(name, path='~/.chronos/dataset', redownload=False, **kwar
...
+    """
+    from bigdl.nano.utils.log4Error import invalidInputError
+    invalidInputError(isinstance(len, int),
+                      f"The parameter len must be int, but found {type(len)}.")
+
+    gen_x = np.linspace(0, (len/1000)*np.pi, len)
+    gen_y = 10 * np.sin(gen_x)
+    df = pd.DataFrame(gen_y, columns=["target"])
+    endtime = time.strftime('%Y-%m-%d %H:%M:%S', time.localtime())
+    df.insert(0, "datetime", pd.date_range(end=endtime, periods=len, freq="T"))
```
TheaperDeng, line 123: freq could also be another parameter.

python/chronos/src/bigdl/chronos/data/repo_dataset.py
```diff
@@ -92,3 +96,36 @@ def get_public_dataset(name, path='~/.chronos/dataset', redownload=False, **kwar
...
+
+    if 'with_split' not in kwargs or not kwargs['with_split']:
+        tsdata = TSDataset.from_pandas(df, dt_col="datetime", target_col="target")
+        return tsdata
+    else:
+        tsdata_train, tsdata_val, tsdata_test = TSDataset.from_pandas(df, dt_col="datetime",
+                                                                      target_col="target", **kwargs)
+        return tsdata_train, tsdata_val, tsdata_test
```
TheaperDeng, line 131: what's the difference between just calling
```python
return TSDataset.from_pandas(df, dt_col="datetime", target_col="target", **kwargs)
```
here

python/chronos/src/bigdl/chronos/data/__init__.py
plusbang: After searching in vscode, there exists "from bigdl.chronos.data.repo_dataset import get_public_dataset" in other documents as well, maybe can create a new PR to change all the documents.

TheaperDeng: yep, we will do it in another PR.

python/chronos/src/bigdl/chronos/data/repo_dataset.py
```diff
@@ -92,3 +96,44 @@ def get_public_dataset(name, path='~/.chronos/dataset', redownload=False, **kwar
...
+
+    >>> from bigdl.chronos.data import gen_synthetic_data
+    >>> tsdata_gen = gen_synthetic_data()
+
+    :param len: int, the number indicates the dataset size. Default to 10000.
+    :param amplitude: float, the number indicates amplitude of the sine function.
+           Default to 10.0.
+    :param angular_freq: float, the number indicates angular frequency of the sine function.
+           Default to 0.01.
+    :param scale: float, the number indicates the standard deviation of the Gaussian noise
```
TheaperDeng, line 115: `noice_amplitude`?

python/chronos/src/bigdl/chronos/data/repo_dataset.py
```diff
@@ -92,3 +96,44 @@ def get_public_dataset(name, path='~/.chronos/dataset', redownload=False, **kwar
                           "Only network_traffic, AIOps, fsi, nyc_taxi, uci_electricity"
                           " uci_electricity_wide "
                           f"are supported in Chronos built-in dataset, while get {name}.")
+
+
+def gen_synthetic_data(len=10000, amplitude=10.0, angular_freq=0.01, scale=1.0,
+                       time_freq="D", **kwargs):
```
TheaperDeng, line 102: add a `seed` here? So that we could reproduce the same output.

plusbang: parameters seed and noise_amplitude are added.

TheaperDeng (approved):
LGTM

PR #5991, opened 2022-09-28
Lazy import BarrierTaskContext in tf spark runner

hkvision:
So that when use predict only, no need to have the constraint of BarrierTaskContext for Spark2.4+

Tested after the modification, spark 2.3.1 works.

@jason-dai @jenniew Take a look.

### DIFF
--- a/python/orca/src/bigdl/orca/learn/tf2/pyspark_estimator.py
+++ b/python/orca/src/bigdl/orca/learn/tf2/pyspark_estimator.py
@@ -350,7 +350,7 @@ class SparkTFEstimator():
             size=self.num_workers,
             model_weights=weights,
             mode="predict",
-            cluster_info=self._get_cluster_info(sc),
+            cluster_info=None,
             model_dir=self.model_dir,
             application_id=self.application_id,
             need_to_log_to_driver=self.need_to_log_to_driver,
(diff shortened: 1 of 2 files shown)


python/orca/src/bigdl/orca/learn/tf2/spark_runner.py
```diff
@@ -215,8 +211,10 @@ def __init__(self, model_creator, compile_args_creator,
         self.setup()
         self.cluster = cluster_info
         if mode == "fit" or mode == "evaluate":
+            from pyspark import BarrierTaskContext
             self.partition_id = BarrierTaskContext.get().partitionId()
         else:
+            from pyspark import TaskContext
```
jason-dai, line 217: Why can't we always use TaskContext to get partitionId?

hkvision: Seems we can?

hkvision: Tested, we can't do this.
When call rdd.barrier(), TaskContext.get() will return None, only BarrierTaskContext will return the actual instance.

python/orca/src/bigdl/orca/learn/tf2/pyspark_estimator.py
```diff
@@ -356,7 +350,7 @@ def predict(self, data, batch_size=None, verbose=1,
             size=self.num_workers,
             model_weights=weights,
             mode="predict",
-            cluster_info=self._get_cluster_info(sc),
+            cluster_info=None,
```
hkvision, line 353: Create cluster_info uses barrier mode for get ips and rank.
Since for predict cluster_info is not used, do not call this function, and input None instead.

Issue #5995, opened 2022-09-28
Nano: Hard to apply inference optimizations with accuracy control on YOLOX with little code modification


y199387:
## Description
YOLOX evaluate the model with [COCOAPI](https://github.com/cocodataset/cocoapi). The evaluation-related code like [this](https://github.com/Megvii-BaseDetection/YOLOX/blob/main/yolox/evaluators/coco_evaluator.py#L116):
```python
for cur_iter, (imgs, _, info_imgs, ids) in enumerate(
            progress_bar(self.dataloader)
        ):
            with torch.no_grad():
                imgs = imgs.type(tensor_type)

                # skip the last iters since batchsize might be not enough for batch inference
                is_time_record = cur_iter < len(self.dataloader) - 1
                if is_time_record:
                    start = time.time()

                outputs = model(imgs)
                if decoder is not None:
                    outputs = decoder(outputs, dtype=outputs.type())

                if is_time_record:
                    infer_end = time_synchronized()
                    inference_time += infer_end - start

                outputs = postprocess(
                    outputs, self.num_classes, self.confthre, self.nmsthre
                )
                if is_time_record:
                    nms_end = time_synchronized()
                    nms_time += nms_end - infer_end

            data_list_elem, image_wise_data = self.convert_to_coco_format(
                outputs, info_imgs, ids, return_outputs=True)
            data_list.extend(data_list_elem)
            output_data.update(image_wise_data)
```
It's hard to convert the code segment to a function like:
```python
def metric(preds, target):
    ...
    return accuracy
```
I need to subclass InferenceOptimizer to apply inference acceleration with accuracy control on YOLOX for now.
```python
from bigdl.nano.pytorch import InferenceOptimizer

def accuracy(model):
        coco_evaluator = exp.get_evaluator(args.batch_size, False)
        coco_evaluator.per_class_AP = True
        coco_evaluator.per_class_AR = True

        ap50_95, ap50, _ = coco_evaluator.evaluate(model, False)
        return ap50



    inference_optimizer = YoloxInferenceOptimizer()
    inference_optimizer.optimize(model, metric=accuracy ,training_data=train_loader)
    inference_optimizer.summary()

class YoloxInferenceOptimizer(InferenceOptimizer):
    def optimize(...):
       super().optimize(...)
       if metric:
            for method, acce_result in self.optimized_model_dict.items():
                with torch.no_grad():
                    result = metric(acce_result["model"], **metric_kwargs)
[...]

PR #6002, opened 2022-09-28
Nano : Enhancements for InferenceOptimizer to support YOLOX

rnwang04:
## Description
To support applying InferenceOptimizer on YOLOX.

### 1. Why the change?

Based on https://github.com/intel-analytics/BigDL/issues/5995, we want to extend `metric` in `InferenceOptimizer.optimize` to support more cases.

### 2. User API changes

Before API supports torchmetrics.Metric object or similar callable object which which takes prediction and target as input.
case1 : torchmetrics.Metric object
```python
                       training_data=train_loader,
                       validation_data=val_loader,
                       metric=torchmetrics.Accuracy(),
                       direction="max",
                       thread_num=1)
```
case2：similar callable object which which takes prediction and target as input. related example: https://github.com/intel-analytics/BigDL/blob/main/python/nano/example/pytorch/inference_optimizer/resnet/inference_pipeline.py
```python
def metric(pred, target):
      value = process(pred, target)
      return value

                       training_data=train_loader,
                       validation_data=val_loader,
                       metric=metric,
                       direction="max",
                       thread_num=1)
```
Now, two additional metric inputs are supported.
case3: A callable object that takes model and validation_data as input
```python
def metric(model, data_loader):
      value = process(model, data_loader)
      return value

                       training_data=train_loader,
                       validation_data=val_loader,
                       metric=metric,
                       direction="max",
                       thread_num=1)
```
case4: A callable object that takes model as input, here validation_data is None, related example: https://github.com/intel-analytics/BigDL/issues/5995
```python
def metric(model):
      value = process(model)
      return value

                       training_data=train_loader,
                       validation_data=None,
                       metric=metric,
                       direction="max",
                       thread_num=1)
```

### 3. Summary of the change

- [x] Extend metric to support more cases
- [x] Add more unittests to test different kinds of metric

### 4. How to test?
- [x] Unit test
- [x] Application test

Inference result of YOLOX：
image

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
+++ b/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
@@ -34,7 +34,9 @@ from bigdl.nano.deps.onnxruntime.onnxruntime_api import PytorchONNXRuntimeModel
 from bigdl.nano.deps.neural_compressor.inc_api import quantize as inc_quantize
 from bigdl.nano.utils.inference.pytorch.model import AcceleratedLightningModule
 from bigdl.nano.utils.inference.pytorch.model_utils import get_forward_args, get_input_example
+from bigdl.nano.utils.inference.pytorch.metrics import NanoMetric
 from bigdl.nano.pytorch.utils import TORCH_VERSION_LESS_1_10, save_model, load_model
+from torchmetrics import Metric
 import warnings
 # Filter out useless Userwarnings
 warnings.filterwarnings('ignore', category=UserWarning, module='pytorch_lightning')
@@ -140,14 +142,12 @@ class InferenceOptimizer:
         :param validation_data: (optional) A pytorch dataloader for accuracy evaluation
                This is only needed when users care about the possible accuracy drop.
         :param metric: (optional) A callable object which is used for calculating accuracy.
-               1. If validation_data is composed of (input_data, target), metric could be
-               a torchmetrics.metric.Metric or similar object or function which takes
-               prediction and target then returns an accuracy value in this
-               calling method `accuracy = metric(pred, target)`.
-               2. Otherwise you can define a custom metric in this calling
-               method `accuracy = metric(model, batch_data)`.
-               3. Metric can even only recieve model as input in this calling
-               method `accuracy = metric(model)`, like coco_evaluator.
+               It supports two kinds of callable object:
+               1. A torchmetrics.Metric object takes prediction and target and returns
+               an accuracy value in this calling method `metric(pred, target)`
+               2. A callable object that takes model and data_loader as input and returns
+               an accuracy value in this calling method
+               `accuracy = metric(model, validation_data)`.
         :param direction: (optional) A string that indicates the higher/lower
                better for the metric, "min" for the lower the better and "max" for the
                higher the better. Default value is "max".
@@ -339,6 +339,9 @@ class InferenceOptimizer:
                        use_ipex: bool = None,
                        accuracy_criterion: float = None) -> Tuple[nn.Module, str]:
         '''
+        According to results of `optimize`, obtain the model with minimum latency under
+        specific restrictions or without restrictions.
+
         :param accelerator: (optional) Use accelerator 'None', 'onnxruntime',
                'openvino', 'jit', defaults to None. If not None, then will only find the
                model with this specific accelerator.
@@ -771,23 +774,9 @@ def _accuracy_calculate_helper(model, metric, data):
     '''
     A quick helper to calculate accuracy
     '''
-    with torch.no_grad():
-        if data is not None:
-            try:
-                # for normal cases, assume validation data is composed of input_data and target
-                metric_list = []
-                sample_num = 0
-                for i, (data_input, target) in enumerate(data):
-                    metric_list.append(
-                        metric(model(data_input), target).numpy() * data_input.shape[0])
-                    sample_num += data_input.shape[0]
-                return np.sum(metric_list) / sample_num
-            except Exception:
-                # metric itself defines how to calculate accuracy
-                return metric(model, data)
-        else:
-            # special cases, i.g. coco_evaluator, which only recieves model as input
-            return metric(model)
+    if isinstance(metric, Metric):
+        metric = NanoMetric(metric)
+    return metric(model, data)
 
 
 def _format_acceleration_option(method_name: str) -> str:
--- a/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
+++ b/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
@@ -302,7 +302,8 @@ class InferenceOptimizer:
                                 result_map[method]["accuracy"] =\
                                     _accuracy_calculate_helper(acce_model, metric,
                                                                validation_data)
-                            except Exception:
+                            except Exception as e:
+                                print(e)
                                 self._calculate_accuracy = False
                                 invalidInputError(
                                     False,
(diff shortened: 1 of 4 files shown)


python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -139,8 +139,15 @@ def optimize(self, model: nn.Module,
...
-               and returns a accuracy value in this calling method `metric(pred, target)`
+        :param metric: (optional) A callable object which is used for calculating accuracy.
+               1. If validation_data is composed of (input_data, target), metric could be
+               a torchmetrics.metric.Metric or similar object or function which takes
+               prediction and target then returns an accuracy value in this
+               calling method `accuracy = metric(pred, target)`.
+               2. Otherwise you can define a custom metric in this calling
+               method `accuracy = metric(model, batch_data)`.
+               3. Metric can even only recieve model as input in this calling
+               method `accuracy = metric(model)`, like coco_evaluator.
```
jason-dai, line 150: The logic seems a little bit messy. Maybe it's simpler if we just require the metric be a callable object that takes model and data_loader (if not None) as input.

We can provide a helper class such as `NanoMetric` to help user compute the result if the user's metric takes (input_data, target), similar to what `_accuracy_calculate_helper` does; the user can then simply wrap their metric using `NanoMetric`

In addition, we can also automatically wrap the metric using `NanoMetric` inside `optimize` if it is a `TorchMetrics` object.

rnwang04: Thanks for such great suggestions. I have modified code according to these comments.  Could you please take a look again to confirm that whether the current implementation is consistent with your vision ? @jason-dai 

Another question is, here I just ask user to provide a callable object that takes model and dataloader as input. Should I impose certain restrictions on this behavior, for example, given a Base class and ask user to inherit this class to implement custom logic?
```python
class BaseMetric(metaclass=ABCMeta):
    def __init__(self, metric: Callable):
        self.metric = metric
    
    @abstractmethod
    def __call__(self, model: nn.Module, data_loader: DataLoader = None):
        pass

class CustomMetric(BaseMetric):
    def __init__(self, metric: Callable):
        self.metric = metric
    
    def __call__(self, model: nn.Module, data_loader: DataLoader = None):
        # custom logic
        pass

metric = Accuracy()
metric = CustomMetric(metric)
```

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -285,9 +295,22 @@ def func_test(model, input_sample):
...
+                        if method == "original":
+                            # test whether metric works
+                            try:
+                                result_map[method]["accuracy"] =\
+                                    _accuracy_calculate_helper(acce_model, metric,
+                                                               validation_data)
+                            except Exception:
+                                self._calculate_accuracy = False
+                                invalidInputError(False,
+                                                  "Your metric is incompatible with "
```
jason-dai, line 307: explain what's the expected metric pattern in the error message

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -750,17 +776,35 @@ def _throughput_calculate_helper(iterrun, baseline_time, func, *args):
...
+    import inspect
+    sig = inspect.signature(function)
+    if len(sig.parameters.values()) != 2:
+        return False
+    param1_name = list(sig.parameters.values())[0].name
+    param2_name = list(sig.parameters.values())[0].name
+    if "pred" in param1_name or "pred" in param2_name:
+        return True
+    if "target" in param2_name or "target" in param2_name:
+        return True
```
jason-dai, line 793: does this make sense?

rnwang04: here we assume user's custom metric function is like this ` function_name(pred=xxx, target=xxx)`, a example is : https://github.com/intel-analytics/BigDL/blob/main/python/nano/example/pytorch/inference_optimizer/resnet/inference_pipeline.py

I think we have to limit the function signature here, otherwise we can't distinguish metric(pred, target) from metric (model, data_loader).

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -139,8 +141,15 @@ def optimize(self, model: nn.Module,
...
-               and returns a accuracy value in this calling method `metric(pred, target)`
+        :param metric: (optional) A callable object which is used for calculating accuracy.
+               It supports two kinds of callable object:
+               1. A torchmetrics.Metric object or similar callable object which takes
+               prediction and target than returns an accuracy value in this calling
+               method `metric(pred, target)`. This requires data in validation_data
+               is composed of (input_data, target).
+               2. A callable object that takes model and validation_data(which could
+               be None) as input and returns an accuracy value in this calling method
+               `accuracy = metric(model, data_loader)`.
```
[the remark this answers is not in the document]
rnwang04: done

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: done

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -750,17 +784,35 @@ def _throughput_calculate_helper(iterrun, baseline_time, func, *args):
...
+    import inspect
+    sig = inspect.signature(function)
+    if len(sig.parameters.values()) != 2:
+        return False
+    param1_name = list(sig.parameters.values())[0].name
+    param2_name = list(sig.parameters.values())[0].name
+    if "pred" in param1_name:
+        return True
+    if "target" in param2_name:
+        return True
```
jason-dai, line 801: `if "pred" in param1_name and "target" in param2_name`, yes?

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -750,17 +784,35 @@ def _throughput_calculate_helper(iterrun, baseline_time, func, *args):
...
-    metric_list = []
-    sample_num = 0
-    with torch.no_grad():
-        for i, (data_input, target) in enumerate(data):
-            metric_list.append(metric(model(data_input), target).numpy() * data_input.shape[0])
-            sample_num += data_input.shape[0]
-    return np.sum(metric_list) / sample_num
+    if isinstance(metric, Metric) or _signature_check(metric) is True:
+        metric = NanoMetric(metric)
+        return metric(model, data)
```
jason-dai, line 811: what if `data` is None?

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -750,17 +784,35 @@ def _throughput_calculate_helper(iterrun, baseline_time, func, *args):
...
     A quick helper to calculate accuracy
     '''
-    metric_list = []
-    sample_num = 0
-    with torch.no_grad():
-        for i, (data_input, target) in enumerate(data):
-            metric_list.append(metric(model(data_input), target).numpy() * data_input.shape[0])
-            sample_num += data_input.shape[0]
-    return np.sum(metric_list) / sample_num
+    if isinstance(metric, Metric) or _signature_check(metric) is True:
```
jason-dai, line 809: What if data is None?

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: will change to this.

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: will add ` invalidInputError(data is not None,xxx)` for this case.
Error message will look like this:
image
Do you think it's ok ?

jason-dai: ok

Issue #4410, opened 2022-04-13
Orca do not support preprocess inputs before training in forward()


zhanglirong1999:
I rewrite yolox's code to run on orca as distributed, but errors were reported when forward(). Because orca directly use  the input to forward() after the data loader, resulting in the inputs not being supported in yolox forward(). Yolox  would preprocess input and target after the data loader 

Is it necessary for orca to support some special preprocess before training?

zhanglirong1999:
This is the link that orca directly call the forward(). But the input exceeds the size of the forward()

[](https://github.com/intel-analytics/BigDL/blob/main/python/orca/src/bigdl/orca/learn/pytorch/training_operator.py#L282)

zhanglirong1999:
This is the link that orca directly call the forward(). But the input exceeds the size of the forward()
[https://github.com/intel-analytics/BigDL/blob/main/python/orca/src/bigdl/orca/learn/pytorch/training_operator.py#L282](url)

zhanglirong1999:
This is the link that orca directly call the forward(). But the input exceeds the size of the forward() in yolox

https://github.com/intel-analytics/BigDL/blob/main/python/orca/src/bigdl/orca/learn/pytorch/training_operator.py#L282

zhanglirong1999:
in yolox, there is some preprocess on input 

https://github.com/Megvii-BaseDetection/YOLOX/blob/main/yolox/core/trainer.py#L100

hkvision:
Probably we can add an argument for users to specify a preprocess function?

zhanglirong1999:
The original yolox GPU version used a prefetcher, throwing away two inputs and only retaining two inputs. This is also a part of its data preprocessing.
https://github.com/Megvii-BaseDetection/YOLOX/blob/7ba9fd298413b79f2d07b6b7a3cdcbc34a9b244d/yolox/data/data_prefetcher.py#L25

PR #6004, opened 2022-09-28
Orca:  Customized collate_fn in dataloader

leonardozcm:
### 1. Why the change?

For third-party dataloader implementations, we can use a `collate_fn` wrapper to implement custom data processing like elements alignment:
```
def reduceWrapper(func):
    def reduceIterElement(batch):
        batch=func(batch)
        assert len(batch)==5, "Should yield inputs, labels, index, time, meta in dataloader"
        return batch[0], batch[1]
    return reduceIterElement

def train_loader_creator(config, batch_size):
    train_loader = loader.construct_loader(config, "train")
    loader.shuffle_dataset(train_loader, 0)
    train_loader.collate_fn = reduceWrapper(train_loader.collate_fn)
    return train_loader
```
It should be more natural than just integrade their code into users' implementation.

~~And it can also collate elements when `data` is dataframe or Xshards for generic usages.~~

https://github.com/intel-analytics/BigDL/issues/4410

### 2. User API changes

We add a new function type argument called `recollate_fn` to estimator.fit(). 

`recollate_fn` takes a batch of dataloader as input and ouput processed batch:
```
def customized_collate_fn(batch):
       *features, labels = batch
       new_features = process_features(features)
       new_labels = process_labels(labels)
       return new_features, new_labels
```

When a dataloader is passed to fit(), it's original `collate_fn` will be wrapped around with `recollate_fn`

By default `recollate_fn` does nothing.

### 3. Summary of the change 

Enable `recollate_fn` in bigdl, ray and spark backend.  

Before this change if we need to collate dataloader manually:
```
def reduceWrapper(func):
    def reduceIterElement(batch):
        batch=func(batch)
        print("reduce called")
        assert len(batch)==5, "Should yield inputs, labels, index, time, meta in dataloader"
        return batch[0], batch[1]
    return reduceIterElement

def train_loader_creator(config, batch_size):
    train_loader = loader.construct_loader(config, "train")
    loader.shuffle_dataset(train_loader, 0)
    train_loader.collate_fn = reduceWrapper(train_loader.collate_fn)
    return train_loader

def validation_data_creator(config, batch_size):
      val_loader = loader.construct_loader(config,"val")
      val_loader.collate_fn = reduceWrapper(val_loader.collate_fn)
      return val_loader

orca_estimator.fit(data=train_loader_creator,
                               validation_data=validation_data_creator,
                               batch_size=cfg.TRAIN.BATCH_SIZE,
                               epochs=cfg.SOLVER.MAX_EPOCH)
```

now:
```
def reduceIterElement(batch):
    print("reduce called")
    assert len(batch)==5, "Should yield inputs, labels, index, time, meta in dataloader"
    return batch[0], batch[1]

orca_estimator.fit(data=train_loader_creator,
                       validation_data=validation_data_creator,
                       batch_size=cfg.TRAIN.BATCH_SIZE,
                       epochs=cfg.SOLVER.MAX_EPOCH,
                       recollate_fn=reduceIterElement)
```

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
@@ -25,7 +25,7 @@ from bigdl.orca.learn.pytorch.training_operator import TrainingOperator
 from bigdl.orca.learn.pytorch.pytorch_pyspark_worker import PytorchPysparkWorker
 from bigdl.orca.learn.utils import maybe_dataframe_to_xshards, dataframe_to_xshards, \
     convert_predict_xshards_to_dataframe, make_data_creator, update_predict_xshards, \
-    reload_dataloader_creator
+    reload_dataloader_creator, empty_recollate_fn
 from bigdl.orca.data import SparkXShards
 from bigdl.orca import OrcaContext
 from bigdl.orca.learn.base_estimator import BaseEstimator
@@ -39,7 +39,7 @@ from bigdl.dllib.utils.utils import get_node_ip
 from bigdl.dllib.utils.log4Error import *
 
 
-def partition_to_creator(partition):
+def partition_to_creator(partition, collate_fn=empty_recollate_fn):
 
     def data_creator(config, batch_size):
         from bigdl.orca.data.utils import ray_partition_get_data_label, index_data, get_size
@@ -67,7 +67,7 @@ def partition_to_creator(partition):
                                                    allow_list=False)
         print("Data size on worker: ", len(label))
         dataset = NDArrayDataset(data, label)
-        data_loader = DataLoader(dataset, **params)
+        data_loader = DataLoader(dataset, **params, collate_fn=collate_fn)
         return data_loader
 
     return data_creator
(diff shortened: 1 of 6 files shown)


python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
```diff
@@ -186,7 +186,8 @@ def fit(self,
             feature_cols=None,
             label_cols=None,
             validation_data=None,
-            callbacks=[]):
+            callbacks=[],
+            recollate_fn=empty_recollate_fn):
```
jason-dai, line 190: shall simply call it `collate_fn`? @hkvision

hkvision: Yes, I think `collate_fn` is better.

hkvision:
Is the current behavior like this? If the data creator returns a DataLoader, then collate_fn in fit won't take effect since user is recommended to add it in its own dataloader.

leonardozcm:
> Is the current behavior like this? If the data creator returns a DataLoader, then collate_fn in fit won't take effect since user is recommended to add it in its own dataloader.

I suppose not, `recollate_fn` is meant to align differences between dataloader and model from highly integrated codebase(slowfast for example)

That is to say, if the model or dataloader defined is on the user side, it is recommended to add it in their own code, but if it is not that convenient  to modify definition code, we  provide the user with one more chance to collate, that is why call it `re-collate`.

leonardozcm:
Current logic is like this:
1. For user's dataloader, we wrap its `collate_fn` with `recollate_fn`:
```
data.collate_fn = dataloader_collate_wrapper(data.collate_fn, recollate_fn)
```
2. For user's dataframe/Xshard/Raydataset, `recollate_fn` will be applied on the dataloader which is constructed automatically。

PR #6016, opened 2022-09-29
Orca: Add requirements

lalalapotter:
## Description

Add requirement_orca.txt for Orca for environment installation.


python/orca/src/setup.py
```diff
@@ -91,7 +91,9 @@ def get_bigdl_packages():
             print("including", package)
     return bigdl_packages
 
-
+# The requirement list contains the main dependencies for users of orca. This list
+# should be carefully curated. If you change it, please reflect
+# the change in the matching file of requirements/orca.
```
hkvision, line 96: python/requirements/orca

python/requirements/orca/requirement_extra.txt
```diff
@@ -0,0 +1,14 @@
...
+boto3
+cloudpickle
+fsspec
+onnx
+onnxruntime
+Pillow
+py4j
+pyarrow
+tensorflow
+tensorflow_estimator
```
hkvision, line 11: will tensorflow_estimator be installed as a dependency of tensorflow?

lalalapotter: I think so, do we need to remove all the dependencies of tensorflow and pytorch?

python/orca/src/setup.py
lalalapotter: fixed

python/requirements/orca/requirement_base.txt
```diff
@@ -0,0 +1,11 @@
+# These are mirrored in python/orca/src/setup.py and python/dllib/src/setup.py 
+# as install_requires, which is what the users of the orca package will install.
+#
+# In short, if you change it here, PLEASE also change it in setup.py.
```
lalalapotter, line 4: shall we add this comment in each requirement?

hkvision: Maybe in base/automl and ray?

python/requirements/orca/requirement_extra.txt
hkvision: In this file, any other dependencies of tensorflow and pytorch?

python/requirements/orca/requirement_base.txt
lalalapotter: Sure.

python/requirements/orca/requirement_extra.txt
lalalapotter: No, but in requirement_base.txt， `packaging`, `numpy`, `six` are also in the requirements of tensorflow.

hkvision: Then it doesn't matter. Even if we don't install tensorflow, we have to install numpy and six.

hkvision (approved):

Issue #5896, opened 2022-09-21
Chronos: from_prometheus api for tsdataset


plusbang:
**This issue is draft version**

## Background

TSDataset currently can be initialized from pandas dataframe and path of parquet file through `from_pandas` and `from_parquet` while some users have adopted Prometheus to collect and store time series data. Maybe we can provide with new API to initialize tsdataset(s) from Prometheus data for specified time period via url.

## Methodology

Utilize prometheus-pandas package to initialize tsdataset(s) from Prometheus.

## API Design
```python
static from_prometheus(prometheus_url, query, starttime, endtime, step, dt_col=None, target_col=None, id_col=None, 
                       extra_feature_col=None, with_split=False, val_ratio=0, test_ratio=0.1, **kwargs)              
```

Parameters: 

· prometheus_url - str, URL of Prometheus server.

· query - str, Prometheus expression query string.

· starttime - Start timestamp of the time period, RFC-3339 string or as a Unix timestamp in seconds.

· endtime - End timestamp of the time period, RFC-3339 string or as a Unix timestamp in seconds.

· step - Query resolution step width in Prometheus duration format or float number of seconds.

(Refer to https://prometheus.io/docs/prometheus/latest/querying/basics/#time-durations for more information about Prometheus duration.)

· dt_col, target_col - Automatically obtained according to the dataframe generated from Prometheus data.

· **kwargs: timeout (Evaluation timeout, Prometheus duration string)

example:
```python
tsdataset = TSDataset.from_prometheus(prometheus_url='http://10.105.191.164:9090/',
                                      query='collectd_cpufreq{cpufreq="0"}',
                                      starttime='2022-09-01T00:00:00Z',
                                      endtime='2022-09-05T00:00:00Z',
                                      step='1s',
                                      ...)
```

PR #6020, opened 2022-09-29
Chronos: add `from_prometheus` to initialize tsdataset(s) from Prometheus data

plusbang:
## Description

Provide with new API to initialize tsdataset(s) from Prometheus data for specified time period via url.

### 1. Why the change?

https://github.com/intel-analytics/BigDL/issues/5896

### 2. User API changes

```python
def from_prometheus(prometheus_url, query, starttime, endtime, step,
                    target_col=None, id_col=None, extra_feature_col=None,
                    with_split=False, val_ratio=0, test_ratio=0.1, **kwargs):
```

Example:
```python
tsdata = TSDataset.from_prometheus(prometheus_url='http://10.105.191.163:30000/',
                                   query='collectd_cpufreq{cpufreq="0"}',
                                   starttime="2022-09-27T00:00:00Z",
                                   endtime="2022-09-28T00:00:00Z",
                                   step="1m")
```

### 3. Summary of the change 
- [x] Generate dataframe from Prometheus data.
- [x] If columns are not explicitly stated, `target_col` is automatically specified according to the Prometheus data, `id_col` is interpreted as only containing a single id and `extra_feature_col` is None.
- [x] Users can also customize columns.

### 4. How to test?
- [ ] Unit test
(Considering the inconvenience of obtaining data from Prometheus server, the API is only tested locally.)

### DIFF
--- a/python/chronos/src/bigdl/chronos/data/tsdataset.py
+++ b/python/chronos/src/bigdl/chronos/data/tsdataset.py
@@ -213,6 +213,79 @@ class TSDataset:
                                      val_ratio=val_ratio,
                                      test_ratio=test_ratio)
 
+    @staticmethod
+    def from_prometheus(prometheus_url,
+                        query,
+                        starttime,
+                        endtime,
+                        step,
+                        target_col=None,
+                        id_col=None,
+                        extra_feature_col=None,
+                        with_split=False,
+                        val_ratio=0,
+                        test_ratio=0.1,
+                        **kwargs):
+        """
+        Initialize tsdataset(s) from Prometheus data for specified time period via url.
+
+        :param prometheus_url: a str indicates url of a Prometheus server.
+        :param query: str, Prometheus expression query string.
+        :param starttime: start timestamp of the specified time period, RFC-3339 string
+               or as a Unix timestamp in seconds.
+        :param endtime: end timestamp of the specified time period, RFC-3339 string
+               or as a Unix timestamp in seconds.
+        :param step: a str indicates query resolution step width in Prometheus duration format
+               or float number of seconds. More information about Prometheus time durations
+               are here:
+               https://prometheus.io/docs/prometheus/latest/querying/basics/#time-durations
+        :param target_col: (optional) a Prometheus expression query str or list indicates the
+               col name of target column in the input data frame. If it is not explicitly stated,
+               then target column is automatically specified according to the Prometheus data.
+        :param id_col: (optional) a Prometheus expression query str indicates the col name of
+               dataframe id. If it is not explicitly stated, then the data is interpreted as
+               only containing a single id.
+        :param extra_feature_col: (optional) a Prometheus expression query str or list indicates
+               the col name of extra feature columns that needs to predict the target column.
+               If it is not explicitly stated, then extra feature column is None.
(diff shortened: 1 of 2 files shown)


python/chronos/src/bigdl/chronos/data/utils/prometheus_df.py
```diff
@@ -0,0 +1,193 @@
...
+    _check_type(endtime, "endtime", (str, float))
+    _check_type(step, "step", (str, float))
+
+    # Generate dateframe
+    Pro_client = Prometheus(prometheus_url)
+    Pro_df = Pro_client.query_range(query, starttime, endtime, step, **kwargs)
+    df = pd.DataFrame(columns=Pro_df.columns.tolist())
+    df.insert(0, "datetime", Pro_df.index)
+    for col in Pro_df.columns.tolist():
+        df[col] = Pro_df[col].values
```
TheaperDeng, line 170: preferably, we don't use capital latter for variables (but we don't need to change it now)

python/chronos/src/bigdl/chronos/data/utils/prometheus_df.py
```diff
@@ -0,0 +1,193 @@
...
+                output_columns[col] = columns[col][0]  # id_col is str
+            else:
+                output_columns[col] = columns[col]
+        if output_columns[col] is not None:
+            if isinstance(output_columns[col], list):
+                output_col_list = output_col_list + output_columns[col]
+            else:
+                output_col_list.append(output_columns[col])
+
+    df = df.loc[:, output_col_list]
```
TheaperDeng, line 192: does `output_col_list` contains user input `extra_feature_col`?

python/chronos/src/bigdl/chronos/data/tsdataset.py
```diff
@@ -213,6 +213,79 @@ def from_parquet(path,
...
+        df, df_columns = GetRangeDataframe(prometheus_url, query, starttime, endtime,
+                                           step, columns=columns, **kwargs)
+        return TSDataset.from_pandas(df,
+                                     dt_col=df_columns["dt_col"],
+                                     target_col=df_columns["target_col"],
+                                     id_col=df_columns["id_col"],
+                                     extra_feature_col=df_columns["extra_feature_col"],
+                                     with_split=with_split,
+                                     val_ratio=val_ratio,
+                                     test_ratio=test_ratio)
```
TheaperDeng, line 287: we may add some white box check (not in this PR).

plusbang: I think it's a good idea, we can try it after this PR

python/chronos/src/bigdl/chronos/data/utils/prometheus_df.py
plusbang: yes, the `output_col_list` is confirmed according to `output_columns`. `output_columns[extra_feature_col]` defaults to be None and will be updated if user specified `extra_feature_col`.

TheaperDeng: okay I see, yeah it should be correct

TheaperDeng (approved):

PR #6039, opened 2022-10-07
Support some xshards operations and add one tutorial

dding3:
## Description
Support below operations in shards.
> sort_values
> max_values
> get_null_sum
> drop_missing_value

and add a shards tutorials


python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -419,10 +418,98 @@ def deduplicates(self):
...
+        """
+        Get the max values of the column name. This is only applicable for SparkXShards
+         of Pandas Series.
+
+        :param col_name column name that need return the max value
+        :return: maximum for the specified columns.
+        """
+        if self._get_class_name() == 'pandas.core.frame.DataFrame':
+            import pandas as pd
+            df = self.to_spark_df()
```
jason-dai, line 453: do we need to convert to spark df here? This can be simply implemented as a reduce function on RDD, yes?

python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -419,10 +418,98 @@ def deduplicates(self):
...
+                              "Currently only support assembleFeatureLabelCols() on"
+                              " XShards of Pandas DataFrame")
+
+        def get_na_sum(df):
+            import pandas as pd
+            series = df.isnull().sum()
+            df = pd.DataFrame({'col': series.index, 'total': series.values})
+            return df
+
+        na_cnt_shard = self.transform_shard(get_na_sum)
```
jason-dai, line 480: use `rdd.map_partition` instead of `transform_shard` to reduce unnecessary caching

python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -419,10 +418,98 @@ def deduplicates(self):
...
+       :return: a new SparkXShards without null values
+        """
+        if self._get_class_name() != 'pandas.core.frame.DataFrame':
+            invalidInputError(False,
+                              "Currently only support assembleFeatureLabelCols() on"
+                              " XShards of Pandas DataFrame")
+
+        null_cnt_shard = self.get_null_sum()
+        missing_data_shards = null_cnt_shard.sort_values(col_names="total", ascending=False)
+        zip_shards = self.zip(missing_data_shards)
```
jason-dai, line 502: can you zip here? after sort, the new partitions do not match the original ones?

python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -419,10 +418,98 @@ def deduplicates(self):
...
+        
+        :param col_names list of column or column naames to sort by
+        :param ascending bool, default True. Specify sort orders
+        :return: a new SparkXShards sorted by the specified columns.
+        """
+        if self._get_class_name() == 'pandas.core.frame.DataFrame':
+            import pandas as pd
+            df = self.to_spark_df()
+            sort_df = df.sort(col_names, ascending=ascending)
+            sort_df = sort_df.coalesce(1)
```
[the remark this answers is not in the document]
dding3: It's used to ensure there is no empty partition. However it's not a good solution. Have updated the code.

python/orca/src/bigdl/orca/data/shard.py
dding3: Yes and have used reduce function to get the max value.

python/orca/src/bigdl/orca/data/shard.py
dding3: Fixed.

python/orca/src/bigdl/orca/data/shard.py
dding3: Updated the code to ensure the new partitions matches the original one.

PR #6050, opened 2022-10-09
Chronos: fix ImportError with pytest.mark.tf2

plusbang:
## Description

In order to easily test uts corresponding to different installation options, uts are marked with labels.
But there exists ImportError with op_tf2, the error is fixed.


TheaperDeng (approved):
LGTM

PR #6052, opened 2022-10-09
Nano: add onnx support for tensorflow

MeouSker77:
## Description

Add onnx support for tensorflow in nano.

### 2. User API changes

User can write the following code to use ONNX acceleration now:

```python
from bigdl.nano.tf.keras import Model

model = Model(...)
spec = tf.TensorSpec(input_shape, dtype)
onnx_model = model.trace(accelerator='onnxruntime', input_sample=spec)
y_hat = onnx_model(x)
```

### 3. Summary of the change 

Add ONNX support for tensorflow in bigdl-nano, and add UT for it.

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...


python/nano/src/bigdl/nano/deps/onnxruntime/tensorflow/tensorflow_onnxruntime_model.py
```diff
@@ -0,0 +1,64 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+import os
+import tensorflow as tf
+import tf2onnx
```
TheaperDeng, line 20: add importerror massage to users?

python/nano/test/onnx/tf/test_onnx.py
```diff
@@ -0,0 +1,43 @@
...
+        onnx_model = model.trace(accelerator='onnxruntime', input_sample=spec)
+
+        y_hat = onnx_model(train_examples[:10])
+        assert y_hat.shape == (10, 10)
+
+        y_hat = onnx_model.predict(train_examples, batch_size=5)
+        assert y_hat.shape == (100, 10)
+
+        onnx_model.compile(metrics=[tf.keras.metrics.CategoricalAccuracy()])
+        acc = onnx_model.evaluate(train_dataset, return_dict=True)['categorical_accuracy']
```
TheaperDeng, line 43: typically, onnx model and tf model should have same acc / predict output, we may check it here.

TheaperDeng (approved):
LGTM

jason-dai:
Maybe also add an example in the tutorials

TheaperDeng:
> Maybe also add an example in the tutorials

we plan to add both openvino and onnxruntime examples on keras later in separate PRs

PR #6057, opened 2022-10-10
Nano: update examples in `nano/tutorial/training/pytorch` to `from bigdl.nano.pytorch import TorchNano`

MeouSker77:
## Description

update examples in `nano/tutorial/training/pytorch` to `from bigdl.nano.pytorch import TorchNano`


jason-dai:
Need to update all examples in `nano/tutorial/training/pytorch/`

PR #6061, opened 2022-10-10
Chronos: make bigdl-chronos[tensorflow] option available

plusbang:
## Description

After previous tests and bug fix, tensorflow backend can be used alone without pytorch in Chronos.
The [tensorflow] installation option is available now.


TheaperDeng (approved):
LGTM

Issue #6023, opened 2022-09-30
Chronos : low quality data detection and repair API discussion


rnwang04:
After detecting the low-quality data, we may want to automatic repair part of them according to our experience.
```python
def quality_check_and_repair_timeseries_dataframe(df, dt_col, repair=False):
    df_res = df.copy(deep=True)
    # 1. timestamp check
    if _timestamp_type_check(df[dt_col]) is False:
        if repair is True and xxx_repair(df_res, dt_col) is False: # inplace modify
            return False, df

    # 2. irregular interval check
    if _time_interval_check(df[dt_col]) is False:
        if repair is True and resample(df_res, dt_col) is False:
            return False, df

    # 3. missing value check
    if _missing_value_check(df, dt_col) is False:
        if repair is True and inpute(df_res, dt_col) is False:
            return False, df

    # 4. pattern check and noise check
    if _pattern_check(df, dt_col) is False:
        return False, df # can't repair

    return True, df_col
```

rnwang04:
@TheaperDeng How do you think of this?

PR #6071, opened 2022-10-11
Chronos: automatic repair low quality data after detection

rnwang04:
## Description

After detecting the low-quality data, we may want to automatic repair part of them according to our experience.

### 1. Why the change?

https://github.com/intel-analytics/BigDL/issues/6023

### 2. User API changes
a new parameter `repair` is supported.
```python
from bigdl.chronos.data import TSDataset
TSDataset.from_pandas(df,
                      dt_col,
                      target_col,
                      id_col=None,
                      extra_feature_col=None,
                      with_split=False,
                      val_ratio=0,
                      test_ratio=0.1,
                      repair=True):
```

```python
from bigdl.chronos.data import TSDataset
TSDataset.from_parquet(path,
                       dt_col,
                       target_col,
                       id_col=None,
                       extra_feature_col=None,
                       with_split=False,
                       val_ratio=0,
                       test_ratio=0.1,
                       repair=True,
                       **kwargs):
```
### 3. Summary of the change 

- [x] repair timestamp type, ut
- [x] repair inconsistent interval, ut
- [x] repair missing data, ut
- [x] add this check_and_repair function to tsdataset initialization
- [x] tsdataset ut

### 4. How to test?
- [x] Unit test
- [x] Application test

### DIFF
--- a/python/chronos/src/bigdl/chronos/data/tsdataset.py
+++ b/python/chronos/src/bigdl/chronos/data/tsdataset.py
@@ -27,6 +27,7 @@ from bigdl.chronos.data.utils.scale import unscale_timeseries_numpy
 from bigdl.chronos.data.utils.resample import resample_timeseries_dataframe
 from bigdl.chronos.data.utils.split import split_timeseries_dataframe
 from bigdl.chronos.data.utils.cycle_detection import cycle_length_est
+from bigdl.chronos.data.utils.quality_inspection import quality_check_timeseries_dataframe
 from bigdl.chronos.data.utils.utils import _to_list, _check_type,\
     _check_col_within, _check_col_no_na, _check_is_aligned, _check_dt_is_sorted
 
@@ -36,12 +37,15 @@ _DEFAULT_ID_PLACEHOLDER = "0"
 
 
 class TSDataset:
-    def __init__(self, data, **schema):
+    def __init__(self, data, repair=True, **schema):
         '''
         TSDataset is an abstract of time series dataset.
         Cascade call is supported for most of the transform methods.
         '''
         self.df = data
+        # detect low-quality data and automatic repair (optional)
+        quality_check_timeseries_dataframe(df=self.df, dt_col=schema["dt_col"],
+                                           repair=repair)
         self.id_col = schema["id_col"]
         self.dt_col = schema["dt_col"]
         self.feature_col = schema["feature_col"].copy()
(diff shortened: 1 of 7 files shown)


python/chronos/src/bigdl/chronos/data/utils/quality_inspection.py
```diff
@@ -57,19 +80,51 @@ def _timestamp_type_check(df_column):
...
+def _timestamp_type_repair(df, dt_col):
+    '''
+    This repair is used to convert object or other non datetime64 timestamp column
+    to datetime dtype.
+    '''
+    try:
+        df[dt_col].astype('datetime64')
+    except:
+        return False
+    df[dt_col] = df[dt_col].astype('datetime64')
```
TheaperDeng, line 92: should we just
```python
try:
    df[dt_col] = df[dt_col].astype('datetime64')
...
```

python/chronos/src/bigdl/chronos/data/tsdataset.py
```diff
@@ -36,12 +37,17 @@
...
 class TSDataset:
-    def __init__(self, data, **schema):
+    def __init__(self, data, repair=True, **schema):
         '''
         TSDataset is an abstract of time series dataset.
         Cascade call is supported for most of the transform methods.
         '''
         self.df = data
+        # detect low-quality data and automatic repair (optional)
+        flag, self.df = quality_check_timeseries_dataframe(df=self.df,
```
TheaperDeng, line 47: where do you want to use this `flag`?

TheaperDeng (commented):
others LGTM, @liangs6212 please also have a look.

python/chronos/src/bigdl/chronos/data/utils/quality_inspection.py
rnwang04: done

python/chronos/src/bigdl/chronos/data/tsdataset.py
rnwang04: Actually I am not sure. There are two possible ways:
one is utilize this flag to propose different message for user like this:
```python
if flag is True:
  loggings.warn("Now that there is no low quality data here.")
elif flag is False:
  loggings.warn("There are still some low quality data.")

```
Another way is just omit this flag
```python
_, self.df = quality_check_timeseries_dataframe(df=self.df,
```

liangs6212: I have no idea, but giving logs via `flag` seems like a good option.

TheaperDeng: we could just omit it since complete warning and repairing information has been provided.

rnwang04: > we could just omit it since complete warning and repairing information has been provided.

done

PR #6075, opened 2022-10-11
Chronos: change import path of `get_public_dataset`

plusbang:
## Description

For users, `from bigdl.chronos.data.repo_dataset import get_public_dataset` is quite hard to remember.
In https://github.com/intel-analytics/BigDL/pull/5983, `from .repo_dataset import get_public_dataset` has been add in init.py.
Therefore, the import path of `get_public_dataset` are all changed in this PR.

### 2. User API changes

If users want to use `get_public_dataset`
```python
from bigdl.chronos.data import get_public_dataset
```


TheaperDeng (approved):
LGTM

PR #6091, opened 2022-10-11
[Chronos] Solve the problem of slow initialization for tf2 seq2seq forecaster

devWangBin:
### 1. Why the change?
Solve the problem of slow initialization for tf2 seq2seq forecaster.

### 2. User API changes
Add new API:
bigdl.chronos.model.tf2.Seq2Seq_keras: **model_creator_auto**

### DIFF
--- a/python/chronos/src/bigdl/chronos/model/tf2/Seq2Seq_keras.py
+++ b/python/chronos/src/bigdl/chronos/model/tf2/Seq2Seq_keras.py
@@ -150,6 +150,22 @@ def model_creator(config):
                         lstm_layer_num=config.get("lstm_layer_num", 2),
                         dropout=config.get("dropout", 0.25),
                         teacher_forcing=config.get("teacher_forcing", False))
+    learning_rate = config.get('lr', 1e-3)
+    model.compile(optimizer=getattr(tf.keras.optimizers,
+                                    config.get("optim", "Adam"))(learning_rate),
+                  loss=config.get("loss", "mse"),
+                  metrics=[config.get("metics", "mse")])
+    return model
+
+
+def model_creator_auto(config):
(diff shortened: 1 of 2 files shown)


python/chronos/src/bigdl/chronos/model/tf2/Seq2Seq_keras.py
```diff
@@ -143,6 +143,22 @@ def from_config(cls, config):
...
+                        teacher_forcing=config.get("teacher_forcing", False))
+    learning_rate = config.get('lr', 1e-3)
+    model.compile(optimizer=getattr(tf.keras.optimizers,
+                                    config.get("optim", "Adam"))(learning_rate),
+                  loss=config.get("loss", "mse"),
+                  metrics=[config.get("metics", "mse")])
+    return model
+
+
+def model_creator_auto(config):
```
TheaperDeng, line 161: add comment here to state why we need this special model creator.

devWangBin: The reason is as follows:
The new **model_creator_auto** function adds two lines of code compared to the original **model_creator**, which will run (model(inputs)) to initialize the weights when creating the model, which is required for the AutoSeq2Seq model.

TheaperDeng: yeah just put a simple version to the src's comment. e.g. add model(inputs) in this model creator to initialize the weights

devWangBin: got it

TheaperDeng: great, will merge once the pr validations pass

PR #6106, opened 2022-10-13
[PPML]fix: Variable name and command error

ShanSimu:
## Description

fix variable name and command error in bigdl-ppml-submit.sh of Gramine


Uxito-Ada (approved):
LGTM

PR #6116, opened 2022-10-13
Nano: pytorch trace onnx & openvino support user customized dynamic_axes.

cyita:
## Description

The current implementation ignores the user passed `dynamic_axes` for the `torch.onnx.export` function.

### 2. User API changes

The user can pass dynamic_axes in trace and quantize.

description:
dynamic_axes : dict or boolean, default: True. By default, the exported model will
           have the first dim of each input as a dynamic batch_size. If dynamic_axes=False, the
           exported model will have the shapes of all input and output tensors set to exactly match
           those given in input_sample. To specify axes of tensors as dynamic (i.e. known only at
           run-time), set dynamic_axes to a dict with schema:
           KEY (str): an input or output name. Each name must also be provided in input_names or
           output_names.
           VALUE (dict or list): If a dict, keys are axis indices and values are axis names. If a
           list, each element is an axis index.

### 4. How to test?
- [ ] Unit test


rnwang04 (commented):
Others LGTM

PR #6119, opened 2022-10-14
[WIP] Chronos: add a seasonality checker to TSDataset's initialization

TheaperDeng:
### 1. Why the change?
1. `statsmodel` is not in our default dependency, which makes the cycle detection not friendly.
2. a seasonality checker is needed to warn our users that their data might not be forecasted easily.

### 2. User API changes
nothing

### 3. Summary of the change 
1. Change the implementation in cycle detection from relying on `statsmodel` to plain numpy implementation
4. Add a new checker in quality check without repair(since we can't repair this one) to 

### 4. How to test?
- [ ] Unit test

### 5. New dependencies

Remove a dependency: `statsmodel`


TheaperDeng:
hold this PR for a better API design due to the high latency add-on

PR #6120, opened 2022-10-14
[PPML] add bigdl jars into classpath in bigdl-ppml-submit.sh

liu-shaojun:
### 1. Why the change?
The versions of some libraries of bigdl and spark are in conflict, bigdl-ppml-submit.sh is to load the library of spark first, so Azure API cannot be called successfully. This submission is to add bigdl into classpath so that bidl jars can be loaded first.


jenniew (approved):
LGTM

PR #6128, opened 2022-10-14
Chronos: add `TSInferenceOptimizer` and use it to replace `Trainer.quantize/trace`

TheaperDeng:
### 1. Why the change?
`Trainer.quantize` and `Trainer.trace` is deprecated, use `InferenceOptimizer` to replace `Trainer`.

### 2. User API changes
Nothing

### 3. Summary of the change 
Add a chronos namespace InferenceOptimizer
Change base forecaster to use this rather than Trainer

### 4. How to test?
- [x] Unit test


rnwang04:
LGTM.

PR #6163, opened 2022-10-18
Chronos: enhance `from_prometheus()` from a single query statement to a query list

plusbang:
## Description

we notice that it will be more flexible if query list is supported.

### DIFF
--- a/python/chronos/src/bigdl/chronos/data/utils/prometheus_df.py
+++ b/python/chronos/src/bigdl/chronos/data/utils/prometheus_df.py
@@ -165,6 +165,8 @@ def GetRangeDataframe(prometheus_url, query_list, starttime, endtime, step, colu
     pro_df = pd.DataFrame()
     for query in query_list:
         query_df = pro_client.query_range(query, starttime, endtime, step, **kwargs)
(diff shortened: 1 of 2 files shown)


python/chronos/src/bigdl/chronos/data/utils/prometheus_df.py
```diff
@@ -156,22 +156,25 @@ def GetRangeDataframe(prometheus_url, query, starttime, endtime, step, columns,
...
-    df = pd.DataFrame(columns=Pro_df.columns.tolist())
-    df.insert(0, "datetime", Pro_df.index)
-    for col in Pro_df.columns.tolist():
-        df[col] = Pro_df[col].values
+    # Generate dataframe according query_list
+    pro_client = Prometheus(prometheus_url)
+    pro_df = pd.DataFrame()
+    for query in query_list:
+        query_df = pro_client.query_range(query, starttime, endtime, step, **kwargs)
+        pro_df = pd.concat([pro_df, query_df], axis=1)
```
TheaperDeng, line 168: whatif pro_df and query_df has different length (maybe caused by missing value in prometheus)

PR #6166, opened 2022-10-18
Add includes and excludes options in InferenceOptimizer

yangw1234:
## Description

Add includes and excludes options in InferenceOptimizer.

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
+++ b/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
@@ -133,6 +133,11 @@ class InferenceOptimizer:
         and record the latency, accuracy and model instance inside the Optimizer for
         future usage. All model instance is setting to eval mode.
 
+        The available methods are "original", "fp32_ipex", "bf16", "bf16_ipex","int8",
+        "jit_fp32", "jit_fp32_ipex", "jit_fp32_ipex_channels_last", "openvino_fp32",
+        "openvino_int8", "onnxruntime_fp32", "onnxruntime_int8_qlinear"
+        and "onnxruntime_int8_integer"
+
         :param model: A torch.nn.Module to be optimized
         :param training_data: A torch.utils.data.dataloader.DataLoader object for training
                dataset. Users should be careful with this parameter since this dataloader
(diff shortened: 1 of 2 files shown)


python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -163,6 +165,10 @@ def optimize(self, model: nn.Module,
                Default: False.
         :param latency_sample_num: (optional) a int represents the number of repetitions
                to calculate the average latency. The default value is 100.
+        :param includes: (optional) a list of acceleration methods that will be included in the
+               search. Default to None meaning including all available methods.
+        :param excludes: (optional) a list of acceleration methods that will be excluded from the
+               search.
```
TheaperDeng, line 171: should we provide all the available choices here in the api doc?

TheaperDeng: and of course the special logic to "original" since it has to be executed, so we may ignore it's presence in `includes` or `excludes`.

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -751,8 +759,25 @@ def _available_acceleration_combination():
                           "openvino": _openvino_checker,
                           "pot": _openvino_checker,
                           "bf16": _bf16_checker}
+    if excludes is None:
+        exclude_set: Set[str] = set()
+    else:
+        exclude_set: Set[str] = set(excludes)
```
TheaperDeng, line 765: should we check to make sure "original" is not in excludes?

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: > should we provide all the available choices here in the api doc?

I think we should provide all available choices name in the api doc.

rnwang04:
Do we need to provide two options to users or just keep `includes` or `excludes` is enough ? @TheaperDeng

yangw1234:
> Do we need to provide two options to users or just keep `includes` or `excludes` is enough ? @TheaperDeng

I think sometimes `includes` is more convenient and sometimes `excludes` is more convenient and users can choose the argument they see fit.

PR #6174, opened 2022-10-19
PPML FL example documents update

Litchilitchy:
## Description
some deletion of useless docs and updates of obsolete docs.


qiuxin2012:
You can use bigdl-assembly-all-with-dependencies.jar instread of ppml.jar

PR #6175, opened 2022-10-19
[WIP]Orca: adjust the format of xshards data input in unit test of Orca torch

Chaselzxy:
### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/orca/src/bigdl/orca/data/utils.py
+++ b/python/orca/src/bigdl/orca/data/utils.py
@@ -344,11 +344,6 @@ def process_spark_xshards(spark_xshards, num_workers):
 def index_data(x, i):
     if isinstance(x, np.ndarray):
         return x[i]
-    elif isinstance(x, dict):
-        res = {}
-        for k, v in x.items():
-            res[k] = v[i]
-        return res
     elif isinstance(x, tuple):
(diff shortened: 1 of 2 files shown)


leonardozcm:
https://github.com/intel-analytics/BigDL/pull/5763/files#diff-652522ccf78fc473ab6e253be32aa5f1383be1f95bea6e1e30303ef7c058ab73R279

I think we don't need this either.

python/orca/src/bigdl/orca/learn/pytorch/training_operator.py
```diff
@@ -276,8 +276,6 @@ def train_batch(self, batch, batch_info):
                 output = self.model(features)
             elif isinstance(features, dict):
                 output = self.model(**features)
-            elif isinstance(features, (tuple, list)):
-                output = self.model(*features)
```
leonardozcm, line 280: remove this in `forward_batch` as well

leonardozcm: I think this is not right here, we should remove the case when features is dict instead of tuple right? @Chaselzxy

leonardozcm:
Close this and assign to @PatrickkZ

PR #6178, opened 2022-10-19
[Chronos]: Support building dist on Windows

hjzin:
### 1. Why the change?

Support building dist on Windows.

### 2. User API changes

N/A
### 3. Summary of the change 

1. make incompatible dependencies and functions only available for linux (ipex, unsetenv(support windows in python 3.9))
2. make some path compatible for windows

To build and install chronos distribution on windows, nano distribution need to be built and installed first.

### DIFF
--- a/python/nano/setup.py
+++ b/python/nano/setup.py
@@ -77,15 +77,24 @@ def setup_package():
                            "keras==2.7.0",
                            "tensorflow-estimator==2.7.0"]
 
-    pytorch_requires = ["torch==1.11.0",
-                        "torchvision==0.12.0",
-                        "pytorch_lightning==1.6.4",
-                        "torchmetrics==0.7.2",
-                        "opencv-python-headless",
-                        "PyTurboJPEG",
-                        "opencv-transforms"
-                        # "intel_extension_for_pytorch==1.11.0"
-                        ]
+    # ipex is only avaliable for linux now
+    if os.name == 'nt':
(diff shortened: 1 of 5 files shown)


python/nano/setup.py
```diff
@@ -77,14 +77,15 @@ def setup_package():
...
+    # ipex is only avaliable for linux now
     pytorch_requires = ["torch==1.11.0",
                         "torchvision==0.12.0",
                         "pytorch_lightning==1.6.4",
                         "torchmetrics==0.7.2",
                         "opencv-python-headless",
                         "PyTurboJPEG",
                         "opencv-transforms",
-                        "intel_extension_for_pytorch==1.11.0"]
+                        "intel_extension_for_pytorch==1.11.0;platform_system!='Windows'"]
```
TheaperDeng, line 88: Could you provide the syntex/doc reference that support the platform check logic here in the PR comment?

TheaperDeng (commented):
others LGTM

python/nano/setup.py
hjzin: > Could you provide the syntex/doc reference that support the platform check logic here in the PR comment?


You can refer to this doc: https://setuptools.pypa.io/en/latest/userguide/dependency_management.html.
And the syntax is here: https://peps.python.org/pep-0508/

TheaperDeng (approved):

PR #6202, opened 2022-10-19
Nano: add gpu-to-cpu feature and uts

TheaperDeng:
## Description

This PR is a collaborated work with @MeouSker77 

### 1. Why the change?
Currently, many pytorch codes written by researchers are written for CUDA specificly, this makes users really frustrating when they want to have a try on CPU servers. This PR proposed an API to solve this problem by only change 1 lines of code.

First round supported apis:
- [x] `torch.Tensor.cuda()`
- [x] `torch.device("cuda")`
- [x] `torch.cuda.amp.autocast`
- [x] `torch.cuda.amp.GradScaler`
- [x] `torch.cuda.set_device`
- [x] `torch.cuda.synchronize`
- [x] `torch.nn.Module.cuda()`
- [x] `torch.cuda.Stream`
- [x] `torch.cuda.current_stream()`
- [x] `torch.cuda.{Float, Int, ...}Tensor`
- [x] `torch.cuda.reset_peak_memory_stats()`
- [x] `torch.cuda.reset_accumulated_memory_stats()`
- [x] `torch.distributed.init_process_group(backend='nccl', ...)`
- [x] `torch.{tensor, rand, ...}(..., device='cuda:0')`
- [x] `tensor.to('cuda:0')`
- [x] `model.to('cuda:0')`
- [x] `torch.load(f, map_location='cuda:0')`

### 2. User API changes

```python
from bigdl.nano.pytorch import patch_torch
patch_torch(gpu_to_cpu=True)

# any pytorch codes that written for CUDA could now run on CPU
```

### 3. Summary of the change 
basically the implementation is based on monkey patch.

- [x] torch version check (since we have not test all versions, I will add the checker after actions are completed the first time)

### 4. How to test?
- [x] Unit test


python/nano/src/bigdl/nano/pytorch/dispatcher.py
```diff
@@ -51,7 +52,7 @@ def _get_patch_map():
     return _mapping_torch
 
 
-def patch_torch():
+def patch_torch(gpu_to_cpu: bool = False):
```
jason-dai, line 55: why not default to true?

TheaperDeng: well because it's experimental for now. But we could set it to True for finding bugs quicker :)

python/nano/src/bigdl/nano/pytorch/dispatcher.py
```diff
@@ -51,7 +52,7 @@ def _get_patch_map():
     return _mapping_torch
 
 
-def patch_torch():
+def patch_torch(gpu_to_cpu: bool = True):
```
jason-dai, line 55: maybe `cuda_to_cpu`

PR #6214, opened 2022-10-20
[PPML] Modify settings in trusted-big-data-ml remote attestation settings

gc-fu:
## Description
Add the following settings to bash.manifest.template:
1. sgx.remote_attestation = "dcap"
2. sys.enable_extra_runtime_domain_names_conf = true

The original setting in first one will generate a warning.

The second one will ensure the gethostname() works correctly in gramine.


Uxito-Ada (approved):
LGTM

Issue #6199, opened 2022-10-19
Nano: bigdl-nano-init not found


Siritao:
When I use conda to activate my env, it says _bigdl-nano-init not found_. But when I directly call bigdl-nano-init in the terminal, it can run successfully. This may be due to the path: 
`>>> bigdl.nano.__file__`
`'/home/cpx/.local/lib/python3.8/site-packages/bigdl/nano/__init__.py'`

`which bigdl-nano-init`
`/home/cpx/.local/bin/bigdl-nano-init`
which is not under /home/cpx/anaconda3/...

PR #6217, opened 2022-10-20
Nano: fix the issue of `bigdl-nano-init`

MeouSker77:
## Description

If user has installed `bigdl-nano` using system's pip, then he can source `bigdl-nano-init` in any conda environment, and this script will mistakenly think that current conda environment has installed `bigdl-nano`, fix this issue: #6199

And, fix another small issue, which will cause `bash: deactivate: command not found`


yangw1234:
> 
I think we should support the scenario where user just uses the system pip (e.g. running a job in a docker container). 

Isn't it the right behavior that if user activates the a conda env, he/she will not have access the command installed in system's pip?

PR #6220, opened 2022-10-20
[Doc] Fix overflow problems caused by long words

Oscilloscope98:
### 1. Why the change?

Sometimes long words such as url and class/method definition in API docs will go overflow the article contents, leading to the appearance of horizontal scroll bar, which makes user harder to see the full contents.

### 2. Summary of the change 
- force middle article contents to break an entire word that cannot be put inside one line (e.g. a long url)
- make inline code block break word if an entire word cannot be put inside one line, or break at normal line break places such as / 
  Example Before:
  image
  Example After:
  image

  Example Before:
  image
  Example After:
  image

### 3. How to test?
- [x] Document test: https://yuwentestdocs.readthedocs.io/en/doc-improve2/index.html

### DIFF
--- a/docs/readthedocs/source/_static/css/custom.css
+++ b/docs/readthedocs/source/_static/css/custom.css
@@ -10,6 +10,18 @@ html[data-theme="light"] {
     --pst-color-info-low-opacity: rgba(1, 113, 195, 0.25);
 }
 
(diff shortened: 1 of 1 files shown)


shane-huang:
LGTM

PR #6227, opened 2022-10-20
add fillna in xshards

songhappy:
## Description
reuse some code of Friesian Table to support Orca Xshards fillna

### 1. Why the change?
to avoid depleted code in orca xshards and friesian table 

### 2. User API changes
no change in friesian table

### 3. Summary of the change 
move scala Friesian Table fillna related implementation to Orca since Friesian depends on Orca
move python friesian.feature.utils fillna related to orca.data.utils 
add fillna in python orca.data.shard.py

### 4. How to test?
- [ ] Unit test

### 5. New dependencies
no dependencies

### DIFF
--- a/python/orca/src/bigdl/orca/data/shard.py
+++ b/python/orca/src/bigdl/orca/data/shard.py
@@ -24,6 +24,16 @@ from bigdl.dllib.utils.log4Error import *
 
 import numpy as np
 
+from typing import (
+    TYPE_CHECKING,
+    TypeVar,
+    Callable,
+    Dict,
+    List,
+    Optional,
+    Tuple,
+    Union,
+)
 
 class XShards(object):
     """
@@ -134,6 +144,7 @@ But got data of type {}
 
 
 class SparkXShards(XShards):
+
     """
 
     A collection of data which can be pre-processed in parallel on Spark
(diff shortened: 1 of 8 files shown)


python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -633,6 +644,19 @@ def zip(self, other):
...
+               columns: Optional[Union[str, List[str]]]):
+        if self._get_class_name() == 'pandas.core.frame.DataFrame':
+            import pandas as pd
+            df = self.to_spark_df()
+            filled_df = fill_na(df, value, columns)
+            data_shards = spark_df_to_pd_sparkxshards(filled_df)
+            return data_shards
+        else:
+            invalidInputError(False,
+                              "Currently only support dedup() on XShards of Pandas DataFrame")
```
dding3, line 657: .... only support fillna() on XShards ...?

python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -633,6 +644,19 @@ def zip(self, other):
                               "The two SparkXShards should have the same number of elements "
                               "in each partition")
 
+    def fillna(self, value: Union[int, float, str, bool],
+               columns: Optional[Union[str, List[str]]]):
```
dding3, line 648: maybe we can also add some comments to this api?

python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -633,6 +644,19 @@ def zip(self, other):
...
+        if self._get_class_name() == 'pandas.core.frame.DataFrame':
+            import pandas as pd
+            df = self.to_spark_df()
```
jason-dai, line 651: why do we need to convert to spark df? I think we can simply do a `mapPartition` and call Pandas `fillna`, yes?

songhappy: yes, the original design of this method wanted to leverage Friesian fillna with performance. but it is not necessary.  closed this PR

songhappy:
closed due to better implementation of just calling pandas dataframe apis

PR #6234, opened 2022-10-21
Chronos: Add how-to guides for exporting ONNX/OpenVINO files to disk

hjzin:
### 1. Why the change?

Add how-to guides for exporting ONNX/OpenVINO files to show the usage of `forecaster.export_onnx_file` and `forecaster.export_openvino_file`.
### 2. User API changes

N/A
### 3. Summary of the change 

- How to export ONNX file to disk: https://zhaojie-doc.readthedocs.io/en/forecaster-export-howto/doc/Chronos/Howto/how_to_export_onnx_files.html
- How to export OpenVINO file to disk: https://zhaojie-doc.readthedocs.io/en/forecaster-export-howto/doc/Chronos/Howto/how_to_export_openvino_files.html


TheaperDeng:
Some comments (put it here since notebook is really hard to comment line by line):
The comment is both effective for onnx and openvino
1. We don't need to call `exit()` for current colab usage. correct me if I am wrong @AndyLuo1029 
2. We need to change
> If you just need to export fp32 onnx files, you could specify dirname only and leave quantized_dirname to the default value: forecaster.export_onnx_file(dirname=ckpt_name)

to

> If you just need to export fp32 onnx files, you could specify dirname only and set quantized_dirname to None: forecaster.export_onnx_file(dirname=ckpt_name, quantized_dirname=None)

The original instruction is right, still the changing is because that some earlier version (2.1.0 and some earlier 2.2.0b) the original `quantized_dirname` was not `None` so errors may be caused if leaving `quantized_dirname` to the default value

TheaperDeng:
I think this PR makes some of the structure needs to be changed (basically those yml files)

PR #6240, opened 2022-10-21
Fixed spark example failing to resolve SPARK_VERSION environment

zzti-bsj:
## Description

Fixed spark example failing to resolve `SPARK_VERSION` environment


zzti-bsj:
the [issue 720](https://github.com/analytics-zoo/arda-docker/issues/720) show that python example can't parse the `$SPARK_VERSION` on the data path. Double quotation marks are used in the `sed` command to parse shell environment variables.

PR #6247, opened 2022-10-21
Update resnet50 inference

hkvision:
- batch_size means batch for each worker, num_steps means steps for each worker
- 50000 val images in total
- conda install pytorch torchvision torchaudio cpuonly -c pytorch
- pip install intel_extension_for_pytorch


hkvision:
@zpeng1898

PR #6251, opened 2022-10-24
[Orca] Support load model instead of  `model_creator` in tf2 ray estimator

sgwhat:
## Description
This PR refactor tf2 ray estimator to support users loading a saved model instead of defining a `model_creator`.

### 1. Why the change?
Some users want to load the model directly but not define a `model_creator`, we could only provide a workaround solution to them and that's really not user-friendly.

### 2. User API changes
The new API accepts `moedl_creator` as None, which will not create a model until users load the model through `load()` API.
```python
trainer= estimator.from_keras(
    workers_per_node=3,
    backend="ray")

trainer.load("/path/to/model.h5")

trainer.predict(df,  feature_cols)
```

### 3. Summary of the change 
Refactor `model_creator` as an optional, users could load their saved model through `estimator.load()`.

### 4. How to test?
- [x] Unit test
- [x] Application test

### DIFF
--- a/python/orca/src/bigdl/orca/learn/tf2/tf_runner.py
+++ b/python/orca/src/bigdl/orca/learn/tf2/tf_runner.py
@@ -379,7 +379,7 @@ class TFRunner:
         invalidInputError(hasattr(self, "model"),
                           "The model has not yet been created. "
                           "Please input a model_creator when creating estimator "
-                          "or use load function of the estimator to load a model")
+                          "or use load function of the estimator to load a model.")
 
         history = self.model.fit(train_dataset,
                                  epochs=self.epoch + epochs,
@@ -441,7 +441,7 @@ class TFRunner:
         invalidInputError(hasattr(self, "model"),
                           "The model has not yet been created. "
                           "Please input a model_creator when creating estimator "
-                          "or use load function of the estimator to load a model")
+                          "or use load function of the estimator to load a model.")
 
         results = self.model.evaluate(dataset, **params)
         if results is None:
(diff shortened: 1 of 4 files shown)


hkvision:
Two questions:
- load the model can only do predict right? What about resume training, is it possible to support? (though not that high priority)
- It's better to have somewhere report a user-friendly error when model_creator is None and user don't call load?

sgwhat:
> Two questions:
> 
> * load the model can only do predict right? What about resume training, is it possible to support? (though not that high priority)
> * It's better to have somewhere report a user-friendly error when model_creator is None and user don't call load?

Yes, will add scope effect to support resume training.

python/orca/src/bigdl/orca/learn/tf2/tf_runner.py
```diff
@@ -323,9 +328,10 @@ def setup_distributed(self, urls, world_rank, world_size):
 
         logger.debug("Creating model with MultiWorkerMirroredStrategy")
         with self.strategy.scope():
```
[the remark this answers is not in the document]
sgwhat: It can only resume predicting now, I'll add the scope effect for resume training soon.

python/orca/src/bigdl/orca/learn/tf2/tf_runner.py
```diff
@@ -370,6 +376,12 @@ def step(self, data_creator, epochs=1, batch_size=32, verbose=1,
         if callbacks:
             replaced_log_dir = process_tensorboard_in_callbacks(callbacks, "fit", self.rank)
 
+        if self.model is None:
+            invalidInputError(False,
+                              "The model has not yet been created. "
+                              "Please input a model_creator when creating estimator "
+                              "or use load function of the estimator to load a model")
```
hkvision, line 383: Can we put this patch into a utility/static function to avoid duplicate code in these functions?

sgwhat: I think we can only duplicate it...

hkvision (approved):

PR #6256, opened 2022-10-24
Chronos: add unit tests for `TSDataset.from_prometheus()`

plusbang:
## Description

Chronos now has a new feature that can fetch data from prometheus database to TSDataset and we have added a prometheus service hosted on one of our servers.
The corresponding unit test is then added.

### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/chronos/test/bigdl/chronos/data/test_tsdataset.py
+++ b/python/chronos/test/bigdl/chronos/data/test_tsdataset.py
@@ -1433,7 +1433,7 @@ class TestTSDataset(TestCase):
         import time
         endtime = time.time()
         starttime = endtime - 1000
-        prometheus_url = os.getenv("PROMETHEUS_URL")
+        prometheus_url = {os.getenv("PROMETHEUS_URL")}
         tsdata = TSDataset.from_prometheus(prometheus_url=prometheus_url,
                                            query='prometheus_ready',
                                            starttime=starttime,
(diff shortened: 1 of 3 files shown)


python/chronos/test/bigdl/chronos/data/test_tsdataset.py
```diff
@@ -1344,3 +1344,75 @@ def test_tsdataset_interval_check_and_repair_for_multi_id(self):
...
                                         target_col=['value', 'extra feature'])
+
+    def test_from_prometheus(self):
+        '''
+        The selected query str is just for getting data from Prometheus server.
+        '''
+        import time
+        endtime = time.time()
+        starttime = endtime - 1000
+        tsdata = TSDataset.from_prometheus(prometheus_url='http://10.112.231.51:19090/',
```
TheaperDeng, line 1355: we may avoid setting the url explicitly, alternatively we may use `os.getenv("PROMETHEUS_URL")` and set this parameter for all action runners like 
https://github.com/intel-analytics/BigDL/blob/c10b4ceee260dd011acfd038328008d048b7623e/python/chronos/test/bigdl/chronos/simulator/test_doppelganger_simulator.py#L58

plusbang: done

PR #6268, opened 2022-10-24
Update Orca install guide

hkvision:


hkvision:
https://hkvision-orca-1024-5.readthedocs.io/en/latest/doc/Orca/Overview/install.html

Issue #6183, opened 2022-10-19
Nano: more friendly install option


TheaperDeng:
With our experience on nano and users' feedback, we find that certain install options can be supported by nano for better usability:

### 1. pytorch, tensorflow version
We currently provide widely support for different version's pytorch (1.9-1.12) and tensorflow (2.7-2.9). Still the installation of `bigdl-nano[pytorch]` and `bigdl-nano[tensorflow]` fixed a set of value (th 1.11, tf 2.7). This makes some of the users who
- want to use features in later version
- has a well defined environment that's supported by bigdl-nano but not the default version
needs to uninstall the default one and install pytorch/tensorflow (and other affected dependencies) and install the new ones again.

Should we provide some options such as `[pytorch111]`, `[pytorch112]`? inspired by [cupy](https://github.com/cupy/cupy/releases).

### 2. inference related dependencies
Currenly most of them are optional, should we add an option to install those (e.g. onnxruntime, onnx, tf2onnx, openvino-dev, neural-compressor...) together? The installation guide is too long for inference related how-to guide and unfriendly to our users.

maybe the option could be `[pytorch-inference]`, `[tensorflow-inference]`?

@rnwang04 please comment if I miss anything for this second part.

### 3. make ipex an optional one
easy reason, it does not support native windows. https://github.com/intel/intel-extension-for-pytorch/blob/master/setup.py#L631-L632

TheaperDeng:
@yangw1234 any comments?

rnwang04:
for inference related dependencies, another candidate name is [pytorch-all], [tensorflow-all].

yangw1234:
I am ok with adding more options in the `extra_installs`.

jason-dai:
Should upgrade to latest TF/PT version

TheaperDeng:
Here is a discussion result with @ruobing-wang for **point 1 pytorch, tensorflow version**

simple Installation will always install the latest release version (so we will update them regularly)
```bash
pip install --pre --upgrade bigdl-nano[pytorch]  # for latest pytorch (currently 1.12) basic dependencies
pip install --pre --upgrade bigdl-nano[pytorch, inference]  # for latest pytorch (currently 1.12) + inference optimizations
```

Users could install pytorch/tensorflow with specificed version, by simply

```bash
pip install --pre --upgrade bigdl-nano[pytorch_110]  # for pytorch 1.10 basic dependencies
[...]

PR #6281, opened 2022-10-25
Nano : Update torch version and add inference requires in setup

rnwang04:
## Description

Currenly most of inference related dependencies are optional, we should add an option to install those (e.g. onnxruntime, onnx, tf2onnx, openvino-dev, neural-compressor...) together.  
The installation guide is too long for inference related how-to guide and unfriendly to our users.

### 1. Why the change?

https://github.com/intel-analytics/BigDL/issues/6183 inference related dependencies

### 2. User API changes
```
# for latest supported pytorch (1.12.1) basic dependencies
pip install --pre --upgrade bigdl-nano[pytorch] 
# for latest supported pytorch (1.12.1) + inference optimizations
pip install --pre --upgrade bigdl-nano[pytorch,inference]
# tensorflow (2.7.0)
pip install --pre --upgrade bigdl-nano[tensorflow]
# tensorflow (2.7.0) + inference optimizations
pip install --pre --upgrade bigdl-nano[tensorflow,inference]
```

### 3. Summary of the change 
- [x] upgrade default torch version to 1.12.1
- [x] provide new installation option : inference 

### 4. How to test?
- [x] Application test

### DIFF
--- a/.github/workflows/nano_unit_tests_pytorch.yml
+++ b/.github/workflows/nano_unit_tests_pytorch.yml
@@ -200,9 +200,9 @@ jobs:
         pytorch-version: [
           # The reason why ipex1.10 is not currently supported is that the optimization of ipex1.10
           # will change the require_grad attribute of layers, which will make some tests of nano fail.
-          "torch==1.9.0+cpu torchvision==0.10.0+cpu torch_ipex==1.9.0",
+          "torch==1.9.0+cpu torchvision==0.10.0+cpu torch_ipex==1.9.0 intel_extension_for_pytorch==1.9.0",
           "",
-          "torch==1.12.1+cpu torchvision==0.13.1+cpu intel_extension_for_pytorch==1.12.100"
+          "torch==1.12.1+cpu torchvision==0.13.1+cpu intel_extension_for_pytorch==1.12.100 torch_ipex==1.12.100"
           # "torch==1.10.1+cpu torchvision==0.11.2+cpu intel_extension_for_pytorch==1.10.0 psutil"
           ]
 
--- a/.github/workflows/nano_unit_tests_pytorch.yml
+++ b/.github/workflows/nano_unit_tests_pytorch.yml
@@ -24,10 +24,9 @@ jobs:
         os: ["ubuntu-20.04"]
         python-version: ["3.7"]
         pytorch-version: [
-          "torch==1.9.0+cpu torchvision==0.10.0+cpu",
-          "torch==1.10.1+cpu torchvision==0.11.2+cpu",
-          "torch==1.12.1+cpu torchvision==0.13.1+cpu",
-          ""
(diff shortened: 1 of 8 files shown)


.github/workflows/nano_unit_tests_pytorch.yml
```diff
@@ -200,9 +200,9 @@ jobs:
         pytorch-version: [
           # The reason why ipex1.10 is not currently supported is that the optimization of ipex1.10
           # will change the require_grad attribute of layers, which will make some tests of nano fail.
-          "torch==1.9.0+cpu torchvision==0.10.0+cpu torch_ipex==1.9.0",
+          "torch==1.9.0+cpu torchvision==0.10.0+cpu torch_ipex==1.9.0 intel_extension_for_pytorch==1.9.0",
```
yangw1234, line 203: I think they changed the ipex's package name from torch_ipex to intel_extension_for_pytorch since 1.10.

image

rnwang04: Thanks for this reminder!  Have revised workflow.

.github/workflows/nano_unit_tests_pytorch.yml
```diff
@@ -24,10 +24,10 @@ jobs:
         os: ["ubuntu-20.04"]
         python-version: ["3.7"]
         pytorch-version: [
-          "torch==1.9.0+cpu torchvision==0.10.0+cpu",
-          "torch==1.10.1+cpu torchvision==0.11.2+cpu",
-          "torch==1.12.0+cpu torchvision==0.13.0+cpu",
-          ""
+          "torch==1.9.0+cpu torchvision==0.10.0+cpu intel_extension_for_pytorch==1.11.0",
```
yangw1234, line 27: Does ipex 1.11.0 works with all previous version?

rnwang04: Yes, and I think this workflow has used ipex1.11.0 before to adapt to torch 1.9-1.11.

https://github.com/intel-analytics/BigDL/blob/685ec6c494df389564cf4fa6bc514fde35169574/python/nano/setup.py#L80-L88

https://github.com/intel-analytics/BigDL/blob/685ec6c494df389564cf4fa6bc514fde35169574/.github/workflows/nano_unit_tests_pytorch.yml#L27-L30

yangw1234: I see. 

I think this is because this workflow only test pytorch-only test cases and no ipex cases. 
It may not work if if you call ipex 1.11.0 using pytorch 1.10 and before.

TheaperDeng (approved):

.github/workflows/nano_unit_tests_pytorch.yml
rnwang04: Thanks, I know what you mean. I haven't specially tested this. (Due to network reasons) I will merge this PR first, and then open a separate PR to test the support of ipex version.

PR #6284, opened 2022-10-25
[PPML] Add PolicyID to AttestationCLI

xiangyuT:
## Description

Current AttestationCLI haven't submit PolicyID to ehsm service so ehsm won't check mrenclave during attestation. Need to add policyID when doing attestation.


Uxito-Ada (approved):
LGTM

PR #6306, opened 2022-10-26
update gramine test-suites

zzti-bsj:
## Description

update gramine test-suites
1.  add all suites as `spark/pyspark` UT scripts input
2. `spark/pyspark` UT scripts will run 3 times for all failed suites
3. update `simplequery` example, delete the extra jar file


ShanSimu (approved):
LGTM

PR #6307, opened 2022-10-26
Nano: Fix: cuda patch takes no effect when running multi-instance training

MeouSker77:
## Description

Fix: CUDA patch takes no effect when running multi-instance training

Add warning about disabling JIT

### 1. Why the change?

CUDA patch takes no effect when running multi-instance training: <https://github.com/analytics-zoo/GPU-CPU-tool/issues/8>

### 2. User API changes

N/A

### 3. Summary of the change 

call `patch_torch` in subprocess automatically and add UTs

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...


TheaperDeng (approved):
LGTM

PR #6311, opened 2022-10-27
Nano : move import of ipex

rnwang04:
## Description

Now ipex and jit is seriously coupled together.  If ipex fails, jit fails too.
Here move the import of ipex into if to decouple them.

### How to test?
- [x] Application test


TheaperDeng:
if we decided to remove the check of bf16+ipex for 1.11, I think this PR should be ok to merge after that.

rnwang04:
> if we decided to remove the check of bf16+ipex for 1.11, I think this PR should be ok to merge after that.

I think so, will merge it now as action check for 1.11 has been removed.

PR #6314, opened 2022-10-27
[WIP] Orca: add pytorch tensorboard callback.

lalalapotter:
## Description
Add PyTorch TensorBoard callback for user convenience and API consistency.

### 1. Why the change?
+ User convenience: users have to use `torch.utils.tensorboard.SummaryWriter()` to record the metrics manually, it is not as convenient as TensorFlow callbacks.
+ API consistency: we have callbacks in TensorFlow Estimator, to unify the usage and estimator API, we need to add PyTorch TensorBoard callback as well.

### 2. User API changes
No changes, add an example here:
```python
orca_estimator = Estimator.from_torch(model=model_creator,
                                              optimizer=optimizer_creator,
                                              loss=criterion,
                                              metrics=[Accuracy()],
                                              use_tqdm=True,
                                              backend=args.backend)
tensorboard_callback = TensorBoardCallback(log_dir = "hdfs://hdfs/path/to/logdir")
stats = orca_estimator.fit(train_data_creator, epochs=epochs, batch_size=batch_size, callbacks=[tensorboard_callback])

```

### 3. Summary of the change 
Add new class `TensorBoardCallback`, which Inherit and extend the class `bigdl.orca.learn.pytorch.callbacks` , record the metrics and loss at the end of the batch or epoch according to the frequency, and store them in the target path at the end of the training, including the remote filesystem.

### 4. How to test?
- [x] Unit test
- [x] Application test: Test on yarn cluster/local with ray/spark backend and store the log in HDFS.

### DIFF
--- a/python/orca/src/bigdl/orca/learn/pytorch/callbacks/tensorboard.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/callbacks/tensorboard.py
@@ -24,7 +24,6 @@ from bigdl.orca.learn.pytorch.callbacks import Callback
 from bigdl.dllib.utils.log4Error import invalidInputError
 
 
-
 class TensorBoardCallback(Callback):
 
     def __init__(
@@ -36,10 +35,10 @@ class TensorBoardCallback(Callback):
         """
         :param log_dir: Log directory of TensorBoard.
         :param freq: Frequency of logging metrics and loss. 
-            Accept values: 'batch' or 'epoch' or integer. When using 'batch', 
+            Accept values: 'batch' or 'epoch' or integer. When using 'batch',
             writes the losses and metrics to TensorBoard after each batch.
-            The same applies for 'epoch'. If using an integer, let's say 1000, 
-            the callback will write the metrics and losses to TensorBoard every 1000 batches. 
+            The same applies for 'epoch'. If using an integer, let's say 1000,
+            the callback will write the metrics and losses to TensorBoard every 1000 batches.
             Note that writing too frequently to TensorBoard can slow down your training.
         :param **kwargs: The keyword arguments will be pased to ``SummaryWriter``.
         """
@@ -119,7 +118,6 @@ class TensorBoardCallback(Callback):
             put_local_dir_tree_to_remote(self.tmp_dir, self.log_dir)
             if os.path.exists(self.tmp_dir):
                 shutil.rmtree(self.tmp_dir)
-
 
     def set_model(self, model):
         self.model = model
--- a/python/orca/src/bigdl/orca/learn/pytorch/callbacks/tensorboard.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/callbacks/tensorboard.py
@@ -69,7 +69,7 @@ class TensorBoardCallback(Callback):
                 writer = SummaryWriter(log_dir=self.tmp_dir, **self.kwargs)
                 for name, value in logs.items():
                     if name not in self.unlog_items:
-                        writer.add_scalar(name, value, epoch)
+                        writer.add_scalar(name, value, batch)
                 writer.close()
 
     def on_epoch_begin(self, epoch):
--- a/python/orca/src/bigdl/orca/learn/pytorch/callbacks/tensorboard.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/callbacks/tensorboard.py
@@ -1,0 +1,130 @@
+#
+# Copyright 2016 The BigDL Authors.
+#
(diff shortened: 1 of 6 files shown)


lalalapotter:
Need to add UT.

python/orca/src/bigdl/orca/learn/pytorch/callbacks/tensorboard.py
```diff
@@ -0,0 +1,134 @@
...
+        """
+        pass
+
+    def on_train_end(self, logs=None):
+        """
+        Called at the end of training.
+        Subclasses should override for any actions to run.
+        """
+        if self._is_rank_zero():
+            put_local_dir_tree_to_remote(self.tmp_dir, self.log_dir)
```
sgwhat, line 118: add `if not local` here?

python/orca/src/bigdl/orca/learn/pytorch/callbacks/tensorboard.py
```diff
@@ -0,0 +1,134 @@
...
+        """
+        Called at the end of a training batch in `fit` methods.
+        Subclasses should override for any actions to run.
+        @param batch: Integer, index of batch within the current epoch.
+        """
+        if self.freq != "epoch" and self._is_rank_zero():
+            if self.freq == "batch" or batch % int(self.freq) == 0:
+                writer = SummaryWriter(log_dir=self.tmp_dir, **self.kwargs)
+                for name, value in logs.items():
+                    if name not in self.unlog_items:
```
sgwhat, line 70: why `self.unlog_items` here?

python/orca/src/bigdl/orca/learn/pytorch/callbacks/tensorboard.py
```diff
@@ -0,0 +1,134 @@
...
+    def on_batch_end(self, batch):
+        """
+        Called at the end of a training batch in `fit` methods.
+        Subclasses should override for any actions to run.
+        @param batch: Integer, index of batch within the current epoch.
+        """
+        if self.freq != "epoch" and self._is_rank_zero():
+            if self.freq == "batch" or batch % int(self.freq) == 0:
+                writer = SummaryWriter(log_dir=self.tmp_dir, **self.kwargs)
+                for name, value in logs.items():
```
sgwhat, line 69: where do the `logs` come from?

python/orca/src/bigdl/orca/learn/pytorch/callbacks/tensorboard.py
lalalapotter: Such situation will be addressed in `put_local_dir_tree_to_remote`, refer to [link](https://github.com/intel-analytics/BigDL/blob/eae520693fa9b542d1f77787fe1c9bea4cb65ec7/python/orca/src/bigdl/orca/data/file.py#L323).

python/orca/src/bigdl/orca/learn/pytorch/callbacks/tensorboard.py
lalalapotter: Because there are some stats that do not need to log in TensorBoard, such as "batch_count" and so on.

python/orca/src/bigdl/orca/learn/pytorch/callbacks/tensorboard.py
sgwhat: we need to support save to local path here? so maybe better to change as 
```python
if is_local_path(...):
    ...
else:
    put_local_dir_tree_to_remote()
```

python/orca/src/bigdl/orca/learn/pytorch/training_operator.py
```diff
@@ -232,7 +232,7 @@ def _train_loop(self, iterator, info, _progress_bar, metric_meters, callbacks):
             self.global_step += 1
             if callbacks is not None:
                 for callback in callbacks:
-                    callback.on_batch_end(batch_idx)
+                    callback.on_batch_end(batch_idx, logs=metrics)
```
sgwhat, line 235: If the metrics is none, we are not able to retrieve any tensorboard events?

lalalapotter: Yes.

sgwhat (approved):

lalalapotter:
Jenkins Validation:
http://10.112.231.51:18888/view/BigDL-2.0-PR/job/BigDL-Orca-PR-Validation/1186/
PR related tests passed, some other tests failed due to network issues.

PR #6318, opened 2022-10-27
Nano:  add more GPU operation support for GPU-CPU tool

MeouSker77:
## Description

add more GPU operation support for GPU-CPU tool, including <https://github.com/analytics-zoo/GPU-CPU-tool/issues/7>


TheaperDeng (approved):
LGTM

Issue #6308, opened 2022-10-26
Chronos: evaluate results are different when dataloader is shuffled or not


plusbang:
When generate two dataloaders with `shuffle=True` and `shuffle=False`, the evaluate results are different:
```python
test_loader_1 = tsdata_test.to_torch_data_loader(batch_size=1, lookback=6, horizon=2, shuffle=False)
test_loader_2 = tsdata_test.to_torch_data_loader(batch_size=1, lookback=6, horizon=2)
```
![image](https://user-images.githubusercontent.com/108676127/197992440-c89039ad-28be-46a7-8225-772c47038928.png)

liangs6212:
This result seems to be as expected? What if both datasets have shuffle=False set?

PR #6326, opened 2022-10-27
Chronos: fix incorrect evaluate results for shuffled dataloader

plusbang:
## Description

For shuffled dataloader, outputs of `evaluate` and `evaluate_with_onnx` are not proper. The reason is that the dataloader is iterated two times to get target values and inference results.

### 1. Why the change?

https://github.com/intel-analytics/BigDL/issues/6308

### 3. Summary of the change 

3.1 If dataloader is shuffled, convert input_data from dataloader to numpy, only iterate once
3.2 add uts

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/chronos/src/bigdl/chronos/forecaster/base_forecaster.py
+++ b/python/chronos/src/bigdl/chronos/forecaster/base_forecaster.py
@@ -932,8 +932,20 @@ class BasePytorchForecaster(Forecaster):
                 invalidInputError(False,
                                   "You must call fit or restore first before calling evaluate!")
             if isinstance(data, DataLoader):
-                input_data = data
-                target = np.concatenate(tuple(val[1] for val in data), axis=0)
+                from torch.utils.data.sampler import RandomSampler
+                if isinstance(data.sampler, RandomSampler):
+                    # If dataloader is shuffled, convert input_data to numpy()
+                    # Avoid to iterate shuffled dataloader two times
+                    input_x = []
+                    input_y = []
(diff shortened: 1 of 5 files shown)


python/chronos/src/bigdl/chronos/forecaster/base_forecaster.py
```diff
@@ -932,8 +932,20 @@ def evaluate(self, data, batch_size=32, multioutput="raw_values", quantize=False
...
+                    input_x = []
+                    input_y = []
+                    for val in data:
+                        input_x.append(val[0].numpy())
+                        input_y.append(val[1].numpy())
+                    input_data = np.concatenate(input_x, axis=0)
+                    target = np.concatenate(input_y, axis=0)
+                else:
+                    input_data = data
+                    target = np.concatenate(tuple(val[1] for val in data), axis=0)
```
TheaperDeng, line 948: We should implement a way to "update" the metric like the implementation in https://torchmetrics.readthedocs.io/en/stable/classification/accuracy.html#functional-interface to continually calculate the metric rather than have a large numpy ndarray in later PR. Maybe we need an issue for this.

TheaperDeng (approved):
LGTM

python/chronos/src/bigdl/chronos/forecaster/base_forecaster.py
plusbang: yeah, I think we can do it later

PR #6329, opened 2022-10-27
Nano : update opset version based on torch version

rnwang04:
## Description

update opset version based on torch version

### How to test?

- [x] Application test


TheaperDeng:
should we try to examine pytorch's version and assign the highest opset version that is supported?

rnwang04:
> should we try to examine pytorch's version and assign the highest opset version that is supported?

Good suggestion, done.

TheaperDeng (approved):
LGTM

PR #6334, opened 2022-10-27
[ppml]: create a workflow for ppml developer to build image based on specific commit id

liu-shaojun:
## Description

This submission creates a workflow (same as manually_build.yml) for ppml developer to build image based on specific commit id, then developers can specify a commit id to build a ppml image and verify.


hzjane (approved):
LGTM.

PR #6336, opened 2022-10-27
Orca: add requirements files in release package.

lalalapotter:
## Description

This PR is to add requirements files in release package for user setting up environment.


Le-Zheng:
LGTM

PR #6339, opened 2022-10-28
Chronos: Speed ​​up initialization weight matrix of tf2 model

liangs6212:
## Description

Using a smaller `numpy.ndarray` can speed up initializing the weights.

### 1. Why the change?

Initializing the weight matrix is ​​too slow.

### 2. Summary of the change 

```python
# server: 7900
# seq2seq_model_creator

no inputs:  Wall time: 25.8 ms

np.zeros(shape=(1, 1, 1)):  Wall time: 128 ms

Input(shape=(None, config['inputs_feature_num']))  Wall time: 728 ms
```
### 4. How to test?
- [ ] Unit test
- [ ] Application test
- [ ] Document test

### DIFF
--- a/python/chronos/src/bigdl/chronos/model/tf2/TCN_keras.py
+++ b/python/chronos/src/bigdl/chronos/model/tf2/TCN_keras.py
@@ -197,7 +197,7 @@ def model_creator(config):
                             kernel_size=config.get("kernel_size", 7),
                             dropout=config.get("dropout", 0.2),
                             repo_initialization=config.get("repo_initialization", True))
-    inputs = Input(shape=(config["past_seq_len"], config["input_feature_num"]))
+    inputs = np.zeros(shape(1, 1, 1))
     # init weights matrix
     model(inputs)
     learning_rate = config.get('lr', 1e-3)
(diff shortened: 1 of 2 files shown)


TheaperDeng (approved):
LGTM

python/chronos/src/bigdl/chronos/model/tf2/TCN_keras.py
```diff
@@ -197,7 +197,7 @@ def model_creator(config):
                             kernel_size=config.get("kernel_size", 7),
                             dropout=config.get("dropout", 0.2),
                             repo_initialization=config.get("repo_initialization", True))
-    inputs = Input(shape=(config["past_seq_len"], config["input_feature_num"]))
+    inputs = np.zeros(shape=(1, config["past_seq_len"], config["input_feature_num"]))
```
TheaperDeng, line 200: so does this mean we only need a "valid" input to go through the model, then the weights are initialized?

liangs6212: Yes, using a small dataset with the same shape can speed up initialization compared to Input.

PR #6354, opened 2022-10-28
Update resnet50 inference

hkvision:


python/orca/example/learn/pytorch/resnet50/inference.py
```diff
@@ -299,10 +307,7 @@ def model_creator(config):
         return model
 
     def optimizer_creator(model, config):
-        optimizer = torch.optim.SGD(model.parameters(), lr=0.1,
```
hkvision, line 302: ValueError: optimizer got an empty parameter list  # for models with jit

PR #6364, opened 2022-10-31
Nano: Add a compatible `CPUInfo` class to check the hardware information

hjzin:
### 1. Why the change?

Now we use `lscpu` to check whether the hardware supports avx512 or bf16, but it's only available in Linux, we need to make it compatible with other platforms like Windows.
### 2. User API changes

Use the following code check the avx512 or bf 16 support:
```
from bigdl.nano.utils import CPUInfo

cpuinfo = CPUInfo()

# check avx512 (return `True` or `False`)
has_avx512 = cpuinfo.has_avx512

# check bf16 (return `True` or `False`)
has_bf16 = cpuinfo.has_bf16
```
### 3. Summary of the change 

1. Add a `CPUInfo` class that can check the hardware information in any platform.
2. Nano needs to install `py-cpuinfo` as a default dependency now.
### 4. How to test?
- [ ] Unit test

### 5. New dependencies

- [ ] New Python dependencies
       - [py-cpuinfo](https://pypi.org/project/py-cpuinfo/)
         `py-cpuinfo` is already a dependency now because it is required by neural-compressor and openvino. But these 2 packages are optional, so we need to make `py-cpuinfo` as a default dependency of Nano to support this feature.


TheaperDeng (approved):
LGTM, I think we should use this dependency consistently in future development to make nano/chronos cross-platform friendly

yangw1234 (approved):
LGTM

PR #6365, opened 2022-10-31
Add gramine encrypting FS to base image.

Hanyu-Jin:
## Description

Enable gramine encrypting file system in base image.

### 1. Why the change?

Enable gramine encrypting FS for fast and convenient data storage.

### 2. User API changes

The default encrypted directory is /ppml/encrypted. You can create a file and write by gramine. The file will be encrypted automatically and can only be read by gramine.
You can also edit fs.mount in bash.manifast.template to add or delete the encrypted directories.

You can encrypt a file and decrypt it in one container or in another container on the same server. However, you cannot use this function between different servers.

### 3. Summary of the change 

Add a new directory, /ppml/encrypted, to base image in Dockerfile.
Mount /ppml/encrypted to gramine in bash.manifast.template.

### DIFF
--- a/ppml/base/Dockerfile
+++ b/ppml/base/Dockerfile
@@ -99,7 +99,8 @@ RUN env DEBIAN_FRONTEND=noninteractive apt-get update && \
     mkdir -p /root/.keras/datasets && \
     mkdir -p /root/.zinc && \
     mkdir -p /root/.m2 && \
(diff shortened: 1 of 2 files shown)


glorysdj (approved):
LGTM

gc-fu (approved):
LGTM

PR #6368, opened 2022-10-31
Nano: add multi-instance inference

MeouSker77:
## Description

Add multi-instance inference for Nano

### 1. Why the change?

Improve inference performance.

### 2. User API changes

The API should be:

```python

model = Model()

# use InferenceOptimizer.to_multi_instance to transform it
model = inference_opt.to_multi_instance(model, num_processes=4)

# predict DataLoader:
preds = model(loader)

# predict a list of inputs:
input_list = []
it = iter(loader)
for i in range(100):
    input_list.append(next(it))
preds = model(input_list)
```

it return a list of inference results, e.g.
```python
for pred in preds:
    do_something()
```

It will split the input dataloader or list and send them to multiple backend processes to run inference, then collect results from backend processes and return a list of inference results.

### 3. Summary of the change 

Add a `to_multi_instance` method for `InferenceOptimizer`, which can be used to transform a normal model to a multi-instance model.

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
+++ b/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
@@ -18,6 +18,7 @@ import torch
 from torch import nn
 import time
 from copy import deepcopy
+import multiprocessing as mp
 from typing import Dict, Callable, Tuple, Optional, List, Union, Sequence
 from torch.utils.data import DataLoader
 from torchmetrics.metric import Metric
@@ -38,6 +39,10 @@ from bigdl.nano.utils.inference.pytorch.dataset import RepeatDataset, remove_bat
 from bigdl.nano.utils.inference.pytorch.dataloader import\
     transform_multiple_input_dataloader_to_inc_mode
 from bigdl.nano.pytorch.utils import TORCH_VERSION_LESS_1_10, save_model, load_model
+from bigdl.nano.common.cpu_schedule import schedule_processors
+
+from .multi_instance import _MultiInstanceModel, _multi_instance_helper
+
 import traceback
 import warnings
 # Filter out useless Userwarnings
@@ -652,6 +657,31 @@ class InferenceOptimizer(BaseInferenceOptimizer):
                  precision(FP32/FP16/BF16/INT8).
         """
         return load_model(path, model)
+
+    @staticmethod
+    def to_multi_instance(model: nn.Module, num_processes: int) -> _MultiInstanceModel:
+        """
+        Transform a model to multi-instance inference model.
+        :param model: The model to transform.
+        :param num_processes: The number of processes which will be used.
+        :return: Model with multi-instance inference acceleration.
+        """
+        p_num = num_processes
+        mgr = mp.Manager()
+        send_queues = [mgr.Queue() for _ in range(p_num)]
+        recv_queues = [mgr.Queue() for _ in range(p_num)]
+        envs = schedule_processors(p_num)
+        ps = []
+        for i in range(p_num):
+            os.environ["KMP_AFFINITY"] = envs[i]['KMP_AFFINITY']
+            os.environ["OMP_NUM_THREADS"] = envs[i]['OMP_NUM_THREADS']
+
+            p = mp.Process(target=_multi_instance_helper,
+                           args=(model, send_queues[i], recv_queues[i]), daemon=True)
+            p.start()
+            ps.append(p)
+
+        return _MultiInstanceModel(model, ps, mgr, send_queues, recv_queues)
 
 
 def _signature_check(function):
--- a/python/nano/test/pytorch/tests/test_inference_pipeline_ipex.py
+++ b/python/nano/test/pytorch/tests/test_inference_pipeline_ipex.py
@@ -356,3 +356,18 @@ class TestInferencePipeline(TestCase):
                                metric=metric,
                                direction="max",
                                thread_num=4)
+
+    def test_multi_instance(self):
+        self.model.eval()
+        inference_opt = InferenceOptimizer()
+        multi_instance_model = inference_opt.to_multi_instance(self.model, num_processes=2)
+
+        input_data = list(map(lambda b: b[0], self.test_loader))
+
+        with torch.no_grad():
+            preds1 = multi_instance_model(input_data)
+            preds2 = [self.model(b) for b in input_data]
+
+            for (pred1, pred2) in zip(preds1, preds2):
+                np.testing.assert_allclose(pred1, pred2, atol=1e-4,
+                                           err_msg=f"\npred1: {pred1}\npred2: {pred2}\n")
--- a/python/nano/test/pytorch/tests/test_inference_pipeline_ipex.py
+++ b/python/nano/test/pytorch/tests/test_inference_pipeline_ipex.py
@@ -358,16 +358,21 @@ class TestInferencePipeline(TestCase):
                                thread_num=4)
 
     def test_multi_instance(self):
-        self.model.eval()
+        model = Net()
+        model.eval()
         inference_opt = InferenceOptimizer()
-        multi_instance_model = inference_opt.to_multi_instance(self.model, num_processes=2)
+        multi_instance_model = inference_opt.to_multi_instance(model, num_processes=2)
 
-        input_data = list(map(lambda b: b[0], self.test_loader))
+        test_loader = create_data_loader(self.data_dir, 1, self.num_workers, data_transform, subset=50, shuffle=False)
+        input_data = list(map(lambda b: b[0], test_loader))
 
         with torch.no_grad():
             preds1 = multi_instance_model(input_data)
-            preds2 = [self.model(b) for b in input_data]
+            print('one:', len(preds1))
+            preds2 = [model(b) for b in input_data]
+            print('two:', len(preds2))
 
-            for (pred1, pred2) in zip(preds1, preds2):
-                np.testing.assert_allclose(pred1, pred2, atol=1e-4,
-                                           err_msg=f"\npred1: {pred1}\npred2: {pred2}\n")
+        for (pred1, pred2) in zip(preds1, preds2):
+            np.testing.assert_allclose(pred1, pred2, atol=1e-4,
+                                        err_msg=f"\npred1: {pred1}\npred2: {pred2}\n")
+        print('three:', len(preds1), len(preds2))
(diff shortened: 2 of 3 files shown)


jason-dai:
Can we split at the dataloader level instead of tensor level?

MeouSker77:
> Can we split at the dataloader level instead of tensor level?

If we split the dataloader, the API may be:
```python
model = InferenceOptimizer.to_distributed(model, num_processes=4)

# start subprocesses and split dataloder in `predict` method
preds = model.predict(dataloader)
... # use preds
```
or
```python
model = InferenceOptimizer.to_distributed(model, num_processes=4)
# start subprocesses and split dataloader in `predict` method
# call `fn` with every output of `forward`
preds = model.predict(dataloader, fn)
```
or
similar to `TorchNano`


The first method requires we must have enough memory to store all inference output,
the second method is not easy to use,
the third method will cause each process can only see a part of the total data

jason-dai:
> > Can we split at the dataloader level instead of tensor level?
> 
> If we split the dataloader, the API may be:
> 
> ```python
> model = InferenceOptimizer.to_distributed(model, num_processes=4)
> 
> # start subprocesses and split dataloder in `predict` method
> preds = model.predict(dataloader)
> ... # use preds
> ```
> 
> or
> 
> ```python
> model = InferenceOptimizer.to_distributed(model, num_processes=4)
> # start subprocesses and split dataloader in `predict` method
> # call `fn` with every output of `forward`
> preds = model.predict(dataloader, fn)
> ```
> 
> or similar to `TorchNano`
> 
> The first method requires we must have enough memory to store all inference output, the second method is not easy to use, the third method will cause each process can only see a part of the total data
> 
> these three methods all need users change their current code

Can we do something like:
```python
model, loader = inference_opt.to_distributed(model, loader, num_processes=4)

with torch.no_grad():
    for X in loader:
        # X contains 4 batch of inputs, and no split_fn is needed
        pred_y = model(X) 
        ...
```

MeouSker77:
> Can we do something like:
> 
> ```python
> from bigdl.nano.pytorch import InferenceOptimizer
> inference_opt = InferenceOptimizer()
> model, loader = inference_opt.to_distributed(model, loader, num_processes=4)
> 
> with torch.no_grad():
>     for X in loader:
>         # X contains 4 batch of inputs, and no split_fn is needed
>         pred_y = model(X) 
>         ...
> ```

the shape of returned tensor may change if we takes this method, user may expect a tensor which dim0=n, but now its dim0=4n

MeouSker77:
And some model's `forward` may receive a dict as input, we do not know how to automatically generate a dict which contains 4 batch, only the user himself knows.

jason-dai:
> > Can we do something like:
> > ```python
> > from bigdl.nano.pytorch import InferenceOptimizer
> > inference_opt = InferenceOptimizer()
> > model, loader = inference_opt.to_distributed(model, loader, num_processes=4)
> > 
> > with torch.no_grad():
> >     for X in loader:
> >         # X contains 4 batch of inputs, and no split_fn is needed
> >         pred_y = model(X) 
> >         ...
> > ```
> 
> the shape of returned tensor may change if we takes this method, user may expect a tensor which dim0=n, but now its dim0=4n

`X` may contain a list of 4 batches, and then inside `model.forward` it can be split into 4 smaller tensors and passed to the individual processes

jason-dai:
> > > Can we do something like:
> > > ```python
> > > from bigdl.nano.pytorch import InferenceOptimizer
> > > inference_opt = InferenceOptimizer()
> > > model, loader = inference_opt.to_distributed(model, loader, num_processes=4)
> > > 
> > > with torch.no_grad():
> > >     for X in loader:
> > >         # X contains 4 batch of inputs, and no split_fn is needed
> > >         pred_y = model(X) 
> > >         ...
> > > ```
> > 
> > 
> > the shape of returned tensor may change if we takes this method, user may expect a tensor which dim0=n, but now its dim0=4n
> 
> `X` may contain a list of 4 batches, and then inside `model.forward` it can be split into 4 smaller tensors and passed to the individual processes

I don't think it matters; the original `loader` returns one batch, and we can simple call the new `loader` 4 times to get 4 batches, no matter what each batch looks like.

jason-dai:
> > > > Can we do something like:
> > > > ```python
> > > > from bigdl.nano.pytorch import InferenceOptimizer
> > > > inference_opt = InferenceOptimizer()
> > > > model, loader = inference_opt.to_distributed(model, loader, num_processes=4)
> > > > 
> > > > with torch.no_grad():
> > > >     for X in loader:
> > > >         # X contains 4 batch of inputs, and no split_fn is needed
> > > >         pred_y = model(X) 
> > > >         ...
> > > > ```
> > > 
> > > 
> > > the shape of returned tensor may change if we takes this method, user may expect a tensor which dim0=n, but now its dim0=4n
> > 
> > 
> > `X` may contain a list of 4 batches, and then inside `model.forward` it can be split into 4 smaller tensors and passed to the individual processes
> 
> I don't think it matters; the original `loader` returns one batch, and we can simply call the `loader` 4 times to get 4 batches, no matter what each batch looks like.

MeouSker77:
> > > > Can we do something like:
> > > > ```python
> > > > from bigdl.nano.pytorch import InferenceOptimizer
> > > > inference_opt = InferenceOptimizer()
> > > > model, loader = inference_opt.to_distributed(model, loader, num_processes=4)
> > > > 
> > > > with torch.no_grad():
> > > >     for X in loader:
> > > >         # X contains 4 batch of inputs, and no split_fn is needed
> > > >         pred_y = model(X) 
> > > >         ...
> > > > ```
> > > 
> > > 
> > > the shape of returned tensor may change if we takes this method, user may expect a tensor which dim0=n, but now its dim0=4n
> > 
> > 
> > `X` may contain a list of 4 batches, and then inside `model.forward` it can be split into 4 smaller tensors and passed to the individual processes
> 
> I don't think it matters; the original `loader` returns one batch, and we can simple call the new `loader` 4 times to get 4 batches, no matter what each batch looks like.

the shape of the returned `pred_y` may be different from that without multi-instance, user may assume the dim0 of `pred_y` is equal to the batch_size of the dataloader

jason-dai:
> > > > > Can we do something like:
> > > > > ```python
> > > > > from bigdl.nano.pytorch import InferenceOptimizer
> > > > > inference_opt = InferenceOptimizer()
> > > > > model, loader = inference_opt.to_distributed(model, loader, num_processes=4)
> > > > > 
> > > > > with torch.no_grad():
> > > > >     for X in loader:
> > > > >         # X contains 4 batch of inputs, and no split_fn is needed
> > > > >         pred_y = model(X) 
> > > > >         ...
> > > > > ```
> > > > 
> > > > 
> > > > the shape of returned tensor may change if we takes this method, user may expect a tensor which dim0=n, but now its dim0=4n
> > > 
> > > 
> > > `X` may contain a list of 4 batches, and then inside `model.forward` it can be split into 4 smaller tensors and passed to the individual processes
> > 
> > 
> > I don't think it matters; the original `loader` returns one batch, and we can simple call the new `loader` 4 times to get 4 batches, no matter what each batch looks like.
> 
> the shape of the returned `pred_y` may be different from that without multi-instance, user may assume the dim0 of `pred_y` is equal to the batch_size of the dataloader, but now the dim0 of `pred_y` is equal to 4*batch_size



> > > > > Can we do something like:
> > > > > ```python
> > > > > from bigdl.nano.pytorch import InferenceOptimizer
> > > > > inference_opt = InferenceOptimizer()
> > > > > model, loader = inference_opt.to_distributed(model, loader, num_processes=4)
> > > > > 
> > > > > with torch.no_grad():
> > > > >     for X in loader:
> > > > >         # X contains 4 batch of inputs, and no split_fn is needed
> > > > >         pred_y = model(X) 
> > > > >         ...
> > > > > ```
> > > > 
> > > > 
> > > > the shape of returned tensor may change if we takes this method, user may expect a tensor which dim0=n, but now its dim0=4n
> > > 
> > > 
> > > `X` may contain a list of 4 batches, and then inside `model.forward` it can be split into 4 smaller tensors and passed to the individual processes
> > 
> > 
> > I don't think it matters; the original `loader` returns one batch, and we can simple call the new `loader` 4 times to get 4 batches, no matter what each batch looks like.
> 
> the shape of the returned `pred_y` may be different from that without multi-instance, user may assume the dim0 of `pred_y` is equal to the batch_size of the dataloader, but now the dim0 of `pred_y` is equal to 4*batch_size

Yes, but I think it is simpler to understand (`pred_y` is just a list of 4 results); the upside is the user no longer needs to provide a split function. 

BTW, `pred_y` may not be a tensor anyway (it can be a list or dictionary itself), and you can also potentially leverage the batch size of the data loader.

MeouSker77:
Ok, I think it's better than current API, I'll change to it

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -653,6 +658,31 @@ def load(path, model: Optional[nn.Module] = None):
...
+        """
+        p_num = num_processes
+        mgr = mp.Manager()
+        send_queues = [mgr.Queue() for _ in range(p_num)]
+        recv_queues = [mgr.Queue() for _ in range(p_num)]
+        envs = schedule_processors(p_num)
+        ps = []
+        for i in range(p_num):
+            os.environ["KMP_AFFINITY"] = envs[i]['KMP_AFFINITY']
+            os.environ["OMP_NUM_THREADS"] = envs[i]['OMP_NUM_THREADS']
```
jason-dai, line 677: shall we restore the env for the main process afterwards?

jason-dai:
Need to apply it to https://github.com/intel-analytics/nuc-ai/tree/main/VideoStylization to check the speedup

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
MeouSker77: yes, fixed it in <https://github.com/intel-analytics/BigDL/pull/6653>

python/nano/test/pytorch/tests/test_inference_pipeline_ipex.py
```diff
@@ -356,3 +356,20 @@ def metric(model, data_loader):
...
                                thread_num=4)
+
+    def test_multi_instance(self):
+        model = Net()
+        model.eval()
+        inference_opt = InferenceOptimizer()
+        multi_instance_model = inference_opt.to_multi_instance(model, num_processes=2)
+
+        test_loader = create_data_loader(self.data_dir, 1, self.num_workers, data_transform, subset=50, shuffle=False)
+        input_data = list(map(lambda b: b[0], test_loader))
```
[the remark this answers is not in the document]
MeouSker77: No, `multi_instance_model` and `model` are the same expect `multi_instance_model` receives a list and return a list.

It can work well as long as the input of `multi_instance_model` is a list of batches.

Here I used all data only for convenience.

PR #6375, opened 2022-11-01
Chronos : modify ut of multi-processes tuning

rnwang04:
## Description

Fix two UT about multi-processes tuning according to feedback from NPG, to avoid bug caused by db duplication caused by abnormal interruption.

### How to test?
- [x] Unit test


TheaperDeng (approved):
LGTM for this PR, still we need to find out the hanging issue that is not reflected in our action.

rnwang04:
> LGTM for this PR, still we need to find out the hanging issue that is not reflected in our action.

Sure, will merge this PR first.

PR #6379, opened 2022-11-01
[ORCA] Refactor `get_model` to support optional `model_creator` with tf2 ray estimator

sgwhat:
## Description

In #6251, we have supported using `estimator.load` instead of `model_creator`. This PR mainly adds a supporting save model in this case.

### 1. Why the change?
Support saving model when `model_creator` is optional.

### 2. User API changes

```python
trainer = Estimator.from_keras(backend="ray")

trainer.load("/path/to/load_model.h5")

trainer.save("/path/to/save_model.h5")
```

### 3. Summary of the change 
When using `estimator.load` instead of `model_creator`, the `estimator.save` will load the model on driver with updating `model_weights` and `optimizer_weights`.

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/orca/src/bigdl/orca/learn/tf2/ray_estimator.py
+++ b/python/orca/src/bigdl/orca/learn/tf2/ray_estimator.py
@@ -591,19 +591,17 @@ class TensorFlow2Estimator(OrcaRayEstimator):
                options for loading from SavedModel.
 
         """
-        self.load_path = filepath
-
-        params = dict(
+        self.load_params = dict(
             filepath=filepath,
             custom_objects=custom_objects,
             compile=compile,
             options=options
         )
         if is_local_path(filepath):
-            ray.get([worker.load_model.remote(**params)
+            ray.get([worker.load_model.remote(**self.load_params)
                      for worker in self.remote_workers])
         else:
-            ray.get([worker.load_remote_model.remote(**params)
+            ray.get([worker.load_remote_model.remote(**self.load_params)
                      for worker in self.remote_workers])
 
     def save_weights(self, filepath, overwrite=True, save_format=None, options=None):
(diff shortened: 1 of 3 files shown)


python/orca/src/bigdl/orca/learn/tf2/ray_estimator.py
```diff
@@ -481,15 +482,13 @@ def predict(self, data, batch_size=None, verbose=1,
 
         return result
 
-    def get_model(self, sample_input=None):
+    def get_model(self):
```
hkvision, line 485: I think we need to keep this API in case users want to get the model to driver?
Just get the model of the first worker and return?

python/orca/src/bigdl/orca/learn/tf2/ray_estimator.py
```diff
@@ -552,26 +551,16 @@ def save(self,
...
+        params = dict(
+            filepath=filepath,
+            overwrite=overwrite,
+            include_optimizer=include_optimizer,
+            save_format=save_format,
+            signatures=signatures,
+            options=options
+        )
+        model_refs = [w.save_model.remote(**params) for w in self.remote_workers]
+        ray.get(model_refs[0])
```
hkvision, line 563: Just use `ray.get(self.remote_workers[0].save_model.remote(**params))`?

python/orca/src/bigdl/orca/learn/tf2/ray_estimator.py
```diff
@@ -619,25 +608,14 @@ def save_weights(self, filepath, overwrite=True, save_format=None, options=None)
...
-            finally:
-                shutil.rmtree(temp_dir)
+        params = dict(
+            filepath=filepath,
+            overwrite=overwrite,
+            save_format=save_format,
+            options=options
+        )
+        weights_refs = [w.save_weights.remote(**params) for w in self.remote_workers]
+        ray.get(weights_refs[0])
```
hkvision, line 618: same as above

python/orca/src/bigdl/orca/learn/tf2/ray_estimator.py
sgwhat: only get model states like weights? Since ray worker cannot deserialize a tf model.

hkvision: oh really? Then just rename to `get_weights`?

hkvision: Or do you think we need to keep this API?

sgwhat: I don't think we need this API, but `get_model` is a static method in `OrcaRayEstimator`. I prefer we rewrite this API as getting model states?

hkvision: Is it possible to do like this:
- If model_creator is defined, do as what we did previously
- If model_creator is not specified, we fetch model load path to local and load + get states from the first worker + write the states to loaded model?

python/orca/src/bigdl/orca/learn/tf2/ray_estimator.py
sgwhat: We need to save models across all workers, as keras doc mentioned: `You need to save and checkpoint on all workers instead of just one. This is because variables whose synchronization=ON_READ triggers aggregation during saving. It's recommended to save to a different path on each worker to avoid race conditions. Each worker saves the same thing.`

PR #6383, opened 2022-11-01
[ORCA] Support continuous training in tf2 ray estimator

sgwhat:
## Description
This PR mainly supports resuming training after using `estimator.load` to load a tf model.

### 1. Why the change?
In the previous design, we only support resuming prediction. 

### 2. User API changes
```python
trainer = estimator.from_keras(backend="ray")

trainer.load("/path/to/model.h5")

#resume training
trainer.fit(train_df, epochs, bactch_size, steps_per_epoch, feature_cols, label_cols)
```

### 3. Summary of the change 
Complete the user training pipeline with loading model.

### 4. How to test?
- [x] Unit test


hkvision (approved):

python/orca/src/bigdl/orca/learn/tf2/tf_runner.py
```diff
@@ -492,8 +492,23 @@ def predict(self, data_creator, batch_size, verbose, steps, callbacks, data_conf
...
+                file_name = os.path.basename(filepath)
+                temp_path = os.path.join(tempfile.mkdtemp(), file_name)
+                if is_file(filepath):
+                    # h5 format
+                    get_remote_file_to_local(filepath, temp_path)
+                else:
+                    # savemodel format
+                    if os.path.exists(temp_path):
+                        os.makedirs(temp_path)
+                    get_remote_dir_to_local(filepath, temp_path)
```
hkvision, line 509: For this code piece can we make it as a utility?
Seems many places use this and there's much duplication? Is it convenient to maintain in the future?

hkvision:
LGTM except the comments above. But can do in a following PR.

python/orca/src/bigdl/orca/learn/tf2/tf_runner.py
sgwhat: sure we could do this later.

sgwhat:
> LGTM except the comments above. But can do in a following PR.

will do this in a following PR.

PR #6387, opened 2022-11-01
Chronos : modify workflow of chronos based on new inference option

rnwang04:
## Description

As Nano support new inference installation option, we can update chronos workflow for updated version dependency.

### How to test?
- [x] Application test


plusbang (commented):
LGTM, but I think action of Chronos how-to guides may also need to be modified.

rnwang04:
> LGTM, but I think action of Chronos how-to guides may also need to be modified.

Done

PR #6405, opened 2022-11-02
Add decomposition and Normalization for tcn (extension: Ntcn, Dtcn, NDtcn)

devWangBin:
### 1. Why the change?

Despired by the methods mentioned in [Are Transformers Effective for Time Series Forecasting?](https://arxiv.org/abs/2205.13504), adding new extensions to tcn model. And the experiment results on the three datasets (electricity, nyc_taxi and NWDAF_INTEL) demonstrate the effectiveness of the added extensions. 

### 2. User API changes
Add new option "**model_set**" to bigdl.chronos.forecaster.TCNForecaster.__init__()
**model_set**: str, must be selected from "tcn", "Ntcn", "Dtcn" and "NDtcn". 1) Choose "tcn" to use the original model. 2) "Ntcn" applies a sample Normalization method which first subtracts the input by the last value of the sequence. Then, the input goes through the tcn layer, and the subtracted part is added back before making the final prediction. 3) "Dtcn" applies a Decomposition scheme that first decomposes a raw data input into a trend component by a moving average kernel and a remainder (seasonal) component. Then, two tcn models are applied to each component and we sum up the two features to get the final prediction. 4) "NDtcn" applies the Normalization and the Decomposition methods at the same time. The value defaults to "tcn".

### 3. Summary of the change 
Add decomposition and Normalization for tcn (extension: Ntcn, Dtcn, NDtcn)

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/python/chronos/src/bigdl/chronos/forecaster/tcn_forecaster.py
+++ b/python/chronos/src/bigdl/chronos/forecaster/tcn_forecaster.py
@@ -55,7 +55,8 @@ class TCNForecaster(BasePytorchForecaster):
                  seed=None,
                  distributed=False,
                  workers_per_node=1,
-                 distributed_backend="ray"):
+                 distributed_backend="ray",
+                 model_set='tcn'):
         """
         Build a TCN Forecast Model.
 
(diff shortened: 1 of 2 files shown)


plusbang (commented):
LGTM

PR #6407, opened 2022-11-02
Fix aggregating stats of torch workers

hkvision:


python/orca/src/bigdl/orca/learn/pytorch/pytorch_ray_estimator.py
```diff
@@ -612,18 +612,30 @@ def shutdown(self, force=False):
...
+        return res_stats
+
     def _process_stats(self, worker_stats):
         stats = {
             "num_samples": sum(
                 stats.pop("num_samples", np.nan) for stats in worker_stats)
         }
 
-        for stat_key in worker_stats[0]:
-            if isinstance(worker_stats[0], numbers.Number):
```
hkvision, line 622: This is wrong, should get the value instead of isinstance on the dict.
Under such code, will always only use the stats of the first worker.
Also profile is not considered, which the value is a dict.

sgwhat (approved):

hkvision:
@sgwhat Take a further look? Put process_stats to utils so that can be used for both ray and pyspark backends. Also add unit tests.

hkvision:
Found a bug here https://github.com/intel-analytics/BigDL/issues/6573 and thus num_samples is currently not tested in the unit test for pyspark backend.

PR #6408, opened 2022-11-02
[Orca] Pytorch Ray Estimator and Runner Refactor Design

leonardozcm:
### 1. Why the change?

In order to :
**1. defined the common behaviors between various estimator and corresponding runner clearly.**
**2. reuse duplicated code.**

In this work we refactor the RayEstimator and the Runner. The first step is to apply this design to mmcv and pytorch, and in the further developing we can follow this pattern, such as supporting pytorch-lightning.

**Note**: The unformed development code is currently under the `experimental` folder, and we will migrate them to the official code area when it is ready.

### 2. User API changes
No

### 3. Summary of the change 

We refactor the abstract class of **torch-based ray-backend** estimator called `BaseRayEstimator` in **base_ray_estimator.py**, and abstract runner class called `BaseRayRunner` in **base_ray_runner.py**

The methods that must be override in an estimator are `fit, predict, evaluate, get_model`, and every estimator's consensuses on RayWorker's RPC are following APIs:

1. Trainer indicates worker's main tasks
```python
class Trainer(metaclass=ABCMeta):    
    @abstractmethod
    def train_epochs(self, **kwargs):
        pass

    @abstractmethod
    def predict(self, **kwargs):
        pass

    @abstractmethod
    def validate(self, **kwargs):
        pass
```

2. ModelIO takes charge of **load and save** stuffs.
```python
class ModelIO(metaclass=ABCMeta):
    def __init__(self):
        self.rank = -1
        self.logger = None

    @abstractmethod
    def get_state_dict(self):
        pass

    @abstractmethod
        pass

    @abstractmethod
    def _save_checkpoint(self, filepath, save_weights_only=False):
        """Save checkpoint."""
        pass
    
    ...
```

3. LifeCycleManager manage the entire setup and shutdown process.
```python
class LifeCycleManager(metaclass=ABCMeta):

    def setup(self, cores_per_node):
        import torch
        torch.set_num_threads(cores_per_node)

    def setup_torch_distribute(self, tcp_store_host, tcp_store_port, world_rank,
                               world_size):
        self._init_torchDDP(tcp_store_host, tcp_store_port, world_rank,
                               world_size)

        self.setup_components()

    ...

    @abstractmethod
    def shutdown(self):
        """Attempts to shut down the worker."""
        pass
```
And an abstract runner could be simply defined as:
```python
class BaseRayRunner(Trainer, LifeCycleManager, ModelIO, metaclass=ABCMeta):
```
A Runner implements all abstract methods mentioned above, is proper to be utilized by an estimator.

In this way, the developer's work can mainly focus on **adapting these behaviors** in the future.

Following work steps:
1. apply this on mmcv estimator
2. refactor torch ray estimator
3. more third-party framework adapting...

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

@jason-dai @hkvision @yangw1234 @sgwhat @PatrickkZ pls take a look?


yangw1234:
Great work! 
But I personally think it is a little hard to reason about such a large scale change. Is it possible to progressively change the code base? 
For example, first refactor all `save` and `load` methods into `ModelIO`. And then `LifeCycleManager`.

jason-dai:
> Great work! But I personally think it is a little hard to reason about such a large scale change. Is it possible to progressively change the code base? For example, first refactor all `save` and `load` methods into `ModelIO`. And then `LifeCycleManager`.

I agree; try to break the PR into several smaller ones.

leonardozcm:
Yes, we will refactor TorchRayEstimator step-by-step.

hkvision:
The overall design looks good. Several comments/questions:

- Is Estimator in ray_estimator.py needed? BaseRayEstimator can directly extend BaseEstimator?
- utils.py is duplicate?
- runner_cls should not be an argument for Estimator init? Users are not supposed to modify the runner cls right?
- Will MMCV EpochBasedRunner.run return anything? Seems no? Then how to align with other estimator fit outputs?
- The class name Trainer is confusing (especially after when introducing PyTorch Lightning). Better change a name. And do you think it can combine with TrainingOperator?
- How is the Callback class used? Would it be combined with Trainer?
- Better test worker_per_node > 1

leonardozcm:
> * Is Estimator in ray_estimator.py needed? BaseRayEstimator can directly extend BaseEstimator?

No. Actually BaseEstimator seems to do nothing when inheriting Estimator class.

> * utils.py is duplicate?

yes, we will merge them when turns to offical code.

> * runner_cls should not be an argument for Estimator init? Users are not supposed to modify the runner cls right?

Yes, but all estimators are built from one api which is `Estimator.from_...`, so I think we will not expose runner_cls in `from_mmcv(*args, **kwargs)` 

> * Will MMCV EpochBasedRunner.run return anything? Seems no? Then how to align with other estimator fit outputs?

we keep this the same with other runner, now it does return somthing like NUM_SAMPLES:
https://github.com/intel-analytics/BigDL/pull/6507 

> * The class name Trainer is confusing (especially after when introducing PyTorch Lightning). Better change a name. And do you think it can combine with TrainingOperator?

Yes, I agree. TrainingOperator is ok. BTW, I think the combination of runner and trainingoperator is a bit bloated. Is it possible to expose the operation of the original trainingoperator to runner?

> * How is the Callback class used? Would it be combined with Trainer?

Callback class is related to our pytorch customized trainning, but I haven't planned how to schedule this yet. Just a placeholder here.

> * Better test worker_per_node > 1

sure

hkvision:
May think and discuss the following in the next stage:
- How to abort the original TrainingOperator
- How to better define the callback usage for users
- Maybe combine ModelIO/Liftcycle one by one with the existing TorchRunner for code merge?

PR #6412, opened 2022-11-03
Orca: add batch_size arguments to model.fit()

lalalapotter:
## Description
This PR is to add `batch_size` arguments while calling function `model.fit()` in tensorflow 2.x estimator, which is to support the mlflow tracking.

### DIFF
--- a/python/orca/src/bigdl/orca/learn/tf2/spark_runner.py
+++ b/python/orca/src/bigdl/orca/learn/tf2/spark_runner.py
@@ -305,7 +305,8 @@ class SparkRunner:
                             initial_epoch=initial_epoch,
                             steps_per_epoch=steps_per_epoch,
                             validation_steps=validation_steps,
-                            validation_freq=validation_freq)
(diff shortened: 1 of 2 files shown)


hkvision:
Comment: in our implementation, we input tf dataset which is already batched. Adding the batch size is mainly used for mlflow tracking, the batch size is the same as the batch in tf dataset, which should have no issue.

hkvision (approved):

jason-dai:
> the batch size is the same as the batch in tf dataset, which should have no issue.

How do you enforce that?

hkvision:
> > the batch size is the same as the batch in tf dataset, which should have no issue.
> 
> How do you enforce that?

The batch size argument in fit will be used to batch the tf dataset, will also be fed into fit. They will always be the same. Should have no issue?

hkvision:
Seems for tf-distributed backend, local batch size for each worker is input_batch_size / worker_size?
But for pytorch, the input_batch_size is the batch size for every single worker?

PR #6424, opened 2022-11-03
move requirements.txt to actions python-requirements

pinggao187:
## Description
move
https://github.com/intel-analytics/BigDL/blob/main/python/orca/dev/example/requirements-ray.txt
https://github.com/intel-analytics/BigDL/blob/main/python/orca/dev/example/requirements-jep.txt
to 
https://github.com/intel-analytics/BigDL/tree/main/.github/actions/python-requirements


glorysdj (approved):
LGTM

Issue #6389, opened 2022-11-01
Nano: Inconsistent dataloader format support for multiple input between `quantize` and `trace` in pytorch `InferenceOptimizer`


TheaperDeng:
## Overview
Multiple input (i.e. `model(x1, x2, ...)` could be a general case for many networks (e.g. GNN, ...). Currently, we support multiple input model both in `quantize` and `trace`, while in different format.

## Current support
For `trace`, we support:
```python
class MultipleTensorDataset(Dataset):
    ...
    def __getitem__(self, idx):
        return self.x1[idx], self.x2[idx], self.y[idx]
```
We call it "trace-mode dataloader" in later paragraphs.

For `quantize`, we support:
```python
class MultipleTensorDataset(Dataset):
    ...
    def __getitem__(self, idx):
        return (self.x1[idx], self.x2[idx]), self.y[idx]
```
We call it "inc dataloader" in later paragraphs.

## what does it cause?
We now recommend users to try `InferenceOptimizer.optimize` first as a unified API
- if users use "trace-mode dataloader", it will cause "fail to convert" error for int8 related items.
- if users use "inc dataloader", there will be a major error so that the application will fail.

## How to solve this?
I believe it's fairly easy to support the transformation between these 2 formats, and we could easily detect which format users are using by inspecting the signature of model's forward.

We could **either**
- only support "trace-mode dataloader" in all APIs and transform it to "inc dataloader" when we need to quantize a model **(my choice)**
- support both formats and transform intelligently between these 2 formats according to the usage

## What I may miss
what about the dict input?

PR #6434, opened 2022-11-03
Nano: add dataloader transformation for `quantize` method

TheaperDeng:
### 1. Why the change?
https://github.com/intel-analytics/BigDL/issues/6389

### 2. User API changes
Nothing is changed

### 3. Summary of the change 
for inc based quantization in `InferenceOptimizer.quantize`, a pre-step is done to detect and transform the dataloader from 
```python
class MultipleTensorDataset(Dataset):
    ...
    def __getitem__(self, idx):
        return self.x1[idx], self.x2[idx], self.y[idx]
```
to
```python
class MultipleTensorDataset(Dataset):
    ...
    def __getitem__(self, idx):
        return (self.x1[idx], self.x2[idx]), self.y[idx]
```
### 4. How to test?
- [x] Unit test

Need a new how-to page to guide users about the supported dataloader.

### DIFF
--- a/python/nano/src/bigdl/nano/utils/inference/pytorch/dataloader.py
+++ b/python/nano/src/bigdl/nano/utils/inference/pytorch/dataloader.py
@@ -20,7 +20,8 @@ from copy import deepcopy
 
 
 def transform_multiple_input_dataloader_to_inc_mode(model, dataloader):
-    if _need_dataloader_type_transformation(model, dataloader):
+    need_transformation, forward_args_len = _need_dataloader_type_transformation(model, dataloader)
+    if need_transformation:
         # define a decorator to change multiple inputs to 2 items
         def tuple_collate_fn_wrapper(func, forward_args_len):
             def collate_fn(batch):
@@ -35,10 +36,6 @@ def transform_multiple_input_dataloader_to_inc_mode(model, dataloader):
 
         # deepcopy the dataloader so that the transformation will not pollute the original one
         new_dataloader = deepcopy(dataloader)
-
-        # get forward method's parameter number
-        forward_args = get_forward_args(model)
-        forward_args_len = len(forward_args)
 
         # add collate fn to the dataloader
         new_dataloader.collate_fn = tuple_collate_fn_wrapper(new_dataloader.collate_fn,
@@ -64,5 +61,5 @@ def _need_dataloader_type_transformation(model, dataloader):
     input_sample = next(iter(dataloader))
     if isinstance(input_sample[0], Sequence):
         if len(input_sample[0]) == forward_args_len:
-            return False
-    return True
+            return False, forward_args_len
+    return True, forward_args_len
--- a/python/nano/src/bigdl/nano/utils/inference/pytorch/dataloader.py
+++ b/python/nano/src/bigdl/nano/utils/inference/pytorch/dataloader.py
@@ -55,7 +55,7 @@ def _need_dataloader_type_transformation(model, dataloader):
     # a special case is 0, this means *args is used in
     # users' forward method, we will also skip it as well
     if forward_args_len <= 1:
-        return False
+        return False, forward_args_len
 
     # check if a dataloader has met inc format
     input_sample = next(iter(dataloader))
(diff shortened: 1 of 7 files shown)


python/nano/src/bigdl/nano/utils/inference/pytorch/dataloader.py
```diff
@@ -0,0 +1,68 @@
...
+                else:
+                    # if multiple y are provided
+                    return tuple(res[:forward_args_len]), tuple(res[forward_args_len:])
+            return collate_fn
+
+        # deepcopy the dataloader so that the transformation will not pollute the original one
+        new_dataloader = deepcopy(dataloader)
+
+        # get forward method's parameter number
+        forward_args = get_forward_args(model)
```
rnwang04, line 40: `forward_args = get_forward_args(model)`
This occurs two times. Maybe move it before 
`if _need_dataloader_type_transformation(model, dataloader):` and pass them into _need_xxx

python/nano/test/pytorch/tests/test_inference_pipeline_ipex.py
```diff
@@ -259,3 +271,34 @@ def test_get_model_with_method_name(self):
             assert isinstance(model, PytorchIPEXJITModel)
         except:
             pass
+
+    def test_multiple_input_dataloader(self):
+        # will not run this test if torch < 1.10
+        if TORCH_VERSION_LESS_1_10:
```
rnwang04, line 277: which skip for this version？

python/nano/test/pytorch/tests/test_inference_pipeline_ipex.py
```diff
@@ -259,3 +271,34 @@ def test_get_model_with_method_name(self):
...
+        InferenceOptimizer.quantize(net,
+                                    accelerator="onnxruntime",
+                                    calib_dataloader=dataloader)
+
+        # int8-onnxruntime
+        InferenceOptimizer.trace(net,
+                                 accelerator="onnxruntime",
+                                 input_sample=dataloader)
+
+        # int8-openvino
```
rnwang04, line 300: here just openvino.
Maybe add an int8-openvino test case.

python/nano/src/bigdl/nano/utils/inference/pytorch/dataloader.py
```diff
@@ -0,0 +1,68 @@
...
+        new_dataloader.collate_fn = tuple_collate_fn_wrapper(new_dataloader.collate_fn,
+                                                             forward_args_len)
+
+        return new_dataloader
+    return dataloader
+
+
+def _need_dataloader_type_transformation(model, dataloader):
+    # get forward method's parameter number
+    forward_args = get_forward_args(model)
```
rnwang04, line 53: get_forward_args() has a potential risk for kwargs.
For example, the forward function of model looks like : 
`model(x, mask=None)`
and dataloader returns `(x, y)`.
This may cause bug, and you maybe add a ut for this case.

rnwang04 (commented):
others LGTM.

python/nano/test/pytorch/tests/test_inference_pipeline_ipex.py
TheaperDeng: for 1.9 ipex, some strange issue might be raised here, since we will gradually close the support for 1.9 so I just skip it here :)

python/nano/test/pytorch/tests/test_inference_pipeline_ipex.py
TheaperDeng: openvino int8 for multiple input case could not run through due to a reason other than dataloader's format, will raise a separate issue and pr for this.

python/nano/src/bigdl/nano/utils/inference/pytorch/dataloader.py
TheaperDeng: change to run `get_forward_args` only 1 time

python/nano/src/bigdl/nano/utils/inference/pytorch/dataloader.py
TheaperDeng: It works well, just add a ut for this

Issue #6006, opened 2022-09-29
Nano: Tensorflow trace(onnx/openvino) api discussion


TheaperDeng:
Well the question here is easy.

Should we put it as before in `model.quantize` and `model.trace`? or we should move all the previous and newly added features to `InferenceOptimizer`'s `trace` and `quantize`.

I would prefer the second choice so that in future we could
1. Use `patch_nano` to handle (nearly) all the training speed-up.
2. Use `InferenceOptimizer` to handle all the inference speed-up.

TheaperDeng:
@rnwang04 @MeouSker77 any thoughts?

TheaperDeng:
> One question here, if we put all the previous and newly added tf features to `InferenceOptimizer`'s `trace` and `quantize`, it seems that this will lead to the tight coupling between Pytorch and Tensorflow ? Is this against our existing installation method: `pip install bigdl nano [tensorflow]` and `pip install bigdl nano [pythoch]`?

Yeah, good question. We may use something like: https://github.com/intel-analytics/BigDL/blob/main/python/chronos/src/bigdl/chronos/utils.py#L39-L93 here to deal with the tf/torch dependency control

MeouSker77:
> Yeah, good question. We may use something like: https://github.com/intel-analytics/BigDL/blob/main/python/chronos/src/bigdl/chronos/utils.py#L39-L93 here to deal with the tf/torch dependency control

It's a possible solution, or maybe we can just create a new `InferenceOptimizer` for tensorflow?

jason-dai:
Maybe we should add an `InferenceOptimizer` for Keras too.

PR #6458, opened 2022-11-07
Nano : support InferenceOptimizer for keras

rnwang04:
## Description

Support InferenceOptimizer for keras.

### 1. Why the change?

We design to provide a similar InferenceOptimizer interface in keras as in Pytorch. https://github.com/intel-analytics/BigDL/issues/6006

### 2. User API changes
```
model = ...
model = NanoModel(inputs=model.inputs, outputs=model.outputs)
# prepare dataset
train_dataset = tf.data.Dataset.from_tensor_slices((train_examples, train_labels))
val_dataset = tf.data.Dataset.from_tensor_slices((val_examples, val_labels))
# prepare optimizer
opt = InferenceOptimizer()
from tensorflow.keras.metrics import CategoricalAccuracy
opt.optimize(model=model,
             training_data=train_dataset,
             validation_data=val_dataset,
             batch_size=32,
             metric=CategoricalAccuracy())
```

### 3. Summary of the change 

Support InferenceOptimizer for keras.

### 4. How to test?
- [x] Unit test
- [x] Application test

### Output sample
image

### DIFF
--- a/python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
+++ b/python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
@@ -1,0 +1,249 @@
+#
+# Copyright 2016 The BigDL Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import os
+import time
+import tensorflow as tf
+from typing import Dict, Optional, List
+from bigdl.nano.utils.inference.common.base_optimizer import BaseInferenceOptimizer
+from bigdl.nano.utils.inference.common.checker import available_acceleration_combination
+from bigdl.nano.utils.inference.common.utils import AccelerationOption,\
+    throughput_calculate_helper, format_optimize_result
+from bigdl.nano.tf.keras import Model as NanoModel
+from bigdl.nano.utils.log4Error import invalidInputError
+from tensorflow.keras import Model as Model
+from tensorflow.data import Dataset
+from tensorflow.keras.metrics import Metric
+
+
+class InferenceOptimizer(BaseInferenceOptimizer):
+
+    # acceleration method combinations, developers may want to register some new
+    # combinations here
+    ALL_INFERENCE_ACCELERATION_METHOD = \
+        {
+            "original": AccelerationOption(),
+            "int8": AccelerationOption(inc=True),
+            "openvino_fp32": AccelerationOption(openvino=True),
+            # "openvino_int8": AccelerationOption(openvino=True, pot=True),
+            "onnxruntime_fp32": AccelerationOption(onnxruntime=True),
+            # "onnxruntime_int8": AccelerationOption(onnxruntime=True, inc=True),
+        }
+
+    def optimize(self, model: Model,
+                 training_data: Dataset,
+                 validation_data: Optional[Dataset] = None,
+                 batch_size: int = 1,
+                 metric: Optional[Metric] = None,
+                 direction: str = "max",
+                 thread_num: Optional[int] = None,
+                 logging: bool = False,
+                 latency_sample_num: int = 100,
+                 includes: Optional[List[str]] = None,
+                 excludes: Optional[List[str]] = None) -> None:
+        '''
+        This function will give all available inference acceleration methods a try
+        and record the latency, accuracy and model instance inside the Optimizer for
+        future usage. All model instance is setting to eval mode.
+
+        The available methods are "original", "openvino_fp32", "onnxruntime_fp32", "int8".
+
+        :param model: A keras.Model to be optimized
+        :param training_data: A tf.data.Dataset object for for training dataset.
+               Users should be careful with this parameter since this dataloader
+               might be exposed to the model, which causing data leak.
+        :param validation_data: (optional) A tf.data.Dataset object for accuracy evaluation.
+               This is only needed when users care about the possible accuracy drop.
+        :param metric: (optional) A tensorflow.keras.metrics.Metric object which is used for
+               calculating accuracy.
+        :param direction: (optional) A string that indicates the higher/lower
+               better for the metric, "min" for the lower the better and "max" for the
+               higher the better. Default value is "max".
+        :param thread_num: (optional) a int represents how many threads(cores) is needed for
+               inference.
+        :param logging: whether to log detailed information of model conversion.
+               Default: False.
+        :param latency_sample_num: (optional) a int represents the number of repetitions
+               to calculate the average latency. The default value is 100.
+        :param includes: (optional) a list of acceleration methods that will be included in the
+               search. Default to None meaning including all available methods. "original" method
+               will be automatically add to includes.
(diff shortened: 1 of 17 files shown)


python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
```diff
@@ -0,0 +1,251 @@
...
+
+class InferenceOptimizer(BaseInferenceOptimizer):
+
+    # acceleration method combinations, developers may want to register some new
+    # combinations here
+    ALL_INFERENCE_ACCELERATION_METHOD: Dict = \
+        {  # type: ignore
+            "original": AccelerationOption(),
+            "int8": AccelerationOption(inc=True),
+            "openvino_fp32": AccelerationOption(openvino=True),
```
yangw1234, line 40: Is it possible to refactor the optimization logic to a class similar to this `AccelerationOption` so that it is clear which options does what e.g.
```python
class Acceleration:
    def __init__(inc:bool, openvion:bool, ...):
        pass
    
    def optimize(model, **other_kwargs):
        # optimization logic
        return optimized_model
```

yangw1234: I think it is not urgent though.

rnwang04: Thanks, will consider this after basic logic is ready and try some refactor.

python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
```diff
@@ -0,0 +1,262 @@
...
+                    model(sample)
+                try:
+                    result_map[method]["latency"], status =\
+                        throughput_calculate_helper(latency_sample_num, baseline_time,
+                                                    func_test, acce_model, input_sample)
+                    if status is False and method != "original":
+                        result_map[method]["status"] = "early stopped"
+                        continue
+                except Exception as e:
+                    result_map[method]["status"] = "fail to forward"
```
yangw1234, line 213: nitpick: maybe also include the reason for failing

python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
```diff
@@ -0,0 +1,262 @@
...
+        else:
+            # test whether accuracy calculation works later
+            # make sure dataset don't have batch
+            batched_validation_data = validation_data.batch(batch_size)
+            self._calculate_accuracy = True
+
+        if os.getenv('OMP_NUM_THREADS') is not None:
+            default_threads: int = int(os.getenv('OMP_NUM_THREADS'))  # type: ignore
+        else:
+            # TODO: how to get and control thread num in tf?
```
yangw1234, line 146: https://www.tensorflow.org/api_docs/python/tf/config/threading

image

python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
```diff
@@ -0,0 +1,262 @@
...
+        This function will give all available inference acceleration methods a try
+        and record the latency, accuracy and model instance inside the Optimizer for
+        future usage. All model instance is setting to eval mode.
+
+        The available methods are "original", "openvino_fp32", "onnxruntime_fp32", "int8".
+
+        :param model: A keras.Model to be optimized
+        :param training_data: An unbatched tf.data.Dataset object for for training dataset.
+               Users should be careful with this parameter since this dataloader
+               might be exposed to the model, which causing data leak.
```
yangw1234, line 98: I am not quite sure I understand the document. Maybe explaining more? (For example, how to be careful?)

python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
```diff
@@ -0,0 +1,262 @@
...
+        else:
+            # TODO: how to get and control thread num in tf?
+            default_threads = None  # type: ignore
+        thread_num = default_threads if thread_num is None else int(thread_num)  # type: ignore
+
+        result_map: Dict[str, Dict] = {}
+
+        batched_training_data = training_data.batch(batch_size)
+        input_sample = next(iter(batched_training_data))
+        # TODO: how to obtain input from output of training_data
```
yangw1234, line 154: Maybe we can use the same input requirement as `tf.keras.model.fit` (https://www.tensorflow.org/api_docs/python/tf/keras/Model#fit), batched or unbatched version.

python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
rnwang04: done

python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
rnwang04: I have tried this locally, and get_xxx_threads() return 0 after we set `OMP_NUM_THREADS`.
And we can't set_xxx_threads in optimize(), it will report :
image

python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
rnwang04: Have update this note.

python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
rnwang04: This maybe a way to obtain more accurate input. But it will occur different input mode from torch.InferenceOptimizer.
How do you think about this ? @TheaperDeng @MeouSker77

MeouSker77: I think it's better to be consistent with the common usage of tensorflow users (i.e. be same as `tf.keras.model.fit`). But the inc, openvino and onnx quantization all requires a dataset which produces both `x` and `y`, so we have to do some changes if we use the same input requirements as `tf.keras.model.fit`. But I think these changes are acceptable.

TheaperDeng: also agree to change the parameter to x, y here. We may do it in later PR.

rnwang04: well, will change the parameter to x, y in next PR.

python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
```diff
@@ -0,0 +1,266 @@
...
+
+        The available methods are "original", "openvino_fp32", "onnxruntime_fp32", "int8".
+
+        :param model: A keras.Model to be optimized
+        :param training_data: An unbatched tf.data.Dataset object which is used for training.
+                              This dataset will be used as calibration dataset for
+                              Post-Training Static Quantization (PTQ), as well as be used for
+                              generating input_sample to calculate latency.
+                              To avoid data leak during calibration, please use training
+                              dataset as much as possible.
```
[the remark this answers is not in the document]
rnwang04: thanks for reminder, will fix this in next PR.

python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
```diff
@@ -0,0 +1,266 @@
...
+                 logging: bool = False,
+                 latency_sample_num: int = 100,
+                 includes: Optional[List[str]] = None,
+                 excludes: Optional[List[str]] = None) -> None:
+        '''
+        This function will give all available inference acceleration methods a try
+        and record the latency, accuracy and model instance inside the Optimizer for
+        future usage. All model instance is setting to eval mode.
+
+        The available methods are "original", "openvino_fp32", "onnxruntime_fp32", "int8".
```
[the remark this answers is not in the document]
rnwang04: I am not sure the known bugs mean what?

PR #6470, opened 2022-11-07
cicd: cancel previous runs in the PR when you push new commits

liu-shaojun:
### 1. Why the change?
According to @leonardozcm 's comment, when a developer pushes several times to one PR, which will create GitHub action jobs for each commit, while developers only want the latest commit to be tested. 

This submission is to cancel the previous running workflow from the same PR when a new workflow is triggered by a new commit. This also reduces the workload of the github runners, making the runner available for pending jobs as soon as possible.

### 4. How to test?
This has been tested manually. When push a new commit to this PR, the previous workflow will be cancelled.

### 5. New dependencies

- [ ] New Python dependencies
       - Dependency1 
       - Dependency2
       - ...
       - Dependency1 and license1
       - Dependency2 and license2
       - ...

### DIFF
--- a/.github/workflows/PR_validation.yml
+++ b/.github/workflows/PR_validation.yml
@@ -34,7 +34,7 @@ permissions:
 jobs:
   changes:
     name: Paths filter
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, Gondolin, ubuntu-20.04-lts]
     outputs:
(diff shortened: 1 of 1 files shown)


pinggao187 (approved):

glorysdj (approved):
LGTM

PR #6493, opened 2022-11-08
Chronos: automatically detect and repair abnormal data

plusbang:
## Description
Currently, we automatically repair data following timestamp check, irregular interval check and missing value check. But there may exist abnormal data when users collect data from Prometheus or other data source.
We may also detect and repair abnormal data during TSDataset initialization.

### 2. User API changes

```python
def repair_abnormal_data(self, mode="relative", threshold=3.0)
```
Parameters:
- mode: detect abnormal data mode, select from "absolute" or "relative".
              "absolute": detect abnormal data by comparing with max and min value.
              "relative": detect abnormal data by comparing with mean value plus/minus several times standard deviation.
- threshold: indicates the range of comparison. It is a 2-dim tuple of float (min_value, max_value) when mode is set to 
                   "absolute" while it is a float number when mode is set to "relative".

### 3. Summary of the change 

3.1 add `_abnormal_value_check` when init TSdataset
3.2 add new method `repair_abnormal_data`
3.3 ut

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/chronos/src/bigdl/chronos/data/utils/quality_inspection.py
+++ b/python/chronos/src/bigdl/chronos/data/utils/quality_inspection.py
@@ -221,11 +221,9 @@ def _abnormal_value_check(df, dt_col, threshold=10):
     This check is used to determine whether there are abnormal values in the data.
     '''
     for column in df.columns:
-        if column == dt_col:
+        if column == dt_col or pd.api.types.is_string_dtype(df[column]):
             continue
         df_col = df[column]
-        if pd.api.types.is_string_dtype(df_col):
-            break # skip columns containing str
         std_val = df_col.std()
         mean_val = df_col.mean()
         df_col = df_col.apply(lambda x: x - mean_val)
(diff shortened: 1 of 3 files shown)


python/chronos/src/bigdl/chronos/data/utils/quality_inspection.py
```diff
@@ -211,3 +214,92 @@ def _missing_value_repair(df):
...
+                flag = False
+                break
+        for val in df_col:
+            if flag is False:
+                break  # skip columns containing str
+            if df_col.std() != 0 and abs((val - df_col.mean()) / df_col.std()) > threshold:
+                logging.warning(f"Some values of column {column} exceeds the mean plus/minus "
+                                f"{threshold} times standard deviation, please call "
+                                f".repair_abnormal_data() to remove abnormal values.")
+                return False
```
TheaperDeng, line 239: we have a really high complexity in this function.
For line 228 - 231, we have O(N)
For line 232 - 239, we have O(N^2) since for every iteration, we calculate std and mean.

I recommend to:
1. no need to have the first loop (started from 228), you may only use `df_col.dtype`
2. calculate std and mean before the second loop
3. try some vectorized method here rather than a for loop

TheaperDeng: same comment to other implements

TheaperDeng (approved):
LGTM

PR #6494, opened 2022-11-08
Chronos: split action workflow

liangs6212:
## Reason

Since `uts` was taking too long to run, we split the operations workflow, including `forecaster` and `other`(both <1 hour).

### 1. How to test?
- [ ] Unit test


plusbang (commented):
LGTM, but I wonder whether we need `RUN_PART1` and `RUN_PART2` in run-pytests-other.sh.

liangs6212:
> I'm not sure, maybe we can leave the changes alone for now.

liangs6212:
> LGTM, but I wonder whether we need `RUN_PART1` and `RUN_PART2` in run-pytests-other.sh.

 I'm not sure, maybe we can leave the changes alone for now.

PR #6502, opened 2022-11-09
Update readme to add Kubernetes settings

gc-fu:
## Description
Add Kubernetes required settings to trusted-deep-learning README

### 1. Why the change?

Provide some tuning options for users who want to use Kubernetes.


glorysdj (approved):
LGTM

PR #6514, opened 2022-11-09
Nano: add more installation options for pytorch (include `pytorch_nightly` / `pytorch_110` / `pytorch_111` / `pytorch_112` / `pytorch_113`)

TheaperDeng:
## Description

This PR implements more options for nano's pytorch installation, this include
```bash
# torch 1.10 and its dependencies
pip install --pre --upgrade bigdl-nano[pytorch_110]

# torch 1.11 and its dependencies
pip install --pre --upgrade bigdl-nano[pytorch_111]

# torch 1.12 and its dependencies
pip install --pre --upgrade bigdl-nano[pytorch_112]  # default "pytorch" option is assigned to 1.12

# torch 1.13 and its dependencies
pip install --pre --upgrade bigdl-nano[pytorch_113]

# torch nightly and its dependencies
pip install --pre --upgrade bigdl-nano[pytorch_nightly] --extra-index-url https://download.pytorch.org/whl/nightly/
```

Due to many reasons, we decided to not support 1.9 in this friendly installation options
- ipex 1.9 does not have the same name as >=1.10
- some of our required inference optimization depdencies does not support 1.9 anymore
- some of our features does not work for 1.9 very well

### DIFF
--- a/.github/workflows/nano_unit_tests_pytorch.yml
+++ b/.github/workflows/nano_unit_tests_pytorch.yml
@@ -54,7 +54,7 @@ jobs:
           $CONDA/bin/conda info
           if [ ! -z "${{matrix.pytorch-version}}" ]; then
             requirements=(${{matrix.pytorch-version}})
-            bash python/nano/dev/build_and_install.sh linux default false ${requirements[0]}
+            bash python/nano/dev/build_and_install.sh linux default false ${requirements[0]} -f https://software.intel.com/ipex-whl-stable
           fi
           source bigdl-nano-init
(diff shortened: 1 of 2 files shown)


rnwang04:
Shall we add some workflow test for this new installation option?
Just worry whether `inference` related dependencies could apply to every torch version.
Or we can do some offline wheel test.

TheaperDeng:
> Shall we add some workflow test for this new installation option? Just worry whether `inference` related dependencies could apply to every torch version. Or we can do some offline wheel test.

yeah, I will definitely add more tests on action for this

PR #6516, opened 2022-11-09
Nano: Add save and load support for tf onnx trace model

MeouSker77:
## Description

Add save and load support for tf onnx trace model


rnwang04 (approved):
LGTM

TheaperDeng (approved):
LGTM

PR #6523, opened 2022-11-10
Nano: refactor PyTorch InferenceOptimizer

rnwang04:
## Description

Refactor the optimization logic to a class TorchAccelerationOption so that it is clear which options does.

### 1. Why the change?
https://github.com/intel-analytics/BigDL/pull/6458#discussion_r1016098714

### 2. User API changes

No change.

### 3. Summary of the change 

* Refactor the optimization logic to a class TorchAccelerationOption

### 4. How to test?
- [x] Unit test
- [x] Application test

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
+++ b/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
@@ -15,19 +15,17 @@
 #
 
 import torch
-import pytorch_lightning as pl
 from torch import nn
 import time
-import numpy as np
 from copy import deepcopy
-from typing import Dict, Callable, Tuple, Optional, List, Set, Union, Sequence
+from typing import Dict, Callable, Tuple, Optional, List, Union, Sequence
 from torch.utils.data import DataLoader
 from torchmetrics.metric import Metric
 from bigdl.nano.utils.inference.common.checker import available_acceleration_combination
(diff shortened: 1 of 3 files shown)


python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -52,28 +49,70 @@
 os.environ['LOGLEVEL'] = 'ERROR'  # remove parital output of inc
 
 
+class TorchAccelerationOption(AccelerationOption):
```
yangw1234, line 52: I think this is better than the current code.

But what I meant is that each `AccelerationOption` has its own `optimize` method.

Say for `bf16_ipex`, we can has a `BF16IPEXAcceleration` class:

```python
class BF16IPEXAcceleration:

    def optimize():
        InferenceOptimizer.quantize(...)

class FP32JITAccelerator:

    def optimize():
        InferenceOptimizer.trace(...)
```

yangw1234: I think this PR can be a first step to simplify.

rnwang04: Thanks for suggestions! Will merge this PR first, and try above method in next PR.

PR #6530, opened 2022-11-10
[Nano] Add missing error messages for `InferenceOptimizer.optimize`

Oscilloscope98:
Add missing error messages for `InferenceOptimizer.optimize`for the case "failed to forward"


rnwang04 (approved):

rnwang04:
LGTM

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -301,6 +301,7 @@ def func_test(model, input_sample):
                         torch.set_num_threads(default_threads)
                         continue
                 except Exception as e:
+                    print(e)
```
yangw1234, line 304: It is better to also print stack trace.

PR #6534, opened 2022-11-10
Chronos: modify the default thread number of `build_openvino`, `build_onnx` and `build_jit`

plusbang:
## Description

The thread number defaults to be None when directly call `predict_with_onnx`, `predict_with_openvino` and `predict_with_jit`. According to test results, this cause a little bad latency performance. Although we have provided `build_openvino`, `build_onnx` and `build_jit` to control, we still need to provide users with better default settings.

### 2. User API changes

None

### 3. Summary of the change 

modify the default thread number from None to 1

### 4. How to test?
- [x] Unit test


TheaperDeng (commented):
Could we first make sure if the `OMP_NUM_THREADS` mis-setting does affect the performance? If that's the case, we could add more information here in the comment.

TheaperDeng (approved):
LGTM

PR #6535, opened 2022-11-10
[PPML] remove proxy set from ehsm

SEULiyao:
## Description

removed proxy url in `ppml/services/ehsm/kubernetes/install-bigdl-ehsm-kms.sh` & `ppml/services/ehsm/kubernetes/README.md`

### 1. Why the change?

proxy is not needed in ehsm.


Uxito-Ada (approved):
lgtm

PR #6538, opened 2022-11-10
Add workflow to test Chronos[pytroch] installation on python 3.8

plusbang:
### 1. Why the change?

Currently Chronos only supports python 3.7.x. We need to support higher version of python and this pr is used to test different installation options on python3.8.

### 2. User API changes

nothing

### 3. Summary of the change 

3.1 add a new workflow to test installation options on python 3.8 platform
3.2 split dependencies of all uts and add marks according to installation options

### 4. How to test?
- [x] Unit test


plusbang:
Although only [pytorch] option is tested in this pr, the pr is big now. I think we can merge it first and test other options based on it. Could you have a look? @TheaperDeng @liangs6212

liangs6212:
> Although only [pytorch] option is tested in this pr, the pr is big now. I think we can merge it first and test other options based on it. Could you have a look? @TheaperDeng @liangs6212

Please make sure all classes contain `op_all`, such as forecaster.
btw, maybe we can rename `onnxrt16` to `inference`? Just a proposal.

TheaperDeng:
> > Although only [pytorch] option is tested in this pr, the pr is big now. I think we can merge it first and test other options based on it. Could you have a look? @TheaperDeng @liangs6212
> 
> Please make sure all classes contain `op_all`, such as forecaster. btw, maybe we can rename `onnxrt16` to `inference`? Just a proposal.

We could leave `op_all` unused or removed for now to avoid too much redundant code.

.github/workflows/chronos-install-option-py38-test.yml
```diff
@@ -0,0 +1,54 @@
...
+# Cancel previous runs in the PR when you push new commits
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: true
+
+# Controls when the action will run. 
+on:
+  # Triggers the workflow on pull request events but only for the main branch
+  pull_request:
+    branches: [ main ]
```
TheaperDeng, line 12: could we add a scheduled trigger for this workflow?

python/chronos/dev/test/run-installation-options.sh
```diff
@@ -25,18 +25,23 @@ if [ -z "${OMP_NUM_THREADS}" ]; then
...
-MOPTIONS=$1
-
-echo "Running chronos tests TF1 and Deprecated API"
-echo $MOPTIONS
-python -m pytest -v -m "${MOPTIONS}" test/bigdl/chronos/forecaster
+# ray stop -f
+
+if [ $1 = true ]; then
+    echo "Running chronos[pytorch] tests"
+    python -m pytest -v -m "not tf2 and not onnxrt16 and not automl and not distributed and not diff_set_all" test/bigdl/chronos/data \
```
TheaperDeng, line 32: we have decided to have 6 ops as we discussed offline
- op_torch
- op_tf2
- op_automl
- op_distributed
- op_inference
- op_diff_set_all

plusbang: I agree we use the above 6 ops, but I think we can do this in later pr (test [tensorflow],[distributed],[inference] options and add these ops based on this pr).

PR #6539, opened 2022-11-10
Nano: Add CUDA unpatch

MeouSker77:
## Description

- Add `unpatch_cuda()`, which will be called automatically in `unpatch_torch()`.
- Protect the patch and unpatch from being called more than once.
- Eliminate the usage of `BIGDL_NANO_PATCH_TORCH` and `BIGDL_NANO_PATCH_CUDA` environment variable.

### 1. Why the change?

The `unpatch_torch()` function will not unpatch CUDA's patch for now, so we fix it in this PR.

### 2. User API changes

N/A

### 3. Summary of the change 

- Replace `setattr` with `replace_attr` helper function, which will not only set new attribute, but also store the original attributes, then use them to do unpatch.
- Use global variables instead of environment variables to indicate the status of patch

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

### 5. New dependencies

N/A

### DIFF
--- a/python/nano/src/bigdl/nano/deps/ray/ray_distributed.py
+++ b/python/nano/src/bigdl/nano/deps/ray/ray_distributed.py
@@ -61,6 +61,7 @@ from bigdl.nano.utils.log4Error import invalidInputError
 from bigdl.nano.pytorch.strategies.ipex.ipex_api import ipex_device, ipex_optimize, \
     create_IPEXAccelerator, to_cpu
 from bigdl.nano.pytorch.utils import TORCH_VERSION_LESS_1_10
+from bigdl.nano.pytorch.dispatcher import get_patch_status
 
 _STEP_OUTPUT_TYPE = Union[torch.Tensor, Dict[str, Any]]
 
@@ -126,10 +127,12 @@ class _RayLauncher(_SpawnLauncher):
             torch_backend = "nccl" if strategy.use_gpu else "gloo"
         strategy._process_group_backend = torch_backend
 
+        patch_status = get_patch_status()
+
         futures = [
             strategy.workers[i].execute.remote(
                 self._wrapping_function,
-                (i, trainer, strategy, function, args, kwargs)
+                (i, trainer, strategy, function, args, kwargs, patch_status)
             )
             for i in range(strategy.num_workers)
         ]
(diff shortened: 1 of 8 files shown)


python/nano/src/bigdl/nano/pytorch/gpu_cpu/gpu_cpu.py
```diff
@@ -41,6 +40,14 @@
                       'empty_quantized', 'empty_strided',
                       'frombuffer', 'from_file']
 
+attrs = []
+
+
+def replace_attr(obj, name: str, value):
+    torch_attr = getattr(obj, name)
+    setattr(obj, name, value)
+    attrs.append((obj, name, torch_attr))
```
TheaperDeng, line 49: what if `patch_cuda` is called twice? `torch_attr` will be the patched cpu version and the unpatch function will be broken.

python/nano/src/bigdl/nano/pytorch/dispatcher.py
```diff
@@ -91,5 +91,7 @@ def unpatch_torch():
     for mapping_iter in mapping_torch:
         setattr(mapping_iter[0], mapping_iter[1], mapping_iter[3])
 
+    unpatch_cuda()
+
```
TheaperDeng, line 95: remove the unpatch todo comment?

MeouSker77: done

python/nano/src/bigdl/nano/pytorch/gpu_cpu/gpu_cpu.py
MeouSker77: fixed it

jason-dai:
Please update the PR description with more details.

MeouSker77:
> Please update the PR description with more details.

done

jason-dai:
> * Use `BIGDL_NANO_PATCH_TORCH` and `BIGDL_NANO_PATCH_CUDA` environment variables to avoid calling patch or unpatch more than once, and telling subprocess whether to call `patch_torch()`.

Is there a way to eliminate the usage of environment variables, which are difficult to manage when you try to deploy the code on a cluster?

MeouSker77:
> Is there a way to eliminate the usage of environment variables, which are difficult to manage when you try to deploy the code in a cluster?

yes, I'll refactor it

python/nano/src/bigdl/nano/deps/ray/ray_distributed.py
```diff
@@ -156,6 +156,8 @@ def _wrapping_function(args_pack: tuple) -> Any:   # type: ignore[override]
         # patch Pytorch and CUDA in subprocess
         if os.environ.get('BIGDL_NANO_PATCH_TORCH') == '1':
             patch_cuda = True if os.environ.get('BIGDL_NANO_PATCH_CUDA') == '1' else False
+            os.environ['BIGDL_NANO_PATCH_TORCH'] = '0'
```
yangw1234, line 159: Are there any places where we document the meaning of `BIGDL_NANO_PATCH_TORCH` and `BIGDL_NANO_PATCH_CUDA`? 
If there are not any, maybe we should add the document.

MeouSker77: They will only be used internally now, they are just used to indicate if patch has been called and whether subprocess should call patch. I'll eliminate their usage.

PR #6547, opened 2022-11-10
[chronos] fix bug of decomposition

devWangBin:
when the decomposition_kernel_size is an even number, the tensor size mismatch error will be reported, fixing the error by changing the padding policy.


liangs6212:
LGTM

Issue #6500, opened 2022-11-09
Orca: init_orca_context function for the cores setting parameters didn't take effect .


bihuanxu:
for AutoTSEstimator, if we set the thread limit on pythorch level for 4 , 
 os.environ['OMP_NUM_THREADS'] = "4"  # cores
 torch.set_num_threads(4)  # please revise this according to the hardware you have

and when call init_orca_context ,  set the cores=4.
init_orca_context(address="localhost:6379", runtime="ray", cores=4)
ray_ctx = OrcaContext.get_ray_context()
ray_ctx.is_local = True

then when set cpus_per_trial=1 for AutoTSEstimator, it show that the cpu cores used for tuning and training is far more than 4 cpu cores. it seem it is not really take effect at all for cores limitation.
auto_estimator = AutoTSEstimator(model='tcn',
                                 search_space="normal",
                                  past_seq_len=look_back,
                                  future_seq_len=horizon,
                                 metric="mse",
                                  cpus_per_trial=1)
if we use htop to see the cpu status, it show that far more than 4 cpu cores are used.

liangs6212:
hi, how did you start the ray service? 

> for AutoTSEstimator, if we set the thread limit on pythorch level for 4 , os.environ['OMP_NUM_THREADS'] = "4" # cores torch.set_num_threads(4) # please revise this according to the hardware you have
> 
> and when call init_orca_context , set the cores=4. init_orca_context(address="localhost:6379", runtime="ray", cores=4) ray_ctx = OrcaContext.get_ray_context() ray_ctx.is_local = True
> 
> then when set cpus_per_trial=1 for AutoTSEstimator, it show that the cpu cores used for tuning and training is far more than 4 cpu cores. it seem it is not really take effect at all for cores limitation. auto_estimator = AutoTSEstimator(model='tcn', search_space="normal", past_seq_len=look_back, future_seq_len=horizon, metric="mse", cpus_per_trial=1) if we use htop to see the cpu status, it show that far more than 4 cpu cores are used.

bihuanxu:
use the following command to launch the ray server.
ray start --head

liangs6212:
> use the following command to launch the ray server. ray start --head

There are many parameters available for "ray start", have you tried "num-cpus"? Like this ray start --head --num-cpus 4, more info, please run ray start --help.

TheaperDeng:
> > use the following command to launch the ray server. ray start --head
> 
[...]

PR #6568, opened 2022-11-11
[ORCA] RayContext supports setting cpu nums through init_orca_context

sgwhat:
## Description
OrcaRayContext supports to create a new Ray instance instead of only connecting to an existing Ray cluster, and also supports users to specify `cores` to specify `num_cpus` resources.

### 1. Why the change?
For https://github.com/intel-analytics/BigDL/issues/6500

### 2. User API changes

In the previous version, we required users to provide `address` for `OrcaRayContext` to connect to an existing Ray cluster, and due to Ray's limitation, users could not specify `num_cpus` to control cluster resources.
```python
from bigdl.orca import init_orca_context

init_orca_context(runtime="ray",  address="localhost:6379")
```

Currently, we provide an option for users, which could support users to create a new local Ray instance and set `num_cpus` as below:
```python
from bigdl.orca import init_orca_context

# set `num_cpus` through `cores`
init_orca_context(runtime="ray",  cores=2)
```

### 3. Summary of the change 
1. Support creating a new Ray instance.
2. Support set `num_cpus`.

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/orca/src/bigdl/orca/common.py
+++ b/python/orca/src/bigdl/orca/common.py
@@ -231,8 +231,6 @@ def init_orca_context(cluster_mode=None, runtime="spark", cores=None, memory="2g
         from bigdl.orca.ray import OrcaRayContext
         ray_ctx = OrcaRayContext(runtime="ray", cores=cores, num_nodes=num_nodes,
                                  **kwargs)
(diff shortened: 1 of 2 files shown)


TheaperDeng (approved):
LGTM

PR #6569, opened 2022-11-11
Chronos: optimize forecaster to reduce memory consumption

hjzin:
### 1. Why the change?

When forecasters are accelerated by various accelerators (onnxruntime, openvino, ...), all the accelerated models will be saved inside forecaster, which causes more memory consumption.
### 2. User API changes

N/A
### 3. Summary of the change 

- The forecaster only saves the latest accelerated model. The accelerated model will be saved at `self.accelerated_model`, and the model type will be saved at `self.accelerated_method`. 
- The quantized model will be saved in a dictionary because multiple frameworks can be specified when calling `.quantize()`. 
### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/chronos/src/bigdl/chronos/forecaster/base_forecaster.py
+++ b/python/chronos/src/bigdl/chronos/forecaster/base_forecaster.py
@@ -87,8 +87,6 @@ class BasePytorchForecaster(Forecaster):
             self.accelerated_model = None  # accelerated model obtained from various accelerators
             self.accelerate_method = None  # str indicates current accelerate method
             self.quantize_models = {}      # quantization models, format: {model_type : model}
-            self.optim_model = None        # accelarated model obtained from .optimize() 
-            self.metadata = None           # str indicates model type of optim_model
 
     def _build_automodel(self, data, validation_data=None, batch_size=32, epochs=1):
         """Build a Generic Model using config parameters."""
@@ -450,7 +448,7 @@ class BasePytorchForecaster(Forecaster):
         '''
         This method will traverse existing optimization methods(onnxruntime, openvino, jit, ...)
         and save the model with minimum latency under the given data and search
-        restrictions(accelerator, precision, accuracy_criterion) in `forecaster.optim_model`.
+        restrictions(accelerator, precision, accuracy_criterion) in `forecaster.accelerated_model`.
         This method is required to call before `predict` and `evaluate`.
         Now this function is only for non-distributed model.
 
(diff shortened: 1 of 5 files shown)


python/chronos/src/bigdl/chronos/forecaster/base_forecaster.py
```diff
@@ -83,15 +83,12 @@ def model_creator_orca(config):
...
-            self.jit_int8 = None  # placeholader jit session for int8 precision
-            self.pytorch_int8 = None  # pytorch model for int8 precision
-            self.optim_model = None  # accelarated model obtained from .optimize()
-            self.metadata = None  # str indicates model type of optim_model
+
+            self.accelerated_model = None  # accelerated model obtained from various accelerators
+            self.accelerate_method = None  # str indicates current accelerate method
+            self.quantize_models = {}      # quantization models, format: {model_type : model}
+            self.optim_model = None        # accelarated model obtained from .optimize() 
+            self.metadata = None           # str indicates model type of optim_model
```
TheaperDeng, line 91: We could just merge `self.optim_model` to `self.accelerated_model ` and `self.metadata` to `self.accelerate_method `

python/chronos/src/bigdl/chronos/forecaster/base_forecaster.py
```diff
@@ -83,15 +83,12 @@ def model_creator_orca(config):
...
-            self.onnxruntime_int8 = None  # onnxruntime session for int8 precision
-            self.openvino_int8 = None  # placeholader openvino session for int8 precision
-            self.jit_int8 = None  # placeholader jit session for int8 precision
-            self.pytorch_int8 = None  # pytorch model for int8 precision
-            self.optim_model = None  # accelarated model obtained from .optimize()
-            self.metadata = None  # str indicates model type of optim_model
+
+            self.accelerated_model = None  # accelerated model obtained from various accelerators
+            self.accelerate_method = None  # str indicates current accelerate method
+            self.quantize_models = {}      # quantization models, format: {model_type : model}
```
TheaperDeng, line 89: leave `self.quantize_models` temporarily here, we shall change `quantize`'s API later to avoid quantize 2 models at the same time.

python/chronos/src/bigdl/chronos/forecaster/base_forecaster.py
TheaperDeng: This means that `forecaster.predict(...)` will automatically use the accelerated model.

hjzin: > This means that `forecaster.predict(...)` will automatically use the accelerated model.

Does this mean `forecaster.predict()` will always use the latest accelerated model no matter what type it is ?

PR #6589, opened 2022-11-14
fix pyspark CVE issue

Le-Zheng:
## Description

Hide pyspark version 

## Reason
image
Need to upgrade pyspark to 3.2.2. solve the issue with inputting the version first.


python/default_pyspark_version.txt
```diff
@@ -0,0 +1 @@
+2.4.6
```
leonardozcm, line 1: It seems that putting dependencies and versions in a **requirments.txt** under each submodule dir is more 'python style'? @hkvision 

Others lgtm.

hkvision: Maybe can we put it here?
https://github.com/intel-analytics/BigDL/tree/main/python/requirements/orca
Add a requirement_pyspark.txt?

leonardozcm: > Maybe can we put it here? https://github.com/intel-analytics/BigDL/tree/main/python/requirements/orca Add a requirement_pyspark.txt?

Sure, but `dllib` you mean? pyspark is binded with dllib.

hkvision: But since dllib doesn't have a requirements folder, if we don't create it, just put it under orca?

python/requirements/orca/requirement_pyspark.txt
```diff
@@ -0,0 +1 @@
+2.4.6
```
hkvision, line 1: Can we change to pyspark==2.4.6 and grep the version out?

Le-Zheng: Any "pyspark==2.4.6" will be scanned as security issue. We need to avoid the key word "pyspark==".

Le-Zheng: @hkvision @leonardozcm Another idea. How about write PYSPARK_VERSION='2.4.6' in dllib setup.py directly instead of putting it in requirement.txt?

hkvision: > @hkvision @leonardozcm Another idea. How about write PYSPARK_VERSION='2.4.6' in dllib setup.py directly instead of putting it in requirement.txt?

I am ok with this one

leonardozcm: > I am ok with this one

I think it is a good idea

Le-Zheng:
jenkins test pass http://10.112.231.51:18888/job/BigDL2.0-Deploy-PYTHON-PyPI-Specific-Spark-Version/26/

Le-Zheng:
Merge first due to an urgent deadline.

PR #6603, opened 2022-11-15
House price analysis

songhappy:
## Description

Port house_price_analysis.py into a notebook for visualization purposes.

###. How to test?
- [ ] N/A


jason-dai:
It seems that you mix pandas and xshards objects/operations in the code, which can be confusing to the user; I wonder if we can make it easier for the user (e.g., applying visualization on xshards objects, or making the distinction between the two clearer)?

PR #6610, opened 2022-11-15
[Doc] Sidebar fix

AndyLuo1029:
### 1. Why the change?
 Fixed an issue where the sidebar‘s content was not fully displayed.

Previous bug: The `bigdl.nano.pytorch.InferenceOptimizer` can't fully display.
image

Current effect: Now it will automatically wrap.
image

### 2. Preview link: [here](https://andytestdoc.readthedocs.io/en/sidebar_fix/doc/PythonAPI/Nano/pytorch.html)

### DIFF
--- a/docs/readthedocs/source/_static/css/custom.css
+++ b/docs/readthedocs/source/_static/css/custom.css
@@ -102,3 +102,8 @@ footer.bd-footer{
 .rendered_html table {
     table-layout: auto !important;
 }
+
+/* for right sidebar word-break */
+/* #bd-toc-nav ul li {
+    word-break: break-word;
+} */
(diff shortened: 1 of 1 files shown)


docs/readthedocs/source/_static/css/custom.css
```diff
@@ -101,4 +101,9 @@ footer.bd-footer{
 /* for tables in notebooks generated by pandas dataframe */
 .rendered_html table {
     table-layout: auto !important;
+}
+
+/* for right sidebar word-break */
+#bd-toc-nav ul li {
+    word-break: break-word;
```
Oscilloscope98, line 108: We are using `overflow-wrap: break-word;` for the middle contents. Is there a specific reason we use `word-break: break-word;` here?

If not, I feel like it is better to use `overflow-wrap: break-word;` here :D

Oscilloscope98: If `overflow-wrap: break-word;` does not take effect, then it LGTM.

Oscilloscope98 (approved):

PR #6613, opened 2022-11-15
Apply mod to some files in gramine to bigdata

ShanSimu:
## Description

Apply the modifications to  bigdl-ppml-submit.sh, spark-driver-template.yaml and spark-executor-template.yaml in gramine to bigdata


Uxito-Ada (approved):
lgtm

PR #6614, opened 2022-11-15
[PPML] update kms-utils

SEULiyao:
## Description

deleted `run-docker-container.sh`
add 'jdk_url' into `build-docker-image.sh`
updated readme

### 1. Why the change?

the script is different from description
jdk_url is needed in `build-docker-image.sh`


Uxito-Ada (approved):
lgtm

Issue #6611, opened 2022-11-15
Nano : reason for DeepViT BF16 fail to forward


rnwang04:
After debugging, I found `fail to forward` is caused by `import intel_extension_for_pytorch as ipex`.
For example, if we just test 'original' / 'bf16' and other methods, everything is ok：
![image](https://user-images.githubusercontent.com/105281011/201870407-af10502d-198b-4fc5-bccc-6020f8796457.png)
But if we test ipex, bf16 will failed:
![image](https://user-images.githubusercontent.com/105281011/201870473-ef88a34d-bb53-4582-8dc1-6e4495159057.png)
and error is raised from torch.layer_norm (RuntimeError: gamma and beta's data type should be float).
This error is caused by this line:
https://github.com/intel-analytics/BigDL/blob/a5032006fed1ef3c6322855a0010f81067aff86b/python/nano/src/bigdl/nano/pytorch/amp/bfloat16.py#L121
model.bfloat16() will make all weight to torch.bfloat16(), but ipex and low leval ONEDNN don't support torch.bfloat16 for torch.layer_norm which is really common in CV.

I will try mixed precision in Nano to support such cases.

PR #6618, opened 2022-11-15
Nano: update quantize with BF16

rnwang04:
## Description

According to some experience from OOB, we found BF16 in Nano exists several problems:
- strict check for ISA, however bf16 model sometimes can run on hardware without bf16 related instructions
- now we use complete precision bf16 model(model.bfloat16()), but this will conflict with ipex, and will cause error in `optimize` when model contains layer norm layer ; by the way, mixed precision model seem to be a more general approach
- lack of `bf16 + channels_last` combination
- save & load function fails for bf16 model
- autocast inside forward will bring about additional  30% overhead

This PR will try to solve first four problems, and **leave the last problem in a separate PR**.

### 1. Why the change?

https://github.com/intel-analytics/BigDL/issues/6611

### 2. User API changes
Add new parameter `channels_last` for BF16, and support save & load for BF16 model.
```
bf16_model = InferenceOptimizer.quantize(model,
                                         precision="bf16",
                                         channels_last=True)
output = bf16_model(input)
assert output.dtype == torch.bfloat16  # just for test
InferenceOptimizer.save(bf16_model, "bf16_model")

bf16_model = InferenceOptimizer.load("bf16_model", model)
output = bf16_model(input)
assert output.dtype == torch.bfloat16  # just for test
```
### 3. Summary of the change 

- Update ISA check of bf16 and ipex bf16. Now only check at initialization, and don't throw exception for bf16 model even if there don't exists bf16 related instruction set.
- change from complete precision model to mixed precision
- Add pamater `channels_last` for BF16
- support save & load for BF16 model

### 4. How to test?
- [x] Unit test
- [x] Application test

### DIFF
--- a/python/nano/src/bigdl/nano/deps/ipex/ipex_inference_bf16_model.py
+++ b/python/nano/src/bigdl/nano/deps/ipex/ipex_inference_bf16_model.py
@@ -46,15 +46,11 @@ class PytorchIPEXJITBF16Model(PytorchIPEXJITModel):
         :param from_load: this will only be set by _load method.
         '''
         if use_ipex:
-            # invalidInputError(
-            #     self._check_cpu_isa,
-            #     errMsg="Applying IPEX BF16 optimization needs the cpu support avx512.",
-            #     fixMsg="Please set use_ipex to False or not set precision to bf16."
-            # )
-            if self._check_cpu_isa is False:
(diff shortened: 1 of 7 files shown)


jason-dai:
LGTM

TheaperDeng (approved):
LGTM

PR #6620, opened 2022-11-15
Nano : Add more combinations in torch InferenceOptimizer

rnwang04:
## Description

Add more combinations in torch InferenceOptimizer based on experience from oob inference.

### 4. How to test?
- [ ] Unit test
- [ ] Application test


python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -191,6 +211,32 @@ def optimize(self, model: nn.Module,
...
+               Defaults to None which represents there is no restriction on accelerators.
+               If not None, then will only travese corresponding methods whose accelerator falls
+               within the specified accelerator tuple.
+        :param precision: (optional) A string tuple that specifys the precision to search.
+               The optional precision are: 'int8', 'bf16', and 'fp32'. Defaults to None which
+               represents no precision limit. If not None, then will only travese corresponding
+               methods whose precision falls within the specified precision tuple.
+        :param use_ipex: (optional) if not None, then will only try methods with/without
+               this specific ipex setting.
+        :param search_model: Here are three modes for optimization:
```
TheaperDeng, line 225: search_mode

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -129,6 +144,10 @@ def optimize(self, model: nn.Module,
                  metric: Optional[Callable] = None,
                  direction: str = "max",
                  thread_num: Optional[int] = None,
+                 accelerator: Optional[Tuple[str]] = None,
+                 precision: Optional[Tuple[str]] = None,
+                 use_ipex: Optional[bool] = None,
```
TheaperDeng, line 149: 1. should we take `channels_last` here as one of the setting that could be specified like something stated in https://github.com/analytics-zoo/nano/issues/214 ?

python/nano/test/pytorch/tests/test_inference_optimizer.py
```diff
@@ -125,6 +127,8 @@ def test_pipeline_without_metric(self):
                                thread_num=1)
 
         acc_model, option = inference_opt.get_best_model()
+        if TORCH_VERSION_LESS_1_10:
+            return
```
TheaperDeng, line 131: 1.9 and there tests has been deprecated in previouse PR, we are good to remove these

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: done

python/nano/test/pytorch/tests/test_inference_optimizer.py
rnwang04: done

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: Three questions here:
1. Does all 4-dim input data are image data ? Or Does channels_last works for all 4-dim data?
2. We don't have fp32_channels_last now, to achieve this, one way is a new class which only do one thing : wrap channels_last; another approach might be achieve this in `PytorchIPEXJITModel`.
3. Do you mean make `channels_last` as a filter like `accelerator` and `precision` ?

rnwang04: if channels_last can apply on openvino / onnxruntime

PR #6622, opened 2022-11-15
[WIP] Orca: extend bigdl-submit CLI.

lalalapotter:
## Description

This PR is to extend bigdl-submit CLI.

## User Usage

For the previous version, we only support users (which cannot install conda) to run programs as below:
```python
1. Download BigDL assembly package.
2. Run spark-submit command:
    spark-submit \
        --master yarn \
        --deploy-mode client \
        --archives /path/to/environment.tar.gz#environment \
        --py-files /path/to/python.zip \
        --jars /path/to/jars \
        train.py
```

For currently, we could do this work like:
```python
bigdl-submit
    --master yarn \
    --deploy-mode client \
    --archives /path/to/environment.tar.gz#environment \
    train.py
```

This new extending `bigdl-submit` CLI could provide users an automatically way to run their programs with none-conda directly, what they need to do is just prepare the conda archive.

### DIFF
--- a/scripts/bigdl-submit
+++ b/scripts/bigdl-submit
@@ -3,20 +3,19 @@
 set -e
 
 # usage
-usage() { echo "Usage: $0 [-a|--archives <string#string> -h|--help]" 1>&2; exit 1; }
+usage() { echo "Usage: $0 [-a|--archives <string#string,...> -d|--deploy-mode <string> -c|--conf <string=string> -h|--help]" 1>&2; exit 1; }
 
 cmd=$@
 # get options
 set +e
-ARGS=`getopt -q -o a:h --long archives:,help -- "$@"`
+ARGS=`getopt -o a:d:c:h: --long archives:,deploy-mode:,conf:,help -- "$@"`
 set -e
 eval set -- "$ARGS"
 
 # extract options and their arguments into variables.
-archive-array=()
-dir-array=()
-dir=$PYSPARK_PYTHON
-deploy-mode=""
+declare -A archives_map
+declare -A conf_map
+dir="environment"
 while true ; do
     case "$1" in
         -a|--archives)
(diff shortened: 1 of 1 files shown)


lalalapotter:
Need to add tests.

hkvision:
Please add the current command how users would use here

scripts/bigdl-submit
```diff
@@ -2,21 +2,64 @@
 
 set -e
 
-#detect paths
+# usage
+usage() { echo "Usage: $0 [-a|--archives <string#string> -h|--help]" 1>&2; exit 1; }
```
hkvision, line 6: will this --archives conflict with spark-submit's archive?

lalalapotter: No, I have store the original spark arguments in `cmd`.

scripts/bigdl-submit
```diff
@@ -2,21 +2,64 @@
...
+
+cmd=$@
+# get options
+set +e
+ARGS=`getopt -q -o a:h --long archives:,help -- "$@"`
+set -e
+eval set -- "$ARGS"
+
+# extract options and their arguments into variables.
+archives="./environment.tar.gz"
```
sgwhat, line 16: Can users name the conda archive randomly?

scripts/bigdl-submit
```diff
@@ -2,21 +2,64 @@
...
 set -e
 
-#detect paths
+# usage
+usage() { echo "Usage: $0 [-a|--archives <string#string> -h|--help]" 1>&2; exit 1; }
+
+cmd=$@
+# get options
+set +e
+ARGS=`getopt -q -o a:h --long archives:,help -- "$@"`
```
sgwhat, line 11: Users could ignore `--` after a option now?

lalalapotter: Yes, I have added break after archives.

scripts/bigdl-submit
```diff
@@ -2,21 +2,145 @@
...
+            then
+                dir=${conf_map["spark.pyspark.python"]}
+            fi
+        fi
+    else
+        if [ $deploy_mode == "cluster" ]
+        then
+            if [ conf_map["spark.yarn.appMasterEnv.PYSPARK_PYTHON"]]
+                dir=${conf_map["spark.yarn.appMasterEnv.PYSPARK_PYTHON"]}
+            fi
```
sgwhat, line 89: Guess we may need `spark.executorEnv.PYSPARK_PYTHON` here?

lalalapotter:
According to customer feedback, their server does not support the use of source to activate the environment, maybe we need to consider other implementation methods.

PR #6625, opened 2022-11-15
Nano: update the metadata saved and loaded for openvino

TheaperDeng:
### 1. Why the change?

Nano serve users as a runtime API rather than a development tool, so we need to preserve the runtime config to disk and make sure they are properly loaded back. Currently we don't load back the config and some of the configurations such as num_thread and precision hint lost when the model is loaded back

With this PR, a typical metadata yml could be
```yaml
ModelType: PytorchOpenVINOModel
config:
  CPU_THREADS_NUM: '4'
  INFERENCE_PRECISION_HINT: f32
weight_path: ov_saved_model.bin
xml_path: ov_saved_model.xml
```

### 2. User API changes
Nothing

### 3. Summary of the change 
add a "config" key when saving and loading.

### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/deps/openvino/core/model.py
+++ b/python/nano/src/bigdl/nano/deps/openvino/core/model.py
@@ -76,6 +76,7 @@ class OpenVINOModel:
         self._compiled_model = self._ie.compile_model(model=self.ie_network,
                                                       device_name=self._device,
                                                       config=config)
+        self.final_config = config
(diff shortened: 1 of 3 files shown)


TheaperDeng:
btw, currently bf16 op might still exist if you test the model that is loaded back. We could fix this issue by this PR. @rnwang04

MeouSker77 (approved):
LGTM

rnwang04 (approved):
LGTM

PR #6633, opened 2022-11-16
Add workflow to test Chronos [tensorflow] [inference] [automl] installation on python 3.8 

plusbang:
### 1. Why the change?

Currently Chronos only supports python 3.7.x. We need to support higher version of python and this pr is used to test [tensorflow] [inference] [automl] installation options on python3.8.

### 2. User API changes

nothing

### 3. Summary of the change 

3.1 add several jobs to test other installation options on python3.8
3.2 split uts according to op_torch / op_tf2 / op_automl / op_distributed / op_inference / op_diff_set_all

### 4. How to test?
- [x] Unit test

### DIFF
--- a/.github/workflows/chronos-install-option-py38-test.yml
+++ b/.github/workflows/chronos-install-option-py38-test.yml
@@ -125,7 +125,7 @@ jobs:
           BIGDL_ROOT: ${{ github.workspace }}
           ANALYTICS_ZOO_ROOT: ${{ github.workspace }}
 
-chronos-pytorch-inference-automl-test:
+  chronos-pytorch-inference-automl-test:
     runs-on: [ self-hosted, Gondolin, ubuntu-20.04-lts ]
     strategy:
       fail-fast: false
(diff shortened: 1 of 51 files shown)


.github/workflows/chronos-install-option-py38-test.yml
```diff
@@ -46,7 +46,116 @@ jobs:
...
+        run: |
+          conda remove -n chronos-py38-env -y --all
+          conda create -n chronos-py38-env -y python==3.8 setuptools==58.0.4 -c ${CONDA_CHANNEL} --override-channels
+          source activate chronos-py38-env
+          pip install pytest==5.4.1
+          apt-get update
+          apt-get install patchelf
+          pip uninstall -y bigdl-friesian bigdl-friesian-spark3 bigdl-dllib bigdl-dllib-spark3 bigdl-orca pyspark bigdl-orca-spark3 bigdl-chronos bigdl-chronos-spark3 bigdl-nano bigdl-friesian bigdl-friesian-spark3
+          pip install --pre --upgrade bigdl-chronos[pytorch,inference,automl]
+          pip install SQLAlchemy==1.4.27
```
TheaperDeng, line 157: we may open an issue to make `SQLAlchemy<=1.4.27` a dependency in our `automl` extra requirement.

TheaperDeng (approved):
LGTM

.github/workflows/chronos-install-option-py38-test.yml
plusbang: yeah, I have opened the issue https://github.com/intel-analytics/BigDL/issues/6717. We can solve it in later pr.

PR #6640, opened 2022-11-16
Nano : support tensor and tensor tuple for InferenceOptimizer.quantize

rnwang04:
### 4. How to test?
- [ ] Unit test
- [ ] Application test


python/chronos/src/bigdl/chronos/forecaster/base_forecaster.py
```diff
@@ -1794,7 +1794,7 @@ def quantize(self, calib_data=None,
                                                   precision='int8',
                                                   accelerator=accelerator,
                                                   method=method,
-                                                  calib_dataloader=calib_data,
+                                                  calib_data=calib_data,
```
TheaperDeng, line 1797: Chronos will refresh the interface later, you may just leave chronos unchanged.

TheaperDeng (commented):
others LGTM

PR #6645, opened 2022-11-16
Orca: Refactor Runner with ModelIO ABC

leonardozcm:
## Description

ModelIO takes charge of load and save stuffs.

All runners should inherit `ModelIO`` as the worker interface and implement the following APIs:
```python
    @abstractmethod
    def get_state_dict(self):
        pass

    @abstractmethod
        pass

    @abstractmethod
    def save_checkpoint(self, filepath, save_weights_only=False):
        """Save checkpoint."""
        pass
```

https://github.com/intel-analytics/BigDL/pull/6408#issue-1432988407

### DIFF
--- a/python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
@@ -33,8 +33,6 @@
 
 from filelock import FileLock
 import logging
-import io
-import itertools
 import os
 import copy
 import tempfile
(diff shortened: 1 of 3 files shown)


python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
```diff
@@ -118,7 +115,7 @@ def __init__(self,
                             format='[%(asctime)s] %(levelname)-8s %(message)s',
                             datefmt='%Y-%m-%d %H:%M:%S'
                             )
-        self.logger = logging.getLogger(__name__)
+        BaseRunner.__init__(self, logging.getLogger(__name__))
```
hkvision, line 118: super(TorchRunner).init?

hkvision (approved):

python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
leonardozcm: this is a typo, there is no need to init logger in baserunner(as logger is not used). removed now.

jason-dai:
> ## Description
> ModelIO takes charge of load and save stuffs.
> 
> All runners should inherit `ModelIO`` as the worker interface and implement the following APIs:

Only PyTorch runners, yes?

hkvision:
> > ## Description
> > ModelIO takes charge of load and save stuffs.
> > All runners should inherit `ModelIO`` as the worker interface and implement the following APIs:
> 
> Only PyTorch runners, yes?

Yes, only for PyTorch.

PR #6649, opened 2022-11-17
Support torch_nano decorator

yangw1234:
## Description

Add a `torch_nano` decorator to more easily support user code with a training loop function.

```python

@torch_nano(use_ipex=True)
def train_loop(model, optimizer, train_data_loader, val_data_loader, *args, **kwargs):
     # training loop logic

if __name__ == "__main__":

    model = ...
    loss_fn = ...
    optimizer = ...
    dataloader = ...
    
    train_loop(model, loss_fn, optimizer, dataloader)

```

todos:

- [x] Unit tests
- [x] example
- [ ] documentation

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/torch_nano.py
+++ b/python/nano/src/bigdl/nano/pytorch/torch_nano.py
@@ -313,7 +313,7 @@ class TorchNano(LightningLite):
         pass
 
 
-def torch_nano(num_processes: int = 1,
+def nano(num_processes: int = 1,
                use_ipex: bool = False,
                strategy: str = "subprocess",
                precision: Union[str, int] = 32,
@@ -321,11 +321,13 @@ def torch_nano(num_processes: int = 1,
 
     def decorator(func):
 
+        # todo check the func signature
+
         def new_func(model, optimizer, train_data_loader, val_data_loader, *inner_args, **inner_kwargs):
             class DecoratedTorchNano(TorchNano):
 
                 def train(self):
-                    _model, _optimizer, _train_loader, _val_loader = self.setup(model, optimizer, train_data_loader, val_data_loader)
+                    _model, _optimizer, (_train_loader, _val_loader) = self.setup(model, optimizer, train_data_loader, val_data_loader)
                     return func(_model, _optimizer, _train_loader, _val_loader, *inner_args, **inner_kwargs)
 
             return DecoratedTorchNano(num_processes=num_processes,
--- a/python/nano/src/bigdl/nano/pytorch/torch_nano.py
+++ b/python/nano/src/bigdl/nano/pytorch/torch_nano.py
@@ -345,13 +345,13 @@ class DecoratedTorchNano(TorchNano):
         _optimizer_pos = []
         _data_loader_pos = []
 
-        _inner_args=list(inner_args)
+        _inner_args = list(inner_args)
         _search_setup_args(_model_pos, _optimizer_pos, _data_loader_pos, _inner_args)
         _search_setup_args(_model_pos, _optimizer_pos, _data_loader_pos, inner_kwargs)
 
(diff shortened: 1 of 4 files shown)


python/nano/src/bigdl/nano/pytorch/torch_nano.py
```diff
@@ -311,3 +311,28 @@ def run(self, *args: Any, **kwargs: Any) -> Any:
...
+
+def torch_nano(num_processes: int = 1,
+               use_ipex: bool = False,
+               strategy: str = "subprocess",
+               precision: Union[str, int] = 32,
+               *args, **kwargs):
+
+    def decorator(func):
+
+        def new_func(model, optimizer, train_data_loader, val_data_loader, *inner_args, **inner_kwargs):
```
yangw1234, line 324: Any suggestions on how to handle the dataloaders here?

what are the "legal" numbers of dataloaders users are allowed to use in `self.setup`?

python/nano/src/bigdl/nano/pytorch/torch_nano.py
```diff
@@ -311,3 +311,28 @@ def run(self, *args: Any, **kwargs: Any) -> Any:
...
         # this is a abstract method, so we must implement it
         pass
+
+
+def torch_nano(num_processes: int = 1,
+               use_ipex: bool = False,
+               strategy: str = "subprocess",
+               precision: Union[str, int] = 32,
+               *args, **kwargs):
+
```
yangw1234, line 321: @MeouSker77 do you think we can detect the cases where users must call `self.backward()` instead of `loss.backward()`

python/nano/src/bigdl/nano/pytorch/torch_nano.py
MeouSker77: > Any suggestions on how to handle the dataloaders here?

maybe we can receive a `Union[DataLoader|List[DataLoader]]` here

> what are the "legal" numbers of dataloaders users are allowed to use in `self.setup`?

`self.setup` only splits dataloaders into multiple subprocesses using Sampler, so it can handle any number of dataloaders

python/nano/src/bigdl/nano/pytorch/torch_nano.py
MeouSker77: yes, we must call `self.backward()` instead of `loss.backward()` only when the strategy is `DeepSpeedStrategy` and has multiple models, we can detect this cases here

python/nano/src/bigdl/nano/pytorch/torch_nano.py
MeouSker77: I think one or two dataloaders is the most common cases.

But maybe we can detect all dataloaders automatically here, e.g.
```python
def new_func(model, optimizer, *inner_args, **inner_kwargs):
    class DecoratedTorchNano(TorchNano):
        def train(self):
            _model, _optimizer = self.setup(model, optmizer)
            
            # setup all dataloaders automatically
            inner_args=list(inner_args)
            for idx, arg in enumerate(inner_args):
                if isinstance(arg, DataLoader):
                    inner_args[idx] = self.setup_dataloaders(arg)
            for key, value in inner_kwargs.items():
                if isintance(value, DataLoader):
                    inner_kwargs[key] = self.setup_dataloaders(value)
            
            return func(_model, _optimizer, *inner_args, **inner_kwargs)
```
And we can give a warning if we don't find dataloader in the arguments.

We can even use the this methods to detect model and optimizers automatically, so that we can support any parameter list, as long as model, optimizers and dataloaders are all in parameter list. But it might be too complicated to do so.

yangw1234: I think this is a great idea. Let me try this.

PR #6661, opened 2022-11-17
fix pyspark version issue

Le-Zheng:
## Description

Fix pyspark version. Make sure the pyspark version is mapping correctly with spark version.


Le-Zheng:
http://10.112.231.51:18888/job/BigDL2.0-Deploy-PYTHON-PyPI-Specific-Spark-Version/32/consoleFull

Issue #6639, opened 2022-11-16
[nano] fixed onnx_path, will throw error if run multi onnx inference jobs in the same time.


qiuxin2012:
https://github.com/intel-analytics/BigDL/blob/69120ccb94b9401364dd1bf72b1413eaec1339d7/python/nano/src/bigdl/nano/deps/onnxruntime/pytorch/pytorch_onnxruntime_model.py#L57
We should use a writable path in random tmp dir, instead of this fixed path.
may get below error:
```
  File "test.py", line 22, in <module>
    optimized_model = trainer.trace(model, accelerator='onnxruntime', input_sample=x[0:64], thread_num=14)
  File "/home/cpx/miniconda3/envs/xin-onnx/lib/python3.7/site-packages/bigdl/nano/utils/util.py", line 31, in wrapped
    return function(*args, **kwargs)
  File "/home/cpx/miniconda3/envs/xin-onnx/lib/python3.7/site-packages/bigdl/nano/pytorch/trainer/Trainer.py", line 311, in trace
    **export_kwargs)
  File "/home/cpx/miniconda3/envs/xin-onnx/lib/python3.7/site-packages/bigdl/nano/pytorch/inference/optimizer.py", line 622, in trace
    simplification=simplification, **export_kwargs)
  File "/home/cpx/miniconda3/envs/xin-onnx/lib/python3.7/site-packages/bigdl/nano/deps/onnxruntime/onnxruntime_api.py", line 39, in PytorchONNXRuntimeModel
    simplification=simplification, **export_kwargs)
  File "/home/cpx/miniconda3/envs/xin-onnx/lib/python3.7/site-packages/bigdl/nano/deps/onnxruntime/pytorch/pytorch_onnxruntime_model.py", line 66, in __init__
    ONNXRuntimeModel.__init__(self, onnx_path, session_options=onnxruntime_session_options)
  File "/home/cpx/miniconda3/envs/xin-onnx/lib/python3.7/site-packages/bigdl/nano/deps/onnxruntime/core/onnxruntime_model.py", line 27, in __init__
    self._build_ortsess(session_options)
  File "/home/cpx/miniconda3/envs/xin-onnx/lib/python3.7/site-packages/bigdl/nano/deps/onnxruntime/core/onnxruntime_model.py", line 63, in _build_ortsess
    self.ortsess = ort.InferenceSession(onnx_path_or_bytes, sess_options=sess_options)
[...]

PR #6663, opened 2022-11-17
Nano: fix the temporary save path of onnx model

MeouSker77:
## Description

Fix the temporary save path of onnx model: #6639


TheaperDeng (approved):
LGTM

PR #6666, opened 2022-11-17
Chronos: a better implementation for cycle length detection

TheaperDeng:
### 1. Why the change?
`statsmodel` is needed for cycle length detection, we remove our requirement for this dependency.
More robust than `statsmodel`'s implementation

### 2. User API changes
Nothing

### 3. Summary of the change 
Change to a `numpy` implementation

### 4. How to test?
- [ ] Unit test

### 5. New dependencies
remove `statsmodel`

### DIFF
--- a/python/chronos/src/bigdl/chronos/data/utils/cycle_detection.py
+++ b/python/chronos/src/bigdl/chronos/data/utils/cycle_detection.py
@@ -18,7 +18,7 @@ from scipy.fftpack import fft, fftfreq
 import numpy as np
 
 
-def cycle_length_est(data, top_k=3, return_acf_score=False, adjust=False):
+def cycle_length_est(data, top_k=3, adjust=False):
     '''
     Detect the cycle of a time series.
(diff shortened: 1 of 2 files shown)


python/chronos/src/bigdl/chronos/data/utils/cycle_detection.py
```diff
@@ -18,7 +18,7 @@
 import numpy as np
 
 
-def cycle_length_est(data, top_k=3):
+def cycle_length_est(data, top_k=3, return_acf_score=False, adjust=False):
```
liangs6212, line 21: We could perhaps add some notes to describe the role of the parameters.

liangs6212:
I have no problem with this pr, when I compare the calculation results I found that the best cycle length of the new function may be different from the previous one, for example, the best cycle length of "nyc_taxi" changed from 48 to 24.
btw, Doesn't mean there's a mistake, just a reminder that we should be aware of the changes.

PR #6667, opened 2022-11-17
Chronos: add dummy encoder choice to lighten the TCN model

TheaperDeng:
### 1. Why the change?

greatly inspired by https://arxiv.org/abs/2205.13504, we choose to adopt the model in a special way - as a special case of TCN.

### 2. User API changes

add a new `dummy_encoder` parameter to `TCNForecaster`

### 3. Summary of the change 
if `dummy_encoder` is set to True, then tcn encoder is removed and TCN becomes a Linear model.

### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/chronos/src/bigdl/chronos/forecaster/tcn_forecaster.py
+++ b/python/chronos/src/bigdl/chronos/forecaster/tcn_forecaster.py
@@ -44,6 +44,7 @@ class TCNForecaster(BasePytorchForecaster):
                  future_seq_len,
                  input_feature_num,
                  output_feature_num,
+                 dummy_encoder=False,
                  num_channels=[16]*3,
                  kernel_size=3,
                  normalization=True,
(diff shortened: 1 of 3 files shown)


devWangBin:
LGTM

python/chronos/src/bigdl/chronos/forecaster/tcn_forecaster.py
```diff
@@ -69,6 +70,9 @@ def __init__(self,
         :param future_seq_len: Specify the output time steps (i.e. horizon).
         :param input_feature_num: Specify the feature dimension.
         :param output_feature_num: Specify the output dimension.
+        :param dummy_encoder: bool, no encoder is applied if True, which will
```
devWangBin, line 73: Maybe we need to add some assert code, to make sure the input_feature_num equals to output_feature_num when dummy_encoder is True.

PR #6673, opened 2022-11-18
[PPML][EHSM] Dkeyserver Fault Tolerance

Uxito-Ada:
## Description

1. Store domainkey persistently by stateful PVC.
2. Label dkeyserver to a fixed node as ckey is generated from CPU which results in that dkeyserver cannot be moved among nodes. The better patch will be released by ehsm, and we temporarily patch in this way to avoid dkeyserver failure.

### 1. Why the change?

Patch for two vulnerabilities (domainkey and ckey) of dkeyserver.

### 2. User API changes

Need to manually select a fixed node for dkeyserver in k8s cluster.

### 3. Summary of the change 

Dkeyserver fault tolerance.

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [x] Application test
- [ ] Document test
- [ ] ...


Uxito-Ada:
test passed.

Issue #6669, opened 2022-11-17
Page not found in document website


jason-dai:
https://bigdl.readthedocs.io/en/latest/doc/UseCase/spark-dataframe.html
https://bigdl.readthedocs.io/en/latest/doc/Orca/QuickStart/orca-autoxgboost-quickstart.html

PR #6676, opened 2022-11-18
Fix quick links

hkvision:
Fix https://github.com/intel-analytics/BigDL/issues/6669


hkvision:
https://hkvision-link-1118.readthedocs.io/en/latest/

PR #6677, opened 2022-11-18
Nano : add individual action for optimizer

rnwang04:
## Description
image

As InferenceOptimizer related ut costs much time, and we found recently `core dumped error in torch1.9 ipex check` occurs frequently, so plan to add individual action check for optimizer to reduce the whole time for ipex action check and better debugging.

### How to test?
- [ ] Unit test
- [ ] Application test


TheaperDeng (approved):
No objection about this, but could we have a reason in pr page just for record?

rnwang04:
> No objection about this, but could we have a reason in pr page just for record?

done

MeouSker77 (approved):
LGTM

Issue #6573, opened 2022-11-11
PyTorch pyspark backend get wrong num_samples


hkvision:
https://github.com/intel-analytics/BigDL/blob/main/python/orca/test/bigdl/orca/learn/ray/pytorch/test_estimator_pyspark_backend.py#L313

When I run this unit test, from the return stats, train num samples is 49 and val num samples is 51, should be 100?

PR #6687, opened 2022-11-21
Fix wrong num_steps in PyTorch pyspark estimator evaluate

hkvision:
Fix https://github.com/intel-analytics/BigDL/issues/6573

1. Bug for Pyspark estimator evaluate: for SparkXShards input, should explicitly set wrap_dataloader=False.
2. Bug for Pyspark estimator fit:
- When return state_dict and stats, should be one list instead of a tuple of dict and list; otherwise worker_stats will have one extra list dimension
- Stats for all workers are not correctly get, res[1] is only the stats for the first worker.

Checked for both spark and ray backends, fit and evaluate all wrap_dataloader=False for SparkXShards.
predict doesn't involve wrap_dataloader.

Tested the stats for both fit and evaluate


hkvision:
PR Validation / Orca-Python-Ray-Py37-Spark3 (pull_request) 
https://github.com/intel-analytics/BigDL/actions/runs/3512360446/jobs/5900045434

PR #6697, opened 2022-11-21
Chronos: add guide for developers to add tags for unit test

plusbang:
## Description

Chronos provides different installation options ([pytorch], [tensorflow], [automl], [inference], [distributed], [all]) but our unit tests are mixed together. 
In fact, this causes some trouble for our nightly tests and installation tests.

This guide demonstrates how to add different tags to Chronos unit-tests.

### 2. User API changes

None

### 4. How to test?
- [x] N/A

### DIFF
--- a/docs/docs/DeveloperGuide/chronos_ut_tag.md
+++ b/docs/docs/DeveloperGuide/chronos_ut_tag.md
@@ -1,0 +1,51 @@
+# Add tags to Chronos unit-test
+
+This page demonstrates how to add different tags to Chronos unit-tests.
+
+## Motivation
+
+Chronos provides different installation options ([pytorch], [tensorflow], [automl], [inference], [distributed], [all]) but our unit tests are mixed together. In fact, this causes some trouble for our nightly tests and installation tests.
+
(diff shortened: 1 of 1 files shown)


rnwang04 (approved):
In general, I can understand the content of this guide, although I think the current tag itself is a bit complicated.
Just two small questions here:
1.  I don't quite understand the meaning of this tag `op_diff_set_all`
2. In the example, why one tag in class and one tag in function ? What's the rules for this behavior?

plusbang:
Chronos now have [pytorch] [tensorflow] [distributed] [automl] [inference] installation options and current tags are designed to correspond to these options. (inevitably complicated ...)
About the two questions:
1) The tag `op_diff_set_all` refers to dependencies('pmdarima==1.8.5', 'prophet==1.1.0', 'tsfresh==0.17.0', 'pyarrow==6.0.1').
2) In the example, all tests of the class need tag `op_torch`, so we can directly add tag `op_torch` to the class instead of adding to every function.
I will add more description around the the two questions in the guide.

PR #6704, opened 2022-11-21
Fix option not working in source bigdl-nano-init

yangw1234:
## Description

Fix option not working in `source bigdl-nano-init`.

Setting OPTIND to 1 before calling `getopts`,

### DIFF
--- a/python/nano/scripts/bigdl-nano-init
+++ b/python/nano/scripts/bigdl-nano-init
@@ -55,6 +55,8 @@ TCMALLOC=1
 # Set TF_ENABLE_ONEDNN_OPTS
 export TF_ENABLE_ONEDNN_OPTS=1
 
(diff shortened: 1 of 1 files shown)


python/nano/scripts/bigdl-nano-init
```diff
@@ -55,6 +55,8 @@ TCMALLOC=1
 # Set TF_ENABLE_ONEDNN_OPTS
 export TF_ENABLE_ONEDNN_OPTS=1
 
+OPTIND=1
```
yangw1234, line 58: `OPTIND` might not be 1 since this is value from the current shell when we use `source bigdl-nano-init`

MeouSker77 (approved):
LGTM

leonardozcm (approved):

PR #6705, opened 2022-11-22
Update tuned malloc conf

yangw1234:
## Description

Update tuned malloc conf as suggested by IPEX team


jason-dai:
Shall we make this default? @yangw1234 @TheaperDeng

PR #6720, opened 2022-11-22
fix `convert_pb_to_xml`

smurf-1119:
## Description

Modify the "mo_cmd" in `bigdl.nano.deps.openvino.core.utils.py`.

### Why the change?

To support convert the tensorflow model to the openvino model.

### How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...


liangs6212:
Perhaps `pb` is difficult to convert to `xml`, perhaps we should use `onnxruntime` as a conversion medium. @TheaperDeng

PR #6725, opened 2022-11-22
Nano: save/load thread control for onnxruntime model

TheaperDeng:
### 1. Why the change?

Same as openvino, we need to recover the runtime env besides the model checkpoint when users call `InferenceOptimizer.load`.

After this PR, a typical metadata.yml would be
```yaml
ModelType: PytorchONNXRuntimeModel
inter_op_num_threads: 1
intra_op_num_threads: 1
onnx_path: onnx_saved_model.onnx

```

### 2. User API changes

nothing

### 3. Summary of the change 

add 2 keys `intra_op_num_threads` and `inter_op_num_threads` to meta data.

### 4. How to test?
- [ ] Unit test


yangw1234:
Will the user load the model in another machine? Maybe they should have the option to override the threading control value?

TheaperDeng:
> Will the user load the model in another machine? Maybe they should have the option to override the threading control value?

Good suggestion, we should design this API (maybe in later PRs) for not only onnxruntime

PR #6727, opened 2022-11-23
Update PyTorch Estimator doc

hkvision:
Correct the inline doc for PyTorch estimator.
Change the order based on importance.
Remove initialization_hook as it is not used anywhere.

The doc for bigdl backend is not included as it will be deprecated in the future.

### DIFF
--- a/python/orca/src/bigdl/orca/learn/pytorch/estimator.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/estimator.py
@@ -25,57 +25,59 @@ class Estimator(object):
                    optimizer,
                    loss=None,
                    metrics=None,
-                   scheduler_creator=None,
                    backend="spark",
                    config=None,
-                   training_operator_cls=TrainingOperator,
-                   initialization_hook=None,
+                   workers_per_node=1,
+                   scheduler_creator=None,
                    scheduler_step_freq="batch",
                    use_tqdm=False,
-                   workers_per_node=1,
                    model_dir=None,
                    sync_stats=False,
                    log_level=logging.INFO,
                    log_to_driver=True,
+                   training_operator_cls=TrainingOperator,
                    ):
         """
         Create an Estimator for PyTorch.
 
-        :param model: A model creator function that returns a PyTorch model.
-               You can directly input a PyTorch model instance for bigdl backend.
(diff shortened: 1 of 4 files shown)


python/orca/src/bigdl/orca/learn/pytorch/estimator.py
```diff
@@ -25,58 +25,61 @@ def from_torch(*,
...
-        :param use_tqdm: parameter for `horovod` and `ray` backends. You can monitor
-               training progress if use_tqdm=True.
-        :param workers_per_node: parameter for `horovod` and `ray` backends. worker
-               number on each node. default: 1.
-        :param model_dir: parameter for `bigdl` and `spark` backend. The path to save model. During
-               the training, if checkpoint_trigger is defined and triggered, the model will be
-               saved to model_dir.
-        :param backend: You can choose "horovod",  "ray", "bigdl" or "spark" as
-               backend. Default: `spark`.
+        :param model: A model creator function that takes the parameter "config" and returns a
```
lalalapotter, line 43: Is it necessary to keep `bigdl` backend related comments? such as PyTorch model if backend="bigdl"

python/orca/src/bigdl/orca/learn/pytorch/estimator.py
```diff
@@ -25,58 +25,61 @@ def from_torch(*,
...
+               Default: None if no validation is involved.
+        :param backend: The distributed backend for the Estimator. One of "spark",  "ray",
+               "bigdl" or "horovod".
+               Default: "spark".
+        :param config: A parameter config dict, CfgNode or any class instance that plays a role of
+               configuration to create model, loss, optimizer, scheduler and data.
+               Default: None if no config is needed.
+        :param workers_per_node: The number of PyTorch workers on each node.
+               Default: 1.
+        :param scheduler_creator: An scheduler creator function that has two parameters "optimizer"
```
lalalapotter, line 61: A scheduler

python/orca/src/bigdl/orca/learn/pytorch/estimator.py
hkvision: It will be deprecated in the future and not recommended and thus removed from the doc.

PR #6732, opened 2022-11-23
Implement MMCVRayEstimator get_model method

PatrickkZ:
## Description

- Implement MMCVRayEstimator get_model() method
- Implement MMCVRayEpochRunner get_state_dict() and load_state_dict()
- add uint test

### 1. Why the change?

To get the state dict of the trained model, or load state dict to a model.

### 2. User API changes

user can get the model state dict through `MMCVRayEstimator` get_model(), or get more info from get_state_dict()
```python
# create a MMCVRayEstimator
estimator = ...

model_state = estimator.get_model()
state_dict = estimator.get_state_dict()
```
`model_state` is an OrderedDict contains the model state, example as below
```
OrderedDict([('fc1.weight', tensor([[1.9995]]))])
```

`state_dict` is a dict contains more information about the model, include `epoch`, `model` and `optimizer`, example as below
```
{'epoch': 20, 'model': OrderedDict([('fc1.weight', tensor([[1.9995]]))]), 'optimizer': {'state': {0: {'momentum_buffer': None}}, 'param_groups': [{'lr': 0.01, 'momentum': 0, 'dampening': 0, 'weight_decay': 0, 'nesterov': False, 'maximize': False, 'initial_lr': 0.01, 'params': [0]}]}}
```

user can also load a state_dict to a model through load_state_dict()
```python
state_dict = {'epoch': 10,
              'model': {
               'fc1.weight': torch.tensor([[3.0]], dtype=torch.float32)}
              # optimizer is optional
           }

estimator.load_state_dict(state_dict)
```

### 3. Summary of the change 

support get/load model state information

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...


python/orca/src/bigdl/orca/learn/pytorch/experimential/mmcv/mmcv_ray_runner.py
```diff
@@ -198,11 +198,19 @@ def validate(self, **kwargs):
 
     def get_state_dict(self):
         """Returns the state of the runner."""
-        pass
+        state = {
+            "epoch": self._epoch,
+            "model": self.model.module.state_dict(),
+            "optimizer": self.optimizer.state_dict()
+        }
```
leonardozcm, line 205: How do mmcv save scheduler(in their scheduler hooks)?

python/orca/src/bigdl/orca/learn/pytorch/experimential/mmcv/mmcv_ray_runner.py
```diff
@@ -198,11 +198,19 @@ def validate(self, **kwargs):
...
+        return state
 
     def load_state_dict(self, state):
         """Sets the state of the model."""
-        pass
+        self.model.module.load_state_dict(state["model"])
```
leonardozcm, line 210: Then I believe our `save_checkpoint` also needs to be fixed as model.module. Can we keep self.model not in DDP mode?

leonardozcm (changes requested):
Redirect to inherit new modelIO.

python/orca/test/bigdl/orca/learn/ray/experimental/test_mmcv_estimator_ray_backend.py
```diff
@@ -218,6 +283,28 @@ def test_run_with_batch_processor(self):
...
+        state_dict = estimator.get_state_dict()
+        assert 'epoch' in state_dict
+        assert 'model' in state_dict
+        assert 'optimizer' in state_dict
+        self.assertEqual(state_dict['epoch'], 30)
+
+        # test get_model()
+        model_state = estimator.get_model()
+        weight = model_state["fc1.weight"].item()
+        assert abs(weight - 2.0) < 0.0001
```
leonardozcm, line 300: we should:
1. train a model, record its loss or something indicates its performance
2. save the model, and then load it
3. validate its performance

python/orca/src/bigdl/orca/learn/pytorch/experimential/mmcv/mmcv_ray_estimator.py
```diff
@@ -96,4 +96,9 @@ def evaluate(self, **kwargs):
         pass
 
     def get_model(self):
-        pass
+        """
+        Returns the learned PyTorch model state dict
+        """
+        state = self.get_state_dict()
+        model_state = state["model"]
+        return model_state
```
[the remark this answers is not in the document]
PatrickkZ: Pytorch provide a way to [save/load the entire model](https://pytorch.org/tutorials/beginner/saving_loading_models.html#save-load-entire-model). but there is a disadvantage

> the serialized data is bound to the specific classes and the exact directory structure used when the model is saved. The reason for this is because pickle does not save the model class itself. Rather, it saves a path to the file containing the class

so Model class must be defined somewhere, but mmcv estimator do not have the Model class, so we can not simply use torch.save/load to get the entire model.

And Pytorch recommends to save the model state dict instead of the entire model. so later in mmcv and pytorch estimator, the get_model method will unify to return the model state dict.

python/orca/src/bigdl/orca/learn/pytorch/experimential/mmcv/mmcv_ray_runner.py
PatrickkZ: mmcv use `LrUpdaterHook` to replace LR scheduler, and do not save scheduler info

PR #6733, opened 2022-11-23
[PPML] Fix xgboost no lib and use copy_bom to generate occlum spark content

hzjane:
## Description
Fix xgboost no lib and use copy_bom to generate occlum spark content


qiyuangong (approved):
LGTM

PR #6738, opened 2022-11-23
Chronos: fix validation process when previous workflow was stopped

TheaperDeng:
### 1. Why the change?

When users stopped the previous training by CTRL+C or something brute, the ramain folder might make next round of validation fail

### 2. User API changes

nothing

### 3. Summary of the change 
delete the folder before fitting

- [x] move the folder to a tmp place rather than pwd.

### 4. How to test?
- [x] Unit test


plusbang (commented):
LGTM, and I think we can use this method in Chronos other functions.

plusbang (approved):

Issue #6675, opened 2022-11-18
InferenceOptimizer context manager proposal


TheaperDeng:
# InferenceOptimizer context manager proposal

**Only a preview idea for discussion, will experimentally implement it once we think this is a good way to move on.**

## Why?

We found many BKCs (`torch.no_grad`, `autocast`, `set_num_threads`, instruction check, framework version check...) that users should apply or set correctly while we could not do them automatically during our optimized model's forwarding such as  `model(x)`, mostly due to the performance degrdation if we keep doing it again and again when users has such a iterative calling:

```python
for data in dataloader:
    model(data)
```

Still, we try our best to not let our users to set too much trivial and error-prone settings or checks. That's why we propose to have a context manager generated by `InferenceOptimizer` together with the best optimized model.

## How does it looks like?

User will get a context manager (this manager does a lot of things, will illustrate in detail later), and use it upon all the inference code they would like to use nano optimized model.

```python
inference_opt = InferenceOptimizer()
inference_opt.optimize(model, ...)
model_nano = inference_opt.get_model("bf16_ipex")
context_manager_nano = inference_opt.get_context_manager("bf16_ipex")

with context_manager_nano:  # <--- The only thing changed
    # all the code users would like to use for inference before
```

I understand some API like `InferenceOptimizer.trace/quantize` may not have a consistent and compatible API as before, we may have some discussion such as add a parameter like `model, context_mgr = InferenceOptimizer.trace(..., context_manager=True)`. 

## What will the context manger do?

- apply `autocast` for bf16 models

- apply thread_control for pytorch framework models

- check if required instruction set is provided on the server (like bf16)

- check if required library version is provided on the server (like we need torch>=1.12 for bf16)

- (possibly) apply good thread control for openvino/onnxruntime by opening a new thread for the code under this context manager

- ...

Since the context manager's `__enter__` is only called once, it should be ok if the latency is a little bit hight.

TheaperDeng:
@MeouSker77 @rnwang04 @yangw1234 @jason-dai Please provide any comment you have.

yangw1234:
What will be called in the `__exit__`. If the code in `__exit__` is not necessary, maybe we can just call those code in `__enter__` when user first call `__call__`?
E.g
```python

class OptimizedModel:

    def __call__:
        if not self.initialized:
             self.setup()
             self.initialized = True
        with torch.no_grad():
            with torch.autocast(enabled=self.bf16_enabled):
                # regular forward code
```

[...]

PR #6747, opened 2022-11-24
Nano: Support new context manager

rnwang04:
## Description

During OOB, we found that:

- autocast inside inference loop will cause ~30% overhead, so we need a autocast context manager outside the loop
- users always forget to use `with torch.no_grad():`

So we decide to provide a context manager inside each model to provide corresponding customized context.
We will implement torch._no_grad and autocast first, and implement thread control in next PR.

### 1. Why the change?

https://github.com/intel-analytics/BigDL/issues/6675

### 2. User API changes

Provide a new **`context_manager` attribute inside each accelerated model**
```python

#  for traced model
acc_model = InferenceOptimizer.trace(...)
with acc_model.context_manager:
    do_somethig()

# for quantized model
acc_model = InferenceOptimizer.quantize(...)
with acc_model.context_manager:
    do_somethig()

# for optimized model
opt = InferenceOptimizer()
opt.optimize(...)
model = opt.get_best_model()
with model.context_manager:
    do_something()
model = opt.get_model(xxx)
with model.context_manager:
    do_something()
```

### 3. Summary of the change 

- [x] Provide two context manager for autocast and no_grad
- [x] UT for Openvino / ONNXRuntime / BF16 / IPEXJIT / IPEX JITBF16 / INC
- [x] UT for optimize and get_model / get_best_model

### 4. How to test?
- [ ] Unit test
- [ ] Application test

### DIFF
--- a/python/nano/src/bigdl/nano/deps/neural_compressor/inc_api.py
+++ b/python/nano/src/bigdl/nano/deps/neural_compressor/inc_api.py
@@ -67,7 +67,7 @@ def quantize(model, dataloader=None, metric=None, **kwargs):
         quantizer = BaseQuantization(**not_none_kwargs)
     model = quantizer.post_training_quantize(model, dataloader, metric)
     if 'pytorch' in not_none_kwargs['framework'] or \
-        ('onnx' in not_none_kwargs['framework'] and onnx_option != 'tensorflow'):
-            from bigdl.nano.pytorch.context_manager import BaseContextManager
-            model.context_manager = BaseContextManager()
+            'onnx' in not_none_kwargs['framework'] and onnx_option != 'tensorflow':
+        from bigdl.nano.pytorch.context_manager import BaseContextManager
+        model.context_manager = BaseContextManager()
     return model
--- a/python/nano/src/bigdl/nano/deps/neural_compressor/inc_api.py
+++ b/python/nano/src/bigdl/nano/deps/neural_compressor/inc_api.py
@@ -65,4 +65,9 @@ def quantize(model, dataloader=None, metric=None, **kwargs):
         # default quantization
         from .core import BaseQuantization
         quantizer = BaseQuantization(**not_none_kwargs)
-    return quantizer.post_training_quantize(model, dataloader, metric)
+    model = quantizer.post_training_quantize(model, dataloader, metric)
+    if 'pytorch' in not_none_kwargs['framework'] or \
+        ('onnx' in not_none_kwargs['framework'] and onnx_option != 'tensorflow'):
+            from bigdl.nano.pytorch.context_manager import BaseContextManager
+            model.context_manager = BaseContextManager()
+    return model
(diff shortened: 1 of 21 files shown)


python/nano/src/bigdl/nano/deps/neural_compressor/inc_api.py
```diff
@@ -65,4 +65,9 @@ def quantize(model, dataloader=None, metric=None, **kwargs):
         # default quantization
         from .core import BaseQuantization
         quantizer = BaseQuantization(**not_none_kwargs)
-    return quantizer.post_training_quantize(model, dataloader, metric)
+    model = quantizer.post_training_quantize(model, dataloader, metric)
+    if 'pytorch' in not_none_kwargs['framework'] or \
+            'onnx' in not_none_kwargs['framework'] and onnx_option != 'tensorflow':
+        from bigdl.nano.pytorch.context_manager import BaseContextManager
+        model.context_manager = BaseContextManager()
```
TheaperDeng, line 72: why we need line 69-71 if we have add the assignment in `PytorchQuantizedModel` and `PytorchONNXRuntimeModel`?

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -364,63 +368,64 @@ def optimize(self, model: nn.Module,
...
-                if self._calculate_accuracy:
-                    # here we suppose trace don't change accuracy,
-                    # so we jump it to reduce time cost of optimize
-                    if precision == "fp32" and method != "original":
-                        result_map[method]["accuracy"] = "not recomputed"
-                    else:
-                        if method == "original":
-                            # test whether metric works
-                            try:
+                    torch.set_num_threads(default_threads)
```
TheaperDeng, line 395: do we still need this?

rnwang04: yeah, we still need this

python/nano/src/bigdl/nano/deps/neural_compressor/inc_api.py
rnwang04: Have deleted the lines.

TheaperDeng (approved):
LGTM for the first commit of ctx mgr

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -338,6 +339,9 @@ def optimize(self, model: nn.Module,
         else:
             sample_size_for_pot = 100
 
+        # patch context manager
+        model.context_manager = BaseContextManager()
```
jason-dai, line 343: Change to `model._nano_context_manager` to avoid potential conflict?

And we should use something like `with InferenceModel.get_context(...)` or `with InferenceModel.set_context(...)` to ensure interface consistency.

rnwang04: Will change to `model._nano_context_manager`.
And we plan to support `with InferenceModel.get_context(...)` in this PR https://github.com/intel-analytics/BigDL/pull/6782

PR #6751, opened 2022-11-24
[ORCA] Support save and load an entire model in PyTroch Pyspark Estimator

sgwhat:
## Description
This PR is mainly proposed to support `save/load` the entire model, pre-PR for https://github.com/intel-analytics/BigDL/pull/6875.

### 1. Why the change?
Optional `model_creator` needs pytorch estimator to support `save/load` an entire model, but we only support the model `state_dict` in the previous design. 

### 2. User API changes
```python
# save the state_dict
est.save(model_path)

# save the entire model
est.save(model_path, entire=True)

# nothing changes in load
est.load(model_path)
```

### 3. Summary of the change 
PyTorch pyspark estimator supports `save/load` an entire model.

### 4. How to test?
- [x] Unit test
- [x] Yarn test


python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
```diff
@@ -466,17 +466,22 @@ def get_state_dict(self):
 
     def load_state_dict(self, state):
         """Sets the state of the model."""
-        for model, state_dict in zip(self.models, state["models"]):
-            model.load_state_dict(state_dict)
+        import collections
+        if isinstance(state, collections.OrderedDict):
+            for model, state_dict in zip(self.models, [state]):
```
hkvision, line 471: Is this reasonable? It will always require self.models to be a singleton list.

hkvision (approved):
LGTM except the concern above.

python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
sgwhat: collections.singletonList? I think it's a java method and `[state]` should always be a python list type.

hkvision: No... I mean since [state] is of length 1, you do the zip, then you assume self.models is of length 1, does it add constraint/limitation to the current design?

sgwhat: yes you're right, this should be updated.

hkvision:
LGTM

PR #6753, opened 2022-11-24
Ncf add feature tf xshard

ForJadeForest:
## Description
This pr is to use xshard data to train tensorflow model NCF.


python/orca/tutorial/NCF/tf_train_xshards.py
```diff
@@ -13,3 +13,79 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+import math
```
hkvision, line 16: put this into step0

PR #6755, opened 2022-11-24
Add regression test for pytorch InferenceOptimizer

MeouSker77:
## Description

Add regression test for pytorch InferenceOptimizer, including single process throughput and multi-process throughput of all optimize options(21 sorts)

### DIFF
--- a/python/nano/benchmark/torch_inference_optimizer/inference.py
+++ b/python/nano/benchmark/torch_inference_optimizer/inference.py
@@ -1,0 +1,70 @@
+import time
+import os
+import sys
+import argparse
+
+import torch
+from torch import nn
+
+from bigdl.nano.pytorch.vision.models import vision
+from bigdl.nano.pytorch import InferenceOptimizer
+
+
+class ResNet18(nn.Module):
+    def __init__(self, num_classes, pretrained=True, include_top=False, freeze=True):
+        super().__init__()
+        backbone = vision.resnet18(pretrained=pretrained, include_top=include_top, freeze=freeze)
+        output_size = backbone.get_output_size()
+        head = nn.Linear(output_size, num_classes)
+        self.model = nn.Sequential(backbone, head)
+
+    def forward(self, x):
+        return self.model(x)
+
+
+options1 = ["original", "fp32_channels_last", "fp32_ipex", "fp32_ipex_channels_last", "bf16",
+            "bf16_channels_last", "bf16_ipex", "bf16_ipex_channels_last", "int8", "int8_ipex",
+            "jit_fp32", "jit_bf16", "jit_fp32_ipex", "jit_fp32_ipex_channels_last",
+            "jit_bf16_ipex", "jit_bf16_ipex_channels_last", ]
+options2 = ["openvino_fp32", "openvino_int8", "onnxruntime_fp32", "onnxruntime_int8_qlinear",
+            "onnxruntime_int8_integer"]
+
+
+def run(args):
+    save_dir = "models"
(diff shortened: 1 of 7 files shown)


python/nano/dev/build_and_install.sh
```diff
@@ -43,7 +43,7 @@ if [ $framework == "basic" ];
 then
   # when framework is specified to "basic", nano will be installed without
   # any framework like Pytorch and TensorFlow
-  pip install $install_options dist/${whl_name}
+  pip install --force-reinstall $install_options dist/${whl_name}
```
MeouSker77, line 46: Here, if we have used this script to install nano, its version will be 2.1.0.dev0. When we try to use this script to install nano again after doing some changes, the version is still 2.1.0.dev0, so pip will refuse to install it again, because there is already a installed nano with same version. Fix this bug.

python/nano/benchmark/torch_inference_optimizer/inference.py
```diff
@@ -0,0 +1,87 @@
...
+        backbone = vision.resnet18(pretrained=pretrained, include_top=include_top, freeze=freeze)
+        output_size = backbone.get_output_size()
+        head = nn.Linear(output_size, num_classes)
+        self.model = nn.Sequential(backbone, head)
+
+    def forward(self, x):
+        return self.model(x)
+
+
+options1 = ["original", "fp32_channels_last", "fp32_ipex", "fp32_ipex_channels_last", "bf16",
```
rnwang04, line 42: why setting options manually other than just set mode=all?

python/nano/benchmark/torch_inference_optimizer/optimize.py
```diff
@@ -0,0 +1,95 @@
...
+        backbone = vision.resnet18(pretrained=pretrained, include_top=include_top, freeze=freeze)
+        output_size = backbone.get_output_size()
+        head = nn.Linear(output_size, num_classes)
+        self.model = nn.Sequential(backbone, head)
+
+    def forward(self, x):
+        return self.model(x)
+
+
+options = ["original", "fp32_channels_last", "fp32_ipex", "fp32_ipex_channels_last", "bf16",
```
rnwang04, line 60: If set this manually, we have to update time to time.
How about obtain this by
```
options = InferenceOptimizer.ALL_INFERENCE_ACCELERATION_METHOD.keys()
```

python/nano/benchmark/torch_inference_optimizer/inference.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def run(args):
+    save_dir = "models"
+    imgs = torch.rand((1000, 1, 3, 128, 128))
+    opt = InferenceOptimizer()
+    if args.option in options1:
+        try:
+            model = ResNet18(10)
+            model = opt.load(os.path.join(save_dir, args.option), model)
```
rnwang04, line 57: load jit model don't need original model.

python/nano/benchmark/torch_inference_optimizer/run.sh
```diff
@@ -0,0 +1,54 @@
...
+source bigdl-nano-unset-env
+source bigdl-nano-init
+
+# cd workload directory
+cd $WORKLOAD_DIR
+
+export OMP_NUM_THREADS=4
+taskset -c 0-3 python optimize.py
+
+options="original fp32_channels_last fp32_ipex fp32_ipex_channels_last bf16 \
```
rnwang04, line 38: the same for this.
Is this any possibility to set this automatic?

rnwang04 (commented):
others LGTM

PR #6756, opened 2022-11-24
Remove cache when Spark DataFrame to SparkXShards

hkvision:
Description and analysis see here: https://github.com/analytics-zoo/orca/issues/136

@jason-dai Take a look and see if you have any comments.

### DIFF
--- a/python/orca/src/bigdl/orca/data/shard.py
+++ b/python/orca/src/bigdl/orca/data/shard.py
@@ -351,34 +351,23 @@ class SparkXShards(XShards):
                 else:
                     dtypes.append(v.dtype)
             if dict_of_batched_ndarray:
-                if num_partitions > self.rdd.getNumPartitions():
-                    def dict_to_unbatched_list(d):
-                        values = [list(d[k]) for k in keys]
-                        return list(zip(*values))
+                def dict_to_unbatched_list(d):
+                    values = [list(d[k]) for k in keys]
+                    return list(zip(*values))
 
-                    def to_batched_dict(iter):
-                        batch_values = list(zip(*iter))
-                        if not batch_values:
-                            return []
-                        batch_ndarrays = [np.stack(v, axis=0).astype(dtype)
-                                          for v, dtype in zip(batch_values, dtypes)]
-                        return [dict(zip(keys, batch_ndarrays))]
+                def to_batched_dict(iter):
+                    batch_values = list(zip(*iter))
+                    if not batch_values:
+                        return []
+                    batch_ndarrays = [np.stack(v, axis=0).astype(dtype)
+                                      for v, dtype in zip(batch_values, dtypes)]
+                    return [dict(zip(keys, batch_ndarrays))]
 
-                    # If number of records in a partition <= 10, may produce empty partition
-                    rdd = self.rdd.flatMap(lambda data: dict_to_unbatched_list(data)) \
-                        .repartition(num_partitions)
-                    repartitioned_shard = self._create(rdd.mapPartitions(
-                        lambda iter: to_batched_dict(iter)), class_name=class_name)
-                else:
-                    rdd = self.rdd.coalesce(num_partitions)
-
-                    def merge_list_of_dict(iter):
-                        iter_list = list(iter)
-                        return [{k: np.concatenate([d[k] for d in iter_list], axis=0)
-                                 for k in keys}]
-
-                    repartitioned_shard = self._create(rdd.mapPartitions(
-                        lambda iter: merge_list_of_dict(iter)), class_name=class_name)
+                # If number of records in a partition <= 10, may produce empty partition
+                rdd = self.rdd.flatMap(lambda data: dict_to_unbatched_list(data)) \
+                    .repartition(num_partitions)
+                repartitioned_shard = self._create(rdd.mapPartitions(
+                    lambda iter: to_batched_dict(iter)), class_name=class_name)
             else:
                 repartitioned_shard = self._create(self.rdd.repartition(num_partitions),
                                                    class_name=class_name)
--- a/python/orca/src/bigdl/orca/data/shard.py
+++ b/python/orca/src/bigdl/orca/data/shard.py
@@ -159,7 +159,6 @@ But got data of type {}
 
 class SparkXShards(XShards):
     """
-
     A collection of data which can be pre-processed in parallel on Spark
     """
     def __init__(self,
@@ -197,6 +196,10 @@ class SparkXShards(XShards):
         :return: An instance of LazySparkXShards.
         """
         return LazySparkXShards(self.rdd, self._get_class_name())
+
+    @property
+    def is_lazy(self):
+        return isinstance(self, LazySparkXShards)
 
     def transform_shard(self, func: Callable, *args) -> "SparkXShards":
         """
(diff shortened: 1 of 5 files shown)


python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -271,35 +270,24 @@ def repartition(self, num_partitions: int) -> "SparkXShards":
         if self._get_class_name() == 'pandas.core.frame.DataFrame':
             import pandas as pd
 
-            if num_partitions > self.rdd.getNumPartitions():
```
hkvision, line 274: @ForJadeForest This would fix the error
```
py4j.protocol.Py4JJavaError: An error occurred while calling z:org.apache.spark.api.python.PythonRDD.collectAndServe.
: org.apache.spark.scheduler.BarrierJobUnsupportedRDDChainException: [SPARK-24820][SPARK-24821]: Barrier execution mode does not allow the following pattern of RDD chain within a barrier stage:
1. Ancestor RDDs that have different number of partitions from the resulting RDD (e.g. union()/coalesce()/first()/take()/PartitionPruningRDD). A workaround for first()/take() can be barrierRdd.collect().head (scala) or barrierRdd.collect()[0] (python).
2. An RDD that depends on multiple barrier RDDs (e.g. barrierRdd1.zip(barrierRdd2)).
```

python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -271,35 +270,24 @@ def repartition(self, num_partitions: int) -> "SparkXShards":
...
-                    if len(dfs) > 0:
-                        return [pd.concat(dfs)]
-                    else:
-                        return iter
+            rdd = self.rdd \
+                .flatMap(lambda df: df.apply(lambda row: (row[0], row.values.tolist()), axis=1)
+                         .values.tolist()) \
+                .partitionBy(num_partitions)
 
-                rdd = self.rdd.coalesce(num_partitions)
```
hkvision, line 301: For Estimators that use Spark barrier mode, if input is spark xshards, we can't use coalesce for repartitions as coalesce can't work together with barrier mode.
```
py4j.protocol.Py4JJavaError: An error occurred while calling z:org.apache.spark.api.python.PythonRDD.collectAndServe.
: org.apache.spark.scheduler.BarrierJobUnsupportedRDDChainException: [SPARK-24820][SPARK-24821]: Barrier execution mode does not allow the following pattern of RDD chain within a barrier stage:
1. Ancestor RDDs that have different number of partitions from the resulting RDD (e.g. union()/coalesce()/first()/take()/PartitionPruningRDD). A workaround for first()/take() can be barrierRdd.collect().head (scala) or barrierRdd.collect()[0] (python).
2. An RDD that depends on multiple barrier RDDs (e.g. barrierRdd1.zip(barrierRdd2)).
```
@jason-dai @jenniew What's your opinion on this?
Even if coalesce can work, it will very likely to cause rather unbalanced partitions?

python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -271,35 +270,24 @@ def repartition(self, num_partitions: int) -> "SparkXShards":
...
+                data = [value[1] for value in list(iter)]
+                if data:
+                    df = pd.DataFrame(data=data, columns=schema['columns']) \
+                        .astype(schema['dtypes'])
+                    return [df]
+                else:
+                    # no data in this partition
+                    return iter
+
+            repartitioned_shard = SparkXShards(rdd.mapPartitions(merge_rows), transient=transient)
```
jason-dai, line 290: `transient=transient` should be used in all the ` repartitioned_shard` objects in this method

python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -377,7 +365,8 @@ def merge_list_of_dict(iter):
                 repartitioned_shard = SparkXShards(self.rdd.repartition(num_partitions))
         else:
             repartitioned_shard = SparkXShards(self.rdd.repartition(num_partitions))
-        self._uncache()
+        if not transient:
+            self._uncache()
```
jason-dai, line 369: Instead of using `transient` here, you may want to explicitly call `SparkXShards.cache` in the estimator

python/orca/src/bigdl/orca/learn/utils.py
```diff
@@ -326,20 +336,22 @@ def add_row(data, results, current):
...
-            result["y"] = df[label_cols[0]].to_numpy()
+            if label_cols:
+                result["y"] = df[label_cols[0]].to_numpy()
 
-        return result
+            yield result
 
-    data = data.transform_shard(to_shard_dict)
+    dict_rdd = data.rdd.mapPartitions(to_shard_dict)
+    data = SparkXShards(dict_rdd, transient=True)
```
jason-dai, line 354: `data = SparkXShards(dict_rdd, transient=True, class_name=...)`

python/orca/src/bigdl/orca/data/shard.py
jenniew: use coalesce can reduce shuffling, but may cause unbalanced partitions.

python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -351,34 +351,23 @@ def merge_rows(iter):
                 else:
                     dtypes.append(v.dtype)
             if dict_of_batched_ndarray:
-                if num_partitions > self.rdd.getNumPartitions():
```
hkvision, line 354: Remove coalesce for SparkXShards of dict as well since it may also be the input of users.

python/orca/src/bigdl/orca/learn/tf2/pyspark_estimator.py
```diff
@@ -372,29 +379,39 @@ def predict(self, data, batch_size=None, verbose=1,
...
         )
 
+        def transform_func(iter, init_param, param):
+            partition_data = list(iter)
+            # res = combine_in_partition(partition_data)
+            param["data_creator"] = make_data_creator(partition_data)
+            return SparkRunner(**init_param).predict(**param)
+
         if isinstance(data, DataFrame):
-            pre_predict_data = data.repartition(self.num_workers)
```
hkvision, line 376: repartition is removed as it is not needed for predict. Also for the result data, the user would expect the partition to be unchanged.

jenniew: If we removed repartition, if the original partition number is big( like 2000 partitions), on each partition it will create a model to do predict. So there will be 2000 models created. If the model is big, there may cause memory issue. I think we can change to coalesce to reduce shuffling.

qiuxin2012: On spark, the task running in the same time equals to the total cores,  Model will be GCed after the partition is finished. So there won't be 2000 models in the same time.

jenniew: even there are number of cores of models are running at the same time, if the model is big, there may have memory issue. @qiuxin2012 , is it right?

qiuxin2012: @jenniew Yes, so user should ensure they have enough memory matching their executor cores, or they should decrease the core number.

python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -175,28 +174,42 @@ def __init__(self,
...
+            self.is_lazy = False
+        else:
+            self.is_lazy = True
         self.type = {}
         if class_name:
             self.type['class_name'] = class_name
 
     @classmethod
-    def _create(cls,
+    def lazy(cls,
```
hkvision, line 185: @jason-dai Using LazySparkXShards would unexpectedly caused uncache for the original xshards, reason not found. Thus removed this subclass and achieve the lazy functionality directly in SparkXShards.
Take a look. See more details here: https://github.com/analytics-zoo/orca/issues/136#issuecomment-1337037891

hkvision:
One job fails due to network and is irrelevant.
Merge it first and then will work on pytorch.

PR #6768, opened 2022-11-25
Orca: Refactor TorchRayEst. with BaseRayEst. ABC

leonardozcm:
## Description

BaseRayEstimator defines basic user APIs with abstract methods and contains common logic such as DDP setup shared by others. BTW this pr doesn't involve modification of pyspark backend as they have only little part in common.

All pytorch ray estimators should inherit `BaseRayEstimator` as the driver interface and implement the following APIs:
```python
    @abstractmethod
    def fit(self, **kwargs):
        """
        Train the model with train data.
        :return: predicted result.
        """

        # Need to preprocess params as pytorch_ray_estimator does.
        #
        # ...
        # self._train_epochs(**kwargs)
        pass

    @abstractmethod
    def predict(self, **kwargs):
        """
        Predict input data.
        :return: predicted result.
        """
        # Need to preprocess params as pytorch_ray_estimator does.
        pass

    @abstractmethod
    def evaluate(self, **kwargs):
        """
        Evaluate model.
        :return: evaluation result as a dictionary of {'metric name': metric value}
        """
        # Need to preprocess params as pytorch_ray_estimator does.
        pass

    @abstractmethod
    def get_model(self):
        """
        Get the trained model.
        :return: Trained model
        """
        # Need to preprocess params as pytorch_ray_estimator does.
        pass
```

https://github.com/intel-analytics/BigDL/pull/6408#issue-1432988407

### DIFF
--- a/python/orca/src/bigdl/orca/learn/pytorch/core/base_ray_estimator.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/core/base_ray_estimator.py
@@ -1,0 +1,267 @@
+#
+# Copyright 2016 The BigDL Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+from bigdl.orca.learn.base_estimator import Estimator
+from bigdl.orca.ray import OrcaRayContext
+from bigdl.orca.learn.pytorch.pytorch_ray_worker import PytorchRayWorker
+from bigdl.orca.data.file import enable_multi_fs_save, enable_multi_fs_load
+
+import numbers
+import numpy as np
+import io
+import torch
+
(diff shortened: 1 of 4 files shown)


python/orca/src/bigdl/orca/learn/pytorch/utils.py
```diff
@@ -37,6 +37,7 @@
 import collections
 import numpy as np
 from contextlib import closing, contextmanager
+import ray
```
hkvision, line 40: do not import ray globally here, put it to local

python/orca/src/bigdl/orca/learn/pytorch/utils.py
```diff
@@ -318,3 +319,12 @@ def mean_reduce_stats(worker_stats, res_stats=None):
         else:
             res_stats[stat_key] = stat_value
     return res_stats
+
+
+def get_driver_node_ip():
```
hkvision, line 324: put it into orca/learn/utils?

python/orca/src/bigdl/orca/learn/pytorch/core/base_ray_estimator.py
```diff
@@ -0,0 +1,264 @@
...
+        Loads the Estimator state (including model and optimizer) from the provided model_path.
+
+        :param model_path: (str) Path to the existing model.
+        """
+        state_dict = torch.load(model_path)
+        self.load_state_dict(state_dict)
+
+    def save_checkpoint(self, model_path):
+        """
+
```
hkvision, line 188: remove the empty line

hkvision (approved):
LGTM except the above minor comments

Issue #6355, opened 2022-10-28
PyTorch Estimator optimizer_creator should be optional


hkvision:
For inference only, optimizer is not needed.

PR #6786, opened 2022-11-28
[ORCA] Orca PyTorch Inference Supports Optional Optimizer

sgwhat:
## Description
One of PR to fix Orca PyTorch inference pipeline. https://github.com/intel-analytics/BigDL/issues/6355

### 1. Why the change?
The previous estimator requires users to set optimizer through `optimizer_creator` for training the model. Btw, the pure inference pipeline (like resnet50 pre-trained model) doesn't need it, and the none-optimizer could cause error happened.

### 2. User API changes
Users could just ignore the `optimizer_creator`.
```python
est = estimator.from_torch(
    model_creator=model_creator,
    backend="spark")
```

### 3. Summary of the change 
`optimzier_creator` could be optional and ignored, which means `optimizer` could be none when doing pure inference pipeline.

### 4. How to test?
- [x] Unit test
- [x] Application test

### 5. Need to do
- [x] Case 1. optimizer_creator
- [x] Case 2. optimizer_creator returns None
- [x] Case 3. optimizer is optional


python/orca/src/bigdl/orca/learn/pytorch/training_operator.py
```diff
@@ -451,12 +453,14 @@ def models(self):
...
-        return self._optimizers[0]
+        if self._optimizers:
+            return self._optimizers[0]
 
     @property
     def optimizers(self):
         """List of optimizers created by the ``optimizer_creator``."""
-        return self._optimizers
+        if self._optimizers:
+            return self._optimizers
```
hkvision, line 463: return None if no optimizer?

python/orca/src/bigdl/orca/learn/pytorch/training_operator.py
```diff
@@ -287,14 +288,15 @@ def train_batch(self, batch, batch_info):
             else:
                 loss = self.criterion(output, target)
 
-        # Compute gradients in a backward pass.
-        with self.timers.record("grad"):
-            self.optimizer.zero_grad()
-            loss.backward()
+        if self.optimizer:
```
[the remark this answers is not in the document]
sgwhat: yes, optimizer must be in train, otherwise backward is pointless, we could remove this.

python/orca/src/bigdl/orca/learn/pytorch/training_operator.py
sgwhat: Since we have assumed optimizers must be in train, we could remove checking the existence.

hkvision (approved):

PR #6787, opened 2022-11-29
Chronos: Improve data process latency by changing apply to for loop

hjzin:
### 1. Why the change?

According to my test, changing `dataframe.groupby().apply()` to for loop will reduce the latency of data processing by 10%. 
### 2. User API changes

N/A
### 3. Summary of the change 

- In some methods, `dataframe.groupby().apply()` is changed to for loop.
### 4. How to test?
- [ ] Unit test


TheaperDeng (approved):
LGTM

PR #6789, opened 2022-11-29
[PPML] python KMS lib

Uxito-Ada:
## Description

Add python KMS support.

### 1. Why the change?

We used to call pykms by python -> scala jars -> spark -> jvm, which is low-performance and has to depend on scala and spark, and this is too heavy for python application such as trusted pytorch. As a result, a pure-python implementation of KMS is needed.

### 2. User API changes

Users are allowed to use kms in their python codes instead of a heavy ppml context.

### 3. Summary of the change 

Add python KMS lib.

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [x] Application test
- [ ] Document test
- [ ] ...

### 5. New dependencies

- [X] New Python dependencies
       requests
       argparse
       cryptography==3.3.2

### DIFF
--- a/python/ppml/src/bigdl/ppml/kms/restcaller.py
+++ b/python/ppml/src/bigdl/ppml/kms/restcaller.py
@@ -66,11 +66,11 @@ def request_parimary_key_ciphertext(ip, port):
     primary_key_ciphertext = post_request(ip, port, action, payload)['keyid']
     return primary_key_ciphertext
 
-def request_data_key_ciphertext(ip, port, encrypted_primary_key):
+def request_data_key_ciphertext(ip, port, encrypted_primary_key, data_key_length):
     action = "GenerateDataKeyWithoutPlaintext"
     payload = {
         "keyid":encrypted_primary_key,
-        "keylen": 32,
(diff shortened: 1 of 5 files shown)


gc-fu (approved):
LGTM

python/ppml/src/bigdl/ppml/kms/restcaller.py
```diff
@@ -0,0 +1,87 @@
...
+        "origin":"EH_INTERNAL_KEY"
+    }
+    primary_key_ciphertext = post_request(ip, port, action, payload)['keyid']
+    return primary_key_ciphertext
+
+def request_data_key_ciphertext(ip, port, encrypted_primary_key):
+    action = "GenerateDataKeyWithoutPlaintext"
+    payload = {
+        "keyid":encrypted_primary_key,
+        "keylen": 32,
```
gc-fu, line 73: Can we add an argument to allow the user to specify the length of the datakey?
Because gramine only supports keylen of 16

Uxito-Ada: ok

PR #6799, opened 2022-11-29
Add shard_size=batch_size when convert Spark DataFrame to SparkXShards

hkvision:
1. Add shard_size=batch_size when convert Spark DataFrame to SparkXShards
2. rename ray_partition_get_data_label related functions to partition_get_data_label as they are not related to ray, the name is confusing.

### DIFF
--- a/python/orca/src/bigdl/orca/learn/utils.py
+++ b/python/orca/src/bigdl/orca/learn/utils.py
@@ -358,7 +358,8 @@ def process_xshards_of_pandas_dataframe(data, feature_cols, label_cols=None, val
         return data
 
 
-def _dataframe_to_xshards(data, feature_cols, label_cols=None, accept_str_col=False, shard_size=None):
+def _dataframe_to_xshards(data, feature_cols, label_cols=None,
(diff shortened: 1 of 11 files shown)


qiuxin2012 (approved):

python/orca/src/bigdl/orca/learn/utils.py
```diff
@@ -358,10 +358,12 @@ def process_xshards_of_pandas_dataframe(data, feature_cols, label_cols=None, val
...
 
 
-def _dataframe_to_xshards(data, feature_cols, label_cols=None, accept_str_col=False):
+def _dataframe_to_xshards(data, feature_cols, label_cols=None,
+                          accept_str_col=False, shard_size=None):
     from bigdl.orca import OrcaContext
     schema = data.schema
-    shard_size = OrcaContext._shard_size
+    if OrcaContext._shard_size:
+        shard_size = OrcaContext._shard_size
```
qiuxin2012, line 366: why OrcaContext._shard_size has a higher priority than current parameter.

PR #6800, opened 2022-11-29
[WIP] Add files needed by distributed encrypted file system (gramine)

gc-fu:
### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

### 5. New dependencies

- [ ] New Python dependencies
       - Dependency1 
       - Dependency2
       - ...
       - Dependency1 and license1
       - Dependency2 and license2
       - ...

### DIFF
--- a/ppml/base/bash.manifest.template
+++ b/ppml/base/bash.manifest.template
@@ -65,9 +65,9 @@ fs.mounts = [
   { path = "/root/.zinc", uri = "file:/root/.zinc" },
   { path = "/root/.cache", uri = "file:/root/.cache" },
   { path = "/usr/lib/gcc", uri = "file:/usr/lib/gcc" },
-  #{ path = "/ppml/trusted-big-data-ml", uri = "file:/ppml/trusted-big-data-ml" },
   { path = "/ppml", uri = "file:/ppml" },
   { type = "encrypted", path = "/ppml/encrypted", uri = "file:/ppml/encrypted", key_name = "_sgx_mrsigner" },
+  { type = "encryptedfsd", path = "/ppml/encryptedfsd", uri = "file:/ppml/encryptedfsd", key_name = "default" },
 ]
 #  { path = "{{ gramine.runtimedir() }}/etc/localtime", uri = "file:/etc" },
 
--- a/ppml/base/encryptedfsd.sh
+++ b/ppml/base/encryptedfsd.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+
+set -x
+
+# Check if the user set EHSM_URL or not
+if [ -z "$EHSM_URL" ]; then
+	echo "[ERROR] Distributed encrypted filesystem is enabled, but EHSM_URL is empty!"
+	echo "[INFO] PPML Application Exit!"
+	exit 1
+fi
+
+if [ -z "$APP_ID" ]; then
+    echo "[ERROR] Distributed encrypted filesystem is enabled, but APP_ID is empty!"
+    echo "[INFO] PPML Application Exit!"
+    exit 1
+fi
+
+if [ -z "$API_KEY" ]; then
+    echo "[ERROR] Distributed encrypted filesystem is enabled, but API_KEY is empty!"
+    echo "[INFO] PPML Application Exit!"
+    exit 1
+fi
+
+if [ -z "$EHSM_PORT" ]; then
+	echo "[INFO] Distributed encrypted filesystem is enabled, but EHSM_PORT is empty!"
+	echo "[INFO] Assume EHSM_PORT is 3000"
+	EHSM_PORT=3000
+fi
+
+# We need to check if encrypted_primary_key and encrypted_data_key exists or not
+if [ -z "$EHSM_PRIMARY_KEY" ]; then
+	echo "[INFO] Distributed encrypted filesystem is enabled, but EHSM_PRIMARY_KEY is empty!"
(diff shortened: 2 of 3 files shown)


Hanyu-Jin (approved):

Uxito-Ada:
Does **Data Encrypter** serve as a client?

gc-fu:
> Does **Data Encrypter** serve as a client?

Data Encrypter uses the same image as the BigDL Worker.  The reason why it is called **Encrypter** is that it is responsible for encrypting the data.

All the Encrypter/Worker uses the KMS client and request the same plaintext data key from KMS using the same encrypted data key and encrypted primary key.

Then, they will write the plaintext key into `/dev/attestation/keys/default`.  Because all the gramine instance writes the same value into `/dev/attestation/keys/default`, they are able to decrypt the contents encrypted on other nodes (using key `/dev/attestation/keys/default`).

Uxito-Ada:
Is Encrypter a docker or k8s pod? 
Where are primary key and data key generated? If Encrypter is a docker (like a client), we can use it as workflow entry to generate keys.
Otherwise, if it is an auto k8s pod, is it safe to pass raw plaintext data to the untrusted k8s cluster?

gc-fu:
> er is a docker (like a client), we can use it as workflow entry to generate keys.
> Otherwise, if it is an auto k8s pod, is it safe to pass raw pl

It can be either Docker or k8s pod.  I assume the user have already generated the encrypted_primary_key and encrypted_data_key.  These keys are provided into the docker container/k8s pod in some way.  The Gramine instance will try to use these keys and APIKEY APPID to contact the KMS service and request the plaintext key in SGX mode.  The plaintext key is later written into /dev/attestation/keys/default which is a virtual file system provided by Gramine also in SGX mode.  It will not be written into disk, thus ensuring the security.

As for the `USING_LOCAL_DKEY`, yes, it is not safe and the user must provide the plaintext key into the docker container or the k8s pod.

Uxito-Ada:
how about the data?
how does plaintext data go into Encrypter?

gc-fu:
> how about the data? how does plaintext data go into Encrypter?

The data will be encrypted in a environment that is trusted by the client.  I will modify the architecture graph according to your suggestions, thanks

ppml/base/bash.manifest.template
```diff
@@ -65,9 +65,9 @@ fs.mounts = [
   { path = "/root/.zinc", uri = "file:/root/.zinc" },
   { path = "/root/.cache", uri = "file:/root/.cache" },
   { path = "/usr/lib/gcc", uri = "file:/usr/lib/gcc" },
-  #{ path = "/ppml/trusted-big-data-ml", uri = "file:/ppml/trusted-big-data-ml" },
   { path = "/ppml", uri = "file:/ppml" },
-  { type = "encrypted", path = "/ppml/encrypted", uri = "file:/ppml/encrypted", key_name = "_sgx_mrsigner" },
+  { type = "encrypted", path = "/ppml/encrypted-fs", uri = "file:/ppml/encrypted-fs", key_name = "_sgx_mrsigner" },
+  { type = "encrypted", path = "/ppml/encrypted-fsd", uri = "file:/ppml/encrypted-fsd", key_name = "default" },
```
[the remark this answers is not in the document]
gc-fu: Ok

ppml/base/encryptedfsd.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
```
glorysdj, line 1: change this script to encrypted-fsd.sh ?

ppml/base/encryptedfsd.sh
```diff
@@ -0,0 +1,71 @@
...
+
+set -x
+
+if [  "$USING_LOCAL_DKEY" == "true" ]; then
```
glorysdj, line 5: USING_LOCAL_DKEY -> USING_LOCAL_DATA_KEY

ppml/base/encryptedfsd.sh
```diff
@@ -0,0 +1,71 @@
...
+    echo "[INFO] Using local key"
+    if [ -z "$LOCAL_DKEY" ]; then
```
glorysdj, line 7: LOCAL_DKEY -> LOCAL_DATA_KEY

ppml/trusted-deep-learning/base/entrypoint.sh
```diff
@@ -63,14 +63,28 @@ runtime_command="$@"
 
 if [ "$SGX_ENABLED" == "true" ]; then
   if [ "$ATTESTATION" ==  "true" ]; then 
+    rm /ppml/temp_command_file || true
     bash attestation.sh
+    if [ "$ENCRYPTEDFSD" == "true" ]; then
```
glorysdj, line 68: ENCRYPTEDFSD -> ENCRYPTED_FSD

ppml/base/encryptedfsd.sh
gc-fu: Fixed

ppml/base/encryptedfsd.sh
gc-fu: Fixed

ppml/trusted-deep-learning/base/entrypoint.sh
gc-fu: Fixed

PR #6802, opened 2022-11-29
Nano: add thread ctrl in context manager

TheaperDeng:
### 1. Why the change?

We have found that thread control could be a tricky setting for users, especially in nano. Since torch's thread num will even affect the performance of onnxruntime and openvino.

### 2. User API changes

nothing, follow the previous context manager usage。
Will add more details once this PR is ready for review

### 3. Summary of the change 
add thread_num in internal APIs.

### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/deps/neural_compressor/pytorch/quantized_model.py
+++ b/python/nano/src/bigdl/nano/deps/neural_compressor/pytorch/quantized_model.py
@@ -57,7 +57,10 @@ class PytorchQuantizedModel(AcceleratedLightningModule):
             with open(tune_cfg_file, 'r') as f:
                 tune_cfg = yaml.safe_load(f)
                 qmodel.tune_cfg = tune_cfg
-        return PytorchQuantizedModel(qmodel, thread_num=status["thread_num"])
+        thread_num = None
+        if status["thread_num"] is not None:
+            thread_num = int(status["thread_num"])
+        return PytorchQuantizedModel(qmodel, thread_num=thread_num)
 
     def _save_model(self, path):
         self.quantized.save(path)
--- a/python/nano/src/bigdl/nano/deps/neural_compressor/pytorch/quantized_model.py
+++ b/python/nano/src/bigdl/nano/deps/neural_compressor/pytorch/quantized_model.py
@@ -58,7 +58,7 @@ class PytorchQuantizedModel(AcceleratedLightningModule):
                 tune_cfg = yaml.safe_load(f)
                 qmodel.tune_cfg = tune_cfg
         thread_num = None
-        if status["thread_num"] is not None:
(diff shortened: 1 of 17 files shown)


python/nano/src/bigdl/nano/deps/neural_compressor/pytorch/quantized_model.py
```diff
@@ -47,7 +57,7 @@ def _load(path, model):
             with open(tune_cfg_file, 'r') as f:
                 tune_cfg = yaml.safe_load(f)
                 qmodel.tune_cfg = tune_cfg
-        return PytorchQuantizedModel(qmodel)
+        return PytorchQuantizedModel(qmodel, thread_num=status["thread_num"])
```
rnwang04, line 60: Do we need below code here ?
```
if status['thread_num'] is None:
            thread_num = None
else:
    thread_num = int(status['thread_num'])
```

python/nano/src/bigdl/nano/pytorch/context_manager.py
```diff
@@ -50,3 +57,17 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_value, exc_tb):
         self.autocast.__exit__(exc_type, exc_value, exc_tb)
         super().__exit__(exc_type, exc_value, exc_tb)
+
+
+def generate_context_manager(accelerator=None, precision="fp32", thread_num=None):
+    '''
+    generate correct context manager according to different situation
+    :param acclerator: str, the accelerator to use, we support "onnxruntime", "openvino"
```
rnwang04, line 65: Why accelerator here? It seems no effect in current code.

rnwang04 (approved):
others LGTM

python/nano/src/bigdl/nano/deps/neural_compressor/pytorch/quantized_model.py
```diff
@@ -20,21 +20,31 @@
...
+from bigdl.nano.pytorch.context_manager import generate_context_manager
 
 
 class PytorchQuantizedModel(AcceleratedLightningModule):
-    def __init__(self, model):
+    def __init__(self, model, thread_num=None):
         super().__init__(model.model)
         self.quantized = model
-        self.context_manager = BaseContextManager()
+        self.thread_num = thread_num
```
jason-dai, line 30: shall we make `thread_num` part of `AcceleratedLightningModule`?

TheaperDeng: Yeah that's definitely correct. We are having a messy inherit situation for inference classes. Some of the class does not call init of their mother class at all, which makes `AcceleratedLightningModule` only function as a dummy base class. I will record this under the refactor umbrella issue.

PR #6804, opened 2022-11-29
Nano: Change the API of keras optmizer and quantize

MeouSker77:
## Description

Change the API of keras `InferenceOptimizer` and `quantize`, now is similar to `keras.Model.fit`

### 1. Why the change?

`keras.Model.fit` use `x` and `y` as input dataset, where `x` is tensors or ndarrays of feature, `y` is tensors or ndarrays of label, it also accepts `x` as a `tf.data.Dataset` returning both feature and label. This PR changes current `calib_dataset` parameter to `x` and `y` to be consistent with `keras.Model.fit`

### 2. User API changes

Before:
```python
# quantize
from bigdl.nano.tf.keras import Model

model = Model()
model.quantize(accelerator=..., calib_dataset=train_dataset)

...

# optimize
from tensorflow.keras import Model

model = Model()
opt = InferenceOptimizer()
opt.optimize(model=model, calib_dataset=train_dataset)
```

Now:
```python
# quantize
from bigdl.nano.tf.keras import Model

model = Model()
model.quantize(accelerator=..., x=train_dataset)
# or
model.quantize(accelerator=..., x=train_features, y=train_labels)

...

# optimize
from tensorflow.keras import Model

model = Model()
opt = InferenceOptimizer()
opt.optimize(model=model, x=train_dataset)
# or
opt.optimize(model=model, x=train_features, y=train_labels)
```

Before `calib_dataset` must be a dataset which returns tuples of (feature, label)

Now `x` and `y` can be a numpy array or tensor, or `x` is a dataset which returns tuples of (feature, label), i. e.

        :param x: Input data which is used for training. It could be:
                  | 1. a Numpy array (or array-like), or a list of arrays (in case the model
                  | has multiple inputs).
                  |
                  | 2. a TensorFlow tensor, or a list of tensors (in case the model has
                  | multiple inputs).
                  |
                  | 3. an unbatched tf.data.dataset. Should return a tuple of (inputs, targets).

                  X will be used as calibration dataset for Post-Training Static Quantization (PTQ),
                  as well as be used for generating input_sample to calculate latency.
                  To avoid data leak during calibration, please use training dataset.
        :param y: Target data. Like the input data x, it could be either Numpy array(s) or
                  TensorFlow tensor(s). It should be consistent with x (you cannot have Numpy inputs
                  and tensor targets, or inversely). If x is a dataset, y will be ignored (since
                  targets will be obtained from x).

Besides, rename the parameter `input_sample` of `model.trace(...)` to `input_spec`

### 3. Summary of the change 

1. Change the API of keras optmizer and quantize: `calib_dataset` to `x` and `y`
2. Rename the parameter `input_sample` of `model.trace(...)` to `input_spec`
3. Infer input shape when tracing onnx model in `model.trace` instead of in `model.quantize` or `opt.optimize`
4. Better `tensors_to_numpy` function
5. Change and add all related UT

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/python/nano/src/bigdl/nano/tf/keras/inference_utils.py
+++ b/python/nano/src/bigdl/nano/tf/keras/inference_utils.py
@@ -57,15 +57,14 @@ class InferenceUtils:
                   | 2. a TensorFlow tensor, or a list of tensors (in case the model has
                   | multiple inputs).
                   |
-                  | 3. an unbatched tf.data.dataset. Should return a tuple of (inputs, targets).
+                  | 3. an unbatched tf.data.Dataset. Should return a tuple of (inputs, targets).
 
                   X will be used as calibration dataset for Post-Training Static Quantization (PTQ),
                   as well as be used for generating input_sample to calculate latency.
                   To avoid data leak during calibration, please use training dataset.
         :param y: Target data. Like the input data x, it could be either Numpy array(s) or
-                  TensorFlow tensor(s). It should be consistent with x (you cannot have Numpy inputs
-                  and tensor targets, or inversely). If x is a dataset, y will be ignored (since
-                  targets will be obtained from x).
+                  TensorFlow tensor(s). Its length should be consistent with x.
+                  If x is a dataset, y will be ignored (since targets will be obtained from x).
         :param precision:       Global precision of quantized model,
                                 supported type: 'int8', defaults to 'int8'.
         :param accelerator:     Use accelerator 'None', 'onnxruntime', 'openvino', defaults to None.
--- a/python/nano/src/bigdl/nano/tf/keras/inference_utils.py
+++ b/python/nano/src/bigdl/nano/tf/keras/inference_utils.py
@@ -13,21 +13,22 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-from typing import List
-from bigdl.nano.deps.neural_compressor.inc_api import quantize as inc_quantzie
+from typing import Optional, List, Union
+import numpy as np
 import tensorflow as tf
 from tensorflow.keras.metrics import Metric
+from bigdl.nano.deps.neural_compressor.inc_api import quantize as inc_quantzie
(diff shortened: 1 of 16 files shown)


jason-dai:
Do we still use `InferenceOptimizer.optimize/quantize/trace` for TF?

MeouSker77:
In fact, now we are using `InferenceOptimizer.optimize` and `model.quantize/trace`

jason-dai:
> In fact, now we are using `InferenceOptimizer.optimize` and `model.quantize/trace`

Why the change then? In particular, I don't think it's a good idea to provide two type of interfaces for the same type of concept.

MeouSker77:
> Why the change then? In particular, I don't think it's a good idea to provide two type of interfaces for the same type of concept.

We have been using `model.quantize/trace` before adding keras optimizer, we will replace it with `InferenceOptimizer.quantize/trace` in the future

python/nano/src/bigdl/nano/tf/keras/inference_utils.py
```diff
@@ -152,10 +173,7 @@ def quantize(self,
             if isinstance(self, KerasONNXRuntimeModel):     # type: ignore
                 onnx_model = self
             else:
-                spec = tf.TensorSpec((self.input_shape), self.dtype)    # type: ignore
-                onnx_model = self.trace(accelerator='onnxruntime',
-                                        input_sample=spec,
-                                        thread_num=thread_num)
+                onnx_model = self.trace(accelerator='onnxruntime', thread_num=thread_num)
```
TheaperDeng, line 176: why don't we need `input_spec` here?

MeouSker77: now it will automatically infer the `input_spec` in`trace`

TheaperDeng (approved):
LGTM

PR #6815, opened 2022-11-30
Chronos: scale using numpy to reduce latency

hjzin:
### 1. Why the change?

This PR is a complement for #6812.
### 3. Summary of the change 

Add numpy implementation for `MaxAbsScaler`, `MinMaxScaler()`, `RobustScaler()`
### 4. How to test?
- [ ] Unit test


python/chronos/src/bigdl/chronos/data/tsdataset.py
```diff
@@ -1041,6 +1041,24 @@ def scale(self, scaler, fit=True):
...
+               test set. The value is defaulted to True.
+        '''
+        col_list = self.target_col + self.feature_col
+        if fit:
+            scaler.fit(self.df[col_list])
+        scaled_data_numpy = scale_timeseries_numpy(self.df[col_list].values, scaler)
+        self.df[col_list] = pd.DataFrame(scaled_data_numpy, columns=col_list)
+        self.scaler = scaler
+        return self
+
```
TheaperDeng, line 1061: move this to `scale` method, we should directly call `scale_timeseries_numpy` for the cases where `fit=False`

TheaperDeng (approved):
LGTM

PR #6821, opened 2022-11-30
[WIP] Orca: remove fsspec global deps.

lalalapotter:
## Description

This PR is to remove fsspec global dependency to avoid using libhdfs.so.


lalalapotter:
Need to add UT for new file ops.

sgwhat (approved):

PR #6822, opened 2022-11-30
Add LazySparkXShards

hkvision:


python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -186,6 +186,21 @@ def __init__(self,
...
+    
+    @classmethod
+    def create(cls,
+               rdd: Union["PipelinedRDD", "RDD"],
+               class_name: str = None) -> "SparkXShards":
+        return cls(rdd, False, class_name)
+    
+    def lazy(self) -> "LazySparkXShards":
+        """
+        Making the current SparkXShards won't affect the current cached rdd.
```
jason-dai, line 198: Creating `LazySparkXShards` doe not change the existing behavior of the current `SparkXShards`; that is, it will still be uncached unless the user explicitly calls `cache()`

python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -1089,6 +1105,20 @@ def sample_to_pdf(self,
         return pdf
 
 
+class LazySparkXShards(SparkXShards):
+    def __init__(self,
+                 rdd: Union["PipelinedRDD", "RDD"],
+                 class_name: str = None) -> None:
+        super(LazySparkXShards, self).__init__(rdd, True, class_name)
+        self.user_cached = True
```
jason-dai, line 1113: Change it to `user_caching`?

And you need to set `self.eager` to false

hkvision: Seems in our logic if transient is True, then eager is always False.

python/orca/src/bigdl/orca/data/shard.py
hkvision: should be `that is, it will still be cached unless the user explicitly calls uncache()`??

Updated the doc, please take a further look.

python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -186,6 +186,21 @@ def __init__(self,
         self.type = {}
         if class_name:
             self.type['class_name'] = class_name
+    
+    @classmethod
+    def create(cls,
+               rdd: Union["PipelinedRDD", "RDD"],
+               class_name: str = None) -> "SparkXShards":
```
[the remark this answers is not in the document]
hkvision: Probably we can't do this? As this create will be used by both SparkXShards and LazySparkXShards.
Make this function as private to _create and add doc to clarify this.
_create for SparkXShards will always create a non-transient SparkXShards, which is always the behavior for the existing implementations.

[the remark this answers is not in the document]
jason-dai: > Probably we can't do this? As this create will be used by both SparkXShards and LazySparkXShards. Make this function as private to _create and add doc to clarify this. _create for SparkXShards will always create a non-transient SparkXShards, which is always the behavior for the existing implementations.

I think it's probably OK for now, if only `transform_shard` and `repartition` are used inside Estimator.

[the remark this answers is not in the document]
hkvision: Yes, at this moment only these two methods may be invoked in the Estimator.

hkvision:
Merge it first. Will further modify this in the PR for estimator.

hkvision:
Remark here:
- For SparkXShards input from users to Estimator (usually should be SparkXShards of pandas dataframe), if call repartition in pyspark backend, can't use coalesce since coalesce don't work with spark barrier mode.
```
py4j.protocol.Py4JJavaError: An error occurred while calling z:org.apache.spark.api.python.PythonRDD.collectAndServe.
: org.apache.spark.scheduler.BarrierJobUnsupportedRDDChainException: [SPARK-24820][SPARK-24821]: Barrier execution mode does not allow the following pattern of RDD chain within a barrier stage:
1. Ancestor RDDs that have different number of partitions from the resulting RDD (e.g. union()/coalesce()/first()/take()/PartitionPruningRDD). A workaround for first()/take() can be barrierRdd.collect().head (scala) or barrierRdd.collect()[0] (python).
2. An RDD that depends on multiple barrier RDDs (e.g. barrierRdd1.zip(barrierRdd2)).
```

Issue #6760, opened 2022-11-25
bigdl-nano auto-ml model select method


Zjq9409:
Bigdl-nano Auto_Ml selects the model based on the last epoch loss, needs to be modified to select the one with the highest accuracy. for example, The Actual output Best trial Value is 0.4935064911842346 (the last epoch), but the model needs to be selected according to the best epoch.

```
Epoch 4:  69%|██████▉   | 9/13 [00:34<00:15,  3.86s/it, loss=1.5, v_num=7, val_loss=1.380, val_acc=0.506, train_loss=1.350, train_acc=0.510] 
Epoch 4:  77%|███████▋  | 10/13 [00:36<00:10,  3.66s/it, loss=1.5, v_num=7, val_loss=1.380, val_acc=0.506, train_loss=1.350, train_acc=0.510]
Epoch 4:  77%|███████▋  | 10/13 [00:36<00:10,  3.66s/it, loss=1.5, v_num=7, val_loss=1.380, val_acc=0.506, train_loss=1.350, train_acc=0.510]
Epoch 4:  77%|███████▋  | 10/13 [00:36<00:10,  3.66s/it, loss=1.29, v_num=7, val_loss=1.380, val_acc=0.506, train_loss=1.350, train_acc=0.510]

Validation: 0it [00:00, ?it/s][A

Validation:   0%|          | 0/3 [00:00<?, ?it/s][A

Validation DataLoader 0:   0%|          | 0/3 [00:00<?, ?it/s][A

Validation DataLoader 0:  33%|███▎      | 1/3 [00:00<00:01,  1.10it/s][A

Validation DataLoader 0:  33%|███▎      | 1/3 [00:00<00:01,  1.10it/s][A

Validation DataLoader 0:  67%|██████▋   | 2/3 [00:01<00:00,  1.18it/s][A

Validation DataLoader 0:  67%|██████▋   | 2/3 [00:01<00:00,  1.18it/s][A

Validation DataLoader 0: 100%|██████████| 3/3 [00:02<00:00,  1.54it/s][A

Validation DataLoader 0: 100%|██████████| 3/3 [00:02<00:00,  1.54it/s][A

                                                                      [A
Epoch 4:  77%|███████▋  | 10/13 [00:39<00:11,  3.90s/it, loss=1.29, v_num=7, val_loss=5.030, val_acc=0.494, train_loss=1.300, train_acc=0.540]
[...]

PR #6826, opened 2022-12-01
Nano: Support best result for each trial in AutoML

rnwang04:
## Description

Now NanoHPO selects model based on the last epoch's metric (for example, last epoch's val loss). However, it's more reasonable to select the one with the best metric.

### 1. Why the change?

https://github.com/intel-analytics/BigDL/issues/6760

### 2. User API changes

New parameter `mode` for `trainer.search`:
```python
model = CustomModel(
    out_dim = space.Categorical(16,32),
    dropout = space.Categorical(0.1,0.2),
    learning_rate = space.Real(0.001,0.01,log=True),
    batch_size = space.Categorical(32,64)
)

trainer = Trainer(
    logger=True,
    checkpoint_callback=False,
    max_epochs=2,
    use_hpo=True,
)

best_model = trainer.search(
    model,
    target_metric='val_loss',
    direction='minimize',
    n_trials=4,
    max_epochs=3,
    mode='best',  # default value is best, you can change to last.
)
```

### 3. Summary of the change 

- support obtain the best metric value for each trial
- provide two modes in `trainer.search`
- related ut

### 4. How to test?
- [x] Unit test
- [x] Application test

### DIFF
--- a/python/nano/src/bigdl/nano/automl/pytorch/_helper.py
+++ b/python/nano/src/bigdl/nano/automl/pytorch/_helper.py
@@ -75,6 +75,38 @@ class LatencyCallback(Callback):
         pl_module.time_avg(batch_latency)
 
 
+def is_better(score, current_score, direction):
+    if current_score is None:
+        return True
+    elif direction == "min":
+        if score < current_score:
+            return True
+        return False
+    elif direction == "max":
+        if score < current_score:
+            return False
+        return True
+
+
+class BestMetricCallback(Callback):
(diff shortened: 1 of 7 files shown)


rnwang04:
> others LGTM. Just one question, so are we now still using the checkpoint of the last epoch or we also switch to the checkpoint of the best epoch

Actually we don't save any checkpoint during AutoML, and if you mean the behavior of checkpoint in fitting, it depends on user themselves. Or you mean in chronos?

PR #6833, opened 2022-12-01
Nano: support getattr from original model

rnwang04:
## Description

During OOB, we found that model's attributes will be losing after applying `InferenceOptimizer.trace` / `InferenceOptimizer.quantize`, as model becomes an attribute of AcceleratedModel, and we can't get original model's attribute directly.

### 1. Why the change?

https://github.com/analytics-zoo/nano/issues/228

### 2. User API changes

No change. But now, after `InferenceOptimizer.trace` / `InferenceOptimizer.quantize`, user can still use original attributes / methods, like following:
```python
model = resnet18(num_classes=10)
#  patch a attr
model.channels = 3
def hello():
    print("hello world!")
# patch a function
model.hello = hello

new_model = InferenceOptimizer.trace(model, precision='bf16',
                                     accelerator="jit", use_ipex=True,
                                     input_sample=x)
    new_model(self.data_sample)
assert new_model.channels == 3
new_model.hello()
```

### 3. Summary of the change 

- support getattr from original model for BF16Model
- support getattr from original model for PytorchIPEXJITModel
- add related ut for above features

### 4. How to test?
- [x] Unit test
- [x] Application test

### DIFF
--- a/python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
+++ b/python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
@@ -17,6 +17,7 @@
 from bigdl.nano.utils.inference.pytorch.model import AcceleratedLightningModule
 from bigdl.nano.pytorch.context_manager import generate_context_manager
 import torch
+from copy import deepcopy
 
 
 class PytorchIPEXJITModel(AcceleratedLightningModule):
@@ -56,7 +57,7 @@ class PytorchIPEXJITModel(AcceleratedLightningModule):
                                                                   thread_num=thread_num)
             return
         self.channels_last = channels_last
-        self.original_model = model
+        self.original_model = model if inplace is False else deepcopy(model)
         self.original_state_dict = model.state_dict()
         self.use_ipex = use_ipex
         self.use_jit = use_jit
@@ -85,6 +86,11 @@ class PytorchIPEXJITModel(AcceleratedLightningModule):
                                                               precision="fp32",
                                                               thread_num=thread_num)
         self.thread_num = thread_num
+        # patch original model's attr to current new model
+        for attr in dir(self.original_model):
+            if attr not in dir(self) and not attr.startswith('_'):
+                setattr(self, attr, getattr(self.original_model, attr))
+        del self.original_model  # save memory
 
     @property
     def forward_args(self):
@@ -101,13 +107,6 @@ class PytorchIPEXJITModel(AcceleratedLightningModule):
 
     def on_forward_end(self, outputs):
         return outputs
-
-    def __getattr__(self, name: str):
-        # automatically unwrap attributes access of model
-        try:
-            return super().__getattr__(name)
-        except AttributeError:
-            return getattr(self.original_model, name)
 
     @property
     def status(self):
--- a/python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
+++ b/python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
@@ -102,6 +102,17 @@ class PytorchIPEXJITModel(AcceleratedLightningModule):
     def on_forward_end(self, outputs):
         return outputs
 
+    def __getattr__(self, name: str):
+        # automatically unwrap attributes access of model
+        print(name)
+        try:
+            return super().__getattr__(name)
+        except:
+            try:
+                return getattr(self.model, name)
+            except AttributeError:
+                pass
+
     @property
     def status(self):
         status = super().status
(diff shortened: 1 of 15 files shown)


python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
```diff
@@ -51,6 +50,7 @@ def __init__(self, model: torch.nn.Module, input_sample=None, use_ipex=False, dt
             self.context_manager = BaseContextManager()
             return
         self.channels_last = channels_last
+        self.original_model = model
```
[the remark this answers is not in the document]
rnwang04: 1. Considering memory consumption, how about monkey patch original attributes in current new code?
2. Will support more cases later.

python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
```diff
@@ -81,6 +83,13 @@ def __init__(self, model: torch.nn.Module, input_sample=None, use_ipex=False, dt
                                                         precision="fp32",
                                                         thread_num=thread_num)
         self.thread_num = thread_num
+        if (self.use_ipex is True or self.use_jit is True) and inplace is False:
+            # patch original model's attr to current new model
+            for attr in dir(self.original_model):
+                if attr not in dir(self) and not attr.startswith('_') and not\
+                        isinstance(getattr(self.original_model, attr), torch.nn.Module):
+                    setattr(self, attr, getattr(self.original_model, attr))
+            del self.original_model  # save memory
```
MeouSker77, line 92: This may not save memory.

If we write:
```python
class A():
    def m(self):
        ...
a = A()
b = getattr(a, "m")
del a
```

Then b will contains object `a` implicitly, because the first parameter of `m` is `self`. `del a` will only delete a reference to object `a`.

rnwang04: Actually, this way has almost the same memory consumption as before, and after our discussion, `self.original_model = model` is just a reference, don't cause too much additional overhead.

I think current implementation (monkey patch) might be an acceptable choice, as if just use `self.original_model = model`, quantization will cause error as some behavior of `nn.Module`.

@MeouSker77 @TheaperDeng

python/nano/src/bigdl/nano/deps/openvino/pytorch/model.py
```diff
@@ -46,17 +46,23 @@ def __init__(self, model, input_sample=None, thread_num=None,
...
             super().__init__(None)
         self.context_manager = generate_context_manager(accelerator="openvino",
                                                         precision="fp32",
                                                         thread_num=thread_num)
+        # patch original model's attr to current new model
+        if isinstance(model, torch.nn.Module):
+            for attr in dir(model):
+                if attr not in dir(self) and not attr.startswith('_') and not\
+                        isinstance(getattr(model, attr), torch.nn.Module):
+                    setattr(self, attr, getattr(model, attr))
```
jason-dai, line 65: Refactor this to a utility method that everyone can call (instead of writing the same code multiple times).

Why do you skip all the attributes starts with '_'? A regular method may depend on these.

And we will save another copy of the original weight? (Maybe it's just a reference and does not matter?)

rnwang04: will revise according to first two points.
and I don't quite understand the third point, another copy of original weight is ?

jason-dai: > will revise according to first two points. and I don't quite understand the third point, another copy of original weight is ?

Both the weight of the new model (e.g., OV or ORT) and the original FP32 PyTorch model?

rnwang04: 1. for `Why do you skip all the attributes starts with '_'? A regular method may depend on these.`
After validation, actually we can just skip them, as these private attributes can be obtained implicitly through the original model, below is a case:
```
class A:
    def __init__(self):
        self._x = 1
        self.channels = 3
    
    def print(self):
        print(self)
        print(self._x + self.channels)


class B:
    def __init__(self, a):
        self.y = 1
        self.channels = a.channels
        self.print = a.print

a = A()
b = B(a)
a.print()
# <__main__.A object at 0x7f5766fa6c10>
# 4
b.print()
# <__main__.A object at 0x7f5766fa6c10>
# 4
```
2. about `save another copy of the original weight`
In fact, this PR don't support to get original attributes after save & load.
If we want to support this, it's not enough to just save model's weight, instead, we may need to save the whole model and then we can load attributes like `channels`. I feel this is a little complex, and the saved files will be larger. Do you think shall we support this ? @jason-dai @TheaperDeng

MeouSker77 (approved):
LGTM

PR #6839, opened 2022-12-01
Orca: Add tensorboard call_back to NCF xshards example.

zpeng1898:


python/orca/tutorial/NCF/pytorch_train_xshards.py
```diff
@@ -59,6 +61,8 @@ def optimizer_creator(model, config):
 
 # Step 4: Distributed training with Orca PyTorch Estimator
 backend = "spark"  # "ray" or "spark"
+callbacks = [TensorBoardCallback(log_dir=os.path.join("runs_epoch"),
+                                 freq="batch")]
```
hkvision, line 65: remove os, change to `runs` and put it in one line

PR #6843, opened 2022-12-01
Nano: Add `inplace` parameter to support model which cannot be copied

MeouSker77:
## Description

Add `inplace` parameter for `InferenceOptimizer.quantize`, `InferenceOptimizer.trace` and `InferenceOptimizer.load` to support models which cannot be copied

### 1. Why the change?

Some models cannot be copied, while IPEX will call `copy.deepcopy` if we don't pass `inpalce=False`, and now `InferenceOptimizer.optimize` will call `copy.deepcopy`, too

### 2. User API changes

`InferenceOptimizer.quantize`, `InferenceOptimizer.trace` and `InferenceOptimizer.load` now have an optional `inplace` parameter. If `inplace=True`, these functions won't call `copy.deepcopy` when tracing ipex model or quantizing int8 model.

### 3. Summary of the change 

- New added `inplace` parameter will be passed to IPEX
- int8 quantize now deepcopy model by default
- other quantize now won't deepcopy model

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
+++ b/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
@@ -17,7 +17,6 @@
 import torch
 from torch import nn
 import time
-from copy import deepcopy
 import multiprocessing as mp
 from typing import Dict, Callable, Tuple, Optional, List, Union, Sequence
 from torch.utils.data import DataLoader
@@ -640,9 +639,6 @@ class InferenceOptimizer(BaseInferenceOptimizer):
                 model which is able to run on Pytorch or ONNXRuntime can be fetched by
                 `quantized_model.model`.
                 """
-                # deepcopy model when using int8 quantize and inplace=False
-                if accelerator is None and not inplace:
-                    model = deepcopy(model)
                 return inc_quantize(model, inc_calib_dataloader, metric,
                                     thread_num=thread_num,
                                     framework=framework,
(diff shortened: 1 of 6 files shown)


python/nano/test/pytorch/tests/test_inference_optimizer.py
```diff
@@ -464,3 +465,25 @@ def test_context_manager(self):
...
+                invalidOperationError(False, "The `deepcopy` function shouldn't be called")
+
+        model = CannotCopyNet()
+        inference_opt = InferenceOptimizer()
+        # int8
+        int8_model = inference_opt.quantize(model, calib_data=self.train_loader, inplace=True)
+        # bf16+ipex
+        bf16_ipex_model = inference_opt.quantize(model, calib_data=self.train_loader, precision='bf16', use_ipex=True, inplace=True)
+        # ipex
+        ipex_model = inference_opt.trace(model, input_sample=self.train_loader, use_ipex=True, inplace=True)
```
TheaperDeng, line 481: could we test like this? I mean we state `inplace=True` for all cases and the model is now very messy which is not a  good choice for cascade calling of `quantize`/`trace` right?

MeouSker77: fix it

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -637,6 +639,9 @@ def quantize(model: nn.Module,
                 model which is able to run on Pytorch or ONNXRuntime can be fetched by
                 `quantized_model.model`.
                 """
+                # deepcopy model when using int8 quantize and inplace=False
```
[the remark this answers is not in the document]
MeouSker77: done

rnwang04 (approved):
LGTM

TheaperDeng (approved):

PR #6845, opened 2022-12-01
Chronos: nightly test different install options on python3.8

plusbang:
## Description

Chronos PR validation is only tested on python3.7. To support python3.8, we need nightly test.

### 2. User API changes

nothing

### 3. Summary of the change 

add trigger
limit numpy version to solve AttributeError

### 4. How to test?
- [x] Unit test


liu-shaojun (approved):

TheaperDeng (approved):
LGTM

PR #6849, opened 2022-12-01
Image shards

songhappy:
## Description
Add methods to support distributed images on Orca Estimator. All the transformations of PIL.Image.Image should be used directly. 

### 1. User API changes
Reading images using spark 
```python
import bigdl.orca.data.image
data_shard = bigdl.orca.data.image.read_images_spark(path)
```
Reading images using PIL 
```python
import bigdl.orca.data.image
data_shard = bigdl.orca.data.image.read_images_pil(path)
```
### 2. Summary of the change 

1. Two methods of read_images_spark and read_images_pil are added to read images from local or hdfs to a SparkXShard. Each record of the  SparkXShard is a  PIL.Image.Image; but if the label function or target_path is provided, each record of the SparkXShard is a tuple of PIL.Image.Image and label, labels are also a PIL.Image.Image if target_path.
2. Add test cases of different scenarios, including with or without label functions or target path; and transformations on  PIL.Image.Image.
3. Add the method of stack_feature_labels in shard.py to assemble features and labels. 

### 3. How to test?
- [ ] Unit test


jason-dai:
Please update the description, and add examples/unit tests to show how it is used.

jason-dai:
> ### 1. User API changes
> Reading images using spark
> 
> ```python
> import bigdl.orca.data.image
> data_shard = bigdl.orca.data.image.read_images_spark(path)
> ```
> 
> Reading images using PIL
> 
> ```python
> import bigdl.orca.data.image
> data_shard = bigdl.orca.data.image.read_images_pil(path)
> ```
> 
 
Change to `bigdl.orca.data.read_images`

What's the difference between the two implementations? I don't think the user need to specify which version to user - it's not very user friendly.

PR #6850, opened 2022-12-02
Nano: Improve context manager in InferenceOptimizer

rnwang04:
## Description

Further improvement of context manager in InferenceOptimizer

### 1. Why the change?

https://github.com/intel-analytics/BigDL/pull/6747
https://github.com/analytics-zoo/nano/issues/210

### 2. User API changes

No change. But now `get_context()` method can be used for multi model case.
```python
input_sample = torch.rand(10, 3, 32, 32)
                       training_data=self.train_loader)

ipex_model = inference_opt.get_model("jit_fp32_ipex")
with InferenceOptimizer.get_context(model, ipex_model):
    ipex_model(input_sample)
```

### 3. Summary of the change 

- make context_manager as a private attribute to avoid potential conflict
- support thread_num for BF16Model
- make get_context() work for multi model case
- related ut for multi-model case
- update some doc
- update API for performance regression test

### 4. How to test?
- [x] Unit test
- [x] Application test

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/amp/bfloat16.py
+++ b/python/nano/src/bigdl/nano/pytorch/amp/bfloat16.py
@@ -168,7 +168,7 @@ class BF16Model(AcceleratedLightningModule):
         model.eval()
         model.load_state_dict(state_dict)
         thread_num = None
-        if status["thread_num"] is not None:
+        if status["thread_num"] is not None and status['thread_num'] != {}:
             thread_num = int(status['thread_num'])
         return BF16Model(model, channels_last=status['channels_last'],
                          thread_num=thread_num)
--- a/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
+++ b/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
@@ -785,13 +785,13 @@ class InferenceOptimizer(BaseInferenceOptimizer):
         invalidInputError(False, "Accelerator {} is invalid.".format(accelerator))
 
     @staticmethod
-    def get_context(model: nn.Module, *args):
+    def get_context(model: nn.Module, *models):
         """
         Obtain corresponding context manager from (multi) model, defaults to BaseContextManager().
 
         :param model: Any model of torch.nn.Module, including all models accelareted by
                InferenceOptimizer.trace/InferenceOptimizer.quantize.
-        :param args: Any model of torch.nn.Module or list of torch.nn.Module, including all models
+        :param models: Any model of torch.nn.Module or list of torch.nn.Module, including all models
                accelareted by InferenceOptimizer.trace/InferenceOptimizer.quantize.
         :return: a context manager if there is no conflict between context managers,
                  otherwise will report RuntimeError.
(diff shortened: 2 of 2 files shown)


python/nano/src/bigdl/nano/pytorch/amp/bfloat16.py
```diff
@@ -160,7 +167,11 @@ def _load(path, model):
         state_dict = torch.load(checkpoint_path)
         model.eval()
         model.load_state_dict(state_dict)
-        return BF16Model(model, channels_last=status['channels_last'])
+        thread_num = None
+        if status["thread_num"] is not None:
```
TheaperDeng, line 171: apologize for not revising this file :(

We need
```python
if status["thread_num"] is not None and status['thread_num'] != {}:
```
since `yaml` will parse a `{}` rather than `None` for `null` in yml file.

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -772,17 +775,57 @@ def trace(model: nn.Module,
         invalidInputError(False, "Accelerator {} is invalid.".format(accelerator))
 
     @staticmethod
-    def get_context(model: nn.Module):
+    def get_context(model: nn.Module, *args):
```
TheaperDeng, line 778: should we just call it `*models`? No strong idea about this.

rnwang04: I think *models might be a good chioce. Do you think we should use 
`get_context(model: nn.Module, *models)`
or just
`get_context(*models)`
as input paramater?

python/nano/src/bigdl/nano/pytorch/amp/bfloat16.py
rnwang04: Done

TheaperDeng (approved):
LGTM

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -782,17 +785,57 @@ def trace(model: nn.Module,
...
+                if thread_num1 is None or thread_num2 is None:
+                    warnings.warn("One of the two models has thread control and the other "
+                                  "does not. The returned context manager will be dominated "
+                                  "by the non-None one.")
+                    if thread_num1 is None:
+                        return new_manager
+                    else:
+                        return manager
+                else:
+                    invalidInputError(False, "Can't obtain a new context manager for model"
```
jason-dai, line 825: Do we want to return error here? We can always return a "common denominator" (e.g., no_grad and autocast with proper thread_num) to the users to run their code.

I think it's better to report warning instead of throwing error to block the user.

rnwang04: In this case, two context managers have different thread number, what should be the proper thread_num?
Or we just report warning and set thread_num=None?

rnwang04: will change to this in next PR.

PR #6858, opened 2022-12-02
Nano: add new parameter for jit's strict mode

TheaperDeng:
### 1. Why the change?
We are adding more **optional** parameters for users to resolve the error they meet.
@Litchilitchy 

### 2. User API changes
```python
InferenceOptimizer.trace(accelerator="jit",...,  jit_strict=False)
InferenceOptimizer.quantize(accelerator="jit",..., jit_strict=False)
```

### 3. Summary of the change 
Add a parameter that will be passed to `torch.jit.trace`
https://pytorch.org/docs/stable/generated/torch.jit.trace.html

### 4. How to test?
- [ ] Unit test


rnwang04 (approved):
LGTM

Litchilitchy (approved):

PR #6860, opened 2022-12-02
Nano: Improve multi-instance inference

MeouSker77:
## Description

- Add `cores_for_each_process` parameter to specify the number of cores used by each process
- Add `cpu_for_each_process` parameter to specify the cores used by each process
- Support each process retrieve tasks dynamically instead of static assignment

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/inference/multi_instance.py
+++ b/python/nano/src/bigdl/nano/pytorch/inference/multi_instance.py
@@ -24,13 +24,13 @@ from bigdl.nano.utils.log4Error import invalidInputError, invalidOperationError
 
 
 class _MultiInstanceModel(torch.nn.Module):
-    def __init__(self, model, ps, send_queues, recv_queues):
+    def __init__(self, model, ps, send_queue, recv_queue):
         super().__init__()
         self.model = model
         self.ps = ps
         self.p_num = len(ps)
-        self.send_queues = send_queues
-        self.recv_queues = recv_queues
+        self.send_queue = send_queue
+        self.recv_queue = recv_queue
 
     def forward(self, input_data: Union[DataLoader, List]) -> List:
         if isinstance(input_data, (DataLoader, list)):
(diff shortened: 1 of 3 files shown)


rnwang04 (approved):
LGTM

python/nano/test/pytorch/tests/test_inference_optimizer.py
```diff
@@ -382,6 +382,25 @@ def test_multi_instance(self):
             np.testing.assert_allclose(pred1, pred2, atol=1e-4,
                                         err_msg=f"\npred1: {pred1}\npred2: {pred2}\n")
 
+    def test_multi_instance_with_specify_cores(self):
+        model = Net()
+        model.eval()
+        inference_opt = InferenceOptimizer()
+        multi_instance_model = inference_opt.to_multi_instance(model, num_processes=2,
+                                                               cpu_for_each_process=[[0], [1]])
```
jason-dai, line 390: I think you should allow the user to simply specify `cores_per_process`, instead of a map of cores for each process

MeouSker77: done

python/nano/src/bigdl/nano/pytorch/inference/multi_instance.py
```diff
@@ -24,40 +24,42 @@
...
-            output = self.recv_queues[idx % self.p_num].get()
-            invalidOperationError(not isinstance(output, Exception), f"{output}")
-            outputs.append(output)
-        return outputs
+        for _i in range(length):
+            idx_and_output = self.recv_queue.get()
+            invalidOperationError(not isinstance(idx_and_output, Exception), f"{idx_and_output}")
+            outputs.append(idx_and_output)
+
+        outputs = sorted(outputs, key=lambda x: x[0])
```
jason-dai, line 50: do you need to sort here? You may store the result using an array directly based on the index, yes?

MeouSker77: changed it in #6893

PR #6862, opened 2022-12-02
Chronos: update numpy and SQLAlchemy version

plusbang:
## Description

During installation test on python3.8, it is found that the version of SQLAlchemy influence tuning with multi-processes.
And the latest version of numpy cause AttributeError.

### 2. User API changes

nothing

### 3. Summary of the change 

update numpy and SQLAlchemy version

### 4. How to test?
- [x] Unit test


TheaperDeng (approved):
LGTM

PR #6869, opened 2022-12-05
Nano: a better support on openvino and onnxruntime for models whose forward args with default value

TheaperDeng:
### 1. Why the change?
There has been report that we have bad support for models whose forward args have default values or other parameters
e.g.
```python
class Net(nn.Module):
    def __init__(self):
        super().__init__()
    def forward(self, x, a=True, b=False):
        if a:
            return x+1
        if b:
            return x-1
        return x
```

We should have a much more intelligent args inspectation to:

- [x] generate dynamic axis for only tensor args
e.g., In this case, `a` and `b` will not generate any dynamic axis.
- [x] allow users not inputting input sample for those args who has default values
e.g., In this case, users could
```python
accmodel = InferenceOptimizer.trace(model,
                                    accelerator="onnxruntime",
                                    input_sample=(torch.rand(2,3,1,1),False,True))
```
or just
```python
accmodel = InferenceOptimizer.trace(model,
                                    accelerator="onnxruntime",
                                    input_sample=torch.rand(2,3,1,1))  # the last 2 parameters will use default values (True, False) automatically.
```
- [ ] (Will do next PR, only for onnxruntime) support better type transformation, not only for tensor

### 2. User API changes
Nothing

### 3. Summary of the change 
1. Inspect the forward args and decide which args is tensor by considering
  - users' type hint
  - default values
2. complete the input sample by using the default values

### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/utils/inference/pytorch/model_utils.py
+++ b/python/nano/src/bigdl/nano/utils/inference/pytorch/model_utils.py
@@ -85,7 +85,7 @@ def get_tensor_args(model):
         # check if defaults is Tensor
         default_args_start_from = args_length - defaults_length
         if i >= default_args_start_from:
-            if type(forward_defaults[i-default_args_start_from]) == torch.Tensor:
+            if type(forward_defaults[i - default_args_start_from]) == torch.Tensor:
                 tensor_args.append(arg)
             continue
         # nothing is provided, we assume it is a Tensor
@@ -115,11 +115,12 @@ def complement_input_sample(model, input_sample):
     # complement the input sample by defaults
     if isinstance(input_sample, Sequence):
         input_sample_complement = input_sample
-        input_sample_complement += forward_defaults[-(len(forward_args)-input_sample_length):]
+        input_sample_complement += forward_defaults[-(len(forward_args) - input_sample_length):]
     else:
         input_sample_complement = []
         input_sample_complement.append(input_sample)
-        input_sample_complement += list(forward_defaults[-(len(forward_args)-input_sample_length):])
+        input_sample_complement +=\
+            list(forward_defaults[-(len(forward_args) - input_sample_length):])
 
     return tuple(input_sample_complement)
 
--- a/python/nano/src/bigdl/nano/utils/inference/pytorch/model_utils.py
+++ b/python/nano/src/bigdl/nano/utils/inference/pytorch/model_utils.py
@@ -27,6 +27,10 @@ def get_forward_args(model):
     '''
     This function is to get all the arguments(excepts *args and **kwargs)
     It will return a list of arg name
+    E.g.
+    def forward(self, a, b=1, c: int = 3, *args, **kwargs):
+        pass
+    it will return ['a', 'b', 'c']
     '''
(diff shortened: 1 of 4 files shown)


python/nano/src/bigdl/nano/utils/inference/pytorch/model_utils.py
```diff
@@ -32,6 +36,95 @@ def get_forward_args(model):
...
+    '''
+    forward_defaults = inspect.getfullargspec(model.forward).defaults
+    if isinstance(model, LightningModule):
+        if not isinstance(model, AcceleratedLightningModule):
+            # forward param list for compiled model
+            forward_defaults = get_forward_defaults(model.model)
+    return forward_defaults
+
+
+def get_forward_annotations(model):
```
rnwang04, line 52: Maybe more comments to explain the three functions, it's a little hard to understand now.

python/nano/src/bigdl/nano/utils/inference/pytorch/model_utils.py
```diff
@@ -32,6 +36,95 @@ def get_forward_args(model):
...
+
+def complement_input_sample(model, input_sample):
+    '''
+    This function will give a complemented input sample
+    Mainly using default value to complete.
+    '''
+    forward_args = get_forward_args(model)
+    forward_defaults = get_forward_defaults(model)
+    input_sample_length = 1
+    if isinstance(input_sample, Sequence):
```
rnwang04, line 104: what if input_sample is a dict？
or now we only support sequence.

python/nano/test/onnx/pytorch/test_onnx.py
```diff
@@ -177,6 +177,64 @@ def test_onnx_context_manager(self):
...
+
+        model = Net()
+
+        x = torch.rand(1,3,1,1)
+        y = 3
+        result_true = model(x, y)
+        # sample with only required parameters (in a tuple)
+        accmodel = InferenceOptimizer.trace(model,
+                                            accelerator="onnxruntime",
+                                            input_sample=(torch.rand(2,3,1,1),3))
```
rnwang04, line 234: in this case, can we only pass in torch.rand(2,3,1,1)?

python/nano/src/bigdl/nano/utils/inference/pytorch/model_utils.py
TheaperDeng: Done

python/nano/src/bigdl/nano/utils/inference/pytorch/model_utils.py
TheaperDeng: Could we support dict as an input sample now? I did not find any ut for this.

python/nano/test/onnx/pytorch/test_onnx.py
TheaperDeng: change a little bit of the uts. Done

Oscilloscope98 (approved):
LGTW

MeouSker77 (approved):
LGTM

PR #6871, opened 2022-12-05
Chronos: Update doc to illustrate deploy mode in `TSDataset`

hjzin:
### 1. Why the change?

Since deploy mode is introduced in `TSDataset`, we need to update the related doc to tell users when and how to use it.

### 2. User API changes

N/A
### 3. Summary of the change 

Update related doc to tell users when and how to use deploy mode.
Preview:
https://zhaojie-doc.readthedocs.io/en/deploy-mode-doc/doc/Chronos/Overview/data_processing_feature_engineering.html#create-a-tsdataset

https://zhaojie-doc.readthedocs.io/en/deploy-mode-doc/doc/Chronos/Howto/how_to_process_data_in_production_environment.html#In-production-environment

### 4. How to test?
- [ ] Document test


TheaperDeng:
In the how to guide, we could
1. in code snip, we could remove the setting of `fit=False` and `is_predict=True` right?
2. in the note, we could only let users set `deploy_mode=True`

PR #6874, opened 2022-12-06
Nano: update join logic in context manager

rnwang04:
## Description

This PR updates the logic of how to join multi context managers in `get_context` as it's better to report warning instead of throwing error to block the user. 
1. If two context managers have difference precision (bf16 and non bf16), we will return `AutocastContextManager()`
2. If only one context manager have `thread_num`, we will set `thread_num` to that value
3. If two context managers have different `thread_num`, we will set `thread_num` to the larger one

### How to test?
- [x] Unit test
- [x] Application test

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
+++ b/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
@@ -808,8 +808,13 @@ class InferenceOptimizer(BaseInferenceOptimizer):
             def is_bf16(x):
                 return isinstance(x, AutocastContextManager)
             if (is_bf16(manager1) ^ is_bf16(manager2)) is True:
-                invalidInputError(False, "Can't obtain a new context manager for BF16 model"
-                                  "and non BF16 model.")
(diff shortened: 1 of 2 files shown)


python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -804,27 +804,31 @@ def obtain_manager(input_model):
                                                             precision="fp32")
             return _context_manager
 
-        def join_manager(manager, new_manager):
+        def join_manager(manager1, manager2):
             def is_bf16(x):
                 return isinstance(x, AutocastContextManager)
-            if (is_bf16(manager) ^ is_bf16(new_manager)) is True:
+            if (is_bf16(manager1) ^ is_bf16(manager2)) is True:
```
jason-dai, line 810: why throw error? just use `autocast` and report warning?

rnwang04: Done. Have used warning to replace error.

jason-dai:
LGTM

PR #6875, opened 2022-12-06
[ORCA] Support Optional `model_creator` with PyTorch Pyspark Estimator

sgwhat:
## Description
This PR based on https://github.com/intel-analytics/BigDL/pull/6751 to support loading an entire torch model instead of defining a model with `model_creator`.

### 1. Why the change?
For user-issues, some users need to load a saved model directly for continuous prediction and training.

### 2. User API changes
Usage with ignoring `model_creator`.
```python
est = estimator.from_torch(optimizer=optimizer_creator, loss=nn.BCELoss(),
                           metrics=Accuracy(), backend="spark")
```

### 3. Summary of the change 
Based on save/load entire model PR https://github.com/intel-analytics/BigDL/pull/6751, PyTorch Estimator could load an entire model on driver, and broadcast model class (while spark executor could not serialize `model` and `model_path`) to executors to replace `model_creator`.

### 4. How to test?
- [x] Unit test
- [x] Yarn test

### 5. To Do
- [x] Complete pipeline after merging https://github.com/intel-analytics/BigDL/pull/6751.


python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
```diff
@@ -116,7 +117,7 @@ def __init__(
 
         sc = OrcaContext.get_spark_context()
 
-        if not isinstance(model_creator, types.FunctionType):
+        if model_creator is not None and not isinstance(model_creator, types.FunctionType):
```
hkvision, line 120: just `if model_creator and ...`?

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
```diff
@@ -102,6 +102,7 @@ def __init__(
             sync_stats=True,
             log_level=logging.INFO,
             model_dir=None,
+            model_class=None,
```
hkvision, line 105: Why do you need the model_class as an argument?
Is this argument supposed to be input/modified by users?

sgwhat: oh yes, this should be removed.

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
```diff
@@ -175,7 +176,8 @@ def __init__(
             cluster_info=self._get_cluster_info(sc),
             **local_init_params)
 
-        self.state_dict = self.driver_runner.get_state_dict()
+        if self.model_creator or self.model_class:
+            self.state_dict = self.driver_runner.get_state_dict()
```
hkvision, line 180: Is the logic here correct? This is the init method of PySparkEstimator and model_class will always be None?

python/orca/src/bigdl/orca/learn/pytorch/pytorch_ray_estimator.py
```diff
@@ -76,6 +76,7 @@ def __init__(
             loss_creator=None,
             metrics=None,
             scheduler_creator=None,
+            model_class=None,
```
hkvision, line 79: Is there unit test for ray backend or you want to put in a following PR?
This design doesn't work for ray backend since the Runner is initialized at the time when init the estimator? You have no way to modify model_class?

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
sgwhat: yes correct, some state_dict like "model" will be None in this case, so we needn't get state_dict here. Instead, `self.state_dict` will get value after `estimator.fit` or `load`.

python/orca/src/bigdl/orca/learn/pytorch/pytorch_ray_estimator.py
sgwhat: It's just for fix ray estimator ut error and not an implementation currently, since ray and pyspark depend on the same `torch_runner`.

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
hkvision: I mean self.model_class will always be None as it is in the init method?

sgwhat: sure.

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
```diff
@@ -605,6 +621,9 @@ def load(self, model_path):
                 self.state_dict = [re.state_dict() for re in res]
         else:
             self.state_dict = res.state_dict()
+        if self.model_creator is None:
+            self.model_creator = lambda config: res
```
hkvision, line 625: I think it is better to put torch.load again in the creator instead of directly pass res
If model_creator is None, we can assume user load the entire model right? Maybe also add a check?

python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
```diff
@@ -176,17 +176,20 @@ def setup_components(self):
...
-        self.models = self.model_creator(self.config)
-        if isinstance(self.models, nn.Sequential) or not isinstance(self.models, Iterable):
-            self.models = [self.models]
-        invalidInputError(all(isinstance(model, nn.Module) for model in self.models),
-                          ("All models must be PyTorch models: {}.".format(self.models)))
+        if self.model_creator:
+            self.models = self.model_creator(self.config)
 
-        if self.optimizer_creator is not None:
+        if self.models:
```
hkvision, line 182: can remove this if to combine with `if self.model_creator`?

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
sgwhat: That cannot be implemented since we load the entire model instead of model state_dict, there will be a serialization error when loading the entire model on spark executor.

hkvision: Why load the entire model on each runner would have serialization error? no serialization is involved...

sgwhat: torch needs to serialize the entire model to get the model class.

PR #6879, opened 2022-12-06
Update shard_size

hkvision:
Put OrcaContext._shard_size in estimators to avoid ambiguity.
https://github.com/intel-analytics/BigDL/pull/6799#discussion_r1040499461

https://github.com/intel-analytics/BigDL/pull/6881#discussion_r1041720913 Change default batch_size for predict to be 32.

### DIFF
--- a/python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
@@ -222,13 +222,14 @@ class PyTorchPySparkEstimator(BaseEstimator):
                 You can also provide custom metrics by passing in a custom training_operator_cls
                 when creating the Estimator.
         """
+        shard_size = OrcaContext._shard_size if OrcaContext._shard_size else batch_size
         data, validation_data = maybe_dataframe_to_xshards(data,
                                                            validation_data=validation_data,
                                                            feature_cols=feature_cols,
                                                            label_cols=label_cols,
                                                            mode="fit",
                                                            num_workers=self.num_workers,
-                                                           shard_size=batch_size)
+                                                           shard_size=shard_size)
 
         sc = OrcaContext.get_spark_context()
         _ = self.create_tcpstore_server()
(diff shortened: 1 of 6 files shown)


python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
```diff
@@ -222,13 +222,14 @@ def fit(self,
                 You can also provide custom metrics by passing in a custom training_operator_cls
                 when creating the Estimator.
         """
+        shard_size = OrcaContext._shard_size if OrcaContext._shard_size else batch_size
```
jason-dai, line 225: batch_size should override shard_size

hkvision: Should not?
batch_size will always be used as the default shard size.
But if users specify OrcaContext._shard_size, they wish to use this value and it should override batch size? Isn't this the logic?

jason-dai: The user should not know whether we are using XShards or not inside Estimator; Estimator should manage its own XShards implementation itself.

hkvision: So you mean not use OrcaContext._shard_size, but always use batch_size and this value is not configurable?

jason-dai: I think it's reasonable that the user should always configure batch_size when using Estimator.

hkvision: Yes, user would configure batch_size; and using batch_size to convert to xhards will always work.
So directly throw OrcaContext._shard_size here?

jason-dai: > Yes, user would configure batch_size; and using batch_size to convert to xhards will always work. So directly throw OrcaContext._shard_size here?

I think you may ignore OrcaContext._shard_size here

hkvision:
@jason-dai Updated. Take a further look.
Add the check for batch_size to be a positive int to make sure it can be used when converting to xshards.
For tf2, batch_size could be None if user input a data_creator and he handles batch by himself. For all other cases, batch_size should be a positive int.

PR #6893, opened 2022-12-07
Nano: improve multi-instance inference again

MeouSker77:
## Description

Remove sort operation in multi-instance inference


rnwang04:
Maybe update `def to_multi_instance(model: nn.Module, num_processes: int,` to `def to_multi_instance(model: nn.Module, num_processes: Optional[int] = 1,` too

jason-dai:
> > Maybe update `def to_multi_instance(model: nn.Module, num_processes: int,` to `def to_multi_instance(model: nn.Module, num_processes: Optional[int] = 1,` too
> 
> done

Why default to 1? Why not default to 2 (or 4)?

PR #6909, opened 2022-12-07
fix jackson mapper CVE

Le-Zheng:
## Description

Jackson mapper CVE issue.


Le-Zheng:
All tests are passed.

PR #6912, opened 2022-12-08
[PPML] Modify bigdl-ppml-submit.sh to enable encrypted-fs in client driver

gc-fu:
## Description

Modify bigdl-ppml-submit.sh to enable encrypted-fsd in client driver
In Spark Client mode, the client will also doing some job which require to access the encrypted-fsd


zzti-bsj (approved):
LGTM

PR #6918, opened 2022-12-08
Nano: add channels_last  convert check

ACupofAir:
### 1. Why the change?
When we pass in a `input_sample` which is not consist of tensor, or one of the item in input doesn't have `.to()` method, it will lead to an error to convert it on `channels_last` is true.

### 2. Summary of the change 
* When user directly use `trace` or `quantize` function with a `input_sample` pass in or use `optimize` function which mean the `input_sample` is **not None**, we will compute a list named `channels_last_available` which instruct which item in the `input_sample` can be convert to `channels_last`.
* When user directly use `trace` or `quantize` function and don't pass in `input_sample`, I make the list `channels_last_available` empty, and the `forward_step` function will convert default, but if there are some errors occure, there is an alert to tell user they should pass in a `input_sample` to try to solve it.

### DIFF
--- a/python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
+++ b/python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
@@ -29,7 +29,6 @@ class PytorchIPEXJITModel(AcceleratedLightningModule):
         dtype=None,
         use_jit=False,
         channels_last=None,
-        channels_last_available=[],
         thread_num=None,
         from_load=False,
         inplace=False,
@@ -68,7 +67,19 @@ class PytorchIPEXJITModel(AcceleratedLightningModule):
             )
             return
         self.channels_last = channels_last
-        self.channels_last_available = channels_last_available
+        if self.channels_last:
+            # try channels_last available
+            channels_last_available = []
+            if input_sample:
+                channels_last_available = [True]*len(input_sample)
+                for idx, input in enumerate(input_sample):
+                    try:
+                        input.to(memory_format=torch.channels_last)
+                    except:
+                        channels_last_available[idx] = False
+                    else:
+                        channels_last_available[idx] = True
+            self.channels_last_available = channels_last_available
         self.original_state_dict = model.state_dict()
         self.use_ipex = use_ipex
         self.use_jit = use_jit
@@ -116,7 +127,8 @@ class PytorchIPEXJITModel(AcceleratedLightningModule):
 
     def forward_step(self, *inputs):
         if self.channels_last is True:
-            if self.channels_last_available: # it mean the input_sample is not None
+            # check if `channels_last_available` has been generated
+            if self.channels_last_available:
                 for idx, input in enumerate(inputs):
                     if self.channels_last_available[idx]:
                         input.to(memory_format=torch.channels_last)
@@ -126,7 +138,7 @@ class PytorchIPEXJITModel(AcceleratedLightningModule):
                 except:
                     print("************************error:***************************\n")
                     print("You should pass in a non_empty input_sample")
-        return self.model(*tuple(inputs))
+        return self.model(*inputs)
 
     def on_forward_end(self, outputs):
         return outputs
--- a/python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
+++ b/python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
@@ -21,9 +21,20 @@ from bigdl.nano.pytorch.utils import patch_attrs_from_model_to_object
 
 
 class PytorchIPEXJITModel(AcceleratedLightningModule):
-    def __init__(self, model: torch.nn.Module, input_sample=None, use_ipex=False, dtype=None,
-                 use_jit=False, channels_last=None, thread_num=None, from_load=False,
-                 inplace=False, jit_strict=True):
+    def __init__(
+        self,
+        model: torch.nn.Module,
+        input_sample=None,
+        use_ipex=False,
+        dtype=None,
+        use_jit=False,
+        channels_last=None,
+        channels_last_available=[],
+        thread_num=None,
+        from_load=False,
+        inplace=False,
+        jit_strict=True,
+    ):
         """
         This is the accelerated model for pytorch and ipex/jit.
         All the external API is based on InferenceOptimizer, so what we have here is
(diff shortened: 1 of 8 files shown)


python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
```diff
@@ -94,16 +110,28 @@ def __init__(self, model: torch.nn.Module, input_sample=None, use_ipex=False, dt
...
+        ]
 
     def on_forward_start(self, inputs):
         return inputs
 
     def forward_step(self, *inputs):
         if self.channels_last is True:
-            inputs = tuple(map(lambda x: x.to(memory_format=torch.channels_last), inputs))
-        return self.model(*inputs)
+            if self.channels_last_available: # it mean the input_sample is not None
```
TheaperDeng, line 124:
```python
# check if `channels_last_available` has been generated
```

python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
```diff
@@ -94,16 +110,28 @@ def __init__(self, model: torch.nn.Module, input_sample=None, use_ipex=False, dt
...
+                for idx, input in enumerate(inputs):
+                    if self.channels_last_available[idx]:
+                        input.to(memory_format=torch.channels_last)
+            else:
```
TheaperDeng, line 128: why don't we just generate the `channels_last_available` here? we have a precious input sample right :)

python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
```diff
@@ -94,16 +110,28 @@ def __init__(self, model: torch.nn.Module, input_sample=None, use_ipex=False, dt
...
+                try:
+                    inputs = tuple(map(lambda x: x.to(memory_format=torch.channels_last), inputs))
+                except:
+                    print("************************error:***************************\n")
+                    print("You should pass in a non_empty input_sample")
+        return self.model(*tuple(inputs))
```
TheaperDeng, line 134: could we just use `*inputs` here?

python/nano/src/bigdl/nano/pytorch/amp/bfloat16.py
```diff
@@ -106,7 +107,18 @@ def on_forward_start(self, inputs):
...
+            if self.channels_last_available: # it mean the input_sample is not None
+                for idx, input in enumerate(inputs):
+                    if self.channels_last_available[idx]:
+                        input.to(memory_format=torch.channels_last)
+            else:
+                try:
+                    inputs = tuple(map(lambda x: x.to(memory_format=torch.channels_last), inputs))
+                except:
+                    print("************************error:***************************\n")
+                    print("You should pass in a non_empty input_sample")
```
TheaperDeng, line 119: same comments as before

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -560,6 +560,19 @@ def quantize(model: nn.Module,
...
+        channels_last_available = []
+        if input_sample:
+            channels_last_available = [True]*len(input_sample)
+            for idx in range(len(input_sample)):
+                try:
+                    input_sample[idx].to(memory_format=torch.channels_last)
+                except:
+                    channels_last_available[idx] = False
+                else:
+                    channels_last_available[idx] = True
```
TheaperDeng, line 574: we should move this logic to `PytorchIPEXJITModel`'s `__init__()`

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -753,6 +769,20 @@ def trace(model: nn.Module,
...
```
TheaperDeng, line 783: same here

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -780,6 +810,7 @@ def trace(model: nn.Module,
             use_jit = (accelerator == "jit")
             return PytorchIPEXJITModel(model, input_sample=input_sample, use_ipex=use_ipex,
                                        use_jit=use_jit, channels_last=channels_last,
+                                       channels_last_available=channels_last_available,
                                        thread_num=thread_num, inplace=inplace,
                                        jit_strict=jit_strict)
         invalidInputError(False, "Accelerator {} is invalid.".format(accelerator))
```
TheaperDeng, line 816: we want some unit test for this fix in https://github.com/intel-analytics/BigDL/blob/main/python/nano/test/pytorch/tests/inference/test_ipex_jit_inference.py

TheaperDeng: we need to make sure `channels_last_available` is well recovered even after `InferenceOptimizer.save/load`

python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
```diff
@@ -117,18 +157,22 @@ def __getattr__(self, name: str):
...
-                       "thread_num": self.thread_num,
-                       "jit_strict": self.jit_strict})
+        status.update(
+            {
+                "use_ipex": self.use_ipex,
+                "use_jit": self.use_jit,
+                "channels_last": self.channels_last,
+                "checkpoint": "ckpt.pth",
+                "thread_num": self.thread_num,
+                "jit_strict": self.jit_strict,
```
TheaperDeng, line 167: should we save the `self.channels_last_available` here? Since we do want it be recoverred after users calling save/load

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -350,6 +350,8 @@ def optimize(self, model: nn.Module,
                                                                thread_num=thread_num)
 
         print("==========================Start Optimization==========================")
+        print('======================DEBUG START: Junes nano======================')
+        print('======================DEBUG  END : Junes nano======================')
```
TheaperDeng, line 354: no need for these 2 lines

python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
TheaperDeng: change to something like
```python
else:
    self.channels_last_available = generate_channels_last_available(inputs)
    for idx, input in enumerate(inputs):
        if self.channels_last_available[idx]:
            input.to(memory_format=torch.channels_last)
```

python/nano/test/pytorch/tests/inference/test_bf16_ipex.py
```diff
@@ -151,6 +152,116 @@ def hello():
             new_model.width
 
 
+    def test_bf16_channels_last_inference_with_jit(self):
+        model = resnet18(num_classes=10)
+
+        x = torch.rand((10, 3, 256, 256))
```
TheaperDeng, line 158: it seems we keep using `x=torch.rand((10, 3, 256, 256))` as the test input sample. An easy way to do this could be
1. add 1 test function in each test file (test_bf16, test_bf16_ipex, test_ipex_jit_inference)
2. For these test function, write your own test model (could be an extremely easy one with 3 input)
3. use (x1, x2, x3) as input, where x1, x3 are 4-dim tensor while x2 is 3-dim.
4. test save/load

ACupofAir:
replaced by #7151

PR #6930, opened 2022-12-09
Nano: fix problem of default values in forward args for onnxruntime int8 quantization

TheaperDeng:
### 1. Why the change?
This change is to make model with kwargs can be run through onnxruntime int8 quantization.

This PR  solves
- When the forward args has only bool and tensor type input
e.g.,
```python
class Net(nn.Module):
    def __init__(self):
        super().__init__()
    def forward(self, x, a=True, b=False):
        if a:
            return x+1
        if b:
            return x-1
        return x
```

- When the forward args has int or other type (with some limit)
e.g.,
```python
class Net(nn.Module):
    def __init__(self):
        super().__init__()
    def forward(self, x, a=3):
        return x + a
```
Here users should always input all the parameter as calib_data, nano will not patch the default values for them (TODO).

### 2. User API changes

Nothing changed

### 3. Summary of the change 

change a minor bug in dataloader_type_transformation

### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/nano/test/onnx/pytorch/test_inc_onnx.py
+++ b/python/nano/test/onnx/pytorch/test_inc_onnx.py
@@ -319,6 +319,25 @@ class TestOnnx(TestCase):
                                                calib_data=(torch.rand(2,3,1,1), 5))
         result_m = accmodel(data, np.array([5]))  # TODO: make this 5
 
+        # default None values
+        class Net(nn.Module):
+            def __init__(self):
+                super().__init__()
+            def forward(self, x, a=None):
+                if a is None:
+                    return x
+                else:
+                    return x + 1
+        model = Net()
+
+        data = torch.rand(1,3,1,1)
+
+        # sample with only required parameters (in a tuple)
+        accmodel = InferenceOptimizer.quantize(model,
+                                               accelerator="onnxruntime",
+                                               calib_data=torch.rand(2,3,1,1))
+        result_m = accmodel(data)
+
 
 if __name__ == '__main__':
     pytest.main([__file__])
--- a/python/nano/test/onnx/pytorch/test_inc_onnx.py
+++ b/python/nano/test/onnx/pytorch/test_inc_onnx.py
@@ -273,6 +273,35 @@ class TestOnnx(TestCase):
         accmodel(x2)
         accmodel(x3)
 
+    def test_onnx_inc_default_values(self):
+        # default bool values
+        class Net(nn.Module):
+            def __init__(self):
+                super().__init__()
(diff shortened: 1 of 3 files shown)


python/nano/test/onnx/pytorch/test_inc_onnx.py
```diff
@@ -240,6 +241,51 @@ def hello():
...
+        model = Net()
+
+        data = torch.rand(1,3,1,1)
+
+        # sample with only required parameters (in a tuple)
+        accmodel = InferenceOptimizer.quantize(model,
+                                               accelerator="onnxruntime",
+                                               calib_data=(torch.rand(2,3,1,1), 5))
+        result_m = accmodel(data, np.array([5]))  # TODO: make this 5
+
```
Oscilloscope98, line 288: For the case I met [error](https://github.com/analytics-zoo/nano/issues/225), the use case is:
```python
elif method == 'onnxruntime_int8_qlinear':
    model = InferenceOptimizer.quantize(model=model,
                                        precision='int8',
                                        accelerator='onnxruntime',
                                        calib_data=input_sample,
                                        input_sample=input_sample,
                                        method='qlinear',
                                        thread_num=4, # need to specify to make sure onnxruntime is using the correct thread
                                        dynamic_axes=dynamic_axes)

elif method == 'onnxruntime_int8_integer':
    model = InferenceOptimizer.quantize(model=model,
                                        precision='int8',
                                        accelerator='onnxruntime',
                                        calib_data=input_sample,
                                        input_sample=input_sample,
                                        method='integer',
                                        thread_num=4, # need to specify to make sure onnxruntime is using the correct thread
                                        dynamic_axes=dynamic_axes)
``` 
in which the `input_sample` is a simple 4d tensor, and model's forward method is like:
```python
 def forward(self, im, augment=False, visualize=False, val=False):
```
I am not quiet understand the use case that do not include a `input_sample` in the `quantize` function, and why passing `(data, np.array([5]))` in the `accmodel`.

TheaperDeng: So the `quantize` will first use `input_sample`(if valid, or `calib_data` will be used) to trace the model and then use `calib_data` to calib the model during the quantization. A special case should be there are `NoneType` or `bool` class parameter in the `forward`, if users want to change the default values of these parameter, they are required to put them in `input_sample`.

Onnxruntime will only take numpy ndarray as input for the prediction, that's why we need to use `np.array([5])` here. We should make `5` available as well.

It's quite tricky, we will put them in trouble shooting page for now.

PR #6931, opened 2022-12-09
Nano: add dynamic_axes parameter for `trace`/`quantize`/`optimize`

rnwang04:
## Description

Add `dynamic_axes` parameter for `trace`/`quantize`/`optimize`, so if input shape will vary during inference, user can specify this parameter manually.

### 1. Why the change?

https://github.com/analytics-zoo/nano/issues/241

### 2. User API changes

Add new `dynamic_axes` parameter for  `trace`/`quantize`/`optimize`, so if input shape will vary during inference, user can specify this parameter manually.

```
dynamic_axes = {}
# 'im' is the param name for input data in the arguments of the model's forward method
dynamic_axes['im'] = {0: 'batch_size', 2: 'width', 3: 'height'}
# Or just specify a list
dynamic_axes['im'] = [0, 2, 3]

model = InferenceOptimizer.trace(model=model,
                                 accelerator='onnxruntime',
                                 input_sample=input_sample,
                                 dynamic_axes=dynamic_axes)
```

### 3. Summary of the change 

- add new paramter `dynamic_axes` for add dynamic_axes parameter for `trace`/`quantize`/`optimize`

### 4. How to test?
- [x] Unit test
- [x] Document test
https://ruonantetdoc.readthedocs.io/en/add_dynamic_axes_and_update_api_doc/doc/PythonAPI/Nano/pytorch.html#bigdl-nano-pytorch-inferenceoptimizer

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
+++ b/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
@@ -581,8 +581,8 @@ class InferenceOptimizer(BaseInferenceOptimizer):
                              at run-time), set dynamic_axes to a dict with schema:
                                 KEY (str): an input or output name. Each name must also be provided
                                 in input_names or output_names.
-                                VALUE (dict or list): If a dict, keys are axis indices and values are
-                                axis names. If a list, each element is an axis index.
+                                VALUE (dict or list): If a dict, keys are axis indices and values
+                                are axis names. If a list, each element is an axis index.
                              If accelerator != 'openvino'/'onnxruntime', it will be ignored.
         :param sample_size: (optional) a int represents how many samples will be used for
                             Post-training Optimization Tools (POT) from OpenVINO toolkit,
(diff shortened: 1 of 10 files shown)


rnwang04:
> LGTM, is there any need to add some ut for this?

I think it's necessary， and I have added ut for openvino/onnxruntime's trace/quantize.

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -245,6 +249,19 @@ def optimize(self, model: nn.Module,
                | grid mode. We will sort and combine according to the value you specified to
                | get the search range.
 
+        :param dynamic_axes: dict or boolean, default to True. By default the exported onnx model
+               will have the first dim of each input as a dynamic batch_size. If dynamic_axes=False,
```
Oscilloscope98, line 253: Should we modify the description for the `by default` case to include the logic added [here](https://github.com/intel-analytics/BigDL/pull/6869)?

Oscilloscope98 (approved):
Others LGTM

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: Done, have changed to `each Tensor input`

PR #6933, opened 2022-12-12
Nano: fix getattr of jit when inplace=True

rnwang04:
## Description

Fix bug of jit related model still don't have original attributes.

### 1. Why the change?

https://github.com/analytics-zoo/nano/issues/228#issuecomment-1342288227

### 2. User API changes

No change

### 3. Summary of the change 

Fix bug of jit related model still don't have original attributes.

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/nano/test/pytorch/tests/inference/test_ipex_jit_inference.py
+++ b/python/nano/test/pytorch/tests/inference/test_ipex_jit_inference.py
@@ -90,10 +90,20 @@ class IPEXJITInference_gt_1_10:
             print("hello world!")
         # patch a function
         model.hello = hello
-        
+
         # test jit + ipex
         new_model = InferenceOptimizer.trace(model, accelerator="jit",
                                              use_ipex=True,
+                                             input_sample=self.data_sample)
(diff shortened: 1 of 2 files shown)


Oscilloscope98 (approved):
LGTM

python/nano/test/pytorch/tests/inference/test_ipex_jit_inference.py
```diff
@@ -116,7 +126,16 @@ def hello():
         new_model.hello()
         with pytest.raises(AttributeError):
             new_model.width
-        
+
+        # test ipex inplace
+        new_model = InferenceOptimizer.trace(model, use_ipex=True, inplace=True)
```
Oscilloscope98, line 131: Will `model` be changed after `inplace=True`? I am not sure will it affect the following tests?

rnwang04: Good catch, although it looks don't affect below test, but I may define a new model here.

PR #6940, opened 2022-12-12
Chronos: Add chronos example to github action

liangs6212:


plusbang (commented):
LGTM, only one question - don't need test on spark 2.4 ?

TheaperDeng (approved):
LGTM, it's ok we currently only test 2.4

PR #6949, opened 2022-12-13
Nano: support ipex quantization as backup for inc ipex quantization

rnwang04:
### 4. How to test?
- [ ] Unit test


python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -689,17 +701,40 @@ def quantize(model: nn.Module,
...
-                                    framework=framework,
-                                    conf=conf,
-                                    approach=approach,
-                                    tuning_strategy=tuning_strategy,
-                                    accuracy_criterion=accuracy_criterion,
-                                    timeout=timeout,
-                                    max_trials=max_trials,
-                                    onnxruntime_session_options=onnxruntime_session_options)
 
+                if framework != 'pytorch_ipex':
```
MeouSker77, line 705: We should avoid too much copy-paste

```python
if framework != 'pytorch_ipex':
    try:
        return inc_quantize(...)
    except Exception:
        framework = 'pytorch_ipex'
return PytorchIPEXQuantizationModel(...)
```

rnwang04: only 'pytorch_ipex' needs try/except and only 'pytorch_ipex' can use `PytorchIPEXQuantizationModel(...)`

rnwang04: Have revised based on the second choice.

PR #6967, opened 2022-12-13
[Nano] Support users to choose `jit.trace`/`jit.script` in `InferenceOptimizer.trace`/`quantize`

Oscilloscope98:
## Description

Support users to choose `jit.trace`/`jit.script` when applying `jit` as the accelerator in `InferenceOptimizer.trace`/`quantize`

### 1. Why the change?

When converting a model to TorchScript, users could either use `jit.trace` or `jit.script`, or even mix both of them. These two converting methods have their own prons and cons. For example, for a model that are both scriptable and traceable, `jit.trace` could always generate a simpler graph (thus probably better performance). However, `jit.trace` would cause problems when the model includes dynamic workflows, it could also capture variables as constants, etc. (Reference: this [blog](https://ppwwyyxx.com/blog/2022/TorchScript-Tracing-vs-Scripting/#Let-Tracing-Generalize))

Currently, in `InferenceOptimizer.trace`/`quantize`, we are only using `jit.trace`. For better flexibility, we should enable users to choose either of them. A try-except logic will also be used (first `jit.trace`, failed then `jit.script`) by default to increase the success rate of converting the model.

### 2. User API changes

Added a `jit_method` parameter to  `InferenceOptimizer.trace`/`quantize` for users to choose whether to use `jit.trace` or `jit.srcipt`. By default,  a try-except logic will be used (first `jit.trace`, failed then `jit.script`) .

```python
# use jit.trace
accmodel = InferenceOptimizer.trace(model, accelerator='jit', 
                                    input_sample=input_sample, 
                                    jit_method='trace')

# use jit.script
accmodel = InferenceOptimizer.trace(model, accelerator='jit', 
                                    input_sample=input_sample, 
                                    jit_method='script')

# use try-except logic
accmodel = InferenceOptimizer.trace(model, accelerator='jit', 
                                    input_sample=input_sample)
```

### 3. Summary of the change 
- Added a `jit_method` parameter to  `InferenceOptimizer.trace`/`quantize`
- Added the missing `jit_strict` parameter when loading a `PytorchIPEXJITBF16Model` model

### 4. How to test?
- [x] Unit test
- [x] Document test for related API: https://yuwentestdocs.readthedocs.io/en/nano-jit-script-trace/doc/PythonAPI/Nano/pytorch.html#bigdl.nano.pytorch.InferenceOptimizer.quantize

### 5. New dependencies

N/A

### 6. Possible further improvements
Try to detect/inform the case that `jit.trace` causes problems in `InferenceOptimizer.optimize` due to problems in dynamic workflows, capturing variables as constants, etc,

### DIFF
--- a/python/nano/src/bigdl/nano/deps/ipex/ipex_api.py
+++ b/python/nano/src/bigdl/nano/deps/ipex/ipex_api.py
@@ -36,7 +36,7 @@ def ipex_device():
 
 def PytorchIPEXJITModel(model, input_sample=None, use_ipex=False,
                         use_jit=False, channels_last=None, thread_num=None,
-                        inplace=False, jit_strict=True):
+                        inplace=False, jit_strict=True, jit_converter=None):
     '''
     :param model: the model(nn.module) to be transform.
     :param input_sample: torch tensor indicate the data sample to be used
@@ -48,16 +48,19 @@ def PytorchIPEXJITModel(model, input_sample=None, use_ipex=False,
     :param thread_num: the thread num allocated for this model.
     :param inplace: whether to perform inplace optimization. Default: ``False``.
     :param jit_strict: Whether recording your mutable container types.
+    :param jit_converter: use ``jit.trace`` or ``jit.script`` to 
+           convert a model to TorchScript.
     '''
     from .ipex_inference_model import PytorchIPEXJITModel
     return PytorchIPEXJITModel(model, input_sample=input_sample, use_ipex=use_ipex,
                                use_jit=use_jit, channels_last=channels_last,
-                               thread_num=thread_num, inplace=inplace, jit_strict=jit_strict)
+                               thread_num=thread_num, inplace=inplace, jit_strict=jit_strict,
+                               jit_converter=jit_converter)
 
 
 def PytorchIPEXJITBF16Model(model, input_sample=None, use_ipex=False,
                             use_jit=False, channels_last=None, thread_num=None,
-                            inplace=False, jit_strict=True):
+                            inplace=False, jit_strict=True, jit_converter=None):
     '''
     :param model: the model(nn.module) to be transform.
     :param input_sample: torch tensor indicate the data sample to be used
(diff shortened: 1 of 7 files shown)


python/nano/test/pytorch/tests/inference/test_bf16_ipex.py
```diff
@@ -150,6 +151,65 @@ def hello():
         with pytest.raises(AttributeError):
             new_model.width
 
+    def test_bf16_ipex_jit_converter(self):
```
rnwang04, line 154: You don't have to put two identical unit tests in two different files.
Now `test_bf.py` only contains ut for `bf16` and `bf16+channels_last`, and all `ipex / jit` related ut are put in `test_bf16_ipex.py` and `test_ipex_jit_inference.py`. So just put ut in the last two files.

python/nano/src/bigdl/nano/deps/ipex/ipex_api.py
```diff
@@ -48,16 +48,19 @@ def PytorchIPEXJITModel(model, input_sample=None, use_ipex=False,
     :param thread_num: the thread num allocated for this model.
     :param inplace: whether to perform inplace optimization. Default: ``False``.
     :param jit_strict: Whether recording your mutable container types.
+    :param jit_converter: use ``jit.trace`` or ``jit.script`` to
```
rnwang04, line 51: In my feeling, `jit_converter` is not intuitive and a little strange.

I think `jit_compiler` may be a choice：
_Scripting a function or nn.Module will inspect the source code, compile it as TorchScript code using the TorchScript compiler, and return a [ScriptModule]_\
_Trace a function and return an executable or [ScriptFunction] that will be optimized using just-in-time compilation._

rnwang04 (commented):
others LGTM.

python/nano/test/pytorch/tests/inference/test_bf16_ipex.py
Oscilloscope98: Thank you for the reminding! I am kind of confused by the file names. Thank you! Have updated,

python/nano/src/bigdl/nano/deps/ipex/ipex_api.py
Oscilloscope98: I also think that `jit_converter` could be not so intuitive.

However, after some investigation, I think one of the biggest difference between `jit.script` and `jit.trace` is that `jit.script` will "compiles" a model into graph, but `jit.trace` will not.

From the direct explanation in TorchScript official documentation:
image
From my understanding, it also shows that the complication of model converted from `jit.trace` will be conducted **jit in time** (but the complication will not be done by `jit.trace` itself).

So, for me, I feel like that `jit_compiler` will by itself already refer to `jit.script`, so may not be a suitable parameter name.

@TheaperDeng Do you have any idea what will be a more suitable name?

TheaperDeng: maybe we could just call it `jit_method` after some discussion with @MeouSker77 @rnwang04

Oscilloscope98: Have renamed the parameter.

TheaperDeng (approved):
LGTM

PR #6975, opened 2022-12-14
[PPML] Update attestation code

hzjane:
## Description
Update attestation code to enable warning result.

### DIFF
--- a/scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/attestation/AttestationCLI.scala
+++ b/scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/attestation/AttestationCLI.scala
@@ -105,8 +105,8 @@ object AttestationCLI {
             if (verifyQuoteResult == 0) {
               System.out.println("Quote Verification Success!")
             } else if (verifyQuoteResult == 1) {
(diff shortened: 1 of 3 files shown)


scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/attestation/AttestationCLI.scala
```diff
@@ -101,19 +102,16 @@ object AttestationCLI {
...
               System.out.println("Quote Verification Success!")
             } else if (verifyQuoteResult == 1) {
-              System.out.println("WARNING:Quote verification passed but BIOS is not up to date." +
-                " result=" + verifyQuoteResult)
+              System.out.println("WARNING:Quote verification passed but BIOS or the software" +
+                " is not up to date.")
             } else if (debug == "true") {
-              System.out.println("ERROR:Quote Verification Fail! In debug mode, continue." +
-                " result=" + verifyQuoteResult)
+              System.out.println("ERROR:Quote Verification Fail! In debug mode, continue.")
```
Uxito-Ada, line 111: May be not print **ERROR** in debug mode, otherwise user may be confused.

PR #6980, opened 2022-12-14
Nano: Update tensorflow example and UT

MeouSker77:
## Description

Update tensorflow example and UT:

- remove `model = Model(model.inputs, model.outputs)` because we don't need it any more


rnwang04 (approved):
LGTM

PR #6981, opened 2022-12-14
Nano: fix install `bigdl-nano` only

MeouSker77:
## Description

Before this PR, if user run `pip install bigdl-nano`,  `pytorch-lightning` won't be installed automatically. However, `Trainer`, `TorchNano` and `InferenceOptimizer` are all depending on `pytorch-lightning`.

User must run `pip install bigdl-nano[pytorch]` to use `Trainer` and `TrochNano`, or run `pip install bigdl-nano[pytorch,inference]` to use all these features

After this PR, if user run `pip install bigdl-nano`, it is equivalent to running `pip install bigdl-nano[pytorch,tensorflow,inference]`, so that he can use all these features

### DIFF
--- a/python/nano/setup.py
+++ b/python/nano/setup.py
@@ -171,7 +171,7 @@ def setup_package():
         url='https://github.com/intel-analytics/BigDL',
         install_requires=install_requires,
         # "" means user run `pip install bigdl-nano`, install all dependencies in this case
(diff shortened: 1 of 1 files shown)


MeouSker77:
closed this PR because it cannot solve this problem

Issue #6968, opened 2022-12-13
[Nano] load pytorch_fx model will get an error when using 2.2.0b20221212 version


NingWang5:
Below is my code to load ckpt of pytorch_fx:
``` 
config = {'input_feature_num':8,
          'output_feature_num':1,
          'past_seq_len':48,
          'future_seq_len':1,
          'kernel_size':3,
          'repo_initialization':True,
          'dropout':0.1,
          'seed': 0,
          'num_channels':[16]*3
          }
model = model_creator(config)
loss = torch.nn.MSELoss()
optimizer = torch.optim.Adam(lr=0.001, params=model.parameters())

lit_model = Trainer.compile(model, loss, optimizer)

fx_q_model = InferenceOptimizer.load("./gpu_ckpt/fx_q_model", lit_model)
```

It will successed when using 2.2.0b20221128 version, but get an error when using the newest version (2.2.0b20221212), and the error log is as following:

![image](https://user-images.githubusercontent.com/113402880/207269515-c88419fc-163a-4a38-b839-e33df42a2434.png)

rnwang04:
Do you save your model again with the newest version (2.2.0b20221212)? or you just load the model which is saved in 2.2.0b20221128 version.

NingWang5:
THX，nano versions of save and load are not unified, no issue now.

TheaperDeng:
> THX，nano versions of save and load are not unified, no issue now.

well that's a good catch for us. Although we may not support some nightly built version's checkpoint, but should we support 2.1.0's checkpoint in 2.2.0? (no strong idea about this) @rnwang04

rnwang04:
> > THX，nano versions of save and load are not unified, no issue now.
> 
> well that's a good catch for us. Although we may not support some old nightly built version's checkpoint, but should we support 2.1.0's checkpoint in 2.2.0? (no strong idea about this) @rnwang04

I agree, maybe we should support such behavior of save & load and I think it's not too difficult for us.

PR #6983, opened 2022-12-14
Nano: support backward compatibility of loading

rnwang04:
## Description

This PR wants to support users to load checkpoint which is saved in earlier version (nano versions of save & load are not unified and load version > save version)

### 1. Why the change?

https://github.com/intel-analytics/BigDL/issues/6968

### 2. User API changes

No change.

### 3. Summary of the change 

- support users to load checkpoint which is saved in earlier version

### 4. How to test?
- [x] Unit test

Have verified this PR in local test with checkpoint save in bigdl-2.1.


TheaperDeng (approved):
LGTM

PR #6985, opened 2022-12-14
Bug fix for spark_df_to_rdd_pd

hkvision:


python/orca/src/bigdl/orca/data/utils.py
```diff
@@ -479,12 +478,12 @@ def f(iter):
     return f
 
 
-def to_pandas(columns: List[str], squeeze: bool=False, index_col: Optional[str]=None,
+def to_pandas(df: "DataFrame", squeeze: bool=False, index_col: Optional[str]=None,
```
hkvision, line 481: The first argument should be df, columns is wrong, and when running the transformation originally, would throw the info:
```
create shards from Spark DataFrame attempted Arrow optimization failed as: name 'df' is not defined. Will try without Arrow optimization
```
df is not defined in line 507.

PR #6986, opened 2022-12-14
Remove save entire model in doc

hkvision:
Not recommended at present. Loading the entire model need to find the definition of the model class.


lalalapotter (approved):
LGTM

PR #6988, opened 2022-12-14
Orca: Deprecate Training_Operator in Torch Estimator

leonardozcm:
## Description

Will Merge this after 2.2.0 release.


python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
```diff
@@ -138,16 +136,15 @@ def __init__(
...
-
         self.model_dir = parse_model_dir(model_dir)
 
         self.model_creator = model_creator
         self.optimizer_creator = optimizer_creator
 
+        if not loss_creator:
+            invalidInputError(False,
+                              "You must provide a loss_creator.")
+
```
hkvision, line 147: This modification is wrong.

PR #6989, opened 2022-12-15
fix org.apache.ivy:ivy Vulnerabilities issue

Le-Zheng:
## Description

image

### 1. Why the change?
org.apache.ivy:ivy has high vulnerability
exclude org.apache.ivy:ivy


Le-Zheng:
all tests are passed.

Issue #242, opened 2016-12-26
Fix the license header in source code


jason-dai:
Change to something like https://github.com/cloudera/livy/blob/master/core/src/main/scala/com/cloudera/livy/Utils.scala

PR #7000, opened 2022-12-15
Nano: Support patching original model'attributes to traced and quantized model

MeouSker77:
## Description

Support accessing original model's attributes after trace and quantization

### 1. Why the change?

Fix issue 3 of [#242](https://github.com/analytics-zoo/nano/issues/242)

### 2. User API changes

N/A

### 3. Summary of the change 

- Patch original model'attributes to traced and quantized model
- Recover the `input_spec` parameter of `InferenceOptimizer.quantize` and `InferenceOptimizer.optimize`

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...


python/nano/src/bigdl/nano/tf/utils.py
```diff
@@ -13,7 +13,70 @@
...
+    :return: `target_obj`
+    """
+    if inspect.ismethod(target_obj.__setattr__):
+        # `target_obj` has custom `__setattr__`
+        wrapper_type = _create_wrapper_type(type(target_obj), target_obj, source_obj)
+        wrapper_obj = wrapper_type.__new__(wrapper_type)
+        return wrapper_obj
+    else:
+        # `target_obj` has no custom `__setattr__`
+        for name in set(dir(target_obj)) - set(dir(target_obj)):
```
TheaperDeng, line 70: should this one be `set(dir(source_obj)) - set(dir(target_obj))` ?

MeouSker77: yes, fix it

python/nano/src/bigdl/nano/tf/utils.py
```diff
@@ -13,7 +13,70 @@
...
+
+def patch_attrs(target_obj: object, source_obj: object) -> object:
+    """
+    Patch attributes of `source_obj` to `target_obj`.
+
+    :param target_obj: target object
+    :param source_obj: source object
+    :return: `target_obj`
+    """
+    if inspect.ismethod(target_obj.__setattr__):
```
TheaperDeng, line 63: maybe this one will cover all the cases, since all obj has `__setattr__` right?

MeouSker77: no, inspect.ismethod('__setattr__') is `True` only when `__setattr__` is defined explicitly by user

TheaperDeng: Yeah, I had a try you are right

TheaperDeng (approved):
LGTM

PR #7001, opened 2022-12-15
dllib: The error message thrown does not contain the real reason

liu-shaojun:
### 1. Why the change?

When using ppml_context.py to encrypt data, the following error was thrown.
MicrosoftTeams-image (14)

This log indicates that the method createPPMLContext does not exist, but in fact, createPPMLContext does exist, the true reason is we specified the wrong path for the primarykey. 

image
https://github.com/intel-analytics/BigDL/blob/b67db454b486c7ac5829cbf55751411ad7636cb1/python/dllib/src/bigdl/dllib/utils/common.py#L631
Whenever the error message contains "does not exist", it is assumed that the method does not exist, and when the path of the key does not exist, it is also considered to be “method does not exist” and then the real cause of the error will be caught and not thrown out.

### DIFF
--- a/python/dllib/src/bigdl/dllib/utils/common.py
+++ b/python/dllib/src/bigdl/dllib/utils/common.py
@@ -630,7 +630,7 @@ def callBigDlFunc(bigdl_type, name, *args):
         except Exception as e:
             error = e
             regex = re.compile(r'Method\s' + name + r'.*does not exist')
-            # if the invoked method exist but something else went wrong, throw the exception 
+            # if the invoked method exist but something else went wrong, throw the exception
             if not regex.match(str(e)):
                 invalidOperationError(False, str(e), cause=e)
         else:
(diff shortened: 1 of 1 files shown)


python/dllib/src/bigdl/dllib/utils/common.py
```diff
@@ -628,8 +629,12 @@ def callBigDlFunc(bigdl_type, name, *args):
             result = callJavaFunc(api, *args)
         except Exception as e:
             error = e
-            if "does not exist" not in str(e):
+            regex = re.compile(r'Method\s' + name + r'.*does not exist')
+            if regex.match(str(e)):
                 invalidOperationError(False, str(e), cause=e)
```
PatrickkZ, line 634: if regex.match, which means the error is method not exist, we should skip this error by doing nothing, not call invalidOperationError

liu-shaojun: updated

PR #7003, opened 2022-12-15
Nano: add `weights_prepack` parameter for `trace`/`quantize`

rnwang04:
## Description

add `weights_prepack` parameter for `trace`/`quantize` to cope with oom.

### 1. Why the change?

The methods related to ipex may go out of memory on machines with small memory.

### 2. User API changes

new `weights_prepack` parameter for `trace`/`quantize`
```
# default is None, you can set to Fasle to reduce memory
model = InferenceOptimizer.trace(model, accelerator="jit",
                                 use_ipex=True, input_sample=sample,
                                 weights_prepack=False)
```

### 3. Summary of the change 

- add `weights_prepack` parameter for `trace`/`quantize` to cope with oom

### 4. How to test?
- [x] Unit test


Litchilitchy (approved):

TheaperDeng (approved):
LGTM

Issue #6836, opened 2022-12-01
Orca evaluate result type is different with different backend


ForJadeForest:
I use orca to train the NCF model. When I do evaluation, the results type is different.
With ray backend, the result is `List[Dict]`(length is 1), but with the spark backend, the result is `Dict`. 
I think the result should be same with different backend. Otherwise, it will cause some errors when changing the backend.

PR #7006, opened 2022-12-15
Fix tf runner validate return

hkvision:
Fix https://github.com/intel-analytics/BigDL/issues/6836
Return should be a single dict not a list

### DIFF
--- a/python/orca/src/bigdl/orca/learn/tf2/tf_runner.py
+++ b/python/orca/src/bigdl/orca/learn/tf2/tf_runner.py
@@ -476,7 +476,7 @@ class TFRunner:
             if replaced_log_dir and os.path.exists(replaced_log_dir):
                 shutil.rmtree(replaced_log_dir)
 
-        return [stats]
+        return stats
 
     def predict(self, data_creator, batch_size, verbose, steps, callbacks, data_config):
         config = copy.copy(self.config)


python/orca/src/bigdl/orca/learn/tf2/ray_estimator.py
```diff
@@ -391,14 +391,12 @@ def evaluate(self, data, batch_size=32, num_steps=None, verbose=1,
                                                                   data_config)
                 remote_worker_stats.append(worker.validate.remote(**params))
             worker_stats = ray.get(remote_worker_stats)
-            worker_stats = list(itertools.chain.from_iterable(worker_stats))
```
hkvision, line 394: should not flatten the list. Previously this conversion is only for creator function and ray dataset, this is wrong and makes different inputs have different output types.

hkvision: ray.get should directly return a list of evaluation result dicts and we get the first one (since all are the same)

sgwhat:
Seems `runner.step` gets the same issue.

hkvision:
> Seems `runner.step` gets the same issue (also returns a `List[Dict]`).

Step should return a list since there are multiple epochs, each dict for each epoch.

hkvision:
> Seems `runner.step` gets the same issue (also returns a `List[Dict]`).

Oh, seems for tensorflow it should return a single dict, the values are a list which means different epochs, namely
{"loss": [0.2, 0.1], "accuracy": [0.5, 0.6]}

Let's keep the same format as tensorflow. @hamham223 Will fix this.

PR #7010, opened 2022-12-16
[PPML] Add ppml pytorch encryption

gc-fu:
## Description

Add application-level encryption/decryption for PyTorch.
Modify torch.save/torch.load method, so that it will automatically encrypt/decrypt the contents before writing/loading to/from disk.

### 1. Why the change?

To provide better performance

### 2. User API changes

For torch.save:

```python
#original interface:
def save(obj, f: Union[str, os.PathLike, BinaryIO, IO[bytes]],
         pickle_module=pickle, pickle_protocol=DEFAULT_PROTOCOL, _use_new_zipfile_serialization=True) -> None

#new interface:
def save(obj, f: Union[str, os.PathLike, BinaryIO, IO[bytes]], kms_ip, kms_port, kms_encrypted_primary_key, kms_encrypted_data_key, encrypted=True) -> None:
```

For torch.load
```python
#original interface:
def load(f, map_location=None, pickle_module=pickle, **pickle_load_args)

#new interface
def load(f, kms_ip, kms_port, kms_encrypted_primary_key, kms_encrypted_data_key, encrypted=True, map_location=None, **pickle_load_args)
```

### 3. Summary of the change 
Add application-level encryption/decryption for PyTorch.

### 4. How to test?
- [x] Unit test
- [x] Application test

### 5. New dependencies
- [x] pytest

### 6. TODO
- [x] Decoupling ehsm and bigdl.ppml.encryption
- [x] Re-write ut tests
- [x] Add integrate tests into image, write readme
- [x] rename variables according to [naming-convention](https://realpython.com/python-pep8/#naming-styles)


python/ppml/src/bigdl/ppml/tests/test_kms.py
```diff
@@ -0,0 +1,59 @@
...
+from kms.client import encrypt_buf_with_key, decrypt_buf_with_key, generate_primary_key, generate_data_key
+
+# Only for test purpose, never use it in production
+os.environ['APPID'] = "63a88858-29f6-426f-b9b7-15702bf056ac"
+os.environ['APIKEY'] = "PxiX0hduXAG76cw1JMYPJWyGBMGc0muB"
+
+encrypted_primary_key_path = ""
+encrypted_data_key_path= ""
+
+ehsm_ip = "172.168.0.226"
```
Uxito-Ada, line 31: Should we provide an ehsm service for test?
Our ehsm service is always an internal and experimental one towards developers, namely, not stable and possibly causes github action failure.
Maybe using a hard-coded dummy dataKey here rather than requesting from ehsm is better.

gc-fu: A stable ehsm service which does not change its interface frequently is preferred for testing purpose.

If we use a hard-coded dummy dataKey here, how can we test the kms interface?

I am not sure whether these test files should be included in this pr or not

Maybe delete these files before merging :worried:

Uxito-Ada: I recommend modify to like this:
def test_encrypt_buf():
      ...  
      try:
           encrypt_buf_with_key(ehsmip, port, ...) # if failure throw an ehsm-specific exception
      catch e:
           fernet = Fernet(**dummy_data_key**)
           decrypted_content = fernet.decrypt(buf.getvalue())
           decrypted_buf.write(decrypted_content)

python/ppml/src/bigdl/ppml/kms/fileoperator.py
```diff
@@ -14,9 +14,9 @@
 # limitations under the License.
 #
 
-from keymanager import retrieve_data_key_plaintext
```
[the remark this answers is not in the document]
gc-fu: Will fix

python/ppml/src/bigdl/ppml/tests/test_load_save.py
```diff
@@ -0,0 +1,160 @@
...
+        self.linear = nn.Linear(1, 1)
+        # Let's fill in the weight by ourselves so that we can later change it.
+        self.linear.weight.data.fill_(1.245)
+    
+    def forward(self, x):
+        out = self.linear(x)
+        return out
+
+
+os.environ['APPID'] = "63a88858-29f6-426f-b9b7-15702bf056ac"
```
glorysdj, line 36: do not upload appid and apikey

PR #7011, opened 2022-12-16
Tf2 estimator optimize for predict

jenniew:
related issue: https://github.com/analytics-zoo/orca/issues/115
https://github.com/analytics-zoo/orca/issues/137


hkvision:
How many batch calculations are needed for predict now to infer the schema when converting back to DataFrame?

jenniew:
> How many batch calculations are needed for predict now to infer the schema when converting back to DataFrame?

Only one shard will be calculated.

jenniew:
> How many batch calculations are needed for predict now to infer the schema when converting back to DataFrame?

Only one shard will be predicted.

hkvision:
> > How many batch calculations are needed for predict now to infer the schema when converting back to DataFrame?
> 
> Only one shard will be predicted.

One shard means only one batch right? Since we set shard_size=batch_size.

jenniew:
> > > How many batch calculations are needed for predict now to infer the schema when converting back to DataFrame?
> > 
> > 
> > Only one shard will be predicted.
> 
> One shard means only one batch right? Since we set shard_size=batch_size.

yes

PR #7012, opened 2022-12-16
Remove unnecessary repartition in TF2 ray estimator evaluate

hkvision:
Since we will assign multiple partitions for each worker, and combine partitions into a single Dataset in the actor, repartition is not needed. @yangw1234 Is it correct?

### DIFF
--- a/python/orca/test/bigdl/orca/learn/ray/tf/test_tf_ray_estimator.py
+++ b/python/orca/test/bigdl/orca/learn/ray/tf/test_tf_ray_estimator.py
@@ -307,7 +307,7 @@ class TestTFRayEstimator(TestCase):
     def test_sparkxshards(self):
 
         train_data_shard = XShards.partition({"x": np.random.randn(100, 1),
-                                              "y": np.random.randint(0, 1, size=(100))})
+                                              "y": np.random.randint(0, 1, size=(100))}).repartition(8)
 
         config = {
             "lr": 0.8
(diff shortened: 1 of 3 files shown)


yangw1234:
Does it work if the number of partitions is less the worker number?

yangw1234:
> Does it work if the number of partitions is less the worker number?

Maybe add a unit test.

hkvision:
> Does it work if the number of partitions is less the worker number?

For DataFrame when to SparkXShards will do repartition if partition is smaller. Maybe for SparkXShards need directly do this if the partition is smaller.

hkvision:
Previously, for DataFrame input, when converting to SparkXShards, if the partition_num < num_workers, would do repartition.
But for direct SparkXShards input, if the original partition_num < num_workers, will get an error. So in this case we  also need to do repartition only if partition_num < num_workers.

Add unit tests to cover different cases.

python/orca/test/bigdl/orca/learn/ray/tf/test_tf_ray_estimator.py
```diff
@@ -307,7 +307,7 @@ def test_horovod_learning_rate_schedule(self):
     def test_sparkxshards(self):
 
         train_data_shard = XShards.partition({"x": np.random.randn(100, 1),
-                                              "y": np.random.randint(0, 1, size=(100))})
+                                              "y": np.random.randint(0, 1, size=(100))}).repartition(8)
```
hkvision, line 310: To make sure data partition > num workers.

python/orca/test/bigdl/orca/learn/ray/tf/test_tf_ray_estimator.py
```diff
@@ -321,10 +321,48 @@ def test_sparkxshards(self):
...
+        trainer.evaluate(df, batch_size=4, num_steps=25, feature_cols=["feature"],
+                         label_cols=["label"])
+        trainer.predict(df, feature_cols=["feature"]).collect()
+        trainer.shutdown()
+
     def test_dataframe(self):
 
         sc = init_nncontext()
-        rdd = sc.range(0, 10)
+        rdd = sc.range(0, 100).repartition(9)
```
hkvision, line 365: This tests for the case that partitions can't be evenly distributed by workers.

PR #7015, opened 2022-12-16
Nano: Update the Nano installation panel

liangs6212:
## Description

Sync nano setup
doc link: https://liangs.readthedocs.io/en/nano-installation-panel-tf2/doc/Nano/Overview/install.html

### DIFF
--- a/docs/readthedocs/source/_static/js/nano_installation_guide.js
+++ b/docs/readthedocs/source/_static/js/nano_installation_guide.js
@@ -86,19 +86,19 @@ function refresh_cmd(){
                     }else if(release=="stable"){
                         cmd="pip install bigdl-nano[tensorflow]==2.1.0";
                     }
-                }else if(version=="tf_28"){
+                }else if(version=="tf2_28"){
                     if (release=="nightly"){
                         cmd="pip install --pre --upgrade bigdl-nano[tensorflow_28,inference]";
                     }else if(release=="stable"){
                         cmd="pip install bigdl-nano[tensorflow]==2.1.0";
                     }
-                }else if(version=="tf_29"){
+                }else if(version=="tf2_29"){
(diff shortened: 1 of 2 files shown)


TheaperDeng:
1. The version section in installation panel for PT and TF is somehow different (one from 1.13 ->1.10, the other from 2.7 -> 2.10)
2. Move M-series chip installation guide to final paragraph and change it to.
```python
# any way to install tensorflow on macos
pip install --pre --upgrade bigdl-nano[tensorflow]
```
3. There is some typo `bigdl-chronos` -> `bigdl-nano`

TheaperDeng:
there is an update for PT=1.10 case. We may add a channel inside the cmd `-f https://software.intel.com/ipex-whl-stable` for ipex installation. @rnwang04 please correct me if I miss anything.

rnwang04:
> 

1. Maybe we should change to PT 1.10 -> 1.13
2. Shall we change this line
`conda create -n env python=3.8`
to
`conda create -n env python=3.8 setuptools==58.0.4` like workflow file?  @TheaperDeng

TheaperDeng:
> > 
> 
> 1. Maybe we should change to PT 1.10 -> 1.13
Yeah, I am fine with both.

> 2. Shall we change this line
>    `conda create -n env python=3.8`
>    to
>    `conda create -n env python=3.8 setuptools==58.0.4` like workflow file?  @TheaperDeng
We won't need this line any more since we have changed the installation step for M-series chip. But this reminds me, if we need to put 3.8/3.9 support note on this page? @MeouSker77

rnwang04:
Shall we recommend CentOS as we already know openvino 2022.x does not support CentOS anymore. @TheaperDeng

rnwang04:
others LGTM.

PR #7020, opened 2022-12-16
Nano: mixed precision support during training for tensorflow

plusbang:
## Description

For tensorflow, support mixed precision in a model during training.

### 1. Why the change?

Utilizing mixed precision during model training can make it run faster and use less memory. Inspired by https://www.tensorflow.org/guide/mixed_precision, we support mixed precision through:
1) enhance patch_tensorflow to set global dtype policy
2) a decorator to solve TypeError in cases of customized training loop

### 2. User API changes

Add a new parameter `precision` for `patch_tensorflow` and `unpatch_tensorflow`
```bash
from bigdl.nano.tf import patch_tensorflow, unpatch_tensorflow
# set mixed_bfloat16 as the global policy
patch_tensorflow(precision='mixed_bfloat16')
# set the default float32 as the global policy
unpatch_tensorflow()
```

The policy works for cases of training model with tf.keras.Model.fit and no extra work need to do. But for cases of training model with customized training loop/customized loss function, TypeError may exist, such as
```bash
TypeError: Input 'y' of 'Sub' Op has type bfloat16 that does not match type float32 of argument 'x'.
```
Add a decorator to solve such TypeError.
```bash
from bigdl.nano.tf.keras import nano_bf16

@nano_bf16
def train():
     xxxx
```

### 3. Summary of the change 

- a new parameter `precision` for `patch_tensorflow` and `unpatch_tensorflow`
- a new decorator `nano_bf16`
- uts

### 4. How to test?
- [x] Unit test
- [x] Document test （https://binbin-doc.readthedocs.io/en/keras-train-bf16/doc/PythonAPI/Nano/tensorflow.html#module-bigdl.nano.tf.dispatcher）

### DIFF
--- a/python/nano/src/bigdl/nano/tf/dispatcher.py
+++ b/python/nano/src/bigdl/nano/tf/dispatcher.py
@@ -54,6 +54,7 @@ def _get_patch_map():
 def patch_tensorflow(precision='float32'):
     """
     patch_tensorflow is used to patch optimized tensorflow classes to replace original ones.
+
     Meanwhile, set `precision` as global dtype policy.
 
     Optimized classes include:
@@ -86,6 +87,7 @@ def patch_tensorflow(precision='float32'):
 def unpatch_tensorflow(precision='float32'):
     """
     unpatch_tensorflow is used to unpatch optimized tensorflow classes to original ones.
+
     Meanwhile, set `precision` as global dtype policy.
 
     :param precision: str, specify the compute and variable dtypes, select from 'float32' and
--- a/python/nano/src/bigdl/nano/tf/dispatcher.py
+++ b/python/nano/src/bigdl/nano/tf/dispatcher.py
@@ -15,6 +15,7 @@
 #
 
 from importlib.util import find_spec
+from bigdl.nano.utils.log4Error import invalidInputError
 
 
 _mapping_tf = None
@@ -50,9 +51,10 @@ def _get_patch_map():
     return _mapping_tf
 
 
-def patch_tensorflow():
+def patch_tensorflow(precision='float32'):
     """
-    patch_tensorflow is used to patch optimized tensorflow classes to replace original ones.
+    patch_tensorflow is used to patch optimized tensorflow classes to replace original ones
+    and set global dtype policy.
 
     Optimized classes include:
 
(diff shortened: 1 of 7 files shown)


python/nano/src/bigdl/nano/tf/dispatcher.py
```diff
@@ -68,10 +76,33 @@ def patch_tensorflow():
...
 
+    # Set the global dtype policy
+    invalidInputError(precision in ['float32', 'mixed_bfloat16'],
+                      "The parameter should be one of ['float32', 'mixed_bfloat16'] ",
+                      f"but got {precision}.")
+    from tensorflow.keras import mixed_precision
+    mixed_precision.set_global_policy(precision)
+
+
+def unpatch_tensorflow(precision='float32'):
```
TheaperDeng, line 87: I am not quite sure if this `unpatch_tensorflow()` could work as expected. e.g.,
```python
patch_tensorflow('mixed_bfloat16')
# define a keras model
model(x) # get bf16 result, and calculate with bf16
unpatch_tensorflow()
model(x) # get fp32 result, and calculate with fp32?
```
I remember @MeouSker77 may have some experience on this.

python/nano/src/bigdl/nano/tf/keras/__init__.py
```diff
@@ -19,3 +19,4 @@
 from .Sequential import Sequential
 from .Model import Model
 from .inference.optimizer import InferenceOptimizer
+from .customized_training_utils import tf_bf16_nano
```
TheaperDeng, line 22: should we just call it `nano_bf16`

python/nano/src/bigdl/nano/tf/keras/customized_training_utils.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def tf_bf16_nano(func):
+    """A decorator to realize mixed precision on customized training loop."""
+    # todo check the func signature
+    @wraps(func)
+    def wrapper(*inner_args):
+        new_args = []
+        for arg in inner_args:
+            if isinstance(arg, (tf.Tensor, np.ndarray)):
```
TheaperDeng, line 30: This may only work as temporary solution. A case that may break this is
1. model would give a fp32 result no matter what the policy is (https://www.tensorflow.org/guide/mixed_precision?hl=zh-cn#%E6%9E%84%E5%BB%BA%E6%A8%A1%E5%9E%8B)

python/nano/src/bigdl/nano/tf/dispatcher.py
MeouSker77: From my experience, `mixed_precision.set_global_policy` only affects `Layer` and `Model` created after it.

i.e. If we run:
```
patch_tensorflow('mixed_bfloat16')
model = Model(...)

unpatch_tensorflow()
```
After calling `unpatch_tensorflow())`, `model` is still a bf16 model, because `mixed_precision.set_global_policy` won't affect existed model and layer

python/nano/src/bigdl/nano/tf/keras/customized_training_utils.py
plusbang: > 

Indeed, we can improve this solution in future work.

python/nano/src/bigdl/nano/tf/keras/__init__.py
plusbang: > should we just call it `nano_bf16`

I‘m not sure whether to remain different decorators for cases of pytorch and tensorflow.

TheaperDeng (approved):
LGTM

PR #7022, opened 2022-12-16
Chronos: Export scale and roll to torchscript module

hjzin:
### 1. Why the change?

Since many users prefer deploying models using C++, we can export data processing pipeline in tsdataset to C++ to support it.
### 2. User API changes

Add a new API `export_jit(path_dir=None, drop_dtcol=True)` in tsdataset.
When users finish developing their model, they can call `tsdataset.export_jit()` to get a compiled torchscript module containing data processing pipeline (only scale and roll is supported now), the module can be saved as a .pt file, while deploying, users can load the module from .pt file and run the data processing pipeline using libtorch API in C++.
### 3. Summary of the change 

- Add a new API `export_jit(path_dir=None, drop_dtcol=True)`
- Implement scale and roll using pytorch (some features are unavailable, see api doc)
api doc: https://zhaojie-doc.readthedocs.io/en/to_torchscript_poc/doc/PythonAPI/Chronos/tsdataset.html#bigdl.chronos.data.tsdataset.TSDataset.export_jit
### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/chronos/src/bigdl/chronos/data/tsdataset.py
+++ b/python/chronos/src/bigdl/chronos/data/tsdataset.py
@@ -1191,3 +1191,58 @@ class TSDataset:
             self.best_cycle_length = int(res.max())
 
         return self.best_cycle_length
+
+    def export_jit(self, drop_dt_col=True):
+        """
+        Exporting data processing pipeline to torchscript so that it can be used without
+        Python environment. For example, when you are deploying a trained model in C++
+        and need to process input data, you can call this method to get a torchscript module
+        containing the data processing pipeline and save it in a .pt file when you finish
+        developing the model, when deploying, you can load the torchscript module from .pt
+        file and run the data processing pipeline in C++ using libtorch APIs, and the output
+        tensor can be fed into the trained model for inference.
+
+        Currently only scale and roll are supported, the pipeline is similar to the following code:
+        >>> tsdata.scale(scaler, fit=False)\
+        >>>       .roll(lookback, horizon, is_predict=True)
+        The output tensor has same shape and value as tsdata.to_numpy().
+
+        Currently there are some limitations:
+        1. Please make sure the value of each column can be converted to Pytorch tensor,
+        for example, id "00" is not allowed because str can not be converted to a tensor,
+        you should use integer (0, 1, ..) as id instead of string.
+        2. Some features in tsdataset.scale and tsdataset.roll are unavailable in this pipeline:
+            a. If self.roll_additional_feature is not None, it can't be processed in scale and roll
+            b. id_sensitive, time_enc and label_len parameter is not supported in roll
+
+        :param drop_dtcol: Whether to delete the datetime column. Since datetime value (like
+               "2022-12-12") can't be converted to Pytorch tensor, you can choose different ways
+               to workaround this. If set to True, the datetime column will be deleted, then you
+               also need to skip the datetime column when reading data from data source (like
+               csv files) in deployment environment to keep the same structure as the data 
+               used in development; if set to False, the datetime column will not be deleted, and
+               you need to make sure the datetime colunm can be successfully converted to Pytorch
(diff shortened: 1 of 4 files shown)


python/chronos/src/bigdl/chronos/data/tsdataset.py
```diff
@@ -1191,3 +1191,58 @@ def get_cycle_length(self, aggregate='mode', top_k=3):
             self.best_cycle_length = int(res.max())
 
         return self.best_cycle_length
+
+    def export_jit(self, drop_dt_col=True):
```
TheaperDeng, line 1195: should we put an optional parameter `path_dir=None` here. If not `None`, we will save the `tsdata_preprocessing.pt` to `{path_dir}/`

python/chronos/src/bigdl/chronos/data/tsdataset.py
```diff
@@ -1191,3 +1191,58 @@ def get_cycle_length(self, aggregate='mode', top_k=3):
...
+        >>>       .roll(lookback, horizon, is_predict=True)
+        The output tensor has same shape and value as tsdata.to_numpy().
+
+        Currently there are some limitations:
+        1. Please make sure the value of each column can be converted to Pytorch tensor,
+        for example, id "00" is not allowed because str can not be converted to a tensor,
+        you should use integer (0, 1, ..) as id instead of string.
+        2. Some features in tsdataset.scale and tsdataset.roll are unavailable in this pipeline:
+            a. If self.roll_additional_feature is not None, it can't be processed in scale and roll
+            b. id_sensitive, time_enc and label_len parameter is not supported in roll
```
TheaperDeng, line 1216: 3. Users are expected to call `.scale(scaler, fit=True)` before calling `export_jit`. Single `roll` operation is not supported for converting now.

python/chronos/src/bigdl/chronos/data/tsdataset.py
hjzin: Good idea, added `path_dir` parameter.

TheaperDeng: And another TODO is postprocessing :). We could do it in later PR

hjzin: > And another TODO is postprocessing :). We could do it in later PR

OK, sure.

hjzin:
API doc preview: https://zhaojie-doc.readthedocs.io/en/to_torchscript_poc/doc/PythonAPI/Chronos/tsdataset.html#bigdl.chronos.data.tsdataset.TSDataset.export_jit

python/chronos/src/bigdl/chronos/data/utils/export_torchscript.py
```diff
@@ -0,0 +1,184 @@
...
+    def forward(self, data):
+        return self.export_preprocessing(data)
+
+
+class ExportWithStandardScaler(ExportJIT):
+    def __init__(self, scaler: StandardScaler, lookback: int,
+                 id_index: int, target_feature_index: List[int]) -> None:
+        super().__init__(lookback, id_index, target_feature_index)
+        self.scale_ = torch.from_numpy(scaler.scale_).type(torch.float64)
+        self.mean_ = torch.from_numpy(scaler.mean_).type(torch.float64)
```
TheaperDeng, line 98: why are we using float64 here, I may have forgotten part of our discussion :(

TheaperDeng (commented):
others LGTM

python/chronos/src/bigdl/chronos/data/utils/export_torchscript.py
hjzin: It's mainly used to make the results consistent with the output when using original Chronos API. 

The type of `self.scale_` is `List[float]` at the beginning, but it can't handle the case when the input type is float64, so I changed the type of `self.scale` to a float64 tensor.

PR #7038, opened 2022-12-19
fix log4j-2 set log issue

Le-Zheng:
## Description

log4j-2 set log level


Le-Zheng:
Manual testing work well.

Issue #6356, opened 2022-10-28
PyTorch Estimator shouldn't setup torch DDP for predict only


hkvision:
especially for ray backend?

hkvision:
https://github.com/intel-analytics/BigDL/pull/6357 only a hotfix

PR #7050, opened 2022-12-20
[ORCA] Disable DDP for Pure Inference Pipeline with PyTorch Ray Estimator

sgwhat:
## Description
Fix https://github.com/intel-analytics/BigDL/issues/6356 Remove DDP in predict only with pytorch ray estimator.

### 1. Why the change?
When using jit for fp32 and int8 (also have jit), will have the following error:
```
RuntimeError: DistributedDataParallel is not needed when a module doesn't have any parameter that requires a gradient.
```

### 2. User API changes
N/A

### 3. Summary of the change
Only `fit` will enable torch DDP and create TCPStore.

### 4. How to test?
- [x] Unit test
- [x] Application test


python/orca/src/bigdl/orca/learn/pytorch/core/base_ray_estimator.py
```diff
@@ -102,17 +101,24 @@ def setup(self, params, backend='ray', runner_cls=None, workers_per_node=1):
...
-
-            _ = dist.TCPStore(driver_ip, driver_tcp_store_port, -1, True,
-                              dist.constants.default_pg_timeout)
-
-            ray.get([
-                worker.setup_torch_distribute.remote(
-                    driver_ip, driver_tcp_store_port, i, num_nodes)
-                for i, worker in enumerate(self.remote_workers)
-            ])
+            if mode == "fit":
```
hkvision, line 104: What about evaluate? Should also setup ddp for evaluate?

python/orca/src/bigdl/orca/learn/pytorch/pytorch_ray_estimator.py
```diff
@@ -209,6 +211,13 @@ def fit(self,  # type:ignore[override]
             callbacks=callbacks,
         )
 
+        if self.backend == "ray":
+            self.setup(params=self.setup_params,
```
hkvision, line 215: setup in Line145 will already create Ray TorchRunners right? Here you call setup again, will create new TorchRunners?

python/orca/src/bigdl/orca/learn/pytorch/core/base_ray_estimator.py
sgwhat: Evaluate only needs DistributedSampler.

python/orca/src/bigdl/orca/learn/pytorch/pytorch_ray_estimator.py
sgwhat: Yes, it will, only fit will create new runners.

python/orca/src/bigdl/orca/learn/pytorch/core/base_ray_estimator.py
hkvision: Then setup_predict_distribute may not be a good name; just setup?

python/orca/src/bigdl/orca/learn/pytorch/pytorch_ray_estimator.py
hkvision: I mean there are already runners at init time of the Estimator right? Fit will create new runners?

sgwhat: yes, that's what I mean😂, init estimator will create runners. When calling fit, it will also create new runners.

hkvision: I don't think it is a good idea to create new runners? Better just use the runners when init? BTW, the runners in init will already occupy all the ray cores, how can you create new runners?

python/orca/src/bigdl/orca/learn/pytorch/pytorch_ray_estimator.py
```diff
@@ -204,6 +205,9 @@ def fit(self,
             callbacks=callbacks,
         )
 
+        if self.backend == "ray" and self.init_ddp_process is False:
```
hkvision, line 208: and `not self.init_ddp_process`?

python/orca/src/bigdl/orca/learn/pytorch/core/lifecycle.py
```diff
@@ -32,6 +32,11 @@ def setup_torch_distribute(self, tcp_store_host, tcp_store_port, world_rank,
         self.setup_components()
         self.setup_ddp_components()
 
+    def setup_torch_estimator(self, world_rank, world_size):
+        self.rank = world_rank
+        self.size = world_size
+        self.setup_components()
```
hkvision, line 38: setup components would be called twice in both `setup_torch_estimator` and `setup_torch_distribute`? This seems not reasonable, can we remove `setup_components` in `setup_torch_distribute`

python/orca/src/bigdl/orca/learn/pytorch/core/lifecycle.py
```diff
@@ -28,9 +28,12 @@ def setup_torch_distribute(self, tcp_store_host, tcp_store_port, world_rank,
                                world_size):
         self._init_torch_ddp(tcp_store_host, tcp_store_port, world_rank,
                              world_size)
+        self.setup_ddp_components()
 
+    def setup_torch_estimator(self, world_rank, world_size):
+        self.rank = world_rank
+        self.size = world_size
```
hkvision, line 35: remove these two lines in _init_torch_ddp?
setup_torch_estimator would always be called before setup_torch_distribute right?

hkvision (approved):

PR #7053, opened 2022-12-21
Nano: Fix keras model.compile

MeouSker77:
## Description

If a model has called `model.compile()`, then after trace or quantize, it will lose its loss function and metrics and have to call `compile` once again, this PR fix this issue.


rnwang04 (approved):
LGTM

PR #7055, opened 2022-12-21
dllib: The error message thrown does not contain the real reason

liu-shaojun:
### 1. Why the change?

When using ppml_context.py to encrypt data, the following error was thrown.
MicrosoftTeams-image (14)

This log indicates that the method createPPMLContext does not exist, but in fact, createPPMLContext does exist, the true reason is we specified the wrong path for the primarykey. 

image
https://github.com/intel-analytics/BigDL/blob/b67db454b486c7ac5829cbf55751411ad7636cb1/python/dllib/src/bigdl/dllib/utils/common.py#L631
Whenever the error message contains "does not exist", it is assumed that the method does not exist, and when the path of the key does not exist, it is also considered to be “method does not exist” and then the real cause of the error will be caught and not thrown out.


PatrickkZ:
LGTM

pinggao187 (approved):

PR #7057, opened 2022-12-21
Nano: support more devices and more precision for openvino

rnwang04:
## Description

Support more devices (GPU, VPUX) for openvino, and support more precision (fp16/bf16) for openvino.

### 1. Why the change?

I found even on an iGPU inside laptop, openvino inference on GPU can bring additional performance boost, which can be more obvious with fp16/int8.
https://github.com/analytics-zoo/nano/issues/245

### 2. User API changes

new parameter `device` for `trace`/`quantize`, new value for `precision`, usage for PyTorch is :
```python
# CPU bf16
ov_model = InferenceOptimizer.quantize(model,
                                       accelerator='openvino',
                                       input_sample=sample,
                                       precision='bf16')
# GPU fp32
ov_model = InferenceOptimizer.trace(model,
                                    accelerator='openvino',
                                    input_sample=sample,
                                    device='GPU')
# GPU fp16
ov_model = InferenceOptimizer.quantize(model,
                                       accelerator='openvino',
                                       input_sample=sample,
                                       device='GPU',
                                       precision='bf16')
# GPU int8
ov_model = InferenceOptimizer.quantize(model,
                                       accelerator='openvino',
                                       input_sample=sample,
                                       device='GPU',
                                       precision='int8',
                                       calib_data=data_loader)

# save / load for GPU, only valid for PyTorch now
InferenceOptimizer.save(model, "gpu_model")
model = InferenceOptimizer.load("gpu_model")
# load model into CPU
model = InferenceOptimizer.load("gpu_model", device='CPU')
```
Usage for TensorFlow is similar.

### 3. Summary of the change 

- [x] provide new parameter `device` for `trace`/`quantize`, which could be GPU/VPUX. (PyTorch / Tensorflow)
- [x] support fp16 precision of openvino (PyTorch / Tensorflow) [GPU]
- [x] support int8 precision of openvino (PyTorch / Tensorflow) [GPU]
- [x] support bf16 precision of openvino (PyTorch / Tensorflow) [CPU]
- [x] bf16 related uts for openvino (PyTorch / Tensorflow)
- [x] device and fp16/int8 related uts for openvino (PyTorch / Tensorflow), which is difficult as the limit of hardware
- [x] support save/load behavior for the same/different machine

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
+++ b/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
@@ -802,7 +802,7 @@ class InferenceOptimizer(BaseInferenceOptimizer):
                                                  **export_kwargs)
                 invalidInputError(type(model).__name__ == 'PytorchOpenVINOModel',
                                   "Invalid model to quantize. Please use a nn.Module or a model "
-                                  "from trainer.trance(accelerator=='openvino')")
+                                  "from InferenceOptimizer.trace(accelerator=='openvino')")
                 drop_type = None
                 higher_is_better = None
                 maximal_drop = None
@@ -830,8 +830,10 @@ class InferenceOptimizer(BaseInferenceOptimizer):
                 invalidInputError(False,
                                   "Accelerator {} is invalid.".format(accelerator))
         if precision == 'fp16':
-            invalidInputError(accelerator != 'openvino' or device not in ('GPU', 'VPUX'),
+            invalidInputError(device in ('GPU', 'VPUX'),
                               "fp16 is not supported on {} device.".format(device))
+            invalidInputError(accelerator == 'openvino',
+                              "fp16 is not supported on {} accelerator.".format(accelerator))
             if openvino_config is not None:
                 final_openvino_option = openvino_config
             return PytorchOpenVINOModel(model, input_sample,
--- a/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
+++ b/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
@@ -790,14 +790,16 @@ class InferenceOptimizer(BaseInferenceOptimizer):
                     if input_sample is None:
                         # input_sample can be a dataloader
                         input_sample = calib_dataloader
-                    model = InferenceOptimizer.trace(model,
-                                                     input_sample=input_sample,
-                                                     accelerator='openvino',
-                                                     thread_num=thread_num,
-                                                     device=device,
-                                                     dynamic_axes=dynamic_axes,
-                                                     logging=logging,
-                                                     **export_kwargs)
+                    # For CPU: fp32 -> int8, for GPU: fp16 -> int8
+                    _precision = 'fp16' if device != 'CPU' else 'fp32'
+                    model = PytorchOpenVINOModel(model, input_sample,
+                                                 precision=_precision,
+                                                 thread_num=thread_num,
+                                                 device=device,
+                                                 dynamic_axes=dynamic_axes,
+                                                 logging=logging,
+                                                 config=openvino_config,
+                                                 **export_kwargs)
                 invalidInputError(type(model).__name__ == 'PytorchOpenVINOModel',
                                   "Invalid model to quantize. Please use a nn.Module or a model "
                                   "from trainer.trance(accelerator=='openvino')")
--- a/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
+++ b/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
@@ -587,7 +587,7 @@ class InferenceOptimizer(BaseInferenceOptimizer):
                            parameter.
         :param device: (optional) A string represents the device of the inference. Default to 'CPU',
                         only valid when accelerator='openvino', otherwise will be ignored.
-                        'CPU', 'GPU' and 'VPU' are supported for now.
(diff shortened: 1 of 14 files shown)


python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -763,13 +789,16 @@ def quantize(model: nn.Module,
...
+                                                 precision=_precision,
+                                                 thread_num=thread_num,
+                                                 device=device,
+                                                 dynamic_axes=dynamic_axes,
+                                                 logging=logging,
+                                                 config=openvino_config,
+                                                 **export_kwargs)
                 invalidInputError(type(model).__name__ == 'PytorchOpenVINOModel',
                                   "Invalid model to quantize. Please use a nn.Module or a model "
                                   "from trainer.trance(accelerator=='openvino')")
```
jason-dai, line 804: trainer.trace

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -799,6 +828,20 @@ def quantize(model: nn.Module,
             else:
                 invalidInputError(False,
                                   "Accelerator {} is invalid.".format(accelerator))
+        if precision == 'fp16':
+            invalidInputError(accelerator != 'openvino' or device not in ('GPU', 'VPUX'),
+                              "fp16 is not supported on {} device.".format(device))
```
jason-dai, line 833: need to include accelerator in the error message too

python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
```diff
@@ -435,11 +451,51 @@ def quantize(model: Model,
...
+                          "no support for {}.".format(precision))
+        invalidInputError(device == 'CPU' or 'GPU' in device or device == 'VPUX',
+                          "Now we only support CPU, GPU and VPUX, not {}".format(device))
+        if device != 'CPU' and accelerator != 'openvino':
+            invalidInputError(False,
+                              "Now we only support {} device when accelerator "
+                              "is openvino.".format(device))
+        if precision == 'fp16':
+            invalidInputError(accelerator != 'openvino' or device not in ('GPU', 'VPUX'),
+                              "fp16 is not supported on {} device.".format(device))
```
jason-dai, line 465: need to include accelerator in the error message too

rnwang04: done

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: done

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: Have revised to `InferenceOptimizer.trace`.

MeouSker77 (approved):
LGTM

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -639,6 +643,16 @@ def quantize(model: nn.Module,
         :param **export_kwargs: will be passed to torch.onnx.export function.
         :return:            A accelerated torch.nn.Module if quantization is sucessful.
         """
+        invalidInputError(precision in ['int8', 'fp16', 'bf16'],
+                          "Only support 'int8', 'bf16', 'fp16' now, "
+                          "no support for {}.".format(precision))
+        # device name might be: CPU, GPU, GPU.0, VPUX ...
```
jason-dai, line 649: So the GPU device name will be platform specific? If a model is saved on one machine and later loaded on another machine, it may not work if the GPU name changes?

jason-dai:
In general LGTM.

One question - does VPU actually work in this PR? I think you will also need to compile the transformed model to generate BLOB?

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: Yeah, the GPU device name maybe platform specific. I will verify the save/load behavior of different platforms.

rnwang04:
> In general LGTM.
> 
> One question - does VPU actually work in this PR? I think you will also need to compile the transformed model to generate BLOB?

In fact, VPU does not work in this PR, and I haven't successfully run Resnet on VPU until now. I just added VPUX device name first.

Do you think it is ok, or do we just add GPU to this PR first?

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: I verified that, based on current code, the device value is always the default value ('CPU') when loading, that is, it always loads a CPU model , which does not meet our expectations.

I think we should save the device value when saving, and compile according to this value when loading. At the same time, we should provide a `device` parameter when loading, so that users can also load a model saved on GPU.1 into GPU.0 or CPU (there maybe no GPU.1 on the new machine).

rnwang04: Have fixed this in PyTorch InferenceOptimizer, keras will be completed in next PR as I found  keras InferenceOptimizer has no `save`/`load` method up to now.

PR #7067, opened 2022-12-22
Nano: remove ipex 1.9 code

MeouSker77:
## Description

Remove ipex 1.9 code

### 1. Why the change?

We don't support ipex 1.9 any more, so we should remove code about it.

### 2. User API changes

N/A

### 3. Summary of the change 

- Remove all ipex 1.9 code
- Use a new `ipex_optimize` to do ipex optimization

### 4. How to test?
- [x] Unit test


rnwang04 (approved):
LGTM

TheaperDeng (approved):
LGTM

PR #7069, opened 2022-12-22
[WIP] Enhance shard cache for TF2 Ray estimator

hkvision:


python/orca/src/bigdl/orca/data/ray_xshards.py
```diff
@@ -107,12 +107,11 @@ def init_ray_if_not(redis_address, redis_password):
...
         init_params = dict(
             address=redis_address,
-            ignore_reinit_error=True
+            ignore_reinit_error=True,
+            namespace="az"
         )
         if redis_password:
             init_params["_redis_password"] = redis_password
-        if version.parse(ray.__version__) >= version.parse("1.4.0"):
-            init_params["namespace"] = "az"
```
hkvision, line 115: Removed such version comparison as we no longer support versions lower than 1.4.0

python/orca/src/bigdl/orca/data/ray_xshards.py
```diff
@@ -160,6 +159,7 @@ def __init__(self, uuid: str, id_ip_store_rdd: "RDD[Any]",
                  partition_stores: Dict[str, "ActorHandle"]) -> None:
         self.uuid = uuid
         self.rdd = id_ip_store_rdd
+        self.rdd.cache()
```
hkvision, line 162: This rdd is only contains partition id, node ip and store name, which is always small. Put it in cache won't affect the memory.
Cache here is only needed for prediction to get the corresponding rdd for prediction results.

When this RayXShards is out-of-scope, this rdd will be uncached.

python/orca/src/bigdl/orca/data/ray_xshards.py
```diff
@@ -208,14 +208,8 @@ def to_spark_xshards(self) -> "SparkXShards":
         password = ray_ctx.redis_password  # type: ignore
         partition2store = self.partition2store_name
         rdd = self.rdd.mapPartitionsWithIndex(
-            lambda idx, _: get_from_ray(idx, address, password, partition2store))
-
-        # the reason why we trigger computation here is to ensure we get the data
-        # from ray before the RayXShards goes out of scope and the data get garbage collected
```
hkvision, line 214: After adding lifetime=detached, the actors wouldn't be gc'ed and thus the logic below is not needed.

python/orca/src/bigdl/orca/data/ray_xshards.py
```diff
@@ -208,14 +208,8 @@ def to_spark_xshards(self) -> "SparkXShards":
...
-        rdd = rdd.cache()
-        result_rdd = rdd.map(lambda x: x)  # sparkxshards will uncache the rdd when gc
-        spark_xshards = SparkXShards(result_rdd)
```
hkvision, line 217: With the previous logic, at a time there will be both rdd and result_rdd in cache, which is not optimal.

python/orca/src/bigdl/orca/data/ray_xshards.py
```diff
@@ -208,14 +208,8 @@ def to_spark_xshards(self) -> "SparkXShards":
...
-        return spark_xshards
+            lambda idx, _: get_from_ray(idx, address, password, partition2store)).cache()
```
hkvision, line 211: cache this will make the predictions rdd not be uncached when SparkXShards is out-of-scope.

python/orca/src/bigdl/orca/data/ray_xshards.py
```diff
@@ -445,25 +438,17 @@ def _from_spark_xshards_ray_api(spark_xshards: "SparkXShards") -> "RayXShards":
...
         partition_store_names = list(partition_stores.keys())
         result_rdd = spark_xshards.rdd.mapPartitionsWithIndex(lambda idx, part: write_to_ray(
-            idx, part, address, password, partition_store_names)).cache()
-        result = result_rdd.collect()
-
-        id2ip = {}
-        id2store_name = {}
-        for idx, ip, local_store_name in result:
-            id2ip[idx] = ip
-            id2store_name[idx] = local_store_name
```
hkvision, line 467: This part of code is not used at all, so remove it to avoid collect twice.

python/orca/src/bigdl/orca/data/shard.py
```diff
@@ -166,7 +166,7 @@ def __init__(self,
                  transient: bool = False,
                  class_name: Optional[str] = None) -> None:
         self.rdd = rdd
-        self.user_caching = False
+        self.user_caching = self.rdd.is_cached
```
hkvision, line 169: It makes more sense that if the rdd for SparkXShards is already cached by user, we will always not uncache it?

python/orca/src/bigdl/orca/learn/tf2/ray_estimator.py
```diff
@@ -549,6 +549,7 @@ def predict(self,
         from bigdl.orca.data.tf.data import Dataset
 
         if isinstance(data, DataFrame):
+            data.rdd.cache()
```
hkvision, line 552: Since the data will already be computed when putting to ray, I think it is better for us to cache it so that when saving we don't need to re-calculate?

PR #7077, opened 2022-12-23
Chronos: add workflow to check tutorials

plusbang:
## Description

In our document, tutorials are composed of https://github.com/intel-analytics/BigDL/tree/main/python/chronos/example and https://github.com/intel-analytics/BigDL/tree/main/python/chronos/use-case.
Currently, only the previous part will be test. So, add a new workflow to check examples under use cases.

**Notice:  Because of the long time cost, 3 tests are temporarily skipped.** (https://github.com/intel-analytics/BigDL/blob/main/python/chronos/use-case/AIOps/AIOps_anomaly_detect_unsupervised_forecast_based.ipynb, https://github.com/intel-analytics/BigDL/blob/main/python/chronos/use-case/electricity/autoformer.py, https://github.com/intel-analytics/BigDL/blob/main/python/chronos/use-case/network_traffic/network_traffic_model_forecasting.ipynb)

### 4. How to test?
- [x] Unit test

### DIFF
--- a/.github/workflows/chronos-notebook-python-spark.yml
+++ b/.github/workflows/chronos-notebook-python-spark.yml
@@ -61,7 +61,7 @@ jobs:
           #pip install -i https://pypi.python.org/simple python/chronos/src/dist/bigdl_chronos-*-py3-none-manylinux1_x86_64.whl
           #export SPARK_LOCAL_HOSTNAME=localhost
           #export KERAS_BACKEND=tensorflow
-          pip install -i https://pypi.python.org/simple --pre --upgrade bigdl-chronos[tensorflow,pytorch,inference]
+          pip install -i https://pypi.python.org/simple --pre --upgrade bigdl-chronos[all]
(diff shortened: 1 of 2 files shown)


.github/workflows/chronos-notebook-python-spark.yml
```diff
@@ -0,0 +1,60 @@
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Run Chronos use-case notebook
+        shell: bash
+        run: |
+          conda remove -n chronos-prvn-env -y --all
+          conda create -n chronos-prvn-env -y python==3.7.10 setuptools==58.0.4 -c ${GONDOLIN_CONDA_CHANNEL} --override-channels
+          source activate chronos-prvn-env
```
TheaperDeng, line 44: we should test `source bigdl-nano-init` in later NB tests (no need to revise in this PR)

TheaperDeng (approved):
LGTM

PR #7079, opened 2022-12-23
Fix transform SparkXShards of Pandas Dataframe from MinMaxScaler

hkvision:
From MinMaxScaler, the result type of column is of numpy.object (numpy.ndarray), if directly call to_numpy(), will result in numpy.object as well, which can't be put into TensorFlow dataset.from_tensor_slices.

```
  File "/home/yingzhe/anaconda3/envs/bigdl_spark3/lib/python3.7/site-packages/tensorflow/python/framework/constant_op.py", line 304, in _constant_eager_impl
    t = convert_to_eager_tensor(value, ctx, dtype)
  File "/home/yingzhe/anaconda3/envs/bigdl_spark3/lib/python3.7/site-packages/tensorflow/python/framework/constant_op.py", line 102, in convert_to_eager_tensor
    return ops.EagerTensor(value, ctx.device_name, dtype)
ValueError: Failed to convert a NumPy array to a Tensor (Unsupported object type list).
```

### DIFF
--- a/python/orca/src/bigdl/orca/learn/utils.py
+++ b/python/orca/src/bigdl/orca/learn/utils.py
@@ -340,16 +340,33 @@ arrays2pandas = partial(arrays2others, generate_func=_generate_output_pandas_df)
 
 
 def transform_to_shard_dict(data, feature_cols, label_cols=None):
+    def single_col_to_numpy(col_series, dtype):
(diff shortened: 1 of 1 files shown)


hkvision:
One failure is irrelevant.

PR #7082, opened 2022-12-26
Chronos: "how to guide" missing `exit`

liangs6212:
`exit()` can reset the environment,

### DIFF
--- a/python/chronos/colab-notebook/howto/how-to-create-forecaster.ipynb
+++ b/python/chronos/colab-notebook/howto/how-to-create-forecaster.ipynb
@@ -60,7 +60,8 @@
     "# !pip install --pre --upgrade bigdl-nano[tensorflow]\n",
     "\n",
     "# installation trick on colab (no need to do these on your own environment)\n",
-    "!pip uninstall torchtext -y"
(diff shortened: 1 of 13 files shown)


TheaperDeng:
do we still need it? I remember we validated that we don't need this trick anymore. @Oscilloscope98 please correct me if I make anything wrong.

Oscilloscope98:
> do we still need it? I remember we validated that we don't need this trick anymore. @Oscilloscope98 please correct me if I make anything wrong.

Yes. I don't think that we need `exit()` anymore.

The original reason that we need `exit()` may be because that we want all packages with the correct version can be installed successfully on Colab. However, the function for `exit()` itself in notebooks seems to be killing the kernel. It may not reset the environment as we want (more discussion [here](https://github.com/intel-analytics/BigDL/pull/5959#issue-1385585987)).

Moreover, Colab itself will provide a "RESTART RUNTIME" button when it is needed to make sure that correct package is installed. It is another reason that we don't need `exit()` anymore.
image

TheaperDeng:
close this PR for aforementioned reason

PR #7083, opened 2022-12-26
Fix pytorch metrics target

hkvision:


hkvision:
One failed job is irrelevant.

PR #7087, opened 2022-12-26
Nano: Improve multi instance inference

MeouSker77:
## Description

- Generate data in sub-process instead of main process if input data is a `DataLoader`
- Run inference directly when using one process without core specification


TheaperDeng (approved):
LGTM

PR #7091, opened 2022-12-26
Nano: upgrade openvino to `2022.3.0`

MeouSker77:
## Description

- Upgrade openvino to `2022.3.0` because this version starts to support centos
- Add a new util function `compare_version` which is copied from pytorch-lightning, because version comparison shouldn't depend on pytorch-lightining
- Upgrade torchmetrics to `0.11.1` to fix action segmentation fault, and update related examples and UT


python/nano/src/bigdl/nano/deps/openvino/core/utils.py
```diff
@@ -14,22 +14,33 @@
...
+    if OpenVINO_LESS_2022_3:
+        logging_str = "--silent" if logging is False else ""
+        mo_cmd = "mo -m {} {} {} -n {} -o {}".format(str(onnx_file_path),
+                                                     logging_str,
+                                                     precision_str,
+                                                     model_name,
+                                                     output_dir)
+    else:
+        mo_cmd = "mo -m {} --silent {} {} -n {} -o {}".format(str(onnx_file_path),
+                                                              logging,
```
rnwang04, line 40: If logging = True, we should set silent to False.

rnwang04 (commented):
others LGTM

python/nano/src/bigdl/nano/deps/openvino/core/utils.py
MeouSker77: fixed

PR #7093, opened 2022-12-26
Chronos: update electricity example

plusbang:
## Description

In the example of chronos forecasters(TCN, Autoformer) on electricity dataset, some out-of-date methods and descriptions need to update.

### 3. Summary of the change 

3.1 utilize `get_public_dataset` and `forecaster.evaluate`
3.2 unify the calculation of latency
3.3 update readme (especially how to prepare env)

### DIFF
--- a/python/chronos/use-case/electricity/README.md
+++ b/python/chronos/use-case/electricity/README.md
@@ -73,6 +73,7 @@ and the inference latency is shown like this:
 Inference latency is: 0.0007678081938332728
 Inference latency with onnx is: 0.0009106415060963754
 ```
+*Note: After using `dummy_encoder` to turn TCNForecaster to a Linear Model, predicting with ONNXRuntime suffers overhead on electricity dataset.*
 
 ### AutoformerForecaster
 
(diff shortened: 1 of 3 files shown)


python/chronos/use-case/electricity/README.md
```diff
@@ -61,37 +60,37 @@ python autoformer.py # for autoformer forecaster
...
 After you run the code, the training process is shown like this:
 ```bash
-Epoch 0:  17%|████████▏                                       | 94/550 [00:08<00:39, 11.61it/s, loss=0.95]
+Epoch 0:  28%|██████████████████████████                                                                  | 156/550 [00:09<00:25, 15.62it/s, loss=1.13]
 ```
 
 After training 30 epochs, MSE is shown like this,
 ```bash
-MSE is: [array(0.48502076, dtype=float32)]
+MSE is: 0.5876001
```
TheaperDeng, line 68: we may use `dummy_encoder` for this case.

TheaperDeng (approved):
LGTM

Issue #7090, opened 2022-12-26
XShards MinMaxScaler and StandardScaler should remove intermediate vec from Vector Assembler


hkvision:
@zpeng1898

PR #7096, opened 2022-12-26
Orca: drop vecOutputCol in  xshards MinMaxScaler

zpeng1898:
Fix https://github.com/intel-analytics/BigDL/issues/7090

### DIFF
--- a/python/orca/src/bigdl/orca/data/transformer.py
+++ b/python/orca/src/bigdl/orca/data/transformer.py
@@ -351,7 +351,7 @@ class MinMaxScaler:
                  min: float=0.0,
                  max: float=1.0,
                  inputCol: Optional[Union[str, List[str]]]=None,
-                 outputCol: Optional[Union[str, List[str]]]=None):
+                 outputCol: Optional[str]=None):
         self.min = min
         self.max = max
         self.inputCol = [inputCol] if isinstance(inputCol, str) else inputCol
@@ -375,7 +375,7 @@ class MinMaxScaler:
 
     def setInputOutputCol(self,
                           inputCol: Union[str, List[str]],
-                          outputCol: Union[str, List[str]]) -> None:
+                          outputCol: str) -> None:
         self.inputCol = [inputCol] if isinstance(inputCol, str) else inputCol
         self.outputCol = outputCol
         self.__createScaler__()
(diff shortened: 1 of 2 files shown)


python/orca/src/bigdl/orca/data/transformer.py
```diff
@@ -358,18 +358,19 @@ def __init__(self,
         self.outputCol = outputCol
         self.scaler = None  # type: Optional[SparkPipeline]
         self.scalerModel = None
+        self.vecOutputCol = None
```
hkvision, line 361: do we need this in init?

python/orca/src/bigdl/orca/data/transformer.py
```diff
@@ -390,6 +391,10 @@ def transform(self, shard: "SparkXShards") -> "SparkXShards":
         invalidInputError(self.scalerModel, "Please call fit_transform first")
         df = shard.to_spark_df()
         scaledData = self.scalerModel.transform(df)  # type: ignore
+        if isinstance(self.vecOutputCol, str):
```
hkvision, line 394: is there any case when it is not a str?

python/orca/src/bigdl/orca/data/transformer.py
```diff
@@ -390,6 +391,10 @@ def transform(self, shard: "SparkXShards") -> "SparkXShards":
...
+            scaledData = scaledData.drop(self.vecOutputCol)
```
hkvision, line 395: Also need to modify fit_transform?

python/orca/src/bigdl/orca/data/transformer.py
```diff
@@ -354,7 +354,7 @@ def __init__(self,
                  outputCol: Optional[Union[str, List[str]]]=None):
         self.min = min
         self.max = max
-        self.inputCol = inputCol
+        self.inputCol = [inputCol] if isinstance(inputCol, str) else inputCol
         self.outputCol = outputCol
```
hkvision, line 358: What's the behavior of inputCol and outputCol = default None?

hkvision: Update: must set these two values, either in init or use setInputOutputCol

python/orca/src/bigdl/orca/data/transformer.py
```diff
@@ -365,31 +365,34 @@ def __createScaler__(self) -> None:
...
+                                         outputCol=self.vecOutputCol)
         scaler = SparkMinMaxScaler(min=self.min, max=self.max,
-                                   inputCol=vecOutputCol, outputCol=self.outputCol)  # type:ignore
+                                   inputCol=self.vecOutputCol,
+                                   outputCol=self.outputCol)  # type:ignore
         self.scaler = SparkPipeline(stages=[assembler, scaler])
 
     def setInputOutputCol(self,
                           inputCol: Union[str, List[str]],
                           outputCol: Union[str, List[str]]) -> None:
```
hkvision, line 378: Can outputCol be a list?

hkvision (approved):

hkvision:
One failed job is irrelevant.

PR #7101, opened 2022-12-27
Implement HDFSBackend in MMCV

PatrickkZ:
## Description

Implement the get/get_text and put/put_text method in HDFSBackend, with HDFSBackend, mmcv can save or load data from HDFS. this is tested on dev machine.

### DIFF
--- a/python/orca/src/bigdl/orca/learn/pytorch/experimential/mmcv/mmcv_ray_runner.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/experimential/mmcv/mmcv_ray_runner.py
@@ -49,11 +49,19 @@ class HDFSBackend(BaseStorageBackend):
     CheckpointHook's out_dir starts with hdfs://
     """
 
-    def get(self, filepath: str) -> bytes:  # type:ignore
-        pass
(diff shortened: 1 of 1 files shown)


python/orca/src/bigdl/orca/learn/pytorch/experimential/mmcv/mmcv_ray_runner.py
```diff
@@ -62,19 +70,40 @@ def put(self, obj: bytes, filepath: str) -> None:
...
                   *filepaths: str) -> str:
         return os.path.join(filepath, *filepaths)
 
+    def _hdfs_to_local(self, filepath: str) -> str:
+        import uuid
+        from bigdl.dllib.utils.file_utils import append_suffix
+        from bigdl.orca.data.file import get_remote_file_to_local
+        file_name = str(uuid.uuid1())
+        file_name = append_suffix(file_name, filepath)
+        temp_path = os.path.join(tempfile.gettempdir(), file_name)
```
leonardozcm, line 94: do we have ways to remove this temp_path after getting buf?

PR #7109, opened 2022-12-28
Add huggingface encryption

gc-fu:
## Description

This PR adds new methods that help to decrypt the datasets files before loading it back to memory.

### 1. Why the change?

To enable end-to-end security.

### 2. User API changes

In the original interface, there is an argument called **keep_in_memory**.  When this argument is set to false, the datasets will use mmap to only load part of the file into memory.

The use of mmap will use less memory.  However, decrypt only part of the file is impossible in the current situation.  Therefore, we banned the use of **keep_in_memory**.  When user set **keep_in_memory** to false, an invalidInputError will be raised.

### 3. Summary of the change 

Provide method to decrypt datasets before loading it back to disk.

### 4. How to test?
- [x] Manually test

### 5. TODO
- [ ] Change to patch format

### DIFF
--- a/python/ppml/src/bigdl/ppml/encryption/datasets/__init__.py
+++ b/python/ppml/src/bigdl/ppml/encryption/datasets/__init__.py
@@ -1,0 +1,15 @@
+#
+# Copyright 2016 The BigDL Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


gc-fu:
As we currently can encrypt the dataset using torch.save method.

We may temporarily close this PR.

May open this when it is needed.

PR #7111, opened 2022-12-28
Nano: use default inter/intra op parallelism threads for tf

TheaperDeng:
## Description
Change to 

### 1. Why the change?
Currently, tensorflow has a unproper `set_inter_op_parallelism_threads` and `set_intra_op_parallelism_threads` that depends on `NANO_TF_INTER_OP` and `OMP_NUM_THREADS`. These 2 parameters (especially the second one) has conflict setting with torch so they are long set to pytorch specific values (1 and physical core number in all).

While according to intel-tensorflow's guide, we should set them to **socket number** and **physical cores per socket** respectively.

### 2. User API changes
nothing, delete a very less known usage `source bigdl-nano-init -f`

### 3. Summary of the change 
1. delete 2 variables `ENABLE_TF_OPTS` and `NANO_TF_INTER_OP`, they were used for `set_inter_op_parallelism_threads` and `set_intra_op_parallelism_threads` only
2. Use default inter/intra op parallelism threads for tf, it seems to be the best setting.

### 4. How to test?
- [ ] Unit test


MeouSker77 (approved):
LGTM

MeouSker77:
We have tried setting `inter=socket_num, intra=core_num_per_socket`, but we found the best configuration is using default inter/intra op parallelism, i.e. do not set it manually. @jason-dai

PR #7112, opened 2022-12-28
Orca: Restore batch before hook `before_train_iter`

leonardozcm:
## Description

In hook `before_train_iter`, a user expects to see the batch in the original format, so we restore batch format before  `before_train_iter` is called. Also we recycle `self.batch` attribute when this iteration is over.

### DIFF
--- a/python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
@@ -484,8 +484,8 @@ class TorchRunner(BaseRunner):
             self.batch = *features, target
         else:
             invalidInputError(False,
-                                "Features should be tensor, list/tuple, "
-                                "but got {}".format(type(features)))
+                              "Features should be tensor, list/tuple, "
+                              "but got {}".format(type(features)))
         self.call_hook(call_backs=callbacks, fn_name="before_train_iter")
 
         # Compute output.
@@ -631,8 +631,8 @@ class TorchRunner(BaseRunner):
             self.batch = *features, target
         else:
             invalidInputError(False,
-                                "Features should be tensor, list/tuple, "
-                                "but got {}".format(type(features)))
+                              "Features should be tensor, list/tuple, "
+                              "but got {}".format(type(features)))
         self.call_hook(call_backs=callbacks, fn_name="before_val_iter")
 
         # compute output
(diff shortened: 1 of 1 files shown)


python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
```diff
@@ -475,20 +475,27 @@ def _train_batch(self, batch, batch_info=None, callbacks=None):
...
         # unpack features into list to support multiple inputs model
+        # and restore batch to what it should be.
         features, target = batch
+        if torch.is_tensor(features):
+            self.batch = features, target
+        elif isinstance(features, (tuple, list)):
+            self.batch = *features, target
+        else:
+            invalidInputError(False,
+                              "Features should be tensor, list/tuple, "
```
hkvision, line 487: tensor __or__ ..., to keep the same as the error below.

python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
```diff
@@ -475,20 +475,27 @@ def _train_batch(self, batch, batch_info=None, callbacks=None):
...
+                              "but got {}".format(type(features)))
+        self.call_hook(call_backs=callbacks, fn_name="before_train_iter")
```
hkvision, line 489: call_backs is a bit strange? change to callbacks?

python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
```diff
@@ -475,20 +475,27 @@ def _train_batch(self, batch, batch_info=None, callbacks=None):
                 calculate averages.
 
         """
-        # TODO: restore batch to what it should be.
-        self.call_hook(call_backs=callbacks, fn_name="before_train_iter")
         # unpack features into list to support multiple inputs model
+        # and restore batch to what it should be.
         features, target = batch
```
hkvision, line 480: is this line correct? Why you first split and then restore and then split (in Line493)?

python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
leonardozcm: sure

python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
leonardozcm: > is this line correct? Why you first split and then restore and then split (in Line493)?

`first split and restore` is meant to recover batch to how user defined in dataloader
the last split(with a star-pack * operation) is to ensure features after `before_train_iter` are fed into self.model (No interal operations on features)

And all actions passed indicates that this adjustment works well as before I think.

PR #7124, opened 2022-12-29
Orca: Assign Some Attrs to Runner and Unifies `model` Definition

leonardozcm:
## Description

In this pr we assign some internal variables like batch and loss to torchrunner's attributes:
```python
batch -> self.batch
output -> self.output
loss -> self.loss
```
What's more we also unifies model and training_model and hide details about ddp with discussing mode in models property:
```python
    @property
    def model(self):
        """
        First or only model(s) created by the ``model_creator``.
        Discuss whether to return ddp model depending on the mode.
        """
        if self._mode == RunnerMode.train:
            if self.training_models:
                return self.training_models[0]
        else:
            if self.models:
                return self.models[0]
```

### DIFF
--- a/python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
@@ -105,10 +105,6 @@ class TorchDistBackend(DistBackend):
         return dist.all_reduce(tensor, *args,
                                **all_reduce_min_kwargs)
 
-class RunnerMode(Enum):
-    train = 0
-    val = 1
-    predict = 2
 
 class TorchRunner(BaseRunner):
     """Manages a PyTorch model for training."""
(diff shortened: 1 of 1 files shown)


python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
```diff
@@ -104,6 +105,10 @@ def all_reduce_min(self, tensor, *args, **kwargs):
         return dist.all_reduce(tensor, *args,
                                **all_reduce_min_kwargs)
 
+class RunnerMode(Enum):
```
hkvision, line 108: Does this PR conflict with https://github.com/intel-analytics/BigDL/pull/7050 ?
Or at least RunnerMode could be combined if we decide to add this

leonardozcm: > Does this PR conflict with #7050 ? Or at least RunnerMode could be combined if we decide to add this

I had synced with Ge about this before, and I think it's beneficial to set a flag like this to indicate runner status.

hkvision: Is this necessary? A runner may have multiple modes right?

leonardozcm: > Is this necessary? A runner may have multiple modes after its creation right?

No, so using string comparision instead?

leonardozcm: synced offline with ge.

PR #7138, opened 2022-12-30
Nano: fix keras onnx model output shape

MeouSker77:
## Description

Fix the output shape of keras onnx model

### 1. Why the change?

`tf.py_function(func, ..., Tout=tf.float32)` will convert the output of `func` to a single Tensor, that means `[t]` or `[[t]]` (`t` is a `tf.Tensor`) all will be converted to `t`.

Besides, if the original model return a single tensor `t`, then after tracing, onnxruntime model will return `[t]`.

These two rules will destroy the output shape when the original output is a single tensor, or a list of tensor with only one element, or a list of list of tensor with only one element, ...

Before this PR, we simply check whether the output is a list and has only one element, if is, then return its first element. We use this method to 

However, this method cannot distinguish whether the original model return `t` or `[t]` (onnxruntime model will return `[t]` in both case). And we don't handle the influence of `tf.py_function`.

This PR fix both, it save the right output shape first, and convert the final output to this right shape.

### 2. User API changes

N/A

### 3. Summary of the change 

N/A

### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/deps/onnxruntime/tensorflow/tensorflow_onnxruntime_model.py
+++ b/python/nano/src/bigdl/nano/deps/onnxruntime/tensorflow/tensorflow_onnxruntime_model.py
@@ -21,7 +21,7 @@ from pathlib import Path
 from tempfile import TemporaryDirectory
 import tensorflow as tf
 from bigdl.nano.utils.util import get_default_args
-from bigdl.nano.tf.utils import KERAS_VERSION_LESS_2_10
+from bigdl.nano.tf.utils import KERAS_VERSION_LESS_2_10, fake_tensor_from_spec
 from bigdl.nano.utils.inference.tf.model import AcceleratedKerasModel
 from bigdl.nano.utils.log4Error import invalidInputError
 
@@ -59,6 +59,14 @@ class KerasONNXRuntimeModel(ONNXRuntimeModel, AcceleratedKerasModel):
                     input_spec = (input_spec, )
                 tf2onnx.convert.from_keras(model, input_signature=input_spec,
                                            output_path=onnx_path, **export_kwargs)
+                # save the nesting level of inference output
+                fake_inputs = [fake_tensor_from_spec(spec) for spec in input_spec]
+                fake_outputs = model(*fake_inputs)
+                self._nesting_level = 0
+                while isinstance(fake_outputs, (tuple, list)) and len(fake_outputs) == 1:
+                    self._nesting_level += 1
+                    fake_outputs = fake_outputs[0]
+
                 self._inputs_dtypes = [inp.dtype for inp in input_spec]
                 self._default_kwargs = get_default_args(model.call)
                 if KERAS_VERSION_LESS_2_10:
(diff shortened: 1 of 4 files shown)


python/nano/test/tf/keras/test_trace_and_quantize.py
```diff
@@ -99,3 +108,16 @@ def test_evaluate_after_trace(self):
             InferenceOptimizer.save(traced_model, tmp_dir_name)
             new_model = InferenceOptimizer.load(tmp_dir_name, model)
         new_model.evaluate(x=x, y=y)
+
+    def test_inference_output_shape(self):
+        model = MyModelReturnList()
+        x = np.random.random((100, 4))
+        traced_model = InferenceOptimizer.trace(model, accelerator="onnxruntime",
+                                                input_spec=tf.TensorSpec(shape=(None, 4), dtype=tf.float32))
+        outputs = traced_model(x, tf.convert_to_tensor(True))
```
rnwang04, line 117: why here needs an additional `tf.convert_to_tensor(True)`?

python/nano/src/bigdl/nano/deps/onnxruntime/tensorflow/tensorflow_onnxruntime_model.py
```diff
@@ -89,8 +100,6 @@ def on_forward_start(self, inputs):
         return inputs
 
     def on_forward_end(self, outputs):
-        if isinstance(outputs, list) and len(outputs) == 1:
-            outputs = outputs[0]
         outputs = self.numpy_to_tensors(outputs)
```
rnwang04, line 103: if output is a multilayer embedded list， does `self.numpy_to_tensors(outputs)` still works?

python/nano/src/bigdl/nano/deps/onnxruntime/tensorflow/tensorflow_onnxruntime_model.py
```diff
@@ -78,7 +86,10 @@ def __call__(self, *args, **kwargs):
                 kwargs[name] = value
         for param in self._call_fn_args_backup[len(inputs):len(self._forward_args)]:
             inputs.append(kwargs[param])
-        return self.call(*inputs)
+        outputs = self.call(*inputs)
```
rnwang04, line 89: If I understand right, here returns original onnx output(a list) ? so maybe it contains _nesting_level at first (not start from 0) ? 
It's just a bit confusing whether this way of adding lists works in all cases

rnwang04 (approved):
others LGTM

python/nano/src/bigdl/nano/deps/onnxruntime/tensorflow/tensorflow_onnxruntime_model.py
MeouSker77: No, here `outputs` is a single Tensor, not onnx output.

MeouSker77: `tf.py_function(func, ..., Tout=tf.float32)` will convert the output of func to a single Tensor

python/nano/test/tf/keras/test_trace_and_quantize.py
MeouSker77: A typo, fixed it

python/nano/src/bigdl/nano/deps/onnxruntime/tensorflow/tensorflow_onnxruntime_model.py
MeouSker77: yes, it works, <https://github.com/intel-analytics/BigDL/blob/79f1df8f472fa858e30ade8a1992cb4a965250d7/python/nano/src/bigdl/nano/utils/inference/tf/model.py#L73>

TheaperDeng (approved):
LGTM

PR #7142, opened 2023-01-03
Chronos: remove channels_last methods from `forecaster.optimize()`

plusbang:
## Description

Considering that channels_last methods make no sense to time series data, remove it from available methods in `forecaster.optimize()`.

### 2. User API changes

no change

### 4. How to test?
- [x] Unit test


TheaperDeng (approved):
LGTM

PR #7151, opened 2023-01-04
Nano: add channels_last convert check

ACupofAir:
### 1. Why the change?
When we pass in a `input_sample` which is not consist of tensor, or one of the item in input doesn't have `.to()` method, it will lead to an error to convert it on `channels_last` is true.

### 2. Summary of the change 
* When user directly use `trace` or `quantize` function with a `input_sample` pass in or use `optimize` function which mean the `input_sample` is **not None**, we will compute a list named `channels_last_available` which instruct which item in the `input_sample` can be convert to `channels_last`.
* When user directly use `trace` or `quantize` function and don't pass in `input_sample`, make the list `channels_last_available` empty, and the `forward_step` function will convert default.
* Add unit-test of methods with channels_last

**warning**: 
* The methods with jit hasn't been handled, I'm working on it.
* This pr is to replace #6918 for the previous pr conflicts with main branch

### DIFF
--- a/python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
+++ b/python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 #
 
+from bigdl.nano.utils.inference.pytorch.model_utils import generate_channels_last_available
 from bigdl.nano.utils.inference.pytorch.model import AcceleratedLightningModule
 from bigdl.nano.pytorch.context_manager import generate_context_manager
 from bigdl.nano.deps.ipex.ipex_api import ipex_optimize
@@ -22,7 +23,7 @@ import torch
 
 class PytorchIPEXJITModel(AcceleratedLightningModule):
     def __init__(self, model: torch.nn.Module, input_sample=None, use_ipex=False, dtype=None,
-                 use_jit=False, channels_last=None, thread_num=None, from_load=False,
+                 use_jit=False, channels_last=None, channels_last_available=[], thread_num=None, from_load=False,
                  inplace=False, jit_strict=True, jit_method=None, weights_prepack=None):
         """
         This is the accelerated model for pytorch and ipex/jit.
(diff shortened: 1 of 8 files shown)


ACupofAir (commented):
LGTM

python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
```diff
@@ -74,6 +76,12 @@ def __init__(self, model: torch.nn.Module, input_sample=None, use_ipex=False, dt
         self.original_model = model
         if self.channels_last:
             self.model = self.model.to(memory_format=torch.channels_last)
+            if channels_last_available: # init from _load, the channels_last_available is not none
+                self.channels_last_available = channels_last_available
+            else:
+                self.channels_last_available = generate_channels_last_available(input_sample)
+        else:
+            self.channels_last_available = []
```
TheaperDeng, line 84: add same logic to if load code block for jit related model

python/nano/src/bigdl/nano/utils/inference/pytorch/model_utils.py
```diff
@@ -198,6 +198,26 @@ def get_input_example(model, input_sample, forward_args):
...
+    '''
+    # try channels_last available
+    if inputs:  # to avoid the situation of inputs == None
+        channels_last_available = [True] * len(inputs)
+        for idx, input in enumerate(inputs):
+            try:
+                input.to(memory_format=torch.channels_last)
+            except Exception as _e:
+                channels_last_available[idx] = False
+            else:
```
rnwang04, line 214: no need for this else

python/nano/test/pytorch/tests/inference/test_ipex_jit_inference.py
```diff
@@ -103,6 +113,49 @@ def test_ipex_jit_inference(self):
         with InferenceOptimizer.get_context(new_model):
             new_model(self.data_sample)
     
+    def test_ipex_channels_last_inference(self):
```
rnwang04, line 116: One suggestion here, may not change your code.
For this case, maybe just define one ut like `test_jit_ipex_channels_last_inference` here and list three cases inside the the function, so you don't need to define model, x1, x2, x3 three times.

rnwang04 (approved):
others LGTM

PR #7152, opened 2023-01-04
Nano: fix bug of openvino cannot infer input shape

rnwang04:
## Description

fix bug of openvino cannot infer input shape.

### 1. Why the change?

If user defines a custom model, which inherent from keras.Model, but don't contain explicit input layer, he will meet below error when trace/quantize with openvino:
```bash
cannot be saved either because the input shape is not available or because the forward pass of the model is not defined.To define a forward pass, please override `Model.call()`. To specify an input shape, either call `build(input_shape)` directly, or call the model on actual data using `Model()`, `Model.fit()`, or `Model.predict()`. 
```

### 2. User API changes

Now, `input_spec` works for openvino accelerator:
```python
# trace
traced_model = InferenceOptimizer.trace(model, accelerator="openvino",
                                        input_spec=tf.TensorSpec(shape=(None, 4), dtype=tf.float32))
# quantize
quantized_model = InferenceOptimizer.quantize(model,
                                              accelerator="openvino",
                                              input_spec=tf.TensorSpec(shape=(None, 4), dtype=tf.float32),
                                              x=np.random.random((100, 4)),
                                              y=np.random.random((100, 5)),
                                              accuracy_criterion = {'relative': 0,
                                                                    'higher_is_better': True})
```

### 3. Summary of the change 

- setting input shape of model by `model.compute_output_shape(input_shape)`
- related uts

### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/deps/openvino/openvino_api.py
+++ b/python/nano/src/bigdl/nano/deps/openvino/openvino_api.py
@@ -85,7 +85,7 @@ def load_openvino_model(path, framework='pytorch', device=None):
                           " Please choose from 'pytorch'/'tensorflow'.")
 
 
-def KerasOpenVINOModel(model, input_spec=None, precision='fp32', 
+def KerasOpenVINOModel(model, input_spec=None, precision='fp32',
                        thread_num=None, device='CPU', config=None,
                        logging=True):
     """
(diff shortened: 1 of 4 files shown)


MeouSker77 (approved):
LGTM

TheaperDeng (approved):
LGTM, I wonder if this way(`compute_output_shape`) works for INC's quantization on model inherited from `keras.Model`? @MeouSker77

rnwang04:
> LGTM, I wonder if this way(`compute_output_shape`) works for INC's quantization on model inherited from `keras.Model`? @MeouSker77

I'm trying this on INC, but INC seems more complex than openvino.

PR #7160, opened 2023-01-04
Nano: provide an internal function to give suggestions to users in a conspicuous way

TheaperDeng:
### 1. Why the change?
We may check if users break or fails to follow some BKMs, while we could not transparently apply them for our users. We prefer to tell our users that they could do something to further improve their performance once they stopped the script/completed the script in a conspicuous way.

### 2. User API changes
nothing, this is an internal function.
For developers, they could register a suggestion any time they want in nano's src.

```python
from bigdl.nano.utils.log4warning import register_suggestion
register_suggestion("You should call `bigdl-nano-init -p`")
```

And when users complete their application, they will see the suggestion **at the bottom of** their output, so that they are unlikely to miss the message.:
```bash
*****************Nano performance Suggestions*****************
You should call `bigdl-nano-init -p`

*****************Nano performance Suggestions*****************
```

The suggestions will also be shown even if our users interrupt the program
```bash
^CTraceback (most recent call last):
  File "test.py", line 6, in <module>
    time.sleep(10)
KeyboardInterrupt

*****************Nano performance Suggestions*****************
You should call `bigdl-nano-init -p`

*****************Nano performance Suggestions*****************
```

### 3. Summary of the change 
1. register a print function to print the suggestions, it won't do anything if no suggestion could be provided

### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/__init__.py
+++ b/python/nano/src/bigdl/nano/__init__.py
@@ -40,7 +40,9 @@ def _check_nano_envs():
 # disable env check for now, as it does not work for tf and windows
 # _check_nano_envs()
 
+
 import atexit
 from bigdl.nano.utils.log4warning import output_suggestions
+
 
 atexit.register(output_suggestions)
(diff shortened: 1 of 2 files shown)


MeouSker77 (approved):
LGTM

rnwang04 (approved):
LGTM

PR #7162, opened 2023-01-04
Nano: support custom model for keras INC 

rnwang04:
## Description

Before Nano can only accept model defined by keras.Model(inputs=.., outputs=..) for inc accelerator.
Now we can support more custom model for inc (>= 1.14)

### 1. Why the change?

Support keras model without input layer for inc.

### 2. User API changes

Now `input_spec` works for `accelerator=None` , you should specify this when your model don't have input layer.
```python
quantized_model = InferenceOptimizer.quantize(model,
                                              accelerator=None,
                                              input_spec=tf.TensorSpec(shape=(None, 4), dtype=tf.float32),
                                              x=np.random.random((100, 4)),
                                              y=np.random.random((100, 5)))
```

### 3. Summary of the change 

- Support keras model without input layer for inc(>= 1.14).
- related uts

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/deps/neural_compressor/tensorflow/quantization.py
+++ b/python/nano/src/bigdl/nano/deps/neural_compressor/tensorflow/quantization.py
@@ -18,6 +18,7 @@ from ..core import BaseQuantization
 from .metric import TensorflowINCMetric
 from .model import KerasQuantizedModel
 from .utils import Dataloader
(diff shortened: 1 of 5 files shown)


plusbang (commented):
LGTM

MeouSker77 (approved):
LGTM

Issue #7161, opened 2023-01-04
Orca: the optimizer_creator and loss_func shouldn't be required in PytorchPysparkEstimator.predict() 


zpeng1898:
When predicting xshards using `PytorchPysparkEstimator`, an error occurs showing that the `optimizer_creator` and `loss` parameters are required. But in fact, they shouldn't be required in predict() after loading the pretrained model.

Code:
![image](https://user-images.githubusercontent.com/42887453/210506654-fe409c15-2a65-440b-ab19-38b67c85c4e4.png)

Problem:
![image](https://user-images.githubusercontent.com/42887453/210506909-9c4afa37-8a32-4bda-a434-9de0a4b32d52.png)

The error disappears after adding the optimizer_creator and loss to the estimator:
![image](https://user-images.githubusercontent.com/42887453/210506866-d16290e6-cb51-47c8-99f7-8587a2b84022.png)

hkvision:
@sgwhat Please take a look.

zpeng1898:
![image](https://user-images.githubusercontent.com/42887453/210511292-609d1a53-fa51-4d94-9915-6e862df60407.png)
![image](https://user-images.githubusercontent.com/42887453/210511352-62d156bd-0d65-4262-bc8a-4ae5cfb788e1.png)

hkvision:
https://github.com/intel-analytics/BigDL/pull/6988/files#diff-9ecfa460f1e73f063bab468a4717117b6d390ba93ce1e5e7c13e49b94533d05fL141
@leonardozcm Please take a look.

leonardozcm:
https://github.com/intel-analytics/BigDL/pull/7168 track here

hkvision:
This should have been fixed. Closing it.

PR #7166, opened 2023-01-04
[Orca] Update optimizer states to support loading model

sgwhat:
## Description
Update for issue https://github.com/intel-analytics/BigDL/issues/7161

### How to test?
- [x] Unit test

PR #7168, opened 2023-01-04
Orca: Make loss_creator optional and discuss criterion when fit and eval

leonardozcm:
## Description

When in predict mode, there should be no loss_creator. So we should require loss_creator when fiting or validating instead of init estimator.
https://github.com/intel-analytics/BigDL/issues/7161?notification_referrer_id=NT_kwDOAZp2hLM1MjM4NTY4MjQxOjI2OTAwMTAw


hkvision:
@sgwhat Take a look. After this PR merge, you can remove the arguments in your PR? https://github.com/intel-analytics/BigDL/pull/7166

hkvision (approved):

PR #7178, opened 2023-01-05
Nano: support mo kwargs in trace/quantize and support VPU fp16 in quantize

rnwang04:
## Description

In this PR, try to expose kwargs of model optimizer to user, and support VPU fp16 precision by forcibly reminding user to pass in `mean_value` parameter to model optimizer.

### 1. Why the change?

After my debugging, I found VPU fp16 works only if we pass `mean_value` paramater to model optimizer.
At the same time, model optimizer has many other params, and user maybe want to specify some of them, I think we should expose these kwargs to user and on the PyTorch side, distinguish between `torch.onnx.export` and `model optimizer`'s kwargs.

### 2. User API changes

Now user can pass in kwargs for `torch.onnx.export` function and `model optimizer` function, for example:
```python
ov_model = InferenceOptimizer.quantize(model,
                                       accelerator="openvino",
                                       input_sample=x,
                                       calib_data=x,
                                       precision='int8',
                                       do_constant_folding=False,  # pass to torch.onnx.export
                                       mean_value=[123.68,116.78,103.94]  # pass to model optimizer
                                       )
```
At the same time, we have supported VPUX fp16, the API looks like (**mean_value is necessary for VPU fp16**) :
```python
openvino_model = InferenceOptimizer.quanize(model,
                                            input_sample=x,
                                            accelerator='openvino',
                                            device='VPUX',
                                            precision='fp16',
                                            mean_value=[123.68,116.78,103.94])
```
For TensorFlow, the usage is similar.

### 3. Summary of the change 

- export kwargs of model optimizer to user
- on PyTorch side, distinguish between `torch.onnx.export` and `model optimizer`'s kwargs.
- related ut for this kwargs
- support VPUX fp16 by forcibly reminding user to pass in `mean_value` parameter to model optimizer.
- related ut for this

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/deps/openvino/core/model.py
+++ b/python/nano/src/bigdl/nano/deps/openvino/core/model.py
@@ -84,13 +84,10 @@ class OpenVINOModel:
         if self.additional_config is not None and self._device == 'CPU':
             # TODO: check addition config based on device
             config.update(self.additional_config)
-        if self._device != 'VPUX' or self._precision == 'int8':
-            # For VPU, now only int8 quantizaton model works
-            # fp16 model always meets compilation error
-            self._compiled_model = self._ie.compile_model(model=self.ie_network,
-                                                          device_name=self._device,
-                                                          config=config)
-            self._infer_request = self._compiled_model.create_infer_request()
+        self._compiled_model = self._ie.compile_model(model=self.ie_network,
+                                                      device_name=self._device,
+                                                      config=config)
+        self._infer_request = self._compiled_model.create_infer_request()
         self.final_config = config
         input_names = [t.any_name for t in self._ie_network.inputs]
         self._forward_args = input_names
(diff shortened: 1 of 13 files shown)


jason-dai:
LGTM.

What does `mean_value` stand for? Mean value of inputs? We may need to add some simple docs on what kwargs that we support in these methods later.

rnwang04:
> > Where should we put such docs? Shall I just add some explanation for `mean_value` or we should explain more potential kwargs in a separate place like an individual how-to guide?
> 
> May in the API doc, explaining what types of arguments are supported (for PyTorch ONNX export and OV model optimizer), and simply referring to the PyTorch/OpenVINO docs; we can also list some important arguments such as `mean_value`.

Have update API doc based on this.

jason-dai:
LGTM. Maybe test more models (including stable diffusion) later.

PR #7180, opened 2023-01-05
[Nano] Add how-to-guide of load/save API for tensorflow inference

HensonMa:
## Description
The preview docs is here https://henrytestnano-doc.readthedocs.io/en/latest/doc/Nano/Howto/index.html.

### DIFF
--- a/docs/readthedocs/source/_toc.yml
+++ b/docs/readthedocs/source/_toc.yml
@@ -125,6 +125,8 @@ subtrees:
                   - file: doc/Nano/Howto/Inference/PyTorch/pytorch_save_and_load_jit
                   - file: doc/Nano/Howto/Inference/PyTorch/pytorch_save_and_load_onnx
                   - file: doc/Nano/Howto/Inference/PyTorch/pytorch_save_and_load_openvino
+                  - file: doc/Nano/Howto/Inference/TensorFlow/tensorflow_save_and_load_onnx
+                  - file: doc/Nano/Howto/Inference/TensorFlow/tensorflow_save_and_load_openvino
                   - file: doc/Nano/Howto/Inference/PyTorch/inference_optimizer_optimize
                   - file: doc/Nano/Howto/Inference/TensorFlow/accelerate_tensorflow_inference_onnx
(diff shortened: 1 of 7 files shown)


HensonMa:
The order of insertion of file toc.yml has issues, need revision...

HensonMa:
Done. 
New preview https://henrytestnano-doc.readthedocs.io/en/latest/doc/Nano/Howto/index.html

HensonMa:
New preview link https://henrytestnano-doc.readthedocs.io/en/latest/doc/Nano/Howto/index.html.

TheaperDeng:
> New preview link https://henrytestnano-doc.readthedocs.io/en/latest/doc/Nano/Howto/index.html.

hint: no need to reply the link many times since they are actually same. :)

HensonMa:
> > New preview link https://henrytestnano-doc.readthedocs.io/en/latest/doc/Nano/Howto/index.html.
> 
> hint: no need to reply the link many times since they are actually same. :)

Oh! I just realized all the generated links are the same ha ha >v<

TheaperDeng:
The how to guides are quite good. Only one concern.

@MeouSker77 Are we requiring all users to input an original model for onnxruntime, openvino keras accelerated model for loading? I understand this function https://github.com/intel-analytics/BigDL/blob/58ba9bad70be5c1b7bf2c3104ed911a64180850c/python/nano/src/bigdl/nano/tf/utils.py#L57-L84
does not allow source_obj to be None right?

If we could support not inputting the original model, we may add a note for this in API doc and how to guides.
If we could not, we may need to support it? Since not all users need the attr from original model right?

MeouSker77:
> Are we requiring all users to input an original model for onnxruntime, openvino keras accelerated model for loading

Yes, now we require original model in all cases.

> If we could not, we may need to support it? Since not all users need the attr from original model right?

You are right, we may support it in future.

TheaperDeng:
> > Are we requiring all users to input an original model for onnxruntime, openvino keras accelerated model for loading
> 
> Yes, now we require original model in all cases.
> 
> > If we could not, we may need to support it? Since not all users need the attr from original model right?
> 
> You are right, we may support it in the future.

Okay, we will just make this a required parameter in this how to guide and change it once we fix this issue.

PR #7183, opened 2023-01-06
[dllib]update keras error message

qiuxin2012:
## Description

[dllib]update keras error message #7004


qiuxin2012:
jenkins passed

PR #7145, opened 2023-01-03
[PPML] Add encryption feature to torch.save/torch.load in nano

gc-fu:
## Description

Move encryption feature of torch.save/torch.load to nano

### 1. User API changes

```python
#original api:
def save(obj, f: Union[str, os.PathLike, BinaryIO, IO[bytes]],
         pickle_module=pickle, pickle_protocol=DEFAULT_PROTOCOL, _use_new_zipfile_serialization=True) -> None:
    pass

#new api:
def save(obj, f: Union[str, os.PathLike, BinaryIO, IO[bytes]], pickle_module = pickle, pickle_protocol=DEFAULT_PROTOCOL, _use_new_zipfile_serialization=True, encryption_key: Optional[str] = None) -> None:
    pass

#similarly for load:
def load(f, map_location=None, pickle_module=pickle, **pickle_load_args):
    pass

def load(f, map_location=None, pickle_module=pickle, decryption_key:Optional[str]=None, **pickle_load_args):
    pass
```

### 2. Summary of the change 

Move encryption feature to torch.save/torch.load

### 3. How to test?
- [x] Unit test

### 4. TODO
- [x] Rename to patch_encryption and move it to bigdl.nano.pytorch.patching

### 5. New dependency
- [x] cryptography

PR #7188, opened 2023-01-06
[PPML] add integration example for kms and patch_encryption

gc-fu:
## Description

As we have changed the api for torch.save and torch.load in ppml, we should also change the example file accordingly.

### 1. Why the change?

Changed according to pr: https://github.com/intel-analytics/BigDL/pull/7145

### 2. User API changes

None.

### 3. Summary of the change 

Provide example file for patch_encryption and kms

### 4. How to test?
- [x] Manually test

wow

Have tested it in SGX mode in image.

### 5. Rebased PR
https://github.com/intel-analytics/BigDL/pull/7191
https://github.com/intel-analytics/BigDL/pull/7197


ppml/trusted-deep-learning/base/load_save_encryption_ex.py
```diff
@@ -0,0 +1,220 @@
...
+    return get_data_key_plaintext(EHSM_IP, EHSM_PORT, encrypted_primary_key_path, encrypted_data_key_path)
+
+
+def save_encrypted_dataset(dataset_path, save_path, secret_key):
+    dataset = load_from_disk(dataset_path, keep_in_memory=True)
+    # This will save the encrypted dataset into disk
+    torch.save(dataset, save_path, encryption_key=secret_key)
+
+
+class Dataset(torch.utils.data.Dataset):
```
glorysdj, line 94: can this dataset be reused?

gc-fu: Fixed in PR: https://github.com/intel-analytics/BigDL/pull/7197

glorysdj (approved):
LGTM

PR #7190, opened 2023-01-06
[PPML] Add overloaded csv method for EncryptedDataFrameWriter

SEULiyao:
## Description

add an overloaded `csv` method, which could accept a key from parameter and add this key into `sparkSession.sparkContext.hadoopConfiguration`, so that dataframe could get this key easily.


qiyuangong (approved):
LGTM

PR #7195, opened 2023-01-06
Remove redundant version comparison in Ray

hkvision:
A PR separated from https://github.com/intel-analytics/BigDL/pull/7069

- Ray versions lower than 1.4.0 are no longer supported, and thus remove the version comparison.
- Change az to bigdl
- Remove the code not used in RayXShards.

### DIFF
--- a/python/orca/src/bigdl/orca/data/ray_xshards.py
+++ b/python/orca/src/bigdl/orca/data/ray_xshards.py
@@ -108,7 +108,7 @@ def init_ray_if_not(redis_address, redis_password):
         init_params = dict(
             address=redis_address,
             ignore_reinit_error=True,
-            namespace="az"
+            namespace="bigdl"
         )
         if redis_password:
             init_params["_redis_password"] = redis_password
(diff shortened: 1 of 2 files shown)


python/orca/src/bigdl/orca/data/ray_xshards.py
```diff
@@ -445,25 +443,14 @@ def _from_spark_xshards_ray_api(spark_xshards: "SparkXShards") -> "RayXShards":
...
         partition_store_names = list(partition_stores.keys())
         result_rdd = spark_xshards.rdd.mapPartitionsWithIndex(lambda idx, part: write_to_ray(
             idx, part, address, password, partition_store_names)).cache()
-        result = result_rdd.collect()
-
-        id2ip = {}
-        id2store_name = {}
-        for idx, ip, local_store_name in result:
-            id2ip[idx] = ip
-            id2store_name[idx] = local_store_name
```
hkvision, line 467: This code piece don't have any impact, thus removed.

PR #7196, opened 2023-01-08
Nano: add tensorflow bf16 inference support for "model transformation" style

TheaperDeng:
### 1. Why the change?
We have now support a easy wrapper for `mixed_precision.set_global_policy(policy)`  in `patch_tensorflow`, while it has quite a few limitation, for the most important one, **it can not support a easy model transformation between fp32 -> bf16**, i.e.,
```python
model = tf.keras.models.load_model(...)  # a model defined and trained in fp32 environment
opt_model = transform_model_fp32_to_bf16(model)
opt_model(data) # here we are using bf16 for calculation, fp32 for weight saving (i.e., amp)
```

In this PR, an implementation is added and verified as proposed by @yangw1234 . We change the `dtype` key of the config (got by `model.get_config`) and set back with a save-and-load trick.

### 2. User API changes
Add a new method in `bigdl.nano.tf.keras.InferenceOptimizer.quantize`
```python
model = tf.keras.models.load_model(...)  # a model defined and trained in fp32 environment

opt_model = InferenceOptimizer.quantize(model, precision="bf16")

opt_model(data) # here we are using bf16 for calculation, fp32 for weight saving (i.e., amp)
```
**There is even no need to call `patch_tensorflow(precision='mixed_bfloat16')`**

A quick usage example:
```python
import tensorflow as tf
from bigdl.nano.tf.keras import Sequential
import numpy as np

x_train = np.random.rand(1000,3,224,224)

model = Sequential([
  tf.keras.layers.Conv2D(256, 1, activation='relu', input_shape=(3,224,224)),
  tf.keras.layers.Flatten(),
  tf.keras.layers.Dense(512, activation='relu'),
  tf.keras.layers.Dense(10, activation='softmax')
])

model.compile(optimizer='adam',
              loss='sparse_categorical_crossentropy',
              metrics=['accuracy'])

model = InferenceOptimizer.quantize(model, precision="bf16")

import time

st = time.time()
for _ in range(10):
    model(x_train)
print(time.time() - st)
```
BF16 is **37.5% faster** than FP32 in this example.

It has been verified by setting `ONEDNN_VERBOSE=1`
bf16 model log is as following
```bash
...
onednn_verbose,exec,cpu,convolution,brgconv_1x1:avx512_core_bf16,forward_training,src_bf16::blocked:acdb:f wei_bf16::blocked:AcdB64a2b:f bia_undef::undef::f dst_bf16::blocked:acdb:f,attr-scratchpad:user ,alg:convolution_direct,mb,53.0029
...
```
it shows that the quantization and model transformation works.

### 3. Summary of the change 
When users call `InferenceOptimizer.quantize(model, precision="bf16")` a few procedures are carried out.
1. get layers by `model.layers`
2. set `_dtype_policy` for these layer
3. save and load the model

Some downside:
1. save-and-load trick is necessary
2. bf16 isa check to be added late
3. a good question is "which layer needs to be transformed to bf16", make all of them bf16 is clearly not the best way (but a good start for now :)
4. some of the implementation is a little bit hacking, the most problematic one is the usage of `_dtype_policy`.

### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/tf/keras/amp/bfloat16.py
+++ b/python/nano/src/bigdl/nano/tf/keras/amp/bfloat16.py
@@ -18,6 +18,7 @@ from tensorflow.keras import mixed_precision
 from tempfile import TemporaryDirectory
 import tensorflow as tf
 
+
 class BF16Model(AcceleratedKerasModel):
     def __init__(self, model, **kwargs):
         super().__init__(model, precision=tf.bfloat16)
@@ -39,7 +40,7 @@ class BF16Model(AcceleratedKerasModel):
         status.update({"ModelType": type(self.target_obj).__name__,
                        "precision": "bfloat16"})
         return status
-    
+
     def _save_model(self, path):
         self.model.save(path)
 
--- a/python/nano/src/bigdl/nano/tf/keras/amp/bfloat16.py
+++ b/python/nano/src/bigdl/nano/tf/keras/amp/bfloat16.py
@@ -28,8 +28,8 @@ class BF16Model(AcceleratedKerasModel):
             layer._dtype_policy = policy_bf16
         with TemporaryDirectory() as temp_dir:
             model.save(temp_dir)
-            model = tf.keras.models.load_model(temp_dir)
-        self.model = model
+            bf16_model = tf.keras.models.load_model(temp_dir)
+        self.model = bf16_model
 
     def forward_step(self, *inputs):
         return self.model(*inputs)
--- a/python/nano/src/bigdl/nano/tf/keras/amp/bfloat16.py
+++ b/python/nano/src/bigdl/nano/tf/keras/amp/bfloat16.py
@@ -1,0 +1,48 @@
+#
+# Copyright 2016 The BigDL Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+from bigdl.nano.utils.inference.tf.model import AcceleratedKerasModel
+from tensorflow.keras import mixed_precision
+from tempfile import TemporaryDirectory
+import tensorflow as tf
+
+class BF16Model(AcceleratedKerasModel):
+    def __init__(self, model, **kwargs):
+        super().__init__(model, precision=tf.bfloat16)
+        # TODO: check bf16 isa
+        policy_bf16 = mixed_precision.Policy('mixed_bfloat16')
+        for layer in model.layers:
+            layer._dtype_policy = policy_bf16
+        with TemporaryDirectory() as temp_dir:
+            model.save(temp_dir)
+            model = tf.keras.models.load_model(temp_dir)
+        self.model = model
+
+    def forward_step(self, *inputs):
+        return self.model(*inputs)
+
+    @property
+    def status(self):
+        status = super().status
+        status.update({"ModelType": type(self.target_obj).__name__,
+                       "precision": "bfloat16"})
+        return status
+    
+    def _save_model(self, path):
+        self.model.save(path)
+
+    @staticmethod
+    def _load(path):
+        return BF16Model(tf.keras.models.load_model(path))
--- a/python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
+++ b/python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
@@ -37,7 +37,7 @@ from bigdl.nano.deps.onnxruntime.onnxruntime_api import KerasONNXRuntimeModel
 from bigdl.nano.deps.openvino.openvino_api import load_openvino_model
 from bigdl.nano.deps.onnxruntime.onnxruntime_api import load_onnxruntime_model
 from bigdl.nano.deps.neural_compressor.inc_api import load_inc_model
-from bigdl.nano.tf.keras.amp import BF16Model
+from bigdl.nano.tf.keras.amp import BF16Model, load_bf16_model
 
 
 class TFAccelerationOption(AccelerationOption):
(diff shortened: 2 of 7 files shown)


python/nano/src/bigdl/nano/utils/inference/tf/model.py
```diff
@@ -41,7 +42,7 @@ def __call__(self, *args, **kwds):
         return super().__call__(*args, **kwds)
 
     def call(self, *inputs):
-        return tf.py_function(self.forward, inputs, Tout=tf.float32)
+        return tf.py_function(self.forward, inputs, Tout=self.precision)
```
MeouSker77, line 45: I'm not sure whether it's necessary, the output of `mixed_bfloat16` may also be float32 ?

python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
```diff
@@ -517,23 +518,27 @@ def quantize(model: Model,
...
+                                            precision=precision,
+                                            thread_num=thread_num,
+                                            device=device,
+                                            config=final_openvino_option,
+                                            logging=logging,
+                                            **kwargs)
+                return patch_attrs(result, model)
+            elif accelerator is None:
+                result = BF16Model(model)
+                return patch_attrs(result, model)
```
MeouSker77, line 541: the same

python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
```diff
@@ -517,23 +518,27 @@ def quantize(model: Model,
...
+                from bigdl.nano.deps.openvino.tf.model import KerasOpenVINOModel    # type: ignore
+                result = KerasOpenVINOModel(model,
+                                            input_spec=input_spec,
+                                            precision=precision,
+                                            thread_num=thread_num,
+                                            device=device,
+                                            config=final_openvino_option,
+                                            logging=logging,
+                                            **kwargs)
+                return patch_attrs(result, model)
```
MeouSker77, line 538: should call `patch_compiled` before `patch_attrs`:
<https://github.com/intel-analytics/BigDL/blob/main/python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py#L648>

python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
```diff
@@ -517,23 +518,27 @@ def quantize(model: Model,
...
-            from bigdl.nano.deps.openvino.tf.model import KerasOpenVINOModel    # type: ignore
-            result = KerasOpenVINOModel(model,
-                                        input_spec=input_spec,
-                                        precision=precision,
-                                        thread_num=thread_num,
-                                        device=device,
-                                        config=final_openvino_option,
-                                        logging=logging,
-                                        **kwargs)
-            return patch_attrs(result, model)
```
MeouSker77, line 536: Here doesn't calling `patch_compiled` is a bug, ruonan will fix it soon

MeouSker77 (approved):
others LGTM

python/nano/src/bigdl/nano/utils/inference/tf/model.py
TheaperDeng: yeah, you are right. We currently make all layers(except the input layer) bf16, we may add a parameter (need design) later for customized requirement (like if the last layer needs to be fp32...), so that we could change our `Tout`.

python/nano/src/bigdl/nano/tf/keras/amp/bfloat16.py
```diff
@@ -0,0 +1,48 @@
...
+class BF16Model(AcceleratedKerasModel):
+    def __init__(self, model, **kwargs):
+        super().__init__(model, precision=tf.bfloat16)
+        # TODO: check bf16 isa
+        policy_bf16 = mixed_precision.Policy('mixed_bfloat16')
+        for layer in model.layers:
+            layer._dtype_policy = policy_bf16
+        with TemporaryDirectory() as temp_dir:
+            model.save(temp_dir)
+            model = tf.keras.models.load_model(temp_dir)
```
jason-dai, line 30: bf16_model=tf.keras.models.load_model(temp_dir)

jason-dai:
> 3. a good question is "which layer needs to be transformed to bf16", make all of them bf16 is clearly not the best way (but a good start for now :)

I wonder if setting global mixed precision policy has the same result on the constructed model?

TheaperDeng:
> > 3. a good question is "which layer needs to be transformed to bf16", make all of them bf16 is clearly not the best way (but a good start for now :)
> 
> I wonder if setting global mixed precision policy has the same result on the constructed model?

no, it does not work

TheaperDeng:
> > 3. a good question is "which layer needs to be transformed to bf16", make all of them bf16 is clearly not the best way (but a good start for now :)
> 
> I wonder if setting global mixed precision policy has the same result on the constructed model?

oh if you mean define a model under global mixed precision policy, then yes, all layer will be bf16 if you don't set dtype explicitly

jason-dai:
> > > 3. a good question is "which layer needs to be transformed to bf16", make all of them bf16 is clearly not the best way (but a good start for now :)
> > 
> > 
> > I wonder if setting global mixed precision policy has the same result on the constructed model?
> 
> oh if you mean define a model under global mixed precision policy, then yes, all layer will be bf16 if you don't set dtype explicitly

Then I don't think this is really an issue as it's the same as TF default behavior.

python/nano/test/tf/keras/test_trace_and_quantize.py
```diff
@@ -177,3 +177,19 @@ def test_inference_output_shape(self):
                                                       input_spec=tf.TensorSpec(shape=(None, 4)), x=x)
         outputs = quantized_model(x)
         assert isinstance(outputs, list) and isinstance(outputs[0], tf.Tensor)
+
+    def test_quantize_bf16(self):
```
yangw1234, line 181: Maybe also test the case where there are customized layers?

python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
TheaperDeng: I will rebase for this fix

python/nano/test/tf/keras/test_trace_and_quantize.py
TheaperDeng: I don't think customized layer will work under this implementation. Customized layer will implement their own `get_config` and `from_config` method, they will set the dtye(if applied) in their `__init__` themselves. If users does not write any logic to set and change the internal dtype, then we could do nothing.

yangw1234: yeah, that makes sense.

PR #7202, opened 2023-01-09
Update download document

hkvision:


hkvision:
https://hkvision-install-0109-2.readthedocs.io/en/latest/doc/Orca/Overview/install.html#download-bigdl-orca

PR #7204, opened 2023-01-09
Nano: fix two missing patch_compiled by new patch_compiled_and_attrs function

rnwang04:
## Description

- fix two missing patch_compiled in keras InferenceOptimizer's quantize
- provide new `patch_compiled_and_attrs` function which contains `patch_compiled` and `patch_attrs`

### How to test?
- [x] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
+++ b/python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
@@ -514,6 +514,7 @@ class InferenceOptimizer(BaseInferenceOptimizer):
                                         config=openvino_config,
                                         logging=logging,
                                         **kwargs)
+            patch_compiled(result, model)
             return patch_attrs(result, model)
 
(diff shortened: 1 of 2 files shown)


MeouSker77 (approved):
LGTM

python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
```diff
@@ -514,6 +514,7 @@ def quantize(model: Model,
                                         config=openvino_config,
                                         logging=logging,
                                         **kwargs)
+            patch_compiled(result, model)
             return patch_attrs(result, model)
```
jason-dai, line 518: Shall we combine these methods into one? Otherwise it's hard to enforce this in the long term.

rnwang04: yeah, you are right. Will construct a new function containing these two functions.

PR #7209, opened 2023-01-09
Update spark-submit section in yarn tutorial

hkvision:
Since BigDL only has two spark versions, the order should be changed to make it clearer. Users should first download BigDL and then download the Spark that matches BigDL.


hkvision:
https://hkvision-yarn-tutorial-0110.readthedocs.io/en/latest/doc/Orca/Tutorial/yarn.html#use-spark-submit

PR #7216, opened 2023-01-10
Change python3.7 to python3.8.

Hanyu-Jin:
## Description

Change python3.7 to python3.8.

### 1. Why the change?

The python version in base image is 3.8.


gc-fu (approved):
LGTM

PR #7218, opened 2023-01-10
Nano: support ONEDNN fusion for jit inside context manager

rnwang04:
## Description

I have tried ONEDNN Graph API based on [this link](https://github.com/pytorch/pytorch/tree/master/torch/csrc/jit/codegen/onednn#pytorch---onednn-graph-api-bridge), and I found ONEDNN fusion can bring obvious acceleration to jit (speedup >= 2x) !

  | Torch Version | FP32 | BF16
-- | -- | -- | --
Original | nightly | 4.15s | 3.3s
ONEDNN | nightly | 0.86s | 0.78s
Original | 1.13.1 | 4.7s | 2.18s
ONEDNN | 1.13.1 | 1.0s | Not supported now
Original | 1.12 | 1.65s | 1.39s
ONEDNN | 1.12 | 0.76s | Not supported now

So here try to support ONEDNN fusion for jit automatically inside context manager.

notes:
1. ONEDNN fusion for fp32 is only supported for versions above torch1.12
2. ONEDNN fusion for bf16 is only supported for versions above torch1.14 [or torch 2.0]
3. speedup is achieved only for static shapes, for dynamic shapes, there is little difference between using ONEDNN fusion or not.

### 1. Why the change?

https://github.com/analytics-zoo/nano/issues/247

### 2. User API changes

new parameter `enable_onednn` for PyTorch InferenceOptimizer's `trace`/`quantize` for user to decide whether use ONEDNN fusion or not.
```python
model = InferenceOptimizer.trace(self.model, accelerator="jit",
                                 use_ipex=True, input_sample=self.data_sample,
                                 enable_onednn=True)
model = InferenceOptimizer.quantize(model, precision='bf16',
                                    accelerator="jit",
                                    use_ipex=True,
                                    input_sample=x,
                                    enable_onednn=True)
```

### 3. Summary of the change 

- support ONEDNN fusion for jit inside context manager
- new parameter `enable_onednn` for PyTorch InferenceOptimizer's trace/quantize
- related uts
- fix mypy error: --install-types failed (no mypy cache directory)

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/context_manager.py
+++ b/python/nano/src/bigdl/nano/pytorch/context_manager.py
@@ -25,10 +25,10 @@ class BaseContextManager(object):
 
     This context manager is used for providing no_grad context only.
     """
-
-    def __init__(self, thread_num=None):
+    def __init__(self, thread_num=None, accelerator=None):
         self.infer_mode = torch.inference_mode(mode=True)
         self.thread_num = thread_num
+        self.accelerator = accelerator
         self.original_thread_num = torch.get_num_threads()
 
     def __enter__(self):
@@ -38,6 +38,8 @@ class BaseContextManager(object):
 
     def __exit__(self, exc_type, exc_value, exc_tb):
         self.infer_mode.__exit__(exc_type, exc_value, exc_tb)
+        if self.accelerator == "jit":
+            torch.jit.enable_onednn_fusion(False)
         torch.set_num_threads(self.original_thread_num)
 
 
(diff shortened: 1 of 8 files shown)


python/nano/src/bigdl/nano/pytorch/context_manager.py
```diff
@@ -46,12 +58,20 @@ class AutocastContextManager(BaseContextManager):
...
     def __enter__(self):
         super().__enter__()
+        if self.accelerator == "jit":
+            if compare_version("torch", operator.le, "1.13.1"):
+                # onednn fusion for bf16 only work for torch version > 1.13
+                if compare_version("torch", operator.ge, "1.12.0"):
+                    # onednn fusion be added to torch from version 1.12
+                    torch.jit.enable_onednn_fusion(False)
+            # Disable AMP for JIT
+            torch._C._jit_set_autocast_mode(False)
```
jason-dai, line 74: are we disabling bf16 when using "jit"?

rnwang04: Not exactly.
_Currently, PyTorch has divergent AMP support in JIT & eager modes, so one should disable JIT AMP support & leverage eager mode AMP support to use BFloat16._
Here we just disable JIT AMP support and use eager mode AMP support.

PR #7219, opened 2023-01-10
Orca: Remove the loop of fit in our example Super Resolution

leonardozcm:
## Description

Call estimator.fit in a loop may raise a big overhead when there is a short period of time for real training, and it seems that our examples are misleading our user to do that.
https://github.com/analytics-zoo/Telecom-CTC-E2E/issues/27#issue-1527156829

So we just apply another way to achieve this.

### DIFF
--- a/python/orca/example/learn/pytorch/super_resolution/super_resolution.py
+++ b/python/orca/example/learn/pytorch/super_resolution/super_resolution.py
@@ -304,7 +304,7 @@ elif opt.backend in ["ray", "spark"]:
     if exists(model_dir):
         shutil.rmtree(model_dir)
 
-    f_model_out_path = model_dir + "/" + "model_epoch_{epoch:02d}.pth"
+    f_model_out_path = join(model_dir, "model_epoch_{epoch:02d}.pth")
 
     stats = estimator.fit(data=train_data_creator,
                           epochs=opt.epochs,
--- a/python/orca/example/learn/pytorch/super_resolution/super_resolution.py
+++ b/python/orca/example/learn/pytorch/super_resolution/super_resolution.py
@@ -313,13 +313,13 @@ elif opt.backend in ["ray", "spark"]:
                                                      save_weights_only=True)])
     for epochinfo in stats:
         print("===> Epoch {} Complete: Avg. Loss: {:.4f}"
-                .format(epochinfo["epoch"], epochinfo["train_loss"]))
+              .format(epochinfo["epoch"], epochinfo["train_loss"]))
 
     val_stats = estimator.evaluate(data=validation_data_creator,
-                                    batch_size=opt.test_batch_size)
+                                   batch_size=opt.test_batch_size)
     print("===> Validation Complete: Avg. PSNR: {:.4f} dB, Avg. Loss: {:.4f}"
-            .format(10 * log10(1. / val_stats["val_loss"]), val_stats["val_loss"]))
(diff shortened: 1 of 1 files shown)


python/orca/example/learn/pytorch/super_resolution/super_resolution.py
```diff
@@ -302,21 +303,27 @@ def optim_creator(model, config):
...
+
+    val_stats = estimator.evaluate(data=validation_data_creator,
+                                    batch_size=opt.test_batch_size)
+    print("===> Validation Complete: Avg. PSNR: {:.4f} dB, Avg. Loss: {:.4f}"
+            .format(10 * log10(1. / val_stats["val_loss"]), val_stats["val_loss"]))
+
+    
+    model = estimator.get_model()
+    last_model_path = f_model_out_path.format(epoch=opt.epoch)
+    torch.save(model, last_model_path)
```
hkvision, line 325: Can we change to estimator.save?

python/orca/example/learn/pytorch/super_resolution/super_resolution.py
```diff
@@ -302,21 +303,27 @@ def optim_creator(model, config):
...
-        model_out_path = model_dir + "/" + "model_epoch_{}.pth".format(epoch)
-        model = estimator.get_model()
-        torch.save(model, model_out_path)
-        print("Checkpoint saved to {}".format(model_out_path))
+    f_model_out_path = model_dir + "/" + "model_epoch_{epoch:02d}.pth"
+
+    stats = estimator.fit(data=train_data_creator,
+                          epochs=opt.epochs,
+                          batch_size=opt.batch_size,
+                          callbacks=[ModelCheckpoint(filepath=f_model_out_path,
```
hkvision, line 311: does {epoch:02d} work to input here?

python/orca/example/learn/pytorch/super_resolution/super_resolution.py
```diff
@@ -299,24 +301,29 @@ def optim_creator(model, config):
...
-        makedirs(model_dir)
+    if exists(model_dir):
+        shutil.rmtree(model_dir)
 
-    for epoch in range(1, opt.epochs + 1):
-        stats = estimator.fit(data=train_data_creator, epochs=1, batch_size=opt.batch_size)
-        for epochinfo in stats:
-            print("===> Epoch {} Complete: Avg. Loss: {:.4f}"
-                  .format(epoch, epochinfo["train_loss"]))
+    f_model_out_path = model_dir + "/" + "model_epoch_{epoch:02d}.pth"
```
hkvision, line 307: use os.path.join?

python/orca/example/learn/pytorch/super_resolution/super_resolution.py
```diff
@@ -299,24 +301,29 @@ def optim_creator(model, config):
         }
     )
 
-    if not exists(model_dir):
-        makedirs(model_dir)
+    if exists(model_dir):
+        shutil.rmtree(model_dir)
```
hkvision, line 305: Does this work for hdfs? Or we only save the model to driver in this example?

leonardozcm: > Does this work for hdfs? Or we only save the model to driver in this example?

I believe this example do not involve hdfs

PR #7224, opened 2023-01-11
Nano: set jemalloc as default

MeouSker77:
## Description

Tcmalloc may cause segmentation fault, and may cause bug when running tensorflow inc quantization. So we use jemalloc as default allocator.

Jemalloc will use more memory than tcmalloc, so we replace ResNet50 with smaller models.


rnwang04 (approved):
LGTM

TheaperDeng (approved):
LGTM

PR #7234, opened 2023-01-12
[Orca] Optimize Spark Dataframe Input Performance with Openvino Estimator

sgwhat:
## Description
This PR applies `spark_df_to_pd_xshards`, which is optimized by arrow to decrease memory usage and improve the inference pipeline performance. 

Performance on Yarn (Spark 2.4.6 Version, without arrow, Minimum Configuration Requirements):
| Input | Executor Num | Executor Memory |
| --- | --- | ---  |
| 2000 Images DataFrame | 2 | 8g | 

Spark 3.1.3 Version still get oom error when converting xshards to spark dataframe (2000 images dataframe).

### 1. Why the change?
In the previous implementation, converting spark dataframe to xshards could cause a lot of memory and time, which may exceed the user's hardware capabilities.

### 2. User API changes
N/A

### 3. Summary of the change
Use arrow-optimized functionalities to improve pipeline performance.

### 4. How to test?
- [ ] Unit test
- [x] Application test

### DIFF
--- a/python/orca/src/bigdl/orca/learn/openvino/estimator.py
+++ b/python/orca/src/bigdl/orca/learn/openvino/estimator.py
@@ -27,7 +27,7 @@ from bigdl.dllib.utils import nest
 from bigdl.dllib.nncontext import init_nncontext
 from bigdl.orca.data.utils import spark_df_to_pd_sparkxshards
 from bigdl.orca.learn.utils import process_xshards_of_pandas_dataframe,\
-                                   add_predict_to_pd_xshards
+    add_predict_to_pd_xshards
 
 from openvino.inference_engine import IECore
 import numpy as np
--- a/python/orca/src/bigdl/orca/learn/openvino/estimator.py
+++ b/python/orca/src/bigdl/orca/learn/openvino/estimator.py
@@ -63,7 +63,6 @@ class OpenvinoEstimator(SparkEstimator):
     def predict(self,  # type: ignore[override]
                 data: Union["SparkXShards", "DataFrame", "np.ndarray", List["np.ndarray"]],
                 feature_cols: Optional[List[str]] = None,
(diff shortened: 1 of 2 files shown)


sgwhat:
Need to add a ut later.

python/orca/src/bigdl/orca/learn/openvino/estimator.py
```diff
@@ -60,6 +63,7 @@ def fit(self, data, epochs, batch_size=32, feature_cols=None, label_cols=None,
     def predict(self,  # type: ignore[override]
                 data: Union["SparkXShards", "DataFrame", "np.ndarray", List["np.ndarray"]],
                 feature_cols: Optional[List[str]] = None,
+                label_cols: Optional[List[str]] = None,
```
hkvision, line 66: label_cols doesn't make sense in predict, is it a must for process_xshards_of_pandas_dataframe? If so, can we make it a fixed value?

sgwhat: not a must, removed.

python/orca/src/bigdl/orca/learn/openvino/estimator.py
```diff
@@ -224,31 +227,28 @@ def predict_transform(dict_data, batch_size):
                                   "x in each shard should be a ndarray or a list of ndarray.")
             return feature_data
 
+        def update_result_shard(data):
```
hkvision, line 230: I think it is better to update in add_predict_to_pd_xshards so that it can be used for other estimators as well?
Also it is not general to always convert to list? Add some check if it is ndarray then convert to list?

sgwhat:
Merge it first to provide users a solution to conduct inference without arrow optimization.

PR #7236, opened 2023-01-12
[Nano] Fix `bigdl.nano.tf.keras.Model` api doc issue after V1Model and V2Model is added

Oscilloscope98:
## Description

Fix the problematic API doc of `bigdl.nano.tf.keras.Model` after merging #7211:

that is, the api doc is currently showing the api for `tf.keras.Model`. This is because we are not injecting functions for class `Model` itself. It seems that `tf.keras.Model` will decide whether to use `V1Model` or `V2Model` during code execution, while autodoc "import the corresponding module and extract the docstring of the given objects" with no execution.

https://github.com/analytics-zoo/nano/issues/239#issuecomment-1378517493

### How to test?
- [x] Unit test
- [x] Document test: https://yuwentestdocs.readthedocs.io/en/nano-tf-model-api-fix/doc/PythonAPI/Nano/tensorflow.html#bigdl-nano-tf-keras

### DIFF
--- a/python/nano/src/bigdl/nano/tf/keras/__init__.py
+++ b/python/nano/src/bigdl/nano/tf/keras/__init__.py
@@ -17,6 +17,6 @@ import tensorflow as tf
 
 
 from .Sequential import Sequential
-from .Model import Model
+from .Model import V2Model as Model
 from .inference.optimizer import InferenceOptimizer
 from .customized_training_utils import nano_bf16


TheaperDeng (approved):
LGTM, but please @yangw1234 have a look either.

BTW, I wonder what should users do when they want to build a model inherited from bigdl.tf.keras.Model

Should they
```python
# graph mode
class MyModel(V1Model):
```
```python
# eager mode
class MyModel(V2Model):
```
?

MeouSker77 (approved):
LGTM

yangw1234:
> LGTM, but please @yangw1234 have a look either.
> 
> BTW, I wonder what should users do when they want to build a model inherited from bigdl.tf.keras.Model
> 
> Should they
> 
> ```python
> # graph mode
> class MyModel(V1Model):
> ```
> 
> ```python
> # eager mode
> class MyModel(V2Model):
> ```
> 
> ?

My understanding is user should always inherit `V2Model` . At the time when the model is created, `keras` will automatically chose if it needs to change the class to `V1Model` according to current mode (in `version_utils.ModelVersionSelector`).

If user inherits the `V1Model`, it will always be a graph mode `V1Model`.

Here is the definition of `V2Model`
image

Here is the definition of `V1Model`
image

yangw1234 (approved):

PR #7238, opened 2023-01-12
Orca: rm unnecessary for loop in examples

hamham223:
### 1. Why the change?

remove unnecessary for loops in examples, users should use more epochs instead.

https://github.com/analytics-zoo/Telecom-CTC-E2E/issues/27


hkvision:
One failure is irrelevant.

PR #7250, opened 2023-01-13
Nano: replace `torch.no_grad()` with `torch.inference_mode()` in context manager 

rnwang04:
## Description

I found torch.inference_mode(True) is more recommended than torch.no_grad() now.
https://pytorch.org/docs/stable/generated/torch.inference_mode.html
_InferenceMode is a new context manager analogous to no_grad to be used when you are certain your operations will have no interactions with autograd (e.g., model training). Code run under this mode gets better performance by disabling view tracking and version counter bumps.Note that unlike some other mechanisms that locally enable or disable grad, entering inference_mode also disables to [forward-mode AD](https://pytorch.org/docs/stable/autograd.html#forward-mode-ad)_
As it disables more things than no_grad, I think it would provide slight better performance.
After my simple test, `torch.inference_mode(True)` has 5% acceleration compared with `torch.no_grad`.

### 1. Why the change?

https://github.com/analytics-zoo/nano/issues/256

### 2. User API changes

No change

### 3. Summary of the change 

replace `torch.no_grad()` with `torhch.inference_mode()` in context manager 

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/context_manager.py
+++ b/python/nano/src/bigdl/nano/pytorch/context_manager.py
@@ -25,17 +25,17 @@ class BaseContextManager(object):
     """
 
     def __init__(self, thread_num=None):
-        self.no_grad = torch.no_grad()
+        self.infer_mode = torch.inference_mode(mode=True)
         self.thread_num = thread_num
         self.original_thread_num = torch.get_num_threads()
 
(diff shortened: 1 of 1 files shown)


MeouSker77 (approved):
LGTM

PR #7252, opened 2023-01-13
Nano: fix save & load for inc ipex quantization

rnwang04:
## Description

INC ipex quantization has a known bug for save & load, that is it don't save related .pt model file.
In INC2.0, this problem has beed fixed, so in Nano, we should also update code to support save & load for INC ipex quantization (version >= 2.0).

### 1. Why the change?

support save & load for INC ipex quantization(version >= 2.0).

### 2. User API changes

New paramter `input_sample` in InferenceOpitmizer.load, which only works for INC ipex quantization.
```python
inc_model = InferenceOptimizer.quantize(model,
                                        precision='int8',
                                        accelerator=None,
                                        method='ipex',
                                        calib_data=data_loader)
with InferenceOptimizer.get_context(inc_model):
    inc_model(self.data_sample)

InferenceOptimizer.save(inc_model,dir_name)
new_model = InferenceOptimizer.load(tmp_dir_name, model,
                                    input_sample=next(iter(data_loader))[0])
    new_model(data_sample)
```

### 3. Summary of the change 

- support save & load for INC ipex quantization(version >= 2.0)
- related ut

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/deps/neural_compressor/inc_api_2.py
+++ b/python/nano/src/bigdl/nano/deps/neural_compressor/inc_api_2.py
@@ -153,15 +153,16 @@ def _quantize(
         timeout=timeout,
         max_trials=max_trials,
     )
+    # For ipex quantization, we should set backend = "ipex"
+    backend = "default" if framework != "pytorch_ipex" else "ipex"
     config = PostTrainingQuantConfig(
         accuracy_criterion=accuracy_criterion,
         tuning_criterion=tuning_criterion,
         approach=approach,
         inputs=inputs,
         outputs=outputs,
+        backend=backend
     )
-    if framework == 'pytorch_ipex':
-        config['backend'] = 'ipex'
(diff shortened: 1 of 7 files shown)


python/nano/src/bigdl/nano/deps/neural_compressor/inc_api_2.py
```diff
@@ -170,7 +172,7 @@ def new_collate_fn(batch):
 
     q_model = quantization.fit(
         model=model,
-        conf=q_conf,
+        conf=config,
```
MeouSker77, line 175: `config` may contain no `framework`, we should use `q_conf`

MeouSker77 (approved):
LGTM

python/nano/src/bigdl/nano/deps/neural_compressor/inc_api_2.py
rnwang04: Done

PR #7263, opened 2023-01-16
Nano: Use subprocess to calculate throughput for original model to achieve thread control

rnwang04:
## Description

Keras use `tf.config.threading.set_inter_op_parallelism_threads()` and `tf.config.threading.set_intra_op_parallelism_threads()` to achieve thread control, however, these two functions can only be called at the beginning of program running. so now in keras `InferenceOptimizer`'s `optimize`, we can't dynamically change the thread number of inference for model without accelerator.
To solve this problem, this PR uses subprocess to start a new subprocess for throughput calculation of original model. In this way, we can spawn a new subprocess which has specified `OMP_NUM_THREADS=thread_num` and setting `inter_op_parallelism_threads=1` and `intra_op_parallelism_threads=thread_num`.

For example, if we set `thread_num=4` when `optimize`, the CPU usage for original model when calculating throughput looks like:
image
which is inaccurate and obviously occupy more threads.
After this PR, the CPU usage looks like:
image

### 1. Why the change?

Use subprocess to calculate throughput for original model to achieve thread control.

### 2. User API changes

No change, just internal update for `InferenceOptimizer`'s `optimize`.

### 3. Summary of the change

- Use subprocess to calculate throughput for original model to achieve thread control
- update related ut for this case

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/tf/keras/inference/_worker.py
+++ b/python/nano/src/bigdl/nano/tf/keras/inference/_worker.py
@@ -42,7 +42,7 @@ if __name__ == "__main__":
     thread_num = sys.argv[2]
     if thread_num != 'None':
         thread_num = int(thread_num)  # type: ignore
-        tf.config.threading.set_inter_op_parallelism_threads(1)
+        tf.config.threading.set_inter_op_parallelism_threads(0)
         tf.config.threading.set_intra_op_parallelism_threads(thread_num)
     params = pickle.load(open(param_file, "rb"))
     latency = throughput_calculate_helper(**params)
--- a/python/nano/src/bigdl/nano/tf/keras/inference/_worker.py
+++ b/python/nano/src/bigdl/nano/tf/keras/inference/_worker.py
@@ -40,7 +40,12 @@ def throughput_calculate_helper(iterrun, func, model, input_sample):
 
 
 if __name__ == "__main__":
-    my_input = sys.argv[1:]
-    params = pickle.load(open(my_input[0], "rb"))
+    param_file = sys.argv[1]
+    thread_num = sys.argv[2]
+    if thread_num != 'None':
+        thread_num = int(thread_num)
+        tf.config.threading.set_inter_op_parallelism_threads(1)
+        tf.config.threading.set_intra_op_parallelism_threads(thread_num)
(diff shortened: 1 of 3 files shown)


python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
```diff
@@ -236,12 +238,37 @@ def optimize(self, model: Model,
...
+                                  "func": func_test,
+                                  "model": acce_model,
+                                  "input_sample": input_sample}
+                        with tempfile.TemporaryDirectory() as temp_dir:
+                            _filename = os.path.join(temp_dir, "params")
+                            cloudpickle.dump(params, open(_filename, "wb"))
+                            if thread_num is not None:
+                                my_env = os.environ.copy()
+                                my_env["OMP_NUM_THREADS"] = str(thread_num)
+                            else:
```
MeouSker77, line 254: Can we calculate throughput in current process if user does not specify `thread_num` ?

python/nano/src/bigdl/nano/tf/keras/inference/_worker.py
```diff
@@ -0,0 +1,49 @@
...
+    time_list = time_list[int(0.1 * iterrun): int(0.9 * iterrun)]
+    return np.mean(time_list) * 1000
+
+
+if __name__ == "__main__":
+    param_file = sys.argv[1]
+    thread_num = sys.argv[2]
+    if thread_num != 'None':
+        thread_num = int(thread_num)  # type: ignore
+        tf.config.threading.set_inter_op_parallelism_threads(1)
```
MeouSker77, line 45: Why set inter op parallel to 1 ? I think it is better to set it to 0 (0 means let tf choose the best configuration) or the number of sockets ?

MeouSker77 (approved):
others LGTM

MeouSker77 (commented):
We may fall back to calculating throughput in current process if starting sub-process caused error

python/nano/src/bigdl/nano/tf/keras/inference/_worker.py
rnwang04: done

PR #7269, opened 2023-01-16
Orca: fix the bug of unnecessary `import pyarrow` in orca/data/image

zpeng1898:
When importing orca `tensorboard callback` locally, the error shows that it needs `import pyarrow`. 
The line `import pyarrow as pa` is moved into specific functions below to avoid unnecessary importing.


hkvision (approved):

hkvision:
The failures are irrelevant.

PR #7270, opened 2023-01-16
Nano: fix attribute access of keras traced/quantized model

MeouSker77:
## Description

fix attribute access of keras traced/quantized model

### 4. How to test?
- [ ] Unit test


rnwang04 (approved):
LGTM

PR #7280, opened 2023-01-16
Add sampler for mmcv validation dataloader 

PatrickkZ:
## Description

- In MMCVRayEstimator, init ddp components when call fit/run, keep sync with #7050

### DIFF
--- a/python/orca/src/bigdl/orca/learn/pytorch/experimential/mmcv/mmcv_ray_estimator.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/experimential/mmcv/mmcv_ray_estimator.py
@@ -75,8 +75,7 @@ class MMCVRayEstimator(BaseRayEstimator):
                       max_epochs=max_epochs,
                       **kwargs)
 
-        if self.backend == 'ray' and not self.init_ddp_process:
-            self.setup_torch_ddp()
+        self.setup_torch_ddp()
 
         success, worker_stats = self._train_epochs(**params)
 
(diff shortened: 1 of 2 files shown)


python/orca/src/bigdl/orca/learn/pytorch/experimential/mmcv/mmcv_ray_estimator.py
```diff
@@ -74,6 +74,10 @@ def fit(self,
                       workflow=workflow,
                       max_epochs=max_epochs,
                       **kwargs)
+
+        if self.backend == 'ray' and not self.init_ddp_process:
+            self.setup_torch_ddp()
```
leonardozcm, line 79: this must be ray backend.

python/orca/src/bigdl/orca/learn/pytorch/experimential/mmcv/mmcv_ray_runner.py
```diff
@@ -169,6 +167,10 @@ def __init__(self,
     def setup_components(self) -> None:
         runner = self.mmcv_runner_creator(self.config)
         self._wrap_from_ebr(runner)
+        # add DistributedSampler for val data loader
+        for hook in self._hooks:
+            if isinstance(hook, EvalHook):
+                hook.dataloader = self.with_sampler(hook.dataloader)
```
leonardozcm, line 173: what if there are hooks only do evaluation on rank0?

leonardozcm (commented):
I believe we may collect the usages of "CTC Hooks" first?

leonardozcm (approved):

PR #7287, opened 2023-01-17
Nano: relax the accuracy restrictions of onnx quantized model

MeouSker77:
## Description

relax the accuracy restrictions of onnx quantized model


rnwang04 (approved):
LGTM

PR #7289, opened 2023-01-17
[PPML] spark I/O mysql example in gramine

ShanSimu:
## Description

add spark I/O mysql example in gramine

### 1. Why the change?

Add ability of connecting mysql


glorysdj (approved):
LGTM

PR #7298, opened 2023-01-17
Update gzip command in FashionMNIST tutorial

hkvision:
-k may not work for all versions of gzip. gzip 1.10 or 1.12 can work, 1.5 fails with error
```
gzip: invalid option -- 'k'
Try `gzip --help' for more information.
```

Removing -k will delete the original zipped files. So change mv to cp so that the original files are still in the cloned folder, and it doesn't matter whether the original files exist in the FashionMNIST/raw folder.


hkvision:
@zpeng1898 Take a look for the current commands?

sgwhat (approved):
Seems `gzip -d` is the best choice currently.

PR #7299, opened 2023-01-17
Chronos: update built-in dataset and import path in examples

plusbang:
## Description

Electricity dataset has been added in built-in datasets, update the usage. Besides, simplify the import path of `Evaluator` in examples.

### 4. How to test?
- [x] Unit test


python/chronos/src/bigdl/chronos/metric/__init__.py
```diff
@@ -13,3 +13,5 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+
+from .forecast_metrics import Evaluator
```
TheaperDeng, line 17: we have a tricky issue here since other metrics may be used here for anomaly detection or other tasks (e.g., classification) in future. We may still to this and make `Evaluator` support more metrics in future.

TheaperDeng (approved):
LGTM

PR #7303, opened 2023-01-17
Orca: Refactor use_tqdm with TqdmCallback

leonardozcm:
## Description

We implement TqdmCallback so we can make our torchrunner a clean structure and simplify our estimator interface.
We may provide more features to users with more callbacks like tqdmCallback.

## API Changes
old:
```python
Estimator.from_torch(*, use_tqdm=True,* )
```
new:
```python
Estimator.fit(*, callbacks=[TqdmCallback()],* )
```
## Tasks
- [x] Append TqdmCallback
- [x] Deprecate use_tqdm in torch estimator.
- [x] Rebase Branch

### DIFF
--- a/python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
@@ -164,7 +164,6 @@ class PyTorchPySparkEstimator(BaseEstimator):
             optimizer_creator=optimizer_creator,
             loss_creator=loss_creator,
             scheduler_creator=scheduler_creator,
-            use_tqdm=use_tqdm,
             config=copy.copy(self.config),
             metrics=metrics,
             size=self.num_workers,
--- a/python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
@@ -37,6 +37,7 @@ from bigdl.orca.data.file import get_remote_file_to_local, put_local_file_to_rem
 from bigdl.dllib.utils.common import get_node_and_core_number
 from bigdl.orca.learn.log_monitor import start_log_server, stop_log_server
 from bigdl.orca.learn.pytorch.callbacks.maincallback import make_only_mainCallback
+from bigdl.orca.learn.pytorch.callbacks.tqdm import TqdmCallback
 from bigdl.orca.learn.utils import find_free_port, find_ip_and_free_port
 from bigdl.dllib.utils.utils import get_node_ip
 from bigdl.dllib.utils.log4Error import invalidInputError
@@ -136,6 +137,7 @@ class PyTorchPySparkEstimator(BaseEstimator):
                               "Must provide a function for model_creator")
 
         self.model_dir = parse_model_dir(model_dir)
+        self.use_tqdm = use_tqdm
 
         self.model_creator = model_creator
         self.optimizer_creator = optimizer_creator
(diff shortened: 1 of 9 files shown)


leonardozcm:
8f6ed022-eb9d-4621-b20f-7bc77a0c24a0
@hkvision does this meet your expectation?

leonardozcm:
https://github.com/ray-project/ray/issues/9233#issuecomment-873707362

It seems pyramid prints can't be avoided by ray and tqdm, for ray will append sth. like "\n" to any line workers print automatically?

hkvision:
I think we'd better keep `use_tqdm=True` for consistency, and add TqdmCallback if True in the code by ourselves, TqdmCallback probably won't be directly used by users.
Otherwise, many examples need to change.

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
```diff
@@ -289,9 +290,10 @@ def fit(self,
         init_params.update(self.worker_init_params)
 
         # Check uniqueness of the MainCallback
-        if not callbacks:
-            callbacks = []
+        callbacks = callbacks or []
         make_only_mainCallback(callbacks)
+        if self.use_tqdm:
+            callbacks.append(TqdmCallback())
```
hkvision, line 296: Do we need to check if Tqdm is already in the callback (explicitly written by users, not recommended, but they can)
Or does it have issue when there are more than 1 Tqdm callback?

hkvision:
After using the callback, can we easily extend the progress bar to evaluate or predict as well?

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
leonardozcm: > Do we need to check if Tqdm is already in the callback (explicitly written by users, not recommended, but they can) Or does it have issue when there are more than 1 Tqdm callback?

There will be two or more progress bars if so? So this could be sticked as an issue.

leonardozcm:
> After using the callback, can we easily extend the progress bar to evaluate or predict as well?

Yes

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
hkvision: Or more generally, we need to better handle such cases:
- If users wrongly input two identical callbacks, we need to correctly handle this (probably just ignore one or raise an error)
- From our side, if users want to write their custom callback, they are supposed to only inherit MainCallBack right? Or they have the flexibility to overwrite other callbacks? In the latter case, we may need to consider whether we suggest them to write everything in one callback or split them according to piece-wise functions.

hkvision:
> > After using the callback, can we easily extend the progress bar to evaluate or predict as well?
> 
> Yes

Then let's add this?

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
leonardozcm: > Or more generally, we need to better handle such cases:
> 
> * If users wrongly input two identical callbacks, we need to correctly handle this (probably just ignore one or raise an error)

Yes, this could be fixed.

> * From our side, if users want to write their custom callback, they are supposed to only inherit MainCallBack right? Or they have the flexibility to overwrite other callbacks? In the latter case, we may need to consider whether we suggest them to write everything in one callback or split them according to piece-wise functions.

Users are allowed to write hooks more than only MainCallBack, however it would be easier if they write down all the logic they expect in MainCallBack. In my imagation it is more like our job to supply small hooks to achieve piece-wise functions?

hkvision: Then to make everything under our control, in our document, we just say users can only override MainCallBack?

leonardozcm: > Then to make everything under our control, in our document, we just say users can only override MainCallBack?


For most of the cases if users aren't going to modify a small part of hook, then all in one maincallback would be nice, but we may just **strongly suggest** them to write codes in only one maincallback

PR #7313, opened 2023-01-18
Nano: support evaluation related features for INC

rnwang04:
## Description

After we have upgrade INC version to 2.0, we have not supported `eval_metric` for this version, and we never support `eval function` for quantization until now, which is an effective tool to support a more flexible evaluation form. However, `eval_metric` and `eval_function` are important for the quantization quality.

### 1. Why the change?

To support `eval_metric` and `eval_function` in `InferenceOptimizer.quantize` for both PyTorch and Tensorflow.

### 2. User API changes

new parameter `eval_function` for `InferenceOptimizer.quantize`, here take PyTorch for example:
```python
def eval_func(model):
    acc = 0
    for sample, label in self.train_loader:
        logits = model(sample)
        pred = logits.argmax(dim=1)
        acc += torch.eq(pred, label).sum().float().item()
    return acc

qmodel = InferenceOptimizer.quantize(pl_model,
                                     calib_data=self.train_loader,
                                     eval_func=eval_func,
                                     accuracy_criterion={'relative': 0.8,
                                                         'higher_is_better': True},
                                     timeout=0,
                                     max_trials=10,
                                     thread_num=8)
```
At the same time, `metric` parameter now actually works for INC 2.0 version.
### 3. Summary of the change 

-  support eval function for INC 2.0 in PyTorch InferenceOptimizer
- support eval function for INC 2.0 in keras InferenceOptimizer
- support eval function for INC 1.x in PyTorch InferenceOptimizer
- support eval function for **INC 1.14.1 or higher version** in keras InferenceOptimizer
- support eval metric for INC 2.0 in PyTorch InferenceOptimizer
- support eval metric for INC 2.0 in keras InferenceOptimizer
- related uts for PyTorch
- related uts for keras
- update comment for `use_ipex` in PyTorch InferenceOptimizer
- update comment for `accuracy_criterion` in `quantize`

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/deps/neural_compressor/core/quantization.py
+++ b/python/nano/src/bigdl/nano/deps/neural_compressor/core/quantization.py
@@ -25,6 +25,7 @@ class BaseQuantization(Quantization):
                  approach='post_training_static_quant',
                  tuning_strategy='bayesian',
                  accuracy_criterion: dict = {'relative': 0.99, 'higher_is_better': True},
+                 eval_func=None,
                  timeout=0,
                  max_trials=1,
                  inputs=None,
@@ -49,6 +50,11 @@ class BaseQuantization(Quantization):
                                      allows relative accuracy loss: 1%. accuracy_criterion = {
                                      'absolute': 0.99, 'higher_is_better':False} means accuracy
                                      < 0.99 must be satisfied.
+        :param eval_func:       A evaluation function which only accepts model as input and return
+                                evaluation value. This parameter provides a higher degree of
+                                freedom than using eval_loader and metric. Default to None meaning
+                                no performance tuning, but it would be better give an evaluation
+                                function to get better quantization performance.
         :param timeout:     Tuning timeout (seconds). Default: 0,  which means early stop.
                             combine with max_trials field to decide when to exit.
         :param max_trials:  Max tune times. Default: 1.
@@ -73,6 +79,9 @@ class BaseQuantization(Quantization):
         cfg.model.inputs = inputs
         cfg.model.outputs = outputs
         super().__init__(qconf)
+        if eval_func is not None:
+            self._eval_func = eval_func
+            self._eval_func.builtin = True
 
     def post_training_quantize(self, model, calib_dataloader=None, metric=None):
         self.sanity_check_before_execution(model, calib_dataloader, metric)
--- a/python/nano/src/bigdl/nano/deps/neural_compressor/core/quantization.py
+++ b/python/nano/src/bigdl/nano/deps/neural_compressor/core/quantization.py
@@ -46,10 +46,12 @@ class BaseQuantization(Quantization):
                             Default: 'post_training_static_quant'.
         :param tuning_strategy:    'bayesian', 'basic', 'mse', 'sigopt'. Default: 'bayesian'.
         :param accuracy_criterion:  Tolerable accuracy drop.
-                                    accuracy_criterion = {'relative': 0.1, 'higher_is_better':True}
-                                     allows relative accuracy loss: 1%. accuracy_criterion = {
-                                     'absolute': 0.99, 'higher_is_better':False} means accuracy
-                                     < 0.99 must be satisfied.
+                                    accuracy_criterion = {'absolute':0.99, 'higher_is_better':False}
+                                    means accuracy loss must be smaller than 0.99. For example, if
+                                    higher_is_better is True, then this requires original metric
+                                    value subtract current metric value be smaller than 0.99.
+                                    accuracy_criterion = {'relative':0.1, 'higher_is_better':True}
+                                    allows relative accuracy loss: 10%.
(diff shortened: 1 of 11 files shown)


python/nano/src/bigdl/nano/deps/neural_compressor/inc_api_2.py
```diff
@@ -93,6 +97,7 @@ def _quantize(
     from neural_compressor.conf.pythonic_config import Config
     from neural_compressor.conf.config import QuantConf
     from neural_compressor.data import DataLoader
+    from neural_compressor.experimental.common import Metric as NCMetric
```
MeouSker77, line 100: why not `from neural_compressor.metric import Metric`

python/nano/src/bigdl/nano/deps/neural_compressor/inc_api_2.py
```diff
@@ -144,8 +191,7 @@ def new_collate_fn(batch):
         accuracy_criterion = AccuracyCriterion(
             higher_is_better=accuracy_criterion.get('higher_is_better', True),
             criterion=criterion,
-            # we calculate `tolerable_loss` by `1 - target_accuracy`
-            tolerable_loss=1.0 - accuracy_criterion[criterion]
+            tolerable_loss=float(accuracy_criterion[criterion])
```
MeouSker77, line 194: shall we rename `accuracy_criterion` to `loss_criterion` ?

python/nano/src/bigdl/nano/deps/neural_compressor/inc_api_2.py
rnwang04: This import is useless now, will delete this line.

python/nano/src/bigdl/nano/deps/neural_compressor/inc_api_2.py
rnwang04: I think we can still use `accuracy_criterion` as we have detailed comment `:param accuracy_criterion:  Tolerable accuracy drop, defaults to None meaning no accuracy control.`

python/nano/src/bigdl/nano/deps/neural_compressor/core/quantization.py
```diff
@@ -49,6 +50,11 @@ def __init__(self,
                                      allows relative accuracy loss: 1%. accuracy_criterion = {
```
MeouSker77, line 50: may update this comment, for my understanding, `'relative': 0.1` means accuracy >= 10% * original_accuracy ? and `'absolute: 0.99'` means accuracy > 0.99 ?

rnwang04: For exmaple, if `'absolute: 0.99'` and `higher_is_better` is True, this requires original metric value subtract current metric value be smaller than 0.99. For `'relative'`, it's similar.
Have revised corresponding comment.

MeouSker77: Oh, I understand. But why we set it to `relative: 0.99` by default, this allows 99% metric value loss ? INC's default value is `relative: 0.01`, why we use 0.99 ?

rnwang04: I think `relative: 0.99` actually means no tunning, and we set this value to ensure quantization always works and don't raise error like `can't find model meeting accuracy_criterion`.
I have tried` relative: 0.01` before and some ut of chronos will failed as can't get quantized model meet such strict accuracy limit.

MeouSker77: well, but I think 0.99 is too relaxed, if the metric value drops to 1%, I think this optimization is pointless. Have we tried other value? such as 0.9, 0.8 ?

rnwang04: `if the metric value drops to 1%, I think this optimization is pointless.`
Of course, in most cases, this will not happen.
However, in our uts, we always use random data for quantization, so that the accuracy of quantized model has no guarantee.
I will try 0.8 but such value will import a potential risk for action failing.

MeouSker77: > I will try 0.8 but such value will import a potential risk for action failing.

We can set its default value to 0.01, as for the UT, we can pass 0.99 to this parameter

rnwang04: Or there is another solution, we set default value to 0.01 as INC's, and adjust related failed uts. How do you think of?

MeouSker77: well, we have the same idea

MeouSker77 (approved):
LGTM

PR #7332, opened 2023-01-19
[PPML] Add Quick Check method to BigDL Remote Attestation Service

xiangyuT:
## Description

Provide a Quick Check method which responses welcome messages at root URL.
Refine document of BigDL Remote Attestation Service.


Uxito-Ada (approved):
lgtm

PR #7340, opened 2023-01-20
[PPML] Add check-service.sh for PPML services

SEULiyao:
## Description

Added a `check-services.sh` containing validations of all PPML services, to test whether these components are ready in current environment like what we do in trusted-real-time.

### How to test?
As described in `ppml/services/README.md`

### DIFF
--- a/ppml/services/README.md
+++ b/ppml/services/README.md
@@ -1,0 +1,15 @@
+##### trouble shooting 
+You can run `check-service.sh` to check whether the ppml services have been correctly initialized.
+
+To test a specific component, pass one or more argument to it among the following: "attestation", "bigDLKMS", "ehsm", "kmsUtils", "pccs".
+
+For example, you can run the following command to check ehsm and pccs service.
+```
+bash check-service.sh ehsm pccs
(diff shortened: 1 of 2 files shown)


ppml/services/README.md
```diff
@@ -0,0 +1,15 @@
+##### trouble shooting 
+You can run `check-service.sh` to check whether the ppml services have been correctly initialized.
+
+To test a specific component, pass one or more argument to it among the following: "attestation", "bigDLKMS", "ehsm", "kmsUtils", "pccs".
```
Uxito-Ada, line 4: modify to the same names as directory folders, like `bigdl-kms`, `kms-utils`

Uxito-Ada:
This should be merged after #7332

Uxito-Ada:
Note in readme that environment variables e.g. `${BIGDL_KMS_IP}` and `${EHSM_IP}` should be exported.
format like:
```bash
export BIGDL_KMS_IP=... # for bigdl-kms
export EHSM_IP=... # for ehsm
...
```

PR #7344, opened 2023-01-20
Remove ANALYTICS_ZOO_HOME in apps

hkvision:


hkvision:
Many places in notebook code uses ANALYTICS_ZOO_HOME...
Not fixing this at this moment.

PR #7363, opened 2023-01-29
Fix k8s tutorial style

hkvision:
Some places are not correctly indented and thus cause the wrong numbering in the page.


hkvision:
https://hkvision-k8s-0129.readthedocs.io/en/latest/doc/Orca/Tutorial/k8s.html#k8s-cluster

PR #7365, opened 2023-01-29
Nano: nano action acceleration and fix

MeouSker77:
## Description

- Try to accelerate nano's actions by splitting them into more smaller actions
- Fix a lot of bugs:
  - Tensorflow's openvino test is still using openvino 2022.2
  - Some UT are still using old API (i.e., `model.trace`)
  - Some UT are testing wrong things

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...


python/nano/test/openvino/tf/test_openvino_quantize.py
```diff
@@ -52,19 +52,24 @@ def test_model_quantize_openvino(self):
...
-        y_hat = openvino_quantized_model.predict(train_examples, batch_size=5)
-        assert y_hat.shape == (100, 10)
-
         preds = model.predict(train_examples)
         openvino_preds = openvino_quantized_model.predict(train_examples)
         np.testing.assert_allclose(preds, openvino_preds, rtol=1e-2)
 
         # Case 3: with config
-        openvino_quantized_model = InferenceOptimizer.quantize(model,
+        openvino_quantized_model = InferenceOptimizer.quantize(openvino_model,
```
rnwang04, line 60: why openvino_model

rnwang04 (approved):
others LGTM

python/nano/test/openvino/tf/test_openvino_quantize.py
MeouSker77: to save time, this will only call `trace` once

PR #7372, opened 2023-01-31
CI/CD:update friesian exampletest action condition

pinggao187:
## Description

code changes are only under python/friesian/example, should only run example test


.github/workflows/PR_validation.yml
```diff
@@ -77,19 +78,27 @@ jobs:
         with:
           filters: |
             friesian:
```
hkvision, line 80: change friesian to friesian-pytest?

PR #7375, opened 2023-01-31
Orca:  add initial_epoch in ray estimator.

lalalapotter:
## Description

This PR is to add `initial_epoch` argument for ray estimator in `estimator.fit()`.


lalalapotter:
We also remove the variable `self.epoch` in class `TFRunner`. Because in TensorFlow the epoch information is not stored in model instance, store it in TFRunner may lead to some misunderstandings. For example, as the experiment below described:
```python
estimator.fit(initial_epoch=2, epochs=4, ...)
# train model from epoch 2 to epoch 4, self.epoch = 2
estimator.fit(initial_epoch=2, epochs=4, ...)
# train model from epoch 2 to epoch 4, self.epoch = 4
estimator.fit(initial_epoch=2, epochs=4, ...)
# train model from epoch 2 to epoch 4, self.epoch = 6
```
As we can see, the self.epoch will lead to a misunderstanding that the model have been trained from 1 to 6 epochs, however the model is trained between 2-4 epochs for three times.

hkvision (approved):

Issue #7347, opened 2023-01-20
Need to manually downgrade protobuf after installing ray?


hkvision:
File "/root/anaconda3/envs/orca/lib/python3.7/site-packages/bigdl/orca/data/ray_xshards.py", line 19, in <module>
    import ray
  File "/root/anaconda3/envs/orca/lib/python3.7/site-packages/ray/__init__.py", line 91, in <module>
    import ray._raylet  # noqa: E402
  File "python/ray/_raylet.pyx", line 104, in init ray._raylet
  File "/root/anaconda3/envs/orca/lib/python3.7/site-packages/ray/exceptions.py", line 5, in <module>
    from ray.core.generated.common_pb2 import RayException, Language, PYTHON
  File "/root/anaconda3/envs/orca/lib/python3.7/site-packages/ray/core/generated/common_pb2.py", line 37, in <module>
    create_key=_descriptor._internal_create_key),
  File "/root/anaconda3/envs/orca/lib/python3.7/site-packages/google/protobuf/descriptor.py", line 755, in __new__
    _message.Message._CheckCalledFromGeneratedFile()
TypeError: Descriptors cannot not be created directly.
If this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0.
If you cannot immediately regenerate your protos, some other possible workarounds are:
 1. Downgrade the protobuf package to 3.20.x or lower.
 2. Set PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python (but this will use pure-Python parsing and will be much slower).

hkvision:
@hamham223 Please take a look.

hamham223:
https://bigdl.readthedocs.io/en/latest/doc/Nano/Overview/known_issues.html#protobuf-version-error

A similar problem in nano:

> protobuf version error
Now pip install ray[default]==1.11.0 will install google-api-core>=2.10.0, which depends on protobuf>=3.20.1. However, nano depends on protobuf==3.19.4, so you will meet this error if you install ray after bigdl-nano. The solution is pip install google-api-core==2.8.2 before installing ray.

dependency:

+ `bigdl-orca` does not require `protobuf`
+ `bigdl-orca[ray]` requires `ray[default]==1.9.2`, which in turn requires `protobuf>=3.15.3`. If protobuf is not installed yet, it will install the newest version 4.21.12 by default, and the user need to down-grade the version.
[...]

PR #7377, opened 2023-01-31
Orca: add protobuf version requirement

hamham223:
### 1. Why the change?
Fix https://github.com/intel-analytics/BigDL/issues/7347

Specify `protobuf` version after install ray.


hkvision (approved):

python/orca/src/setup.py
```diff
@@ -37,7 +37,8 @@
 VERSION = open(os.path.join(bigdl_home, 'python/version.txt'), 'r').read().strip()
 
 RAY_DEP = ['ray[default]==1.9.2', 'aiohttp==3.8.1', 'async-timeout==4.0.1', 'aioredis==1.3.1',
-           'hiredis==2.0.0', 'setproctitle', 'psutil', 'prometheus-client==0.11.0']
+           'hiredis==2.0.0', 'setproctitle', 'psutil', 'prometheus-clienst==0.11.0',
```
TheaperDeng, line 40: I believe this should be `prometheus-client`?

TheaperDeng (commented):
Could be a severe problem

PR #7383, opened 2023-01-31
Nano: fix keras attribute access

MeouSker77:
## Description

Fix attribute access of keras traced/quantized model

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/tf/keras/amp/bfloat16.py
+++ b/python/nano/src/bigdl/nano/tf/keras/amp/bfloat16.py
@@ -19,31 +19,16 @@ from tempfile import TemporaryDirectory
 import tensorflow as tf
 
 
-class BF16Model(AcceleratedKerasModel):
-    def __init__(self, model, **kwargs):
-        super().__init__(model, precision=tf.bfloat16)
-        # TODO: check bf16 isa
-        policy_bf16 = mixed_precision.Policy('mixed_bfloat16')
-        for layer in model.layers:
-            layer._dtype_policy = policy_bf16
-        with TemporaryDirectory() as temp_dir:
-            model.save(temp_dir)
-            bf16_model = tf.keras.models.load_model(temp_dir)
-        self.model = bf16_model
+def BF16Model(model):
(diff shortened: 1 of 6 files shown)


python/nano/src/bigdl/nano/tf/keras/amp/bfloat16.py
```diff
@@ -19,31 +19,16 @@
...
-        return status
-
-    def _save_model(self, path):
-        self.model.save(path)
-
-    @staticmethod
-    def _load(path):
-        return BF16Model(tf.keras.models.load_model(path))
+def load_bf16_model(path):
+    return tf.keras.models.load_model(path)
```
rnwang04, line 34: Here use default save behavior for normal keras model.
However, we just save mode weight in that case.
I am not sure after such save/load, whether the model is still a mixed bf16 model.

rnwang04: However, action can pass for current code, we may fix the saving behavior in next PR.

rnwang04 (approved):

python/nano/src/bigdl/nano/tf/keras/amp/bfloat16.py
TheaperDeng: > Here use default save behavior for normal keras model. However, we just save mode weight in that case. I am not sure after such save/load, whether the model is still a mixed bf16 model.

I think it's a good point. We may need to do some manual test and add something after
https://github.com/intel-analytics/BigDL/blob/9aae3013910eda8cac8c72e4b81364bf2249969a/python/nano/test/tf/inference/test_trace_and_quantize.py#L201-L215
to check if the loaded model still work in bf16.

Or maybe just `return BF16Model(tf.keras.models.load_model(path))` right? Since `BF16Model` is now just a function

MeouSker77: yes, this UT is testing wrong things now, we'll fix it and test more things in the next PR

PR #7390, opened 2023-02-01
add docker SNAPSHOT tag

Le-Zheng:
## Description

Add docker SNAPSHOT tag


hkvision:
Will we still override the SNAPSHOT every day? How can we get the correspondence between SNAPSHOT and pip nightly version?

Le-Zheng:
> Will we still override the SNAPSHOT every day? How can we get the correspondence between SNAPSHOT and pip nightly version?

@glorysdj @pinggao187 can we release nightly image tagged like intelanalytics/bigdl-k8s
:2.3.0b20230131 which is the same version as pypi? I'm a bit worried that it will increase space on Dockerhub.

hkvision:
What's the status of this PR?

Le-Zheng:
This pr will be merge for adding "SNAPSHOT" tag and will crate another pr for adding monthly docker release.

PR #7392, opened 2023-02-01
Nano: add libjemalloc.dylib

MeouSker77:
## Description

Add `libjemalloc.dylib` and update `.gitignore`, and continue to fix killed

- `*.so.*` for `libturbojpeg.so.0.2.0`
- `*.dylib` for `libjemalloc.dylib`


TheaperDeng (approved):
LGTM

PR #7396, opened 2023-02-01
[Nano] Improve How-to Guides Navigations

Oscilloscope98:
## Description

Improve navigations for how-to guides

### 1. Why the change?

As we are adding more guides, the left sides bar including more and more titles with no classification, which would be very time-consuming for readers to find the guide they want, especially when they has already opened a guide and leaving away the index page.

### 2. Summary of the change 
- Classified the how-to guides titles in the left side nav bar
   | before | after |
   |:--------|:------|
   |image|image|
- Added index page for each sub/subsubsections. For example, for the training optimization section, we have
  
- Moved the install related how-to guides inside a Install folder, and fixed related links
- Removed deprecated option `enable_auto_doc_ref` for `recommonmark`, which resolved ~200 build warnings

**Note** that after this navigation improvement, there are one more place we need to change when we are adding new guides, that is, the index page in the specific folder where the new nblink file is added.

### 3. How to test?
- [x] Document test: https://yuwentestdocs.readthedocs.io/en/improve-nano-howto-nav/doc/Nano/Howto/index.html


TheaperDeng (approved):
LGTM

PR #7398, opened 2023-02-01
Nano: fix keras traced/quantized model evaluate

MeouSker77:
## Description

Nano: fix keras traced/quantized model's `evaluate` method

### 1. Why the change?

When using `CategoricalAccuracy` as metric, calling `evaluate` will raise error.

### 2. User API changes

N/A

### 3. Summary of the change 

- `tf.py_function` will loss the shape info of the returned Tensor. We first save the shape of output, and using `set_shape` method to recover the shape info of the Tensor returned by `tf.py_function`.
- Add more test about `evaluate` method.

### 4. How to test?
- [x] Unit test


rnwang04 (approved):
LGTM

PR #7418, opened 2023-02-03
Nano: update comment for input spec

rnwang04:
## Description

Update comment for input spec in keras InferenceOptimizer to make it more clear

### How to test?
- [x] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
+++ b/python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
@@ -136,8 +136,10 @@ class InferenceOptimizer(BaseInferenceOptimizer):
                   If x is a dataset, y will be ignored (since targets will be obtained from x).
         :param validation_data: (optional) An unbatched tf.data.Dataset object for accuracy
                evaluation. This is only needed when users care about the possible accuracy drop.
-        :param input_spec: A (tuple or list of) ``tf.TensorSpec`` defining the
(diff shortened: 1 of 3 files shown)


python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
```diff
@@ -136,8 +136,10 @@ def optimize(self, model: Model,
                   If x is a dataset, y will be ignored (since targets will be obtained from x).
         :param validation_data: (optional) An unbatched tf.data.Dataset object for accuracy
                evaluation. This is only needed when users care about the possible accuracy drop.
-        :param input_spec: A (tuple or list of) ``tf.TensorSpec`` defining the
-                           shape/dtype of the input.
+        :param input_spec: (optional) A (tuple or list of) ``tf.TensorSpec``
+                           defining the shape/dtype of the input. This is only required when
+                           you have a custom Keras model(no input/output layer is explicitly
```
Oscilloscope98, line 141: A space seems to be missing between before the "("

Oscilloscope98 (approved):
Others LGTM :)

python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
rnwang04: done

MeouSker77 (approved):
LGTM

PR #7420, opened 2023-02-03
Nano: fix bug of context manager

rnwang04:
## Description

Fix bug when the second context manager is None and update related ut

### How to test?
- [x] Unit test


Oscilloscope98 (approved):
LGTM

PR #7422, opened 2023-02-03
Chronos: init bigdl-nano in nightly tests

plusbang:
## Description

For most users, it's common to use `source bigdl-nano-init` to enable local accelerations.
Test this in nightly tests.
Besides, remove unnecessary dependencies in nightly tests.

### 4. How to test?
- [x] Unit test


TheaperDeng (approved):
LGTM

Issue #7319, opened 2023-01-19
Update usage of Depracate API (get_weights) in TensorFlow 2 Optimizer.


lalalapotter:
Encountered following error when I run LSTM model with orca pyspark tf2 estimator:

```shell
Traceback (most recent call last):
  File "lstm_pollution.py", line 33, in <module>
    steps_per_epoch= df.count() // batch_size)
  File "/home/cengguang/anaconda3/envs/bigdl2.1/lib/python3.7/site-packages/bigdl/orca/learn/tf2/pyspark_estimator.py", line 199, in fit
    lambda iter: transform_func(iter, init_params, params)).collect()
  File "/home/cengguang/anaconda3/envs/bigdl2.1/lib/python3.7/site-packages/pyspark/rdd.py", line 816, in collect
    sock_info = self.ctx._jvm.PythonRDD.collectAndServe(self._jrdd.rdd())
  File "/home/cengguang/anaconda3/envs/bigdl2.1/lib/python3.7/site-packages/py4j/java_gateway.py", line 1257, in __call__
    answer, self.gateway_client, self.target_id, self.name)
  File "/home/cengguang/anaconda3/envs/bigdl2.1/lib/python3.7/site-packages/pyspark/sql/utils.py", line 63, in deco
    return f(*a, **kw)
  File "/home/cengguang/anaconda3/envs/bigdl2.1/lib/python3.7/site-packages/py4j/protocol.py", line 328, in get_return_value
    format(target_id, ".", name), value)
py4j.protocol.Py4JJavaError: An error occurred while calling z:org.apache.spark.api.python.PythonRDD.collectAndServe.
: org.apache.spark.SparkException: Job aborted due to stage failure: Could not recover from a failed barrier ResultStage. Most recent failure reason: Stage failed because barrier task ResultTask(12, 0) finished unsuccessfully.
org.apache.spark.api.python.PythonException: Traceback (most recent call last):
  File "/home/cengguang/anaconda3/envs/bigdl2.1/lib/python3.7/site-packages/pyspark/python/lib/pyspark.zip/pyspark/worker.py", line 377, in main
    process()
  File "/home/cengguang/anaconda3/envs/bigdl2.1/lib/python3.7/site-packages/pyspark/python/lib/pyspark.zip/pyspark/worker.py", line 372, in process
    serializer.dump_stream(func(split_index, iterator), outfile)
[...]

PR #7423, opened 2023-02-03
Orca: Support for Tensorflow 2.11.0 Breaking Changes to Optimizer

sgwhat:
## Description

See https://github.com/intel-analytics/BigDL/issues/7319, tf team makes a breaking change to optimize in version 2.11.0.

### 1. Why the change?

`tf.keras.optimizers` does not support `get_weights` anymore.

### 2. User API changes

N/A

### 3. Summary of the change 

Seems optimizer weight is not needed when enabling `model_dir`, so we could remove it currently and find another way to instead.

### 4. How to test?
- [x] Unit test


sgwhat:
`opt.weights()` is no longer needed, since we will use `model_saved_path` instead for continuous training specially when enabling `model_dir`.

lalalapotter (approved):

sgwhat:
Some errors may not related to this PR.

PR #7436, opened 2023-02-06
Chronos: remove redundant `no_grad` and change internal impl to `inference_mode`

TheaperDeng:
### 1. Why the change?
`torch.inference_mode` has been involved by https://github.com/intel-analytics/BigDL/pull/7250 and this PR is to make Nano and Chronos aligned.

Recent benchmark results shows that

- code within `torch.no_grad()` for PT1.13(our default version) is ~10% slower than 1.12 (TCN is tested for this case)
- `torch.inference_mode` could effectively solve the problem

### 2. User API changes
nothing

### 3. Summary of the change 

- Some internal implementation is changed to `torch.inference_mode` (including utils and autoformer)
- Some external implementation's (e.g., benchmark tool & examples) `torch.no_grad` are deleted
- TCMF is not changed since we currently does not want to change legacy code.

### 4. How to test?
- [ ] Unit test


plusbang (approved):
LGTM

PR #7438, opened 2023-02-06
Nano/Chronos: add `packaging` to nano's setup to avoid no module error

TheaperDeng:
### 1. Why the change?
There 2 kinds of users may meet `No module named 'packaging'` error (raised from nano's src) when they normally use Nano/Chronos
1. M-series chip users will install `pip install bigdl-nano[tensorflow]`
2. Chronos users who does not need deep-learning framework/inference optimization

### 2. User API changes
nothing

### 3. Summary of the change 
1. add a new (small) dependency to nano
2. fix some platform arch statement

### 4. How to test?
- [ ] Unit test

### 5. New dependencies
- [x] New Python dependencies
       - `packaging`


MeouSker77 (approved):
LGTM

Issue #7456, opened 2023-02-07
What data should i pass while working with AutoFormers


pankajr141:
Working with most of the forecasters seems straight forward.

Eg
<pre>
model = TCNForecaster(past_seq_len=train_x.shape[1],
                      future_seq_len=train_y.shape[1],
                      input_feature_num=train_x.shape[-1],
                      output_feature_num=train_y.shape[-1],
                      num_channels = num_channels,
                      kernel_size=kernel_size,
                      normalization=False,
                      dropout=dropout,
                      lr=lr,
                      dummy_encoder = False,
                      seed=0)
model.fit((train_x, train_y), batch_size=batch_size, epochs=epochs)
</pre>

However in case of AutoFormers it seems model.fit  input data needs 4 numpy array instead of 2.

<pre>
model.fit((train_x, train_y, train_x_enc, train_y_enc), batch_size=batch_size, epochs=epochs)`
</pre>

Came to this conclusion based on errors and below documentation

https://bigdl.readthedocs.io/en/latest/doc/PythonAPI/Chronos/forecasters.html
<b>data – train data, as numpy ndarray tuple (x, y, x_enc, y_enc)</b>

also in code i can see that 4 numpy arrays are being used.
https://github.com/intel-analytics/BigDL/blob/main/python/chronos/src/bigdl/chronos/forecaster/autoformer_forecaster.py

From documentation it is not clear how can we generate those additional dataset, can you please point to any link which can help clarify it.

TheaperDeng:
Yeah, autoformer needs 2 **more** array as input compared to other models. We provided methods in `TSDataset` to generate them directly.

If you prefer to use roll and to_numpy to get numpy ndarrays, please find an `time_enc` parameter in following document
https://bigdl.readthedocs.io/en/latest/doc/PythonAPI/Chronos/tsdataset.html#bigdl.chronos.data.tsdataset.TSDataset.roll

If you prefer to use to_torch_data_loader to get pytorch dataloader, please find an `time_enc` parameter in following document
https://bigdl.readthedocs.io/en/latest/doc/PythonAPI/Chronos/tsdataset.html#bigdl.chronos.data.tsdataset.TSDataset.to_torch_data_loader

This use-case may help you as well: https://github.com/intel-analytics/BigDL/tree/main/python/chronos/use-case/electricity

pankajr141:
thnx for the update

PR #7461, opened 2023-02-08
Chronos: update some missing description in autoformerforecaster api doc

TheaperDeng:
## Description

update some missing description in autoformerforecaster api doc per #7456


rnwang04 (approved):
LGTM

PR #7466, opened 2023-02-08
Nano: use context manager automatically inside multi-instance inference

MeouSker77:
## Description

Use context manager automatically inside multi-instance inference


python/nano/src/bigdl/nano/pytorch/inference/multi_instance.py
```diff
@@ -42,7 +43,14 @@ def forward(self, input_data: Union[DataLoader, List]) -> List:
...
             # run inference in current process directly
-            with torch.no_grad():
+            if isinstance(self.model, AcceleratedLightningModule):
+                # If this model is optimized by `InferenceOptimizer`
+                # we should use context manager
+                from bigdl.nano.pytorch import InferenceOptimizer
+                context = InferenceOptimizer.get_context(self.model)
+            else:
+                context = torch.inference_mode(True)
+            with context:
```
rnwang04, line 53: maybe just `with InferenceOptimizer.get_context(self.model):` and remove above lines

PR #7468, opened 2023-02-08
Nano: support CPU fp16 when accelerator is openvino

rnwang04:
## Description

support CPU fp16 when accelerator is openvino for PyTorch and Keras.

### 1. Why the change?

To support CPU fp16 when accelerator is openvino.

### 2. User API changes
Now precision `fp16` can be used when device is CPU.
```python
ov_fp16_model = InferenceOptimizer.quantize(model,
                                            accelerator='openvino',
                                            input_sample=x,
                                            precision='fp16')
```

### 3. Summary of the change 

- support CPU fp16
- add ut for this
- fix some wrong ut

### 4. How to test?
- [x] Unit test


MeouSker77 (approved):

MeouSker77:
LGTM

PR #7478, opened 2023-02-08
Nano: check affinity core number and omp thread number for both PT and TF

TheaperDeng:
### 1. Why the change?
Users will use nano's system variable setting, but use `taskset`, `numactl`, `KMP_AFFINITY` as well, a typical usage is
```bash
source bigdl-nano-init  # this will set OMP_NUM_THREADS to physical core number, e.g., 16
taskset -c 0-3 python test.py  # this will only use 4 cores
```
If openmp thread number is set larger than affinity core number, the program will be run really slow.

### 2. User API changes
nothing. A warning will be raised to users like below if openmp thread number is set larger than affinity core number
```bash
*****************Nano performance Suggestions*****************
CPU Affinity is set to this program and 2 cores are binded. While PyTorch OpenMP code block will use 3 cores, which may cause severe performance downgrade. Please set `OMP_NUM_THREADS` to 2.
CPU Affinity is set to this program and 2 cores are binded. While Tensorflow OpenMP code block will use 3 cores, which may cause severe performance downgrade. Please set `OMP_NUM_THREADS` to 2.

*****************Nano performance Suggestions*****************
```

### 3. Summary of the change 
Add the check to `__init__.py` of `bigdl.nano.tf` and `bigdl.nano.pytorch`.

### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/utils/common/affinity_core.py
+++ b/python/nano/src/bigdl/nano/utils/common/affinity_core.py
@@ -1,0 +1,44 @@
+#
+# Copyright 2016 The BigDL Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import os
+
+
+def get_affinity_core_num():
+    # When using proclist to bind cores,
+    # `os.sched_getaffinity(0)` will return only the first bound core,
+    # so we need to parse KMP_AFFINITY manually in this case
+    KMP_AFFINITY = os.environ.get("KMP_AFFINITY", "")
+    if "proclist" not in KMP_AFFINITY:
+        affinity_core_num = len(os.sched_getaffinity(0))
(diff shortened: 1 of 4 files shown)


MeouSker77 (approved):
LGTM

python/nano/src/bigdl/nano/utils/common/affinity_core.py
```diff
@@ -0,0 +1,44 @@
...
+import os
+
+
+def get_affinity_core_num():
+    # When using proclist to bind cores,
+    # `os.sched_getaffinity(0)` will return only the first bound core,
+    # so we need to parse KMP_AFFINITY manually in this case
+    KMP_AFFINITY = os.environ.get("KMP_AFFINITY", "")
+    if "proclist" not in KMP_AFFINITY:
+        affinity_core_num = len(os.sched_getaffinity(0))
```
MeouSker77, line 26: I remember `compact,1,0` will also cause `len(os.sched_getaffinity(0))` returns 1

python/nano/src/bigdl/nano/utils/common/affinity_core.py
```diff
@@ -0,0 +1,44 @@
...
+            for n in proclist_str.split(','):
+                if '-' not in n:
+                    core_list.append(int(n))
+                else:
+                    start, end = n.split('-')
+                    core_list.extend(range(int(start), int(end) + 1))
+            affinity_core_num = len(core_list)
+        except Exception as _e:
+            warnings.warn(f"Failed to parse KMP_AFFINITY: '{KMP_AFFINITY}'."
+                          f" Will use default thread number: {preset_thread_nums}")
```
MeouSker77, line 42: may remove this line, we won't set thread number now

PR #7484, opened 2023-02-09
Orca: Fix conflicts about use_tqdm and TqdmCallback

leonardozcm:
## Description

If users wrongly input two identical tqdmcallbacks(users pass use_tqdm=true and tqdmcallback as the same time), we need to correctly handle this (Just ignore one)


python/orca/src/bigdl/orca/learn/pytorch/callbacks/tqdm.py
```diff
@@ -27,6 +27,13 @@
     pass
 
 
+def is_tqdm_exited(callbacks):
```
hkvision, line 30: not exited, change to exists

hkvision (approved):

PR #7487, opened 2023-02-09
Orca: remove unnecessary import in src

hamham223:
### 1. Why the change?

Remove those unused `import` in orca src files.

### 2. How to test?

- [x] Orca Unit Test

### DIFF
--- a/python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_worker.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_worker.py
@@ -29,11 +29,9 @@
 # limitations under the License.
 
 
-from pyspark import BarrierTaskContext
 from bigdl.orca.learn.pytorch.torch_runner import TorchRunner
 import torch.distributed as dist
 import logging
-from bigdl.orca.learn.utils import save_pkl
 import os
 import tempfile
 import copy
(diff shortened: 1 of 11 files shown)


python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_worker.py
```diff
@@ -29,7 +29,6 @@
 # limitations under the License.
 
 
-from subprocess import call
 from pyspark import BarrierTaskContext
```
hkvision, line 32: this is also duplicated?

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_worker.py
```diff
@@ -29,7 +29,6 @@
...
 from bigdl.orca.learn.pytorch.torch_runner import TorchRunner
 import torch.distributed as dist
```
hkvision, line 34: logging and save pkl are not needed?

python/orca/src/bigdl/orca/learn/tf2/ray_estimator.py
```diff
@@ -25,7 +24,6 @@
 
 from bigdl.dllib.utils import log4Error
 from bigdl.dllib.utils.file_utils import is_local_path
-from bigdl.orca import OrcaContext
 from bigdl.orca.data.file import enable_multi_fs_save, enable_multi_fs_load
 from bigdl.orca.data.ray_xshards import RayXShards
 from bigdl.orca.learn.dl_cluster import RayDLCluster
```
hkvision, line 29: Any is not needed

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_worker.py
hamham223: `logging` is used in Line 47. `save_pkl` is duplicated

PR #7493, opened 2023-02-09
Nano: fix multi-instance on Windows

MeouSker77:
## Description

Fix multi-instance bugs on Windows:
- all logical cores will be mapped to the same physical core
- specifying cores used by each process will cause error
- catch the `OSError` exception caused by `import torch` after `import tensorflow`

### 4. How to test?
- [x] Manually test

### DIFF
--- a/python/nano/src/bigdl/nano/utils/common/schedule.py
+++ b/python/nano/src/bigdl/nano/utils/common/schedule.py
@@ -41,8 +41,10 @@ def get_cpu_info():
         logical_cores = int(subprocess.check_output(
             get_logical_core_args, universal_newlines=True).splitlines()[4].split("=")[1])
         for i in range(logical_cores):
-            for j in range(physical_cores):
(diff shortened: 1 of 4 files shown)


python/nano/src/bigdl/nano/utils/common/schedule.py
```diff
@@ -41,8 +41,10 @@ def get_cpu_info():
         logical_cores = int(subprocess.check_output(
             get_logical_core_args, universal_newlines=True).splitlines()[4].split("=")[1])
         for i in range(logical_cores):
-            for j in range(physical_cores):
-                get_physical_core[i] = j
+            # We don't know how to query which cores are physical cores on Windows,
+            # so we use all logical cores
+            get_physical_core[i] = i
+            get_socket[i] = 0
```
rnwang04, line 47: This line seems to be repeated with the following two lines. Maybe you can delete the following two lines.

PR #7508, opened 2023-02-13
Nano: update save & load and related ut

rnwang04:
## Description

Update loading logic for PyTorch and Keras, now original model is not necessary for accelerator=None, only when you want to contain original attributes, you can pass it.

### 1. Why the change?

In the past, for keras, user must pass in original model when loading model, however, in PyTorch, we can't pass in original model to obtain some original attributes.
Now we want to unify the two behaviors: original model is not necessary for accelerator=None, only when you want to contain original attributes, you can pass it.

### 2. User API changes
Update `InferenceOptimizer.load` for PyTorch and keras.
```python
# save & load without original model, for accelerator != None
InferenceOptimizer.save(openvino_model, dir_name)
load_model = InferenceOptimizer.load(dir_name) 

# save & load with original model, for accelerator == None or you wan to contain original attributes
InferenceOptimizer.save(openvino_model, tmp_dir_name)
load_model = InferenceOptimizer.load(tmp_dir_name, model=model)
```

### 3. Summary of the change 

- Update loading behavior for PyTorch and Keras
- Update related ut

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
+++ b/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
@@ -1160,9 +1160,14 @@ class InferenceOptimizer(BaseInferenceOptimizer):
 
         :param path: Path to model to be loaded. Path should be a directory.
         :param model: Required FP32 model to load pytorch model, it is needed if:
-               1. you accelerated the model with accelerator=None by
-               InferenceOptimizer.trace/InferenceOptimizer.quantize.
-               2. you want to the loaded model contains the attributes of original model.
+               1. you accelerate the model with accelerator=None by
+               InferenceOptimizer.trace()/InferenceOptimizer.quantize().
+               2. you accelerate the model with InferenceOptimizer.optimize() and
+               get_model()/get_best_model(), and the best method or the method you
+               specify don't contain accelerator 'onnxruntime'/'openvino'/'jit'.
+               If you are not sure what optimization method is used, we recommend that
+               you always pass in the original model for this case.
+               3. you want to the loaded model contains the attributes of original model.
         :param input_sample: Input sample for your model, could be a Tensor or a tuple.
                Only valid for inc ipex quantization model, otherwise will be ignored.
         :param inplace: whether to perform inplace optimization. Default: ``False``.
--- a/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
+++ b/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
@@ -1159,10 +1159,10 @@ class InferenceOptimizer(BaseInferenceOptimizer):
         Load a model from local.
 
         :param path: Path to model to be loaded. Path should be a directory.
-        :param model: Required FP32 model to load pytorch model, it is needed if you accelerated
-               the model with accelerator=None by InferenceOptimizer.trace/
-               InferenceOptimizer.quantize. model should be set to None if you choose
-               accelerator="onnxruntime"/"openvino"/"jit".
+        :param model: Required FP32 model to load pytorch model, it is needed if:
+               1. you accelerated the model with accelerator=None by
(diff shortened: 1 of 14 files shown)


python/nano/src/bigdl/nano/pytorch/utils.py
```diff
@@ -144,24 +145,26 @@ def load_model(path, model: pl.LightningModule = None, input_sample=None,
...
+        result = load_ipexjit_model(path, model, inplace=inplace)
     if model_type == 'PytorchIPEXJITBF16Model':
-        return load_ipexjitbf16_model(path, model, inplace=inplace)
+        result = load_ipexjitbf16_model(path, model, inplace=inplace)
     if model_type == 'PytorchIPEXQuantizationModel':
-        return load_ipex_quantization_model(path, model, inplace=inplace)
+        result = load_ipex_quantization_model(path, model, inplace=inplace)
     if model_type == 'BF16Model':
         return load_bf16_model(path, model)
+    if result is not None:
```
MeouSker77, line 163: may give some error message if `result is None`

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -1159,10 +1159,10 @@ def load(path, model: Optional[nn.Module] = None, input_sample=None,
         Load a model from local.
 
         :param path: Path to model to be loaded. Path should be a directory.
-        :param model: Required FP32 model to load pytorch model, it is needed if you accelerated
-               the model with accelerator=None by InferenceOptimizer.trace/
-               InferenceOptimizer.quantize. model should be set to None if you choose
-               accelerator="onnxruntime"/"openvino"/"jit".
+        :param model: Required FP32 model to load pytorch model, it is needed if:
+               1. you accelerated the model with accelerator=None by
+               InferenceOptimizer.trace/InferenceOptimizer.quantize.
```
MeouSker77, line 1164: How can user know whether to pass this parameter if they get optimized model by `.optimize()` ?

MeouSker77 (approved):
others LGTM

python/nano/src/bigdl/nano/pytorch/utils.py
rnwang04: I think should not raise error here, as if result is None, it maybe a normal nn.Module and it will enter below logic.

MeouSker77: yes, you're right

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: Add more comment for this, thanks!

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
```diff
@@ -424,15 +424,28 @@ def save(model: pl.LightningModule, path):
...
+               Only valid for inc ipex quantization model, otherwise will be ignored.
+        :param inplace: whether to perform inplace optimization. Default: ``False``.
+        :param device: A string represents the device of the inference. Default to None.
+               Only valid for openvino model, otherwise will be ignored.
         :return: Model with different acceleration(None/OpenVINO/ONNX Runtime/JIT) or
                  precision(FP32/FP16/BF16/INT8).
         """
-        return load_model(path, model)
+        return load_model(path, model, input_sample=input_sample,
+                          inplace=inplace, device=device)
```
TheaperDeng, line 451: just a question, in what case do we recommend use `Trainer.save/load`

TheaperDeng (approved):
LGTM

python/nano/src/bigdl/nano/pytorch/trainer/Trainer.py
rnwang04: maybe when user only accelerate training part?

PR #7509, opened 2023-02-13
Nano: add Unet ut for PyTorch openvino & ipex+jit

rnwang04:
## Description

Add UNet related UT in PR action to prevent changes on Nano side from causing stable diffusion to fail.

### 1. Why the change?
https://github.com/intel-analytics/BigDL/pull/7500

### 2. User API changes
None, just add two unittests.

### 3. Summary of the change

- add twos UNet related UT
- extend Tensor conditional_args from `torch.Tensor` to `(torch.Tensor, torch.FloatTensor, torch.LongTensor)` to cope more Tensor types

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/nano/test/openvino/pytorch/test_openvino.py
+++ b/python/nano/test/openvino/pytorch/test_openvino.py
@@ -308,21 +308,26 @@ class TestOpenVINO(TestCase):
         from diffusers.models import UNet2DConditionModel
         unet = UNet2DConditionModel(sample_size=64,
                                     cross_attention_dim=10,
-                                    attention_head_dim=1)
+                                    attention_head_dim=1,
+                                    down_block_types=("CrossAttnDownBlock2D", "DownBlock2D"),
+                                    block_out_channels=(32, 64),
+                                    up_block_types=("UpBlock2D", "CrossAttnUpBlock2D"),
+                                    layers_per_block=1)
         latent_shape = (2, 4, 8, 8)
         image_latents = torch.randn(latent_shape, device = "cpu", dtype=torch.float32)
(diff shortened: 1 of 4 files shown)


python/nano/test/openvino/pytorch/test_openvino.py
```diff
@@ -303,3 +303,33 @@ def test_openvino_trace_kwargs(self):
...
+        unet(image_latents, torch.Tensor([980]).long(), encoder_hidden_states)
+        dynamic_axes= {"sample": [0],
+                        "encoder_hidden_states": [0],
+                        "unet_output": [0]}
+        nano_unet = InferenceOptimizer.trace(unet, accelerator="openvino",
+                                                input_sample=input_sample,
+                                                input_names=["sample", "timestep",
+                                                            "encoder_hidden_states", "return_dict"],
+                                                output_names=["unet_output"],
+                                                dynamic_axes=dynamic_axes,
```
TheaperDeng, line 329: I wonder what will happen if we leave this `dynamic_axes` to be autogenerated

TheaperDeng (commented):
others LGTM

python/nano/test/openvino/pytorch/test_openvino.py
rnwang04: Just validate it, no need for `dynamic_axes`, `input_names` and `output_names` now. Will update the ut. @cyita

rnwang04: If leave this `dynamic_axes` to be autogenerated, then model can't deal with dynamic shape which is showed in uts.

cyita (approved):

cyita:
LGTM

PR #7513, opened 2023-02-13
Orca: Support tqdm progress bar in evaluation and predict

leonardozcm:
## Description

In this PR we support tqdm progress bar in evaluaion like training does.
In predict mode it is hard to apply a progress bar on it for uncountable iterable dataset, so we may just notify users with some steady log output in one line:
```python
    def after_pred_iter(self, runner):
        if self._is_rank_zero(runner):
            print("\r rank0 predict batch_idx: {%d}" % runner.batch_idx, end="")
```

### DIFF
--- a/python/orca/src/bigdl/orca/learn/pytorch/callbacks/tqdm.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/callbacks/tqdm.py
@@ -44,7 +44,7 @@ class TqdmCallback(Callback):
     def before_train_epoch(self, runner):
         if self._is_rank_zero(runner):
             desc = "{}/{}e".format(runner.epochs + 1,
-                                runner.num_epochs)
+                                   runner.num_epochs)
 
             invalidInputError(tqdm is not None,
                               "tqdm is not installed, please install with 'pip install tqdm'")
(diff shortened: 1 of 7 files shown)


python/orca/src/bigdl/orca/learn/pytorch/callbacks/tqdm.py
```diff
@@ -44,7 +44,7 @@ def after_train_iter(self, runner):
     def before_train_epoch(self, runner):
         if self._is_rank_zero(runner):
             desc = "{}/{}e".format(runner.epochs + 1,
-                                   runner.num_epochs)
+                                runner.num_epochs)
```
hkvision, line 47: why this change?

python/orca/src/bigdl/orca/learn/pytorch/callbacks/tqdm.py
```diff
@@ -53,6 +53,29 @@ def before_train_epoch(self, runner):
...
+            invalidInputError(tqdm is not None,
+                              "tqdm is not installed, please install with 'pip install tqdm'")
+            runner._progress_bar = tqdm(total=len(runner.val_loader),
+                                        desc=desc,
+                                        unit="batch",
+                                        leave=False)
+
+    def after_pred_iter(self, runner):
+        if self._is_rank_zero(runner):
+            print("\r rank0 predict batch_idx: {%d}" % runner.batch_idx, end="")
```
hkvision, line 77: remove `rank0`? Just Predict batch_idx: ...?
For users they don't even know rank0

leonardozcm: sure

PR #7515, opened 2023-02-13
[Nano] Make tf bf16 quantization not inplace

Oscilloscope98:
## Description

Make sure that no changes are made to the original model during `bigdl.nano.tf.keras.InferenceOptimizer.quantize(model, precision="bf16")`

### 1. Why the change?

Current implementation of TensorFlow bf16 quantization makes change to the original model, which is not intuitive to use for users.

### 2. User API changes

N/A

### 3. Summary of the change 

- Make sure that no changes are made to the original model during `bigdl.nano.tf.keras.InferenceOptimizer.quantize(model, precision="bf16")`
- Update related uts
- Update relate how-to guides

### 4. How to test?
- [x] Unit test
- [x] Document test: https://yuwentestdocs.readthedocs.io/en/fix-nano-tf-bf16-inplace/doc/Nano/Howto/Inference/TensorFlow/tensorflow_inference_bf16.html

### DIFF
--- a/python/nano/src/bigdl/nano/tf/keras/amp/bfloat16.py
+++ b/python/nano/src/bigdl/nano/tf/keras/amp/bfloat16.py
@@ -19,16 +19,16 @@ import tensorflow as tf
 
 
 def BF16Model(model):
-    model_policys = []
+    original_model_policies = []
     policy_bf16 = mixed_precision.Policy('mixed_bfloat16')
     for layer in model.layers:
-        model_policys.append(layer._dtype_policy)
+        original_model_policies.append(layer._dtype_policy)
         layer._dtype_policy = policy_bf16
     with TemporaryDirectory() as temp_dir:
         model.save(temp_dir)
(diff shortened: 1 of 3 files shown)


python/nano/src/bigdl/nano/tf/keras/amp/bfloat16.py
```diff
@@ -19,12 +19,16 @@
 
 
 def BF16Model(model):
+    model_policys = []
```
TheaperDeng, line 22: could we call it `original_model_policies`, seems more understandable. (no strong opinion on this, feel free to ignore)

TheaperDeng (commented):
others LGTM

python/nano/src/bigdl/nano/tf/keras/amp/bfloat16.py
Oscilloscope98: I agree, has updated :D

PR #7516, opened 2023-02-13
Chronos: fix quantization related ut

plusbang:
## Description

To avoid "Error: -10 10:32:58 [ERROR] Specified timeout or max trials is reached! Not found any quantized model which meet accuracy goal. Exit.", fix quantization related ut.

### 4. How to test?
- [x] Unit test


rnwang04 (approved):
LGTM

PR #7517, opened 2023-02-13
Nano: upgrade to ipex 1.13.100

MeouSker77:
## Description

IPEX model and optimizer doesn't support `load_state_dict()` before, so we use some workaround.
This issue should be fixed in version 1.13.100, so we upgrade to this version and won't use workaround with pytorch 1.13


TheaperDeng (approved):
LGTM

rnwang04 (approved):
LGTM

PR #7520, opened 2023-02-13
Nano: use subprocess to calculate throughput for inc quantizated model

rnwang04:
## Description

In https://github.com/intel-analytics/BigDL/pull/7263, I use subprocess to calculate throughput for original model to achieve thread control.
Now use subprocess to calculate throughput for inc quantizated model, so that all model's throughput could be calculated rightly in keras `InferenceOptimizer.optimize`.

### How to test?
- [x] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/tf/keras/inference/_worker.py
+++ b/python/nano/src/bigdl/nano/tf/keras/inference/_worker.py
@@ -19,6 +19,8 @@ import sys
 import numpy as np
 import pickle
 import tensorflow as tf
+import os
+from bigdl.nano.tf.keras import InferenceOptimizer
 
 
 def throughput_calculate_helper(iterrun, func, model, input_sample):
@@ -45,5 +47,10 @@ if __name__ == "__main__":
         tf.config.threading.set_inter_op_parallelism_threads(0)
         tf.config.threading.set_intra_op_parallelism_threads(thread_num)
     params = pickle.load(open(param_file, "rb"))
+    if params["method"] != "original":
+        model_dir = os.path.dirname(param_file)
+        qmodel = InferenceOptimizer.load(model_dir, params["model"])
+        params["model"] = qmodel
+    del params["method"]
     latency = throughput_calculate_helper(**params)
     print(latency)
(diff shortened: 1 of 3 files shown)


python/nano/src/bigdl/nano/tf/keras/inference/_worker.py
```diff
@@ -45,5 +47,10 @@ def throughput_calculate_helper(iterrun, func, model, input_sample):
         tf.config.threading.set_inter_op_parallelism_threads(0)
         tf.config.threading.set_intra_op_parallelism_threads(thread_num)
     params = pickle.load(open(param_file, "rb"))
```
MeouSker77, line 49: may use `cloudpickle` to load because you saved it by `cloudpickle`

python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
```diff
@@ -247,15 +247,19 @@ def optimize(self, model: Model,
...
                                   "func": func_test,
-                                  "model": acce_model,
-                                  "input_sample": input_sample}
+                                  "model": model,  # save original model
+                                  "input_sample": input_sample,
+                                  "method": method}
                         with tempfile.TemporaryDirectory() as temp_dir:
+                            if method != "original":
+                                # save accelerated model
+                                InferenceOptimizer.save(acce_model, temp_dir)
```
MeouSker77, line 262: besides, I wonder if `cloudpickle` can dump inc model directly.

MeouSker77 (approved):
others LGTM

python/nano/src/bigdl/nano/tf/keras/inference/optimizer.py
rnwang04: I have tried this, `cloudpickle` can't dump inc model.

MeouSker77: all right

python/nano/src/bigdl/nano/tf/keras/inference/_worker.py
rnwang04: Done

TheaperDeng (approved):
LGTM

PR #7525, opened 2023-02-13
Chronos: reduce overhead during inference

plusbang:
## Description

If the input data is smaller than one batch, we don't need to use loop to divide input data and call `np.concatenate`.
This simplification can help reduce overhead during inference.

### 2. User API changes

None

### 4. How to test?
- [x] Unit test


TheaperDeng (approved):
LGTM

PR #7529, opened 2023-02-14
Image coco

songhappy:
## Description
Support read coco2017 images on SparkXShards. 

### 1. Why the change?
coco2017 dataset format is different from regular image datasets

### 2. User API changes
read_coco(file_path)

### DIFF
--- a/python/orca/src/bigdl/orca/data/image/preprocessing.py
+++ b/python/orca/src/bigdl/orca/data/image/preprocessing.py
@@ -13,7 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-from PIL import Image, ImageDraw
 from bigdl.orca import OrcaContext
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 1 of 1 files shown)


python/orca/src/bigdl/orca/data/image/preprocessing.py
```diff
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+from PIL import Image, ImageDraw
```
hkvision, line 16: Do not import PIL globally, otherwise you will make it a must dependency

python/orca/src/bigdl/orca/data/image/preprocessing.py
```diff
@@ -90,7 +90,7 @@ def read_images_pil(file_path: str,
 
     def load_image(iterator):
         for f in iterator:
-            img = open_image(f)
+            img = open_image(f).convert("RGB")
```
hamham223, line 93: Cause error: 
```
logits and labels must have the same first dimension, got logits shape [51200,3] and labels shape [153600]
```

Nightly Job: https://github.com/intel-analytics/BigDL/actions/runs/4427765949/jobs/7775579178

PR #7531, opened 2023-02-14
read voc data

songhappy:
## Description

to support VOC on SparkXSpards

### 1. Why the change?
voc image datasets are different from regular images 

### 2. User API changes
read_voc(file_path)


leonardozcm (approved):
Better setting aside a little time for reviewing next time

PR #7540, opened 2023-02-14
Refactor tf1/tf2/ray related actions

hkvision:
- TF1 with ray is only for automl -> rename the job to automl [This will only work for python3.7 since python 3.8 can't install tf1.15]
- For all other cases, ray should go together with tf2


pinggao187 (approved):

hkvision:
https://github.com/intel-analytics/BigDL/actions/runs/4181108967/jobs/7246655985

PR #7541, opened 2023-02-14
Nano/Chronos: Refactor nano's utils (2/3)

MeouSker77:
## Description

Refactor nano's utils (2/3):
- remove `python/nano/utils/log4Error.py`
- replace all `from bigdl.nano.utils.log4Error import invalidInputError` with `from bigdl.nano.utils.common import invalidInputError` in chronos

TODO:
step 3: move `nano/pytorch/utils.py` to `nano/utils/pytorch` and move `nano/tf/utils.py` to `/nano/utils/tf`

NOTE:
I think we shouldn't use `invlidInputError` in a chronos example <https://github.com/intel-analytics/BigDL/blob/90a13b2db84979085788e30248802e9de0dd8041/python/chronos/example/tcmf/run_electricity.py#L78>


rnwang04 (approved):
LGTM

MeouSker77:
APRT

PR #7463, opened 2023-02-08
Orca: setup unit test clean env

hamham223:
### 1. Why the change?

Add a clean orca test enviornment, only install bigdl-orca necessary libs.

### 2. Summary of the change 

+ Create a new test job only with bigdl-orca necessary dependencies
+ The job will run `test/bigdl/orca/data/test_xshards_partition.py` and `test/bigdl/orca/test_orca_py37_import.py` 

### 3. How to test?

The following unit test must be completed ✅ 
- [x] [Orca-Python-Py37-Basic](https://github.com/intel-analytics/BigDL/actions/runs/4131525920/jobs/7139431678)
- [x] [Orca-Python-Ray-Py37-Spark3](https://github.com/intel-analytics/BigDL/actions/runs/4131525920/jobs/7139432567) (`orca/data`)
- [x] [Orca-Python-Horovod-Tf2-Py37-Spark3](https://github.com/intel-analytics/BigDL/actions/runs/4131525920/jobs/7139432272) (`orca/data/test_read_parquet_images.py`)

### 4. New dependencies

+ New Python dependencies
       - `test_orca_py37_import.py`

PR #7545, opened 2023-02-14
Orca: rm test `test_xshards partition`

hamham223:
### 1. Why the change?

Since the pytest sciprt `orca/data/test_xshards_partition.py` is already tested in the `Orca-Python-Py37-Basic`, PR: https://github.com/intel-analytics/BigDL/pull/7463, it can be removed from the old test.

Fix the issue in pr: https://github.com/intel-analytics/BigDL/pull/7503
### 2. How to test?

The following Orca Unit Test should be completed ✅ 
- [ ] Orca-Python-Ray-Py37-Spark3
- [ ] Orca-Python-Ray-Py38-Spark3


hkvision:
Added in https://github.com/intel-analytics/BigDL/pull/7540/files#diff-ecaa2b3ce216f8cb1c12eaf2d39bb7d6bc3f34c6111cbe9dd478a366a52b8ff7R67

PR #7584, opened 2023-02-16
Chronos: update dockerfile

plusbang:
## Description

Update dockerfile to prepare for uploading chronos docker image to dockerhub.

### 1. Why the change?

When build docker image based on previous dockerfile, the image size is ~6GB. As a base image, it is really huge.

### 3. Summary of the change 

- Modify default value and simplify apt-get installation in dockerfile (The image size will be 1.48GB)
- Update readme and how-to guide

### 4. How to test?
- [x] Local test
- [x] Document test
https://binbin-doc.readthedocs.io/en/dockerfile/doc/Chronos/Howto/docker_guide_single_node.html#

### DIFF
--- a/docker/chronos-nightly/Dockerfile
+++ b/docker/chronos-nightly/Dockerfile
@@ -20,8 +20,8 @@ WORKDIR /opt/work
 ARG DEBIAN_FRONTEND=noninteractive
 ENV TZ=Asia/Shanghai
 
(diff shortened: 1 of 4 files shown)


TheaperDeng (approved):
LGTM, when we upload the docker to dockerhub, we may revise the document again to let users fetch the container directly from dockerhub

PR #7585, opened 2023-02-16
[PPML] Changed app_key to api_key

SEULiyao:
## Description

changed some parameter names from `app_key` to `api_key`

### 1. Why the change?

the parameter called `app_key` before is named `api_key` now


Uxito-Ada (approved):
lgtm

PR #7587, opened 2023-02-16
[PPML] Change torchserve start script to allow booting multiple workers

gc-fu:
## Description
As stated in Gramine documentation:

`We also recommend to use core pinning via taskset or even isolating cores via isolcpus or disabling interrupts on cores via nohz_full.`

If we want to boot multiple instance of gramine, we'd better pin cores that gramine can use by using `taskset`.

### 1. Why the change?

To provide better throughput, and less enclave memory usage

### 2. User API changes

The new api will be:
```bash
./start-torchserve-sgx.sh [-c <configfile>] [-b <core# for each backend worker>] [-f <core# for frontend worker>]
```

### 3. Summary of the change 
Change torchserve start scripts to allow booting multiple gramine worker instances.

### 4. How to test?
- [x] Manually test

Tested with 16G enclave memory.

### DIFF
--- a/ppml/trusted-deep-learning/base/start-backend-sgx.sh
+++ b/ppml/trusted-deep-learning/base/start-backend-sgx.sh
@@ -1,15 +1,19 @@
 #!/bin/bash
 port=9000
-while getopts "p:" opt
+core=""
+while getopts ":c:p:" opt
 do
     case $opt in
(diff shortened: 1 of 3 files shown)


glorysdj (approved):
LGTM

PR #7588, opened 2023-02-16
Nano: Fix affinity detection

MeouSker77:
## Description

"compact,1,0" will cause `os.sched_getaffinity(0)` return 1, fix it.


rnwang04 (approved):
LGTM

PR #7590, opened 2023-02-16
fix bigdl-nano-init --perf

qiuxin2012:
## Description

--perf 's LD_PRELOAD's env is wrong, fix it.

### 1. Why the change?

--perf 's LD_PRELOAD should not contains iomp.so

### 2. User API changes

none

### 3. Summary of the change 

--perf 's LD_PRELOAD should not contains iomp.so

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...


MeouSker77 (approved):
LGTM

PR #7593, opened 2023-02-16
Remove 246/312/313 from script names

hkvision:
To make our scripts easier to maintain
- Remove spark specific versions (246, 313) from script names, use 2 and 3 only, in the future each spark version uses one script.
- spark312 is not used and thus removed.


pinggao187 (approved):

leonardozcm (approved):

hkvision:
Only one failure is due to network. Merge it first. Tell me if there are any issues.

PR #7594, opened 2023-02-17
[PPML] Fix 0.29.4 bug

hzjane:
## Description
Fix 0.29.4 performance bug and add a ENV to disable glibc per thread arena to reduce the usage of EPC.


hzjane:
https://github.com/intel-analytics/BigDL/actions/runs/4200259305/jobs/7286097897

PR #7597, opened 2023-02-17
Nano: refactor nano utils (3/3)

MeouSker77:
## Description

Refactor nano's utils (3/3):
- move `nano/pytorch/utils.py` to `nano/utils/pytorch/*`
- move `nano/tf/utils.py` to `nano/utils/tf/*`


python/nano/src/bigdl/nano/utils/pytorch/inspect.py
```diff
@@ -17,7 +17,6 @@
 
 import inspect
 import torch
-from bigdl.nano.pytorch.model import AcceleratedLightningModule
```
TheaperDeng, line 20: why do we move this to each function internally?

TheaperDeng (approved):
LGTM

PR #7603, opened 2023-02-17
Chronos: move `import` to reduce inference overhead

plusbang:
## Description

During performance tests on icx5, I found move `from bigdl.chronos.pytorch.utils import _pytorch_fashion_inference` and `from bigdl.nano.utils.common import invalidInputError` to front can reduce inference overhead.

TCNForecaster, 4 cores, lookback 96, horizon 96
onnx latency:
| dataset | latency(before moving) | latency(after moving) |
| ---- | ---- | ---- |
| electricity | 0.140ms | 0.135ms |
| ETTh1 | 0.056ms | 0.053ms |

### 4. How to test?
- [x] Unit test
- [x] Document test
(Because https://github.com/intel-analytics/BigDL/pull/4618, document test is necessary)
https://binbin-doc.readthedocs.io/en/move-import/doc/PythonAPI/Chronos/forecasters.html


TheaperDeng (commented):
The PR LGTM, but please hold for some time, we need to check the failed Action. (it seems unrelated to this PR)

TheaperDeng (approved):
LGTM

PR #7605, opened 2023-02-17
[Nano] Add installation option for users to choose stock TensorFlow

Oscilloscope98:
## Description

Add installation option for users to choose stock TensorFlow when installing Nano

### 1. Why the change?

- For TensorFlow users, they may want to use `tensorflow` instead of `intel-tensorflow` for some cases.
- For the further integration of ITEX (release version 1.0.0 and 1.1.0), which currently requires stock TensorFlow, the installation option of stock TensorFlow may also be needed. 

### 2. User API changes

N/A

### 3. Summary of the change 

- Add installation option for users to choose stock TensorFlow when installing Nano
- Removed the manually-added requirements for `keras` and `tensorflow-estimator`
- Add related workflows for action tests

### 4. How to test?
- [x] Unit test

### 5. New dependencies

N/A

**TO DO:** In the next PR, we need to update the installation panel.

### DIFF
--- a/python/nano/setup.py
+++ b/python/nano/setup.py
@@ -104,19 +104,19 @@ def setup_package():
                                "tensorflow-estimator==2.10.0; (platform_machine=='x86_64' or platform_machine == 'AMD64')"]
 
     # options for stock tensorflow
-    stock_tensorflow_27_requires = ["tensorflow==2.7.0; (platform_machine=='x86_64' or platform_machine == 'AMD64')",
+    stock_tensorflow_27_requires = ["tensorflow==2.7.4; (platform_machine=='x86_64' or platform_machine == 'AMD64')",
                                     "keras==2.7.0; (platform_machine=='x86_64' or platform_machine == 'AMD64')",
                                     "tensorflow-estimator==2.7.0; (platform_machine=='x86_64' or platform_machine == 'AMD64')"]
 
-    stock_tensorflow_28_requires = ["tensorflow==2.8.0; (platform_machine=='x86_64' or platform_machine == 'AMD64')",
+    stock_tensorflow_28_requires = ["tensorflow==2.8.4; (platform_machine=='x86_64' or platform_machine == 'AMD64')",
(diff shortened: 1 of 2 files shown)


python/nano/setup.py
```diff
@@ -74,43 +74,64 @@ def download_libs(url: str):
...
+                                    "tensorflow-estimator==2.8.0; (platform_machine=='x86_64' or platform_machine == 'AMD64')"]
+
+    stock_tensorflow_29_requires = ["tensorflow==2.9.0; (platform_machine=='x86_64' or platform_machine == 'AMD64')",
+                                    "keras==2.9.0; (platform_machine=='x86_64' or platform_machine == 'AMD64')",
+                                    "tensorflow-estimator==2.9.0; (platform_machine=='x86_64' or platform_machine == 'AMD64')"]                         
+
+    stock_tensorflow_210_requires = ["tensorflow==2.10.0; (platform_machine=='x86_64' or platform_machine == 'AMD64')",
+                                    "keras==2.10.0; (platform_machine=='x86_64' or platform_machine == 'AMD64')",
+                                    "tensorflow-estimator==2.10.0; (platform_machine=='x86_64' or platform_machine == 'AMD64')"]  
+
```
TheaperDeng, line 122: for stock tensorflow, should we install the latest patch of each minor version? like 2.7.4, 2.8.4, 2.9.3, 2.10.1.

TheaperDeng (commented):
others LGTM

python/nano/setup.py
Oscilloscope98: Have updated each minor version to the latest one.

BTW, I noticed that the installation of `tensorflow` and `intel-tensorflow` will install the corresponding version of `keras` and `tensorflow-estimator` automatically. Is there any reason that we manually specify them? Could we remove them?

Oscilloscope98: Have removed the manually-added requirements for `keras` and `tensorflow-estimator` as discussion.

TheaperDeng (approved):
LGTM

PR #7616, opened 2023-02-20
Nano: add compressed saving logic for accelerator=None/ipex

TheaperDeng:
### 1. Why the change?
Currently, `InferenceOptimizer.save` save fp32 parameter for all models with `accelerator=None/"ipex"/"jit"`, this PR add a new saving and loading logic for compressed saving file.

- [x] original model
- [x] bf16 model
- [x] ipex model
- [ ] jit model

### 2. User API changes
A new parameter `compress_to_bf16` is added to `InferenceOptimizer.save`.
```python
InferenceOptimizer.save(opt_model, path)  # uncompressed file saved
InferenceOptimizer.save(opt_model, path, compress_to_bf16=True)  # compressed file saved

InferenceOptimizer.load(path, ...)  # both files can be loaded the same way
```

### 3. Summary of the change 
The logic basic works like this
image

### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/deps/ipex/ipex_inference_bf16_model.py
+++ b/python/nano/src/bigdl/nano/deps/ipex/ipex_inference_bf16_model.py
@@ -117,4 +117,5 @@ class PytorchIPEXJITBF16Model(PytorchIPEXJITModel):
                                        jit_strict=status.get('jit_strict', True),
                                        jit_method=status.get('jit_method', None),
                                        weights_prepack=status.get('weights_prepack', None),
-                                       enable_onednn=status.get('enable_onednn', True))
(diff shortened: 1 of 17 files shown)


jason-dai:
Change to `InferenceOptimizer.save(opt_model, path, compression='bf16') `? We may add other compression support in future

TheaperDeng:
> Change to `InferenceOptimizer.save(opt_model, path, compression='bf16') `? We may add other compression support in future

Yeah, you are right, the API has been revised. Still, dtypes other than bf16 could have a really difficult implementation since fp16 or fp8 need scaling so they are not planned for now.

TheaperDeng:
@rnwang04 merge this PR for quick iter

Issue #7574, opened 2023-02-16
Pytorch Pyspark Estimator cannot load checkpoint from hdfs


jenniew:
Pytorch Pyspark Estimator cannot load checkpoint from hdfs.

The error is like below:
Traceback (most recent call last):
  File "/opt/work/jwang/BigDL-jennie-2/python/orca/example/learn/pytorch/bert/sentiment_huggingface.py", line 1263, in <module>
    main()
  File "/opt/work/jwang/BigDL-jennie-2/python/orca/example/learn/pytorch/bert/sentiment_huggingface.py", line 1085, in main
    est.load_checkpoint("hdfs://almaren-node-105:8020/user/root/jwang/sst2/a.ckpt")
  File "/opt/work/jwang/BigDL-jennie-2/dist/lib/bigdl-friesian-spark_3.1.3-2.3.0-SNAPSHOT-python-api.zip/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py", line 702, in load_checkpoint
  File "/opt/work/jwang/BigDL-jennie-2/dist/lib/bigdl-friesian-spark_3.1.3-2.3.0-SNAPSHOT-python-api.zip/bigdl/orca/learn/pytorch/core/modelIO.py", line 62, in load_checkpoint
  File "/opt/work/jwang/wj-transformer/lib/python3.8/site-packages/torch/serialization.py", line 771, in load
    with _open_file_like(f, 'rb') as opened_file:
  File "/opt/work/jwang/wj-transformer/lib/python3.8/site-packages/torch/serialization.py", line 275, in _open_file_like
    return _open_buffer_reader(name_or_buffer)
  File "/opt/work/jwang/wj-transformer/lib/python3.8/site-packages/torch/serialization.py", line 260, in __init__
    _check_seekable(buffer)
  File "/opt/work/jwang/wj-transformer/lib/python3.8/site-packages/torch/serialization.py", line 348, in _check_seekable
    f.seek(f.tell())
  File "pyarrow/io.pxi", line 274, in pyarrow.lib.NativeFile.tell
  File "pyarrow/io.pxi", line 206, in pyarrow.lib.NativeFile.get_random_access_file
  File "pyarrow/io.pxi", line 235, in pyarrow.lib.NativeFile._assert_seekable
OSError: only valid on seekable file

Example code:
est = Estimator.from_torch(...)
est.save_checkpoint("hdfs://xxx/xx.ckpt")
[...]

PR #7619, opened 2023-02-20
[ORCA] Refactor Load Checkpoint with PyTorch Pyspark

sgwhat:
## Description
To fix https://github.com/intel-analytics/BigDL/issues/7574, remove fsspec and pyarrow using.

### 1. Why the change?
Fix bug caused by pyarrow when loading checkpoint from remote storage like HDFS.

### 2. User API changes
N/A

### 3. Summary of the change 
Re-write this part according to the `load` implementation.

### 4. How to test?
- [x] Unit test
- [x] Yarn test

### DIFF
--- a/python/orca/src/bigdl/orca/learn/pytorch/core/modelIO.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/core/modelIO.py
@@ -22,7 +22,6 @@ from abc import ABCMeta, abstractmethod
 
 import torch
 
-from bigdl.orca.data.file import get_remote_file_to_local
 from bigdl.dllib.utils.log4Error import invalidInputError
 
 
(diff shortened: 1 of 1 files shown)


python/orca/src/bigdl/orca/learn/pytorch/core/modelIO.py
```diff
@@ -14,10 +14,15 @@
 # limitations under the License.
 #
 
-from abc import ABCMeta, abstractmethod
+import os
```
lalalapotter, line 17: Shall we avoid these global dependencies?

hkvision: import os globally should be ok😂
but if some package is only used once, then it is better practice to make it local import

lalalapotter (approved):

sgwhat:
one action failed but not related.

PR #7620, opened 2023-02-20
[PPML]fix: more malloc arena is needed to run pyspark

ShanSimu:
## Description

more malloc arena is needed to run pyspark


qiyuangong (approved):
LGTM

PR #7623, opened 2023-02-20
Orca: Save the tensorflow model in NCF tutorial in h5 format

zpeng1898:
## Description

Save the tensorflow model in NCF tutorial in h5 format.

### 1. Why the change?

Sometimes saving tf model in default format `Tensorflow SavedModel` on hdfs requires the permission of `mkdir`, and saving tf model as a `HDF5` single file does not need to get the permission.

### DIFF
--- a/python/orca/tutorial/NCF/tf_train_spark_dataframe.py
+++ b/python/orca/tutorial/NCF/tf_train_spark_dataframe.py
@@ -106,8 +106,7 @@ for k, v in eval_stats.items():
 
 
 # Step 6: Save the trained TensorFlow model and processed data for resuming training or prediction
-# TODO: fix save model to HDFS
-est.save(os.path.join(args.model_dir, "NCF_model"), save_format="h5")
+est.save(os.path.join(args.model_dir, "NCF_model.h5"))
 save_model_config(config, args.model_dir, "config.json")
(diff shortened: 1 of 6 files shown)


python/orca/tutorial/NCF/tf_train_spark_dataframe.py
```diff
@@ -107,7 +107,7 @@ def model_creator(config):
 
 # Step 6: Save the trained TensorFlow model and processed data for resuming training or prediction
 # TODO: fix save model to HDFS
-est.save(os.path.join(args.model_dir, "NCF_model"))
+est.save(os.path.join(args.model_dir, "NCF_model"), save_format="h5")
```
hkvision, line 110: directly use NCF_model.h5?

python/orca/tutorial/NCF/tf_train_spark_dataframe.py
```diff
@@ -107,7 +107,7 @@ def model_creator(config):
 
 # Step 6: Save the trained TensorFlow model and processed data for resuming training or prediction
 # TODO: fix save model to HDFS
```
hkvision, line 109: remove this

hkvision:
Also need to change resume train and predict accordingly

PR #7629, opened 2023-02-21
Nano: fix bigdl-nano-init and turbojpeg

MeouSker77:
## Description

- Fix loading `libturbojpeg.so.0.2.0` error on windows
- Fix OS detection in `bigdl-nano-init`

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/vision/datasets/datasets.py
+++ b/python/nano/src/bigdl/nano/pytorch/vision/datasets/datasets.py
@@ -53,7 +53,13 @@ class ImageFolder(torchvision.datasets.ImageFolder):
             target and transforms it.
         """
         super(ImageFolder, self).__init__(root, transform, target_transform)
-        self.jpeg: Optional[TurboJPEG] = None
+        try:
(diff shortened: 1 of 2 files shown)


python/nano/src/bigdl/nano/pytorch/vision/datasets/datasets.py
```diff
@@ -62,16 +68,14 @@ def _read_image_to_bytes(self, path: str):
         return img_str
 
     def _decode_img_libjpeg_turbo(self, img_str: str):
-        if self.jpeg is None:
-            self.jpeg = TurboJPEG(lib_path=local_libturbo_path)
         bgr_array = self.jpeg.decode(img_str)
```
rnwang04, line 71: if self.jpeg is None?

MeouSker77: before this PR, `self.jpeg` is initialized to `None` on line 56, so line 66 will execute once and only once

rnwang04 (approved):
LGTM

PR #7633, opened 2023-02-21
Nano: support compressed saving logic for JIT model

rnwang04:
## Description

Support compressed saving logic for `accelerator=jit` by saving state dict of bf16 model, and when loading, load the state_dict to fp32 and then convert again.

### 1. Why the change?

https://github.com/intel-analytics/BigDL/pull/7616
https://github.com/analytics-zoo/nano/issues/294

### 2. User API changes
makes `compression='bf16'` works for JIT model.
```python
# for pure jit
opt_model = InferenceOptimizer.trace(model,
                                     accelerator="jit",
                                     input_sample=input_sample,
                                     use_ipex=False)
InferenceOptimizer.save(opt_model, dir, compression="bf16")
opt_model_load = InferenceOptimizer.load(dir, model=model,
                                         input_sample=input_sample)
# for jit ipex
opt_model = InferenceOptimizer.trace(model,
                                     accelerator="jit",
                                     input_sample=input_sample,
                                     use_ipex=True)
InferenceOptimizer.save(opt_model, dir, compression="bf16")
opt_model_load = InferenceOptimizer.load(dir, model=model,
                                         input_sample=input_sample)
```

### 3. Summary of the change 

- support compressed saving logic for `accelerator=jit` by saving state dict of bf16 model, and when loading, load the state_dict to fp32 and then convert again
image

- update and complete related uts

### 4. How to test?
- [x] Unit test

TODO:
locate why jit model will generate different result for the same input sample.

### DIFF
--- a/python/nano/src/bigdl/nano/deps/ipex/ipex_inference_bf16_model.py
+++ b/python/nano/src/bigdl/nano/deps/ipex/ipex_inference_bf16_model.py
@@ -19,7 +19,9 @@ from .ipex_inference_model import PytorchIPEXJITModel
 from bigdl.nano.pytorch.context_manager import generate_context_manager
 from bigdl.nano.utils.common import _avx512_checker
 from bigdl.nano.utils.common import invalidInputError
+from bigdl.nano.utils.pytorch import transform_state_dict_to_dtype
 import torch
+import copy
 
 
 class PytorchIPEXJITBF16Model(PytorchIPEXJITModel):
@@ -94,17 +96,25 @@ class PytorchIPEXJITBF16Model(PytorchIPEXJITModel):
         return status
 
     @staticmethod
-    def _load(path, model, inplace=False):
+    def _load(path, model, inplace=False, input_sample=None):
         status = PytorchIPEXJITBF16Model._load_status(path)
         checkpoint_path = path / status['checkpoint']
         if status["use_jit"]:
-            if status["use_ipex"]:
-                import intel_extension_for_pytorch as ipex
-            model = torch.jit.load(checkpoint_path)
-            model.eval()
-            if status["use_ipex"]:
-                model = torch.jit.freeze(model)
-            from_load = True
+            if status['compression'] == "bf16":
+                state_dict = torch.load(checkpoint_path, map_location='cpu')
+                if status['compression'] == "bf16":
+                    state_dict = transform_state_dict_to_dtype(state_dict, dtype="fp32")
+                model = copy.deepcopy(model)
+                model.load_state_dict(state_dict)
+                from_load = False
+            else:
+                if status["use_ipex"]:
+                    import intel_extension_for_pytorch as ipex
+                model = torch.jit.load(checkpoint_path)
+                model.eval()
+                if status["use_ipex"]:
+                    model = torch.jit.freeze(model)
+                from_load = True
         else:
             state_dict = torch.load(checkpoint_path)
             model.eval()
@@ -117,7 +127,9 @@ class PytorchIPEXJITBF16Model(PytorchIPEXJITModel):
             thread_num = None
         if thread_num is not None:
             thread_num = int(status['thread_num'])
-        return PytorchIPEXJITBF16Model(model, use_ipex=status['use_ipex'],
+        return PytorchIPEXJITBF16Model(model,
+                                       input_sample=input_sample,
+                                       use_ipex=status['use_ipex'],
                                        use_jit=status['use_jit'],
                                        channels_last=status['channels_last'],
                                        channels_last_available=status['channels_last_available'],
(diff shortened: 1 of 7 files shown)


python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
```diff
@@ -203,16 +206,25 @@ def status(self):
...
-                import intel_extension_for_pytorch as ipex
-            model = torch.jit.load(checkpoint_path)
-            model.eval()
-            model = torch.jit.freeze(model)
-            from_load = True
+            if status['compression'] == "bf16":
+                state_dict = torch.load(checkpoint_path, map_location='cpu')
+                if status['compression'] == "bf16":
+                    state_dict = transform_state_dict_to_dtype(state_dict, dtype="fp32")
+                model = copy.deepcopy(model)
```
MeouSker77, line 217: Maybe we should decide whether to  `deepcopy` model based on `inplace` parameter ?

python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
```diff
@@ -203,16 +206,25 @@ def status(self):
...
-            if status["use_ipex"]:
-                import intel_extension_for_pytorch as ipex
-            model = torch.jit.load(checkpoint_path)
-            model.eval()
-            model = torch.jit.freeze(model)
-            from_load = True
+            if status['compression'] == "bf16":
+                state_dict = torch.load(checkpoint_path, map_location='cpu')
+                if status['compression'] == "bf16":
+                    state_dict = transform_state_dict_to_dtype(state_dict, dtype="fp32")
```
MeouSker77, line 216: Can we convert loaded state dict to 'fp32' directly ? Maybe it was not 'fp32' before saving.

MeouSker77 (approved):
others LGTM

python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
rnwang04: Have revised it, thanks

python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
rnwang04: Now it seems we only support fp32 for JIT? maybe we can update this code when we actually meet new use case?
Or obtain the dtype of model by something like `list(model.parameters())[0].dtype`

python/nano/src/bigdl/nano/deps/ipex/ipex_inference_bf16_model.py
```diff
@@ -94,17 +96,25 @@ def status(self):
         return status
 
     @staticmethod
-    def _load(path, model, inplace=False):
+    def _load(path, model, inplace=False, input_sample=None):
```
TheaperDeng, line 99: A quick thought, we have fairly vague definition what should be done in `_load` and what should be done in `__init__` when users call `InferenceOptimizer.load`, we may have a better (at least internal) understanding for this.

python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
```diff
@@ -150,7 +153,7 @@ def __init__(self, model: torch.nn.Module, input_sample=None, use_ipex=False, dt
                                                          strict=jit_strict)
                         except Exception:
                             self.model = torch.jit.script(self.model)
-                    self.model = torch.jit.freeze(self.model)
+                    # self.model = torch.jit.freeze(self.model)
```
TheaperDeng, line 156: this line is commented, I understand it's called in `_load` if user load the model, but when it's called when user call `trace`

python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
TheaperDeng: yeah, this might be a problem, we should have a better logic in `transform_state_dict_to_dtype`. Currently we use `is_floating_point` to decide which parameter needs to be revised.

python/nano/src/bigdl/nano/deps/ipex/ipex_inference_bf16_model.py
```diff
@@ -94,17 +96,25 @@ def status(self):
...
+            if status['compression'] == "bf16":
+                state_dict = torch.load(checkpoint_path, map_location='cpu')
+                if status['compression'] == "bf16":
+                    state_dict = transform_state_dict_to_dtype(state_dict, dtype="fp32")
+                model = copy.deepcopy(model)
+                model.load_state_dict(state_dict)
+                from_load = False
+            else:
+                if status["use_ipex"]:
+                    import intel_extension_for_pytorch as ipex
```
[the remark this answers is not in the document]
rnwang04: yeah it should happen before load.

python/nano/src/bigdl/nano/deps/ipex/ipex_inference_bf16_model.py
rnwang04: we may update here later like give more detailed comment.

python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
rnwang04: will remove it, it's a mistake here.

PR #7638, opened 2023-02-21
Chronos: support numpy in and out onnx/openvino models without converting to tensors

plusbang:
## Description

When call `predict_with_onnx` or `predict_with_openvino`, type transforms between tensor and numpy cause a little bit overhead.
In this PR, support numpy in and out onnx/openvino models without converting to tensors to reduce inference overhead.

### 1. Why the change?

Reduce inference overhead by reducing type transforms between tensor and numpy.
Experiment results are as following:

TCNForecaster, 4 cores, lookback 96, horizon 96
| dataset | onnx latency(before) | onnx latency(after) | openvino latency(before) | openvino latency(after) |
| ---- | ---- | ---- | ---- | ---- |
| ETTh1 | 0.054ms | 0.045ms | 0.130ms | 0.116ms |

### 2. User API changes

Chronos - None
Nano - a new param `output_tensors` in `InferenceOptimizer.trace`
```bash
def trace(..., output_tensors: bool = True, **kwargs)
# output_tensors: boolean, default to True and output of the model will be Tensors, only valid when accelerator='onnxruntime' or accelerator='openvino', otherwise will be ignored. If output_tensors=False, output of the export model will be ndarray.
```

### 3. Summary of the change 

- add a new param `output_tensors` in `InferenceOptimizer.trace`, `PytorchONNXRuntimeModel`, `PytorchOpenVINOModel`
- during inference of forecasters, provide `_numpy_inference`(for numpy input case) and `_tensor_inference`(for tensor input case)
- uts (forecaster related, onnx trace related, openvino trace related)

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/chronos/src/bigdl/chronos/pytorch/utils.py
+++ b/python/chronos/src/bigdl/chronos/pytorch/utils.py
@@ -81,7 +81,8 @@ def _pytorch_fashion_inference(model, input_data, batch_size=None, output_tensor
     else:
         # Reduce time of type transforms between tensor and numpy
         if isinstance(input_data, DataLoader):
-            input_sample_list = list(map(lambda x: x[0].numpy(), input_data))
+            input_sample = tuple(batch[0].numpy() for batch in input_data)
+            input_sample_list = [np.concatenate(input_sample, axis=0)]
         elif isinstance(input_data, np.ndarray):
             input_sample_list = [input_data]
         else:
--- a/python/chronos/src/bigdl/chronos/pytorch/utils.py
+++ b/python/chronos/src/bigdl/chronos/pytorch/utils.py
@@ -19,7 +19,7 @@ import numpy as np
 from torch.utils.data.dataloader import DataLoader
 
 
-def _inference(model, input_sample_list, batch_size):
+def _tensor_inference(model, input_sample_list, batch_size):
     if batch_size is None:
         # this branch is only to speed up the inferencing when batch_size is set to None.
         return model(*input_sample_list).numpy()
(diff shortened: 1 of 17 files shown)


rnwang04 (approved):
Generally LGTM.
We can add `output_tensors` in quantize in next PR.

python/chronos/src/bigdl/chronos/pytorch/utils.py
```diff
@@ -50,12 +69,24 @@ def _pytorch_fashion_inference(model, input_data, batch_size=None):
...
     '''
-    with torch.inference_mode():
-        if isinstance(input_data, list):
-            input_sample_list = list(map(lambda x: torch.from_numpy(x), input_data))
-        elif isinstance(input_data, DataLoader):
-            input_sample_list = [torch.cat(tuple(batch[0] for batch in input_data), dim=0)]
-        else:
-            input_sample_list = [torch.from_numpy(input_data)]
-        yhat = _inference(model, input_sample_list, batch_size=batch_size)
+    if hasattr(model, "output_tensors") and not model.output_tensors:
```
TheaperDeng, line 72: should we add a `output_tensor` parameter and default to True for this `_pytorch_fashion_inference`? `_pytorch_fashion_inference` is an internal function so we could always call the correct `output_tensor` (maybe by maintain a `self.optimized_model_output_tensors`.

only check 1 bool could be much faster than 2 attr access (i.e., `hasattr` and `model.output_tensors`)

python/nano/test/onnx/pytorch/test_onnx.py
```diff
@@ -333,6 +333,29 @@ def forward(self, x):
...
+        loss = nn.CrossEntropyLoss()
+        optimizer = torch.optim.Adam(model.parameters(), lr=0.01)
+        trainer = Trainer(max_epochs=1)
+
+        pl_model = Trainer.compile(model, loss, optimizer)
+        x = torch.rand((10, 3, 256, 256))
+        y = torch.ones((10, ), dtype=torch.long)
+        ds = TensorDataset(x, y)
+        train_loader = DataLoader(ds, batch_size=2)
+        trainer.fit(pl_model, train_loader)
```
TheaperDeng, line 347: maybe we could just use a pretrained resnet 18 to reduce the ut time?

python/nano/test/openvino/pytorch/test_openvino.py
```diff
@@ -364,3 +364,26 @@ def test_openvino_trace_stable_diffusion_unet(self):
...
+        model = mobilenet_v3_small(num_classes=10)
+
+        x = torch.rand((10, 3, 256, 256))
+        y = torch.ones((10, ), dtype=torch.long)
+
+        pl_model = Trainer.compile(model, loss=torch.nn.CrossEntropyLoss(),
+                                   optimizer=torch.optim.SGD(model.parameters(), lr=0.01))
+        ds = TensorDataset(x, y)
+        dataloader = DataLoader(ds, batch_size=2)
+        trainer.fit(pl_model, dataloader)
```
TheaperDeng, line 379: same here.

TheaperDeng (commented):
@rnwang04 @MeouSker77 please also have a look at the nano part.

MeouSker77 (approved):
LGTM, just a little question, shall we also change the signature of `Trainer.trace` ? @TheaperDeng

PR #7646, opened 2023-02-22
Add data sampler for val_dataloader when train with MMCVRayRunner

PatrickkZ:
## Description

user need to provide a `mmcv_runner_creator` when they train with MMCVRayEstimator, they may register a DistEvalHook in the runner creator.
```python
def runner_creator(config):
    model = ...
    optimizer = ...
    if validate:
        val_dataloader = build_dataloader()
        runner.register_hook(
            DistEvalHook(val_dataloader, **eval_cfg), priority='LOW')
```

the runner will be created in `MMCVRayRunner.setup_component()`, at this time, ddp env was not initialized, so the val_dataloder in DistEvalHook is not correctly build with a sampler. so add sampler for val_dataloder if there is a DistEvalHook.

### DIFF
--- a/python/orca/src/bigdl/orca/learn/pytorch/core/lifecycle.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/core/lifecycle.py
@@ -76,7 +76,7 @@ class LifeCycle(metaclass=ABCMeta):
             world_size=world_size)
         self.backend = "torch-distributed"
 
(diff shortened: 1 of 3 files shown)


leonardozcm (commented):
add an ut when worker_per_node=2?

leonardozcm (approved):

PR #7621, opened 2023-02-20
Nano: fix `__getattr__` of PytorchIPEXJITModel and update attributes related uts

rnwang04:
## Description

Fix `RecursionError: maximum recursion depth exceeded while calling a Python object` when calling a non-existing method on a loaded `PytorchIPEXJITModel`/`PytorchIPEXJITBF16Model` and fix/complete/update attributes related uts to make sure accelerated model behave normally.

### 1. Why the change?

Fix `RecursionError: maximum recursion depth exceeded while calling a Python object` when calling a non-existing method on a loaded `PytorchIPEXJITModel`/`PytorchIPEXJITBF16Model`.

### 2. User API changes

No change, just update inside logic about how to obtain additional attributes of `PytorchIPEXJITModel`/`PytorchIPEXJITBF16Model`.

### 3. Summary of the change 

- Fix `RecursionError: maximum recursion depth exceeded while calling a Python object` when calling a non-existing method on a loaded `PytorchIPEXJITModel`/`PytorchIPEXJITBF16Model` 
- Fix/complete/update attributes related uts of IPEXJIT/OpenVINO/ONNXRuntime/INC/BF16

### 4. How to test?
- [x] Unit test

Todo:
`TestIPEXBF16::test_ipex_jit_channels_last_3d_inference` will fail with quantization, locate this issue later and raise another PR for this issue

PR #7647, opened 2023-02-22
Nano: fix ut of channels_last_3d_inference

rnwang04:
## Description

fix ut of `TestIPEXBF16::test_ipex_jit_channels_last_3d_inference` which fail as lack of context manager.

### 1. Why the change?

https://github.com/intel-analytics/BigDL/pull/7621

### How to test?
- [x] Unit test

### DIFF
--- a/python/nano/test/pytorch/tests/inference/test_bf16_ipex.py
+++ b/python/nano/test/pytorch/tests/inference/test_bf16_ipex.py
@@ -354,18 +354,20 @@ class Pytorch1_11:
         model = DummyModelWith3d()
         x1 = torch.rand(32, 3, 3, 224, 224) # 5-dim input test
         x2 = 3
-        # TODO: quantize will fail here, will fix it in next PR
-        ipex_jit_channels_last_model = InferenceOptimizer.trace(model, accelerator="jit", 
-                                                                use_ipex=True, precision='bf16',
-                                                                input_sample=(x1, x2),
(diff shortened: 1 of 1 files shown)


python/nano/test/pytorch/tests/inference/test_bf16_ipex.py
```diff
@@ -354,18 +354,20 @@ def test_ipex_jit_channels_last_3d_inference(self):
...
-        # TODO: quantize will fail here, will fix it in next PR
-        ipex_jit_channels_last_model = InferenceOptimizer.trace(model, accelerator="jit", 
-                                                                use_ipex=True, precision='bf16',
-                                                                input_sample=(x1, x2),
-                                                                enable_onednn=True,
-                                                                channels_last=True)
+        ipex_jit_channels_last_model = InferenceOptimizer.quantize(model,
+                                                                   accelerator="jit", 
+                                                                   use_ipex=True,
+                                                                   precision='bf16',
```
TheaperDeng, line 360: just a thought, should we add some error message to users who use `precision` in `trace` in future PR?

TheaperDeng (approved):

python/nano/test/pytorch/tests/inference/test_bf16_ipex.py
rnwang04: Good suggestion! I will add such behavior in next PR.

PR #7654, opened 2023-02-22
Nano: fix saving model compiled without metrics in keras

rnwang04:
## Description

fix saving model compiled without metrics in keras

```bash
kwargs["metrics"] = user_metric._name
AttributeError: 'NoneType' object has no attribute '_name'
```

### How to test?
- [x] Unit test


MeouSker77 (approved):
LGTM

PR #7663, opened 2023-02-23
Chronos: fix quantization of openvino

plusbang:
## Description

There exists error when run:
```python
forecaster.quantize(calib_data=train_loader, val_data=val_loader, metric="mae", framework='openvino')
```
In this PR, fix it.

### 2. User API changes

None

### 3. Summary of the change 

- modify check of `metric` in `PytorchOpenVINOMetric`
- remove `_str2metric` and use `_str2optimizer_metric` of which parameter order is (pred, target)
- add openvino quantization related uts

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/deps/openvino/pytorch/metric.py
+++ b/python/nano/src/bigdl/nano/deps/openvino/pytorch/metric.py
@@ -21,8 +21,11 @@ from torchmetrics import Metric
 
 class PytorchOpenVINOMetric(BaseOpenVINOMetric):
     def __init__(self, metric, higher_better=True):
-        forecast_metric = hasattr(metric, '__name__') and\
-                          metric.__name__ in ['mae', 'mse', 'rmse', 'mape', 'smape', 'r2']
+        # to support openvino quantization of chronos forecasters
+        if hasattr(metric, '__name__'):
+            forecast_metric = metric.__name__ in ['mae', 'mse', 'rmse', 'mape', 'smape', 'r2']
+        else:
+            forecast_metric = False
         invalidInputError(isinstance(metric, Metric) or forecast_metric,
                           "Please provide an instance of torchmetrics.Metric.")
         super().__init__(metric, higher_better)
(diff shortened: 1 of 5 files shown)


python/nano/src/bigdl/nano/deps/openvino/pytorch/metric.py
```diff
@@ -21,7 +21,12 @@
 
 class PytorchOpenVINOMetric(BaseOpenVINOMetric):
     def __init__(self, metric, higher_better=True):
-        invalidInputError(isinstance(metric, Metric),
+        # to support openvino quantization of chronos forecasters
+        if hasattr(metric, '__name__'):
+            forecast_metric = metric.__name__ in ['mae', 'mse', 'rmse', 'mape', 'smape', 'r2']
+        else:
+            forecast_metric = False
+        invalidInputError(isinstance(metric, Metric) or forecast_metric,
```
TheaperDeng, line 29: This is a valid tmp change for chronos but may not be a good change for nano.

Issues here is that openvino quantization does not really need a torchmetric for metric, but also support a function with 2 tensor input like shown below.
https://github.com/intel-analytics/BigDL/blob/a896e8520ceb175e4d7a9aa1397757a749c95788/python/nano/src/bigdl/nano/deps/openvino/core/metric.py#L43

**Should we just relax the check in line 26 and only check the signature (2 args).**

BTW, please also change the parameter order in

https://github.com/intel-analytics/BigDL/blob/eba2c223a8142e7a6c6a57c8161669dc327da425/python/chronos/src/bigdl/chronos/forecaster/base_forecaster.py#L2037

This may work for some symmetric metric but errors may happen for some asymmetric metrics.

@MeouSker77 please provide any thoughts why a strict torchmetric check is placed here, I should have forgot something.

MeouSker77: > @MeouSker77 please provide any thoughts why a strict torchmetric check is placed here, I should have forgot something.

I don't know why, either. And I also think this check is unnecessary.

plusbang: > 

okk, I will try to relax the strict torchmetric check here

TheaperDeng (approved):
LGTM

PR #7665, opened 2023-02-23
Nano/Chronos: update the requirements txt for chronos and nano

TheaperDeng:
## Description

Add requirements.txt for bigdl-chronos[all] and each installation option for bigdl-nano for security scaning.


Le-Zheng:
LGTM

PR #7672, opened 2023-02-23
Orca: Refactor Testing Structures (Ray-ctx)

hamham223:
### 1. Why the change?

Refactor orca testing structures. 

Several steps to go:
+ Refactor Pytest `Ray-ctx`
+ Refactor Pytest `Horovod`
    + Merge pytorch / tf tests?
    + Shorten test time
+ Refactor Pytest `Ray-Py37/Py38`
+ Remove unnecessary envs, maybe only need to keep 3
    + py37
    + py38
    + ctx (runner configuration is different)

Future Prospection:
+ Eliminate random fail
+ Minimize package requirements

This PR only does the first step.

### 2. Summary of the change 

Two parts

+ test scripts
+ location of testing python sciprts and configurations

### 4. How to test?

- [ ] Orca unit test


hamham223:
Perhaps have a workflow try first @hkvision  :hugs:

python/orca/test/bigdl/orca/learn/ray/ctx/conftest.py
```diff
@@ -0,0 +1,41 @@
+#
```
hkvision, line 1: I don't think we need this file. There are already setup methods in each test file.

hamham223: Still need `conf = {"spark.python.worker.reuse": "false"}`.

hamham223: Removed, worth trying 😃

PR #7678, opened 2023-02-24
[Nano] New class to save model signature

hoshibara:
### 1. Why the change?

The current data processing logic is scattered in multiple files, and many logics are not clear enough. It can be unified into a set of APIs, which is convenient for calling and reusing, and can reduce the coupling degree of the code at the same time.

### 3. Summary of the change 

* Add class ModelInfo to save model signature info

### 4. How to test?
- [ ] Unit test


python/nano/src/bigdl/nano/utils/pytorch/model_info.py
```diff
@@ -0,0 +1,141 @@
...
+
+import torch
+from bigdl.nano.pytorch.model import AcceleratedLightningModule
+
+
+class ModelInfo:
+    def __init__(self, model):
+        self.forward_args = ModelInfo.get_forward_args(model)
+        self.forward_defaults = ModelInfo.get_forward_defaults(model)
+        self.forward_annotations = ModelInfo.get_forward_annotations(model)
```
TheaperDeng, line 27: please add a `self.forward_fullargspec` here for possible future development.

TheaperDeng (commented):
please also add some ut for this class to check if model is not changed after the initialization.

@rnwang04 @MeouSker77 please also have a look.

This class is designed and added for future development. Since we use forward signature almost every where to reformat user's input, and this class may help us make the code easier to understand.

TheaperDeng (approved):
LGTM

python/nano/src/bigdl/nano/utils/pytorch/model_info.py
```diff
@@ -0,0 +1,135 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import inspect
+from typing import Optional, Union
+
+import torch
+from bigdl.nano.pytorch.model import AcceleratedLightningModule
```
MeouSker77, line 21: Don't import `AcceleratedLightningModule` here.
Because `bigdl.nano.pytorch.xxxx` will import `bigdl.nano.utils.pytorch.xxxx`, if you import `bigdl.nano.pytorch.xxxx` here and add this class to `nano/utils/python/__init__.py`, it may cause circular references, then import recursively.
You can import `AcceleratedLightningModule` inside thd function body.

python/nano/src/bigdl/nano/utils/pytorch/model_info.py
```diff
@@ -0,0 +1,135 @@
...
+        It will return a dict of {args: annotations}
+        E.g.
+        def forward(self, a, b=1, c: int = 3, *args, **kwargs):
+            pass
+        it will return {'c': <class 'int'>}
+        """
+        self.forward_annotations = self.forward_fullargspec.annotations
+
+    @staticmethod
+    def get_forward_fullargspec(model):
```
MeouSker77, line 59: Besides, there are already `get_forward_args` and  `get_condition_args` in <https://github.com/intel-analytics/BigDL/blob/5b99f80837186865780b18818545afb0c2ff49d1/python/nano/src/bigdl/nano/utils/pytorch/inspect.py#L22>, why not use them directly ?

hoshibara: Exactly, what we want to do is to aggregate data preprocess logic in one or two classes. In addition, the previous `get_forward_args`, `get_forward_defaults` and `get_forward_annotations` have redundancy code.

MeouSker77: oh, I see

PR #7680, opened 2023-02-24
Pytorch Modelcheckpoint to support step

jenniew:
Pytorch Modelcheckpoint to support checkpoint by steps.
If every_n_train_steps = None or <=0, then it will save model checkpoint after each epoch.


leonardozcm:
Duplicated with https://github.com/intel-analytics/BigDL/pull/7712

PR #7682, opened 2023-02-24
Nano: allow pipeline to specify CPU cores for each stage

MeouSker77:
## Description

Improve pipeline: Allow pipeline to specify CPU cores for each stage.
And fix nano nightly test: windows doesn't support pickle local object

### 1. Why the change?

For better performance

### 2. User API changes

```python
from bigdl.nano.pytorch import Pipeline

pipeline = Pipeline([
    ("preprocess", preprocess, {"core_num": 4}),
    ("inference", inference, {"core_num": 8})
])
```

### 3. Summary of the change 

- Allow pipeline to specify CPU cores for each stage.
- Add a new utils `EnvContext` to simplify setting environment variables temporarily.

### 4. How to test?

- [ ] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/inference/pipeline.py
+++ b/python/nano/src/bigdl/nano/pytorch/inference/pipeline.py
@@ -52,8 +52,9 @@ class Pipeline():
             ])
         ```
         will create a pipeline which has stage "preprocess" and "inference",
-        and the "preprocess" stage will call `preprocess(input)` with 4 CPU cores,
-        while the "inference" stage will call `model(input)` with 8 CPU cores.
+        and the "preprocess" stage will call `preprocess(input)` with 4 CPU cores
+        for openmp worload, while the "inference" stage will call `model(input)`
+        with 8 CPU cores for openmp workload.
 
         :param stages: A list of configurations for each stage, each stage should consist of
             a 'name'(str), a 'function'(Callable), and a 'config'(dict).
--- a/python/nano/src/bigdl/nano/pytorch/inference/pipeline.py
+++ b/python/nano/src/bigdl/nano/pytorch/inference/pipeline.py
@@ -14,12 +14,17 @@
 # limitations under the License.
 #
 
+import os
+import warnings
 import itertools
 import multiprocessing as mp
 from multiprocessing.connection import Connection
 from typing import List, Tuple, Dict, Callable
 
 from torch import nn
+
+from bigdl.nano.utils.common import schedule_workers
+from bigdl.nano.utils.common import EnvContext
 
 
 class Pipeline():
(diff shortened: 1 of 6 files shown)


python/nano/src/bigdl/nano/utils/common/env.py
```diff
@@ -62,3 +62,37 @@ def _find_library(library_name: str, priority_dir: Union[str, None] = None) -> U
...
+        if self.new_env is not None:
+            os.environ.update(self.new_env)
+
+    def __exit__(self, exc_type, exc_value, exc_tb):
+        # it seems `os.environ = self.env_backup` will cause bug
+        for k, v in self.env_backup.items():
+            if k not in os.environ:
+                os.environ.pop(k)
+            else:
+                os.environ[k] = v
```
TheaperDeng, line 98: does this exit pop out those key only in `new_env`?

MeouSker77: oh, it's a bug, I'll fix it

MeouSker77: fixed it

rnwang04 (approved):
LGTM

python/nano/src/bigdl/nano/pytorch/inference/pipeline.py
```diff
@@ -91,10 +97,23 @@ def run(self, inputs):
...
+        if isinstance(config.get("core_num", None), int):
+            core_num = config["core_num"]
+            cores = self.cores[:core_num]
+            self.cores = self.cores[core_num:]
+            if len(cores) < core_num:
+                warnings.warn(f"stage {name} requires {core_num} cores,"
+                              f" but there are only {len(cores)} cores left")
+            subprocess_env["KMP_AFFINITY"] = (f"granularity=fine,proclist"
+                                              f"=[{','.join([str(i) for i in cores])}],explicit")
+            subprocess_env["OMP_NUM_THREADS"] = str(len(cores))
```
TheaperDeng, line 110: I think we may just allow users to state the environment variables themselves. `core_num` seems to be a good shortcut for users but shortcut always has some limitations (e.g., this `core_num` only controls the openmp workload's thread num with a specific affinity strategy)

python/nano/src/bigdl/nano/pytorch/inference/pipeline.py
```diff
@@ -42,13 +47,13 @@ def __init__(self, stages: List[Tuple]):
...
-                ("inference", model, {}),
+                ("preprocess", preprocess, {"core_num": 4}),
+                ("inference", model, {"core_num": 8}),
             ])
         ```
         will create a pipeline which has stage "preprocess" and "inference",
-        and the "preprocess" stage will call `preprocess(input)`, while the
-        "inference" stage will call `model(input)`.
+        and the "preprocess" stage will call `preprocess(input)` with 4 CPU cores,
+        while the "inference" stage will call `model(input)` with 8 CPU cores.
```
TheaperDeng, line 56: +"for openmp workload" maybe

TheaperDeng (approved):

python/nano/src/bigdl/nano/pytorch/inference/pipeline.py
MeouSker77: We may provide another key named `env` for users to override the environment variables of subprocess ? e.g.
`pipeline = Pipeline([("preprocess", preprocess, {"core_num": 4, "env": {"KMP_AFFINITY": "granularity=fine,none"}}])`

python/nano/src/bigdl/nano/pytorch/inference/pipeline.py
MeouSker77: done

yangw1234:
Do we have documentation on when and how to use the `pipeline` feature?

MeouSker77:
no, I think it's still in a very early state

> Do we have documentation on when and how to use the `pipeline` feature?

PR #7528, opened 2023-02-13
Orca:  Polish Resnet50 Inference Example

leonardozcm:
## Description

https://github.com/intel-analytics/BigDL/issues/6955

PR #7687, opened 2023-02-24
Orca: Support Customized Pocket Attributes in Torchrunner

leonardozcm:
## Description

We need to provide users with such a storage place to assign their own attributes to be utilized across hooks. At the same time we    would better avoid them to directly assign these attrs to torchrunner.

https://github.com/intel-analytics/BigDL/pull/7528#discussion_r1112634530

### DIFF
--- a/python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
@@ -131,7 +131,7 @@ class TorchRunner(BaseRunner):
         self.sync_stats = sync_stats
         self.epoch_stats = None  # The state saved in every epoch
         self._mode = 'val'  # By default we don't use ddp model
-        self.pocket = dict() # Used to store some customized attributes
+        self._pocket = dict() # Used to store some customized attributes
 
     def _create_loss(self):
         if not self.loss_creator:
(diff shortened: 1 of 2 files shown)


python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
```diff
@@ -131,6 +131,7 @@ def __init__(self,
         self.sync_stats = sync_stats
         self.epoch_stats = None  # The state saved in every epoch
         self._mode = 'val'  # By default we don't use ddp model
+        self.pocket = dict() # Used to store some customized attributes
```
hkvision, line 134: What's the difference between using a dict VS directly assigning `runner.images = ...` in the hook?

python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
```diff
@@ -748,6 +749,21 @@ def call_hook(self, callbacks, fn_name: str) -> None:
         for hook in callbacks:
             if hasattr(hook, fn_name):
                 getattr(hook, fn_name)(self)
+    
+    def put(self, k, v):
+        self.pocket[k] = v
```
hkvision, line 754: better handle some overwrite cases? (e.g. print a warning?)

python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
leonardozcm: > What's the difference between using a dict VS directly assigning `runner.images = ...` in the hook?

To avoid users assign and occupy runner's reserved attrs like `batch` or `loss` by mistake, so we open up a new field for them to store whatever they want.

python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
leonardozcm: sure

python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
hkvision: Theoretically if users don't assign the reserved values, the two ways are equivalent right?
So the question is do we have way to prevent users to override reserved values? Probably no for now?

leonardozcm: > Theoretically if users don't assign the reserved values, the two ways are equivalent right? So the question is do we have way to prevent users to override reserved values? Probably no for now?

Yes, but we can keep this for now i think? For 1. most of time users are not awared of what attribute names are availble.  2. There two kinds of way of assignment are not conflicting.

PR #7688, opened 2023-02-24
Chronos: add `inference_mode` in context manager

plusbang:
## Description

Currently, chronos context manager only provides thread control and autocast.
Recent experiment results shows that if provide `inference_mode` in context manager, inference overhead can be less.
For example, on ETTh1 dataset, onnx latency: 0.044ms -> 0.038ms

### 2. User API changes

None

### 3. Summary of the change 

- add `inference_mode` in context manager
- remove `with torch.inference_mode()` in some internal implementation
- add `with torch.inference_mode()` and `with self.cxt_manager` in `tspipeline` and `evaluate()`
- move import in `tspipeline`

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/chronos/src/bigdl/chronos/autots/tspipeline.py
+++ b/python/chronos/src/bigdl/chronos/autots/tspipeline.py
@@ -22,6 +22,8 @@ import numpy as np
 
 from bigdl.chronos.data import TSDataset
 from bigdl.chronos.metric.forecast_metrics import Evaluator
+from bigdl.chronos.pytorch.utils import _pytorch_fashion_inference
+from bigdl.nano.utils.common import invalidInputError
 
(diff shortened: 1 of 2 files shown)


python/chronos/src/bigdl/chronos/pytorch/context_manager.py
```diff
@@ -37,10 +37,11 @@ class ForecasterContextManager(object):
     """
     This context manager is used for Pytorch Model Inference.
```
TheaperDeng, line 38: we should add this `inference_mode` to dummy context mgr right?

TheaperDeng (approved):

TheaperDeng (approved):
LGTM

Issue #7626, opened 2023-02-20
Enhancement on PyTorch Estimator spark backend


hkvision:
https://github.com/analytics-zoo/Telecom-CTC-E2E/issues/26

- Do not broadcast when model_dir is None
- Only rank0 worker return stats

PR #7698, opened 2023-02-27
[ORCA] Reduce Training Time Cost with PyTorch Pyspark Estimator

sgwhat:
## Description
For https://github.com/intel-analytics/BigDL/issues/7626 to improve pytorch pyspark backend training performance.

### 1. Why the change?
To minimize the overhead during training with pytorch pyspark backend.

### 2. User API changes
N/A

### 3. Summary of the change 
Disable broadcast for first training and setup only rank 0 worker will return model states to reduce spark overhead. 

### 4. How to test?
- [x] Unit test
- [x] Application test

### DIFF
--- a/python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
@@ -141,6 +141,7 @@ class PyTorchPySparkEstimator(BaseEstimator):
 
         self.model_creator = model_creator
         self.optimizer_creator = optimizer_creator
+        self.state_dict = None
 
         num_nodes, cores_per_node = get_node_and_core_number()
         self.num_workers = num_nodes * workers_per_node
--- a/python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
@@ -363,11 +363,13 @@ class PyTorchPySparkEstimator(BaseEstimator):
             worker_stats = res
         else:
             # state dicts of all runners would be the same
-            state_list = [item for item in res if isinstance(item, dict)]
+            for item in res:
+                if isinstance(item, dict):
+                    self.state_dict = item
+                    break
             # Each runner would return a list of worker stats for different epochs
             worker_stats = [item for item in res if isinstance(item, list)]
 
-        self.state_dict = state_list[0]
         epoch_stats = list(map(list, zip(*worker_stats)))
         if reduce_results:
             for i in range(len(epoch_stats)):
--- a/python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
@@ -183,7 +183,6 @@ class PyTorchPySparkEstimator(BaseEstimator):
             cluster_info=self._get_cluster_info(sc),
             **local_init_params)
 
-
     def create_tcpstore_server(self) -> 'TCPStore':
         import torch.distributed as dist
         server_store = dist.TCPStore(self.ip, self.tcp_store_port, -1, True,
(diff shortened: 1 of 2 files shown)


python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_worker.py
```diff
@@ -126,32 +126,38 @@ def setup_distributed(self, mode, cluster_info, driver_ip, driver_tcp_store_port
     def train_epochs(self, data_creator, epochs=1, batch_size=32, profile=False,
                      wrap_dataloader=None, callbacks=None,
                      validation_data_creator=None):
-        self.load_state_dict(self.state_dict.value)
+        if self.state_dict:
+            self.load_state_dict(self.state_dict.value)
```
hkvision, line 130: Will this get improvement?
Since for the logic in estimator, if there's model creator, there will always be self.state_dict?

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_worker.py
```diff
@@ -126,32 +126,38 @@ def setup_distributed(self, mode, cluster_info, driver_ip, driver_tcp_store_port
...
 
         if self.model_dir is not None:
             return [stats_list]
         else:
-            return [state_dict, stats_list]
+            if self.rank == 0:
+                state_dict = self.get_state_dict()
+                return [state_dict, stats_list]
+            else:
+                return [stats_list]
```
hkvision, line 154: maybe we can combine these two if parts?

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_worker.py
```diff
@@ -126,32 +126,38 @@ def setup_distributed(self, mode, cluster_info, driver_ip, driver_tcp_store_port
...
         if self.rank == 0:
             if self.model_dir is not None:
+                state_dict = self.get_state_dict()
                 save_pkl(state_dict, os.path.join(self.model_dir, "state.pkl"))
 
         if self.model_dir is not None:
             return [stats_list]
         else:
-            return [state_dict, stats_list]
+            if self.rank == 0:
```
hkvision, line 150: I think we should also modify Line363 in pytorch_pyspark_estimator?
```
            self.state_dict = res[0]  # state dicts of all runners would be the same
            # Each runner would return a list of worker stats for different epochs
            worker_stats = [item for item in res if isinstance(item, list)]
```
rank0 worker's result doesn't necessarily comes first?

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_worker.py
sgwhat: `self.state_dict` will be none if `epoch == 0`.

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
```diff
@@ -360,10 +362,12 @@ def transform_func(iter, init_param, param):  # type:ignore
             self.state_dict = PyTorchPySparkEstimator._get_state_dict_from_remote(self.model_dir)
             worker_stats = res
         else:
-            self.state_dict = res[0]  # state dicts of all runners would be the same
+            # state dicts of all runners would be the same
+            state_list = [item for item in res if isinstance(item, dict)]
```
hkvision, line 366: stats_list not state_list?

hkvision, line 366: This should just be self.state_dict?

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
```diff
@@ -360,10 +362,12 @@ def transform_func(iter, init_param, param):  # type:ignore
...
             # Each runner would return a list of worker stats for different epochs
             worker_stats = [item for item in res if isinstance(item, list)]
 
+        self.state_dict = state_list[0]
```
hkvision, line 370: this is not necessary, see the comment above

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
```diff
@@ -360,7 +362,11 @@ def transform_func(iter, init_param, param):  # type:ignore
             self.state_dict = PyTorchPySparkEstimator._get_state_dict_from_remote(self.model_dir)
             worker_stats = res
         else:
-            self.state_dict = res[0]  # state dicts of all runners would be the same
+            # state dicts of all runners would be the same
```
hkvision, line 365: This comment should be modified to: only the state_dict of the rank 0 worker would be returned?

hkvision (approved):

Issue #7348, opened 2023-01-24
[Orca] Support multiple outputs of orca estimator predict


songhappy:
Current Orca estimator predict only handles model with one output. Many models have multiple outputs, e.g. image object detection models: detection.fasterrcnn_resnet50_fpn, fasterrcnn_mobilenet_v3_large_320_fpn, detection.retinanet_resnet50_fpn the output is a dictionary of "boxes", "labels", "scores", Orca estimator need to handle it correctly.

https://pytorch.org/tutorials/intermediate/torchvision_tutorial.html

hkvision:
What do you think is the reasonable way for this? For example,
- For Spark DataFrame, we output the df with more than one extra columns, how to get the names of the outputs? (Currently the colname is always "prediction") [For SparkXShards of pandas df, same]
- For SparkXShards of dict, the output shard dict should be {"y": [y1, y2, ]}?

sgwhat:
> 



> What do you think is the reasonable way for this? For example,
> 
> * For Spark DataFrame, we output the df with more than one extra columns, how to get the names of the outputs? (Currently the colname is always "prediction") [For SparkXShards of pandas df, same]
> * For SparkXShards of dict, the output shard dict should be {"y": [y1, y2, ]}?

For spark df, maybe we could retrieve the columns' names from the model or users provided?

hkvision:
probably we can add an optimal argument for output_col_names?
Can we always get the names from models? (probably yes?

sgwhat:
> probably we can add an optimal argument for output_col_names? Can we always get the names from models? (probably yes?

I think for tensorflow multi-output model, most cases will have a parameter `output` when creating the model. So it's possible for us to get the names, am I right?

hkvision:
Probably we can use the optimizations in https://github.com/intel-analytics/BigDL/pull/5952 to support multi-outputs as well as enhance performance converting rdd to dataframe with arrow.

sgwhat:
Currently we have supported multi-output prediction model with both pytorch and tensorflow, for the limitations and next step preview plan, we are going to:
- [ ] Support model outputs with different shapes of results.
- [ ] Support automatically allocating output column names.

hkvision:
Seems automatically allocating output column names is not feasible in PyTorch?

sgwhat:
> Seems automatically allocating output column names is not feasible in PyTorch?

[...]

PR #7704, opened 2023-02-27
Orca: Support Multiple outputs in prediction

leonardozcm:
### 1. Why the change?

We used to assume our predict output is a single ndarray, but there are many models have multiple outputs. This PR enables multiple outputs in prediction with hook mechanism.

### 2. User API changes

Add callback point `on_pred_forward` in MainCallback for users to implement their own logic.

### 3. Summary of the change 

1. Add callback point `on_pred_forward` in MainCallback
2. Recursively aggregate various forms of outputs.

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

Related issue: https://github.com/intel-analytics/BigDL/issues/7348


python/orca/src/bigdl/orca/learn/utils.py
```diff
@@ -180,31 +185,31 @@ def flatten(data):
...
+            return {k: convert_elem(v) for k, v in elem.items()}
+        # scalar in basic type
+        elif not isinstance(elem, np.ndarray) and len(elem.shape) == 0:
+            return float(elem.item(0))
         # np ndarray
         else:
-            dim = len(pair[1].shape)
-            if dim == 1:
+            dim = len(elem.shape)
+            if dim in [0, 1]:
```
sgwhat, line 203: Shall we consider dim is 0?

leonardozcm: we will revert this in next pr.

PR #7707, opened 2023-02-28
pytorch object detection example

songhappy:
### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

### 5. New dependencies

- [ ] New Python dependencies
       - Dependency1 
       - Dependency2
       - ...
       - Dependency1 and license1
       - Dependency2 and license2
       - ...

### DIFF
--- a/ppml/trusted-big-data-ml/scala/docker-occlum/kubernetes/run_spark_tpch.sh
+++ b/ppml/trusted-big-data-ml/scala/docker-occlum/kubernetes/run_spark_tpch.sh
@@ -4,7 +4,7 @@ ${SPARK_HOME}/bin/spark-submit \
     --master k8s://https://${kubernetes_master_url}:6443 \
     --deploy-mode cluster \
     --name spark-tpch \
-    --class main.scala.TpchQuery \
+    --class com.intel.analytics.bigdl.ppml.examples.tpch.TpchQuery \
     --conf spark.rpc.netty.dispatcher.numThreads=32 \
     --conf spark.kubernetes.container.image=intelanalytics/bigdl-ppml-trusted-big-data-ml-scala-occlum:2.3.0-SNAPSHOT \
     --conf spark.kubernetes.container.image.pullPolicy="IfNotPresent" \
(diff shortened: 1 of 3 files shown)


python/orca/tutorial/xshards/images/torchvision_fineturning_exp.py
```diff
@@ -0,0 +1,213 @@
...
+            for it in iterator:
+                feature, target = it[0], it[1]
+                features.append(feature)
+                for key in keys:
+                    targets[key].append(target[key])
+            out = {'x': np.array(features).astype(np.float32),
+                   'y': targets}  # dictionary of list
+            return [out]
+        rdd = data.rdd.mapPartitions(lambda x: per_partition(x))
+        return SparkXShards(rdd)
```
leonardozcm, line 154: Actually maybe you do not need this, for this sort of logic has already been implemented in https://github.com/intel-analytics/BigDL/blob/main/python/orca/src/bigdl/orca/data/utils.py#L211
```python
def partition_get_data_label(partition_data,
                             allow_tuple=True,
                             allow_list=True,
                             has_label=True):
    """
    :param partition_data: The data partition from Spark RDD, which should be a list of records.
    :param allow_tuple: Boolean. Whether the model accepts a tuple as input. Default is True.
    :param allow_list: Boolean. Whether the model accepts a list as input. Default is True.
    :param has_label: Boolean. Whether the data partition contains labels.
    :return: Concatenated data for the data partition.
    """
    data_list = [data['x'] for data in partition_data]
    label_list = [data['y'] for data in partition_data]

    data = _convert_list_tuple(combine(data_list),
                               allow_tuple=allow_tuple, allow_list=allow_list)
    if has_label:
        label = _convert_list_tuple(combine(label_list),
                                    allow_tuple=allow_tuple, allow_list=allow_list)
    else:
        label = None

    return data, label
```
in `combine` https://github.com/intel-analytics/BigDL/blob/main/python/orca/src/bigdl/orca/data/utils.py#L178-L181:
```python
def combine(data_list):
    item = data_list[0]
    if isinstance(item, dict):
        res = {}
        for k, v in item.items():
            res[k] = np.concatenate([data[k] for data in data_list], axis=0)
```
This will combine the format that you want here as:
```
{'x': [img1, img2, ...], 'y': {'boxes': [ndarray1, ndarray2,...]}}
```

python/orca/tutorial/xshards/images/torchvision_fineturning_exp.py
```diff
@@ -0,0 +1,213 @@
...
+        print(len(batch[0]))
+        print(type(batch))
+        print(type(batch[0]))
+        print(type(batch[0][0]))
+        print(type(batch[0][1]))
+        print(batch[0][0].shape)
+        print(batch[0][1].keys())
+        for key in batch[0][1].keys():
+            print(batch[0][1][key].shape)
+        out = tuple(zip(*batch))
```
leonardozcm, line 181: You can select target keys ["boxes", "labels", "masks"] here

leonardozcm: out -> (tuple of image, tuple of dict)

python/orca/tutorial/xshards/images/torchvision_fineturning_exp.py
songhappy: moved to another PR

Issue #7105, opened 2022-12-27
Orca: Hook Mechanism TaskBoard


leonardozcm:
## Following Tasks

- [ ] Insert Base Hook Func into TorchRunner https://github.com/intel-analytics/BigDL/pull/7100
- [ ] In `before_train_iter`, restore batch: [x1, x2], y -> x1, x2, y
- [ ] Assign some attrs to runner(self.batch, self.loss etc.)
- [ ] Implement MainCallBacks
- [ ] Migrate Existing CallBacks: TensorBoard, ModelCheckPoint
- [ ] CleanTorchRunner
- [ ] User Interfaces: estimator.fit
- [ ] Example: Resnet50 Inference
- [ ] Doc

jenniew:
For huggingface based training:
- [ ] Support model input is a dictionary which includes labels. Also in current torch runner, if feature is not list nor tensor, it will trigger error. We may remove this restriction. Maybe put all these data processing and check to hook function and provide default implementation to support various data source type. 
- [ ] Support train with max steps
- [ ] Support ModelCheckpoint on step_frequency

leonardozcm:
- [ ] Pre-check tqdmcallback
- [ ] Add tqdm supports in evaluate and predict
- [ ] Suggest users to write all logics in Maincallback.

songhappy:
Orca estimator needs to support multiple outputs
https://github.com/intel-analytics/BigDL/blob/main/python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py#L647
https://github.com/intel-analytics/BigDL/issues/7348

leonardozcm:
> Orca estimator needs to support multiple outputs code: https://github.com/intel-analytics/BigDL/blob/main/python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py#L647
> 
> issue: #7348

https://github.com/intel-analytics/BigDL/actions/runs/4219544101/jobs/7325009597

leonardozcm:
https://github.com/intel-analytics/BigDL/issues/7433#issuecomment-1439073600

PR #7712, opened 2023-02-28
Support save model checkpoint by step frequency

PatrickkZ:
## Description

Implement `after_train_iter`  method in ModelCheckpoint callback to support saving model checkpoint by step frequency. 
### 1. Why the change?
related issue: https://github.com/intel-analytics/BigDL/issues/7105#issuecomment-1419996858

### 2. User API changes

Add two parameters when creating the ModelCheckpoint callback.
- **by_epoch**: save chekpoint by epoch or by iteration, default is True
- **interval**: The saving period. If `by_epoch=True`, interval indicates epochs, otherwise it indicates iterations. Default is -1, which means "never".

To save model checkpoint by step frequency, we can create a ModelCheckpoint callback as below:
```python
# create estimator
estimator = ...

# create callbacks
callbacks = [
    ModelCheckpoint(filepath=ckpt_save_path, # filepath should contains 'epoch' or 'iter' like this 'my/path/test-{iter:02d}'
    save_weights_only=True,
    by_epoch=False,
    interval=5)
]

# use callbacks when training
estimator.fit(data, batch_size=batch_size, epochs=epochs,
    callbacks=callbacks,
    feature_cols=["feature"],
    label_cols=["label"])
```

### 3. Summary of the change 

Implement `after_train_iter`  method in ModelCheckpoint callback to support saving model checkpoint by step frequency. 

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/python/orca/src/bigdl/orca/learn/pytorch/callbacks/model_checkpoint.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/callbacks/model_checkpoint.py
@@ -28,10 +28,10 @@ class ModelCheckpoint(Callback):
     FILE_EXTENSION = ".ckpt"
 
     def __init__(self,
-                 filepath=None,
-                 save_weights_only=False,
-                 by_epoch=True,
-                 interval=-1,
+                 filepath: str = "",
+                 save_weights_only: bool = False,
+                 by_epoch: bool = True,
+                 interval: int = -1,
                  ):
         """
         ModelCheckpoint callback is used in conjunction with training using estimator.fit() to save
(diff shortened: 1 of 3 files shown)


jenniew:
> 

I think if by_epoch=True(default value), and interval=-1(default value), we may need to save checkpoint at each epoch, which is the common action just like pytorch lightning modelcheckpoint.

python/orca/src/bigdl/orca/learn/pytorch/callbacks/model_checkpoint.py
```diff
@@ -57,12 +64,25 @@ def after_train_epoch(self, runner):
         be called during TRAIN mode.
         :param epoch:  Integer, index of epoch.
         """
+        if not self.by_epoch:
+            return
```
jenniew, line 68: We need to check self.interval, and only checkpoint at specific epochs.

python/orca/src/bigdl/orca/learn/pytorch/callbacks/model_checkpoint.py
```diff
@@ -30,6 +30,8 @@ class ModelCheckpoint(Callback):
     def __init__(self,
                  filepath=None,
                  save_weights_only=False,
+                 by_epoch=True,
+                 interval=-1,
                  ):
```
jenniew, line 35: Please add type hint for the init function

leonardozcm (approved):
LGTM

jenniew (approved):

python/orca/src/bigdl/orca/learn/pytorch/callbacks/model_checkpoint.py
```diff
@@ -123,8 +150,8 @@ def _format_checkpoint_name(cls, dirname, filename, stats=None):
             for group in groups:
                 name = group[1:]
 
-                if "epoch" not in name:
-                    warnings.warn("We only support filepath with {epoch} for now.")
+                if "epoch" not in name and "iter" not in name:
+                    warnings.warn("We only support filepath with {epoch} or {iter} for now.")
```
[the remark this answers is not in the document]
PatrickkZ: the file path should have the format {epoch} or {iter}
```python
# example
filepath='my/path/sample-{epoch:02d}'
filepath='my/path/sample-{iter}'
```

PR #7737, opened 2023-03-02
Nano: fix trace custom model which doesn't support `compute_output_shape`

MeouSker77:
## Description

Fix trace custom models which do not support `compute_output_shape`.
For models which do not support the `compute_output_shape` method, we now fake a tensor from input spec to do a inference(except for onnx, onnx does not need this inference), so it can be traced by openvino, and can be quantized by inc.
However, these models will not support `evaluate` method.


python/nano/src/bigdl/nano/deps/onnxruntime/tensorflow/tensorflow_onnxruntime_model.py
```diff
@@ -53,17 +53,10 @@ def __init__(self, model, input_spec=None, onnxruntime_session_options=None,
...
-                elif input_spec is not None:
-                    input_shape = tensor_spec_to_shape(input_spec)
-                    self._output_shape = model.compute_output_shape(input_shape)
-                else:
-                    invalidInputError(False,
-                                      "Subclassed model must specify `input_spec` parameter.")
+                self._output_shape = try_compute_output_shape(model, input_spec,
+                                                              try_fake_inference=False)
+                if input_spec is None and hasattr(model, "input_shape"):
+                    input_spec = input_spec = tf.TensorSpec(model.input_shape, model.dtype)
```
Oscilloscope98, line 59: duplicated `input_spec` here?

Oscilloscope98 (approved):
Others LGTM

python/nano/src/bigdl/nano/deps/onnxruntime/tensorflow/tensorflow_onnxruntime_model.py
MeouSker77: fixed

rnwang04 (approved):
LGTM

PR #7738, opened 2023-03-02
[PPML] Add ppml Decrypt code

hzjane:
## Description
Add ppml Decrypt code


Uxito-Ada:
Can we combine `Encrypt` and `Decrypt` into one and append an argument named `--action` instead of repeating codes?

hzjane:
> Can we combine `Encrypt` and `Decrypt` into one and append an argument named `--action` instead of repeating codes?



> Can we combine `Encrypt` and `Decrypt` into one and append an argument named `--action` instead of repeating codes?

Good idea！

qiyuangong (approved):
LGTM

Uxito-Ada:
Pls also modify *Encrypt and upload data* section in PPML README after `Encrypt` API is updated.

PR #7749, opened 2023-03-02
Nano: add output_tensors parameter for `quantize` and `optimize`

plusbang:
## Description

Add `output_tensors` parameter for `quantize` and `optimize`, so users can specify the output is numpy.ndarray. (avoid some unnecessary type conversion)

### 1. Why the change?

follow up https://github.com/intel-analytics/BigDL/pull/7638.

### 2. User API changes

add new param `output_tensors` in `InferenceOptimizer.quantize` and `InferenceOptimizer.optimize`
```bash
def optimize(..., output_tensors: bool = True, ...)
def quantize(..., output_tensors: bool = True, ...)
# output_tensors: boolean, default to True and output of the model will be Tensors, only valid when accelerator='onnxruntime' or accelerator='openvino', otherwise will be ignored. If output_tensors=False, output of the export model will be ndarray.
```

### 3. Summary of the change 

- add new param `output_tensors` in `InferenceOptimizer.quantize` and `InferenceOptimizer.optimize`
- add related uts
- modify internal implementation in Chronos

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/nano/test/onnx/pytorch/test_inc_onnx.py
+++ b/python/nano/test/onnx/pytorch/test_inc_onnx.py
@@ -432,6 +432,27 @@ class TestOnnx(TestCase):
         result_m = accmodel(x1,x2,x3,x4, np.array([2]))
         assert abs(torch.sum(result_m).item()) < 1e-5 + 6
 
+    def test_onnx_quantize_output_tensors(self):
+        model = ResNet18(10, pretrained=True, include_top=False, freeze=True)
+        loss = nn.CrossEntropyLoss()
+        optimizer = torch.optim.Adam(model.parameters(), lr=0.01)
+
+        pl_model = Trainer.compile(model, loss, optimizer)
+        x = torch.rand((10, 3, 256, 256))
+        y = torch.ones((10, ), dtype=torch.long)
+        ds = TensorDataset(x, y)
+        train_loader = DataLoader(ds, batch_size=2)
+
+        onnx_model = InferenceOptimizer.quantize(pl_model, accelerator='onnxruntime',
+                                                 method='qlinear', calib_data=train_loader)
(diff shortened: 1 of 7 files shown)


python/nano/test/onnx/pytorch/test_inc_onnx.py
```diff
@@ -432,6 +432,27 @@ def forward(self, x1, x2, x3, x4, a=None, b=None, c=None):
...
+        onnx_model = InferenceOptimizer.quantize(pl_model, accelerator='onnxruntime',
+                                                 method='qlinear', calib_data=train_loader)
+        test_onnx_model = InferenceOptimizer.quantize(pl_model, accelerator="onnxruntime",
+                                                      method='qlinear', calib_data=train_loader,
+                                                      output_tensors=False)
+        for x, y in train_loader:
+            model.eval()
+            forward_res_tensor = onnx_model(x).numpy()
+            forward_res_numpy = test_onnx_model(x)
+            np.testing.assert_almost_equal(forward_res_tensor, forward_res_numpy, decimal=5)
```
MeouSker77, line 455: I remember `np.testing.assert_almost_equal` can also be used to compare tensors, so I think we can check whether `forward_res_numpy` is ndarray explicitly.

MeouSker77 (approved):
others LGTM

python/nano/test/onnx/pytorch/test_inc_onnx.py
plusbang: > I remember `np.testing.assert_almost_equal` can also be used to compare tensors, so I think we can check whether `forward_res_numpy` is ndarray explicitly.

yeah, `np.testing.assert_almost_equal` indeed support tensors as input. I add type check in uts directly.

PR #7750, opened 2023-03-02
Pytorch Estimator support more input type

jenniew:
Pytorch Estimator support more input type.

### DIFF
--- a/python/orca/src/bigdl/orca/learn/pytorch/callbacks/maincallback.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/callbacks/maincallback.py
@@ -15,7 +15,9 @@
 #
 from .base import Callback
 from bigdl.dllib.utils.log4Error import invalidInputError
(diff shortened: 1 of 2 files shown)


python/orca/src/bigdl/orca/learn/pytorch/callbacks/maincallback.py
```diff
@@ -91,3 +93,29 @@ def on_val_forward(self, runner):
         Any behavior inconsistent with the default training behavior should be overridden here.
         """
         self.on_iter_forward(runner)
+
+    def before_train_iter(self, runner):
+        features, target = runner.batch
+        if torch.is_tensor(features):
+            runner.batch = features, target
+        elif isinstance(features, (tuple, list)):
+            runner.batch = *features, target
```
leonardozcm, line 102: I don't think this is a good idea to expose this to users, for these lines are intended to restore `batch` to the format users expects, see pr https://github.com/intel-analytics/BigDL/pull/5763
You may discuss whether runner.batch could be unpack in  `features, target = runner.batch`, if not then we assume it is a custom data format like a single dict {'x': x,'y':y}.

PR #7751, opened 2023-03-02
Pytorch Estimator support train on steps

jenniew:
Pytorch Estimator support train on max steps.

### DIFF
--- a/python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
@@ -199,6 +199,7 @@ class PyTorchPySparkEstimator(BaseEstimator):
     def fit(self,
             data: Union['SparkXShards', 'SparkDataFrame', Callable[[Dict, int], 'DataLoader']],
             epochs: int=1,
+            max_steps: Optional[int]=None,
             batch_size: int=32,
             profile: bool=False,
             reduce_results: bool=True,
@@ -294,6 +295,7 @@ class PyTorchPySparkEstimator(BaseEstimator):
 
         params = dict(
             epochs=epochs,
+            max_steps=max_steps,
             batch_size=batch_size,
             profile=profile,
             callbacks=callbacks
--- a/python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
@@ -219,6 +219,8 @@ class PyTorchPySparkEstimator(BaseEstimator):
                takes config and batch_size as argument and returns a PyTorch DataLoader for
                training.
         :param epochs: The number of epochs to train the model. Default is 1.
+        :param max_steps: The max steps to train the model. Default is None.
+         If max_steps > 0, `epochs` would be ignored.
         :param batch_size: Total batch size for all workers used for training. Each worker's batch
                size would be this value divide the total number of workers. Default is 32.
                If your training data is a function, you can set batch_size to be the input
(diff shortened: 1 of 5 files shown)


python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
```diff
@@ -244,13 +244,18 @@ def train_epochs(self, data_creator, epochs=1, batch_size=32, profile=False,
         self.val_loader = val_loader
         self.call_hook(callbacks=callbacks, fn_name="before_run")
 
+        if max_steps is not None:
+            epochs = math.ceil(max_steps / len(self.train_loader))
```
leonardozcm, line 248: So `max_step` indicates a global step?

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
```diff
@@ -199,6 +199,7 @@ def _get_cluster_info(self, sc):
     def fit(self,
             data: Union['SparkXShards', 'SparkDataFrame', Callable[[Dict, int], 'DataLoader']],
             epochs: int=1,
+            max_steps: Optional[int]=None,
```
leonardozcm, line 202: Also in ray estimator?

python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
```diff
@@ -361,17 +367,19 @@ def train_epoch(self, ...):
...
 
         self.call_hook(callbacks=callbacks, fn_name="on_lr_adjust")
 
         return metric_meters.summary(sync_stats=self.sync_stats,
                                      dist_backend=self.dist_backend)
 
-    def _train_loop(self, iterator, metric_meters, callbacks):
+    def _train_loop(self, iterator, metric_meters, callbacks, max_steps=None):
         for batch_idx, batch in enumerate(iterator):
+            if max_steps is not None and self.global_step >= max_steps:
```
leonardozcm, line 381: I prefer to integrate this into a EarlyStopping Callback after this merges? @hkvision We shall take it carefully to add any new arg in torchrunner, for they are really hard to be maintained.

jenniew: earlystop is to stop with metric criteria. I don't think we need to integrate training with max steps to earlystop as max step means training only specific step without any condition. We may refer to pytorch lightning, max steps is a trainer options. EarlyStop is callback to stop when some condition met.

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
jenniew: Yes, will also add to ray estimator.

python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
jenniew: global step is runner status, max steps is a training option for just training specific steps.

leonardozcm: Yes, I mean the max_step is **compared** with the global step across all ranks?

python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
leonardozcm: You can keep this in this pr, and we may refactor this just like what we do in TqdmCallback? @hkvision  I think training only specific steps is also one kind of condition. What about we set up a flag like `runner.stop`, when any condition reaches we change `runner.stop` to True and runner checks it and exits in EarlyStop hooks.

PR #7757, opened 2023-03-03
add tdx cc deep learning

Le-Zheng:
## Description

Add tdx cc deep learning Dockerfile


glorysdj:
The Docker images can be used by both TDXCC and TDXVM, we may organize them as:
/tdx/docker/trusted-bigd-data
/tdx/docker/trusted-deep-learning
/tdx/docker/trusted-dl-serving

glorysdj (approved):
LGTM

PR #7762, opened 2023-03-03
Orca: fix ut `test_reinit_raycontext.py`

hamham223:
### 1. Why the change?

+ Fix the issue in `test_reinit_raycontext.py` and re-enable the ut.
+ The Process Monitor only works in cluster mode, so the corresponding test are commented.
+ api fix: `ray_ctx._ray_on_spark_context.ray_processesMonitor.process_infos`

### 2. How to test?
- [ ] Orca-Pytest-Ray-Py37-Spark3
- [ ] Orca-Pytest-Ray-Py38-Spark3

### DIFF
--- a/python/orca/test/bigdl/orca/ray/test_reinit_raycontext.py
+++ b/python/orca/test/bigdl/orca/ray/test_reinit_raycontext.py
@@ -54,9 +54,10 @@ class TestUtil(TestCase):
             ray_ctx.stop()
             sc.stop()
             time.sleep(3)
-            for process_info in ray_ctx.ray_processesMonitor.process_infos:
-                for pid in process_info.pids:
-                    assert not psutil.pid_exists(pid)
+            # TODO: add the process monitor test in cluster mode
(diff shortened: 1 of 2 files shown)


hamham223:
**Now all uts under `test/bigdl/orca/` are tested!!!** 🎉

python/orca/test/bigdl/orca/ray/test_reinit_raycontext.py
```diff
@@ -54,9 +54,10 @@ def test_local(self):
             ray_ctx.stop()
             sc.stop()
             time.sleep(3)
-            for process_info in ray_ctx.ray_processesMonitor.process_infos:
-                for pid in process_info.pids:
-                    assert not psutil.pid_exists(pid)
+            # TODO: add the process monitor test in cluster mode
```
hkvision, line 57: for local mode, we have changed the implementation to directly call ray.init and thus there is no processmonitor.

PR #7766, opened 2023-03-05
Nano: TF customized training loop multi-process training

TheaperDeng:
### 1. Why the change?

Currently we only support TF multiprocess training for keras model with default training loop (i.e., `model.fit`). While many of our users design their own training loop as shown in this [tensorflow document](https://www.tensorflow.org/guide/keras/writing_a_training_loop_from_scratch). 

This PR add the support for this kind of models with customized training loop with a **2-lines code change** for users.

Proof points has been carried out on 2 models, and it does speed-up the training process.

### 2. User API changes

Users will have 2 decorators.
```python
from bigdl.nano.tf.keras import nano_multiprocessing
from bigdl.nano.tf.keras import nano
```

- `nano_multiprocessing`: users need to add this decorator to their `train_step` (typically an tf function), should be OK to use together with `nano_bf16` (need validation)
- `nano(num_processes=...)`: users need to add this decorator to their customized training loop (typically a loop calling `train_step`)

Here is an easy example
```python
from bigdl.nano.tf.keras import nano_multiprocessing, nano
import tensorflow as tf

tf.random.set_seed(0)

global_batch_size = 32
model = tf.keras.Sequential([tf.keras.layers.Dense(1, input_shape=(1,))])
optimizer = tf.keras.optimizers.SGD()

dataset = tf.data.Dataset.from_tensors(([1.], [1.])).repeat(128).batch(
    global_batch_size)

loss_object = tf.keras.losses.BinaryCrossentropy(from_logits=True)

# @nano_bf16  <-- Users could use bf16 and multiple processing together
@nano_multiprocessing  # <-- Just remove this line to run on 1 process
@tf.function
def train_step(inputs, model, loss_object, optimizer):
    features, labels = inputs

    with tf.GradientTape() as tape:
        predictions = model(features, training=True)
        loss = loss_object(labels, predictions)

    gradients = tape.gradient(loss, model.trainable_variables)
    optimizer.apply_gradients(zip(gradients, model.trainable_variables))
    return loss

@nano(num_processes=4)  # <-- Just remove this line to run on 1 process
def train_whole_data(model, dataset, loss_object, optimizer, train_step):
    for inputs in dataset:
        print(train_step(inputs, model, loss_object, optimizer))

if __name__ == "__main__":
    train_whole_data(model, dataset, loss_object, optimizer, train_step)
    print(model.predict(next(dataset.as_numpy_iterator())[0])[0])  # <-- single process and multiple process will get same result
```

### 3. Summary of the change 
nothing special is implemented. `backend` and `subprocess_worker` is reused as in default training loop model.

Users' args put to the customized training loop is serialized and changed (different types, like model, dataset, loss, optimizer..., in different way)

### 4. How to test?
- [ ] Unit test

please @plusbang help me to complete some critical TODOs

### DIFF
--- a/python/nano/src/bigdl/nano/tf/keras/__init__.py
+++ b/python/nano/src/bigdl/nano/tf/keras/__init__.py
@@ -19,4 +19,4 @@ import tensorflow as tf
 from .Sequential import Sequential
 from .Model import V2Model as Model
 from .inference.optimizer import InferenceOptimizer
-from .customized_training_utils import nano_bf16
+from .customized_training_utils import nano_bf16, nano_multiprocessing, nano
--- a/python/nano/src/bigdl/nano/tf/keras/customized_training_utils.py
+++ b/python/nano/src/bigdl/nano/tf/keras/customized_training_utils.py
@@ -50,7 +50,7 @@ class nano_multiprocessing(object):
         # TODO: to validate if we really could support kwargs
         per_replica_losses = mirrored_strategy.run(self.func, args=args, kwargs=kwargs)
         return mirrored_strategy.reduce(tf.distribute.ReduceOp.SUM, per_replica_losses,
-                                axis=None)
+                                        axis=None)
 
 
 def nano(num_processes):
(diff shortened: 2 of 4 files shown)


plusbang:
After some discussion, todo and issues are listed in https://github.com/analytics-zoo/Samsung-MSCRED/issues/5.

TheaperDeng (commented):
nothing from my side, requested some other stackholders' review

PR #7483, opened 2023-02-09
Nano: support jit int8 quantization for pytorch

Siritao:
## Description

Add `jit + int8` option to `InferenceOptimizer.quantize()`, which applies __FX Graph Mode Quantization__ and __JIT__ provided by pytorch.

### 1. Why the change?

Previously, Nano's int8 quantization supports accelerators like `onnxruntime` and `openvino`, or pure `ipex` / `inc` quantization. Since `TorchScript` is also a great accelerator to serialize and optimize pytorch models, we provide this approach to Nano users.

### 2. User API changes

New option in `InferenceOptimizer.quantize()` for pytorch:
```
int8_model = InferenceOptimizer.quantize(original_model, precision='int8', accelerator='jit',
                                         calib_data, q_config, input_sample, channels_last,
                                         thread_num, jit_strict, jit_method, enable_onednn)
```

### 3. Summary of the change 

original_model -> fx quantization (prepare, calibrate, convert) -> jit trace / script & freeze

### 4. How to test?

- [ ] Unit test

### Bug to be investigated
Note that enabling ONEDNN together with fx quantization leads to error on ResNet18 model, so we may disable it by default when using jit + int8.

PR #7771, opened 2023-03-06
Nano: fix the too long time in PytorchIPEXQuantizationModel for loop

ACupofAir:
## Description
Add a small change to make the  ipex_quantize model for loop has a limit times

### 1. Why the change?
* add sample_size property for calib_data
* change the for loop logic in ipex_quantize_model


python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -805,6 +805,10 @@ def quantize(model: nn.Module,
                                                                      calib_dataloader,
                                                                      input_sample)
 
+            # Add sample_size property to calib_dataloader to make the for loop in 
+            # `ipex_quantization_model.py` load sample items less than this value
+            calib_dataloader.sample_size = sample_size
```
TheaperDeng, line 810: should we make this an private member with some ugly name to avoid duplicated member like `_nano_calib_sample_size` for example?

TheaperDeng (commented):
wait for #7483 merging and update this logic to that PR

ACupofAir:
I'm sorry that I have some problem in rebase this branch to main. So I create a new pr #7804 
All the changes have move to #7804

PR #7785, opened 2023-03-07
support ndarray of tensors in torch_runner

songhappy:
## Description
Current torch_runner accepts a torch tensor as a feature if only one input is needed, a list or tuple of tensors is unpacked when a couple of inputs when needed.  
When users often have self-defined collate_fn, an array of tensors is needed.

### DIFF
--- a/python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
@@ -426,7 +426,7 @@ class TorchRunner(BaseRunner):
         # and restore batch to what it should be.
         features, target = batch
         if torch.is_tensor(features) or \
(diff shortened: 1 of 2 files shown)


python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
```diff
@@ -425,7 +425,8 @@ def _train_batch(self, batch, callbacks=None):
         # unpack features into list to support multiple inputs model
         # and restore batch to what it should be.
         features, target = batch
-        if torch.is_tensor(features):
+        if torch.is_tensor(features) or \
+                (isinstance(features, np.ndarray) and torch.is_tensor(features[0])):
```
leonardozcm, line 429: May we have a ut about this?

PR #7791, opened 2023-03-07
Orca: fix tf estimator ut random fail

hamham223:
### 1. Why the change?

Fix https://github.com/analytics-zoo/arda-docker/issues/729

Optimizer: SGD -> Adam

```python
            for i in range(len(pre_opt_weights)):
                assert np.array_equal(pre_opt_weights[i], after_opt_weights[i])
```

### 2. How to test?
- [x] Orca-Pytest-Ray-Py37-Spark3
- [x] Orca-Pytest-Ray-Py38-Spark3


hamham223:
@sgwhat Let's run the test again to see whether the random failure is eliminated?

hamham223:
Hooray! no piles of `nan` anymore

hamham223:
@hkvision take a look? Now `Orca-Pytest-Ray-Py37/38-Spark3` are the longest two tests, we might need to split them into two parts?

hkvision:
> @hkvision take a look? Now `Orca-Pytest-Ray-Py37/38-Spark3` are the longest two tests, we might need to split them into two parts?

One thing we can do is to put mxnet related tests into horovod?

python/orca/test/bigdl/orca/learn/spark/test_tf2_estimator.py
```diff
@@ -41,7 +41,7 @@ def compile_args(config):
     else:
         lr = 1e-3
     args = {
-        "optimizer": tf.keras.optimizers.SGD(lr),
+        "optimizer": tf.keras.optimizers.Adam(lr),
```
hkvision, line 44: SGD will possibly get gradient disappear and get nan weights; thus changed to adam

PR #7793, opened 2023-03-07
[PPML] Add pyspark tpch example

hzjane:
## Description
Add pyspark tpch example


ppml/trusted-big-data-ml/scala/docker-occlum/py-examples/tpch/main.py
```diff
@@ -0,0 +1,68 @@
+#
```
qiyuangong, line 1: Can we move TPCH into zoo-tutorial? In that case, both occlum and Gramine can use it by adding it into docker image?

hzjane: Good idea， i will do it.

qiyuangong (approved):
LGTM

PR #7796, opened 2023-03-07
[ORCA] Support Prediction Multi-output with Orca Estimator

sgwhat:
## Description
This pull request primarily adds support for a multi-output model with a Spark DataFrame input in both tf2 and pytorch estimator. Previously, only single-output models were supported.

### 1. Why the change?
As described at https://github.com/intel-analytics/BigDL/issues/7348, the orca estimator currently does not support a multi-output inference pipeline.

### 2. User API changes
We have added a new parameter, output_cols, for users to provide output column information.

```python
from bigdl.orca.learn.tf2 import Estimator

def model_creator(config):
    # define a multi-output model
    return model

estimator = Estimator.from_keras(model_creator=model_creator, backend="spark")
estimator.predict(df, feature_cols=["input"], output_cols=["tf_output_1", "tf2_output_2"])
```

- Single-dimensional prediction output:

| index | input_1 |  tf_output_1 | tf_output_2 |
| --- | --- | --- | --- |
| 0 | [[[0.535554111003... | [0.34790271520614... | [0.02838393114507... |
| 1 | [[[0.880806922912... | [0.29939782619476... | [-0.0265996530652... |

- Multi-dimensional prediction output:

| index | input_1 | input_2 | tf_output_1 | tf_output_2 |
| --- | --- | --- | --- | --- |
| 0 | [[[0.022802922874... | [[[0.198862820863... | [[[-0.76845872402... | [[[-0.18314521014... |
| 1 | [[[0.502752780914... | [[[0.679012894630... | [[[-0.35531336069... | [[[-0.26589363813... |

### 3. Summary of the change 
1. Added `output_cols` to support multi-output models.
2. Reconstructed the Spark DataFrame as the return result with the provided columns.

### 4. How to test?
- [x] Unit test
- [x] Application test

### 5. To Do
- [x] Align dict format implementation with pytorch estimator (processing by @leonardozcm )
- [ ] Automatically allocate output columns name when users do not specify it (next pr)
- [ ] Support output with different lengths (next pr)
- [x] Support multi-dimension with each output label (like 3d)


python/orca/src/bigdl/orca/learn/utils.py
```diff
@@ -203,11 +219,17 @@ def combine(pair):
                 structType = FloatType()
                 for _ in range(dim):
                     structType = ArrayType(structType)
-                row = Row(*([pair[0][col] for col in pair[0].__fields__] + [pair[1].tolist()]))
+                row_values = [pair[0][col] for col in pair[0].__fields__]
+                for value in pair[1]:
+                    row_values.append(value.tolist())
```
leonardozcm, line 224: This may conflict with the case that ndarray is a 2 or more dimensional array?

sgwhat:
Currently, we still need users to specify the output labels (aka output df columns' name), we plan to achieve this functionality in the next pr, which could avoid this pr being too complex to review.

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
```diff
@@ -420,6 +420,7 @@ def predict(self,
                 data: Union['SparkXShards', 'SparkDataFrame'],
                 batch_size: int=32,
                 feature_cols: Optional[List[str]]=None,
+                output_cols: Optional[List[str]]=None,
```
hkvision, line 423: So in this case, can we support the previous single output column to specify the column name, i.e. instead of "prediction "(default), user can specify the output_col

sgwhat: Seems it's not supported currently? @leonardozcm

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
```diff
@@ -475,7 +476,9 @@ def predict(self,
                                               shard_size=batch_size)
 
             pred_shards = self._predict_spark_xshards(xshards, init_params, params)
-            result = convert_predict_xshards_to_dataframe(data, pred_shards)
+            result = convert_predict_xshards_to_dataframe(data,
```
hkvision, line 479: what about SparkXShards? we also need to support that

leonardozcm: We have supported this in previous pr https://github.com/intel-analytics/BigDL/pull/7704/files#diff-ce59ed5b2127ca719be83567a5d7443351f2491296219720aade783069860748R155-R179

python/orca/src/bigdl/orca/learn/pytorch/utils.py
```diff
@@ -353,6 +353,22 @@ def index_concatenate(x, axis=0):
         return res
 
 
+def split_predict_cols(y):
+    if not isinstance(y, (list, tuple)):
+        return {"prediction": y}
+    output_cols = [f"output_{i}" for i in range(len(y))]
+    if isinstance(y, np.ndarray):
+        # Single output
+        y = [y]
```
hkvision, line 362: the logic here seems strange. For the single output case, it is already considered in Line358 right?

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
leonardozcm: Our xshards are still working the same as before, and we will plan to support this in the future.

python/orca/src/bigdl/orca/learn/pytorch/utils.py
leonardozcm: Yes, this is not working and should be aborted.

python/orca/src/bigdl/orca/learn/utils.py
```diff
@@ -208,11 +210,25 @@ def convert_elem(elem):
                 return elem.tolist()
 
     def combine(pair):
-        return Row(*([pair[0][col] for col in pair[0].__fields__] +
-                     [convert_elem(pair[1])]))
+        if not output_cols:
+            if isinstance(pair[1], (list, tuple)) and len(pair[1]) == 1:
```
hkvision, line 214: what does this indicate? A singleton list?

python/orca/src/bigdl/orca/learn/utils.py
```diff
@@ -208,11 +210,25 @@ def convert_elem(elem):
...
+                return Row(*([pair[0][col] for col in pair[0].__fields__] +
+                             convert_elem(pair[1])))
```
sgwhat, line 216: row should be <[input_1], [input_2], [DenseVector]>.

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
leonardozcm: For now our output_col logic is like:
1. If single output, user **does not and should not** pass output_col to.predict()
2. if multi-output, user should pass a list of output_col to predict(), and the number of output_cols is exactly corresponding to the predict columns.

hkvision: For multi-output, output_cols is always required right? where do you add this assert?

python/orca/src/bigdl/orca/learn/utils.py
hkvision: shouldn't it be Row(x1, x2, output)? * will do the unpack for lists.

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
hkvision: If so, the interface would be better if single output can also specify the output_col

python/orca/src/bigdl/orca/learn/utils.py
leonardozcm: Yes, if users do not specified a output_col name, all output in pair[1] should be stacked into one column.
Note that pair[1] is always a list, in which case:
1. if pair[1] is a singleton list, we should return [a]
2. if pair[1] is not a singleton list, we should stack them as [[a, b, c]]

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
sgwhat: yes it's need to add an assert, and we plan to add a func (in next pr) to allocate output_cols automatically in case users don't specify them.

python/orca/src/bigdl/orca/learn/utils.py
sgwhat: that depends o, if pair[1] is ndarray, output (Dense Vector) is fine, if pair[1] is a list of ndarray, we should convert it to [Dense Vector].

hkvision: so why should we discuss the case for a singleton list?

sgwhat:
Will separate torch and tf2 implementation.

python/orca/src/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py
```diff
@@ -420,6 +420,7 @@ def predict(self,
                 data: Union['SparkXShards', 'SparkDataFrame'],
                 batch_size: int=32,
                 feature_cols: Optional[List[str]]=None,
```
hkvision, line 422: Add the code doc for output_cols

python/orca/src/bigdl/orca/learn/tf2/pyspark_estimator.py
```diff
@@ -398,6 +399,9 @@ def predict(self,
         :param data_config: An optional dictionary that can be passed to data creator function.
         :param feature_cols: Feature column name(s) of data. Only used when data is a Spark
                DataFrame or an XShards of Pandas DataFrame. Default: None.
+        :param output_cols: Column name(s) of the model output data. Only used when data is
+               a Spark DataFrame, note the order of column name(s) should be consistent with the
+               model output data. Default: None.
```
hkvision, line 404: Need to add description of the default behavior. Maybe in a next PR after we finish this.

sgwhat: sure.

python/orca/test/bigdl/orca/learn/test_tf2_basic.py
```diff
@@ -354,6 +371,54 @@ def test_checkpoint_model(self):
...
+            model_creator=model_creator,
+            verbose=True,
+            config={"lr": 0.2},
+            workers_per_node=2,
+            backend="spark")
+
+        pred_res = estimator.predict(df,
+                                     feature_cols=["input_1", "input_2"],
+                                     output_cols=["score_output", "class_output"])
+        pred_res.collect()
```
hkvision, line 415: need to check the output column names for pred_res are as expected.

hkvision: also may need to check the output column data, each is an ndarray?

leonardozcm (approved):

PR #7805, opened 2023-03-08
torch object detection tutorial

songhappy:
## Description

add a notebook for pytorch object detection


leonardozcm:
@hkvision hi, would you mind taking a look at this?

jenniew:
I think you may add prediction part to let user see the training result.

jenniew (commented):
I think you may add prediction part to let user see the training result.

jenniew (approved):
LGTM

PR #7807, opened 2023-03-08
Nano: quick fix the calculation of output shape and latency

MeouSker77:
## Description

quick fix the calculation of output shape and latency


rnwang04 (approved):
LGTM

PR #7809, opened 2023-03-08
[PPML]check if sgx is available

ShanSimu:
## Description

check if sgx is available

### DIFF
--- a/ppml/services/check-service.sh
+++ b/ppml/services/check-service.sh
@@ -1,6 +1,34 @@
 #!/bin/bash
 # Acceptable arguments: attestation, bigdl-kms, ehsm, kms-utils, pccs, all
 
+sgx() {
+    echo "Detecting SGX..."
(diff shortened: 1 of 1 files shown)


ppml/services/check-service.sh
```diff
@@ -1,6 +1,34 @@
...
+sgx() {
+    echo "Detecting SGX..."
+    if [ -e "/dev/sgx" ] && [ -e "/dev/sgx_enclave" ] && [ -e "/dev/sgx_provision" ]; then
+        echo "SGX driver is installed."
+    else
+        echo "SGX driver isn't installed."
+    fi
+
+    echo "Detecting aesmd..."
+    if [ -e "/var/run/aesmd/aesm.socket" ]; then
```
Uxito-Ada, line 13: Maybe check status of aesmd system service and grep whether `Active: active (running)` is better than brutally `ls`

ShanSimu: good idea

Uxito-Ada (approved):
Lgbm

PR #7822, opened 2023-03-09
Orca: Make training by step a maxstepcallback

leonardozcm:
## Description

In this pr we integrate our support of `maxstep` training in a callback, just like what we do in Tqdmcallback.
We shall take it carefully to add any new arg in torchrunner, for they are really hard to be maintained.


jenniew (approved):
LGTM

PR #7827, opened 2023-03-09
Limit sgx memory usage of torchserve frontend to 8g.

Hanyu-Jin:
## Description

Limit sgx memory usage of torchserve frontend to 8g.

### 1. Why the change?

Limit sgx memory usage of torchserve processes to 8g and reduce total sgx memory.

### 2. User API changes

None.

### 3. Summary of the change 

Add serveral JVM settings to frontend.


glorysdj (approved):
LGTM

PR #7833, opened 2023-03-10
Nano: fix tf bf16 conversion for functional API model with Conv2DTranspose layer

MeouSker77:
## Description

Fix tf bf16 conversion for functional API model with Conv2DTranspose layer. i.e. issue 1 in <https://github.com/analytics-zoo/nano/issues/318>

### 1. Why the change?

Model with Conv2DTranspose layer cannot be saved after modifying its dtype policy,
and model with custom object cannot be simply loaded by current code

### 2. User API changes

N/A

### 3. Summary of the change 

If save or load failed, use `from_config` to create bf16 model.

### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/tf/keras/amp/bfloat16.py
+++ b/python/nano/src/bigdl/nano/tf/keras/amp/bfloat16.py
@@ -33,6 +33,7 @@ def BF16Model(model):
         # save or load operation may fail
         with TemporaryDirectory() as temp_dir:
             model.save(temp_dir)
+            # todo: fix loading custom object
             bf16_model = tf.keras.models.load_model(temp_dir)
     except Exception as _e:
         # if a functional model failed to save or load, try `from_config`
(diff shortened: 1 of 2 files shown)


yangw1234:
Can we add a argument that allow user to pass their "custom object"?

MeouSker77:
> Can we add a argument that allow user to pass their "custom object"?

Yes, will add this argument in next PR

Oscilloscope98 (approved):
LGTM

python/nano/src/bigdl/nano/tf/keras/amp/bfloat16.py
```diff
@@ -13,25 +13,42 @@
...
+        # save or load operation may fail
         with TemporaryDirectory() as temp_dir:
             model.save(temp_dir)
+            # todo: fix loading custom object
             bf16_model = tf.keras.models.load_model(temp_dir)
+    except Exception as _e:
+        # if a functional model failed to save or load, try `from_config`
+        if hasattr(model, "input_shape"):
+            config = model.get_config()
+            bf16_model = tf.keras.Model.from_config(config)
```
rnwang04, line 42: why `tf.keras.Model.from_config(config)` will return a bf16 model?

rnwang04 (approved):
LGTM

PR #7839, opened 2023-03-13
Nano: fix thread num setting of openvino 2022.3

MeouSker77:
## Description

Fix thread num setting of openvino 2022.3, 2022.3 use `INFERENCE_NUM_THREADS` to set thread num, not `CPU_THREADS_NUM`

<https://docs.openvino.ai/2022.3/ovms_docs_performance_tuning.html?highlight=cpu_threads_num#scalability>


rnwang04 (approved):
LGTM

PR #7840, opened 2023-03-13
[Nano] Bug fix when enabling channels last on variable length input models

hoshibara:
## Description

[Enabling channels last on variable length input models may cause forward to fail](https://github.com/analytics-zoo/nano/issues/319)


rnwang04 (approved):
LGTM

PR #7849, opened 2023-03-13
[Nano] Concat rest part of inputs when inputs larger than training data

hoshibara:
## Description

Supplementary modification of #7840 . 
It will concat rest part of input when the length of input larger than training data.


rnwang04:
LGTM

Issue #7779, opened 2023-03-06
Orca: ut `test_string_input` failed in tf 2.9.0


hamham223:
As title, the following ut failed in tf 2.9.0. Seems the failure is not due to experimental api but tf versions.

```python
    def test_string_input(self):

        def model_creator(config):
            import tensorflow as tf
            vectorize_layer = tf.keras.layers.experimental.preprocessing.TextVectorization(
                max_tokens=10, output_mode='int', output_sequence_length=4)
            model = tf.keras.models.Sequential()
            model.add(tf.keras.Input(shape=(1,), dtype=tf.string))
            model.add(vectorize_layer)
            return model

        from bigdl.orca import OrcaContext
        from pyspark.sql.types import StructType, StructField, StringType
        spark = OrcaContext.get_spark_session()
        schema = StructType([StructField("input", StringType(), True)])
        input_data = [["foo qux bar"], ["qux baz"]]
        input_df = spark.createDataFrame(input_data, schema)
        estimator = Estimator.from_keras(model_creator=model_creator)
        output_df = estimator.predict(input_df, batch_size=1, feature_cols=["input"])
        output = output_df.collect()
        print(output)
```

+ code: https://github.com/intel-analytics/BigDL/blob/main/python/orca/test/bigdl/orca/learn/ray/horovod/test_tf2_horovod_backend.py#L952
+ ut failed in tf29: https://github.com/intel-analytics/BigDL/actions/runs/4340973173/jobs/7580060537#step:6:10493
+ ut passed in tf23: https://github.com/intel-analytics/BigDL/actions/runs/4321730334/jobs/7576715426#step:6:7325
+ api change: `tf.keras.layers.experimental.preprocessing.TextVectorization` -> `tf.keras.layers.TextVectorization`
+ reference: https://www.tensorflow.org/api_docs/python/tf/keras/layers/TextVectorization

PR #7850, opened 2023-03-13
Fix test string input for tf 2.9.0

hkvision:
Fix https://github.com/intel-analytics/BigDL/issues/7779

Seems for later tf versions, TextVectorization layers must be given a vocabulary or be adapted. (previously tf 2.3.0 doesn't have such constraint)
https://www.tensorflow.org/api_docs/python/tf/keras/layers/TextVectorization

The vocabulary for the layer must be either supplied on construction or learned via adapt().


hkvision:
https://github.com/intel-analytics/BigDL/actions/runs/4402905561/jobs/7714378744 py38 test passed.
py37 test failed several times due to the environment, raised an issue: https://github.com/analytics-zoo/arda-docker/issues/796

PR #7853, opened 2023-03-14
Fix Nightly-test "ExampleTest-Ray-Ctx-Py37-Spark3"

hamham223:
### 1. Why the change?

Fix https://github.com/intel-analytics/BigDL/actions/runs/4406399922/jobs/7728776700

### 2. How to test?
Nightly Test:
- [x] Orca-ExampleTest-Ray-Ctx-Py37-Spark3


hamham223:
This example test does not have `setup conda env` and `remove env before, and thus lots of python packages are left on the runner...

PR #7856, opened 2023-03-14
Nano: quick fix torch optimizer

MeouSker77:
## Description

quick fix torch optimizer: If `input_sample` is some custom class which is not Dict or Tensor, we should pass it directly


rnwang04 (approved):
LGTM

PR #7859, opened 2023-03-14
Chronos: fix nightly build test

plusbang:
## Description

Although ut fail, the nb test can still pass. To solve this problem, add `exit_status_0` in test script.
MicrosoftTeams-image (7)

And fix https://github.com/intel-analytics/BigDL/actions/runs/4417465437/jobs/7753468945#step:6:1457


liu-shaojun (approved):
LGTM

PR #7862, opened 2023-03-14
Nano: fix tf load model with custom objects

MeouSker77:
## Description

fix tf load model with custom objects


rnwang04 (approved):
LGTM, but it would be better to have a related ut or example.

PR #7874, opened 2023-03-15
Chronos: update `ThresholdDetector` and anomaly detect example

plusbang:
## Description

Related to reference use case, update `ThresholdDetector` and anomaly detect example.

### 2. User API changes

change `anomaly_indexes` of `ThresholdDetector` to support obtaining anomaly indexes of new time series.
new api:
```python
 def anomaly_indexes(self, y=None, y_pred=None):
# Gets the indexes of the anomalies from new time series
```

### 3. Summary of the change 

- more function of `ThresholdDetector`
- new anomaly detect example with TCNForecaster
- update readme

### 4. How to test?
- [x] Unit test


TheaperDeng (approved):
The notebook looks great. We may want to add CI/CD test script for this one in later PR.

plusbang:
> The notebook looks great. We may want to add CI/CD test script for this one in later PR.

yeah, we temporarily skip this test before (https://github.com/intel-analytics/BigDL/blob/main/python/chronos/dev/app/run-app-tests-pip.sh#L52). We can restore it later.

Issue #7481, opened 2023-02-09
version.txt for Python is not changed


hkvision:
https://github.com/intel-analytics/BigDL/blob/main/python/version.txt
Why this is not changed?
Also for branch2.2 it is unchanged.

For the release job, we should change this file instead of manually input the version to the script. @pinggao18

Le-Zheng:
Shall we remove this file since it seems not be use anywhere?

hkvision:
I don't think we can remove it.
We do need to have some place in the code to identify the version, instead of always manually input it in our Jenkins jobs.

hkvision:
@hamham223 Please help fix this.

hkvision:
Also work with @Le-Zheng and @liu-shaojun to make sure when we release, we use the version in version.txt instead of manually input the version.

PR #7881, opened 2023-03-15
CI/CD: update `version.txt`

hamham223:
### 1. Why the change?

https://github.com/intel-analytics/BigDL/issues/7481

update `version.txt` to `2.3.0.dev0`


.github/workflows/nightly_build.yml
```diff
@@ -296,7 +296,7 @@ jobs:
     - name: Build package
       run: |
         export TIMESTAMP=`date '+%Y%m%d'`
-        export PYPI_VERSION=2.3.0
+        export PYPI_VERSION=$(cat ./python/version.txt | head -1)
```
hkvision, line 299: need to modify the release job instead of the nightly job.

.github/workflows/nightly_build.yml
```diff
@@ -346,7 +346,7 @@ jobs:
...
```
hkvision, line 349: this cat will get 2.3.0.dev0 right? this is wrong, for nightly build the version should be 2.3.0b20230317
We can keep this in nightly build, only make sure we use version.txt when we release

hamham223: ok, change `version.txt` in to `2.3.0`, restore changes in `nightly_build.yml`

hkvision: yes, change back and we can manually modify this?

hkvision: no, version.txt should always be the correct version, in main it should be dev0

hkvision: just change this file back to 2.3.0 and we can manually modify this

.github/workflows/bigdl-core-release-pypi.yml
```diff
@@ -5,11 +5,11 @@ on:
...
-        description: 'pypi version (e.g. 2.1.0, 2.2.0b20220927)'
-        required: true
-        default: '2.2.0b20221205'
-        type: string
+      # version:
+      #   description: 'pypi version (e.g. 2.1.0, 2.2.0b20220927)'
+      #   required: true
+      #   default: '2.2.0b20221205'
+      #   type: string
       branch:
```
hkvision, line 13: just delete them?

hamham223:
@Le-Zheng Could you plz take a look?

Le-Zheng:
LGTM

hkvision (approved):

PR #7887, opened 2023-03-16
Nano: support Mapping input_sample in torch optimizer

MeouSker77:
## Description

Support Mapping input_sample in torch optimizer

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
+++ b/python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
@@ -19,7 +19,7 @@ from torch import nn
 import time
 import sigfig
 import multiprocessing as mp
-from typing import Dict, Callable, Tuple, Optional, List, Union, Sequence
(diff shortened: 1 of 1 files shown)


python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -410,13 +410,18 @@ def optimize(self, model: nn.Module,
                     val_dataset = RepeatDataset(sample=(val_sample, val_label), num=1)
                     validation_data = DataLoader(val_dataset, batch_size=1)
                     validation_data = remove_batch_dim_fn(validation_data)
-
+        # jit cannot handle `Mapping`, so we convert it to `dict`
+        if isinstance(input_sample, (Mapping, MappingView)):
```
hoshibara, line 414: https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes
Maybe MappingView cannot convert to dict

hoshibara (approved):

PR #7888, opened 2023-03-16
[PPML] Change dl-serving image README position

gc-fu:
## Description

Change dl-serving image README position


Hanyu-Jin (approved):
LGTM

PR #7890, opened 2023-03-16
Orca[test]: NCF tutorial on ray backend

hamham223:
## Description

as title, test the running time of two backends.


hamham223:
env is fixed, job: https://github.com/intel-analytics/BigDL/actions/runs/4432873082/jobs/7779989904

PR #7892, opened 2023-03-16
[PPML]fix: put MALLOC_ARENA_MAX into sgx template

ShanSimu:
## Description

put MALLOC_ARENA_MAX into sgx template


xiangyuT (approved):
LGTM

PR #7898, opened 2023-03-17
Chronos: add anomaly detection example to docker and test script

plusbang:
## Description

Related to reference use case, add anomaly detection example to docker and CI/CD test script.

### 4. How to test?
- [x] Unit test


TheaperDeng (approved):
LGTM

PR #7912, opened 2023-03-21
[PPML]delete MALLOC_ARENA_MAX  in driver.yaml

ShanSimu:
## Description

delete MALLOC_ARENA_MAX  in driver.yaml because it will cause unknown memory errors


hzjane (approved):
LGTM

PR #7915, opened 2023-03-21
change default hadoop user name from root to os.env[HADOOP_USER_NAME] or current user

### LINKED ISSUE #7905: default to user 'root' when init_nncontext with yarn-cluster and yarn-local modes
I just noticed that bigdl seems to default to 'root' when init_nncontext is called with yarn-cluster and yarn-local modes. That leads to be permission errors if the actual user is not root. This does not happen when using 'local' or 'spark-submit' for init_nncontext(). 

Along with the same context, what is the proper configuration place to change application settings, such as the username, the working path, the number of executors, executor memory, driver memory, and the number of slaves used in Spark/Yarn.

qiuxin2012:
## Description

change default hadoop user name in init_nncontext_on_yarn

### 1. Why the change?

fixed #7905 

### 2. User API changes

None

### 3. Summary of the change 

change default hadoop user name in init_nncontext_on_yarn

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/python/dllib/src/bigdl/dllib/nncontext.py
+++ b/python/dllib/src/bigdl/dllib/nncontext.py
@@ -65,7 +65,7 @@ def init_spark_on_yarn(hadoop_conf,
                        extra_python_lib=None,
                        penv_archive=None,
                        additional_archive=None,
-                       hadoop_user_name="root",
+                       hadoop_user_name=None,
                        spark_yarn_archive=None,
                        spark_log_level="WARN",
                        redirect_spark_log=True,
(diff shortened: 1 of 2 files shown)


python/dllib/src/bigdl/dllib/nncontext.py
```diff
@@ -93,7 +93,8 @@ def init_spark_on_yarn(hadoop_conf,
            Default to be None.
     :param additional_archive: Comma-separated list of additional archives to be uploaded and
            unpacked on executors. Default to be None.
-    :param hadoop_user_name: The user name for running the yarn cluster. Default to be 'root'.
+    :param hadoop_user_name: The user name for running the yarn cluster. Default to be None.
+           None means os.environ["HADOOP_USER_NAME"], or current user if HADOOP_USER_NAME is unset.
```
hkvision, line 97: Change to `Default to be None, which will use the system configured user.`?

hkvision (approved):

PR #7923, opened 2023-03-22
[PPML] Update LgbmClassifierTrain.scala

hzjane:
## Description
Update LgbmClassifierTrain.scala to enable distributed training.

### DIFF
--- a/scala/dllib/src/main/scala/com/intel/analytics/bigdl/dllib/example/nnframes/lightGBM/LgbmClassifierTrain.scala
+++ b/scala/dllib/src/main/scala/com/intel/analytics/bigdl/dllib/example/nnframes/lightGBM/LgbmClassifierTrain.scala
@@ -40,7 +40,7 @@ object LgbmClassifierTrain {
         StructField("petal length", DoubleType, true),
         StructField("petal width", DoubleType, true),
         StructField("class", StringType, true)))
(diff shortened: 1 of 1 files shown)


scala/dllib/src/main/scala/com/intel/analytics/bigdl/dllib/example/nnframes/lightGBM/LgbmClassifierTrain.scala
```diff
@@ -40,7 +40,7 @@ object LgbmClassifierTrain {
         StructField("petal length", DoubleType, true),
         StructField("petal width", DoubleType, true),
         StructField("class", StringType, true)))
-      val df = spark.read.schema(schema).csv(params.inputPath)
+      val df = spark.read.schema(schema).csv(params.inputPath).repartition(sc.defaultParallelism)
```
Uxito-Ada, line 43: Since not every user wants to enable parallel training, pls set this to a configurable option
And as for the partition number, maybe the count of executors is better than default parallelism

Uxito-Ada:
spark.default.parallelism defaults to the number of all cores on all machines, while on K8S CPU cores are virtual and elastic resources and its count varies. The number of executors is better as the number of pods are fixed.

PR #7935, opened 2023-03-23
[ORCA] Enhance functionality to automatically allocate 'output_cols' for supporting multi-output predictions

sgwhat:
## Description
To support the automatic generation of output column names when creating a Spark DataFrame as the prediction result.

### 1. Why the change?
This improvement aims to assist users who may not be familiar with the output labels of a model. In the previous implementation of PR #7796, users were required to specify the `output_cols` to create a Spark DataFrame as the output. 

Now, we have enhanced the functionality of the Orca Estimator so that users no longer need to specify `output_cols`. Orca Estimator will automatically parse the model's outputs and generate output column names for the Spark DataFrame.

### 2. User API changes
We have improved the functionality of PR #7796 to aid users who may not fully comprehend their model by eliminating the requirement to specify the `output_cols `parameter. In situations where `output_cols` are not defined in the `estimator.predict()` function, the output will be generated according to the following description:

```python
from bigdl.orca.learn.tf2 import Estimator

def model_creator(config):
    # define a multi-output model
    # outputs=["class_output", "score_output"]
    return model

estimator = Estimator.from_keras(
                 model_creator=model_creator,
                 backend="spark")
estimator.predict(df, feature_cols=["input"])
```

- Prediction output (with automatically generate output labels):

| index | input_1 | input_2 |  tf_output_1 | tf_output_2 |
| --- | --- | --- | --- | --- |
| 0 | [[[0.139809682965... | [[[0.224684208631... | [[[-0.31654295325... | [[[1.004208207130... |
| 1 | [[[0.588949799537... | [[[0.614766180515... | [[[0.011196002364... | [[[1.346148848533... |

### 3. Summary of the change 

- The `output_cols` parameter has been removed from `convert_predict_rdd_to_dataframe`, as it was only intended as a temporary option.
- When using the TF2 estimator, output labels will be automatically generated as `tf_output_x`. For the PyTorch estimator, output labels will be generated as `torch_output_x`.
- In the previous implementation, the multi-output results of the model would be compressed under the `prediction` column label. This PR has addressed this issue.

### 4. How to test?
- [x] Unit test
- [ ] Application test


leonardozcm (approved):
Need to consider the case that torch model's output is a tuple of tensors, which indicates it is a batch instead of multi-output.
like https://pytorch.org/tutorials/intermediate/torchvision_tutorial.html

PR #7939, opened 2023-03-23
[PPML]fix: change the setting of MALLOC_ARENA_MAX in yaml

ShanSimu:
## Description

change the setting of MALLOC_ARENA_MAX in yaml because the former way is wrong. yaml file can't read environment value directly.


Uxito-Ada (approved):
lgtm

PR #7940, opened 2023-03-23
Chronos: support context manager in AutoformerForecaster

plusbang:
## Description

In previous work, we have supported context manager in `BaseForecaster`.
According to some benchmark experiments, with context manager, inference latency of `AutoformerForecaster` can reduced 5%-10%.
In this pr, also support context manager in `AutoformerForecaster`.

### 1. Why the change?

Similar to https://github.com/intel-analytics/BigDL/pull/7279.

### 2. User API changes

Provide a new `get_context` method for forecasters to implement thread control and inference_mode.
```python
with forecaster.get_context(thread_num=xxx):
    for _ in range(n):
        forecaster.predict(data)
```
It's recommend to use above context manager in inference loop.

But, if users don't use context manager, the following loop still works. (first call is much slower)
```python
for _ in range(n):
    forecaster.predict(data)
```

### 3. Summary of the change 

- `get_context` method
- In `predict`, we will generate `ForecasterContextManager` if users don't use context manager; otherwise, just use `DummyForecasterContextManager`

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/chronos/src/bigdl/chronos/forecaster/autoformer_forecaster.py
+++ b/python/chronos/src/bigdl/chronos/forecaster/autoformer_forecaster.py
@@ -30,6 +30,8 @@ from bigdl.chronos.data import TSDataset
 from bigdl.nano.automl.hpo.space import Space
 from bigdl.chronos.forecaster.utils_hpo import GenericTSTransformerLightningModule, \
     _config_has_search_space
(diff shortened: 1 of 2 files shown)


TheaperDeng (approved):
LGTM

PR #7948, opened 2023-03-24
Orca: [test] rm conf `spark.python.worker.reuse` (spark backend ut)

hamham223:
### 1. Why the change?

This is a test, see what can happen..


python/orca/src/bigdl/orca/learn/tf2/tf_runner.py
```diff
@@ -59,7 +59,7 @@ def find_free_port():
 def _try_import_strategy():
     """Late import for TensorFlow"""
     import tensorflow as tf
-    return tf.distribute.experimental.MultiWorkerMirroredStrategy
+    return tf.distribute.MultiWorkerMirroredStrategy
```
hamham223, line 62: `tf.distribute.experimental.MultiWorkerMirroredStrategy` this is now deprecated, remove the `experimental`

PR #7958, opened 2023-03-29
[Nano] support keyword argument for `PytorchIPEXJITModel` model

rnwang04:
## Description

Support keyword argument for `PytorchIPEXJITModel` and add related UT.

### How to test?
- [x] Unit test


TheaperDeng (approved):
LGTM, we may examine all of our models in future to relax this keyword argument for available ones

PR #7969, opened 2023-03-31
[NANO] Add safe path check in multi-process worker

ACupofAir:
## Description

* To prevent users from entering illegal paths.
* Add path available check before load model when multi-process invoking worker .


TheaperDeng:
Don't merge this PR until safechecks are passed

TheaperDeng:
close this PR since a similar one has been merged

PR #7971, opened 2023-03-31
Nano: further support tensorflow customized training loop with multi-process

plusbang:
## Description

Complete remaining TODOs after https://github.com/intel-analytics/BigDL/pull/7766.

### 1. Why the change?

In previous work (https://github.com/intel-analytics/BigDL/pull/7766), we start to support tensorflow multi-process training with customized training loop.
In fact, some critical questions still need to solve and validate.

Proof points has been carried out on model of special use-case, and it does solve remaining problems and speed-up the training process.

### 2. User API changes

Users will have a decorator to handle customized loss function
```bash
from bigdl.nano.tf.keras import nano_multiprocessing_loss

@nano_multiprocessing_loss(**kwargs)
def loss_object(x, pred, **kwargs):
    xxxx
```
nano_multiprocessing_loss: users need to add this decorator to their loss function.

### 3. Summary of the change 

- add a new decorator @nano_multiprocessing_loss to handle customized loss function
- support thread settings for core affinity for stock tensorflow
- fix loss computation and other operations within `MultiWorkerMirroredStrategy`
- support dataset created by `from_generator`

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/tf/keras/customized_training_utils.py
+++ b/python/nano/src/bigdl/nano/tf/keras/customized_training_utils.py
@@ -108,12 +108,12 @@ class _Nano_Customized_Training(object):
                 if isinstance(arg, tf.data.Dataset):
                     warnings.warn('If dataset is created by `from_generator`, please initiate '
                                   '`_GeneratorState` of the dataset first, that is '
-                                  'dataset._GeneratorState = dataset._GeneratorState(generator).' 
+                                  'dataset._GeneratorState = dataset._GeneratorState(generator). '
                                   'Otherwise, there exists errors because of a known limitation: '
                                   'https://www.tensorflow.org/api_docs/python/tf/data/Dataset'
                                   '#from_generator')
                     if hasattr(arg, '_GeneratorState') and \
-                        hasattr(arg._GeneratorState, '_generator'):
+                       hasattr(arg._GeneratorState, '_generator'):
                         # support dataset created by `from_generator`
                         train_ds_gen = arg._GeneratorState._generator
                         train_ds_signature = arg.element_spec
@@ -167,6 +167,7 @@ class _Nano_Customized_Training(object):
 
         return histrories[0]
 
+
 def _train_func(target_path, *args):
     mirrored_strategy = tf.distribute.MultiWorkerMirroredStrategy()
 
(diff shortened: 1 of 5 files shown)


python/nano/src/bigdl/nano/tf/keras/customized_training_utils.py
```diff
@@ -169,46 +185,79 @@ def _train_func(target_path, *args):
...
+            # deserialize others
+            if arg[0] == "others":
+                with open(arg[1], 'rb') as f:
+                    actrual_args[i] = cloudpickle.load(f)
+                    if callable(actrual_args[i]) and \
+                       isinstance(actrual_args[i], nano_multiprocessing):
+                        actrual_args[i] = partial(actrual_args[i],
+                                                  mirrored_strategy=mirrored_strategy)
+
+        with open(target_path, 'rb') as f:
```
TheaperDeng, line 225: @ACupofAir please have a look at this PR and add some security related change after its merging

python/nano/src/bigdl/nano/utils/tf/backend.py
```diff
@@ -59,8 +59,13 @@ def run_subprocess(self, target, args=..., nprocs=1, envs=None) -> Any:
...
                     if key not in envs[i]:
                         envs[i][key] = val
+                if 'OMP_NUM_THREADS' in envs[i]:
+                    set_op_thread = envs[i]['OMP_NUM_THREADS']
+                else:
+                    # 0 means the system picks an appropriate number
+                    set_op_thread = '0'
                 ex_list.append(subprocess.Popen([sys.executable, f"{cwd_path}/subprocess_worker.py",
-                                                 temp_dir], env=envs[i]))
+                                                 temp_dir, set_op_thread], env=envs[i]))
```
TheaperDeng, line 68: @MeouSker77 please have a look at the changes in `backend.py` and `subprocess_worker.py`. It looks good to me but I may miss something.

TheaperDeng (approved):
LGTM

python/nano/src/bigdl/nano/utils/tf/backend.py
TheaperDeng: Note: This change is for stock tensorflow since they are not built with openmp and may have some thread control issue by using OMP_NUM_THREADS

PR #7972, opened 2023-03-31
Chronos: quick fix `Evaluator.plot`

plusbang:
## Description

When use `Evaluator.plot` to plot more than 10 images, ValueError exists.
In this pr, fix this bug and modify related ut.

### 4. How to test?
- [x] Unit test


TheaperDeng (approved):
LGTM merge after action passed

PR #7975, opened 2023-04-03
[PPML] Fix trusted-deep-learning base image

gc-fu:
## Description

The previous build environment is a mess.  This pr will fix that issue

### 1. Why the change?

To provide clean build environment.

The PyTorch version is also upgraded to `1.13.1` so that it is compatible with `bigdl-nano` nightly-build.


Hanyu-Jin (approved):
LGTM

PR #7976, opened 2023-04-03
[ORCA] Fix for TF2 Pyspark Estimator Getting Stuck on Kubernetes

sgwhat:
## Description
Fix issue with the description in https://github.com/analytics-zoo/arda-docker/issues/808.

### 1. Why the change?
Change the method to get current worker ip, which could support getting k8s pods' addresses instead of only reaching nodes' addresses in cluster.

### 2. User API changes
N/A

### 3. Summary of the change 
Re-write the part to get workers' addressed to init Grpc group.
1. Divide worker address into `worker address` and `executor address`.
2. `executor address` is mainly used for getting the global rank.
3. `worker address` is mainly used for allocating pod or node's ip to grpc group. 

### 4. How to test?
- [x] Unit test
- [x] Test on K8s
- [x] Test on Yarn
- [x] Test on Local

### DIFF
--- a/python/orca/src/bigdl/orca/learn/tf2/pyspark_estimator.py
+++ b/python/orca/src/bigdl/orca/learn/tf2/pyspark_estimator.py
@@ -109,14 +109,19 @@ class SparkTFEstimator():
 
     def _get_cluster_info(self, sc):
         def get_worker_address(iter):
+            res = []
             worker_ip = get_node_ip()
             worker_port = find_free_port()
-            res = []
-            res.append(f"{worker_ip}:{worker_port}")
+            address = [f"{worker_ip}:{worker_port}"]
+            address.extend(find_ip_and_free_port(iter))
+            res.append(address)
             return res
+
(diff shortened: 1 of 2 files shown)


sgwhat:
Still processing the pod ip with worker rank.

hkvision (approved):

python/orca/src/bigdl/orca/learn/tf2/pyspark_estimator.py
```diff
@@ -108,8 +108,18 @@ def __init__(self,
             self.log_server_thread = start_log_server(self.ip, self.port)
 
     def _get_cluster_info(self, sc):
-        cluster_info = self.workerRDD.barrier().mapPartitions(find_ip_and_free_port).collect()
-        return cluster_info
+        def get_worker_address(iter):
+            worker_ip = get_node_ip()
```
hkvision, line 112: Remark: using this function can get the k8s virtual ip while using Sparkbarriercontext can only get the physical node ip

PR #7993, opened 2023-04-04
Tdx DL serving

Le-Zheng:
## Description
Tdx DL serving image


ppml/tdx/docker/trusted-dl-serving/Dockerfile
```diff
@@ -0,0 +1,187 @@
+ARG BIGDL_VERSION=2.3.0-SNAPSHOT
```
glorysdj, line 1: can we from ppml dl serving image or copy from it?

PR #7995, opened 2023-04-05
Pytorch Estimator Evaluation support customer input

jenniew:
In pytorch estimator evaluation, , the input format is hardcoded  to (features, target). When user give different input,  target=self.batch[-1] would throw error. 

So change to support custom input. And the default data input support will move to MainCallback class.

See issue: https://github.com/analytics-zoo/orca-llm/issues/1#issuecomment-1496569468

### DIFF
--- a/python/orca/src/bigdl/orca/learn/pytorch/callbacks/maincallback.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/callbacks/maincallback.py
@@ -93,6 +93,7 @@ class MainCallback(Callback):
         Any behavior inconsistent with the default training behavior should be overridden here.
         """
         self.on_iter_forward(runner)
+        runner.target = runner.batch[-1]
 
     def on_pred_forward(self, runner):
         """
--- a/python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
+++ b/python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
@@ -576,7 +576,8 @@ class TorchRunner(BaseRunner):
 
         # User should not see batch from last iteration
         # Assume target is always the last one in a batch.
-        target = self.batch[-1]
+        if hasattr(self, "target"):
+            target = self.target
         del self.batch
 
         if hasattr(self, "loss"):


python/orca/src/bigdl/orca/learn/pytorch/callbacks/maincallback.py
```diff
@@ -93,6 +93,7 @@ def on_val_forward(self, runner):
         Any behavior inconsistent with the default training behavior should be overridden here.
         """
         self.on_iter_forward(runner)
+        runner.target = runner.batch[-1]
```
leonardozcm, line 96:
```python
class MainCallback(Callback):
    """
    MainCallback is a one-of-a-kind callback that contains hook functions:
        - `on_iter_forward`
        - `on_iter_backward`
        - `on_lr_adjust`

    These methods are somewhat special, because only one special MainCallback
    should be allowed to implement these methods among all callbacks, otherwise
    there will propagate forward and backward twice.
    """
    def on_iter_forward(self, runner):
        """
        If `on_train_forward` and `on_val_forward` are not overridden,
        this will be called during forward when training and validating.
        Any behavior inconsistent with the default forward behavior should be overridden here.
        """
        # unpack features into features and targets
        *features, target = runner.batch
        runner.target = target
        # Forward features
        runner.output = runner.model(*features)
        # Ensure `targetL` and `outputL` are always in a list format.
        targetL = [target] if not isinstance(target, (list, tuple)) else target
        outputL = [runner.output] if not isinstance(runner.output, (list, tuple)) else runner.output
        # Compute loss
        runner.loss = runner.criterion(*outputL, *targetL)
```
move this to on_iter_forward

python/orca/src/bigdl/orca/learn/pytorch/torch_runner.py
```diff
@@ -576,7 +576,8 @@ def forward_batch(self, batch, callbacks=None):
 
         # User should not see batch from last iteration
         # Assume target is always the last one in a batch.
-        target = self.batch[-1]
+        if hasattr(self, "target"):
+            target = self.target
```
leonardozcm, line 580: del self.target

PR #8002, opened 2023-04-07
[Nano] Nightly-build UT issue: fix tfds version to less than 4.9

HensonMa:
According to the issue posted here, we should restrict tensorflow-datasets version as less than 4.9

### DIFF
--- a/.github/actions/nano/setup-tensorflow-env/action.yml
+++ b/.github/actions/nano/setup-tensorflow-env/action.yml
@@ -85,6 +85,6 @@ runs:
         pip install ConfigSpace optuna
 
         # required by NoteBook and Tutorial test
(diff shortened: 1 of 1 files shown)


TheaperDeng (approved):
LGTM

.github/actions/nano/setup-tensorflow-env/action.yml
```diff
@@ -85,6 +85,6 @@ runs:
         pip install ConfigSpace optuna
 
         # required by NoteBook and Tutorial test
-        pip install tensorflow-datasets jinja2 jupyter nbconvert nbmake
+        pip install tensorflow-datasets<4.9 jinja2 jupyter nbconvert nbmake
```
TheaperDeng, line 88: valid this installation syntax

PR #8009, opened 2023-04-10
Nano: replace part of the pickle to json to avoid security issue

TheaperDeng:
## Description
replace part of the unnecessary usage of pickle to json to avoid security issue


rnwang04 (approved):
LGTM

PR #8014, opened 2023-04-10
Fix benchmark security issue brought by sql injection

HensonMa:
## Description
Currently, users have direct access to file `python/benchmark/run.py`. It allows them to pass arguments to change the SQL contents, some of which may do harm to the data. Hence, we test the method that moving the file to a inner secure repo and fetch it during benchmarking.


HensonMa:
APRT

TheaperDeng (approved):

PR #8017, opened 2023-04-11
[Nano] Support torch 2.0 in setup and action test

rnwang04:
## Description

Support torch 2.0 in setup and action test

### 1. Why the change?

Support torch 2.0 in setup and action test to provide new installation option for user and support more 2.0 related features in the future.

### 2. User API changes

No change.

### 3. Summary of the change 

- [x] upgrade action default python version (in pytorch action test) from 3.7 to 3.8.16 as torch 2.0 only support python 3.8 and higher versions
- [x] torch2.0 requires inc >= 2.1 and inc 2.1 has a little different API with inc 2.0
- [x] ipex 2.0 update the check for AVX2 instruction sets, add related avx2_checker
`SystemExit: ERROR! Intel® Extension for PyTorch* only works on machines with instruction sets equal or newer than AVX2, which are not detected on the current machine.`
- [x] PyTorch Lightning scheduler issue
`pytorch_lightning.utilities.exceptions.MisconfigurationException: The provided lr scheduler `LambdaLR` doesn't follow PyTorch's LRScheduler API. You should override the `LightningModule.lr_scheduler_step` hook with your own logic if you are using a custom LR scheduler.`
we can't fix it except we upgrade pytorch lightning. Just skip related tests now.
We may upgrade pytorch lightning in next PR and update related code later.

### 4. How to test?
- [x] Unit test

### 5. What to do after this PR
- Upgrade PyTorch-Lightning to 2.0.0
- check or report the AVX2 behaviour of ipex


.github/workflows/nano_unit_tests_pytorch.yml
```diff
@@ -189,6 +190,7 @@ jobs:
             bash python/nano/dev/build_and_install.sh linux default false ${requirements[0]} -f https://software.intel.com/ipex-whl-stable
             pip install ${requirements[1]}
           fi
+          pip install numpy==1.23.4
```
TheaperDeng, line 193: is `numpy>1.23.4` not supported by nano or any of its dependencies?

python/nano/src/bigdl/nano/utils/common/checker.py
```diff
@@ -62,3 +67,14 @@ def _avx512_checker():
...
     return cpuinfo.has_avx512
+
+
+def _avx2_checker():
+    # copied from https://github.com/intel/intel-extension-for-pytorch
+    try:
+        if compare_version("intel_extension_for_pytorch", operator.ge, "2.0.0"):
+            import intel_extension_for_pytorch._isa_help as isa
+            return isa._check_isa_avx2()
+    except SystemExit:
```
TheaperDeng, line 78: we may meet `ImportError` here for non-ipex users right? This may cause some minor issues for a non-AVX2 && non-ipex user

python/nano/src/bigdl/nano/utils/common/checker.py
```diff
@@ -16,6 +16,11 @@
 
 from importlib.util import find_spec
 from bigdl.nano.utils.common import CPUInfo
+import os
+import operator
+import subprocess
+from bigdl.nano.utils.common import invalidInputError
```
TheaperDeng, line 22: why we need `os`, `subprocess` and `invalidInputError`

python/nano/src/bigdl/nano/utils/common/checker.py
```diff
@@ -62,3 +67,14 @@ def _avx512_checker():
     '''
     cpuinfo = CPUInfo()
     return cpuinfo.has_avx512
+
+
+def _avx2_checker():
```
TheaperDeng, line 72: This checker is used to avoid avx2 checking issue involved by IPEX (since AVX2 is involved by [Haswell](https://en.wikipedia.org/wiki/Haswell_(microarchitecture)) processor in 2013, very few users are using previous products). Should we post an issue to IPEX team?

python/nano/test/run-nano-pytorch-ray-tests.sh
```diff
@@ -11,7 +11,10 @@ set -e
 echo "# Start testing"
 start=$(date "+%s")
 python -m pytest -s ${NANO_RAY_TEST_DIR}/test_ray_trainer.py ${NANO_RAY_TEST_DIR}/test_torch_nano_ray.py
-python -m pytest -s ${ANALYTICS_ZOO_ROOT}/python/nano/test/pytorch/tests/train/trainer/test_scale_lr.py -k 'ray'
+version=$(python -c "import torch;print(torch.__version__)")
+if [ ${version} != "2.0.0+cu117" ];then  # workaround as we have not upgrade pl
```
TheaperDeng, line 15: should we use sth similar to `startwith`?

python/nano/src/bigdl/nano/utils/common/checker.py
rnwang04: Just forget to delete them ：） Have fixed.

python/nano/src/bigdl/nano/utils/common/checker.py
rnwang04: Yeah, I will raise a related issue to IPEX team. But we still need this checker to avoid avx2 checking issue as least for ipex2.0 .

python/nano/src/bigdl/nano/utils/common/checker.py
rnwang04: Thanks for reminder! will add ImportError here : )

.github/workflows/nano_unit_tests_pytorch.yml
rnwang04: when numpy>1.23.4, ` np.bool `has been completely abandoned. However, inc has `np.bool` itself, so if numpy>1.23.4, we can't even import inc.

python/nano/test/run-nano-pytorch-ray-tests.sh
rnwang04: Done

TheaperDeng (approved):
LGTM

PR #8020, opened 2023-04-11
[PPML] Add MAA part in attestation document

xiangyuT:
## Description

Refine and add `How to use and configure attestation policy in Microsoft Azure Attestation (MAA) with AttestationCLI` in AttestationCLI document.


hzjane (approved):
LGTM

PR #8027, opened 2023-04-11
[Nano] Support `example_kwarg_inputs` for `jit.trace` when torch>=2.0

rnwang04:
## Description

Support `example_kwarg_inputs` paramter for `jit.trace` when torch>=2.0 to support more diverse input forms (especially when None is included in the input).

### 1. Why the change?

Fix issue8 of https://github.com/analytics-zoo/nano/issues/327.

### 2. User API changes

new paramater  `example_kwarg_inputs`  for `InferenceOptimizer.trace` and `InferenceOptimizer.quantize`.
```python
class JumpInputNet(nn.Module):
    def __init__(self):
        super().__init__()
        self.dense1 = nn.Linear(10, 1)
        self.dense2 = nn.Linear(10, 1)

    def forward(self, x1, x2=None, x3=None):
        if x3 is not None:
            return self.dense1(x1) + self.dense2(x3)
        else:
            return self.dense1(x1)

opt_model = InferenceOptimizer.trace(model, accelerator="jit",
                                     input_sample=None,
                                     example_kwarg_inputs={'x1':x1, 'x3':x3})
output1 = opt_model(x1, x3)
```

### 3. Summary of the change 

- expose new paramater  `example_kwarg_inputs`  for `InferenceOptimizer.trace` and `InferenceOptimizer.quantize`.
- related UT for `PytorchIPEXJITModel`, `PytorchIPEXJITBF16Model` and `PytorchJITINT8Model`

### 4. How to test?
- [x] Unit test
- [x] Local unit test

### DIFF
--- a/python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
+++ b/python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
@@ -15,13 +15,12 @@
 #
 
 from bigdl.nano.utils.pytorch import generate_channels_last_available,\
-    apply_proper_channels_last, transform_state_dict_to_dtype
+    apply_proper_channels_last, transform_state_dict_to_dtype, \
+    patch_attrs_from_model_to_object, jit_convert
 from bigdl.nano.pytorch.model import AcceleratedLightningModule
 from bigdl.nano.pytorch.context_manager import generate_context_manager
 from bigdl.nano.deps.ipex.ipex_api import ipex_optimize
 from bigdl.nano.utils.common import invalidInputError, compare_version
-from bigdl.nano.utils.pytorch import patch_attrs_from_model_to_object
-from bigdl.nano.utils.pytorch import transform_state_dict_to_dtype
 import operator
 import torch
 import copy
@@ -73,9 +72,9 @@ class PytorchIPEXJITModel(AcceleratedLightningModule):
                to load the compressed file if compression is set other than "fp32".
                Currently, "bf16" and "fp32"(default) are supported.
         :param example_kwarg_inputs: keyword arguments of example inputs that will be passed
-               to ``torch.jit.trace``. Default to None. Either this argument or input_sample
-               should be specified when use_jit is ``True`` and torch > 2.0,
-               otherwise will be ignored.
+               to ``torch.jit.trace``. Default to ``None``. Either this argument or
+               ``input_sample`` should be specified when ``use_jit`` is ``True`` and
+               torch > 2.0, otherwise will be ignored.
         """
         super().__init__(model)
         if from_load:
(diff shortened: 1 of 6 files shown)


python/nano/src/bigdl/nano/deps/ipex/ipex_quantization_model.py
```diff
@@ -52,6 +55,10 @@ def __init__(self, model: torch.nn.Module, calib_data, q_config=None,
         :param from_load: this will only be set by _load method.
         :param inplace: whether to perform inplace optimization. Default: ``False``.
         :param jit_strict: Whether recording your mutable container types.
+        :param example_kwarg_inputs: keyword arguments of example inputs that will be passed
```
MeouSker77, line 58: may change None, input_sample and use_jit to ` ``None`` `, ` ``input_sample`` ` and ` ``use_jit`` `

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
```diff
@@ -745,6 +746,12 @@ def quantize(model: nn.Module,
                                only valid when accelerator='onnxruntime' or accelerator='openvino',
                                otherwise will be ignored. If output_tensors=False, output of the
                                export model will be ndarray.
+        :param example_kwarg_inputs: a pack of keyword arguments of example inputs that will be
```
MeouSker77, line 749: Why this comment is different from others

python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
```diff
@@ -124,33 +129,68 @@ def __init__(self, model: torch.nn.Module, input_sample=None, use_ipex=False, dt
...
+                                        example_kwarg_inputs=example_kwarg_inputs)
+                                else:
+                                    self.model = torch.jit.trace(self.model, input_sample,
+                                                                 check_trace=False,
+                                                                 strict=jit_strict)
                             except Exception:
                                 self.model = torch.jit.script(self.model)
                         if self.use_ipex:
                             self.model = torch.jit.freeze(self.model)
             else:
```
MeouSker77, line 163: There is too much redundant code here.
For example, this `if` `else`  can be removed, you only need to remove line 128 and this `else` branch, then change line 130 to
```python
with torch.cpu.amp.autocast(enabled=dtype == torch.bfloat16):
```

python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
```diff
@@ -124,33 +129,68 @@ def __init__(self, model: torch.nn.Module, input_sample=None, use_ipex=False, dt
                 with torch.no_grad():
                     with torch.cpu.amp.autocast():
                         if self.jit_method == 'trace':
-                            self.model = torch.jit.trace(self.model, input_sample,
-                                                         check_trace=False,
-                                                         strict=jit_strict)
+                            if compare_version("torch", operator.ge, "2.0"):
```
MeouSker77, line 132: And I think you should define a new function to do jit trace instead of copy paste everywhere

MeouSker77 (approved):
others LGTM

python/nano/src/bigdl/nano/deps/ipex/ipex_quantization_model.py
rnwang04: Done

python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
rnwang04: Good suggestion, thanks! Have revised to this form.

python/nano/src/bigdl/nano/deps/ipex/ipex_inference_model.py
rnwang04: Good suggestion, thanks! Have revised to this form.

python/nano/src/bigdl/nano/pytorch/inference/optimizer.py
rnwang04: all ipex models have the same comment.

PR #8034, opened 2023-04-12
Orca: fix loading model without optimizer

hamham223:
### 1. Why the change?

Fix the issue when the model does not have an optimizer.

optimizer

### 2. How to test?

- [ ] Unit test


hamham223:
@sgwhat  Could you please take a look?

sgwhat (approved):
LGTM

PR #8035, opened 2023-04-12
[Nano] Support `**kwargs` for `AcceleratedLightningModule`/`OpenVINO`/`ONNXRuntime`

rnwang04:
## Description

Previously, both OpenVINO and ONNXRuntime optimized models did not support passing in keyword parameters, resulting in **inconsistent user behavior** when calling the accelerated model, such as `model(x1=x1, x2=x2)` and `opt_model(x1, x2)`.

This PR will enable the optimized models of OpenVINO and ONNXRuntime to also support keyword arguments.

### 1. Why the change?

Fix **issue2(missing kwargs which cause rewritten forward function)** of https://github.com/analytics-zoo/nano/issues/327

### 2. User API changes
After this PR, user can pass in keyword arguments for OpenVINO and ONNXRuntime optimized models(trace and quantize), below code takes OpenVINO fp32 as an example:
```python
class MultipleInputNet(nn.Module):
    def __init__(self):
        super().__init__()
        self.dense1 = nn.Linear(10, 1)
        self.dense2 = nn.Linear(10, 1)

    def forward(self, x1, x2):
        return self.dense1(x1) + 2 * self.dense2(x2)

net = MultipleInputNet()
x1 = torch.randn(32, 10)
x2 = torch.randn(32, 10)

ov_model = InferenceOptimizer.trace(net,
                                    accelerator="openvino",
                                    input_sample=(x1,x2))
with InferenceOptimizer.get_context(ov_model):
    output1 = ov_model(x1, x2).numpy()
    # keyword argument is supported now
    output2 = ov_model(x1, x2=x2).numpy()
    output3 = ov_model(x1=x1, x2=x2).numpy()
```

### 3. Summary of the change 
- Support `**kwargs` for `AcceleratedLightningModule`
- Support `**kwargs` for `OpenVINO`'s trace and quantize, related UT
- Support `**kwargs` for `ONNXRuntime`'s trace and quantize, related UT
- Support int and float input for ONNXRuntime, update related UT

### 4. How to test?
- [ ] Unit test


python/nano/src/bigdl/nano/deps/openvino/core/model.py
```diff
@@ -40,17 +40,22 @@ def on_forward_start(self, inputs):
         self._model_exists_or_err()
         return inputs
 
-    def forward_step(self, *inputs):
+    def forward_step(self, *inputs, **kwargs):
         flattened_inputs = []
         _flatten(inputs, flattened_inputs)
+        args_length = len(inputs)
+        if kwargs is not None and len(kwargs) > 0:
+            for arg in self.forward_args[args_length:]:
```
MeouSker77, line 48: maybe the `kwargs` contains the same argument as the `inputs`, I mean if the signature is `def forward(a, b, c)`, then user call `model(t1, t2, b=t3, c=t4)`, I think we should give an error in such case, not just ignore the `b=t3`.

python/nano/src/bigdl/nano/utils/pytorch/tensor_convert.py
```diff
@@ -0,0 +1,49 @@
...
+    return to_numpy(tensors)
+
+
+def cope_with_keyword_arguments(kwargs):
+    # inplace convert kwargs
+    for k in kwargs.keys():
+        if isinstance(kwargs[k], tuple):
+            kwargs[k] = numpy_to_tensors(kwargs[k])
+        if isinstance(kwargs[k], torch.Tensor):
+            kwargs[k] = kwargs[k].numpy()
```
MeouSker77, line 47: may change to `x.cpu().detach().numpy()` ?

MeouSker77 (approved):
others LGTM

python/nano/src/bigdl/nano/utils/pytorch/tensor_convert.py
rnwang04: done

python/nano/src/bigdl/nano/deps/openvino/core/model.py
rnwang04: Good catch, thanks! Have fixed this.

python/nano/src/bigdl/nano/deps/onnxruntime/core/onnxruntime_model.py
```diff
@@ -28,13 +28,13 @@ def __init__(self, onnx_filepath, session_options=None):
...
 
-    def forward_step(self, *inputs):
+    def forward_step(self, *inputs, **kwargs):
         '''
         This function run through the onnxruntime forwarding step
         '''
         flattened_inputs = []
         _flatten(inputs, flattened_inputs)
-        if len(self._forward_args) != len(flattened_inputs):
+        if len(kwargs) == 0 and len(self._forward_args) != len(flattened_inputs):
```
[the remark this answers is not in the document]
rnwang04: Have changed to the second way. thanks!

PR #8038, opened 2023-04-12
Nano: add more error information about customized training loop

plusbang:
## Description

Add some error message to help users more easily enable customized training loop.

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/tf/keras/customized_training_utils.py
+++ b/python/nano/src/bigdl/nano/tf/keras/customized_training_utils.py
@@ -27,6 +27,7 @@ from bigdl.nano.utils.common import schedule_processors
 from bigdl.nano.tf.keras.distributed_utils import _find_free_port
 from tensorflow.keras.losses import Loss
 import warnings
+from bigdl.nano.utils.common import invalidInputError
 
 
 def nano_bf16(func):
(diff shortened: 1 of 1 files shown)


python/nano/src/bigdl/nano/tf/keras/customized_training_utils.py
```diff
@@ -132,6 +133,22 @@ def __call__(self, *args, **kwargs):
                     cloudpickle.dump(arg, f)
                     new_args.append(("others", os.path.join(temp_dir, f"args_{i}.pkl")))
 
+            # check necessary objects
+            arg_instance = list(map(lambda x: x[0], new_args))
+            invalidInputError("model" in arg_instance, "Please check whether 'model' in the input "
+                                                       "parameter in train function.")
```
TheaperDeng, line 139: this is quite strange to understand for users since we do not require a parameter called `'model'`. We may just make this "Please check if a model inherited from nn.Module is inputted as parameter in train function."

python/nano/src/bigdl/nano/tf/keras/customized_training_utils.py
```diff
@@ -132,6 +133,22 @@ def __call__(self, *args, **kwargs):
...
+                                                           "input parameter in train function.")
+            invalidInputError("loss" in arg_instance, "Please check whether 'loss' in the input "
+                                                      "parameter in train function. Moreover, if "
+                                                      "you use a customized loss, please add "
+                                                      "'nano_multiprocessing_loss' decorator.")
+            invalidInputError("dataset" in arg_instance, "Please check whether 'dataset' in the "
+                                                         "input parameter in train function. "
+                                                         "Moreover, if dataset is created by "
+                                                         "'from_generator', please initiate "
+                                                         "'_GeneratorState' of dataset first.")
```
TheaperDeng, line 150: same for these three.

python/nano/src/bigdl/nano/tf/keras/customized_training_utils.py
```diff
@@ -225,7 +242,11 @@ def loss_object(*args, **kwargs):
         with open(target_path, 'rb') as f:
             target_func = cloudpickle.load(f)
 
-        res = target_func(*actrual_args)
+        try:
+            res = target_func(*actrual_args)
+        except TypeError:
+            invalidInputError(False, "Please check whether add 'nano_multiprocessing' decorator "
```
TheaperDeng, line 248: "if you have added"

TheaperDeng (approved):
LGTM

PR #8045, opened 2023-04-13
Chronos: fix thread num error when use `get_context` and `predict_with_onnx` together

plusbang:
## Description

When we call `forecaster.get_context(thread_num=4)` and `forecaster.predict_with_onnx`, we receive warning to set OMP_NUM_THREADS=1. Obviously, it's incorrect.
In this pr, fix it.

### 2. User API changes

None

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/chronos/src/bigdl/chronos/pytorch/context_manager.py
+++ b/python/chronos/src/bigdl/chronos/pytorch/context_manager.py
@@ -57,6 +57,8 @@ class ForecasterContextManager(object):
         # call `torch.set_num_threads` at most once
         self.forecaster.thread_num = set_pytorch_thread(self.thread_num,
                                                         self.forecaster.thread_num)
+        self.forecaster.optimized_model_thread_num = \
+            set_pytorch_thread(self.thread_num, self.forecaster.optimized_model_thread_num)
         self.forecaster.context_enabled = True
         self.infer_mode.__enter__()
         if self.bf16_enable:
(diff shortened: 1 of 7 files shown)


python/chronos/src/bigdl/chronos/pytorch/context_manager.py
```diff
@@ -57,6 +57,9 @@ def __enter__(self):
         # call `torch.set_num_threads` at most once
         self.forecaster.thread_num = set_pytorch_thread(self.thread_num,
                                                         self.forecaster.thread_num)
+        if hasattr(self.forecaster, 'optimized_model_thread_num'):
```
TheaperDeng, line 60: In what case a forecaster does not has this `optimized_model_thread_num`? remind me please :)

TheaperDeng (commented):
others LGTM

python/chronos/src/bigdl/chronos/pytorch/context_manager.py
plusbang: > In what case a forecaster does not has this `optimized_model_thread_num`? remind me please :)

yeah, autoformer forecaster now only has `thread_num`, and I think we don't need to add `optimized_model_thread_num` to it currently.

TheaperDeng: ok make sense

TheaperDeng (approved):

PR #8047, opened 2023-04-14
[WIP]Nano: support deepspeed inference

MeouSker77:
## Description (Now it's only a draft)

Support deepspeed multi-process inference

### 2. User API changes

maybe
```python
...
model = InferenceOptimizer.to_deepspeed(model, nproc=n)
...  # use new model to do inference as before
```

### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/deps/deepspeed/api.py
+++ b/python/nano/src/bigdl/nano/deps/deepspeed/api.py
@@ -1,0 +1,93 @@
+#
+# Copyright 2016 The BigDL Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import os
+import sys
+import json
+import uuid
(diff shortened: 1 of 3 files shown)


MeouSker77:
I'm not sure why its performance is so poor, the following test code will print:
```python
import torch
import time

from torchvision.models import resnet50

from bigdl.nano.deps.deepspeed.api import DeepSpeedModel

inputs = torch.rand((16, 3, 512, 512))

t1 = 0
model = resnet50(pretrained=False)
model.eval()
with torch.inference_mode():
    for i in range(8):
        st = time.time()
        model(inputs)
        if i >= 3:
            t1 += time.time() - st

t2 = 0
deepspeed_model = DeepSpeedModel(model, 2)
with torch.inference_mode():
    for i in range(8):
        st = time.time()
        deepspeed_model(inputs)
        if i >= 3:
            t2 += time.time() - st
deepspeed_model.shutdown()

t3 = 0
multi_model = InferenceOptimizer.to_multi_instance(model, 2)
with torch.inference_mode():
    for i in range(8):
        st = time.time()
        multi_model([inputs[:8], inputs[8:]])
        if i >= 3:
            t3 += time.time() - st


print(f'single process: {t1}')
print(f'deepspeed: {t2}')
print(f'multiprocess: {t3}')
```

```
single process: 3.9192421436309814
deepspeed: 4.322096824645996
multiprocess: 1.9595201015472412
```

It is slower than single process(tested on a 2 socket clx)，and the more processes deepspeed uses, the slower it is.
If deepspeed only uses 1 process, the time will be about 3s, but if it uses 4 processes, the time will be about 8s.

Any ideas ? @yangw1234 

Btw, we can install this [branch](https://github.com/delock/DeepSpeedSYCLSupport/tree/gma/cpu_support) to use deepspeed with CPU.

PR #8048, opened 2023-04-14
Chronos: improve prvd and nb tests with more python versions

plusbang:
## Description

Related to https://github.com/analytics-zoo/chronos/issues/16, we may need to enable prvd and nb tests with both python3.8 and python3.9.

**Note**: to use orca on python3.8, we must use spark3. So we now directly rely on `orca-spark3` in setup and only test **spark3** package.

### 3. Summary of the change 

- enable all tests with python 3.8 and 3.9
- simplify, remove unnecessary workflow
- directly rely on `orca-spark3` in setup (to do this, change release order, release spark3 first and spark2 later)
- modify related description in document

### 4. How to test?
- [x] Unit test
- [x] Document test (https://binbin-doc.readthedocs.io/en/workflow-py38/doc/Chronos/Overview/install.html)


python/chronos/colab-notebook/howto/how_to_export_torchscript_files.ipynb
```diff
@@ -261,7 +261,7 @@
     "postprocess_output = tsdata_test.unscale_numpy(forecasting_output)\n",
     "\n",
     "# compare the results\n",
-    "assert_array_almost_equal(output_tensor.numpy(), postprocess_output)"
+    "assert_array_almost_equal(output_tensor.numpy(), postprocess_output, decimal=3)"
```
TheaperDeng, line 264: decimal=3 is a quite relax assertion, what about =5

python/chronos/use-case/AIOps/AIOps_anomaly_detect_unsupervised.ipynb
```diff
@@ -188,7 +188,7 @@
     {
      "data": {
       "text/plain": [
-       "<matplotlib.axes._subplots.AxesSubplot at 0x7f0d227e4400>"
+       "<AxesSubplot:title={'center':'cpu_usage of machine 1932'}, xlabel='time_step'>"
```
TheaperDeng, line 191: why is this notebook changed?

plusbang: > why is this notebook changed?

Because we change the input para of `set_params` in https://github.com/intel-analytics/BigDL/pull/8104, we need to fix this notebook.

TheaperDeng (approved):

PR #8059, opened 2023-04-17
Orca: fix type checking

hamham223:
### 1. Why the change?

ignore those strict type checking...

### 2. How to test?
- [x] Python-Style-Check
- [x] Orca-Pytest-Py37-AutoML


leonardozcm (approved):

python/orca/src/bigdl/orca/automl/auto_estimator.py
```diff
@@ -138,7 +138,7 @@ def fit(self,
                 Union[Callable, Tuple["ndarray", "ndarray"], "DataFrame"]]=None,
             metric: Optional[Union[Callable, str]]=None,
             metric_mode: Optional[str]=None,
-            metric_threshold: Optional[Union["Function", "Float", "Integer"]]=None,
+            metric_threshold: Optional[Union["Function", "float", "int"]]=None,
```
hamham223, line 141: Is this change ok? Need to use `Callable`?

PR #8060, opened 2023-04-17
[Nano] support non-tensor input to export onnx/openvino

HensonMa:
## Description
Export models to onnx will automatically flatten the inputs from dictionary, Tuple and List, but users may keep feed non-flattened inputs to the traced model. Hence, work needs to be done for both pre-processing user's unconventional inputs to let it be able to be traced, and post-processing the inputs to flattened ones to let the traced model able to do inference. 

## Issue
Currently `dynamic_axes` hasn't correctly been set. Hence, dynamic batch size has not supported yet, which needs further experiments and investigation. 

## Things about to do
- [x] Support fixed-size tensors tracing and inference for input samples as dictionary, Tuple and List.
- [ ] Write corresponding unit tests
- [ ] Investigate the issue stated above and enable dynamic batch size.


python/nano/test/onnx/pytorch/test_onnx.py
```diff
@@ -283,6 +284,45 @@ def forward(self, x1, x2):
...
+
+class MultipleInputNet_non_list_dict(nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.dense1 = nn.Linear(12, 1)
+        self.dense2 = nn.Linear(11, 1)
+        self.dense3 = nn.Linear(10, 1)
+        self.dense4 = nn.Linear(9, 1)
+
+    def forward(self, bbox: torch.Tensor, x: Dict):
```
TheaperDeng, line 322: what will happen if the forward is designed as `forward(self, x: Dict, bbox: torch.Tensor)`?

python/nano/src/bigdl/nano/utils/pytorch/convert.py
```diff
@@ -42,6 +43,17 @@ def export_to_onnx(model, input_sample=None, onnx_path="model.onnx", dynamic_axe
     forward_args = get_forward_args(model)
     tensor_args = get_conditional_args(model)
     input_sample = get_input_example(model, input_sample, forward_args)
+
+    if isinstance(input_sample, Sequence) and isinstance(input_sample[-1], Dict):
```
TheaperDeng, line 47: what if dict in input_sample is not at last?

python/nano/src/bigdl/nano/utils/pytorch/convert.py
```diff
@@ -42,6 +43,17 @@ def export_to_onnx(model, input_sample=None, onnx_path="model.onnx", dynamic_axe
     forward_args = get_forward_args(model)
     tensor_args = get_conditional_args(model)
     input_sample = get_input_example(model, input_sample, forward_args)
+
```
TheaperDeng, line 46: add some comment about what line 47-56 is doing, feel free to add an example in the comment. It may take future developer some time to understand :)

python/nano/src/bigdl/nano/utils/pytorch/convert.py
HensonMa: The logic I design the test cases is divided into two situations. 
- The **dict** type is at last
- The **dict** type is not at the last (any position except the last). 

Pytorch onnx team has a function **_decide_input_format()** to process the input sample. 
- When the **dict** type is not at the last, we don't need to do anything additional, just feed the original input _sample.
- When encountering the situation that the last item in the input sample is a **dict** type, the function will treat it as a dictionary containing optional arguments and resolve it  (read the keys of the dictionary, check if it is among the arguments of the forward function. If yes, then append it to the input list. If no, do nothing when the arg has no default value). So if we simply leave the input of **dict** type at the last, it won't be put in the input list by onnx by default, so I did a small trick here.  

For reference, the **_decide_input_format()** is here. 
https://github.com/pytorch/pytorch/blob/main/torch/onnx/utils.py#:~:text=def%20_decide_input_format(model%2C%20args)%3A

python/nano/test/onnx/pytorch/test_onnx.py
HensonMa: Please refer to the comments in the next conversation for reference. 
And I think the last test case contains this situation.

python/nano/src/bigdl/nano/utils/pytorch/convert.py
HensonMa: Sure! I'll leave completed comments later

python/nano/src/bigdl/nano/utils/pytorch/convert.py
TheaperDeng: yeah now I fully understand the logic, the `keys[0] not in forward_args` also contributes in the whole logic. Please do add some comment to this code snap :)

PR #8061, opened 2023-04-17
[Nano] Fix too strict OpenVINO testcase

cyita:
## Description

Some nodes may fail due to the too strict asserts: https://github.com/intel-analytics/BigDL/actions/runs/4718444558/jobs/8368056491

### 1. Why the change?

Some nodes may fail due to the too strict asserts: https://github.com/intel-analytics/BigDL/actions/runs/4718444558/jobs/8368056491

### 3. Summary of the change 

Reshape without compile: 30ms -> 50ms

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...


rnwang04 (approved):
LGTM

PR #8074, opened 2023-04-18
[Nano] multiprocessing support for pipeline

hoshibara:
## Description

Add multiprocessing support for Nano pipeline API. It can be useful in Alphafold optimization.

```python
pipeline = Pipeline([
    ("preprocess", preprocess, {"core_num": 8, 'worker_num': 8}),
    ("inference", model, {"core_num": 8, 'worker_num': 8}),
])
pipeline.run(input_list)
```

### DIFF
--- a/python/nano/src/bigdl/nano/pytorch/inference/pipeline.py
+++ b/python/nano/src/bigdl/nano/pytorch/inference/pipeline.py
@@ -45,8 +45,8 @@ class Pipeline():
         For example,
         ```
             pipeline = Pipeline([
-                ("preprocess", preprocess, {"core_num": 4, 'worker_num': 1}),
-                ("inference", model, {"core_num": 8, 'worker_num': 1}),
+                ("preprocess", preprocess, {"cores_per_worker": 4, 'worker_num': 1}),
+                ("inference", model, {"cores_per_worker": 8, 'worker_num': 1}),
             ])
         ```
         will create a pipeline which has stage "preprocess" and "inference",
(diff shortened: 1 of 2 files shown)


jason-dai:
Shall we specify "total core#" or "core# per worker"? @TheaperDeng

TheaperDeng:
> Shall we specify "total core#" or "core# per worker"? @TheaperDeng

We could call it `"core_num_per_worker"`, since for total core setting we may meet a ”inexhaustible-issue“ and I think `"core_num_per_worker"` and `worker_num` should be clear for our users to know what is happening

jason-dai:
> > Shall we specify "total core#" or "core# per worker"? @TheaperDeng
> 
> We could call it `"core_num_per_worker"`, since for total core setting we may meet a ”inexhaustible-issue“ and I think `"core_num_per_worker"` and `worker_num` should be clear for our users to know what is happening

Maybe just `cores_per_worker` and `worker_num`, with `worker_num` default to 1?

python/nano/src/bigdl/nano/pytorch/inference/pipeline.py
```diff
@@ -88,46 +84,57 @@ def run(self, inputs):
...
         """
         outputs = []
-        for i in inputs:
-            self.send_.send(i)
+        for idx, value in enumerate(inputs):
+            self.send_.put((idx, value))
             outputs.append(None)
 
-        outputs = [self.recv_.recv() for _ in outputs]
+        outputs = [self.recv_.get() for _ in outputs]
```
MeouSker77, line 91: maybe wen can avoid sorting,
```python
for _i in range(len(outputs)):
    idx, output = self.recv_.get()
    outputs[idx] = output
```

MeouSker77 (approved):
others LGTM

PR #8077, opened 2023-04-18
Fix serving and ppml vulnerabilities

Litchilitchy:
## Description

* fix serving vulnerability: upgrade `net.minidev:json-smart` to 2.4.9
* add wait termination for gRPC clients in ppml vulnerability

### DIFF
--- a/scala/serving/pom.xml
+++ b/scala/serving/pom.xml
@@ -335,6 +335,11 @@
             <artifactId>avatica-core</artifactId>
             <version>1.22.0</version>
         </dependency>
+        <dependency>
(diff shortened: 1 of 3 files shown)


scala/serving/pom.xml
```diff
@@ -335,6 +335,11 @@
             <artifactId>avatica-core</artifactId>
             <version>1.22.0</version>
         </dependency>
+        <dependency>
```
glorysdj, line 338: why this dependency is needed?

glorysdj (approved):
LGTM

PR #8052, opened 2023-04-14
[PPML] Support gramine-sgx LGBM training in trusted-machine-learning

SEULiyao:
## Description

Added `ppml/trusted-machine-learning/bash.manifest.template` to make LGBM work properly in trusted-machine-learning image
Modified `THIS_DIR` in `ppml/trusted-machine-learning/custom-image/make-sgx.sh` to `/ppml`
Updated dockerfile for machine learning base image and custom image

PR #8099, opened 2023-04-21
[PPML] Support SGX  LightGBM training

SEULiyao:
## Description

transferred from #8052 
Support SGX LightGBM training


Uxito-Ada (approved):
lgtm

PR #8100, opened 2023-04-21
Orca: Fix ut `Colab-Python-Py37-Pytorch` (branch 2.3)

hamham223:
### 1. Why the change?

Fix the ut.
PR in main branch: https://github.com/intel-analytics/BigDL/pull/8093


Le-Zheng:
LGTM

PR #8104, opened 2023-04-23
Chronos: enhance `ThresholdDetector` to better handle anomaly detection of prediction data

plusbang:
## Description

Currently, `ThresholdDetector` only supports anomaly detection of true data, we need to enable it supporting anomaly detection of prediction data as well.

Besides, we now support pattern/trend anomaly detection in internal implementation.

### 2. User API changes

Despite of previous attr `threshold`, we now support two other attrs `pattern_threshold` and `trend_threshold`
```bash
set_params(self, mode="default", ratio=0.01, pattern_threshold=math.inf, trend_threshold=(-math.inf, math.inf), dist_measure=EuclideanDistance()):
# pattern_threshold: a single value, specify absolute distance threshold between real data and predicted data to detect pattern anomaly.
# trend_threshold: a tuple composed of min_threshold and max_threshold, specify min and max threshold for real data to detect trend anomaly.
```

### 3. Summary of the change 

- internal implementation of `ThresholdDetector`
- related uts
- related use case

### 4. How to test?
- [x] Unit test


TheaperDeng (approved):
LGTM, leave refining API for future

plusbang:
> LGTM, leave refining API for future

yeah, we can provide better API in later PR.

PR #8137, opened 2023-04-26
[Nano] Upgrade python to 3.8

rnwang04:
## Description

Upgrade python version from 3.7.10 to 3.8.16 in all Nano action tests and update related installation doc.

### 1. Why the change?

Start from torch2.0, python 3.7 is not supported anymore. So we should upgrade to python 3.8 as 3.7 is too old and will be deprecated soon.

### 2. User API changes

No change.

### 3. Summary of the change 

- Upgrade python version from 3.7.10 to 3.8.16 in all Nano action tests
- Update related installation doc
- upgrade python version in [performance test](https://github.com/intel-analytics/BigDL/blob/86231d9b96212550ee0773ffcf8f4af50be25289/.github/workflows/performance-regression-test.yml#L30) and [nano nightly test](https://github.com/intel-analytics/BigDL/blob/86231d9b96212550ee0773ffcf8f4af50be25289/.github/workflows/nano-nightly-test.yml#L31)

### 4. How to test?
- [x] Unit test

Note:
now pytorch-lightning 1.6.4 requires `protobuf<3.20`, however, tensorflow-datasets requires `protobuf>=3.20`.
We may upgrade protobuf after we upgrade pytorch-lightning.

### DIFF
--- a/.github/workflows/nano-nightly-test.yml
+++ b/.github/workflows/nano-nightly-test.yml
@@ -28,7 +28,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.7", "3.8", "3.9"]
+        python-version: ["3.8", "3.9"]
 
     steps:
     - uses: actions/checkout@v3
@@ -61,7 +61,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.7", "3.8", "3.9"]
+        python-version: ["3.8", "3.9"]
 
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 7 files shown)


rnwang04:
> LGTM, may also upgrade python version in [performance test](https://github.com/intel-analytics/BigDL/blob/86231d9b96212550ee0773ffcf8f4af50be25289/.github/workflows/performance-regression-test.yml#L30) and [nano nightly test](https://github.com/intel-analytics/BigDL/blob/86231d9b96212550ee0773ffcf8f4af50be25289/.github/workflows/nano-nightly-test.yml#L31)

Thanks for reminder! Have upgraded these python version too.

PR #8146, opened 2023-04-28
[Nano] Add how-to-guide for pytorch async pipeline

HensonMa:
Preview link here https://henrytestnano-pipeline-doc.readthedocs.io/en/latest/doc/Nano/Howto/Inference/PyTorch/accelerate_pytorch_inference_async_pipeline.html


TheaperDeng:
For the introduction, please change to:

> 
> If there are multiple stages (e.g., preprocessing, model inferencing, postprocessing, ...) in your AI application, an asynchronized pipeline that parallelize different stages could significantly improve the inference throughput of the whole pipeline.
> 
> In a pipeline with N stages could work like this, where multiple (not necessarily to be the same number) workers could be enabled for each stage. The data will flow along each stages.
> image

TheaperDeng (approved):

Issue #8151, opened 2023-05-01
RuntimeError: Socket Timeout when running pyspark pytorch Orca Estimator training on multiple nodes with big models


jenniew:
Running Pytorch Pyspark Estimator training on multiple nodes with big models often got RuntimeError: Socket Timeout- when workers init_process_group. The trace back is as below:
Exception: Traceback (most recent call last):
  File "/opt/spark/python/lib/pyspark.zip/pyspark/worker.py", line 604, in main
    process()
  File "/opt/spark/python/lib/pyspark.zip/pyspark/worker.py", line 594, in process
    out_iter = func(split_index, iterator)
  File "/opt/spark/python/lib/pyspark.zip/pyspark/rdd.py", line 2863, in func
  File "/opt/bigdl-2.3.0-SNAPSHOT/python/bigdl-orca-spark_3.1.3-2.3.0-SNAPSHOT-python-api.zip/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py", line 370, in <lambda>
  File "/opt/bigdl-2.3.0-SNAPSHOT/python/bigdl-orca-spark_3.1.3-2.3.0-SNAPSHOT-python-api.zip/bigdl/orca/learn/pytorch/pytorch_pyspark_estimator.py", line 367, in transform_func
  File "./bigdl-orca-spark_3.1.3-2.3.0-SNAPSHOT-python-api.zip/bigdl/orca/learn/pytorch/pytorch_pyspark_worker.py", line 93, in __init__
    self.setup_distributed(self.mode, cluster_info, driver_ip, driver_tcp_store_port)
  File "./bigdl-orca-spark_3.1.3-2.3.0-SNAPSHOT-python-api.zip/bigdl/orca/learn/pytorch/pytorch_pyspark_worker.py", line 116, in setup_distributed
    self.setup_torch_distribute(tcp_store_host=driver_ip,
  File "./bigdl-orca-spark_3.1.3-2.3.0-SNAPSHOT-python-api.zip/bigdl/orca/learn/pytorch/core/lifecycle.py", line 29, in setup_torch_distribute
    self._init_torch_ddp(tcp_store_host, tcp_store_port, world_rank,
  File "./bigdl-orca-spark_3.1.3-2.3.0-SNAPSHOT-python-api.zip/bigdl/orca/learn/pytorch/core/lifecycle.py", line 72, in _init_torch_ddp
    dist.init_process_group(
[...]

PR #8152, opened 2023-05-01
Increase client TCPStore timeout to support big model in Pytorch Pyspark Estimator training

jenniew:
Increase client TCPStore timeout to support big model in Pytorch Pyspark Estimator training.

Related issue: https://github.com/intel-analytics/BigDL/issues/8151


songhappy:
LGTM

songhappy (approved):

Issue #8138, opened 2023-04-26
Cannot set compile to false when load model without model_creator


lalalapotter:
When load model with out model_creator to do inference as follows:
```shell
est = Estimator.from_keras(backend="spark", workers_per_node=1)
est.load("hdfs:///path/to/saved_model", compile=False)
est.predict(data=test,
            batch_size=batch_size,
            feature_cols=["input_1", "input_2", "input_3"])
est.shutdown()
```
, encountered following error:

```shell
Traceback (most recent call last):
  File "tiny_example.py", line 174, in <module>
    feature_cols=["input_1", "input_2", "input_3"])
  File "/disk0/yarn/nm/usercache/kai/appcache/application_1668477395550_2041/container_1668477395550_2041_01_000002/bigdl-spark_2.4.6-2.3.0-SNAPSHOT-python-api.zip/bigdl/orca/learn/tf2/pyspark_estimator.py", line 458, in predict
  File "/disk0/yarn/nm/usercache/kai/appcache/application_1668477395550_2041/container_1668477395550_2041_01_000002/bigdl-spark_2.4.6-2.3.0-SNAPSHOT-python-api.zip/bigdl/orca/learn/utils.py", line 178, in convert_predict_xshards_to_dataframe
  File "/disk0/yarn/nm/usercache/kai/appcache/application_1668477395550_2041/container_1668477395550_2041_01_000002/bigdl-spark_2.4.6-2.3.0-SNAPSHOT-python-api.zip/bigdl/orca/learn/utils.py", line 214, in convert_predict_rdd_to_dataframe
  File "/disk0/yarn/nm/usercache/kai/appcache/application_1668477395550_2041/container_1668477395550_2041_01_000002/pyspark.zip/pyspark/sql/session.py", line 58, in toDF
  File "/disk0/yarn/nm/usercache/kai/appcache/application_1668477395550_2041/container_1668477395550_2041_01_000002/pyspark.zip/pyspark/sql/session.py", line 746, in createDataFrame
  File "/disk0/yarn/nm/usercache/kai/appcache/application_1668477395550_2041/container_1668477395550_2041_01_000002/pyspark.zip/pyspark/sql/session.py", line 390, in _createFromRDD
  File "/disk0/yarn/nm/usercache/kai/appcache/application_1668477395550_2041/container_1668477395550_2041_01_000002/pyspark.zip/pyspark/sql/session.py", line 361, in _inferSchema
[...]

PR #8157, opened 2023-05-04
[ORCA] Support custom objects with tf2 pyspark estimator (none model_creator)

sgwhat:
## Description
Fix and update for https://github.com/intel-analytics/BigDL/issues/8138.

### 1. Why the change?
In some cases, we may need to support users in specifying tf load model parameters, including `custom objects` and `compile`, which were not available in the previous version.

### 2. User API changes
For the new usage:
```python
estimator = Estimator.from_keras(backend="spark")
estimator.load(trained_model_path, custom_object=None, compile=False)
```

### 3. Summary of the change 
Support users to specify more load model params when calling `estimator.load( )` with none model_creator.

### 4. How to test?
- [x] Unit test
- [x] Test on Yarn

### DIFF
--- a/python/orca/src/bigdl/orca/learn/tf2/spark_runner.py
+++ b/python/orca/src/bigdl/orca/learn/tf2/spark_runner.py
@@ -297,7 +297,7 @@ class SparkRunner:
                 self.setup_local(self.custom_objects, self.compile)
         else:
             # create local model for predict
-            self.setup_local(self.custom_object, self.compile)
+            self.setup_local(self.custom_objects, self.compile)
 
     def setup(self):
         import tensorflow as tf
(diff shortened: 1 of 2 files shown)


python/orca/src/bigdl/orca/learn/tf2/spark_runner.py
```diff
@@ -297,13 +306,13 @@ def setup(self):
         os.environ["KMP_BLOCKING_TIME"] = self.config.get("KMP_BLOCKING_TIME",
                                                           os.environ.get("KMP_BLOCKING_TIME", "0"))
 
-    def setup_local(self):
+    def setup_local(self, custom_objects, compile):
```
lalalapotter, line 309: Maybe modify the function to `def setup_local(self)`, and use class member variables in function

lalalapotter (approved):
Others LGTM.

PR #8159, opened 2023-05-05
Reduce SGX EPC of trusted-dl-serving and manually_build_for_testing tests

Hanyu-Jin:
## Description

Reduce EPCM.

### 1. Why the change?

To exclude the influence of EPCM.


qiyuangong (approved):
LGTM

PR #8160, opened 2023-05-05
Add MKL_DYNAMIC to bigdl-nano sgx profile

gc-fu:
## Description

I am currently testing the newest bigdl-nano nightly-release.  It seems that not setting `MKL_DYNAMIC` still leads to performance drop

Without `MKL_DYNAMIC`, the performance is around 6 while adding it leads to a throughput of 9.


python/nano/scripts/bigdl-nano-init
```diff
@@ -307,6 +307,7 @@ if [ "${OPENMP}" -eq 1 ]; then
             export MKL_NUM_THREADS=$core_
             export OMP_PROC_BIND="close"
             export OMP_SCHEDULE="STATIC"
+            export MKL_DYNAMIC="FALSE"
```
MeouSker77, line 310: may also unset it in `bigdl-nano-unset-env`

MeouSker77 (approved):

python/nano/scripts/bigdl-nano-init
gc-fu: Will fix, thx

MeouSker77 (approved):
LGTM

PR #8163, opened 2023-05-05
[Nano] Support python 3.10 in nightly test

rnwang04:
## Description

Support python 3.10 in nightly test and installation doc

### 1. Why the change?

To support python 3.10 for Nano

### 2. User API changes

None

### 3. Summary of the change 

- Support python 3.10 in nightly test and installation doc
- upgrade onnxruntime-extensions to 0.7.0 in setup 
- upgrade ray to 1.13.0 in workflow
- change to `from collections.abc import Iterable`

### 4. How to test?
- [x] Unit test

### DIFF
--- a/.github/actions/nano/setup-pytorch-env/action.yml
+++ b/.github/actions/nano/setup-pytorch-env/action.yml
@@ -81,7 +81,7 @@ runs:
 
         # required by Ray test
         pip install google-api-core==2.8.2
-        pip install ray[default]==1.11.0 prometheus_client==0.13.0
+        pip install ray[default]==1.13.0 prometheus_client==0.13.0
 
         # required by Automl test
         pip install ConfigSpace optuna
--- a/python/nano/setup.py
+++ b/python/nano/setup.py
@@ -167,10 +167,7 @@ def setup_package():
 
     inference_requires = ["onnx==1.12.0",
                           "onnxruntime==1.12.1",
-                          "onnxruntime-extensions==0.4.2; platform_system!='Darwin' and \
-                          python_full_version<'3.10.0'",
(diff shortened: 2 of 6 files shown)


.github/actions/nano/setup-pytorch-env/action.yml
```diff
@@ -81,7 +81,7 @@ runs:
 
         # required by Ray test
         pip install google-api-core==2.8.2
-        pip install ray[default]==1.11.0 prometheus_client==0.13.0
+        pip install ray[default]==1.13.0 prometheus_client==0.13.0
```
TheaperDeng, line 84: I think it's all right to do this for now, but we should bear in mind that orca only support 1.9.2 and there is a gap between nano's ray and orca's ray.

python/nano/setup.py
```diff
@@ -167,7 +167,10 @@ def setup_package():
 
     inference_requires = ["onnx==1.12.0",
                           "onnxruntime==1.12.1",
-                          "onnxruntime-extensions==0.4.2; platform_system!='Darwin'",
+                          "onnxruntime-extensions==0.4.2; platform_system!='Darwin' and \
+                          python_full_version<'3.10.0'",
+                          "onnxruntime-extensions==0.7.0; platform_system!='Darwin' and \
+                          python_full_version>='3.10.0'",
```
TheaperDeng, line 173: should we just bump to 0.7.0 for all python versions? I think it still support py37 ~ py39 https://pypi.org/project/onnxruntime-extensions/#files

TheaperDeng (approved):
others LGTM

python/nano/setup.py
rnwang04: Done

PR #8165, opened 2023-05-05
[Nano] Revert reshape if compile failed

cyita:
## Description

If reshape success but fails to compile, the user cannot reshape to the same size again since the network has already reshaped to the specific size.

### 1. Why the change?

Fix this bug.

### 2. User API changes

Now the reshape function will return reshape status: success, and the error_msg if reshape failed.

### 3. Summary of the change 

1. Revert the reshaping process if compile failed.
2. Return success status and error message.

### 4. How to test?
- [ ] Unit test


rnwang04 (approved):
LGTM

PR #8167, opened 2023-05-06
Refine orchestrate-tdxvm.sh

liu-shaojun:
This submission refines orchestrate-tdxvm.sh to let the user specify how many VMs they need.


glorysdj (approved):
LGTM

Issue #6838, opened 2022-12-01
TypeError: 'NoneType' object is not callable when run TF2 Estimator ray backend


hkvision:
```
Stopping orca context
Exception ignored in: <function SparkXShards.__del__ at 0x7f522a93add0>
Traceback (most recent call last):
  File "/home/yingzhe/anaconda3/envs/bigdl_spark3/lib/python3.7/site-packages/bigdl/orca/data/shard.py", line 811, in __del__
  File "/home/yingzhe/anaconda3/envs/bigdl_spark3/lib/python3.7/site-packages/bigdl/orca/data/shard.py", line 238, in uncache
  File "/home/yingzhe/anaconda3/envs/bigdl_spark3/lib/python3.7/site-packages/pyspark/rdd.py", line 315, in unpersist
  File "/home/yingzhe/anaconda3/envs/bigdl_spark3/lib/python3.7/site-packages/py4j/java_gateway.py", line 1296, in __call__
  File "/home/yingzhe/anaconda3/envs/bigdl_spark3/lib/python3.7/site-packages/py4j/java_gateway.py", line 1260, in _build_args
  File "/home/yingzhe/anaconda3/envs/bigdl_spark3/lib/python3.7/site-packages/py4j/java_gateway.py", line 1246, in _get_args
  File "/home/yingzhe/anaconda3/envs/bigdl_spark3/lib/python3.7/site-packages/py4j/java_collections.py", line 490, in can_convert
  File "/home/yingzhe/anaconda3/envs/bigdl_spark3/lib/python3.7/abc.py", line 139, in __instancecheck__
TypeError: 'NoneType' object is not callable
Exception ignored in: <function SparkXShards.__del__ at 0x7f522a93add0>
Traceback (most recent call last):
  File "/home/yingzhe/anaconda3/envs/bigdl_spark3/lib/python3.7/site-packages/bigdl/orca/data/shard.py", line 811, in __del__
  File "/home/yingzhe/anaconda3/envs/bigdl_spark3/lib/python3.7/site-packages/bigdl/orca/data/shard.py", line 238, in uncache
  File "/home/yingzhe/anaconda3/envs/bigdl_spark3/lib/python3.7/site-packages/pyspark/rdd.py", line 315, in unpersist
  File "/home/yingzhe/anaconda3/envs/bigdl_spark3/lib/python3.7/site-packages/py4j/java_gateway.py", line 1296, in __call__
[...]

PR #8171, opened 2023-05-08
Orca: enhance shard cache logic

hamham223:
### 1. Why the change?

1. To fix the [`TypeError: 'NoneType' object is not callable when run TF2 Estimator`](https://github.com/intel-analytics/BigDL/issues/6838). 
2. Enhance the shard cache logic.

### 2. Summary of the change 

1. Init `SparkXshards.user_caching` with `rdd.is_cached`, as it's possible that the rdd is already cached by user
2. User api `uncache()` will first try to access `self.rdd` by querying `self.rdd.is_cached`
    + If it's not reachable, namely the rdd does not exist or has been destroyed by spark, then catch the exception **(All kind of exceptions? or just Py4jError+TypeError?)**
    + If reachable, call `rdd.unpersist()` to uncache
3. `__del__` will call `_uncache()`, meaning that SparkXshards only uncaches automatically if the rdd is not cached by user

### 3. How to test?
- [x] Unit test
- [x] Test on yarn
- [x] Test on k8s
- [x] Test on Local
No other kind of exceptions found up to now.

### DIFF
--- a/python/orca/src/bigdl/orca/data/shard.py
+++ b/python/orca/src/bigdl/orca/data/shard.py
@@ -166,7 +166,7 @@ class SparkXShards(XShards):
                  transient: bool = False,
                  class_name: Optional[str] = None) -> None:
         self.rdd = rdd
-        self.user_caching = self.rdd.is_cached()
+        self.user_caching = self.rdd.is_cached
         if transient:
             self.eager = False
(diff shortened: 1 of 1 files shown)


cyita (approved):
LGTM

PR #8173, opened 2023-05-09
Nano: make get_cgroup_cpuset function flexible for a subset of resource

TheaperDeng:
### 1. Why the change?
The `schedule_processors` in nano which will be frequently used in multi-instance training/inferencing now will generate core binding strategy for _all cores_, while ignore the setting of
1. docker container's `--cpuset-cpu` setting
2. if users only want to use a subset of resource (e.g., `bigdl-nano-init -r 4` for 4 cores only)

### 2. User API changes
nothing

### 3. Summary of the change 
- `get_cgroup_cpuset` in `python/nano/src/bigdl/nano/utils/common/schedule.py` is rewrite by checking `/sys/fs/cgroup/cpuset/cpuset.cpus` if it exists.
- read OMP_NUM_THREADS to revise the cpuset value

### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/nano/src/bigdl/nano/utils/common/schedule.py
+++ b/python/nano/src/bigdl/nano/utils/common/schedule.py
@@ -111,6 +111,12 @@ def schedule_workers(num_workers: int,
             p2l[physical_core] = logical_core
     p_cores = sorted(p_cores_set)
 
+    # we check the OMP_NUM_THREADS setting as the only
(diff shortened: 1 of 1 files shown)


MeouSker77 (approved):
LGTM

gc-fu (approved):
LGTM

PR #8180, opened 2023-05-09
Only mount necessary etc files to Occlum

qzheng527:
## Description

There is no need mount the whole `/etc` directory to Occlum with hostfs mount which is not safe.


hzjane (approved):
LGTM

PR #8189, opened 2023-05-11
Nano: fix nightly test

MeouSker77:
## Description

fix nano nightly test


TheaperDeng (approved):
LGTM, we should use this env variable reading block for other vars in future :)

PR #8192, opened 2023-05-11
Move SGX and ATTESTATION from entrypoint to scripts in DL-Serving.

Hanyu-Jin:
## Description

Move SGX and ATTESTATION from entrypoint to scripts in DL-Serving.

### 1. Why the change?

Cannot start torchserve with attestation in entrypoint.


xiangyuT (approved):
LGTM

PR #8193, opened 2023-05-12
Orca: add requirement in docker.

lalalapotter:
## Description

Add requirement and install pypi library in docker.


docker/orca/install-orca-env.sh
```diff
@@ -1,6 +1,8 @@
 wget https://raw.githubusercontent.com/intel-analytics/BigDL/main/python/requirements/orca/requirements_automl.txt
+wget https://raw.githubusercontent.com/lalalapotter/BigDL/main/python/requirements/orca/requirements.txt
```
hkvision, line 2: revert the order?

hkvision (approved):

docker/orca/install-orca-env.sh
lalalapotter: Sure.

PR #8195, opened 2023-05-12
feat: unzip python dependency in bigdata toolkit

ShanSimu:
## Description

unzip python dependency in bigdata toolkit


hzjane (approved):
LGTM

PR #8206, opened 2023-05-17
[PPML] Add filepath parameter to save quote

xiangyuT:
## Description

Current quote output may be messed with logs. Add filepath parameter to save quote.


gc-fu (approved):
LGTM

qiyuangong (approved):
LGTM

PR #8207, opened 2023-05-17
[Nano] Update ipex to 2.0.1 for nano and restore ipex related tests

rnwang04:
## Description

Update ipex to 2.0.100 for Nano and restore ipex related tests

### 1. Why the change?

Follow up of https://github.com/intel-analytics/BigDL/pull/8017.
As ipex 2.0.1 has fixed  strange AVX2 behaviour in runner, we can upgrade to ipex 2.0.1 and restore ipex related tests.
image

### 2. User API changes

- Change default value of  `enable_onednn` to False

### 3. Summary of the change 

- Update ipex to 2.0.100 for Nano and restore ipex related tests
- Set default value of `enable_onednn` to False as some known issue and it will cause `RuntimeError: bad optional access` when torch=2.0 & use jit & input contain kwargs
- Jump the inplace UT as a new found issue of ipex 2.0.100 that it will lose weight of model during inplace optimization, has reported related issue to ipex team
### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/nano/setup.py
+++ b/python/nano/setup.py
@@ -120,7 +120,7 @@ def setup_package():
     # ipex is only avaliable for linux now
     pytorch_20_requires = ["torch==2.0.0",
                            "torchvision==0.15.1",
-                           "intel_extension_for_pytorch==2.0.0;platform_system=='Linux'"]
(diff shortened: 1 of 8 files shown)


MeouSker77 (approved):
LGTM

PR #8210, opened 2023-05-18
Orca: modify dockerfile, change ADD to COPY

lalalapotter:
## Description

This PR is to modify dockerfile, change ADD to COPY, due to the error in hadolint:
```shell
Use COPY instead of ADD for files and folders
```


sgwhat (approved):
LGTM

PR #8213, opened 2023-05-19
Chronos: fix SDL error of Dockerfile

plusbang:
## Description

Modify Chronos Dockerfile, replace `ADD` with `COPY`, because error in hadolint:
```bash
Use COPY instead of ADD for files and folders
```


TheaperDeng (approved):
LGTM

PR #8215, opened 2023-05-19
[PPML]feat: add encryption and decryption functions to BankDataFilter Example

ShanSimu:
## Description

add the encrypted data reading function to BankDataFilter Example


hzjane (approved):
LBTM

PR #8223, opened 2023-05-24
[PPML]doc: how to generate service account secret

ShanSimu:
## Description

doc: how to generate service account secret


hzjane (approved):
LGTM

PR #8225, opened 2023-05-25
LLM: add first round files

TheaperDeng:
### 1. Why the change?
Add a new package to bigdl called `bigdl.llm`

### 2. User API changes
nothing, still work in progress.

### 3. Summary of the change 
add initial setuptools and init files


rnwang04 (approved):
LGTM

PR #8226, opened 2023-05-25
[ORCA] Replace Optimizer_weights for Upgrading to TensorFlow 2.12

sgwhat:
## Description
TensorFlow version needs to be upgraded to 2.12.0, meanwhile there exists breaking changes related to `tf.keras.optimizer`. 

### 1. Why the change?
In 2.11.0, `tf.keras.optimizer` has a different set of public APIs from the old optimizer (breaking changes), so we need to support the new usage for TensorFlow version upgrading.

### 2. User API changes

N/A

### 4. How to test?
- [x] Unit test
- [x] Continuous training check

### DIFF
--- a/python/orca/src/bigdl/orca/learn/tf2/spark_runner.py
+++ b/python/orca/src/bigdl/orca/learn/tf2/spark_runner.py
@@ -271,7 +271,8 @@ class SparkRunner:
                                         zero_grads = [tf.zeros_like(w) for w in grad_vars]
                                         self.model.optimizer.apply_gradients(
                                             zip(zero_grads, grad_vars))
-                                        self.model.optimizer.set_weights(self.optimizer_weights.value)
+                                        self.model.optimizer.set_weights(
+                                            self.optimizer_weights.value)
                                     self.strategy.run(build_optimizer)
                                 if self.model_weights:
                                     self.model.set_weights(self.model_weights.value)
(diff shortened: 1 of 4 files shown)


python/orca/src/bigdl/orca/learn/tf2/spark_runner.py
```diff
@@ -265,15 +265,14 @@ def __init__(self,
...
+                                if self.model_variables:
+                                    def build_optimizer():
                                         grad_vars = self.model.trainable_weights
                                         zero_grads = [tf.zeros_like(w) for w in grad_vars]
                                         self.model.optimizer.apply_gradients(
                                             zip(zero_grads, grad_vars))
-                                        self.model.optimizer.set_weights(
-                                            self.optimizer_weights.value)
-                                    self.strategy.run(load_opt_weights)
+                                        self.model.optimizer.build(self.model_variables)
```
hkvision, line 274: is this compatible with older tf versions?

hkvision:
For the test, I think we can modify 1 or 2 github actions to tf 2.12 and keep the others tf 2.9 so that we can test both versions.

python/orca/src/bigdl/orca/learn/tf2/spark_runner.py
sgwhat: nope, fixing this part.

sgwhat:
Passed UT, some failed but not caused by this PR, also passed training and inference in https://github.com/intel/BigDL-Recommender-System-Toolkit/blob/main/DEVCATALOG.md.

lalalapotter (approved):

PR #8227, opened 2023-05-25
[PPML] Fix PPML Decryption error caused by specified partition size

hzjane:
## Description
Fix ppml encrypt bug

### DIFF
--- a/scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/utils/Encrypt.scala
+++ b/scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/utils/Encrypt.scala
@@ -18,8 +18,12 @@ package com.intel.analytics.bigdl.ppml.utils
 
 import com.intel.analytics.bigdl.ppml.utils.Supportive
 import com.intel.analytics.bigdl.ppml.PPMLContext
-import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.{Dataset, Row, SparkSession}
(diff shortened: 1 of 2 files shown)


qiyuangong (approved):
LGTM

scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/utils/Encrypt.scala
```diff
@@ -101,6 +104,9 @@ object Encrypt extends Supportive {
       opt[String]('h', "header")
         .action((x, c) => c.copy(header = x))
         .text("whether to write header to the csv file, default is false")
+      opt[Int]('p', "partition")
```
qiuxin2012, line 107: I think you can provide a default partition number, like ( file size / the HDFS block size) + 1

scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/utils/Encrypt.scala
```diff
@@ -109,7 +115,8 @@ case class EncryptArguments(inputDataSourcePath: String = "input_data_path",
                             cryptoMode: String = "aes/cbc/pkcs5padding",
                             dataSourceType: String = "csv",
                             action: String = "encrypt",
-                            header: String = "false"
+                            header: String = "false",
+                            nPartition: Int = 4
```
qiuxin2012, line 119: 4 is not good default number, I think you can set it 0. When this nPartition < 1, you can detect the partition number automaticly ( file size / the HDFS block size) + 1.

PR #8236, opened 2023-05-30
LLM: Support package/quantize for llama.cpp/redpajama.cpp on Windows

rnwang04:
## Description

Initial support of llama.cpp on Windows by uploading manually llama.dll and quantize-llama.exe and downloading them in `setup.py`, as well as support `quantize` on Windows.

Files need to be uploaded for Windows :
- [x] llama.dll
- [x] quantize-llama.exe
- [x] gptneox.dll
- [x] quantize-gptneox.exe
- [ ] bloomz.dll
- [ ] quantize-bloomz.exe
image

bloomz will be added in the future as bloomz.cpp don't have Windows support now.
Linux support will be refactored in the future.

### 1. Why the change?
To support these cpp repo on Windows without exposing these repo public.

### 2. User API changes
package for Windows:
`python setup.py bdist_wheel`
package for Linux:
`python setup-linux.py bdist_wheel`
quantization for Windows is the same as Linux
```python
from bigdl.llm.ggml import quantize

quantize(input_path=r".\merged_gpt4all_llama_7b\ggml-model-f16.bin",
         output_path=r".\merged_gpt4all_llama_7b\ggml-model-q4_0.bin",
         model_family="llama",
         dtype="q4_0")
```

### 3. Summary of the change 

- new setup.py for Windows
- Initial support of llama.cpp on Windows by uploading manually llama.dll and quantize-llama.exe and download them in setup.py
- support quantize on Windows
- move all .exe/.dll/.lib files to libs directory
- update llama-cpp-python's CMakeFiles.txt to obtain `quantize-llama.exe`
- update redpajama.cpp's CMakeFiles.txt and related pybinding's CMakeFiles.txt to obtain `quantize-gptneox.exe`

### 4. How to test?
- [x] Unit test
- [x] Local test

### DIFF
--- a/python/llm/setup.py
+++ b/python/llm/setup.py
@@ -48,6 +48,7 @@ def get_llm_packages():
 lib_urls = [
     "https://sourceforge.net/projects/analytics-zoo/files/bigdl-llm/llama.dll",
     "https://sourceforge.net/projects/analytics-zoo/files/bigdl-llm/quantize-llama.exe",
+    "https://sourceforge.net/projects/analytics-zoo/files/bigdl-llm/gptneox.dll",
     # TODO: add bloomz and redpajama
 ]
 
@@ -66,7 +67,8 @@ def setup_package():
     
     package_data = [
         "libs/llama.dll",
-        "libs/quantize-llama.exe"
+        "libs/quantize-llama.exe",
+        "libs/gptneox.dll",
     ]
 
     for url in lib_urls:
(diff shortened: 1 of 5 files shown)


TheaperDeng (approved):
We will merge this PR for quick iter

python/llm/setup.py
```diff
@@ -44,7 +45,36 @@ def get_llm_packages():
...
+    if not os.path.exists(libso_file):
+        urllib.request.urlretrieve(url, libso_file)
+
+
 def setup_package():
+    
+    package_data = [
+        "libs/llama.dll",
+        "libs/quantize-llama.exe",
+        "libs/gptneox.dll",
```
TheaperDeng, line 72: are we missing `quantize-gptneox.exe` here?

rnwang04: yeah it's missing here

PR #8242, opened 2023-06-01
[PPML] Add 01.org & key to fix image build failed

hzjane:
## Description
Add 01.org & key to fix image build failed


gc-fu (approved):
LGTM

PR #8243, opened 2023-06-01
[LLM] Enable UT workflow logics for LLM

HensonMa:
## Description

Enable UT workflow logics for LLM by a naive pass test.

### DIFF
--- a/python/llm/test/basics/test_install.py
+++ b/python/llm/test/basics/test_install.py
@@ -1,0 +1,13 @@
+import bigdl.llm
+import pytest
+from unittest import TestCase
+
+
+class Test_LLM_Basics(TestCase):
+
+    def test_naive(self):
+        pass
+
+
+if __name__ == '__main__':
+    pytest.main([__file__])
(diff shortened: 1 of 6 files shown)


python/llm/test/basics/test_install.py
```diff
@@ -0,0 +1,30 @@
+#
```
TheaperDeng, line 1: change the dir name to python/llm/test/packaging

python/llm/test/basics/test_install.py
```diff
@@ -0,0 +1,30 @@
...
+
+import bigdl.llm
+import pytest
+from unittest import TestCase
+
+
+class Test_LLM_Basics(TestCase):
+
+    def test_naive(self):
+        pass
```
TheaperDeng, line 26: try to import something here

python/llm/test/basics/test_install.py
HensonMa: resolved.

python/llm/test/basics/test_install.py
HensonMa: resolved.

TheaperDeng (approved):

PR #8228, opened 2023-05-26
[llm] Initial support of package and quantize

rnwang04:
## Description

To support quantization and further inference, at first, we need involve theses dynamic library and executable file (libllama.so / quantize) into our llm package.
In this PR, try to obtain dynamic library and executable file of three cpp repos(llama.cpp / bloomz.cpp / redpajama.cpp) by skbuild + submodule on Linux.
Also support uniform quantization python API for llama.cpp / bloomz.cpp / redpajama.cpp.

### 1. Why the change?

To obtain dynamic library and executable file of three cpp repos, and provide an uniform  quantization API

### 2. User API changes

Basic usage of quantization is supported.
```python
from bigdl.llm.ggml import quantize

quantize(input_path="./merged_gpt4all_llama_7b/ggml-model-f16.bin",
         output_path="./merged_gpt4all_llama_7b/ggml-model-q4_0.bin",
         model_family="llama",
         dtype="q4_0")
```

### 3. Summary of the change 

- obtain dynamic library and executable file of three cpp repos(llama.cpp / bloomz.cpp / redpajama.cpp) by skbuild + submodule on Linux, finish related CMakeLists.txt  [**only for Linux now**]
```bash
/disk8/miniconda3/envs/test/lib/python3.9/site-packages/bigdl_llm-2.4.0.dev0-py3.9-linux-x86_64.egg/bigdl/llm$ ls bin
quantize-bloomz  quantize-gptneox  quantize-llama
/disk8/miniconda3/envs/test/lib/python3.9/site-packages/bigdl_llm-2.4.0.dev0-py3.9-linux-x86_64.egg/bigdl/llm$ ls lib
libbloom.so  libgptneox.so  libllama.so
```
- provide an uniform quantization API

### 4. How to test?
- [x] Unit test

PR #8244, opened 2023-06-01
[llm] Add convert_model api

ACupofAir:
## Description
combine #8235 and #8228  to one api `convert_model`

### 1. User API changes

Basic usage of convert a model to  int4-quantized model

```python
from bigdl.llm.ggml import convert_model
model_path = .'/models/vicuna-13b-1.1'
convert_model(model_path=model_path, output_path=None, model_family='llama')
```

### 2. Summary of the change 

* add `convert_model` api for bigdl.llm.ggml
* fix a path name replace problem of pathlib.Path in `quantize.py`

### 3. How to test?
successfully test on vicuna-13b-1.1

### DIFF
--- a/python/llm/src/bigdl/llm/ggml/convert_model.py
+++ b/python/llm/src/bigdl/llm/ggml/convert_model.py
@@ -1,0 +1,35 @@
+from bigdl.llm.ggml.convert import _convert_to_ggml
+from bigdl.llm.ggml.quantize import quantize
+from pathlib import Path
+import time
+
+
+def convert_model(model_path: str,
+                  output_path: str = None,
+                  model_family: str = 'llama',
+                  dtype: str = 'q4_0'):
+    """
+    Convert Hugging Face llama-like / gpt-neox-like / bloom-like model to lower precision
+
+    :param model_path: str, path of model, for example `./llama-7b-hf`.
+    :param output_path: Save path of output quantized model. Default to `None`.
+            If you don't specify this parameter, quantized model will be saved in
+            the same directory as the input and just replace precision with quantize_type
+            like `./ggml-model-q4_0.bin`.
+    :param model_family: Which model family your input model belongs to. Default to `llama`.
+            Now only `llama`/`bloomz`/`gptneox` are supported.
+    """
+
+    model_name = Path(model_path).stem
+    tmp_ggml_file_path = f'/tmp/{model_name}_{int(time.time())}'
+    _convert_to_ggml(model_path=model_path,
+                     outfile_dir=tmp_ggml_file_path,
+                     model_family=model_family,
+                     outtype="fp16")
+    
+    tmp_ggml_file_path = next(Path(tmp_ggml_file_path).iterdir())
+
+    quantize(input_path=tmp_ggml_file_path,
+             output_path=output_path,
+             model_family=model_family,
+             dtype=dtype)
--- a/python/llm/src/bigdl/llm/ggml/convert_model.py
+++ b/python/llm/src/bigdl/llm/ggml/convert_model.py
@@ -25,7 +25,6 @@ def convert_model(input_path: str,
     if dtype == 'int4':
         dtype = 'q4_0'
 
-
     model_name = Path(input_path).stem
     tmp_ggml_file_path = f'/tmp/{model_name}_{int(time.time())}'
     _convert_to_ggml(model_path=input_path,
--- a/python/llm/src/bigdl/llm/ggml/convert_model.py
+++ b/python/llm/src/bigdl/llm/ggml/convert_model.py
@@ -22,6 +22,7 @@ def convert_model(input_path: str,
             Now only int4 supported.
     """
 
+    dtype = dtype.lower()
     if dtype == 'int4':
         dtype = 'q4_0'
 
--- a/python/llm/src/bigdl/llm/ggml/convert_model.py
+++ b/python/llm/src/bigdl/llm/ggml/convert_model.py
@@ -4,14 +4,14 @@ from pathlib import Path
 import time
 
 
-def convert_model(model_path: str,
+def convert_model(input_path: str,
                   output_path: str = None,
                   model_family: str = 'llama',
                   dtype: str = 'q4_0'):
     """
     Convert Hugging Face llama-like / gpt-neox-like / bloom-like model to lower precision
 
-    :param model_path: str, path of model, for example `./llama-7b-hf`.
+    :param input_path: str, path of model, for example `./llama-7b-hf`.
     :param output_path: Save path of output quantized model. Default to `None`.
             If you don't specify this parameter, quantized model will be saved in
             the same directory as the input and just replace precision with quantize_type
@@ -20,9 +20,9 @@ def convert_model(model_path: str,
             Now only `llama`/`bloomz`/`gptneox` are supported.
     """
 
-    model_name = Path(model_path).stem
+    model_name = Path(input_path).stem
     tmp_ggml_file_path = f'/tmp/{model_name}_{int(time.time())}'
-    _convert_to_ggml(model_path=model_path,
+    _convert_to_ggml(model_path=input_path,
                      outfile_dir=tmp_ggml_file_path,
                      model_family=model_family,
                      outtype="fp16")
--- a/python/llm/src/bigdl/llm/ggml/convert_model.py
+++ b/python/llm/src/bigdl/llm/ggml/convert_model.py
@@ -5,8 +5,8 @@ import time
 
 
 def convert_model(input_path: str,
-                  output_path: str = None,
-                  model_family: str = 'llama',
+                  output_path: str,
+                  model_family: str,
                   dtype: str = 'int4'):
     """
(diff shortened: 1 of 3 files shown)


jason-dai:
> ```python
> from bigdl.llm.ggml import convert_model
> model_path = .'/models/vicuna-13b-1.1'
> convert_model(model_path=model_path, output_path=None, model_family='llama')
> ```

- Change `model_path` to `input_path`
- Do we want to have a default value for `model_family`? Since bigdl-llm is supposed to support general llm, a default value may not make sense @TheaperDeng

jason-dai:
And it's not clear to me what the expected input model format is; do we check if the input format is valid?

TheaperDeng:
We need to specify what 

> > ```python
> > from bigdl.llm.ggml import convert_model
> > model_path = .'/models/vicuna-13b-1.1'
> > convert_model(model_path=model_path, output_path=None, model_family='llama')
> > ```
> 
> * Change `model_path` to `input_path`
> * Do we want to have a default value for `model_family`? Since bigdl-llm is supposed to support general llm, a default value may not make sense
> * How does the user specify the output type?
> 
> In addition, we need to unify the `model_family` handling in bigdl-llm, including:
> 
> * Exact name of the families: e.g., people are using "gptneox" or "redpajama" in different places; maybe we should define these families in a central places
> * Is there a default model family - probably not
> * Check if the input model family is valid.
> 
> @TheaperDeng

- We will use `gptneox` instead of `redpajama`, currently the later one is only used as some branch-name or repo-name we build at very early stage, which will not be exposed to users. After https://github.com/intel-analytics/BigDL/pull/8246, I don't think we will see any `redpajama` in bigdl repo.
- We may leave the model family a required parameter for now. In next PR, we may directly infer the family from the model dir's config.json (if existed)
- We need to add detailed structure requirement in our comment **in this PR** (including the file structure for tokenizer, modelbin, and other files)

TheaperDeng:
what we support now is what users directly pull from huggingface hub.

Like these
https://huggingface.co/bigscience/bloomz-7b1/tree/main
https://huggingface.co/decapoda-research/llama-7b-hf/tree/main

jason-dai:
> what we support now is what users directly pull from huggingface hub.
> 
> Like these https://huggingface.co/bigscience/bloomz-7b1/tree/main https://huggingface.co/decapoda-research/llama-7b-hf/tree/main

Need to document how the input can be obtained (e.g., in the python doc, a model convert cli tool, etc.)

jason-dai:
> * We will use `gptneox` instead of `redpajama`, currently the later one is only used as some branch-name or repo-name we build at very early stage, which will not be exposed to users. After [[LLM]Support for linux package (llama, NeoX) & quantize (llama) #8246](https://github.com/intel-analytics/BigDL/pull/8246), I don't think we will see any `redpajama` in bigdl repo.

I think the command line tool currently uses `redpajama` @hoshibara 

We need to have a central utility to manage all the model families

python/llm/src/bigdl/llm/ggml/convert_model.py
```diff
@@ -4,14 +4,14 @@
...
-def convert_model(model_path: str,
+def convert_model(input_path: str,
                   output_path: str = None,
                   model_family: str = 'llama',
                   dtype: str = 'q4_0'):
     """
     Convert Hugging Face llama-like / gpt-neox-like / bloom-like model to lower precision
 
-    :param model_path: str, path of model, for example `./llama-7b-hf`.
+    :param input_path: str, path of model, for example `./llama-7b-hf`.
```
jason-dai, line 14: Need to document what is expected in the `input_path` folder, and how to obtain the folder for the user

ACupofAir: - I will add the document of input_path format in next pr, this pr  is being relied upon by other developers and should be merged as soon as possible. 
- I add an issue #8254  to track the document format problem.

python/llm/src/bigdl/llm/ggml/convert_model.py
```diff
@@ -0,0 +1,35 @@
+from bigdl.llm.ggml.convert import _convert_to_ggml
+from bigdl.llm.ggml.quantize import quantize
+from pathlib import Path
+import time
+
+
+def convert_model(input_path: str,
+                  output_path: str = None,
+                  model_family: str = 'llama',
+                  dtype: str = 'q4_0'):
```
jason-dai, line 10: `dtype` should be `int4` by default, which maps to `int4_0`;

python/llm/src/bigdl/llm/ggml/convert_model.py
```diff
@@ -0,0 +1,40 @@
...
+            If you don't specify this parameter, quantized model will be saved in
+            the same directory as the input and just replace precision with quantize_type
+            like `./ggml-model-q4_0.bin`.
+    :param model_family: Which model family your input model belongs to. Default to `llama`.
+            Now only `llama`/`bloomz`/`gptneox` are supported.
+    :param dtype: Which quantized precision will be converted
+            Now only int4 supported.
+    """
+
+    if dtype == 'int4':
```
jason-dai, line 25: first change `dtype` to lower case first

python/llm/src/bigdl/llm/ggml/convert_model.py
```diff
@@ -0,0 +1,40 @@
+from bigdl.llm.ggml.convert import _convert_to_ggml
+from bigdl.llm.ggml.quantize import quantize
+from pathlib import Path
+import time
+
+
+def convert_model(input_path: str,
+                  output_path: str = None,
+                  model_family: str = 'llama',
```
jason-dai, line 9: remove default value 'llama'

python/llm/src/bigdl/llm/ggml/convert_model.py
```diff
@@ -0,0 +1,40 @@
...
+    :param output_path: Save path of output quantized model. Default to `None`.
+            If you don't specify this parameter, quantized model will be saved in
+            the same directory as the input and just replace precision with quantize_type
+            like `./ggml-model-q4_0.bin`.
+    :param model_family: Which model family your input model belongs to. Default to `llama`.
+            Now only `llama`/`bloomz`/`gptneox` are supported.
+    :param dtype: Which quantized precision will be converted
+            Now only int4 supported.
+    """
+
```
jason-dai, line 24: you need to check if the input values are valid

python/llm/src/bigdl/llm/ggml/convert_model.py
```diff
@@ -0,0 +1,40 @@
...
+                  model_family: str = 'llama',
+                  dtype: str = 'int4'):
+    """
+    Convert Hugging Face llama-like / gpt-neox-like / bloom-like model to lower precision
+
+    :param input_path: str, path of model, for example `./llama-7b-hf`.
+    :param output_path: Save path of output quantized model. Default to `None`.
+            If you don't specify this parameter, quantized model will be saved in
+            the same directory as the input and just replace precision with quantize_type
+            like `./ggml-model-q4_0.bin`.
```
jason-dai, line 18: I don't think we should support `None`; we may support model meta info in future @TheaperDeng ?

python/llm/src/bigdl/llm/ggml/convert_model.py
ACupofAir: * The `input_path` has checked in `_convert_to_ggml` function
* The `dtype` has checked in `quantize` function

python/llm/src/bigdl/llm/ggml/convert_model.py
TheaperDeng: I will take this issue for input path format description

python/llm/src/bigdl/llm/ggml/convert_model.py
TheaperDeng: yeah a good way for future is to require a dir path here.

python/llm/src/bigdl/llm/ggml/convert_model.py
jason-dai: The guideline is always to check and report error to the user as early as possible

TheaperDeng (approved):
TODOs are recorded in https://github.com/intel-analytics/BigDL/issues/8254

TheaperDeng:
Some further improvement will be added in https://github.com/intel-analytics/BigDL/issues/8254. Merge this PR for now to speed-up some dependent work.

python/llm/src/bigdl/llm/ggml/quantize.py
```diff
@@ -38,7 +39,7 @@
                           "q8_0": 7}
 
 _quantize_type = {"llama": _llama_quantize_type,
-                  "bloomz": _bloomz_quantize_type,
+                  "bloom": _bloom_quantize_type,
                   "gptneox": _gptneox_quantize_type}
 
 _valid_types = set(list(_llama_quantize_type.keys()) + list(_bloomz_quantize_type.keys()) +
```
Oscilloscope98, line 45: `_bloomz_quantize_type.keys()` => `_bloom_quantize_type.keys()`
Seems that you forgot to change here

PR #8245, opened 2023-06-01
[PPML] Add Attestation Lib implemented in Python

xiangyuT:
## Description

Enable attestation in Python so that there is no need to install Java and other jars to do attestation in LLM, etc.


Uxito-Ada (approved):
lgtm

PR #8249, opened 2023-06-01
[ORCA] Add Github Action for TF 2.12 Upgrade

sgwhat:
## Description
Add a new action for TF 2.12, since we need to support TF 2.10 (also the previous version) and TF 2.12 in the meanwhile.

### How to test?
- [x] Unit test


sgwhat:
All test passed.

Le-Zheng (approved):

PR #8257, opened 2023-06-02
[LLM] Fix linux load libs for NeoX and llama

cyita:
## Description

Fix linux load libs for NeoX and llama.

### 1. Why the change?

The current code cannot load the correct libs on Linux platform.

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/python/llm/src/bigdl/llm/utils/utils.py
+++ b/python/llm/src/bigdl/llm/utils/utils.py
@@ -1,3 +1,19 @@
+#
+# Copyright 2016 The BigDL Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


python/llm/src/bigdl/llm/utils/utils.py
```diff
@@ -0,0 +1,32 @@
...
+import sys
+from bigdl.llm.utils.common import invalidInputError, invalidOperationError
+
+
+def get_avx_flags():
+    avx = ""
+    if sys.platform != "win32":
+        import subprocess
+        msg = subprocess.check_output(["lscpu"]).decode("utf-8")
+        if "avx512_vnni" in msg:
```
MeouSker77, line 26: we can support avx512 without avx512_vnni since this PR: <https://github.com/analytics-zoo/bloomz.cpp/commit/affa28b2cbfecc7d8d6feeb9b3a3fe35b1c7c070>

cyita: Any plans for llama and neox?

MeouSker77: llama and neox already support it too

cyita: Updated in https://github.com/intel-analytics/BigDL/pull/8266

PR #8260, opened 2023-06-02
[LLM] add convert's python deps for LLM

HensonMa:
Add quantize's required python deps for LLM (numpy, torch, transformers, and sentencepiece)

### DIFF
--- a/python/llm/setup.py
+++ b/python/llm/setup.py
@@ -42,7 +42,7 @@ BIGDL_PYTHON_HOME = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 VERSION = open(os.path.join(BIGDL_PYTHON_HOME, 'version.txt'), 'r').read().strip()
 llm_home = os.path.join(os.path.dirname(os.path.abspath(__file__)), "src")
 libs_dir = os.path.join(llm_home, "bigdl", "llm", "libs")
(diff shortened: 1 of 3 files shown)


rnwang04:
quantize don't need such deps but convert needs.
Not sure convert has any limitations for transformers? @plusbang

HensonMa:
> quantize don't need such deps but convert needs. Not sure convert has any limitations for transformers? @plusbang

Oh, yes, it's this file's python deps. https://github.com/intel-analytics/BigDL/blob/main/python/llm/src/bigdl/llm/utils/convert_util.py

let me change the name of the deps group

python/llm/setup.py
```diff
@@ -128,6 +129,10 @@ def setup_package():
 
     for url in lib_urls[platform_name]:
         download_libs(url, change_permission=change_permission)
+    
+    install_requires=[]
+    if "--all" in sys.argv:
+        install_requires += CONVERT_DEP
```
TheaperDeng, line 135: let's use `extra_requires` in metadata for this

HensonMa: resolved

plusbang:
> quantize don't need such deps but convert needs. Not sure convert has any limitations for transformers? @plusbang

yeah, dependency `transformers` is necessary during converting.

PR #8268, opened 2023-06-05
Upgrade ppml-base image to use ubuntu:22.04

gc-fu:
## Description

The chatllm requires ubuntu:22.04, so there it is.


gc-fu:
Task suspended, not needed anymore.

PR #8277, opened 2023-06-06
Orca: fix API compatibility for tf2.12 in ray backend Estimator.

lalalapotter:
## Description

This PR is to fix API compatibility in ray backend Estimator, including:
+  [tf.keras.Model.save(filepath, overwrite=True, save_format=None, **kwargs)](https://www.tensorflow.org/api_docs/python/tf/keras/Model#save)
+ [tf.keras.optimizers](https://www.tensorflow.org/api_docs/python/tf/keras/optimizers)

How to test?
- [x] Unit test
- [x] Application test


sgwhat (approved):
LGTM and we need to add a github action for ray ut.

PR #8281, opened 2023-06-06
[PPML] Refine pyspark simplequery and add to Occlum

hzjane:
## Description
Refine pyspark simplequery and add to Occlum


ShanSimu (approved):
LGTM

Issue #8261, opened 2023-06-03
bigdl-llm transformers-like API first round design


TheaperDeng:
## Load / Initialization
We will support a `AutoModelForCausalLM` (just same name as the one in `transformers` for now). The goal is to privide them a consitent way for `from_pretrained` method.

#### Option 1: Convert and Load
Users will make a 2-line call for this way.
1. Call `convert_model` to make a disk-to-disk convertion for huggingface model -> ggml low-precision model
2. Call `AutoModelForCausalLM.from_pretrained` to load the model that has been converted.

**Requirement for /path/to/huggingface/checkpoint will be illustrated in API Design**

```python
from bigdl.llm.transformers import AutoModelForCausalLM
from bigdl.llm.ggml import convert_model

convert_model(input_path="/path/to/huggingface/checkpoint",
              output_path="/path/to/low-precision/converted-model",
              model_family="xxxx")

llm = AutoModelForCausalLM.from_pretrained("/path/to/low-precision/converted-model", model_family="xxxx")
```
#### Option 2: Load directly from huggingface checkpoint
Users could also directly load their huggingface model checkpoint, and we will convert it for them.

```python
from bigdl.llm.transformers import AutoModelForCausalLM

llm = AutoModelForCausalLM.from_pretrained("/path/to/huggingface/checkpoint", model_family="xxxx", cache_dir="/path/to/low-precision/converted-model")
```

#### Option 3: Load directly from huggingface hub
Users could also directly load their huggingface hub name, and we will download and convert it for them.

```python
from bigdl.llm.transformers import AutoModelForCausalLM

llm = AutoModelForCausalLM.from_pretrained("/path/to/huggingface/hub", model_family="xxxx", cache_dir="/path/to/low-precision/converted-model")
```

### API Design
```python
from bigdl.llm.transformers import AutoModelForCausalLM

llm = AutoModelForCausalLM.from_pretrained(pretrained_model_name_or_path,
                                           model_family,
                                           dtype="int4",
                                           cache_dir=None)
'''
:param pretrained_model_name_or_path: We support 3 kinds of pretrained model checkpoint
1. path for huggingface checkpoint that are directly pulled from hugginface hub. This should be a dir path that contains: weight bin, tokenizer config, tokenizer.model(required for llama) and added_tokens.json(if applied). For lora fine tuned model, the path should be pointed to a merged weight.
[...]

PR #8283, opened 2023-06-06
LLM: Support `generate(max_new_tokens=...)`, `tokenize` and `decode` for transformers-like API

TheaperDeng:
### 1. Why the change?
Support `generate(max_new_toekens=...)`, `tokenize` and `decode` for transformers-like API
#8261 

### 2. User API changes
```python
from bigdl.llm.ggml import AutoModelForCausalLM

llm = AutoModelForCausalLM.from_pretrained("gpt4all-model-q4_0.bin", model_family="llama", n_threads=28)

tokens = llm.tokenize("Q: Tell me something about Intel. A:")
tokens_id = llm.generate(tokens, max_new_tokens=32)
print(llm.decode(tokens_id))
```

### 3. Summary of the change 
1. Support `max_new_tokens` by stop when the tokens are reached
2. support `tokenize` by an extra encoding logic
3. support decode by an extra decoding logic


cyita (approved):
LGTM

python/llm/src/bigdl/llm/ggml/model/gptneox/gptneox.py
```diff
@@ -225,7 +225,7 @@ def __init__(
         if self.verbose:
             print(gptneox_cpp.gptneox_print_system_info().decode("utf-8"), file=sys.stderr)
 
-    def tokenize(
+    def _tokenize(
```
cyita, line 228: Need to modify tokenize in gptneox.py -> _tokenize. Will fix it in the next PR.

PR #8284, opened 2023-06-07
LLM: add bloom dll/exe in setup

rnwang04:
## Description
As we have supported bloom in Windows and update related files to sourceforge, we can now add dll and exe of bloom into setup

### How to test?
- [x] Unit test


MeouSker77 (approved):
LGTM

PR #8289, opened 2023-06-07
Modify the backend script.

Hanyu-Jin:
## Description

As Frontend use TCP socket, Backend also needs to use the same method.

### 1. Why the change?

Fix the backend to use TCP socket.

### 2. User API changes

None.


hzjane (approved):
LGTM

PR #8294, opened 2023-06-08
[PPML] Support Gramine and Amber in python AttestationCLI

xiangyuT:
## Description

Support Gramine and Amber in python AttestationCLI


gc-fu (approved):
LGTM

PR #8296, opened 2023-06-08
LLM: add new readme as first version document

TheaperDeng:
### 1. Why the change?
lack of document and API description.

### 2. User API changes
Nothing

### 3. Summary of the change 
https://github.com/TheaperDeng/BigDL/blob/bigdl-llm-readme/python/llm/README.md

### DIFF
--- a/python/llm/README.md
+++ b/python/llm/README.md
@@ -10,7 +10,7 @@ Users could use `bigdl-llm` to
 Currently `bigdl-llm` has supported
 - Precision: INT4
 - Model Family: llama, gptneox, bloom
-- Platform: Linux, Windows
+- Platform: Ubuntu 20.04 or later, CentOS 7 or later, Windows 10/11
 - Device: CPU
 
 ## Installation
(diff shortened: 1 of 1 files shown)


python/llm/README.md
```diff
@@ -1,47 +1,129 @@
...
-Feel free to explore different options and experiment with the llama/gptneox/bloom models using
-llm-cli!
\ No newline at end of file
+#### langchain integration
+TODO
+
+## Examples
+We prepared several examples in https://github.com/intel-analytics/BigDL/tree/main/python/llm/example
+
+## Dynamic library BOM
```
jason-dai, line 119: Is this dependency on user machine?

TheaperDeng: This dependency is not needed for users. But I provide these information for compatibility reason. Users who are using lower version `GLIBC` may meet compatibility issue.

The version we use to build the dynamic lib and executable files are "as-low-as-possible" to increase the compatibility.

jason-dai: > This dependency is not needed for users. But I provide these information for compatibility reason. Users who are using lower version `GLIBC` may meet compatibility issue.
> 
> The version we use to build the dynamic lib and executable files are "as-low-as-possible" to increase the compatibility.

SO this is actually user system dependency - for linux users they need to install glibc v9+, yes? how about windows users?

TheaperDeng: Yeah, I will add this requirement. The upgrade of GLIBC could be really tricky so most of our users will use the default GLIBC version provided by their OS. In this case, I will state we support Windows 10/11 (10 not validated), Ubuntu 20.04 or later, CentOS 7 or later (bloom could not support CentOS).

rnwang04 (approved):
LGTM

PR #8297, opened 2023-06-08
LLM: avoid unnecessary `import torch` except converting process

plusbang:
## Description

Avoid unnecessary `import torch` except converting process.

### 1. Why the change?

To avoid segmentation fault when initiate a `Bloom` instance.

### 4. How to test?
- [x] Local test


MeouSker77 (approved):
LGTM

PR #8298, opened 2023-06-08
add convert-gptq-to-ggml.py to bigdl-llama

qiuxin2012:
## Description

add python script to convert gptq quantized 4bits llama model to ggml type.

### 1. Why the change?

add python script to convert gptq quantized 4bits llama  model to ggml type.
Usage:

### 2. User API changes

a new python script to convert gptq 4bits llama model to ggml type.
Usage:
```
convert-gptq-to-ggml.py llamaXXb-4bit.pt tokenizer.model out.bin
```
For example:
```
python convert-gptq-to-ggml.py vicuna-13B-1.1-GPTQ-4bit-128g/vicuna-13B-1.1-GPTQ-4bit-128g.compat.no-act-order.pt vicuna-13B-1.1-GPTQ-4bit-128g/tokenizer.model vicuna-13B-1.1-gptq-4bit-ggml.bin
```

### 3. Summary of the change 

add python script convert-gptq-to-ggml.py
The script provide a way to convert gptq quantized 4bits llama model to ggml type. But don't support act-order and safetensors.

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/python/llm/src/bigdl/llm/gptq/convert/convert-gptq-to-ggml.py
+++ b/python/llm/src/bigdl/llm/gptq/convert/convert-gptq-to-ggml.py
@@ -26,6 +26,7 @@ import struct
 import numpy as np
 import torch
 from sentencepiece import SentencePieceProcessor
+from bigdl.llm.utils.common.log4Error import invalidInputError
 
 if len(sys.argv) != 4:
     print("Usage: convert-gptq-to-ggml.py llamaXXb-4bit.pt tokenizer.model out.bin\n")
@@ -34,6 +35,8 @@ if len(sys.argv) != 4:
 fname_model = sys.argv[1]
 fname_tokenizer = sys.argv[2]
 dir_out = sys.argv[3]
+invalidInputError(fname_model.endswith(".pt"),
+    "only support pytorch's .pt format now.")
 
 model = torch.load(fname_model, map_location="cpu")
 
@@ -152,6 +155,7 @@ def convert_q4_with_groupsize(src_name, dst_name, permute=False):
 
     # Q4_1 does not support bias; good thing the bias is always all zeros.
     # assert not np.any(g_idx)
+    invalidInputError(np.all(g_idx[:-1] <= g_idx[1:]), "Act-order is not supported, please use a no act-order model.")
     if not np.any(zeros):
         ftype = 2  # Q4_0
     else:
--- a/python/llm/src/bigdl/llm/gptq/convert/convert-gptq-to-ggml.py
+++ b/python/llm/src/bigdl/llm/gptq/convert/convert-gptq-to-ggml.py
@@ -41,7 +41,7 @@ def write_header(fout, shape, dst_name, ftype_cur):
 def convert_non_q4(src_name, dst_name, model, fout):
     v = model[src_name]
     shape = v.shape
-    print("Processing non-Q4 variable: " + src_name +\
+    print("Processing non-Q4 variable: " + src_name +
           " with shape: ", shape, " and type: ", v.dtype)
     if len(shape) == 1:
         print("  Converting to float32")
@@ -160,7 +160,7 @@ def convert_gptq2ggml(model_path, tokenizer_path, output_path):
 
     n_vocab, n_embd = model['model.embed_tokens.weight'].shape
     n_layer = 1 + max(int(m.group(1)) for name in model
-                      if (m:=re.match(r'model\.layers\.([0-9]+)', name)))
+                      if (m := re.match(r'model\.layers\.([0-9]+)', name)))
 
     # hardcoded:
     n_mult = 256
--- a/python/llm/src/bigdl/llm/gptq/convert/convert-gptq-to-ggml.py
+++ b/python/llm/src/bigdl/llm/gptq/convert/convert-gptq-to-ggml.py
@@ -160,7 +160,7 @@ def convert_gptq2ggml(model_path, tokenizer_path, output_path):
 
     n_vocab, n_embd = model['model.embed_tokens.weight'].shape
     n_layer = 1 + max(int(m.group(1)) for name in model
-                      if m := re.match(r'model\.layers\.([0-9]+)', name))  # NOQA
+                      if m := re.match(r'model\.layers\.([0-9]+)', name))  # noqa
 
     # hardcoded:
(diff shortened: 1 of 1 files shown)


qiuxin2012:
TODO: add windows support

jason-dai:
I think we may need two command line tools for model conversion with the same arguments:
- `convert_pytorch_model` (or `convert_hf_model`?)
- `convert_gptq_model`

Or one `convert_model` command line tool to support both format? And where should we put the model conversion tool? @TheaperDeng

jason-dai:
> ### 2. User API changes
> a new python script

Please describe how it is used

> 
> ### 3. Summary of the change
> add python script convert-gptq-to-ggml.py The script provide a way to convert gptq quantized 4bits model to ggml type. But don't support act-order and safetensors.

Can you check if the input format is expected and report error if not?

jason-dai:
> > I think we may need two command line tools for model conversion with the same arguments:
> > 
> > * `convert_pytorch_model` (or `convert_hf_model`?)
> > * `convert_gptq_model`
> > 
> > Or one `convert_model` command line tool to support both format? And where should we put the model conversion tool? @TheaperDeng
> 
> We have a `convert_model` entry point now, I believe put it there with some extra parameters(args) should be suitable

What do you mean by `convert_model` entry point? Is it a script?

jason-dai:
> > I think we may need two command line tools for model conversion with the same arguments:
> > 
> > * `convert_pytorch_model` (or `convert_hf_model`?)
> > * `convert_gptq_model`
> > 
> > Or one `convert_model` command line tool to support both format? And where should we put the model conversion tool? @TheaperDeng
> 
> We have a `convert_model` entry point now, I believe put it there with some extra parameters(args) should be suitable

What do you mean by `convert_model` entry point? Is it a script?

python/llm/src/bigdl/llm/gptq/convert/convert-gptq-to-ggml.py
```diff
@@ -0,0 +1,237 @@
...
+import sys
+import json
+import struct
+import numpy as np
+import torch
+from sentencepiece import SentencePieceProcessor
+
+if len(sys.argv) != 4:
+    print("Usage: convert-gptq-to-ggml.py llamaXXb-4bit.pt tokenizer.model out.bin\n")
+    sys.exit(1)
```
jason-dai, line 32: Is this for llama only?

TheaperDeng:
> > > I think we may need two command line tools for model conversion with the same arguments:
> > > 
> > > * `convert_pytorch_model` (or `convert_hf_model`?)
> > > * `convert_gptq_model`
> > > 
> > > Or one `convert_model` command line tool to support both format? And where should we put the model conversion tool? @TheaperDeng
> > 
> > 
> > We have a `convert_model` entry point now, I believe put it there with some extra parameters(args) should be suitable
> 
> What do you mean by `convert_model` entry point? Is it a script?

Yes, it's a python script but could work like a cmd line tool:
https://github.com/intel-analytics/BigDL/blob/main/python/llm/setup.py#L179

It work like this:
https://github.com/TheaperDeng/BigDL/blob/85d6b243bea829b7f7c55c5dd723f8e5b9cecb3d/python/llm/README.md#convert-your-model

python/llm/src/bigdl/llm/gptq/convert/convert-gptq-to-ggml.py
```diff
@@ -0,0 +1,237 @@
...
+        for j in range(i, i + (32 // bits)):
+            zeros[:, j] = (qzeros[:, col] >> (bits * (j - i)) & (2 ** bits - 1)) + 1
+        i += 32 // bits
+        col += 1
+    return zeros
+
+def convert_q4_with_groupsize(src_name, dst_name, permute=False):
+    qzeros = model[f"{src_name}.qzeros"].numpy()
+    zeros = qzeros_to_zeros(qzeros).T
+    scales = model[f"{src_name}.scales"].numpy().T
```
jason-dai, line 149: Are we using fp16 or fp32 scale/zero? I think the latest llama.cpp use fp16 by default, and we have changed to fp32 in our implementation for q4_0. @MeouSker77

MeouSker77: yes, we have changed to fp32 for q4_0

jason-dai: we should probably test the latency & accuracy of fp32 vs. fp16

shane-huang:
IMHO, using one command `convert_model` with options like `pytorch` or `gptq` is better, as later we may support more model format, or if the format name is changed we don't have to change the command name. Btw, can we infer the format based on the model suffix or something? 

Another thing, shall we rename the command for conversion and running to `bigdl-llm-convert`, and `bigdl-llm`? That'll be easier to search for.

jason-dai:
> IMHO, using one command `convert_model` with options like `pytorch` or `gptq` is better, as we may support more model formats in future. Btw, can we infer the format based on the model suffix or something?
> 
> Another thing, shall we rename the command for conversion and running to `bigdl-llm-convert`, and `bigdl-llm`? That'll be easier to search for. Or maybe just using one command `bigdl-llm-cli` and use `bidl-llm-cli convert` for conversion and other options compatible with llama-cpp cli.
> 
> @jason-dai @TheaperDeng

I don't think one can easily infer the model format or family.

And it's probably simpler in concept to have a separate model conversion tool; we can call it `bigdl-llm-convert` but maybe the name is a bit long. Or we can just use `llm-convert` (and similar `llm-cli`)?

python/llm/src/bigdl/llm/gptq/convert/convert-gptq-to-ggml.py
qiuxin2012: yes

jason-dai: The we should ask the user always specify `model_family=llama` in the command line

qiuxin2012:
`:=` is new from python3.8, our style check is based on 3.7. So need to remove `:=` to solve the style check.

PR #8299, opened 2023-06-08
lightgbm and model encryption in PPMLContext

Uxito-Ada:
## Description

LightGBM model (classifier, ranker and regressor) encrypted IO on PPMLContext Spark

### 1. Why the change?

Previous model encryption in BigDL does not support LightGBM, which is required by customers

### 2. User API changes

add LightGBM model save and load methods in PPMLContext

### 3. Summary of the change 

as title

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [x] Application test
- [ ] Document test
- [ ] ...


Uxito-Ada:
Passed the test: first, train a classification model from base classifier, then save the trained model in ciphertext, final reload the encrypted model and use it to predict:
image

plusbang (approved):

Uxito-Ada:
Passed python example:
image

PR #8302, opened 2023-06-08
LLM: optimize namespace and remove unused import logic

plusbang:
## Description

Currently, when we inference with binary files, some unused dependencies are still needed because messy namespace.
In this PR, optimize namespace and import logic.

### 2. User API changes

Previous:
```python
from bigdl.llm.ggml.model.bloom import Bloom
```

Now:
```python
from bigdl.llm.models import Llama, Bloom, Gptneox
```

### 3. Summary of the change 

- internal implementation of `LazyImport`
- modify import path of `Llama`, `Bloom`, `Gptneox`
- change default value of `n_threads` to 2 (only temporary workaround)

After installing `bigdl-llm`, **only `typing_extensions` dependency is necessary** to inference using converted ggml binary checkpoint.

### 4. How to test?
- [x] Unit test
- [x] Local test

### DIFF
--- a/python/llm/src/bigdl/llm/models.py
+++ b/python/llm/src/bigdl/llm/models.py
@@ -19,5 +19,6 @@
 # Otherwise there would be module not found error in non-pip's setting as Python would
 # only search the first bigdl package and end up finding only one sub-package.
 
-from .log4Error import invalidInputError, invalidOperationError
-from .lazyimport import LazyImport
+from bigdl.llm.ggml.model.llama import Llama
+from bigdl.llm.ggml.model.gptneox import Gptneox
+from bigdl.llm.ggml.model.bloom import Bloom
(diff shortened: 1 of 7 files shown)


python/llm/src/bigdl/llm/models.py
```diff
@@ -0,0 +1,24 @@
...
+#
+
+# This would makes sure Python is aware there is more than one sub-package within bigdl,
+# physically located elsewhere.
+# Otherwise there would be module not found error in non-pip's setting as Python would
+# only search the first bigdl package and end up finding only one sub-package.
+
+from bigdl.llm.ggml.model.llama import Llama
+from bigdl.llm.ggml.model.gptneox import Gptneox
+from bigdl.llm.ggml.model.bloom import Bloom
```
TheaperDeng, line 24: @shane-huang please confirm these spelling and if we need to make some letter capital.

TheaperDeng (approved):
LGTM

python/llm/src/bigdl/llm/models.py
plusbang: > @shane-huang please confirm these spelling and if we need to make some letter capital.

@jason-dai Now, API is changed to
```python
from bigdl.llm.models import Llama, Bloom, Gptneox
```

PR #8303, opened 2023-06-09
Update MPI Estimator to support Pytorch IPEX training

jenniew:
Update MPI Estimator to support Pytorch IPEX training


python/orca/src/bigdl/orca/learn/mpi/mpi_estimator.py
```diff
@@ -139,8 +142,11 @@ def fit(self, data, epochs=1, batch_size=32, validation_data=None, validate_batc
                               validate_batch_size, train_func, validate_func, train_batches,
                               validate_batches, validate_steps), f)
         self.mpi_runner.scp_file("mpi_train_data.pkl", self.dir)
-        self.mpi_runner.run("{}/mpi_train.py".format(self.dir), pkl_path=self.dir)
-        self.mpi_runner.shutdown_plasma()
+        self.mpi_runner.run("{}/mpi_train.py".format(self.dir),
+                            mpi_options=mpi_options,
+                            pkl_path=self.dir)
+        if isinstance(data, DataFrame) or isinstance(data, RDD):
```
hkvision, line 148: You also need to add this if when launch_plasma

python/orca/src/bigdl/orca/learn/mpi/mpi_estimator.py
```diff
@@ -308,62 +313,137 @@ def create_plasma_dataloader(config):
...
-                + "loss {:.6f}, ".format(average_loss)
-            )
-            total_iter = 0
-            total_samp = 0
-
-        should_validate = valid_ld and ((validate_steps > 0 and (j + 1) % validate_steps == 0)
-                                        or (j + 1 == train_batches))
-        if should_validate:
-            validate_func(config, model, valid_ld, metrics, validate_batches)
+    step = 0
```
hkvision, line 325: @leonardozcm Take a look at this part.

python/orca/src/bigdl/orca/learn/mpi/mpi_runner.py
```diff
@@ -59,8 +59,10 @@ def get_ip(iter):
         print("Hosts: ", self.hosts)
         self.processes_per_node = processes_per_node
         self.env = env if env else {}
+        import getpass
```
hkvision, line 62: put this into the beginning with other imports?

python/orca/src/bigdl/orca/learn/mpi/mpi_estimator.py
```diff
@@ -308,62 +313,137 @@ def create_plasma_dataloader(config):
...
+    for i in range(epochs):
```
hkvision, line 326: Since you still use the for loop, why do you need to change train_loop? You can put step in config after train_loop?

python/orca/src/bigdl/orca/learn/mpi/mpi_estimator.py
jenniew: launch_plasma  is already under the condition if isinstance(data, DataFrame)

hkvision: ok. Then I think we can remove RDD, since RDD is not supported in fit methods of all estimators.

python/orca/src/bigdl/orca/learn/mpi/mpi_estimator.py
jenniew: what do you mean? step is to store current step. Use current step to check whether need to print/evaluate when print_frequency/val_step is met.

hkvision: I mean seems you can just keep train_epoch, why do you want users to override the whole train logic?

python/orca/src/bigdl/orca/learn/mpi/mpi_runner.py
jenniew: ok

python/orca/src/bigdl/orca/learn/mpi/mpi_estimator.py
jenniew: if users want to use huggging face trainer, we can not let them to pass their own train_epoch function because they cannot provide such train_epoch function if using huggingface trainer. We let them to pass whole train logic provide more flexibility to them. If they don't pass custom train function, they will use our default whole train loop.

python/orca/src/bigdl/orca/learn/mpi/mpi_estimator.py
jenniew: Actually after training, the input 'data' which was DataFrame has been changed to RDD. When we want to shutdown plasma,  the data is RDD. Will change to if isinstance(data, RDD): self.mpi_runner.shutdown_plasma()

hkvision: I don't think it is reasonable to change the original data format.

python/orca/src/bigdl/orca/learn/mpi/mpi_estimator.py
leonardozcm: alpaca is trained by batch I remember?

python/orca/src/bigdl/orca/learn/mpi/mpi_runner.py
```diff
@@ -85,7 +87,9 @@ def run(self, file, **kwargs):
             mpi_config.extend(["-genv", "OMP_NUM_THREADS={}".format(mpi_env["OMP_NUM_THREADS"])])
         if len(self.remote_hosts) > 0:
             mpi_config.extend(["-hosts", ",".join(self.hosts)])
-        cmd.extend(mpi_config)
+        if mpi_options:
+            mpi_config += mpi_options
```
leonardozcm, line 91: Will here be any duplicated options?

python/orca/src/bigdl/orca/learn/mpi/mpi_runner.py
```diff
@@ -17,6 +17,7 @@
 import os
 import sys
 import subprocess
+import getpass
```
leonardozcm, line 20: is this used?

python/orca/src/bigdl/orca/learn/mpi/mpi_estimator.py
jenniew: updated

python/orca/src/bigdl/orca/learn/mpi/mpi_runner.py
jenniew: If there're duplicated options, then the mpi_options will overwrite previous ones.

python/orca/src/bigdl/orca/learn/mpi/mpi_estimator.py
```diff
@@ -308,62 +313,137 @@ def create_plasma_dataloader(config):
...
+#         if previous_iteration_time:
+#             iteration_time = current_time - previous_iteration_time
+#         else:
+#             iteration_time = 0
+#         previous_iteration_time = current_time
+#         x, y = next(train_iter)
+#         o = model(x, y)
+#         l = loss(o, y)
+#         l_np = l.detach().cpu().numpy()
+#         y_np = y.detach().cpu().numpy()
```
leonardozcm, line 414: remove this if we don't need this

python/orca/src/bigdl/orca/learn/mpi/mpi_train.py
```diff
@@ -61,7 +61,10 @@
...
     validate_batches = None
 
-for i in range(epochs):
-    config["epoch"] = i
-    train_func(config, model, train_ld, train_batches, optimizer, loss, scheduler,
-               validate_func, valid_ld, metrics, validate_batches, validate_steps)
+# for i in range(epochs):
+#     config["epoch"] = i
+#     train_func(config, model, train_ld, train_batches, optimizer, loss, scheduler,
+#                validate_func, valid_ld, metrics, validate_batches, validate_steps)
```
leonardozcm, line 67: same as above

leonardozcm (commented):
and do we also need to update the doc?

python/orca/src/bigdl/orca/learn/mpi/mpi_estimator.py
jenniew: sure

jenniew:
> and do we also need to update the doc?

ok

leonardozcm (approved):
LGTM

PR #8304, opened 2023-06-09
LLM: fix convert path error of gptneox and bloom on windows

rnwang04:
## Description

Fix convert path error of gptneox and bloom on Windows which is found by validating examples.

### How to test?
- [x] Local test
- [x] Unit test


plusbang (approved):
LGTM

PR #8310, opened 2023-06-09
[LLM] Add unified default value for cli programs

hoshibara:
## Description

Now, the default parameter values depend on the respective cli programs, and we need to unify them so that users have a unified experience when running different models.

## Unified Parameters
* `threads`
* `n_predict`


plusbang (approved):
LGTM

PR #8313, opened 2023-06-12
Nano: fix bug of `@nano(use_ipex=True)`

plusbang:
## Description

To fix bug reported by @gc-fu : when use `@nano(use_ipex=True)` to accelerate training loop, the output model seems not trained

### 2. User API changes

None

### 4. How to test?
- [x] Unit test
- [x] Local test


TheaperDeng (approved):
LGTM

PR #8316, opened 2023-06-12
LLM: first add `_tokenize`, `detokenize` and `_generate` for bloom pybinding

plusbang:
## Description

Add `tokenize`, `detokenize` and `generate` methods for Bloom pybinding. And make API compatible with Llama and Gptneox. 
**For those unsupported parameters of Bloom, errors will be raised if users input them. The default behavior is to ignore these settings temporarily.**
(More enhancements are on the way.)

### 2. User API changes

- transformers-like API
```python
from bigdl.llm.ggml.transformers import AutoModelForCausalLM

llm = AutoModelForCausalLM.from_pretrained(pretrained_model_name_or_path=your_model_path,
                                           model_family='bloom')
tokens = llm.tokenize(b"Learning English is")
tokens_id = llm.generate(tokens, max_new_tokens=10)
output = llm.batch_decode(tokens_id)
print(output)
```
- directly call model
```python
from bigdl.llm.models import Bloom

llm = Bloom(your_model_path)
tokens = llm._tokenize(b"Learning English is")
for token in llm._generate(tokens):
    print(llm.detokenize([token]).decode("utf-8", errors="ignore"))
```

### 3. Summary of the change 

- internal implementation of `Bloom._tokenize`, `Bloom.detokenize` and `Bloom._generate`
- internal implementation of `Bloom._supported_call` and `Bloom._supported_generate`, in which unsupported parameters will be checked
- `__init__`, `__call__`, `_tokenize` and `_generate` are compatible with Llama and Gptneox.

### 4. How to test?
- [ ] Unit test
- [x] Local test

### DIFF
--- a/python/llm/src/bigdl/llm/ggml/model/bloom/bloom.py
+++ b/python/llm/src/bigdl/llm/ggml/model/bloom/bloom.py
@@ -302,7 +302,7 @@ class Bloom(GenerationMixin):
     def free(self):
         bloom_free(self.ctx)
 
-    def _tokenize(self, text: bytes, add_bos: bool = True) -> List[int]:
+    def _tokenize(self, text: bytes, add_bos: bool = False) -> List[int]:
         """Tokenize a string.
 
         Args:
@@ -315,7 +315,7 @@ class Bloom(GenerationMixin):
             A list of tokens.
         """
         invalidInputError(self.ctx is not None, "The attribute `ctx` of `Bloom` object is None.")
-        return bloom_tokenize(self.ctx, text, add_bos)
+        return bloom_tokenize(self.ctx, text, False)
 
     def detokenize(self, tokens: List[int]) -> bytes:
         """Detokenize a list of tokens.
@@ -327,10 +327,10 @@ class Bloom(GenerationMixin):
             The detokenized string.
         """
         invalidInputError(self.ctx is not None, "The attribute `ctx` of `Bloom` object is None.")
-        output = b""
+        output = ""
         for token in tokens:
             output += bloom_detokenize(self.ctx, token)
-        return output
+        return output.encode('utf-8')
 
     def forward(self, input_ids: List[int]) -> int:
         return bloom_forward(ctx=self.ctx,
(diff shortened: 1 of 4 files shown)


shane-huang:
Shall we also make the API compatible with Llama and Gptneox? langchain requires the same parameters in the __init__ and __call__ interfaces in all models. 

I know some of the parameters are not supported in Bloom right now. Actually many of them are not needed in many use cases. We just need to confirm for each parameter: 1) what is the default behavior if the parameter is not passed by user 2) what is the behavior if user passed a value (e.g. `lora_path` is not supported, but if user passed it in, we may need to raise errors).

python/llm/src/bigdl/llm/ggml/model/bloom/bloom.py
```diff
@@ -87,15 +109,73 @@ def __init__(self,
         self.ctx = bloom_load(bytes(model_path, encoding='utf-8'), n_ctx, n_threads)
         invalidInputError(self.ctx is not None, f"Failed to load model from {model_path}")
         self.n_ctx = n_ctx
+        self.n_parts = n_parts
+        self.n_gpu_layers = n_gpu_layers
+        self.f16_kv = f16_kv
         self.seed = seed
```
shane-huang, line 115: I compared the default values of the parameters in langchain integrations. https://github.com/intel-analytics/BigDL/blob/main/python/llm/src/bigdl/llm/langchain/llms/bigdlllm.py. Most of the values are the same as the default values here. 

However there're some differences:
- `seed = -1` is -1 working in bloom?  https://github.com/intel-analytics/BigDL/blob/main/python/llm/src/bigdl/llm/langchain/llms/bigdlllm.py#L98
- `n_gpu_layers = None` https://github.com/intel-analytics/BigDL/blob/main/python/llm/src/bigdl/llm/langchain/llms/bigdlllm.py#L120

Please kindly check which is the proper default values. Maybe

MeouSker77: > * `seed = -1` is -1 working in bloom?  https://github.com/intel-analytics/BigDL/blob/main/python/llm/src/bigdl/llm/langchain/llms/bigdlllm.py#L98

if `seed < 0`, bloom will use current timestamp as random seed

shane-huang: LGTM. I'll validate langchain bloom path after you merge this PR.

PR #8322, opened 2023-06-13
Fix missing import of ipex int8 model

rnwang04:
## Description

Fix missing `import intel_extension_for_pytorch` of ipex int8 model which caused very severe performance degradation of loaded model.

### How to test?
- [x] Unit test
- [x] Local test


TheaperDeng (approved):
LGTM

PR #8326, opened 2023-06-14
LLM: update `convert_model` to support int8

rnwang04:
## Description

Support int8 quantization in `convert_model`.

### 1. Why the change?

To support int8 quantization.

### 2. User API changes
dtype `int8` is valid for `convert_model`.
```python
from bigdl.llm.ggml import convert_model

output_path = convert_model(input_path=model_dir,
                            output_path="./quantized_gpt4all",
                            model_family="llama",
                            dtype='int8')
print(output_path)
```

### 3. Summary of the change 

- Support int8 quantization in `convert_model`

### 4. How to test?
- [x] Unit test


TheaperDeng (approved):
LGTM

PR #8328, opened 2023-06-14
[LLM] LLM windows daily test

cyita:
## Description

LLM windows cicd init.

- [x] Add new pytest.


.github/workflows/bigdl-llm-winlog.yml
```diff
@@ -0,0 +1,22 @@
...
+  workflow_dispatch:
+
+jobs:
+  bigdl-llm-win:
+    runs-on: [self-hosted, Gondolin, ubuntu-20.04-lts]
+    steps:
+      - name: Download log
+        run: |
+          wget ftp://10.112.231.51:8821/llm/win_llm_test.log --ftp-user=${{ secrets.FTP_USERNAME }} --ftp-password=${{ secrets.FTP_PASSWORD }}
+          cat win_llm_test.log
```
TheaperDeng, line 21: so basically we need to manual check if there is any error or not?

TheaperDeng (commented):
others LGTM

.github/workflows/bigdl-llm-winlog.yml
cyita: Yes.

Le-Zheng (approved):

PR #8330, opened 2023-06-15
[PPML] Add read and write aliyun OSS using jindo sdk readme

hzjane:
## Description
Add read and write aliyun OSS using jindo sdk readme


qiyuangong (approved):
LGTM

PR #8331, opened 2023-06-15
Update the version of NANO in dl serving.

Hanyu-Jin:
## Description

Update the version of NANO in dl serving.

### 1. Why the change?

Models are built under 2.3.0.

### 2. User API changes

None.


ShanSimu (approved):
lgtm

PR #8336, opened 2023-06-15
Remove python3-pip in apt.

Hanyu-Jin:
## Description
Remove python3-pip in apt.

### 1. Why the change?

Pip is already installed in base image.


xiangyuT (approved):
LGTM

PR #8342, opened 2023-06-15
LLM: fix match in setup.py

rnwang04:
## Description
Fuzzy search can cause incorrect downloading of `quantize-llama` when only `quantize-llama.exe` exists
Change to a file name hard match to avoid this issue.

### How to test?
- [x] Unit test


Oscilloscope98 (approved):
LGTM

PR #8346, opened 2023-06-15
LLM: add basic uts related to inference

plusbang:
## Description

Add basic uts related to inference (`Llama/Bloom/Gptneox.__call__`)

### 4. How to test?
- [x] Unit test


Oscilloscope98 (approved):
LGTM, we may need to add tests for outputs perplexity later

plusbang:
> LGTM, we may need to add tests for outputs perplexity later

yeah, we need to explore other method later instead of directly `assert` checking to detect text completion outputs.

PR #8350, opened 2023-06-16
Try to build Torchserve from source code instead of intalling it from pip.

Hanyu-Jin:
## Description

Try to build Torchserve from source code instead of intalling it from pip.

### 1. Why the change?

As several codes are modified in Backend, it's inconvenient to install it and modify in Dockerfile.

### 2. User API changes

None


xiangyuT (approved):
LGTM

PR #8355, opened 2023-06-19
LLM: fix typos in llm_convert

leonardozcm:
## Description

Fix typos in readme and add return values in llm-convert


shane-huang (approved):
LGTM

PR #8358, opened 2023-06-19
LLM: Fix ggml return value

leonardozcm:
## Description
ggml convert should just return its own value.
```python
        return ggml_convert_model(input_path=model,
                                  output_path=outfile,
                                  model_family=model_family,
                                  dtype=outtype,
                                  **_used_args,
                                  )
```


cyita (approved):
LGTM

PR #8370, opened 2023-06-20
llm lora support

cyita:
## Description

LLM convert supports Lora.

### 1. Why the change?

The user should merge the Lora with the base model before converting -> User can pass the base model and Lora id or path when converting.

### 2. User API changes

convert_model.py /d1/llm/models/llama-7b-hf --model-format pth --outfile /d1/llm/models/llama-7b-int4/ --model-family llama **--lora-id-or-path /d1/llm/models/gpt4all-lora** -p /d1/llm/models/temp_path

### 3. Summary of the change 

LLM convert supports Lora.

### 5. New dependencies

pip install peft

### DIFF
--- a/python/llm/src/bigdl/llm/utils/lora_util.py
+++ b/python/llm/src/bigdl/llm/utils/lora_util.py
@@ -1,0 +1,86 @@
+#
+# Copyright 2016 The BigDL Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# ===========================================================================
+#
+# This file is adapted from
+# https://github.com/tloen/alpaca-lora/blob/main/export_hf_checkpoint.py
+#
+# Copyright 2023 Rohan Taori, Ishaan Gulrajani, Tianyi Zhang, Yann Dubois, Xuechen Li
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


python/llm/README.md
```diff
@@ -58,6 +58,33 @@ llm_convert(model="/path/to/vicuna-13B-1.1-GPTQ-4bit-128g/",
             model_family="llama")
 ```
 
+#### Convert your model with lora adapter
+##### Precondition
+1. The base model should be in **the huggingface format**, you can use this [script](https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/convert_llama_weights_to_hf.py) from transformers to convert.
+2. The lora model should contains `adapter_config.json` and `adapter_model.bin`.
```
TheaperDeng, line 64: this precondition is only for llama, we may just remove it?

python/llm/README.md
```diff
@@ -58,6 +58,33 @@ llm_convert(model="/path/to/vicuna-13B-1.1-GPTQ-4bit-128g/",
...
+```
+Here is an example to use `llm_convert` python API.
+```python
+from bigdl.llm import llm_convert
+llm_convert(model="/path/to/llama-7b-hf/",
+            outfile="/path/to/llama-7b-int4/",
+            model_format="pth",
+            lora_id_or_path="/path/to/gpt4all-lora",
+            model_family="llama")
+```
```
TheaperDeng, line 86: This part could be a small paragraph in https://github.com/jason-dai/BigDL/tree/llm-readme/python/llm#cli-tools, with no detailed parameter explanation. (But please be sure to save the for later document migration on readthedoc).

python/llm/src/bigdl/llm/utils/lora_util.py
```diff
@@ -0,0 +1,75 @@
...
+    # merge weights - new merging method from peft
+    lora_model = lora_model.merge_and_unload()
+    lora_model.train(False)
+
+    lora_model.save_pretrained(output_path)
+    tokenizer.save_pretrained(output_path)
+    return output_path
+
+if __name__ == '__main__':
+    merge_and_export_lora_model("huggyllama/llama-7b", "tloen/alpaca-lora-7b", "./hf_ckpt")
```
TheaperDeng, line 75: maybe we could just remove this?

cyita:
bloom 7b1 + haonan-li/bactrian-zh-bloom-7b1-lora convert memory usage:
merge Lora: peak: ~15g
convert & quantize: 23g

python/llm/README.md
cyita: updated

python/llm/src/bigdl/llm/utils/lora_util.py
cyita: updated

PR #8371, opened 2023-06-20
Support directly quantizing huggingface transformers into 4bit format

yangw1234:
### 1. Why the change?

Zero effort supporting ggml 4bit quantization for any LLMs defined in transformers.
Zero effort supporting all transformers API.

### 2. User API changes

Users can use the `quantize_4bit` function to quantized a fp32 transformers model into a quantized model. 

`quantize_4bit` does not take any additional arguments except the original model and return the quantized model.

The returned quantized model is the same transformers model, which directly support all transformers. API.

```python
  model = LlamaForCausalLM.from_pretrained(model_path, torch_dtype=torch.float32)

  ###### Quantize model into ggml q4_0 format ######
  from bigdl.llm.transformers import quantize_4bit
  model = quantize_4bit(model)
  print(model)
  ###################################################
```

### 3. Summary of the change 

1. depends on adding `ggml_quantize_q4_0` and `ggml_compute_forward_mul_mat_q_fp32` API to libllama.so

### 4. How to test?
Unit tests will be added

### 5. New dependencies
None


jason-dai:
Is it possible to use the `transformer` like API?
```python
model = AutoModelForCausalLM.from_pretrained("model_path", load_in_4bit=True)
```

python/llm/src/bigdl/llm/transformers/quantize.py
```diff
@@ -0,0 +1,98 @@
+#
```
jason-dai, line 1: change to `convert.py`

python/llm/example/transformers_quantize_4bit.py
```diff
@@ -0,0 +1,38 @@
+#
```
jason-dai, line 1: change to transformer_4bit

python/llm/src/bigdl/llm/transformers/linear_4bit.py
```diff
@@ -0,0 +1,196 @@
...
+
+import torch
+import ctypes
+
+QK = 32
+QK = 64 # read this value from libllama.so
+scale_size_in_bytes = 4
+block_size_in_bytes = QK // 2 + scale_size_in_bytes
+
+def ggml_quantize(tensor: torch.Tensor):
```
jason-dai, line 62: change to `ggml_convert_int4`

python/llm/src/bigdl/llm/transformers/linear_4bit.py
```diff
@@ -0,0 +1,196 @@
+#
```
jason-dai, line 1: change to `linear_int4`?

yangw1234:
> Is it possible to use the `transformer` like API?
> 
> ```python
> model = AutoModelForCausalLM.from_pretrained("model_path", load_in_4bit=True)
> ```

Do you mean using 
1. providing our own `AutoModelForCausalLM`, like `bigdl.llm.transformers.AutoModelForCausalLM.from_pretrained("model_path", load_in_4bit=True)`,
2.  or still letting user use `transformers.AutoModelForCausalLM.from_pretrained("model_path", load_in_4bit=True)` but changing its behaviors to using `bigdl.llm` instead of `bitsandbytes`.

I think the former is certainly easy to do. 

The latter would be hard unless we let user install our version of `bitsandbytes`, because the 4bit/8bit logic is hardcoded (in many places) with `bitsandbytes` in `transformers`.

yangw1234:
> > Is it possible to use the `transformer` like API?
> > ```python
> > model = AutoModelForCausalLM.from_pretrained("model_path", load_in_4bit=True)
> > ```
> 
> Do you mean using
> 
> 1. providing our own `AutoModelForCausalLM`, like `bigdl.llm.transformers.AutoModelForCausalLM.from_pretrained("model_path", load_in_4bit=True)`,
> 2. or still letting user use `transformers.AutoModelForCausalLM.from_pretrained("model_path", load_in_4bit=True)` but changing its behaviors to using `bigdl.llm` instead of `bitsandbytes`.
> 
> I think the former is certainly easy to do.
> 
> The latter would be hard unless we let user install our version of `bitsandbytes`, because the 4bit/8bit logic is hardcoded (in many places) with `bitsandbytes` in `transformers`.

Or something like this in `bigdl.llm.__init__.py`, which I think is very tricky. I am not sure if it will work. 

```python
transformers.AutoModelForCausalLM.from_pretrained = bigdl.llm.transformers.from_pretrained
```

jason-dai:
We can providing our own `AutoModelForCausalLM`; potentially we can also do `patch_transformer` as we did for `patch_cuda` before, but not for now

python/llm/src/bigdl/llm/ggml/transformers/convert.py
```diff
@@ -0,0 +1,100 @@
...
+# limitations under the License.
+
+
+import torch
+import torch.nn as nn
+from accelerate import init_empty_weights
+from bigdl.llm.ggml.transformers.linear_int4 import LinearInt4, ParamsInt4
+import warnings
+
+def _replace_with_4bit_linear(model, modules_to_not_convert=None, current_key_name=None):
```
jason-dai, line 43: _replace_with_int4_linear

python/llm/example/transformers_int4.py
```diff
@@ -0,0 +1,38 @@
...
+
+if __name__ == '__main__':
+    model_path = 'decapoda-research/llama-7b-hf'
+    model = LlamaForCausalLM.from_pretrained(model_path, torch_dtype=torch.float32)
+
+    ###### Quantize model into ggml q4_0 format ######
+    from bigdl.llm.ggml.transformers import ggml_convert_int4
+    model = ggml_convert_int4(model)
+    print(model)
+    ###################################################
```
jason-dai, line 28: Change to:
```python
model = AutoModelForCausalLM.from_pretrained("model_path", ..., load_in_4bit=True)
```

yangw1234: > Change to:
> 
> ```python
> from bigdl.llm.transformers import AutoModelForCausalLM
> model = AutoModelForCausalLM.from_pretrained("model_path", ..., load_in_4bit=True)
> ```

We also have a `bigdl.llm.ggml.transformers.AutoModelForCausalLM`, should we merge these two classes?

jason-dai: Yes. I think it support three cases today:
1) load bigdl-llm converted INT4 models
2) load huggingface model and convert using bigdl-llm convert_model
3) load huggingface model and convert linear layers to INT4

We need to define the right APIs for different cases.

python/llm/src/bigdl/llm/transformers/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from .convert import ggml_convert_int4
+from .model import AutoModelForCausalLM, AutoModel
+from transformers import AutoTokenizer
```
jason-dai, line 19: do we need this? we should not unconditionally depends `transformer` - import it only when needed

python/llm/example/transformers_int4.py
```diff
@@ -0,0 +1,36 @@
...
+import torch
+import os
+from bigdl.llm.transformers import AutoModelForCausalLM
+from transformers import LlamaTokenizer
+
+if __name__ == '__main__':
+    model_path = 'decapoda-research/llama-7b-hf'
+
+    # load_in_4bit=True in bigdl.llm.transformers will convert the model into
+    # ggml int4 format, which can directly run on intel CPUs.
```
jason-dai, line 26: will convert the relevant layers in the model into int4 format

jason-dai:
Some minor comments; otherwise LGTM

jason-dai:
> ### 2. User API changes
> Users can use the `quantize_4bit` function to quantized a fp32 transformers model into a quantized model.
> 
> `quantize_4bit` does not take any additional arguments except the original model and return the quantized model.
> 
> The returned quantized model is the same transformers model, which directly support all transformers. API.
> 
> ```python
>   model = LlamaForCausalLM.from_pretrained(model_path, torch_dtype=torch.float32)
> 
>   ###### Quantize model into ggml q4_0 format ######
>   from bigdl.llm.transformers import quantize_4bit
>   model = quantize_4bit(model)
>   print(model)
>   ###################################################
> ```

Please update the API change description

rnwang04 (approved):
Have verified on server with latest so.
Now llama model is ok  with `AutoModelForCausalLM.from_pretrained(model_path, trust_remote_code=True, load_in_4bit=True)` but chatglm model will fail.

PR #8372, opened 2023-06-21
[PPML] refine notebook service

Uxito-Ada:
## Description

support more TEE modes in notebook service and expose the service through VIP

### 1. Why the change?

to refine trusted bigdata frontend in TEE e.g. SGX and TDX

### 2. User API changes

need to set notebookExternalIP in values.yaml, which is an VIP to use as service IP

### 3. Summary of the change 

as title

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [x] Application test
- [ ] Document test
- [ ] ...


xiangyuT (approved):
LGTM

PR #8378, opened 2023-06-21
LLM: fix `ModuleNotFoundError` when use llm-cli

plusbang:
## Description

When inference using llm-cli, there exists `ModuleNotFoundError` of some dependencies which are only necessary in converting.
Fix in this PR.

### 4. How to test?
- [ ] Unit test


leonardozcm (approved):
lgtm

PR #8382, opened 2023-06-21
LLM: raise warning instead of error when use unsupported parameters

plusbang:
## Description

Currently, some parameters are unsupported in bloom and starcoder. When users input these unsupported parameters, errors are raised. Raise warning messages are better.

So replace errors with warnings.

### 4. How to test?
- [ ] Unit test


rnwang04 (approved):
LGTM

PR #8389, opened 2023-06-25
transformer api refactor

shane-huang:
## Description

Separate the partial optimization and complete native optimization in transformer API. Specifically,
- `AutoModelForCasualLM` for partial optimization models (bitesandbytes way)
- `BigdlForCasualLM` for complete native optimization models
 
In this PR, 

- move namespace `bigdl.llm.transformers.ggml` to `bigdl.llm.transformers`
- add `BigdlForCasualLM` interface to load ggml format only 
- merged updates from Yang's PR (#8371 )  
- revised README transformer API usage
- revised transformer API example 

### 1. Why the change?

Original API is a bit confusing to user. 

### 2. User API changes

`bigdl.llm.transformers.BigdlForCasualLM`
`bigdl.llm.transformers.AutoModelForCasualLM`


python/llm/example/transformers/int4_pipeline.py
```diff
@@ -62,27 +45,27 @@ def inference(llm, repo_id_or_model_path, model_family, prompt):
...
         # due to its out-of-date 'tokenizer_class' defined in its tokenizer_config.json.
         #
         # for most cases, you could use `AutoTokenizer`.
-        if model_family == 'llama':
-            from transformers import LlamaTokenizer
-            tokenizer = LlamaTokenizer.from_pretrained(repo_id_or_model_path)
-        else:
-            from transformers import AutoTokenizer
-            tokenizer = AutoTokenizer.from_pretrained(repo_id_or_model_path)
+        # if model_family == 'llama':
```
shane-huang, line 53: 1. Shall we still demonstrate the way using huggingface tokenizer? User needs to specify two different paths for tokenizer and model. 
The usage will be like
```python
           from bigdl.llm.transformers import BigdlForCausalLM
               llm = BigdlForCausalLM.from_pretrained("/path/to/ggml/bin",...)
            from transformers import LlamaTokenizer
            tokenizer = LlamaTokenizer.from_pretrained('/path/to/huggingface/repo')

            tokens_id = tokenizer(prompt).input_ids
            output_tokens_id = llm.generate(tokens_id, max_new_tokens=32)
            output = tokenizer.batch_decode(output_tokens_id)            
```

2. Shall we separate conversion and load/inference in the example, or just remove the convert and write it in README? Support the user will convert once and load/inference many times?

3. Do we still support download and loading pth/gptq models directly from huggingface using `BigdlForCausalLM`? Will it be confusing? Currently I removed them so user has to explicitly convert the model and then use `BigdlForCausalLM` to load it. 

@jason-dai

PR #8395, opened 2023-06-26
LLM: Check GPTQ outfile directory

leonardozcm:
## Description

To align with huggingface API, in case outfile do not exist, makedir manually.


plusbang (approved):
LGTM, but please wait tests pass.

PR #8400, opened 2023-06-26
Optimize transformer int4 loading memory

yangw1234:
## Description
Optimize transformer int4 loading memory

### DIFF
--- a/python/llm/example/transformers_int4.py
+++ b/python/llm/example/transformers_int4.py
@@ -32,7 +32,7 @@ if __name__ == '__main__':
     if model_path == 'decapoda-research/llama-7b-hf':
         # load_in_4bit=True in bigdl.llm.transformers will convert
         # the relevant layers in the model into int4 format
-        model = AutoModelForCausalLM.from_pretrained(model_path, load_in_4bit=True)
+        model = AutoModelForCausalLM.from_pretrained(model_path, load_in_4bit=True, low_cpu_mem_usage=True)
         tokenizer = LlamaTokenizer.from_pretrained(model_path)
 
         input_str = "Once upon a time, there existed a little girl who liked to have adventures. She wanted to go to places and meet new people, and have fun"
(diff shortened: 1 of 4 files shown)


python/llm/example/transformers_int4.py
```diff
@@ -46,7 +46,8 @@
         print(output_str)
         print(f'Inference time: {end-st} s')
     elif model_path == 'THUDM/chatglm-6b':
-        model = AutoModel.from_pretrained(model_path, trust_remote_code=True, load_in_4bit=True)
+        model = AutoModel.from_pretrained(model_path, trust_remote_code=True, load_in_4bit=True, low_cpu_mem_usage=True)
+        model.to(torch.float32)
```
yangw1234, line 50: directly using ChatGLM in transformers on cpu also requires `model.to(torch.float32)` or `model.to(torch.bfloat16)`

jason-dai: Can we automatically cast non-INT4 layers to FP32 in our conversion? For BF16, can we use auto mixed precision?

yangw1234: Yes.
And using BF16 actually perform worse than fp32 in this case.

python/llm/example/transformers_int4.py
```diff
@@ -32,7 +32,7 @@
     if model_path == 'decapoda-research/llama-7b-hf':
         # load_in_4bit=True in bigdl.llm.transformers will convert
         # the relevant layers in the model into int4 format
-        model = AutoModelForCausalLM.from_pretrained(model_path, load_in_4bit=True)
+        model = AutoModelForCausalLM.from_pretrained(model_path, load_in_4bit=True, low_cpu_mem_usage=True)
```
jason-dai, line 35: do we still need `low_cpu_mem_usage`?

yangw1234: Yes, otherwise transformer will use twice as much memory to load the model.

We can automatically set that value to true, but this will require user to install HF `accelerate`. Do you want it to be set automatically?

jason-dai: shall we add HF `accelerate` to our dependency? @TheaperDeng @shane-huang

yangw1234: @jason-dai 

`accelerate` (about 200k in size) is already in the dependencies. `low_cpu_mem_usage` will be set automatically when user sets `load_in_4bit`

jason-dai:
LGTM

Le-Zheng (approved):

PR #8410, opened 2023-06-28
Compile model archiver.

Hanyu-Jin:
## Description

Compile model archiver.

### 1. Why the change?

Model encryption is added to model archiver.

### 2. User API changes

None.


gc-fu (approved):
lgtm

PR #8412, opened 2023-06-29
Fix pentest issues

cyita:
## Description

Fix pentest issues.

### 3. Summary of the change 

Improper Resource Shutdown or Release of Resource

- [x] \BigDL-main\scala\friesian\src\main\java\com\intel\analytics\bigdl\friesian\serving\utils\EncodeUtils.java:27
- [ ] \BigDL-main\scala\orca\src\main\java\com\intel\analytics\bigdl\orca\tf\TFNetNative.java:136 -- cannot solve since this function returns the resource. 
- [x] \BigDL-main\scala\friesian\src\main\java\com\intel\analytics\bigdl\friesian\serving\utils\EncodeUtils.java:45

Stack Trace Exception Information Disclosure


glorysdj (approved):
LGTM

PR #8414, opened 2023-06-29
Pentest security issue: PPML - Key Exposed

Romanticoseu:
Change the encryption algorithm in the genrsa process from weak DES3 encryption to AES256, and increase the key length from 2048 to 3072.


Uxito-Ada:
pls verify this by mouting it as k8s secret for SparkPi ssl key

PR #8418, opened 2023-06-29
Orca: enable SSL certificate verification.

lalalapotter:
## Description

This PR is to enable SSL certificate verification when we connect the s3 file system for file operations.

### DIFF
--- a/python/orca/src/bigdl/orca/data/file.py
+++ b/python/orca/src/bigdl/orca/data/file.py
@@ -52,7 +52,7 @@ def open_text(path: str) -> List[str]:
         import boto3
         s3_client = boto3.Session(
             aws_access_key_id=access_key_id,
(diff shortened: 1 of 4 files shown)


hkvision (approved):

lalalapotter:
If `verify` is `False`, SSL verification will be disabled. If `verify` is not set to `False`, may be `None` or file path of CA bundle. When it is set to the default value `None`, CA bundle will be found in following configurations:

1. The environment variable AWS_CA_BUNDLE
2. Configuration in the ~/.aws/config file:
```
[default]
ca_bundle = path/to/cabundle.pem
```

Therefore, here we just delete the `verify` argument, and enable the SSL verification.

qiyuangong (approved):
LGTM

PR #8421, opened 2023-06-29
FIx Command Injection issues in Bigdl

leonardozcm:
## Description

As pentest recommended, we:
1. Validate and sanitize all parameters that the user provides to the application
2. By passing the command and its arguments as a list, you can ensure that the command string is not interpreted as a single command, thereby preventing command injection.

### DIFF
--- a/python/dllib/src/setup.py
+++ b/python/dllib/src/setup.py
@@ -20,6 +20,7 @@ import os
 import sys
 from shutil import copyfile, copytree, rmtree
 from setuptools import setup
+import subprocess
 
 long_description = '''
 BigDL DLlib is a distributed deep learning library for Apache Spark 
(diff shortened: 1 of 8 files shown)


qiyuangong:
Please fix https://github.com/intel-analytics/BigDL/actions/runs/5410872384/jobs/9832904701?pr=8421

python/dllib/src/setup.py
```diff
@@ -20,6 +20,7 @@
 import sys
 from shutil import copyfile, copytree, rmtree
 from setuptools import setup
+import subprocess
```
hkvision, line 23: why need this import?

qiyuangong (approved):
LGTM

PR #8445, opened 2023-07-03
Orca: remove hdfs cli.

lalalapotter:
## Description

This PR is to avoid using `hdfs` CLI in implementation. Because there may be servers that did not config or install `hdfs` CLI.


python/orca/src/bigdl/orca/data/file.py
```diff
@@ -572,10 +526,10 @@ def get_remote_dir_to_local(remote_dir: str, local_dir: str) -> int:
 def get_remote_files_with_prefix_to_local(remote_path_prefix: str,
                                           local_dir: str) -> Union[str, int]:
     prefix = os.path.basename(remote_path_prefix)
-    if remote_path_prefix.startswith("hdfs"):  # hdfs://url:port/file_path
-        cmd = 'hdfs dfs -get {}* {}'.format(remote_path_prefix, local_dir)
-        process = subprocess.Popen(cmd, shell=True)
-        return process.wait()
+    if remote_path_prefix.startswith("hdfs"):  # hdfs://(url:port/file_path
```
hkvision, line 529: this line change is wrong?

hkvision (approved):

PR #8447, opened 2023-07-03
Pentest fix: first fix command injection in `pickle.load()`

plusbang:
## Description

To fix this issue, add verification before deserializing.
- Add internal SafePickle in chronos module
- Fix `chronos/src/bigdl/chronos/autots/tspipeline.py` for example

### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/chronos/src/bigdl/chronos/utils.py
+++ b/python/chronos/src/bigdl/chronos/utils.py
@@ -91,3 +91,39 @@ class LazyImport:
         module = importlib.import_module(module_name, package=self.pkg)
         function = getattr(module, function_name)
         return function(*args, **kwargs)
+
+
+# Refer to this guide https://www.synopsys.com/blogs/software-security/python-pickling/
+# To safely use python pickle
+class SafePickle:
+    """
+    Example:
+        >>> from bigdl.chronos.utils import SafePickle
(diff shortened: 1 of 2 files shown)


python/chronos/src/bigdl/chronos/utils.py
```diff
@@ -91,3 +91,41 @@ def __call__(self, *args, **kwargs):
...
+    """
+    @classmethod
+    def dump(self, obj, file, *args, **kwargs):
+        import pickle
+        import hmac
+        import hashlib
+        pickle.dump(obj, file, *args, **kwargs)
+        try:
+            pickled_data = pickle.dumps(obj)
+            digest = hmac.new(b'shared-key', pickled_data, hashlib.sha1).hexdigest()
```
glorysdj, line 115: use a constant variable for SHARED_KEY?

plusbang: > use a constant variable for SHARED_KEY?

The SHARED_KEY is only used for HMAC generation, `pickled_data` is actually authenticated using HMAC. When data is changed, `digest` is different.

glorysdj (approved):
LGTM

PR #8448, opened 2023-07-03
[LLM] Change default runner for LLM Linux tests to the ones with AVX512

Oscilloscope98:
## Description

Added runner with AVX512 for LLM tests.

### 1. Why the change?

Added server runner with AVX512 as default to replace the Godolin runner (AVX2). Because AVX2 + sever is not our target users.

### 2. User API changes

N/A

### 3. Summary of the change 

- Change runners for inference related tests on Linux to the ones with AVX512 (on clx-018)
- Change runners for nightly convert tests to the ones with AVX512
- Optimize convert tests procedure to reduce peak disk memory
- Add job (runner) specific env variable `THREAD_NUM` for inference-related tests

### 4. How to test?
- [x] Unit test

### DIFF
--- a/.github/workflows/llm-nightly-test.yml
+++ b/.github/workflows/llm-nightly-test.yml
@@ -65,7 +65,7 @@ jobs:
         shell: bash
         run: |
           $CONDA_HOME/bin/conda env remove -y -n bigdl-init-llm || true
-          $CONDA_HOME/bin/conda create -n bigdl-init-llm -y python==3.9 setuptools==58.0.4 -c ${GONDOLIN_CONDA_CHANNEL} --override-channels
+          $CONDA_HOME/bin/conda create -n bigdl-init-llm -y python==3.9 setuptools==58.0.4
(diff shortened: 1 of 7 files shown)


.github/workflows/llm-nightly-test.yml
```diff
@@ -14,7 +14,7 @@ on:
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
   llm-nightly-convert-test:
-    runs-on: [ self-hosted, Gondolin, ubuntu-20.04-lts ]
+    runs-on: [ llm, AVX512, ubuntu-20.04-lts ]
```
hoshibara, line 17: Maybe `self-hosted` label need to be added.
[ref](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/using-self-hosted-runners-in-a-workflow)
other LGTM

hoshibara (approved):

PR #8449, opened 2023-07-03
LLM: Support AutoModelForSeq2SeqLM transformer API

leonardozcm:
## Description

Support AutoModelForSeq2SeqLM transformer API

```python
from bigdl.llm.transformers import AutoModelForSeq2SeqLM

model_name="/root/changmin/bart-base"
tokenizer = AutoTokenizer.from_pretrained(model_name)
model = AutoModelForSeq2SeqLM.from_pretrained(model_name, load_in_4bit=True)

inputs = tokenizer("BART model pre-trained on English language. It was introduced in the paper BART: Denoising Sequence-to-Sequence Pre-training for Natural Language Generation, Translation, and Comprehension by Lewis et al. and first released in this repository.", return_tensors="pt")
outputs = model.generate(**inputs)

summary = tokenizer.batch_decode(outputs,
                                 skip_special_tokens=True,
                                 spaces_between_special_tokens=False,)

print("Abstract: ", summary)
# Abstract: ['BART model pre-trained on English language. It was introduced in the paper BART']

```


shane-huang:
In transformer docs lots of models have their own model class. I'm afraid AutoXXX may not be able to load all of them. Maybe we should making transformer int4 general applicable to all transformer classes instead of adding classes one by one like this?

There're several possible options:
 
1) use class decorators
2) import (selected or all) model classes or detect the classes imported in scope, and use routine to automatically replace all of them

What do you think?

@yangw1234 @jason-dai

leonardozcm:
> How about add this as a new transformer example ?

That sounds good, for this bart model is small enough(500MB) for ut integration.

jason-dai:
> In transformer docs lots of models have their own model class. I'm afraid AutoXXX may not be able to load all of them. Maybe we should make transformer int4 general applicable to all transformer model classes instead of adding support one by one like this?
> 
> There're several possible options:
> 
> 1. use class decorators
> 2. import (selected or all) model classes or detect the classes imported in scope, and use routine to automatically replace all of them
> 
> What do you think?
> 
> @yangw1234 @jason-dai

Let's first verify more models and add more examples before adding additional classes.

jason-dai:
Is this ready for merge?

leonardozcm:
> Is this ready for merge?

Yes

leonardozcm:
> Is this ready for merge?

yes i think

shane-huang (approved):
you can merge after verifying the PR validation failure.

PR #8462, opened 2023-07-05
LLM: transformer int4 save and load

leonardozcm:
## Description

In this PR, we added support to save the results of the GGML int4 format conversion within the transformer API, which eliminates the need for redundant conversions in subsequent calls. However, please note that the **`ggml_matmul_src1_x_src0_t`** operation cannot currently be recorded in **`AutoConfig`**, and this operation is not included in the core Hugging Face infrastructure. Therefore, when using **`model.save_pretrained`**, it can only save the original PyTorch implementation of the model and cannot fully restore the converted model.

For loading the saved converted model, we forcefully modify the model's definition and the tensor shape of int4 weights without quantization, so that we can load the quantized model at last.

```python
from bigdl.llm.transformers import AutoModelForSeq2SeqLM

model_name="/root/changmin/bart-base"
# model_name="/root/changmin/BigDL/bart"
save_dir = "./bart"
tokenizer = AutoTokenizer.from_pretrained(model_name)
# Convert and save
model = AutoModelForSeq2SeqLM.from_pretrained(model_name, load_in_4bit=True)
model.save_pretrained(save_dir)
# Load back
model = AutoModelForSeq2SeqLM.from_pretrained(save_dir)

inputs = tokenizer("BART model pre-trained on English language. It was introduced in the paper BART: Denoising Sequence-to-Sequence Pre-training for Natural Language Generation, Translation, and Comprehension by Lewis et al. and first released in this repository.", return_tensors="pt")
outputs = model.generate(**inputs)

summary = tokenizer.batch_decode(outputs,
                                 skip_special_tokens=True,
                                 spaces_between_special_tokens=False,)

print("Abstract: ", summary)
# Abstract: ['BART model pre-trained on English language. It was introduced in the paper BART']

```

ggml int4 model will be marked as ggml_linear_int4_converted=True in config.json:
image

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/model.py
+++ b/python/llm/src/bigdl/llm/transformers/model.py
@@ -16,6 +16,7 @@
 
 import transformers
 import torch
+from .utils import extract_local_archive_file, load_state_dict, load
 
 
 class _BaseAutoModelClass:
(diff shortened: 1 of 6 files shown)


jason-dai:
> ```python
> # Save:
> model = AutoModelForSeq2SeqLM.from_pretrained(model_name, load_in_4bit=True, save_convert_pretrained=save_dir)
> ```

Change to 
```python
model = AutoModelForSeq2SeqLM.from_pretrained(model_name, load_in_4bit=True)
model.save_pretrained(...)
```

> ```python
> # Load
> model = AutoModelForSeq2SeqLM.from_pretrained(save_dir, load_convert_pretrained=True)
> ```

change to 
```python
model = AutoModelForSeq2SeqLM.from_pretrained(save_dir)
```

python/llm/src/bigdl/llm/transformers/model.py
```diff
@@ -29,12 +31,49 @@ def from_pretrained(cls,
...
+                                                      variant)
+            state_dict = load_state_dict(archive_file)
+            load(model, state_dict)
+            del state_dict
+        elif load_in_4bit:
             from .convert import ggml_convert_int4
             model = model.to("cpu")
             model = ggml_convert_int4(model)
+            model.config.update({"bigdl_linear_int4_converted": True,
+                                 "quantized": True})
```
jason-dai, line 75: why do we need "quantized"?

python/llm/src/bigdl/llm/transformers/model.py
```diff
@@ -29,12 +31,49 @@ def from_pretrained(cls,
...
+                                                      subfolder,
+                                                      variant)
+            state_dict = load_state_dict(archive_file)
+            load(model, state_dict)
+            del state_dict
+        elif load_in_4bit:
             from .convert import ggml_convert_int4
             model = model.to("cpu")
             model = ggml_convert_int4(model)
+            model.config.update({"bigdl_linear_int4_converted": True,
```
jason-dai, line 74: change to “bigdl_transformers_int4”

python/llm/src/bigdl/llm/transformers/model.py
leonardozcm: This is not used, I will remove this.

jason-dai:
LGTM

hkvision (approved):

PR #8467, opened 2023-07-06
Disable mmap by default for better performance

MeouSker77:
## Description

Disable mmap by default for better performance


rnwang04 (approved):
LGTM

PR #8477, opened 2023-07-07
[LLM] Add transformers int4 example for MOSS

Oscilloscope98:
## Description

Add transformers int4 examples for MOSS 16B.

On SPR-01,  ['nlp/moss-moon-003-sft'](https://huggingface.co/fnlp/moss-moon-003-sft) loaded in 4 bit can bring ~4.5x speed-up except the first token. And both English & Chinese outputs are supported well.

On client machine, due the large model size (~32G), it may take long time for loading the model 'fnlp/moss-moon-003-sft' in 4 bit. Maybe it would be more suitable to run 'fnlp/moss-moon-003-sft' on client machine after the saving and loading of transformers int4 models function is supported.

Detailed performance for MOSS: https://github.com/analytics-zoo/nano/issues/386#issuecomment-1623438837


plusbang (approved):
LGTM

shane-huang:
let's separate the models into different folders and add built-in prompts if necessary.

shane-huang (commented):
let's separate the models into different folders and add built-in prompts if necessary.

Oscilloscope98:
This will be added into #8498 leter

PR #8479, opened 2023-07-07
LLM: fix inconsistency between output token number and `max_new_token`

plusbang:
## Description

When run native int4 pipeline, the actual output token number is always 1 more than `max_new_token` parameter.
In this PR, fix it.

### 2. User API changes

None

### 4. How to test?
- [x] Unit test
- [x] Local test


cyita (approved):
LGTM

PR #8480, opened 2023-07-09
Add bigdl llm sgx image

gc-fu:
### 1. Why the change?
Add SGX image for bigdl-llm.

### 2. How to test?
- [x] Manually tested


glorysdj (approved):
LGTM

PR #8484, opened 2023-07-10
Add readme to deprecate trusted-big-data-ml

liu-shaojun:
## Description

Add readme to deprecate trusted-big-data-ml, trusted-big-data-ml folder should be deleted in next release.


glorysdj (approved):
LGTM

PR #8485, opened 2023-07-10
[LLM] Add workflow_dispatch for llm unittest workflow

hoshibara:
## Description

Some workflows of LLM lack the function of manual triggering, which needs to be added for testing.


Oscilloscope98 (approved):
LGTM

PR #8491, opened 2023-07-10
[LLM] Initial Transformers example refactor

Oscilloscope98:
## Description

Separate transformers examples into `native_int4` / `transformers_int4` folder, so that we could:
1. Add per-model examples into it transformers_int4 later
2. Avoid circular import error caused by folder named `transformers`


plusbang (approved):
LGTM

PR #8497, opened 2023-07-11
Support vnni check

MeouSker77:
## Description

Add avx_vnni and avx512_vnni checker

## Usage

```python
from bigdl.llm.utils.isa_checker import check_avx_vnni
from bigdl.llm.utils.isa_checker import check_avx512_vnni

if check_avx_vnni():
    ...

if check_avx512_vnni():
    ...
```

### DIFF
--- a/python/llm/setup.py
+++ b/python/llm/setup.py
@@ -199,7 +199,7 @@ def setup_package():
     for url in lib_urls[platform_name]:
         download_libs(url, change_permission=change_permission)
 
-    all_requires = []
+    all_requires = ['py-cpuinfo']
     all_requires += CONVERT_DEP
 
     metadata = dict(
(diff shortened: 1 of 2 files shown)


MeouSker77:
I test it on i9 windows (support avx_vnni but doesn't support avx512_vnni), clx (support avx512_vnni but doesn't support avx_vnni) and spr (support both avx_vnni and avx512_vnni), all works correctly

MeouSker77:
According to 
image
<https://www.felixcloutier.com/x86/cpuid>

we can check avx-vnni by setting eax to 7, ecx to 1, then use cpuid instruction, if returned eax's bit4 is 1, it means avx-vnni is supported

jason-dai:
Can we also check avx (that is, avx1)?

MeouSker77:
> Can we also check avx (that is, avx1)?

added it

jason-dai:
LGTM

rnwang04 (approved):
LGTM

PR #8503, opened 2023-07-12
[LLM] Merge the llm.cpp build and the pypi release

hoshibara:
### 1. Why the change?

In the current version, the compilation process of llm.cpp has nothing to do with the release of the pypi package. When there are rapid iterations and compilation failures, the latest binaries may not be available for pypi package releases.
Combining the two ensures that the latest binaries must be used when releasing a pypi package.

### 3. Summary of the change 

* Create a new workflow `llm-pypi-release` for combine two steps
* Nightly build workflow will call `llm-pypi-release` to release pypi whl

### 4. How to test?
- [ ] Manually test


hoshibara:
llm.cpp build step has been merged.
In this workflow, we will use `artifact` to [storge built binary files](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts).
The `setup.py` is being modified.

liu-shaojun (approved):

PR #8504, opened 2023-07-12
Orca: Fix mxnet python shell blocked issue

leonardozcm:
## Description

using `shell=True`  will call out a python interaction terminal


leonardozcm:
image

hkvision (approved):
Need to remove example test from pr action before the merge.

PR #8505, opened 2023-07-12
LLM: fix chatglm2 output

plusbang:
## Description

Related to https://github.com/analytics-zoo/nano/issues/434
Because the output linear layer of chatglm2 is converted, the output quality becomes poor.

Add an extra argument `keep_in_fp32_modules` so that different models can choose specific layers not to convert. The default value is `[]` and this parameter only works when `load_in_4bit/load_in_low_bit/bigdl_transformers_low_bit=True`.

### 2. User API changes

```python
# chatglm2 for example
from bigdl.llm.transformers import AutoModel
model = AutoModel.from_pretrained(model_path, load_in_4bit=True, keep_in_fp32_modules=['output_layer'])
```

### 3. Summary of the change 

- Add an extra argument `keep_in_fp32_modules` in `ggml_convert_quant`
- modify example (btw, we are working to add more examples, the example for chatglm2 is temporary)

### 4. How to test?
- [ ] Unit test
- [x] Local test

### DIFF
--- a/python/llm/example/transformers/transformers_int4/transformers_int4_pipeline.py
+++ b/python/llm/example/transformers/transformers_int4/transformers_int4_pipeline.py
@@ -47,7 +47,7 @@ if __name__ == '__main__':
         print('Output:', output_str)
         print(f'Inference time: {end-st} s')
     elif model_path == 'THUDM/chatglm-6b':
-        model = AutoModel.from_pretrained(model_path, trust_remote_code=True, load_in_4bit=True)
+        model = AutoModel.from_pretrained(model_path, trust_remote_code=True, load_in_4bit=True, keep_in_fp32_modules=['output_layer'])
         tokenizer = AutoTokenizer.from_pretrained(model_path, trust_remote_code=True)
 
         input_str = "晚上睡不着应该怎么办"
--- a/python/llm/src/bigdl/llm/transformers/model.py
+++ b/python/llm/src/bigdl/llm/transformers/model.py
@@ -46,6 +46,13 @@ class _BaseAutoModelClass:
         if load_in_4bit or load_in_low_bit or bigdl_transformers_low_bit:
             # Speed up when loading model
             kwargs["low_cpu_mem_usage"] = True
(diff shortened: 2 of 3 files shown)


python/llm/example/transformers/transformers_int4/transformers_int4_pipeline.py
```diff
@@ -47,7 +47,7 @@
         print('Output:', output_str)
         print(f'Inference time: {end-st} s')
     elif model_path == 'THUDM/chatglm-6b':
-        model = AutoModel.from_pretrained(model_path, trust_remote_code=True, load_in_4bit=True)
+        model = AutoModel.from_pretrained(model_path, trust_remote_code=True, load_in_4bit=True, keep_in_fp32_modules=['output_layer'])
```
jason-dai, line 50: layers_not_to_convert=['output_layer']

python/llm/src/bigdl/llm/transformers/model.py
```diff
@@ -46,6 +46,13 @@ def from_pretrained(cls,
         if load_in_4bit or load_in_low_bit or bigdl_transformers_low_bit:
             # Speed up when loading model
             kwargs["low_cpu_mem_usage"] = True
+            # Specify some layers not be converted to ensure accuracy
+            keep_in_fp32_modules = kwargs.pop("keep_in_fp32_modules", [])
```
jason-dai, line 50: add python doc

python/llm/example/transformers/transformers_int4/transformers_int4_pipeline.py
jason-dai: actually change `keep_in_fp32_modules` to `layers_using_high_precision`

plusbang: > actually change `keep_in_fp32_modules` to `layers_using_high_precision`

I think this PR is now unnecessary, reasons can be found [here](https://github.com/analytics-zoo/nano/issues/434#issuecomment-1635201788)

PR #8506, opened 2023-07-13
[PPML] Add bigdl-llm-demo dependencies in image

gc-fu:
## Description

Add bigdl-llm-demo dependencies in `trusted-bigdl-llm` image.

This also include some changes required for better fschat performance.


ShanSimu (approved):
lgtm

PR #8519, opened 2023-07-13
LLM: fix memory access violation

MeouSker77:
Fix bug in <https://github.com/analytics-zoo/nano/issues/355>

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/linear_quant.py
+++ b/python/llm/src/bigdl/llm/transformers/linear_quant.py
@@ -153,13 +153,13 @@ def ggml_matmul_src1_x_src0_t(src0: torch.Tensor,
     if src1.dtype != torch.float32:
         src1 = src1.float()
 
-    src0_ptr = src0.data_ptr() + (src0.storage_offset() * src0.element_size())
-    src1_ptr = src1.data_ptr() + (src1.storage_offset() * src1.element_size())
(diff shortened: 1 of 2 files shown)


python/llm/src/bigdl/llm/transformers/linear_quant.py
```diff
@@ -153,13 +153,13 @@ def ggml_matmul_src1_x_src0_t(src0: torch.Tensor,
     if src1.dtype != torch.float32:
         src1 = src1.float()
 
-    src0_ptr = src0.data_ptr() + (src0.storage_offset() * src0.element_size())
-    src1_ptr = src1.data_ptr() + (src1.storage_offset() * src1.element_size())
```
MeouSker77, line 157: I think the address returned by `.data_ptr()` is always right, even this tensor is using others tensor's `Storage`, so there is no need to add `storage_offset * element_size`.

MeouSker77:
image
In this crash, `src_1_data` argument of `ggml_compute_forward_mul_mat_q_fp32` is `0x1eef311dfc0`, and the error is caused by reading `0x1eef311e060`.

`src_1_data` has 4096 fp32 elements, `0x1eef311dfc0 + 4096*4` > `0x1eef311e060`, so I think `src_1_data` is wrong, so we shouldn't add `storage_offset() * element_size()`

plusbang (approved):
LGTM

PR #8530, opened 2023-07-14
[LLM] Fix the problem of workflow cancellation after pr merge

hoshibara:
## Description

Fix the problem of workflow cancellation after pr merge.


MeouSker77 (approved):
LGTM

PR #8531, opened 2023-07-14
Add load_low_bit save_load_bit to AutoModelForCausalLM

qiuxin2012:
## Description

Add example for save/load api.
Add load_low_bit save_load_bit to AutoModelForCausalLM

### 1. Why the change?
Split load_low_bit from from_pretrained
Add example for save/load model.

### 2. User API changes
AutoModelForCausalLM's API is changed:
1. from_pretrained is used to load float or hf model into low-bit model.
2. save_low_bit is used to save low-bit model a folder.
3. load_low_bit is used to load serialized low-bit model

### 3. Summary of the change 

Add load_low_bit and save_low_bit to AutoModelForCausalLM
Add example for save/load model.

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/python/llm/example/transformers/transformers_low_bit/README.md
+++ b/python/llm/example/transformers/transformers_low_bit/README.md
@@ -1,0 +1,43 @@
+# BigDL-LLM Transformers INT4 Inference Pipeline for Large Language Model
+
+In this example, we show a pipeline to apply BigDL-LLM low-bit optimizations to any Hugging Face Transformers model, and then run inference on the optimized low-bit model.
+
+## Prepare Environment
+We suggest using conda to manage environment:
+```bash
+conda create -n llm python=3.9
+conda activate llm
+
+pip install --pre --upgrade bigdl-llm[all]
+```
+
+## Run Example
+```bash
+python ./transformers_low_bit_pipeline.py --model-path decapoda-research/llama-7b-hf --low-bit sym_int5 --save-path ./llama-7b-sym_int5
+```
+arguments info:
+- `--repo-id-or-model-path`: str value, argument defining the huggingface repo id for the large language model to be downloaded, or the path to the huggingface checkpoint folder, the value is 'decapoda-research/llama-7b-hf' by default.
+- `--low-bit`: str value, options are sym_int4, asym_int4, sym_int5, asym_int5 or sym_int8. The model's linear will be loaded into corresponding low-bit type. sym_int4 means symmetric int 4, asym_int4 means asymmetric int 4, etc.
+- `--save-path`: str value, the path to save the low-bit model. Then you can load the low-bit directly.
+- `--load-path`: optional str value. The path to load low-bit model.
+
+
+## Sample Output for Inference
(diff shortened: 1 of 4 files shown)


python/llm/src/bigdl/llm/transformers/model.py
```diff
@@ -21,6 +21,10 @@
 from bigdl.llm.utils.common import invalidInputError
 
 
+def save_low_bit(self, *args, **kwargs):
+    self.save_pretrained(*args, **kwargs)
```
jason-dai, line 25: we should remove `save_pretrained` method and only use `save_low_bit`

python/llm/example/transformers/transformers_low_bit/README.md
```diff
@@ -0,0 +1,43 @@
...
+pip install --pre --upgrade bigdl-llm[all]
+```
+
+## Run Example
+```bash
+python ./transformers_low_bit_pipeline.py --model-path decapoda-research/llama-7b-hf --low-bit sym_int5 --save-path ./llama-7b-sym_int5
+```
+arguments info:
+- `--repo-id-or-model-path`: str value, argument defining the huggingface repo id for the large language model to be downloaded, or the path to the huggingface checkpoint folder, the value is 'decapoda-research/llama-7b-hf' by default.
+- `--low-bit`: str value, options are sym_int4, asym_int4, sym_int5, asym_int5 or sym_int8. The model's linear will be loaded into corresponding low-bit type. sym_int4 means symmetric int 4, asym_int4 means asymmetric int 4, etc.
```
jason-dai, line 20: `--low-bit`: str value, options are sym_int4, asym_int4, sym_int5, asym_int5 or sym_int8. (sym_int4 means symmetric int 4, asym_int4 means asymmetric int 4, etc.). Relevant low bit optimizations will be applied to the model.

python/llm/src/bigdl/llm/transformers/model.py
qiuxin2012: If I remove save_pretrained, how can I call `save_pretrained` in save_low_bit?

qiuxin2012: Just record, we can delete save_pretrained like this:
```python
        import types
        from functools import partial
        save_func = partial(model.save_pretrained)
        save_low_bit_func = partial(save_low_bit, save_func=save_func)  # call this instead of save_pretrained in save_low_bit
        model.save_low_bit = types.MethodType(save_low_bit_func, model)
        from transformers.modeling_utils import PreTrainedModel
        del PreTrainedModel.save_pretrained
```

jason-dai: No need to delete for now

qiuxin2012: Model loaded with load_low_bit can be saved using save_load_bit correctly.

leonardozcm (approved):
lgtm

PR #8534, opened 2023-07-14
pin bigdl-llm version

gc-fu:

### DIFF
--- a/ppml/trusted-bigdl-llm/base/Dockerfile
+++ b/ppml/trusted-bigdl-llm/base/Dockerfile
@@ -13,7 +13,7 @@ ARG https_proxy
 ADD utils.diff /opt/utils.diff
 ADD llm_cli.diff /opt/llm_cli.diff
 
-RUN pip3 install --pre --upgrade bigdl-llm[all] && \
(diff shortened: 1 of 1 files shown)


ShanSimu (approved):
lgtm

jason-dai:
Instead of using an arbitrary nightly build, add your use case as a unit test in bigdl-llm to make sure it always works.

gc-fu:
> Instead of using an arbitrary nightly build, add your use case as a unit test in bigdl-llm to make sure it always works.

The reason why we pin `bigdl-llm` version is that **ppml** keeps another copy of `llm.cpp` which is not as updated as the `intel-analytics/llm.cpp` repo.

So, the new nightly release of `bigdl-llm` will break the code because the `.so` files generated by `ppml-llm.cpp` is not updated.  

We keep `ppml-llm.cpp` because we need to use thread pool instead of creating threads on every token inference as it will make it impossible to run in `SGX` environment.

Our use case is a little bit special and need SGX environment and contains self-modified code in `llm.cpp`, so it is quite difficult to add it in unit test.

We know that use an arbitrary nightly release is bad, so we will replace this once we have a new release of `bigdl-llm` (Maybe bigdl-llm 2.4.0?).

PR #8537, opened 2023-07-14
[LLM] Setup.py & llm-cli update for windows vnni binary files

hoshibara:
## Description

* setup.py will contain new built vnni files.
* llm-cli will check if platform has vnni instruction to determine which exec files will be called.


MeouSker77 (approved):
LGTM

PR #8555, opened 2023-07-18
[PPML]doc: fix the format of service-account-token

ShanSimu:
## Description

fix the format of service-account-token


Uxito-Ada (approved):
lgbm

PR #8559, opened 2023-07-18
Transfer the MAR decompress function from frontend to backend.

Hanyu-Jin:
## Description

Transfer the MAR decompress function from frontend to backend.

### 1. Why the change?

 To ensure the security.

### 2. User API changes

None.


glorysdj (approved):
LGTM

PR #8565, opened 2023-07-19
[LLM] Add dependency file check in setup.py

hoshibara:
## Description

* Add dependency file check in setup.py


Oscilloscope98 (approved):
lgtm

PR #8569, opened 2023-07-19
[PPML] Add attestation dep to image

gc-fu:
## Description

Add attestation-related dependencies to bigdl-llm tdx image


xiangyuT (approved):
LGTM

PR #8584, opened 2023-07-21
fix xgboost in spark2.4.6

Le-Zheng:
## Description

xgboost with Spark 2.4.6 needs lower version


Le-Zheng:
spark 2.4.6 jenkins: http://10.112.231.51:18888/job/BigDL-DLLIB-NB-UnitTests/690/

qiuxin2012 (approved):

PR #8587, opened 2023-07-21
[PPML] Change entrypoint for bigdl-llm tdx image

gc-fu:
## Description

 Change entrypoint for bigdl-llm tdx image.


ShanSimu (approved):
lgtm

PR #8590, opened 2023-07-21
[LLM] Add ` torch_dtype='auto'` when loading in 4 bit for transformers int4 examples

Oscilloscope98:
## Description

Add `torch_dtype='auto'` when loading in 4 bit for transformers int4 examples

### 1. Why the change?

https://github.com/analytics-zoo/nano/issues/451#issuecomment-1642901502
https://huggingface.co/docs/transformers/main_classes/model#transformers.PreTrainedModel.from_pretrained.torch_dtype

If we do not specify `torch_dtype='auto'` when loading model in 4 bit, default fp32 dtype could be used instead of what is specified in model's `config.json`. That is, 16 bit model may be loaded in fp32 at first, which could lead to 4XGB memory required for XB param models.

`torch_dtype='auto'` will at first check the `torch_dtype` specified in model's `config.json` to load the model. That is, 16 bit model will be loaded in 16 bit at first, which requires 2XGB memory for XB param models (as we expected).

**We will validate the improvement of `torch_dtype='auto' for each example later**

### DIFF
--- a/python/llm/example/transformers/transformers_int4/baichuan/generate.py
+++ b/python/llm/example/transformers/transformers_int4/baichuan/generate.py
@@ -42,6 +42,7 @@ if __name__ == '__main__':
     # which convert the relevant layers in the model into INT4 format
     model = AutoModelForCausalLM.from_pretrained(model_path,
                                                  load_in_4bit=True,
+                                                 torch_dtype='auto',
                                                  trust_remote_code=True)
 
     # Load tokenizer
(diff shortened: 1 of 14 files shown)


jason-dai:
Instead of changing all the applications, maybe we should set it in our `AutoModel.from_pretrained` method?

plusbang (approved):
LGTM

Oscilloscope98:
Close as #8592 has been merged

PR #8593, opened 2023-07-24
[PPML] Fix bigdl-llm-tdx image's entrypoint

gc-fu:
## Description

Make it compatible when user do not supply any command-line arguments.


xiangyuT (approved):
LGTM

PR #8597, opened 2023-07-24
[LLM] add CausalLM and Speech UT

Litchilitchy:
## Description

add CausalLM and Speech model UTs for LLM module

### DIFF
--- a/python/llm/test/inference/test_transformers_api.py
+++ b/python/llm/test/inference/test_transformers_api.py
@@ -87,7 +87,7 @@ class TestTransformersAPI(unittest.TestCase):
             end = time.time()        
         print('Output:', transcription)
         print(f'Inference time: {end-st} s')
-        res = 'Mr. Quilter is the apostle of the middle classes and we are glad to welcome his gospel.' in transcription
+        res = 'Mr. Quilter is the apostle of the middle classes and we are glad to welcome his gospel.' in transcription[0]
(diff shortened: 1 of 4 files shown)


python/llm/test/inference/test_transformers_api.py
```diff
@@ -17,11 +17,11 @@
...
 import unittest
 import os
-
+import pytest
 import time
 import torch
-from bigdl.llm.transformers import AutoModel
-from transformers import AutoTokenizer
+from bigdl.llm.transformers import AutoModel, AutoModelForCausalLM, AutoModelForSpeechSeq2Seq
+from transformers import AutoTokenizer, AutoModel as TAutoModel, AutoModelForCausalLM as TAutoModelForCausalLM
```
Oscilloscope98, line 24: Seems that this `AutoModel as TAutoModel, AutoModelForCausalLM as TAutoModelForCausalLM` is not needed to import

Oscilloscope98 (commented):
Seems that you forgot to change https://github.com/intel-analytics/BigDL/blob/main/python/llm/test/run-llm-inference-tests.sh#L12 and https://github.com/intel-analytics/BigDL/blob/main/python/llm/test/run-llm-inference-tests.sh#L12 :)

Oscilloscope98 (approved):
LGTM

Issue #8501, opened 2023-07-12
Deprecated OpenVINO 2020.3.2 and TensorFlow 1.15


qiyuangong:
Due to recent CVE issues and frameworks update, these packages are out of date and may cause security issues.

- [ ] Deprecated OpenVINO 2020.3.2 
- [ ] Deprecated TensorFlow 1.15
- [ ] Exclude OpenVINO 2020.3.2 building from BigDL
- [ ] Exclude TensorFlow 1.15 building from BigDL

PR #8599, opened 2023-07-24
[Orca] Deprecate TensorFlow 1.15.0 API

sgwhat:
## Description
Due to recent CVE issues and frameworks update, we need to deprecate the outdated packages to avoid security issues.

### 1. Why the change?
See https://github.com/intel-analytics/BigDL/issues/8501.

### 2. User API changes
N/A

### 3. Summary of the change 
1. Add an error to remind users that tensorflow 1.15.0 APIs have been deprecated.
2. Deprecate Apis including `tf.estimator` and `tfpark`.

### 4. How to test?
- [ ] Unit test

### 5. To Do
- [x] Deprecate apis

### DIFF
--- a/python/orca/src/bigdl/orca/data/tf/tf1_data.py
+++ b/python/orca/src/bigdl/orca/data/tf/tf1_data.py
@@ -21,13 +21,13 @@ from bigdl.orca.tfpark import TFDataset
 from bigdl.dllib.utils.log4Error import invalidInputError
 
 from typing import TYPE_CHECKING, Optional, Any
+import logging
 if TYPE_CHECKING:
     from bigdl.orca.data.tf.data import MapDataset
     from pyspark.rdd import RDD
 
 
-invalidInputError(False,
-                  "TF1 API has been deprecated, you may use TF2 Estimator instead.")
+logging.warning("TF1 API has been deprecated, you may use TF2 Estimator instead.")
 
 
 class TF1Dataset(TFDataset):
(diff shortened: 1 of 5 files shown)


python/orca/src/bigdl/orca/data/tf/tf1_data.py
```diff
@@ -25,6 +25,9 @@
     from bigdl.orca.data.tf.data import MapDataset
     from pyspark.rdd import RDD
 
+invalidInputError(False,
```
hkvision, line 28: Can we just put this in the init of the data/tf&tfpark related packages?

yangw1234:
`invalidInputError(False, ...)` will make the file or package unusable, in that case, can we also delete all the code in the file or package?

sgwhat:
> `invalidInputError(False, ...)` will make the file or package unusable, in that case, can we also delete all the code in the file or package?

We will keep these tf1.15 APIs preserved. Due to security issues, we have temporarily disabled these APIs. When there are user demands for related functionalities, we will restore these features.

hkvision:
> > `invalidInputError(False, ...)` will make the file or package unusable, in that case, can we also delete all the code in the file or package?
> 
> Maybe we could keep these tf1.15 APIs preserved😄. Due to security issues, we have temporarily disabled these APIs. When there are user demands for related functionalities, we will restore these features.

Agree, better to keep these code in case they are needed in the future.

qiyuangong (approved):
LGTM

sgwhat:
Some actions failed but were not relevant.

PR #8600, opened 2023-07-24
LLM: Reduce Peak Memory when loading transformer int4 models

leonardozcm:
## Description

1. In huggingface transformer from_pretrained api, it will initialize model weights(on ram) if there are any keys missing in state_dict: https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L3129
```python
            set_initialized_submodules(model, _loaded_keys)
            # This will only initialize submodules that are not marked as initialized by the line above.
            model.apply(model._initialize_weights)
```
This will waste the efforts of what `low_cpu_mem_usage` does, for if we will pass an empty state_dict here, there would be nothing in `_loaded_keys ` which will make hf init the whole models(4XGB) here.

2. Actually we don't need to set an empty dict here, when specifing `low_cpu_mem_usage` state_dict will be automatically set as `None `https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L2688
In another word, **`low_cpu_mem_usage` must be used when loading files from disk.**

## Experiments(On CPX)

Peak memory observed on Baichuan13B load_low_bit:
Start point:
image

Peak memory before this PR(time during is  92.51276755332947):
image

Peak memory after this PR(time during is  14.094629764556885):
image


leonardozcm:
BTW if there is any model still triggers OOM when loading, we may specify `offload_state_dict = True ` to enable np.memmap to  place big tensors on disk and read&write part of them, but still don't know whether R&W will behave when inference.
https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L2092-L2095

yangw1234 (approved):
LGTM

Oscilloscope98 (approved):
LGTM

PR #8601, opened 2023-07-24
LLM: Mute shape mismatch output

leonardozcm:
## Description

In this PR we mute huggingface logger when loading int4 model for a clean output.

Before:
```log
(changmim-llm) cpx@cpx-2:~/changmin/BigDL/python/llm$ python test_memory_load.py 
Some weights of LlamaForCausalLM were not initialized from the model checkpoint at /disk1/changmin/vicuna-13b-int4 and are newly initialized because the shapes did not match:
- model.layers.0.self_attn.q_proj.weight: found shape torch.Size([13926400]) in the checkpoint and torch.Size([5120, 5120]) in the model instantiated
- model.layers.0.self_attn.k_proj.weight: found shape torch.Size([13926400]) in the checkpoint and torch.Size([5120, 5120]) in the model instantiated
- model.layers.0.self_attn.v_proj.weight: found shape torch.Size([13926400]) in the checkpoint and torch.Size([5120, 5120]) in the model instantiated
......
You should probably TRAIN this model on a down-stream task to be able to use it for predictions and inference.
time during is  14.336107730865479
```

After:
```log
(changmim-llm) cpx@cpx-2:~/changmin/BigDL/python/llm$ python test_memory_load.py 
time during is  14.293046474456787
```

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/model.py
+++ b/python/llm/src/bigdl/llm/transformers/model.py
@@ -125,6 +125,8 @@ class _BaseAutoModelClass:
 
         # Avoid KeyError
         kwargs["ignore_mismatched_sizes"] = True
+        # Avoid reading from local file at the first initialization
+        kwargs["state_dict"] = {}
 
         # Maybe needed when extract_local_archive_file
         subfolder = kwargs.get("subfolder", "")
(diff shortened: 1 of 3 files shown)


hkvision:
Will this change impact other logs?
Also what's the root cause of this log? Why shape is mismatch?

leonardozcm:
> Will this change impact other logs?
We only apply changes on modelutils.logger, with all warnings disabled when from_pretrained. I believe there is no way to extract a single warning from a logger?

> Also what's the root cause of this log? sWhy shape is mismatch?
pls refer to https://github.com/intel-analytics/BigDL/pull/8462/files#diff-b546c6863dc39322e7aafe9e7a8924e1b92c5cd6d52ea0337904160b2eea7272R39-R41

PR #8617, opened 2023-07-26
[PPML] Pin pydantic version in fastchat image

gc-fu:
## Description

Pin pydantic version because this error here:

> pydantic.errors.PydanticImportError: BaseSettings has been moved to the pydantic-settings package. See https://docs.pydantic.dev/2.0/migration/#basesettings-has-moved-to-pydantic-settings for more details.


hzjane (approved):
LGTM

PR #8618, opened 2023-07-26
[LLM] Fix cancel flag causing nightly builds to fail

hoshibara:
## Description

* Set the cancel flag to false, enabling this flag causes the binary build step to be canceled.


Oscilloscope98 (approved):
LGTM

PR #8619, opened 2023-07-26
[LLM] Remove concurrency group for binary build workflow

hoshibara:
## Description

This [known issue](https://github.com/orgs/community/discussions/5435) will cause the binary build step in nightly tests to be canceled unexpectedly. We had to remove the concurrency group to ensure all binary build steps worked.


Oscilloscope98 (approved):
LGTM

Issue #8615, opened 2023-07-26
openllama 13b model with BigdlNativeForCausalLM use transformer int4 will cause error during generate


ZhaoqiongZ:
model: openllama-13b:  https://huggingface.co/openlm-research/open_llama_13b
usage:
```
from bigdl.llm.transformers import BigdlNativeForCausalLM
from transformers import AutoTokenizer
import time

from bigdl.llm.transformers import AutoModelForCausalLM


llm = BigdlNativeForCausalLM.from_pretrained(

        pretrained_model_name_or_path="/home/zhaoqion/llm/openllama-out/bigdl_llm_llama_q4_0.bin",

        model_family="llama",

        n_threads=16)

# # llm = AutoModelForCausalLM.from_pretrained("/home/zhaoqion/llm/open_llama_13b", load_in_4bit=True)
# # llm.save_low_bit("/home/zhaoqion/llm/openllama_int4")
# llm = AutoModelForCausalLM.load_low_bit("/home/zhaoqion/llm/openllama_int4")

tokenizer = AutoTokenizer.from_pretrained("/home/zhaoqion/llm/open_llama_13b")

start = time.time()

tokens_id = tokenizer("Q: What is the book you recommend for learning math?\nA:",  return_tensors="pt").input_ids
token_time = time.time()
# print("token input",tokens_id)
output_tokens_id = llm.generate(tokens_id, max_new_tokens=256)
generate_time = time.time()

output = tokenizer.batch_decode(output_tokens_id)
```

BigdlNativeForCausalLM will trigger the error

```
bigdl-llm: loading model from /home/zhaoqion/llm/openllama-out/bigdl_llm_llama_q4_0.bin
loading bigdl-llm model: format     = ggjt v3 (latest)
loading bigdl-llm model: n_vocab    = 32000
loading bigdl-llm model: n_ctx      = 512
loading bigdl-llm model: n_embd     = 5120
loading bigdl-llm model: n_mult     = 256
loading bigdl-llm model: n_head     = 40
loading bigdl-llm model: n_layer    = 40
loading bigdl-llm model: n_rot      = 128
loading bigdl-llm model: ftype      = 2 (mostly Q4_0)
loading bigdl-llm model: n_ff       = 13824
loading bigdl-llm model: n_parts    = 1
loading bigdl-llm model: model size = 13B
loading bigdl-llm model: ggml ctx size = 6595.81 MB
[...]

PR #8620, opened 2023-07-26
LLM: support tensor input of native int4 `generate`

plusbang:
## Description

Related to https://github.com/intel-analytics/BigDL/issues/8615.

### 1. Why the change?

Currently, our native int4 `generate` method only support int/List(int) type as input token id. 
So when use `tokens_id = tokenizer(prompt, return_tensors="pt").input_ids` which is quite common in huggingface, there exists error.
Support tensor input in this PR.

### 2. User API changes

Now, we also support usage:
```python
tokens_id = tokenizer(prompt, return_tensors="pt").input_ids # Use HuggingFace transformers tokenizer
output_tokens_id = llm.generate(tokens_id)
```

### 4. How to test?
- [x] Unit test
- [x] Local test

### DIFF
--- a/python/llm/src/bigdl/llm/ggml/model/generation/utils.py
+++ b/python/llm/src/bigdl/llm/ggml/model/generation/utils.py
@@ -101,9 +101,9 @@ class GenerationMixin:
 
     def generate(
         self,
-        inputs: Union[Optional[Sequence[int]],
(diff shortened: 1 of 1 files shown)


MeouSker77 (approved):
LGTM

PR #8621, opened 2023-07-26
[LLM] Add needs flag for nightly build

hoshibara:
## Description

* Add `needs` flag to ensure nightly build runs in the correct order


Oscilloscope98 (approved):
LGTM

PR #8645, opened 2023-08-01
LLM: Disable transformer api `pretraining_tp`

leonardozcm:
## Description

To avoid linear parallel involved F.linear operation, we turn off `pretraining_tp` in large models.
https://github.com/analytics-zoo/nano/issues/482

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/model.py
+++ b/python/llm/src/bigdl/llm/transformers/model.py
@@ -72,7 +72,7 @@ class _BaseAutoModelClass:
             kwargs["low_cpu_mem_usage"] = True
             # set default torch_dtype='auto'
             kwargs["torch_dtype"] = kwargs.get("torch_dtype", 'auto')
-            # Avoid tensor parallel F.Linear Operations 
+            # Avoid tensor parallel F.Linear Operations
             if "pretraining_tp" in config_dict:
                 kwargs["pretraining_tp"] = 0
             q_k = load_in_low_bit if load_in_low_bit else "sym_int4"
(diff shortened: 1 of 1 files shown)


yangw1234 (approved):

python/llm/src/bigdl/llm/transformers/model.py
```diff
@@ -72,6 +72,9 @@ def from_pretrained(cls,
             kwargs["low_cpu_mem_usage"] = True
             # set default torch_dtype='auto'
             kwargs["torch_dtype"] = kwargs.get("torch_dtype", 'auto')
+            # Avoid tensor parallel F.Linear Operations
+            if "pretraining_tp" in config_dict:
+                kwargs["pretraining_tp"] = 0
```
jason-dai, line 77: should it be 1?

leonardozcm: I think both 0 and 1 will prevent this problem, only if self.pretraining_tp <= 1
```
        if self.pretraining_tp > 1: 
```

jason-dai: what the default value for 7B/13B?

leonardozcm: They don't have pretraining_tp this entry in their config.json

jason-dai:
https://huggingface.co/docs/transformers/main/model_doc/llama2
```
Setting config.pretraining_tp to a value different than 1 will activate the more accurate but slower computation of the linear layers, which should better match the original logits.
```

leonardozcm:
OK I have set this to default 1 as a precaution

PR #8648, opened 2023-08-01
Chronos: fix nightly version installation

plusbang:
## Description

According to the failed nightly action https://github.com/intel-analytics/BigDL/actions/runs/5716336510/job/15501367462#step:6:236, there exists error on nightly version installation.
The root cause is conflict on protobuf version of nano and orca.

Fix the nightly version installation in this PR. (This is only a work around before nano upgrade protobuf.)

### 2. User API changes

N/A

### 4. How to test?
- [x] Unit test
- [x] Local test


rnwang04 (approved):
LGTM

PR #8650, opened 2023-08-01
[LLM] Disable chatglm llm-cli unittest

hoshibara:
## Description

Because clx lacks avx_vnni, it is necessary to disable chatglm test to ensure that unittest & nightly test can work.


Oscilloscope98 (approved):
LGTM. We may add it back later

PR #8654, opened 2023-08-02
[PPML]Feat: support FastChat on Gramine

ShanSimu:
## Description

Support FastChat in Gramine

### 1. Why the change?

There may be customer needs in the future

### 2. Summary of the change 

Add FastChat support to intelanalytics/bigdl-ppml-trusted-bigdl-llm-gramine


gc-fu (approved):
LGTM

PR #8658, opened 2023-08-02
LLM first example test and template

Litchilitchy:
## Description

LLM first example test and template


.github/workflows/llm_example_tests.yml
```diff
@@ -0,0 +1,78 @@
...
+          python -m pip install --upgrade setuptools==58.0.4
+          python -m pip install --upgrade wheel
+
+      - name: Download llm binary
+        uses: ./.github/actions/llm/download-llm-binary
+
+      - name: Run LLM install (all) test
+        uses: ./.github/actions/llm/setup-llm-env
+        env:
+          ANALYTICS_ZOO_ROOT: ${{ github.workspace }}
```
hkvision, line 68: Can we change this to BIGDL_ROOT?

.github/actions/llm/example-test/action.yml
```diff
@@ -0,0 +1,28 @@
...
+            wget -r -nH --no-verbose --cut-dirs=1 $LLM_FTP_URL/${ORIGINAL_LLAMA2_PATH:2} -P $LLM_DIR
+        fi
+
+        source $CONDA_HOME/bin/activate bigdl-llm-test
+        std=$(python python/llm/example/transformers/transformers_int4/llama2/generate.py --repo-id-or-model-path $ORIGINAL_LLAMA2_PATH)
+        echo "the output of the example is: " 
+        echo $std
+        if [[ ! $std == *"AI is a term"* ]]; then
+          echo "The expected output is not met."
+          return 1
```
hkvision, line 26: Put this into a separate script so that other examples are easier to be added.

.github/workflows/llm_example_tests.yml
```diff
@@ -0,0 +1,80 @@
+name: LLM Example Test
+
+# Cancel previous runs in the PR when you push new commits
+concurrency:
+  group: ${{ github.workflow }}-llm-example-tests-${{ github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: true
+
+# Controls when the action will run. 
+on:
```
Oscilloscope98, line 9: Should we trigger the example test nightly? Maybe there is no need to trigger it every PR. The example test will be heavy later.
Refer to the "on" condition here: https://github.com/intel-analytics/BigDL/blob/main/.github/workflows/llm-nightly-test.yml#L9

Oscilloscope98: BTW, should we just put it into the nightly test yml?

python/llm/dev/test/run-example-tests.sh
```diff
@@ -0,0 +1,15 @@
+# LLAMA2 example test
+if [ ! -d $ORIGINAL_LLAMA2_PATH ]; then
```
Oscilloscope98, line 2: I do not find the place for defining `ORIGINAL_LLAMA2_PATH`, should we put the downloading into the yml file as other tests?

python/llm/dev/test/run-example-tests.sh
```diff
@@ -0,0 +1,15 @@
...
+    echo "Directory $ORIGINAL_LLAMA2_PATH not found. Downloading from FTP server..."
+    wget -r -nH --no-verbose --cut-dirs=1 $LLM_FTP_URL/${ORIGINAL_LLAMA2_PATH:2} -P $LLM_DIR
+fi
+
+source $CONDA_HOME/bin/activate bigdl-llm-test
```
Oscilloscope98, line 7: No conda is needed for now. We have removed the conda for llm tests totally.

python/llm/dev/test/run-example-tests.sh
```diff
@@ -0,0 +1,15 @@
...
+std=$(python python/llm/example/transformers/transformers_int4/llama2/generate.py --repo-id-or-model-path $ORIGINAL_LLAMA2_PATH)
```
Oscilloscope98, line 8: We may need taskset for better test performance.
Please refer to here: https://github.com/intel-analytics/BigDL/blob/main/python/llm/test/run-llm-inference-tests.sh#L14 :)

.github/workflows/llm_example_tests.yml
hkvision: I suppose this should be first run in the PR to make sure it succeeds before the merge. Then we can move to nightly only.

Oscilloscope98: https://github.com/intel-analytics/BigDL/blob/main/.github/workflows/llm-nightly-test.yml#L9 The on condition here will trigger the test for the PR related to the example yml, and trigger this test nightly after merging this PR.

python/llm/dev/test/run-example-tests.sh
Litchilitchy: > Should we put the downloading into the yml file as other tests?

@hkvision what do you think?

hkvision: I think putting the download together with run script is better since in the future we only need to edit one file?

Oscilloscope98: If later we put the example tests directly into the nightly test yml, they may share the downloaded models, so it may be better then to put the downloading into the yml files. But anyway, we could keep the workflow as it for now and change them later. :)

hkvision: `if [ ! -d $ORIGINAL_LLAMA2_PATH ]` there is path check, if the models are downloaded elsewhere, here still won't download again, no?

Litchilitchy: > `if [ ! -d $ORIGINAL_LLAMA2_PATH ]` there is path check, if the models are downloaded elsewhere, here still won't download again, no?

yes, as long as the working directory is the same

Oscilloscope98: Yes, the only thing is for other tests, the model may not be downloaded yet before we run into this sh file. Thus for all other tests, the downloading part is put into the yml file before the sh file, so can be safely shared by all tests.

Oscilloscope98: Anyway, we could keep as it for now :)

Oscilloscope98 (approved):
LGTM

PR #8662, opened 2023-08-02
optimize chatglm2 long sequence

yangw1234:
### 1. Why the change?

optimize chatglm2 long sequence

### 2. User API changes

No API changes.

### 3. Summary of the change 

1. change the tensor layout from [seq_length, batch_size, num_head, hidden_size] to [batch_size, num_head, seq_length, hidden_size] to make attention faster.
2. move the (expand -> reshape) operation before KV cache to avoid large memory copy, (at the expense of larger KV cache)
3. pre-allocate KV cache optimization

### 4. How to test?
manually tested it

### 5. New dependencies

- [ ] New Python dependencies
       - Dependency1 
       - Dependency2
       - ...
       - Dependency1 and license1
       - Dependency2 and license2
       - ...

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/model.py
+++ b/python/llm/src/bigdl/llm/transformers/model.py
@@ -24,6 +24,9 @@ from .utils import extract_local_archive_file, \
     fix_key
 from bigdl.llm.ggml.quantize import ggml_tensor_qtype
 from bigdl.llm.utils.common import invalidInputError, MuteHFLogger
+import sys
+import importlib
+
 
 
 def save_low_bit(self, *args, **kwargs):
@@ -104,6 +107,14 @@ class _BaseAutoModelClass:
         else:
             # todo implement 4.28.0 ~ 4.30.2
             pass
+
+        modeling_module_name = model.__class__.__module__
+        module = importlib.import_module(modeling_module_name)
+        from bigdl.llm.transformers.models.chatglm import chatglm_attention_forward_8eb45c
+        convert_forward(model,
+                        module.SelfAttention,
+                        chatglm_attention_forward_8eb45c
+                        )
 
     @classmethod
     def load_convert(cls, q_k, *args, **kwargs):
--- a/python/llm/src/bigdl/llm/transformers/model.py
+++ b/python/llm/src/bigdl/llm/transformers/model.py
@@ -110,11 +110,15 @@ class _BaseAutoModelClass:
 
         modeling_module_name = model.__class__.__module__
         module = importlib.import_module(modeling_module_name)
-        from bigdl.llm.transformers.models.chatglm import chatglm_attention_forward_8eb45c
+        from bigdl.llm.transformers.models.chatglm import chatglm_attention_forward_8eb45c, cross_attn_forward
         convert_forward(model,
                         module.SelfAttention,
                         chatglm_attention_forward_8eb45c
                         )
+        convert_forward(model,
+                module.CoreAttention,
+                cross_attn_forward
+                )
 
     @classmethod
     def load_convert(cls, q_k, *args, **kwargs):
--- a/python/llm/src/bigdl/llm/transformers/model.py
+++ b/python/llm/src/bigdl/llm/transformers/model.py
@@ -108,17 +108,18 @@ class _BaseAutoModelClass:
             # todo implement 4.28.0 ~ 4.30.2
             pass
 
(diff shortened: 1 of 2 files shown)


python/llm/src/bigdl/llm/transformers/model.py
```diff
@@ -105,6 +107,19 @@ def optimize(cls, model):
...
+            module = importlib.import_module(modeling_module_name)
+            from bigdl.llm.transformers.models.chatglm import chatglm_attention_forward_8eb45c
+            from bigdl.llm.transformers.models.chatglm import cross_attn_forward_8eb45c
+            convert_forward(model,
+                            module.SelfAttention,
+                            chatglm_attention_forward_8eb45c
+                            )
+            convert_forward(model,
+                            module.CoreAttention,
+                            cross_attn_forward_8eb45c)
```
jason-dai, line 121: should be core_attn_forward?

python/llm/src/bigdl/llm/transformers/model.py
```diff
@@ -105,6 +107,19 @@ def optimize(cls, model):
             # todo implement 4.28.0 ~ 4.30.2
             pass
 
+        if "chatglm2" in model.config._name_or_path:
```
jason-dai, line 110: do we need to check the transformers `version`?

python/llm/src/bigdl/llm/transformers/model.py
```diff
@@ -105,6 +107,19 @@ def optimize(cls, model):
...
+            from bigdl.llm.transformers.models.chatglm import chatglm_attention_forward_8eb45c
+            from bigdl.llm.transformers.models.chatglm import cross_attn_forward_8eb45c
+            convert_forward(model,
+                            module.SelfAttention,
+                            chatglm_attention_forward_8eb45c
+                            )
+            convert_forward(model,
+                            module.CoreAttention,
+                            cross_attn_forward_8eb45c)
+
```
jason-dai, line 122: we also need to call optimize in `load_low_bit`? Alternatively, we may refactor `load_convert` and `ggml_convert_quant` so that `load_pretrained` and `load_low_bit` call the same underlying method

jason-dai:
> move the (expand -> reshape) operation before KV cache to avoid large memory copy, (at the expense of larger KV cache)

What's the impact on client?

python/llm/src/bigdl/llm/transformers/model.py
yangw1234: There is no `chatglm` model in `transformers`. The model implementation is directly placed into the model repo. I have not find a way to check the version of the model implementation.

yangw1234:
> > move the (expand -> reshape) operation before KV cache to avoid large memory copy, (at the expense of larger KV cache)
> 
> What's the impact on client?

work in progress

yangw1234:
> > move the (expand -> reshape) operation before KV cache to avoid large memory copy, (at the expense of larger KV cache)
> 
> What's the impact on client?

Ideally, we should provide an efficient `scaled_dot_product_attention` that support multi-query attention without memory copy, then we can reduce the KV cache memory to its original size.

jason-dai:
> > > move the (expand -> reshape) operation before KV cache to avoid large memory copy, (at the expense of larger KV cache)
> > 
> > 
> > What's the impact on client?
> 
> Ideally, we should provide an efficient `scaled_dot_product_attention` that support multi-query attention without memory copy, then we can reduce the KV cache memory to its original size.

Is it easy to do? Or we can merge server specific optimizations first.

jason-dai:
> > > move the (expand -> reshape) operation before KV cache to avoid large memory copy, (at the expense of larger KV cache)
> > 
> > 
> > What's the impact on client?
> 
> Ideally, we should provide an efficient `scaled_dot_product_attention` that support multi-query attention without memory copy, then we can reduce the KV cache memory to its original size.

Is it easy to do? Or we can merge server specific optimizations first.

yangw1234:
> > move the (expand -> reshape) operation before KV cache to avoid large memory copy, (at the expense of larger KV cache)
> 
> What's the impact on client?

Updated client data. 

Client actually benefits more from this PR, probably because of its lower memory bandwidth. 

As for the additional memory overhead, I guess it is fine, because this PR just revert the memory savings introduced by multi-query attention and get back to the same level as llama.

jason-dai (approved):
LGTM

PR #8671, opened 2023-08-03
[LLM] Remove llm workflow conda dependency

hoshibara:
## Description

Remove llm workflow's dependency on conda. In the workflow, Python installed with setup-python is more convenient to manage, and the requirements for the runner are also lower.


Oscilloscope98 (approved):
LGTM

PR #8672, opened 2023-08-03
[LLM] chatglm pybinding support

MeouSker77:
## Description

add initial chatglm pybinding support, but need to wait for pyd/so file release

## usage

```python
from bigdl.llm.ggml.model.chatglm.chatglm import ChatGLM


model = ChatGLM("chatglm2-q4_0.bin", use_mmap=True/False, n_threads=24, n_ctx=512)

prompt = ""

print(model(prompt))

for output in model(prompt, stream=True):
    print(output)
```


plusbang (approved):
LGTM

PR #8677, opened 2023-08-04
[LLM] add chatglm pybinding binary file release

MeouSker77:
## Description

only release Windows avx-vnni version because the linux avx-vnni runner can not setup python 3.9 successfully, will fix it soon


.github/workflows/llm-binary-build.yml
```diff
@@ -312,6 +319,11 @@ jobs:
     needs: check-windows-avx2-vnni-artifact
     if: needs.check-windows-avx2-vnni-artifact.outputs.if-exists == 'false'
     steps:
+      - uses: actions/checkout@v3
```
hoshibara, line 322: We don't need a checkout for this job. The following code contains the procedure for checking out llm.cpp.

hoshibara (approved):
LGTM

PR #8680, opened 2023-08-04
bandit action

Le-Zheng:
## Description

add bandit action for scan /python

### DIFF
--- a/.github/workflows/sdl_bandit.yml
+++ b/.github/workflows/sdl_bandit.yml
@@ -1,0 +1,29 @@
+name: Scanner-bandit
+on:
+  schedule:
+    - cron: '0 3 * * 0' # GMT time, 3:00 GMT == 11:00 China Every Sunday
+#  push:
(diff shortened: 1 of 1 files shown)


.github/workflows/sdl_bandit.yml
```diff
@@ -0,0 +1,35 @@
...
+jobs:
+  bandit-python:
+    runs-on: [self-hosted, SDL]
+    steps:
+    - name: "bandit python"
+      run: |
+        export CI=false
+        source activate bigdl-bandit 
+        bandit -r python -f html -o /root/bandit_report.html -c /root/ipas_default.config
+    - name: HTML Preview
```
qiyuangong, line 31: Need to print the result with cat or other command.

qiyuangong: https://github.com/marketplace/actions/python-bandit-with-code-annotations

PR #8682, opened 2023-08-04
fix: delete entrypoint set of base image and llm image

ShanSimu:
## Description

delete entrypoint set of base image and llm image

### Result

Before this PR, `docker run -itd $DOCKER_IMAGE` works but `docker run -itd $DOCKER_IMAGE bash` doesn't. Now both work.

### Technical Details

https://stackoverflow.com/questions/21553353/what-is-the-difference-between-cmd-and-entrypoint-in-a-dockerfile
Default `entrypoint` is `/bin/sh -c` and default `CMD` is `bash`. After the docker container starts, it will execute `entrypoint`+`CMD`. The part after the docker image name of `docker run` will replace the default `CMD`. If `entrypoint` is set to `/bin/bash`, `docker run -itd $DOCKER_IMAGE bash` will cause docker to run `bash bash`.


gc-fu (approved):
LGTM

Issue #8681, opened 2023-08-04
Missing files while saving low bit pretrained model due to low version transformers lib.


lalalapotter:
When use transformers(<=4.30.0) to save low bit pretrained model, report following error:

```shell
OSError: /disk7/zcg/baichuan does not appear to have a file named configuration_baichuan.py. Checkout 'https://huggingface.co//disk7/zcg/baichuan/None' for available files.
```
Example code：
```python
from bigdl.llm.transformers import AutoModelForCausalLM
from transformers import AutoTokenizer

model_path = "baichuan-inc/Baichuan-13B-Chat"
model = AutoModelForCausalLM.from_pretrained(model_path,
                                             load_in_4bit=True,
                                             trust_remote_code=True)

tokenizer = AutoTokenizer.from_pretrained(model_path,
                                          trust_remote_code=True)

# Save pretrained model
save_dir = "./baichuan/"
model.save_low_bit(save_dir)

# Load pretrained model again
model = AutoModelForCausalLM.load_low_bit(save_dir, trust_remote_code=True)
```

After testing, we can avoid this issue by using transformers(>= 4.31.0), maybe need to set in `setup.py`

lalalapotter:
@plusbang Could you please check if we can set transformers version >= 4.31.0?

leonardozcm:
> @plusbang Could you please check if we can set transformers version >= 4.31.0?

Like, as far as we know, is there any side effect if we set transformers != 4.29?

plusbang:
> @plusbang Could you please check if we can set transformers version >= 4.31.0?

`transformers==4.31.0` is released on July 19. During recent performance experiments (later than July 19), `transformers==4.31.0` is installed when install nightly version `bigdl-llm[all]` and I didn't encounter problems of using high version `transformers`. BTW, I think you can check LLM Github Action before limiting `transformers` version.

lalalapotter:
Set transformer lib version in PR #8683 . Close the issue.

PR #8683, opened 2023-08-04
LLM: set transformers lib version.

lalalapotter:
## Description

This PR is to set transformers lib version to solve issue #8681 .


leonardozcm (approved):
LGTM

PR #8688, opened 2023-08-04
[PPML] Enable TLS option in entrypoint

gc-fu:


ppml/tdx/docker/trusted-bigdl-llm-fschat/entrypoint.sh
```diff
@@ -172,21 +179,46 @@ else
     model_path=$MODEL_PATH
   fi
 
-  controller_address="http://$controller_host:$controller_port"
+  if [[ -n $ENABLE_TLS ]]; then
```
qiyuangong, line 182: In entrypoint, we can directly use `ENABLE_TLS`.

qiyuangong: https://github.com/apache/spark/blob/master/bin/load-spark-env.sh#L33

PR #8693, opened 2023-08-07
Add usedforsecurity=False flag for test use of MD5 hash

xiangyuT:
### 1. Why the change?

Bandit scan has found use of weak MD5 hash, and if it's not for security use it can be allowed with `usedforsecurity=False` flag.


sgwhat (approved):
LGTM

PR #8697, opened 2023-08-07
[PPML] Remove Legacy Decrypt and Encrypt scala example

Romanticoseu:
## Description

Delete scala ppml example file


qiyuangong:
We also need to remove this line 
https://github.com/intel-analytics/BigDL/blob/4a602092262970a47fb61c424fc7a0659365da8e/scala/ppml/src/main/scala/com/intel/analytics/bigdl/ppml/attestation/admin/RegisterMrEnclave.scala#L19

qiyuangong (approved):
LGTM

PR #8702, opened 2023-08-08
[LLM] Binary build with ubuntu 22.04

hoshibara:
## Description

Try to update runner to ubuntu 22.04. Compiled binaries will be tested for compatibility on centos and ubuntu 16.04-22.04.


hoshibara:
binaries built on ubuntu 22.04 cannot run on aep-004's ubuntu 20.04 runner.

PR #8709, opened 2023-08-09
[PPML] Provide libtdx_attest.so to do attestation on Anolis

xiangyuT:
## Description

There are some differences in Anolis kernel, so the standard DCAP lib will fail to generate TD report and TD quote. We can provide a patch or a reference libtdx_attest.so to fix this problem.


gc-fu (approved):
LGTM

PR #8710, opened 2023-08-09
[LLM] Linux vnni build with ubuntu 18.04

hoshibara:
## Description

Ubuntu 18.04 and 20.04 cannot update the binutils version when installing gcc 11, resulting in the inability to correctly compile binary files of the linux avx vnni version. 
Update the binutils version so that ubuntu 18.04 can be compiled normally.


MeouSker77 (approved):
LGTM

PR #8716, opened 2023-08-10
[LLM] Add bigdl-core-xe dependency when installing bigdl-llm[xpu]

hoshibara:
## Description

Add bigdl-core-xe dependency when installing bigdl-llm[xpu]


rnwang04 (approved):
maybe update example readme too (remove installation of xe)

PR #8718, opened 2023-08-10
Enable Nano to receive memory files as input.

Hanyu-Jin:
## Description

Enable PytorchQuantizedModel, PytorchIPEXJITModel and PytorchIPEXJITBF16Model to receive memory files as input.

### 1. Why the change?

To avoid plain text saved on disk.

### 2. User API changes

None.

### 3. Summary of the change 

If path is a dict, get memory files directly.


rnwang04:
please rebase latest commit and run action again

rnwang04 (approved):
LGTM, merge it after all action pass

PR #8719, opened 2023-08-10
Modify Dockerfile and fix bugs.

Hanyu-Jin:
## Description

Modify inc in Dockerfile to receive model directly.
Fix bugs in k8s scripts.

### 1. Why the change?

To receive memory files instead of path as input.

### 2. User API changes

None.


xiangyuT (approved):
LGTM

PR #8729, opened 2023-08-11
Nano: fix action of ipex

rnwang04:
## Description

fix action caused by avx512 & ipex.

### 4. How to test?
- [x] Unit test


Hanyu-Jin (approved):
LGTM

PR #8732, opened 2023-08-11
[LLM] Revert compile OS for llm build workflow

hoshibara:
## Description

Through [this problem](https://github.com/intel-analytics/BigDL/actions/runs/5827919438/job/15804958337?pr=8713#step:6:615), we found that there was a problem with the installation method of the previous test environment, which caused the bigdl-llm package to not be updated as expected during the test, resulting in the failure of the previous test. 
After changing the compilation environment, the binary files of the avx512 instruction set cannot run correctly.


MeouSker77 (approved):
LGTM

Issue #8715, opened 2023-08-10
ChatGLM2 Native INT4 model outputs English without space by bigdl-llm0808. 


KiwiHana:
ChatGLM2 Native INT4 model outputs English without space by bigdl-llm0808 or 0809. 
OS: windows 11, 13900H

```
Output =  alongtheway.Oneday,sheheardaboutafestivalcalled"TheColorfulAdventure".

Thefestivalwasallaboutcelebratetheendoftheyearandthestartofanewone.Thegirlwasfascinatedbytheidea.Shedecidedtogoandhaveablast.

Shepackedsomesnacksandherfavoritetoy,andsetoffonherjourney.Shetraveledthroughthecity,takinginallthesightsandsounds.Shemetsomenewfriendsalongtheway,andhadagreattime.

Atthefestival,thecolorfuldecorationsandlivelymusicfilledtheair.Thegirl'sheartwasfilledwithjoyasshetookpartinthecelebrations.Shedancedandsangalongtothemusic,andevenwonaprizeinagame.

Thefestivalwasagreatadventureforthegirl,andshecouldn'twaittoseewhatthenewyearhadinstore.Sheknewthatshewouldhavemanymoreadventuresandexcitingexperiences,andshewasreadyforthem.

Theend.
```

Attach code:

```
from bigdl.llm.ggml.model.chatglm.chatglm import ChatGLM
model = ChatGLM("D:\\PC_LLM_UI\\benchmark\\checkpoint\\ggml-chatglm2-6b-q4_0.bin", n_threads=20)
response = ""
prompt = "Once upon a time, there existed a little girl who liked to have adventures. She wanted to go to places and meet new people, and have fun"
for chunk in model(prompt, temperature=0.95,top_p=0.8, stream=True,max_tokens=512):
    response += chunk['choices'][0]['text']
    #print(chunk['choices'][0]['text'])
print("Output = ",response)
```

PR #8733, opened 2023-08-11
LLM: fix ChatGLM2 native int4 stream output

plusbang:
## Description

Fix https://github.com/intel-analytics/BigDL/issues/8715. Besides, when `stream=False` and prompt is not end with space, the generation output is not starting with space. 

Code:
```python
from bigdl.llm.ggml.model.chatglm.chatglm import ChatGLM
model = ChatGLM(CONVERTED_MODEL_PATH, n_threads=20)
prompt = "Once upon"

response = ""
# output
response = model(prompt, temperature=0.95,top_p=0.8, stream=False,max_tokens=32)
print(f"Output = '{response['choices'][0]['text']}'")

response = ""
# stream output
for chunk in model(prompt, temperature=0.95,top_p=0.8, stream=True,max_tokens=32):
    response += chunk['choices'][0]['text']
print(f"Stream Output = '{response}'")
```
Sample output:
```bash
Output = 'Once upon a time in a land far away, there lived a man named Jack. Jack was a farmer, and he lived in a small village surrounded by fields and forests'
Stream Output = ' a time, in a land far, far away, there lived a wise old man. The wise old man had a long white beard and a bushy white'
```

### 2. User API changes

N/A

### 4. How to test?
- [ ] Unit test
- [x] Local test


MeouSker77 (approved):
LGTM

Issue #8730, opened 2023-08-11
Starcoder Native INT4 model can't stop when set max_tokens


KiwiHana:
Starcoder Native INT4 model can't stop when set max_tokens=512. Need to set stop token? Which is stop token of starcoder?

```
from bigdl.llm.transformers import BigdlNativeForCausalLM
# LLama2 starcoder load 
def load(model_path, model_family, n_threads,n_ctx):
    llm = BigdlNativeForCausalLM.from_pretrained(
        pretrained_model_name_or_path=model_path,
        model_family=model_family,
        n_threads=n_threads,
        n_ctx=n_ctx)
    return llm

model = load(model_path=model_all_local_path + "\\bigdl_llm_starcoder_q4_0.bin",
      model_family='starcoder',
      n_threads=20,n_ctx=4096)

response = ""
prompt = "Write a Python function class Solution to solve the following question.
Given an array of integers, return indices of the two numbers such that they add up to a specific target.
You may assume that each input would have exactly one solution, and you may not use the same element twice.
Example:
Given nums = [2, 7, 11, 15], target = 9,
Because nums[0] + nums[1] = 2 + 7 = 9,
return [0, 1]."


for chunk in model(prompt, temperature=0.95,top_p=0.8, stream=True,max_tokens=512):
    response += chunk['choices'][0]['text']

print(response)
```
output:
token count input:  119
token count output:  539
time cost(s):  172.5168182849884
First token latency(ms):  10017.410278320312
After token latency(ms/token) 302.04350930607455

```
#  
#  
#  In[  ]:
class  Solution(object):
        def  twoSum(self,  nums,  target):
                """
                :type  nums:  List[int]
                :type  target:  int
                :rtype:  List[int]
                """
                for  i  in  range(len(nums)):
                        for  j  in  range(i+1,len(nums)):
                                if  nums[i]+nums[j]==target:
                                        return  [i,j]
#  In[  ]:
<fim_prefix><fim_suffix>n(self.path,  'w')  as  f:
                        f.write(self.content)
        def  __str__(self):
                return  self.path
<fim_middle>import  os
import  shutil
import  tempfile
from.  import  utils
class  File:
        def  __init__(self,  content,  path=None):
                self.content  =  content
                self.path  =  path  or  tempfile.mktemp()
        def  __enter__(self):
                self.create()
                return  self
        def  __exit__(self,  exc_type,  exc_val,  exc_tb):
                self.remove()
        def  create(self):
[...]

PR #8734, opened 2023-08-11
LLM: support stop for starcoder native int4 stream

rnwang04:
## Description

Support stop words for starcoder native int4 stream

### 1. Why the change?

Fix https://github.com/intel-analytics/BigDL/issues/8730

### 2. User API changes

No change.

### 3. Summary of the change 

Support stop words for starcoder native int4 stream

### 4. How to test?
- [x] Unit test

## some sample output
```python
response = ""
prompt = "Write a Python function to generate the nth Fibonacci number"

for chunk in model(prompt, stream=True, max_tokens=512, stop=['<fim_prefix>', '<fim_middle>']):
    response += chunk['choices'][0]['text']

print(response)
```

```bash
.

# The Fibonacci numbers are the numbers in the following integer sequence.
# 0, 1, 1, 2, 3, 5, 8, 13, 21, 34, 55, 89, 144, ……..

# In mathematical terms, the sequence Fn of Fibonacci numbers is defined by the recurrence relation
# Fn = Fn-1 + Fn-2 where F1 = 0 and F2 = 1.

# Example:
# Input: 4
# Output: 3
# Explanation:
# Fibonacci numbers are:
# 0, 1, 1, 2, 3, 5, 8, 13, 21, 34
# Hence, the 4th Fibonacci number is 3.

# Solution:
# 1. Recursive
# 2. Iterative
# 3. Dynamic Programming

# Recursive
class Solution:
    def fib(self, n: int) -> int:
        if n == 0:
            return 0
        if n == 1:
            return 1
        return self.fib(n-1) + self.fib(n-2)

# Iterative
class Solution:
    def fib(self, n: int) -> int:
        if n == 0:
            return 0
        if n == 1:
            return 1
        a, b = 0, 1
        for _ in range(n-1):
            a, b = b, a+b
        return b

# Dynamic Programming
class Solution:
    def fib(self, n: int) -> int:
        if n == 0:
            return 0
        if n == 1:
            return 1
        dp = [0, 1]
        for _ in range(2, n+1):
            dp.append(dp[-1] + dp[-2])
        return dp[-1]
```

### DIFF
--- a/python/llm/src/bigdl/llm/ggml/model/starcoder/starcoder.py
+++ b/python/llm/src/bigdl/llm/ggml/model/starcoder/starcoder.py
@@ -295,6 +295,10 @@ class Starcoder(GenerationMixin):
                 text = self.detokenize([token]).decode("utf-8", errors="ignore")
                 if text.endswith("<|endoftext|>"):
                     print('\n')
+                    return
+                elif text is not None and text in stop:
+                    print('\n')
+                    return
                 else:
                     yield {
                         "id": completion_id,


shane-huang (approved):
LGTM.

PR #8763, opened 2023-08-16
LLM: reduce 1st token latency and update example

rnwang04:
## Description

I discovered that removing `half()` can further improve 2+ token's performance, but will hurt 1st token latency.
So here force first token use torch.float16 and update related examples.

### 1. Why the change?

https://github.com/analytics-zoo/nano/issues/411#issuecomment-1679938137

### 2. User API changes
```python
from bigdl.llm.transformers import AutoModel
model = AutoModel.from_pretrained(args.model_dir, load_in_4bit=True, optimize_model=False, trust_remote_code=True)
model = model.to('xpu')
with torch.inference_mode():
     inputs = tokenizer([prompt], return_tensors="pt").to('xpu')
     output = model.generate(**inputs, do_sample=False, max_new_tokens=32)
```

### 3. Summary of the change 

- update example about new usage
- force first token always use torch.float16
- update readme of GPU benchmark script

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/llm/dev/benchmark/README.md
+++ b/python/llm/dev/benchmark/README.md
@@ -54,9 +54,16 @@ tokenizer = AutoTokenizer.from_pretrained(model_path, trust_remote_code=True)
 prompt = "今天睡不着怎么办"
  
 with torch.inference_mode():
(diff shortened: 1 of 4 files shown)


cyita (approved):
LGTM

PR #8766, opened 2023-08-16
Re organize llm test

liu-shaojun:
This submission adds llm-example-test into llm-nightly-test.yml so that we only need to monitor the status of llm-nightly-test.yml every workday.


hkvision (approved):

hoshibara (approved):
LGTM

PR #8779, opened 2023-08-21
[LLM] change default n_threads of native langchain API

MeouSker77:
## Description

change default n_threads of native langchain API  from 2 to all logical cores


rnwang04 (approved):
LGTM

PR #8790, opened 2023-08-23
[PPML]fix: output mr_enclave to customer in Gramine building

ShanSimu:
## Description

fix wrong usage of output mr_enclave to customer in Gramine building


xiangyuT (approved):
LGTM

PR #8794, opened 2023-08-24
[LLM] Fix error when runner installing bigdl-llm

hoshibara:
## Description

[Log](https://github.com/intel-analytics/BigDL/actions/runs/5951694068/job/16163881052#step:27:697)

Workflow throw an error when installing bigdl-llm package. Maybe it caused by pip.


hoshibara:
Fix with installing rust on windows runner

PR #8796, opened 2023-08-24
[LLM] Performance test

Litchilitchy:
## Description

First performance test, using LLAMA2
* baselines are set on user servers which are dedicated to run benchmark tests, others without server auth could not access these baselines
* added a `do_print` option in benchmark utils to not print sensitive performance results
* the code of performance tests are usually the counter part of example test plus `BenchmarkWrapper`


python/llm/dev/benchmark/pipelines/llama2_test.py
```diff
@@ -0,0 +1,94 @@
...
+
+### RESPONSE:
+"""
+
+if __name__ == '__main__':
+
+    parser = argparse.ArgumentParser(description='Predict Tokens using `generate()` API for Llama2 model')
+    parser.add_argument('--repo-id-or-model-path', type=str, default="meta-llama/Llama-2-7b-chat-hf",
+                        help='The huggingface repo id for the Llama2 (e.g. `meta-llama/Llama-2-7b-chat-hf` and `meta-llama/Llama-2-13b-chat-hf`) to be downloaded'
+                                ', or the path to the huggingface checkpoint folder')
```
hkvision, line 46: this part is not well formatted.

hkvision (approved):

hkvision:
We can merge this as the first version.

python/llm/dev/benchmark/pipelines/llama2_test.py
```diff
@@ -0,0 +1,94 @@
+#
```
hkvision, line 1: We may need to better design this. It is not convenient to add one test py for each model.
Probably one naive way is to add benchmarkwrapper using commands (e.g. sed), grep the output time and do the comparison in bash file?

Litchilitchy: It would be great to use one script to test all the models, but they don't have the same inputs, and some like whisper even needs extra dependencies and datasets...

python/llm/dev/benchmark/pipelines/llama2_test.py
Litchilitchy: I missed this comment.. will refine in the next PR

hkvision: Sure.

python/llm/dev/benchmark/pipelines/llama2_test.py
hkvision: I mean probably we can use one script like testing examples, and in the script we run the generate.py of all models, grep the result and do the perf test?

Litchilitchy: Yeah this is possible

PR #8801, opened 2023-08-24
update illustration in PPML readme

Uxito-Ada:
## Description

update illustration in PPML readme

### 1. Why the change?

add latest features into doc

### 2. User API changes

no

### 3. Summary of the change 

as above


glorysdj (approved):
LGTM

Issue #8802, opened 2023-08-24
voiceassistant tranformersint4 example does not work


nedo99:
Hi,

Tried running transformers_int4 example voice assistant example from [here ](https://github.com/intel-analytics/BigDL/blob/main/python/llm/example/langchain/transformers_int4/voiceassistant.py) with the command:
```
python transformers_int4/voiceassistant.py -r openai/whisper-medium -m meta-llama/Llama-2-7b-chat-hf
```
but getting following error:
```
_PopenSelector = selectors.SelectSelector
AttributeError: partially initialized module 'selectors' has no attribute 'SelectSelector' (most likely due to a circular import)
```

nedo99:
Another issue with this example is that this part:
```
# Following code are complete the same as the use-case
    voiceassitant_chain = LLMChain(
        llm=llm,
        prompt=prompt,
        verbose=True,
        llm_kwargs={"max_new_tokens":args.max_new_tokens},
        memory=ConversationBufferWindowMemory(k=2),
    )
```

throws an exception:
```
pydantic.error_wrappers.ValidationError: 1 validation error for LLMChain
llm_kwargs
  extra fields not permitted (type=value_error.extra)

```

Removing ` llm_kwargs={"max_new_tokens":args.max_new_tokens},` solves the issue, but then the max tokens is much bigger.

jason-dai:
@yangw1234 please take a look

nedo99:
> For the first issue, it is because there `math.py` in `transformer_int4/` directory causing a name collision. We will fixed it. You can try moving the `voiceassistant.py` file to another directory and run again.
> 
> For the second issue, it looks like the `langchain` version does not match. Would you mind sharing which version are you using?

Sure:
```
Name: langchain
Version: 0.0.184
Summary: Building applications with LLMs through composability
Home-page: https://www.github.com/hwchase17/langchain
Author: 
Author-email: 
License: MIT
Location: /home/nhadzic/miniconda3/envs/llm/lib/python3.9/site-packages
Requires: aiohttp, async-timeout, dataclasses-json, numexpr, numpy, openapi-schema-pydantic, pydantic, PyYAML, requests, SQLAlchemy, tenacity

```

yangw1234:
> LLMChain

I think upgrading langchain will fix the issue.

The langchain library evolves really fast. The `llm_kwargs` is added in `langchain==0.0.205` and the latest version is `langchain==0.0.273`.

nedo99:
[...]

PR #8805, opened 2023-08-24
rename math.py in example to avoid conflict

yangw1234:
## Description

rename math.py in example to avoid conflict

fix issue reported in https://github.com/intel-analytics/BigDL/issues/8802


jason-dai (approved):
LGTM

PR #8807, opened 2023-08-25
[LLM] Fix llm test workflow repeatedly download model files

hoshibara:
## Description

Fix bash commands which cause converted llm models are repeatedly downloaded.


Oscilloscope98 (approved):
LGTM

PR #8827, opened 2023-08-28
[LLM] add new API for optimize any pytorch models

shane-huang:
## Description

Add a new API for optimize any PyTorch models

### 1. Why the change?

Using `optimize_model`, low-bit optimizations can be applied on any pytorch model, not just transformers. 

### 2. User API changes

```python
#create any pytorch models
model = ...
###  add the following 2 lines
from bigdl.llm import optimize_model
model = optimize_model(model, low_bit="sym_int4", optimize_llm=False)
# continue to use the model
model.forward(...)
```

### 3. Summary of the change 
Effect of using `optimize_model` on OpenAI Whisper tiny model.
```python
import whisper
model = whisper.load_model("tiny")
st = time.time()
result = model.transcribe("audio.mp3", verbose=True, language="English")
print(result["text"])
print(f"Using {time.time()-st}s")
```
Before:
```
[00:00.000 --> 00:04.000]  Book Me a Reservation for Midday at French Camp Academy.
 Book Me a Reservation for Midday at French Camp Academy.
Using 3.2975785732269287s
```
After:
```
[00:00.000 --> 00:04.000]  Book Me a Reservation for Midday at French Camp Academy.
 Book Me a Reservation for Midday at French Camp Academy.
Using 0.4424586296081543s
```

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

### 5. New dependencies
n/a

### DIFF
--- a/python/llm/src/bigdl/llm/optimize.py
+++ b/python/llm/src/bigdl/llm/optimize.py
@@ -18,19 +18,20 @@ from .transformers import ggml_convert_quant
 from bigdl.llm.ggml.quantize import ggml_tensor_qtype
 from bigdl.llm.utils.common import invalidInputError
 
+
 def optimize_model(model, low_bit, optimize_llm=True):
     """
     A method to optimize any pytorch models.
-    
+
     :param model: The original PyTorch model (nn.module)
-    :param low_bit: Supported low-bit options are "sym_int4", "asym_int4", "sym_int5", 
+    :param low_bit: Supported low-bit options are "sym_int4", "asym_int4", "sym_int5",
         "asym_int5" or "sym_int8".
     :param optimize_llm: Whether to further optimize llm model.
-        
+
     return: The optimized model.
     """
     invalidInputError(low_bit in ggml_tensor_qtype,
-                    f"Unknown load_in_low_bit value: {low_bit}, expected:"
-                    f" sym_int4, asym_int4, sym_int5, asym_int5 or sym_int8.")
+                      f"Unknown load_in_low_bit value: {low_bit}, expected:"
+                      f" sym_int4, asym_int4, sym_int5, asym_int5 or sym_int8.")
     qtype = ggml_tensor_qtype[low_bit]
     return ggml_convert_quant(model, qtype=qtype, optimize_model=optimize_llm)
--- a/python/llm/src/bigdl/llm/optimize.py
+++ b/python/llm/src/bigdl/llm/optimize.py
@@ -1,0 +1,36 @@
+#
+# Copyright 2016 The BigDL Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


python/llm/src/bigdl/llm/optimize.py
```diff
@@ -0,0 +1,36 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from .transformers import ggml_convert_quant
+from bigdl.llm.ggml.quantize import ggml_tensor_qtype
+from bigdl.llm.utils.common import invalidInputError
+
+def optimize_model(model, load_in_low_bit, optimize_atten=True):
```
jason-dai, line 21: change to optimize_model(model, low_bit='sym_int4', optimize_llm=True)

python/llm/src/bigdl/llm/optimize.py
```diff
@@ -18,19 +18,19 @@
 from bigdl.llm.ggml.quantize import ggml_tensor_qtype
 from bigdl.llm.utils.common import invalidInputError
 
-def optimize_model(model, load_in_low_bit, optimize_atten=True):
+def optimize_model(model, low_bit, optimize_llm=True):
```
jason-dai, line 22: low_bit default to 'sym_int4'

python/llm/src/bigdl/llm/optimize.py
shane-huang: ok. I'll also change `optimize_llm=False`, since it may not apply to all models (e.g. latest transformer version)

jason-dai: > I'll also change `optimize_llm=False`, since it may not apply to all models (e.g. latest transformer version)

You should keep its behavior the same as `from_pretrained`, which I believe defaults to True.

python/llm/src/bigdl/llm/optimize.py
```diff
@@ -19,7 +19,7 @@
 from bigdl.llm.utils.common import invalidInputError
 
 
-def optimize_model(model, low_bit, optimize_llm=True):
+def optimize_model(model, low_bit='sym_int4', optimize_llm=False):
```
jason-dai, line 22: what's the default value of `pretrained`? We should make it consistent

shane-huang: I'll check with @yangw1234 about what is the preferred default value.

python/llm/src/bigdl/llm/optimize.py
jason-dai: > `from_pretrained` defaults to True and do optimizations aggressively.

I think we need to do the same as it helps llm models

python/llm/src/bigdl/llm/optimize.py
jason-dai: just set it to true

shane-huang: done

jason-dai:
Need to add save/load API, something like:
```python
#save
model = optimize_model(...)
model.save_low_bit(...)
```

```python
#load
model = Net(...)
load_low_bit(model, ...)
```

@leonardozcm

PR #8838, opened 2023-08-29
support transformer int4 + amx int4

MeouSker77:
## Description

Note:
1. AMX will be enabled automatically on SPR
2. AMX cannot be used with bf16 (at least for now), means we must disable autocast on SPR


rnwang04 (approved):
LGTM

Issue #8837, opened 2023-08-29
Bark-small load_in_4bit got error


violet17:
Dear,
Thanks for your great work.

I run bark small(https://huggingface.co/suno/bark-small) on A770， and  get error when using load_in_4bit.
Here are my codes:

```
import torch
from transformers import AutoProcessor
import intel_extension_for_pytorch as ipex
from bigdl.llm.transformers import AutoModel
import scipy
import time

model_path = "./bark-small"
processor = AutoProcessor.from_pretrained(model_path)
model = AutoModel.from_pretrained(model_path,load_in_4bit=True, optimize_model=False)
model = model.to("xpu")

text = "Hello, my name is Suno. And, uh — and I like pizza. [laughs] But I also have other interests such as playing tic tac toe."
t0 = time.time()
with torch.inference_mode():
    inputs = processor(text=[text], return_tensors="pt")
    t1 = time.time()
    print("Cost time--1: ",  t1- t0)
    inputs = inputs.to("xpu")
    speech_values = model.generate(**inputs, do_sample=True)
    print("Cost time--2: ", time.time() - t1)

print("Cost time: ", time.time() - t0)

sampling_rate = model.generation_config.sample_rate
scipy.io.wavfile.write("bark_out.wav", rate=sampling_rate, data=speech_values.cpu().numpy().squeeze())
print("save audio to bark_out.wav---")
```
Running log:
```
$ python bark_bigdl.py 
/home/adc2/miniconda3/envs/llm/lib/python3.9/site-packages/torchvision/io/image.py:13: UserWarning: Failed to load image Python extension: ''If you don't plan on using image functionality from `torchvision.io`, you can ignore this warning. Otherwise, there might be something wrong with your environment. Did you have `libjpeg` or `libpng` installed before building `torchvision` from source?
  warn(
Traceback (most recent call last):
  File "/home/adc2/crystal/llm/bark_bigdl.py", line 13, in <module>
    model = AutoModel.from_pretrained(model_path,load_in_4bit=True, optimize_model=False)
  File "/home/adc2/miniconda3/envs/llm/lib/python3.9/site-packages/bigdl/llm/transformers/model.py", line 82, in from_pretrained
    model = cls.load_convert(q_k, optimize_model, *args, **kwargs)
  File "/home/adc2/miniconda3/envs/llm/lib/python3.9/site-packages/bigdl/llm/transformers/model.py", line 96, in load_convert
[...]

PR #8841, opened 2023-08-29
LLM: Enable attempting loading method automatically

leonardozcm:
## Description

https://github.com/intel-analytics/BigDL/issues/8837#issuecomment-1696883932

We are enforcing the experimental feature "low_cpu_mem_usage" to alleviate memory pressure when loading large models. This feature avoids initializing the parameters when initializing empty model itself (by put everything on a meta device). However,  some models like the bark contains a operation named "weights_norm" , which normalizes the weights during initialization, necessitating non-empty weights. This conflicts with the design of the meta device, and this is why "low_cpu_mem_usage" doesn't work on bark.

So in this PR,  we first load model with "low_mem_cpu_usage=True", if an exception is caught, then throw a warning and proceed to handle it using the original loading method.
```log
(changmim-llm) cpx@cpx-2:/disk1/changmin/BigDL/python/llm$ python test_bark.py 
Failed to load models with `low_cpu_mem_usage` specified, will fall to traditional load method with higher memory consumption.

/disk1/changmin/BigDL/python/llm/src/bigdl/llm/transformers/convert.py:103: UserWarning: No linear modules were found in your model. This can happen for some architectures such as gpt2 that uses Conv1D instead of Linear layers. Please double check your model architecture, or submit an issue on github if you think this is a bug.
  warnings.warn(
Cost time--1:  0.009249687194824219
The attention mask and the pad token id were not set. As a consequence, you may observe unexpected behavior. Please pass your input's `attention_mask` to obtain reliable results.
Setting `pad_token_id` to `eos_token_id`:10000 for open-end generation.
Cost time--2:  36.95219945907593
Cost time:  36.9615421295166
save audio to bark_out.wav---
```

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/convert.py
+++ b/python/llm/src/bigdl/llm/transformers/convert.py
@@ -85,12 +85,13 @@ def _replace_with_quant_linear(model, qtype, modules_to_not_convert=None,
 
         # Remove the last key for recursion
         if len(list(module.children())) > 0:
-            _, has_been_replaced = _replace_with_quant_linear(
+            _, _flag = _replace_with_quant_linear(
                 module,
                 qtype,
                 modules_to_not_convert,
                 current_key_name,
             )
+            has_been_replaced = _flag or has_been_replaced
     return model, has_been_replaced
 
 
--- a/python/llm/src/bigdl/llm/transformers/model.py
+++ b/python/llm/src/bigdl/llm/transformers/model.py
@@ -22,6 +22,10 @@ from .utils import extract_local_archive_file, \
 from bigdl.llm.ggml.quantize import ggml_tensor_qtype
 from bigdl.llm.utils.common import invalidInputError
 import torch
+import copy
+import logging
+
+logger = logging.getLogger(__name__)
 
 
 def save_low_bit(self, *args, **kwargs):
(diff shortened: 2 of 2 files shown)


python/llm/src/bigdl/llm/transformers/convert.py
```diff
@@ -85,12 +85,13 @@ def _replace_with_quant_linear(model, qtype, modules_to_not_convert=None,
...
         # Remove the last key for recursion
         if len(list(module.children())) > 0:
-            _, has_been_replaced = _replace_with_quant_linear(
+            _, _flag = _replace_with_quant_linear(
                 module,
                 qtype,
                 modules_to_not_convert,
                 current_key_name,
             )
+            has_been_replaced = _flag or has_been_replaced
```
leonardozcm, line 94: Otherwise has_been_replaced will always be overwritten by the last one iteration.

python/llm/src/bigdl/llm/transformers/model.py
```diff
@@ -98,7 +102,19 @@ def load_convert(cls, q_k, optimize_model, *args, **kwargs):
...
-        model = cls.HF_Model.from_pretrained(*args, **kwargs)
+        # In case it needs a second try,
+        # `from_pretrained`` may pop items out in dict
+        # and lead to args missing.
+        _args = copy.deepcopy(args)
+        _kwargs = copy.deepcopy(kwargs)
+        try:
+            model = cls.HF_Model.from_pretrained(*args, **kwargs)
+        except NotImplementedError:
+            logger.warning("Failed to load models with `low_cpu_mem_usage` specified, "
```
jason-dai, line 113: info instead of warning

PR #8731, opened 2023-08-11
LLM: Implement hf `low_cpu_mem_usage` with 1xbinary file peak memory on transformer int4

leonardozcm:
## Description

**`load_low_bit_cpumem` will be change to `load_low_bit` when everything works well as before**

 https://github.com/analytics-zoo/nano/issues/506
The code is extracted from https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L2073
In this PR, we have implemented the mechanism to load the Transformer INT4 version using only a 1X model size (including peak memory), taking inspiration from the `from_pretrained` method in Hugging Face's library.

        * `low_cpu_mem_usage` algorithm:

        This is an experimental function that loads the model using ~1x model size CPU memory

        Here is how it works:

        1. load state_dict for the first time and save which state_dict keys we have
        2. drop state_dict before the model is created, since the latter takes 1x model size CPU memory
        3. after the model has been instantiated switch to the meta device all params/buffers that
        are going to be replaced from the loaded state_dict
        4. load state_dict 2nd time
        5. replace the params/buffers from the state_dict

        Currently, it can't handle deepspeed ZeRO stage 3 and ignores loading errors

There is little difference between our implement and hf's,  for we save all  state_dict keys at `save_low_bit` in a single file, and we can just read that file when loading low bit model so that we don't need to repeat step 1~2 mentioned above.

## Experiments
This PR involves a lot of changes, so we may tested on models as more as possible.

1. Baichuan13B(int4 model with 7.7GB size)
Before this PR we may need at least 17GB(at peak memory) to load a int4 model, for we may load original model(+1x model size) ,convert and load .bin from disk again(+ 1x model size), so at peak memory it may take at least 2* model size memory.

After this PR we make assigning model weights a in-place operation, and it will cost 8.1GB memory to load int4 model at most.
```log
torch dtype torch.bfloat16
--------------------before hf loading--------------------

Used: 130.70 GB
Increased: 0.01 GB
len(model) is 283
--------------------before ggml convert--------------------

Used: 130.70 GB
Increased: 0.00 GB
--------------------after ggml convert--------------------

Used: 130.70 GB
Increased: 0.00 GB
len(model) is 283
--------------------before model_class._load_pretrained_model--------------------

Used: 130.70 GB
Increased: 0.00 GB

Memory Usagebefore _load_pretrained_model :
Used: 130.70 GB


Memory Usagebefore load params :
Used: 130.70 GB


Memory Usageafter load params :
[...]

PR #8847, opened 2023-08-30
LLM: Update ut save_low_bit

leonardozcm:
## Description

https://github.com/intel-analytics/BigDL/pull/8731#issue-1846155669
As this description says, our `save_low_bit`'s behavior differ from hf `save_pretrained` now, for we will save an extra load_keys.json to avoid repeating load step.


Oscilloscope98 (approved):
LGTM

PR #8855, opened 2023-08-30
Fix llama attention optimization for XPU

yangw1234:
## Description

Need to allocate KV cache on device.


yangw1234:
perf data listed in https://github.com/analytics-zoo/nano/issues/546

jason-dai (approved):
LGTM

PR #8856, opened 2023-08-31
[LLM] Downgrade amx build gcc version and remove avx flag display

hoshibara:
* Downgrade gcc version to 11 for amx building step
* Remove printing avx flag in llm-cli bash


MeouSker77 (approved):
LGTM

PR #8858, opened 2023-08-31
[PPML] Fix tdx attest vsock port setting

xiangyuT:
## Description

Add `port=4050` when attestation api is needed for tdx coco.


hzjane (approved):
LGTM

PR #8861, opened 2023-08-31
[LLM] add protobuf in bigdl-llm dependency

MeouSker77:
## Description

add protobuf in bigdl-llm dependency


Oscilloscope98 (approved):
LGTM

PR #8877, opened 2023-09-01
LLM: Fix discards in `optimize_model` with non-hf models  and add openai whisper example

leonardozcm:
## Description

Fix the requirement for the source of the model in `optimize()`, for it strongly ask for model's `.config` attribute which only exits in huggingface transformer models.
And add new an exmaple with openai-tiny.

### DIFF
--- a/python/llm/example/pytorch-model/openai-whisper/recognize.py
+++ b/python/llm/example/pytorch-model/openai-whisper/recognize.py
@@ -1,4 +1,5 @@
 import whisper
+import time
 import librosa
 import argparse
 from bigdl.llm import optimize_model
@@ -26,11 +27,15 @@ if __name__ == '__main__':
                             target_sr=target_sr)
 
     # Load whisper model under pytorch framework
-    model = whisper.load_model(args.model_size)
+    model = whisper.load_model(args.model_name)
 
     # With only one line to enable bigdl optimize on a pytorch model
     model = optimize_model(model, low_bit="sym_int4", optimize_llm=True)
 
+    st = time.time()
     result = model.transcribe(audio, verbose=True, language=args.language)
+    end = time.time()
+    print(f'Inference time: {end-st} s')
 
+    print('-'*20, 'Output', '-'*20)
     print(result["text"])
--- a/python/llm/example/pytorch-model/openai-whisper/recognize.py
+++ b/python/llm/example/pytorch-model/openai-whisper/recognize.py
@@ -47,7 +47,7 @@ if __name__ == '__main__':
     model = whisper.load_model(args.model_name)
 
     # With only one line to enable bigdl optimize on a pytorch model
-    model = optimize_model(model, low_bit="sym_int4", optimize_llm=True)
+    model = optimize_model(model, low_bit="sym_int4")
 
     st = time.time()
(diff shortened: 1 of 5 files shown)


python/llm/example/pytorch-model/openai-whisper/recognize.py
```diff
@@ -0,0 +1,58 @@
+import whisper
```
plusbang, line 1: This line maybe removed to L18?

plusbang (approved):
Other LGTM

python/llm/example/pytorch-model/openai-whisper/recognize.py
```diff
@@ -0,0 +1,58 @@
...
+    target_sr = 16000
+    audio = librosa.resample(y,
+                            orig_sr=sr,
+                            target_sr=target_sr)
+
+    # Load whisper model under pytorch framework
+    model = whisper.load_model(args.model_name)
+
+    # With only one line to enable bigdl optimize on a pytorch model
+    model = optimize_model(model, low_bit="sym_int4", optimize_llm=True)
```
jason-dai, line 50: optimize_llm should default to true

python/llm/src/bigdl/llm/transformers/convert.py
```diff
@@ -114,7 +115,11 @@ def ggml_convert_quant(model, qtype, optimize_model=True, device="cpu"):
         pass
 
     if optimize_model:
-        model = optimize(model)
+        try:
+            model = optimize(model)
+        except:
+            logger.info("It is not supported that optimizing a model isn't belong to"
```
jason-dai, line 121: what do you mean?

leonardozcm: `optimize()`  is only compatible with the Hugging Face API, and take effects on `chatglm` and `llama` family more specifically.
We apply `optimize()` on other pytorch model is likely to fail, `optimize()` strongly assume it is [a hf format model](https://github.com/intel-analytics/BigDL/blob/main/python/llm/src/bigdl/llm/transformers/convert.py#L142):
```python
if "chatglm2" in model.config._name_or_path:
```

Do we need to explore a way more general to extend its working range?

leonardozcm: I see

jason-dai: If there is any issue, you should fix `optimize` instead

python/llm/src/bigdl/llm/transformers/convert.py
```diff
@@ -129,6 +130,14 @@ def convert_forward(m, target_m, new_forward):
 def optimize(model):
     from packaging import version
     from bigdl.llm.transformers.models.llama import llama_attention_forward_4_31
+    from transformers.modeling_utils import PreTrainedModel
+
+    # All huggingface format models are inherited from `PreTrainedModel`
+    if not isinstance(model, PreTrainedModel):
+        logger.info("It is not supported that optimizing a model isn't belong to"
+                    "huggingface transformers with `optimize_model=True` for now.")
```
jason-dai, line 138: Only HuggingFace Transformers models are currently supported for further optimizations

Issue #8870, opened 2023-08-31
Pytorch/TF Estimator with Ray backend cannot run on Databricks


jenniew:
When create PyTorch/TF Estimator with Ray backend, it will give this error:

Py4JJavaError                             Traceback (most recent call last)
<command-3809454374114932> in <module>
     17 criterion = nn.CrossEntropyLoss()
     18 
---> 19 estimator = Estimator.from_torch(model=model_creator,
     20                                  optimizer=optimizer_creator,
     21                                  loss=criterion,

/databricks/python/lib/python3.8/site-packages/bigdl/orca/learn/pytorch/estimator.py in from_torch(model, optimizer, loss, metrics, backend, config, workers_per_node, scheduler_creator, use_tqdm, model_dir, sync_stats, log_level, log_to_driver)
     92         if backend in {"horovod", "ray"}:
     93             from bigdl.orca.learn.pytorch.pytorch_ray_estimator import PyTorchRayEstimator
---> 94             return PyTorchRayEstimator(model_creator=model,
     95                                        optimizer_creator=optimizer,  # type:ignore
     96                                        loss_creator=loss,

/databricks/python/lib/python3.8/site-packages/bigdl/orca/learn/pytorch/pytorch_ray_estimator.py in __init__(self, model_creator, optimizer_creator, loss_creator, metrics, scheduler_creator, config, use_tqdm, backend, workers_per_node, sync_stats, log_level)
    109 
    110         # todo remove ray_ctx to run on workers
--> 111         ray_ctx = OrcaRayContext.get()
    112         if not isinstance(model_creator, types.FunctionType):  # Torch model is also callable.
    113             invalidInputError(False,

/databricks/python/lib/python3.8/site-packages/bigdl/orca/ray/raycontext.py in get(cls, initialize)
    101             ray_ctx = OrcaRayContext._active_ray_context
    102             if initialize and not ray_ctx.initialized:
--> 103                 ray_ctx.init()
    104             return ray_ctx
    105         else:

/databricks/python/lib/python3.8/site-packages/bigdl/orca/ray/raycontext.py in init(self, driver_cores)
     75                 results = ray.init(**self.ray_args)
     76         else:
---> 77             results = self._ray_on_spark_context.init(driver_cores=driver_cores)
     78             self.num_ray_nodes = self._ray_on_spark_context.num_ray_nodes  # type: ignore
[...]

PR #8879, opened 2023-09-01
[ORCA] Hot-fix pytorch ray beckend error on databricks cluster

sgwhat:
## Description
Fix https://github.com/intel-analytics/BigDL/issues/8870.

### 1. Why the change?
`python_loc` is a file but not a directory in databricks, that's the reason of this issue.

### 2. User API changes
N/A

### 3. Summary of the change 
`os.path.isdir` -> `os.path.exists`.

### 4. How to test?
- [ ] Unit test
- [ ] Application test


sgwhat:
One ut failed but not related.

PR #8883, opened 2023-09-04
Add two parameters to backend entrypoint.

Hanyu-Jin:
## Description

Add two parameters to backend entrypoint.

### 1. Why the change?

Occupy too much epcm may lead to error in SGX.

### 2. User API changes

None.

### 3. Summary of the change 

Set `SAVED_ON_DISK=true` to save model files on disk and set `SECURED_DIR` to set the saved directory.


xiangyuT (approved):
LGTM

PR #8871, opened 2023-09-01
transformer int4 and native int4's benchmark script for 32 256 1k 2k input

qiuxin2012:
## Description

transformer int4 and native int4's benchmark script for 32 256 1k 2k input

### 1. Why the change?

transformer int4 and native int4's benchmark script for 32 256 1k 2k input

### 2. User API changes

None

### 3. Summary of the change 

Usage:
```
python native_benchmark.py --repo-id meta-llama/Llama-2-7b-chat-hf --local-model-hub ~/IntelAnalytics/models/ --prompt-len 2048  --infer-times 2  --save-path /tmp/llama-7b

python transformers_benchmark.py --repo-id 'THUDM/chatglm2-6b' --local-model-hub ~/IntelAnalytics/models --prompt-len 1024 --max-tokens 128   --infer-times 2
```
The scripts will load models from local-model-hub if it's provided.

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

### 5. New dependencies

- [ ] New Python dependencies
       - Dependency1 
       - Dependency2
       - ...
- [ ] New Java/Scala dependencies and their license
       - Dependency1 and license1
       - Dependency2 and license2
       - ...

PR #8899, opened 2023-09-06
[llm]Merge benchmark_utils.py and gpu_benchmark_utils.py

qiuxin2012:
## Description

Merge benchmark_utils.py and gpu_benchmark_utils.py

### 1. Why the change?

Too many duplicated code.

### 2. User API changes

None

### 3. Summary of the change 

Merge benchmark_utils.py and gpu_benchmark_utils.py.

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...


python/llm/dev/benchmark/benchmark_util.py
```diff
@@ -510,9 +510,12 @@ class BenchmarkWrapper:
     learn more about decoding strategies refer to the [text generation strategies guide](../generation_strategies).
     """
     
-    def __init__(self, model, do_print=True):
+    def __init__(self, model, do_print=False):
```
rnwang04, line 513: shall we leave the default value as true?

python/llm/dev/benchmark/benchmark_util.py
```diff
@@ -510,9 +510,12 @@ class BenchmarkWrapper:
...
         self.model = model
         self.do_print = do_print
+        self.encoder_time = 0.0
+        self.first_cost = 0.0
+        self.rest_cost_mean = 0.0
```
rnwang04, line 518: it seems that these three variables may be unnecessary ?

rnwang04 (commented):
others LGTM

python/llm/dev/benchmark/benchmark_util.py
qiuxin2012: To resolve pycharm's warming: `instance attribute defined outside __init__`

python/llm/dev/benchmark/benchmark_util.py
qiuxin2012: > shall we leave the default value as true?
From the disscusion in https://github.com/intel-analytics/BigDL/pull/8871, we want to hide the print by default.

rnwang04 (approved):
LGTM

Issue #8895, opened 2023-09-05
BigDL AutoProphet function not working fine when executing spark-submit to AWS EKS cluster 


SjeYinTeoIntel:
Facing issue when running below codes.
from bigdl.orca import init_orca_context
init_orca_context(cluster_mode="local", cores=1)
forecaster = AutoProphet(optimizer=opt, loss=los)
_______________________________________________________________________________________________________________________
Logs:
23/09/05 10:45:09 WARN SparkContext: Using an existing SparkContext; some configuration may not take effect.
23-09-05 10:45:09 [Thread-4] INFO  Engine$:461 - Find existing spark context. Checking the spark conf...
23-09-05 10:45:09 [Thread-4] WARN  Engine$:467 - Engine.init: Can not find spark.shuffle.reduceLocality.enabled. For details please check https://bigdl-project.github.io/master/#APIGuide/Engine/
23-09-05 10:45:09 [Thread-4] WARN  Engine$:467 - Engine.init: Can not find spark.shuffle.blockTransferService. For details please check https://bigdl-project.github.io/master/#APIGuide/Engine/
23-09-05 10:45:09 [Thread-4] WARN  Engine$:467 - Engine.init: Can not find spark.scheduler.minRegisteredResourcesRatio. For details please check https://bigdl-project.github.io/master/#APIGuide/Engine/
23-09-05 10:45:09 [Thread-4] WARN  Engine$:467 - Engine.init: Can not find spark.scheduler.maxRegisteredResourcesWaitingTime. For details please check https://bigdl-project.github.io/master/#APIGuide/Engine/
23-09-05 10:45:09 [Thread-4] WARN  Engine$:467 - Engine.init: Can not find spark.speculation. For details please check https://bigdl-project.github.io/master/#APIGuide/Engine/
23-09-05 10:45:09 [Thread-4] WARN  Engine$:467 - Engine.init: Can not find spark.serializer. For details please check https://bigdl-project.github.io/master/#APIGuide/Engine/
23-09-05 10:45:09 [Thread-4] WARN  Engine$:470 - Engine.init: spark.driver.extraJavaOptions should be -Dlog4j2.info, but it is -Dcom.amazonaws.services.s3.enableV4=true. For details please check https://bigdl-project.github.io/master/#APIGuide/Engine/
cls.getname: com.intel.analytics.bigdl.dllib.utils.python.api.Sample
BigDLBasePickler registering: bigdl.dllib.utils.common  Sample
[...]

PR #8901, opened 2023-09-06
[ORCA] Hot-fix ray_tune_search issue on k8s cluster

sgwhat:
## Description
Fix issue https://github.com/intel-analytics/BigDL/issues/8895 as a temp solution, we will add support for k8s later.

### 1. Why the change?

Ray_tune_search requires to create hdfs directories on cluster, which caused error on k8s.

### 2. User API changes
N/A

### 3. Summary of the change 
When the cluster does not support `hadoop` cmd, setup `remote_directory` to None (same as local).

### 4. How to test?
- [x] Unit test
- [x] Application test


hkvision (approved):

sgwhat:
One ut failed but not related.

PR #8902, opened 2023-09-06
Add known issue in arc voice assistant example

cyita:
## Description

Add known issue in arc voice assistant example readme.

### 1. Why the change?
Kiwi noticed that the speech_recognition library may occasionally skip recording due to low volume.


leonardozcm (approved):
LGTM and also add on cpu side.

PR #8906, opened 2023-09-07
[WIP] optimize llama xpu rope

yangw1234:
## Description

optimize llama xpu rope


yangw1234:
performance listed in https://github.com/analytics-zoo/nano/issues/559 and https://github.com/intel-analytics/llm.cpp/pull/73

Issue #8909, opened 2023-09-07
Fail to install latest `bigdl-core-xe` through `pip install --pre --upgrade bigdl-llm[xpu] -f https://developer.intel.com/ipex-whl-stable-xpu`


plusbang:
If there exists old version `bigdl-llm` and `bigdl-core-xe` in conda environment, fail to install the latest `bigdl-core-xe` through `pip install --pre --upgrade bigdl-llm[xpu] -f https://developer.intel.com/ipex-whl-stable-xpu`.

Then there may exists error because the old version of `bigdl-core-xe` as following:
![image](https://github.com/intel-analytics/BigDL/assets/108676127/4138c763-a22c-419c-ab36-be20fc830178)

**Although we can manually `pip install --pre --upgrade bigdl-core-xe` to solve this problem, but we only provide `pip install --pre --upgrade bigdl-llm[xpu] -f https://developer.intel.com/ipex-whl-stable-xpu` in our examples.**

PR #8913, opened 2023-09-07
LLM: update setup to specify bigdl-core-xe version

rnwang04:
## Description

Keep bigdl-core-xe's version is the same with bigdl-llm.

### 1. Why the change?
https://github.com/intel-analytics/BigDL/issues/8909

### 2. User API changes

No change

### 3. Summary of the change 

Keep bigdl-core-xe's version is the same with bigdl-llm.

### 4. How to test?
- [x] Unit test
- [x] Local test


plusbang (approved):
LGTM

PR #8917, opened 2023-09-07
[nano] remove tensorflow dependency

### LINKED ISSUE #8886: Authenticated Local Privilege Escalation vulnerability in Intel Optimization for Tensorflow
Improper buffer restrictions in the Intel(R) Optimization for Tensorflow software before version 2.12 may allow an authenticated user to potentially enable escalation of privilege via local access.

https://github.com/intel-analytics/BigDL/blob/main/python/requirements/nano/requirements_tensorflow.txt

MeouSker77:
## Description

It's hard work to support intel-tensorflow 2.12, so we remove it from our dependency, and require user to install it manually if they want to use tensorflow.

fix #8886

### 2. User API changes

`pip install bigdl-nano[tensorflow]` or command like `pip install bigdl-nano[tensorflow_27/tensorflow_28/stock_tensorflow_27/stock_tensorflow_28]` are no longer supported

### 4. How to test?
- [x] Unit test

### DIFF
--- a/.github/actions/nano/setup-tensorflow-env/action.yml
+++ b/.github/actions/nano/setup-tensorflow-env/action.yml
@@ -48,7 +48,7 @@ runs:
         if [ $OS = "centos" ]; then
           source $CONDA_HOME/bin/activate nano-tensorflow
         fi
-        pip install intel-tensorflow==2.10.0 tf2onnx==1.13.0
+        pip install intel-tensorflow==2.9.1 tf2onnx==1.13.0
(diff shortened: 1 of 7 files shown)


jason-dai:
Do we need to update `nano` document?

MeouSker77:
> Do we need to update `nano` document?

yes, not only document, lots of examples and tutorials are also need to update

rnwang04 (approved):
LGTM

PR #8922, opened 2023-09-07
update llm latency benchmark.

lalalapotter:
## Description

This PR is to update llm latency benchmark.

### DIFF
--- a/python/llm/dev/benchmark/all-in-one/README.md
+++ b/python/llm/dev/benchmark/all-in-one/README.md
@@ -1,7 +1,7 @@
 # All in One Benchmark Test
 All in one benchmark test allows users to test all the benchmarks and record them in a result CSV. Users can provide models and related information in `config.yaml`.
 
-Before running, make sure to have [bigdl-llm](../../../README.md) installed.
+Before running, make sure to have [bigdl-llm](../../../README.md) and [bigdl-nano](../../../../nano/README.md) installed.
 
 ## Config
 Config YAML file has following format
@@ -28,4 +28,10 @@ test_api:
 run `python run.py`, this will output results to `results.csv`.
 
 For SPR performance, run `bash run-spr.sh`.
-For ARC performance, run `bash run-arc.sh`
+> **Note**
+>
+> In `run-spr.sh`, we set optimal environment varaible by `source bigdl-nano-init -c`, `-c` stands for disabling jemalloc. Enabling jemalloc may lead to latency increasement after multiple trials.
+>
+> The value of `OMP_NUM_THREADS` should be the same as the cpu cores specified by `numactl -C`.
+
+For ARC performance, run `bash run-arc.sh`.
--- a/python/llm/dev/benchmark/all-in-one/run-spr.sh
+++ b/python/llm/dev/benchmark/all-in-one/run-spr.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
-
+source bigdl-nano-init -c
+export OMP_NUM_THREADS=48
+export TRANSFORMERS_OFFLINE=1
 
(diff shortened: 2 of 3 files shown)


hkvision (approved):

python/llm/dev/benchmark/all-in-one/run-spr.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
-
+source bigdl-nano-init -c
```
jason-dai, line 2: Do we need `nano` besides `bigdl-nano-init`? If not, we can simple copy the script to `llm` and eliminate the dependency on `nano`

lalalapotter: Sure, maybe we can add a script named `bigdl-llm-init` and exclude jemalloc related part.

python/llm/dev/benchmark/all-in-one/run-spr.sh
```diff
@@ -1,5 +1,7 @@
...
+export OMP_NUM_THREADS=48
+export TRANSFORMERS_OFFLINE=1
 
 # set following parameters according to the actual specs of the test machine
 numactl -C 0-47 -m 0 python $(dirname "$0")/run.py
```
jason-dai, line 7: maybe accept core# as a parameter to automatically set `OMP_NUM_THREADS` and `numactl` values

python/llm/dev/benchmark/all-in-one/README.md
```diff
@@ -1,7 +1,7 @@
 # All in One Benchmark Test
 All in one benchmark test allows users to test all the benchmarks and record them in a result CSV. Users can provide models and related information in `config.yaml`.
 
-Before running, make sure to have [bigdl-llm](../../../README.md) installed.
+Before running, make sure to have [bigdl-llm](../../../README.md) and [bigdl-nano](../../../../nano/README.md) installed.
```
jason-dai, line 4: do we really need `nano`?

python/llm/dev/benchmark/all-in-one/run-spr.sh
lalalapotter: `numactl -C` not only stands for number of cores but also index of cores, so parameter core# can only work for `OMP_NUM_THREADS`. Will try to find a simpler way to set the variables.

python/llm/dev/benchmark/all-in-one/README.md
lalalapotter: I think if we add `bigdl-llm-init` for BigDL-LLM, there should be no need for nano.

python/llm/dev/benchmark/all-in-one/run-spr.sh
jason-dai: always use 0-`core#-1` for numactl

lalalapotter: In this way, we could not specify the socket of CPU. For example, if we are going to use second socket, the command should be `numactl -C 48-95 ...`

jason-dai: I don't think there is such need in bechnmarking

lalalapotter: Sure, then we will add parameter core# and automatically set `OMP_NUM_THREADS` and `numactl` values

python/llm/dev/benchmark/all-in-one/README.md
hkvision: We can add the nano env into llm in a later PR.

PR #8932, opened 2023-09-08
[PPML] Fix incorrect usage in docs of Finetuning to enable TDX

xiangyuT:
### 1. Why the change?

The actual value should be `tdx` instead of `TDX`.


Uxito-Ada (approved):
lgtm

PR #8945, opened 2023-09-11
[PPML] Enable TLS in Attestation API Serving for LLM finetuning

xiangyuT:
## Description

Add `enableTLS` flag to enable TLS in Attestation API Serving for LLM finetuning.


glorysdj (approved):
LGTM

PR #8960, opened 2023-09-12
LLM: add gpu example of chinese-llama-2-7b

rnwang04:
## Description

To provide an arc example of chinese-llama-2-7b.

### 1. Why the change?

To provide an arc example of chinese-llama-2-7b, especially to illustrate setting `use_cache=True` for chinese-llama-2-7b to obatin optimal performance.

### 4. How to test?
- [x] Unit test


shane-huang:
Merge for now and later revise examples README together.

shane-huang (approved):

PR #8965, opened 2023-09-13
LLM: reduce GPU memory for `optimize_model=True`

rnwang04:
## Description

For long sequence generation, we may encounter out of GPU memory issue as large kv_cache.
This PR try to optimize this by manually empty xpu cache to reduce GPU memory for long sequence generation.

### 4. How to test?
- [x] Unit test
- [x] Local test

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/models/chatglm.py
+++ b/python/llm/src/bigdl/llm/transformers/models/chatglm.py
@@ -67,6 +67,8 @@ def attention_fn(
         cache_v = cache_v.permute(1, 2, 0, 3)
         past_length = cache_k.size(2)
         if cache_k.stride()[1] <= cache_k.size(2) * cache_k.size(3):
+            if device == torch.device('xpu:0'):
+                torch.xpu.empty_cache()
             max_cache_length = past_length + cur_length + KV_CACHE_ALLOC_BLOCK_LENGTH
             new_cache_k, new_cache_v = create_kv_cache(batch_size,
                                                        self.num_attention_heads_per_partition,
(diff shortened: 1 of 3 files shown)


python/llm/src/bigdl/llm/transformers/models/chatglm.py
```diff
@@ -67,6 +67,8 @@ def attention_fn(
         cache_v = cache_v.permute(1, 2, 0, 3)
         past_length = cache_k.size(2)
         if cache_k.stride()[1] <= cache_k.size(2) * cache_k.size(3):
+            if device == torch.device('xpu:0'):
```
jason-dai, line 70: how about other variations (e.g., `xpu:1`)?

jason-dai: And what if there are multiple GPU's in the system?

rnwang04: Have revised to judge device type.

jason-dai: > Have revised to judge device type.

If there are multiple GPU's on the same system, will this create correctness problem?

rnwang04: In my understanding, this function just clear unused cache, so it may not cause correctness problem, but I am not very sure.

jason-dai: > In my understanding, this function just clear unused cache

Then it's probably fine

jason-dai (approved):
LGTM

PR #8973, opened 2023-09-14
fix bloom-176b int overflow

MeouSker77:
## Description

same to <https://github.com/intel-analytics/llm.cpp/pull/93>


rnwang04 (approved):
LGTM

PR #8975, opened 2023-09-14
LLM: add `optimize_model` example for bert

plusbang:
## Description

Verified the acceleration of `optimize_model` API on bert-large-uncased model.
Add related example.

### 4. How to test?
- [ ] Unit test
- [x] Local test

### DIFF
--- a/python/llm/example/pytorch-models/bert/README.md
+++ b/python/llm/example/pytorch-models/bert/README.md
@@ -1,0 +1,48 @@
+# BERT
+In this directory, you will find examples on how you could use BigDL-LLM `optimize_model` API to accelerate BERT models. For illustration purposes, we utilize the [bert-large-uncased](https://huggingface.co/bert-large-uncased) as reference BERT models.
+
+## Requirements
+To run these examples with BigDL-LLM, we have some recommended requirements for your machine, please refer to [here](../README.md#recommended-requirements) for more information.
+
+## Example: Extract the feature of given text
+In the example [extract_feature.py](./extract_feature.py), we show a basic use case for a BERT model to extract the feature of given text, with BigDL-LLM INT4 optimizations.
+### 1. Install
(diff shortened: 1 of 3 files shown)


python/llm/example/pytorch-models/bert/extract_feature.py
```diff
@@ -0,0 +1,53 @@
...
+                        help='The huggingface repo id for the BERT (e.g. `bert-large-uncased`) to be downloaded'
+                             ', or the path to the huggingface checkpoint folder')
+    parser.add_argument('--text', type=str, default="What is AI?",
+                        help='Text to extract features')
+
+    args = parser.parse_args()
+    model_path = args.repo_id_or_model_path
+
+    # Load model
+    model = BertModel.from_pretrained(model_path)
```
shane-huang, line 37: shall we also use the options `torch_dtype=auto` and `load_low_cpu_memory_usage=True`?

python/llm/example/pytorch-models/bert/README.md
```diff
@@ -0,0 +1,48 @@
...
+pip install --pre --upgrade bigdl-llm[all] # install the latest bigdl-llm nightly build with 'all' option
+```
+
+### 2. Run
+After setting up the Python environment, you could run the example by following steps.
+
+#### 2.1 Client
+On client Windows machines, it is recommended to run directly with full utilization of all cores:
+```powershell
+python ./extract_feature.py --text 'What is AI?'
```
shane-huang, line 26: maybe replace `what is AI` with `This is an example text for feature extraction.`

python/llm/example/pytorch-models/bert/extract_feature.py
plusbang: > shall we also use the options `torch_dtype=auto` and `load_low_cpu_memory_usage=True`?

yeah, it would be better if we set it. Have updated.

shane-huang (approved):
LGTM.

Issue #8942, opened 2023-09-11
Whisper got error "expected scalar type Byte but found Half"  using load_low_bit 


violet17:
Whisper got error "expected scalar type Byte but found Half"  using `load_low_bit`  version `2.4.0b20230910` on A770.
no error using `from_pretrained`

runing log:
```
$ python test_whisper.py  --whisper-repo-id-or-model-path ./whisper-medium-int4-0910/
/home/a/miniconda3/envs/llm/lib/python3.9/site-packages/torchvision/io/image.py:13: UserWarning: Failed to load image Python extension: ''If you don't plan on using image functionality from `torchvision.io`, you can ignore this warning. Otherwise, there might be something wrong with your environment. Did you have `libjpeg` or `libpng` installed before building `torchvision` from source?
  warn(
Converting and loading models...
loading whisper----1
loading whisper----2
loading whisper----Done
audio_file:  /home/a/crystal/llm/pyaudio_out.wav
audio:  <speech_recognition.audio.AudioData object at 0x7f8ec0bf1bb0> frame_data:  [0.         0.         0.         ... 0.00015259 0.         0.00054932]
/home/a/miniconda3/envs/llm/lib/python3.9/site-packages/transformers/generation/utils.py:1369: UserWarning: Using `max_length`'s default (448) to control the generation length. This behaviour is deprecated and will be removed from the config in v5 of Transformers -- we recommend using `max_new_tokens` to control the maximum length of the generation.
  warnings.warn(
Traceback (most recent call last):
  File "/home/a/crystal/llm/test_whisper.py", line 168, in <module>
    predicted_ids = whisper.generate(input_features)
  File "/home/a/miniconda3/envs/llm/lib/python3.9/site-packages/transformers/models/whisper/modeling_whisper.py", line 1749, in generate
    outputs = super().generate(
  File "/home/a/miniconda3/envs/llm/lib/python3.9/site-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "/home/a/miniconda3/envs/llm/lib/python3.9/site-packages/transformers/generation/utils.py", line 1538, in generate
    return self.greedy_search(
  File "/home/a/miniconda3/envs/llm/lib/python3.9/site-packages/transformers/generation/utils.py", line 2362, in greedy_search
    outputs = self(
[...]

PR #8977, opened 2023-09-14
LLM: Save transformer models with `tie_word_embeddings=False` option

leonardozcm:
## Description

For symmetrical embedding architecture, they will duplicate the input embedding matrix to output embedding, on the purpose of saving only one shard of embedding weights on disk. Since we have already resize the shape of output embedding, this 'tie_word_embeddings' operation is ineffective, and because the two embeddings' weights have already been aligned during 'from_pretrained', there is no need to perform this operation during the second load.

llama2 is not involved on this PR.:
```
### HUMAN:
What is AI?

### RESPONSE:

### HUMAN:
What is AI?

### RESPONSE:

AI, or artificial intelligence, refers to the ability of machines to perform tasks that would typically require human intelligence, such as learning, problem-solving,
```

related issue:
https://github.com/intel-analytics/BigDL/issues/8942
https://github.com/analytics-zoo/nano/issues/581

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/model.py
+++ b/python/llm/src/bigdl/llm/transformers/model.py
@@ -121,6 +121,7 @@ class _BaseAutoModelClass:
         model = model.to("cpu")
         model = ggml_convert_low_bit(model, qtype, optimize_model)
         model.config.update({"bigdl_transformers_low_bit": q_k})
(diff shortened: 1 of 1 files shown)


cyita (approved):
LGTM

PR #8980, opened 2023-09-14
[bigdl-llm] Remove serving-dep from all_requires

gc-fu:
## Description

Pin fastchat version & remove serving_dependencies from all_requires


xiangyuT (approved):
LGTM

PR #8981, opened 2023-09-14
LLM: add auto torch dtype in bf16 benchmark.

lalalapotter:
## Description

This PR is to add `torch_dtype="auto"` in bf16 benchmark when loading mode, to avoid using fp32 to load for all models.


leonardozcm (approved):
LGTM

PR #8999, opened 2023-09-18
LLM: Add one-click installer for Windows

JinBridger:
## Description

Add one-click installer for Windows.

- Tested on ChatGLM2-6b, Baichuan-13B-Chat, Baichuan2-7B-Chat
- Tested on another Windows machine without Python.

For usage, please refer to README.md.


python/llm/portable-installer/chat.py
```diff
@@ -0,0 +1,111 @@
...
+def stream_chat(model,
+                tokenizer,
+                stopping_criteria,
+                input_str,
+                chat_history):
+    prompt = format_prompt(input_str, chat_history)
+    # print(prompt)
+    input_ids = tokenizer([prompt], return_tensors="pt")
+    streamer = TextIteratorStreamer(tokenizer, skip_prompt=True, skip_special_tokens=True)
+    generate_kwargs = dict(input_ids, streamer=streamer, max_new_tokens=512, stopping_criteria=stopping_criteria)
```
Oscilloscope98, line 54: `use_cache=True` should be added into generation config, not in the `from_pretrained` function. Refer to [here](https://huggingface.co/docs/transformers/main_classes/text_generation#transformers.GenerationConfig.use_cache)

python/llm/portable-installer/chat.py
```diff
@@ -0,0 +1,111 @@
...
+                                                    low_cpu_mem_usage=True,
+                                                    torch_dtype="auto",
+                                                    trust_remote_code=True,
+                                                    use_cache=True)
+    else:
+        return AutoModel.from_pretrained(model_path,
+                                         low_cpu_mem_usage=True,
+                                         torch_dtype="auto",
+                                         trust_remote_code=True,
+                                         use_cache=True)
```
Oscilloscope98, line 84: Maybe we should add a try except logic here to catch the case for the model which cannot be loaded by `AutoModelForCausalLM` and `AutoModel`, and print a related message.

Later we could find a better way to support more model class

python/llm/portable-installer/setup.bat
```diff
@@ -0,0 +1,20 @@
...
+
+:: enable run site.main() automatically
+cd .\python-embed
+set "search=#import site"
+set "replace=import site"
+powershell -Command "(gc python39._pth) -replace '%search%', '%replace%' | Out-File -encoding ASCII python39._pth"
+cd ..
+
+:: install pip packages
+%python-embed% -m pip install bigdl-llm[all] transformers_stream_generator
```
Oscilloscope98, line 20: Maybe also add dependency: `tiktoken` and `einops`.

`einops` is required for mpt, falcon and qwen;
`tiktoken` is required for qwen

python/llm/portable-installer/README.md
```diff
@@ -0,0 +1,12 @@
+# BigDL-LLM One-Click Installer For Windows
+
+## Supported models
+
+- ChatGLM2-6b
+- Baichuan-13B-Chat
+- Baichuan2-7B-Chat
```
Oscilloscope98, line 7: We could change the title to `Verified Models`, as I believe this one-click tool can work for much more models.

python/llm/portable-installer/setup.bat
shane-huang: I think we'd better separate the setup scripts/README and user scripts&README. We use setup.bat to build the zip, and the zip should only contain the user scripts and user guide.

shane-huang: In user guide please add a step-by-step usage guide with screen snapshots.

shane-huang: Also, you can upload a gif of the process in github to show the entire process.

python/llm/portable-installer/chat.py
JinBridger: Added.

python/llm/portable-installer/README.md
JinBridger: Done.

Oscilloscope98:
It would be better to change the question in the demo gif :)

For example, change "谁创造了你" to "AI是什么“ etc. I feel a little strange to see the name of another company in the demo video.

JinBridger:
> It would be better to change the question in the demo gif :)
> 
> For example, change "谁创造了你" to "AI是什么“ etc. I feel a little strange to see the name of another company in the demo video.

Yes, I recorded another gif and now there's no company name in the gif.

python/llm/portable-installer/setup.bat
JinBridger: Already done as you said  :)

jason-dai:
Do not upload pictures to this repo; see gif examples at https://github.com/intel-analytics/BigDL/blob/main/README.md

python/llm/portable-installer/README.md
```diff
@@ -0,0 +1,33 @@
+# BigDL-LLM One-Click Installer For Windows: User Guide
```
shane-huang, line 1: Since we actually didn't require user to install anything, shall we call it "BigDL-LLM Portable"?

JinBridger: Already done as you said. :)

python/llm/portable-executable/setup.bat
```diff
@@ -0,0 +1,23 @@
...
+
+:: enable run site.main() automatically
+cd .\python-embed
+set "search=#import site"
+set "replace=import site"
+powershell -Command "(gc python39._pth) -replace '%search%', '%replace%' | Out-File -encoding ASCII python39._pth"
+cd ..
+
+:: install pip packages
+%python-embed% -m pip install bigdl-llm[all] transformers_stream_generator tiktoken einops
```
Oscilloscope98, line 20: Do we also need `colorama` here?

JinBridger: It is installed as a dependency when install other packages. Anyway, it's better to add it. I've added it in the latest commit.

shane-huang (approved):
LGTM. We'll add GPU version in another PR.

jason-dai:
Is there an installation/setup guide? @shane-huang @Oscilloscope98

Oscilloscope98:
> Is there an installation/setup guide? @shane-huang @Oscilloscope98

We currently have a very simple setup-guide for how to build the zip file: https://github.com/intel-analytics/BigDL/blob/main/python/llm/portable-executable/setup.md

PR #9000, opened 2023-09-19
Speed up gpt-j in gpubenchmark

qiuxin2012:
## Description

Speed up gpt-j in gpubenchmark

### 1. Why the change?

gpt-j need additional optimization code.

### 2. User API changes

none


cyita (approved):
LGTM

python/llm/dev/benchmark/all-in-one/run.py
```diff
@@ -294,6 +294,9 @@ def run_transformer_int4_gpu(repo_id,
     print(">> loading of model costs {}s".format(end - st))
 
     model = model.to('xpu')
+    if isinstance(model, GPTJForCausalLM):
```
jason-dai, line 297: Move this to the if-else above:

```python
 if repo_id in ...
        model = ...
 else:
        model = ...
```

PR #9001, opened 2023-09-19
[LLM] Fix GPU performance tests

Oscilloscope98:
Fix LLM performance tests for GPU.
Currently, `bigdl-core-xe` cannot be successfully installed with default version `2.4.0.dev0`. We find a way to make sure latest `bigdl-core-xe`  is always installed for GPU related actions.

https://github.com/intel-analytics/BigDL/actions/runs/6222941880/job/16888205721


rnwang04 (approved):
LGTM

PR #9003, opened 2023-09-19
Add xpu image for bigdl-llm

gc-fu:
## Description

Add an image that supports running `BigDL-LLM[xpu]` in container

### How to test?
- [x] Manually test

In image, running `sycl-ls`:

image

Running inference:
image

### DIFF
--- a/docker/llm/inference/xpu/docker/Dockerfile
+++ b/docker/llm/inference/xpu/docker/Dockerfile
@@ -1,22 +1,18 @@
-FROM intel/oneapi-basekit
+FROM intel/oneapi-basekit:latest
 
 ENV http_proxy $HTTP_PROXY
 ENV https_proxy $HTTP_PROXY
 
 ENV TZ=Asia/Shanghai
 
-# Otherwise will require user to input timezone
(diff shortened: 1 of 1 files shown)


docker/llm/inference/xpu/docker/Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM intel/oneapi-basekit
```
glorysdj, line 1: need to specify the version of  intel/oneapi-basekit

docker/llm/inference/xpu/docker/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+ENV http_proxy $HTTP_PROXY
+ENV https_proxy $HTTP_PROXY
+
+ENV TZ=Asia/Shanghai
+
+# Otherwise will require user to input timezone
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
```
glorysdj, line 9: please use one RUN to install all of the dependencies

PR #9018, opened 2023-09-20
add export merged model example

yangw1234:
## Description

add export merged model example


python/llm/example/gpu/qlora_finetuning/qlora_finetuning.py
```diff
@@ -71,7 +72,8 @@
             warmup_steps=20,
             max_steps=200,
             learning_rate=2e-4,
-            fp16=False, # fp16 is not supported yet
+            save_steps=100,
+            fp16=True, # fp16 is not supported yet
```
jason-dai, line 76: is supported?

PR #9020, opened 2023-09-21
MPT model optimize for long sequence

jenniew:
Optimize KV cache in MPT model to reduce long sequence latency.

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/models/mpt.py
+++ b/python/llm/src/bigdl/llm/transformers/models/mpt.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 # Some parts of this file is adapted from
-# https://huggingface.co/mosaicml/mpt-7b/blob/main/attention.py
+# https://huggingface.co/mosaicml/mpt-7b-chat/blob/main/attention.py
 #
 
 import warnings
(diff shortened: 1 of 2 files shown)


python/llm/src/bigdl/llm/transformers/models/mpt.py
```diff
@@ -0,0 +1,157 @@
...
+    kv_seq_len = k.shape[-1]
+    if past_key_value is not None:
+        if len(past_key_value) != 0:
+        #     k = torch.cat([past_key_value[0], k], dim=3)
+        #     v = torch.cat([past_key_value[1], v], dim=2)
+            cache_k = past_key_value[0].permute(0, 1, 3, 2)
+            cache_v = past_key_value[1]
+            kv_seq_len += cache_k.shape[-2]
+            if cache_k.stride()[1] <= cache_k.size(2) * cache_k.size(3):
+                if device.type == 'xpu':
```
sgwhat, line 76: we can remove this, since it has been merge into `create_kv_cache`.

python/llm/src/bigdl/llm/transformers/models/mpt.py
```diff
@@ -0,0 +1,139 @@
...
+                                                           dtype=cache_k.dtype,
+                                                           device=device)
+                new_cache_k[:] = cache_k
+                new_cache_v[:] = cache_v
+                cache_k = new_cache_k
+                cache_v = new_cache_v
+            key_states, value_states = append_kv_cache(cache_k, cache_v, k.permute(0, 1, 3, 2), v)
+            k = key_states.permute(0, 1, 3, 2)
+            v = value_states
+        else:
```
sgwhat, line 93: maybe `elif use_cache:`?

jenniew: There is no use_cache flag in this model. In this model, just check whether len(past_key_value) != 0 to see if it is the first time since initially past_key_value=()

sgwhat (approved):
LGTM

PR #9024, opened 2023-09-21
LLM: Fix empty cache

rnwang04:
## Description

Today we find after moving `torch.xpu.empty_cache()` into `create_kv_cache`, will incur several problems like strange output of llama and slowndown of chatglm.
Move this functino to model file again.

### 1. Why the change?

https://github.com/analytics-zoo/nano/issues/594

### 2. User API changes

No change.

### 4. How to test?
- [x] Unit test
- [x] Local test

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/models/baichuan.py
+++ b/python/llm/src/bigdl/llm/transformers/models/baichuan.py
@@ -70,6 +70,8 @@ def baichuan_attention_forward_7b(
         cache_k = past_key_value[0]
         cache_v = past_key_value[1]
         if cache_k.stride()[1] <= cache_k.size(2) * cache_k.size(3):
+            if device.type == 'xpu':
+                torch.xpu.empty_cache()
             # allocate new
             new_cache_k, new_cache_v = create_kv_cache(bsz,
                                                        self.num_heads,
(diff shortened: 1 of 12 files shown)


python/llm/src/bigdl/llm/transformers/models/baichuan.py
```diff
@@ -70,6 +70,8 @@ def baichuan_attention_forward_7b(
         cache_k = past_key_value[0]
         cache_v = past_key_value[1]
         if cache_k.stride()[1] <= cache_k.size(2) * cache_k.size(3):
+            if device.type == 'xpu':
+                torch.xpu.empty_cache()
```
jason-dai, line 74: better to have two versions: `init_kv_cache` and `extend_kv_cache`

qiuxin2012:
chatglm2 is OK with this PR.

python/llm/src/bigdl/llm/transformers/models/baichuan.py
rnwang04: I will try such refactor, we can merge this pr first to better user experience.

MeouSker77 (approved):
LGTM

PR #9026, opened 2023-09-21
add README.md for bigdl-llm-cpu image

Zhengjin-Wang:
add README.md and also modify Dockerfile

### DIFF
--- a/docker/llm/inference/cpu/docker/README.md
+++ b/docker/llm/inference/cpu/docker/README.md
@@ -1,0 +1,34 @@
+## Build/Use BigDL-LLM cpu image
+
+### Build Image
+```bash
+docker build \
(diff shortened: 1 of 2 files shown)


docker/llm/inference/cpu/docker/README.md
```diff
@@ -0,0 +1,34 @@
...
+        --memory="32G" \
+        --name=CONTAINER_NAME \
+        --shm-size="16g" \
+        $DOCKER_IMAGE
+```
+
+
+After the container is booted, you could get into the container through `docker exec`.
+
+
```
glorysdj, line 34: need an example of how to do the inference, may be a link to existing doc

glorysdj (approved):
LGTM

PR #9035, opened 2023-09-22
LLM: Change benchmark bf16 load format.

lalalapotter:
## Description

This PR is to change benchmark bf16 load format to `torch.bfloat16`. The reasons are as follows:
1. If `torch_dtype` is not set, the model will load in fp32 format. Therefore, Large model, such as Bloom-176b and falcon-180b need to use bfloat16 format to load model for saving memory usage.
2. If `torch_dtype` is set to `"auto"`, the model will load in the configured format in model file. According to our test results, the `autocast` function will not convert all the model weights to bfloat16 and therefore will have a latency performance drop compare with setting `torch_dtype` to `torch.bfloat16`.

### DIFF
--- a/python/llm/dev/benchmark/all-in-one/run.py
+++ b/python/llm/dev/benchmark/all-in-one/run.py
@@ -176,7 +176,7 @@ def run_pytorch_autocast_bf16(repo_id,
     st = time.perf_counter()
     if repo_id in ['THUDM/chatglm-6b', 'THUDM/chatglm2-6b']:
         # TODO: need verify chatglm family run bf16.
-        invalidInputError(False, "Currently pytorch do not support bfloat 16 on cpu for chatglm models.")
+        invalidInputError(False, "Currently pytorch do not support bfloat16 on cpu for chatglm models.")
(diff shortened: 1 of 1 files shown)


python/llm/dev/benchmark/all-in-one/run.py
```diff
@@ -176,18 +176,16 @@ def run_pytorch_autocast_bf16(repo_id,
     st = time.perf_counter()
     if repo_id in ['THUDM/chatglm-6b', 'THUDM/chatglm2-6b']:
         # TODO: need verify chatglm family run bf16.
-        model = AutoModel.from_pretrained(model_path, trust_remote_code=True, torch_dtype='auto').float()
-        #model = AutoModel.from_pretrained(model_path, trust_remote_code=True, torch_dtype='auto').bfloat()
-        tokenizer = AutoTokenizer.from_pretrained(model_path, trust_remote_code=True)
+        invalidInputError(False, "Currently pytorch do not support bfloat 16 on cpu for chatglm models.")
```
hkvision, line 179: bfloat16, no space

hkvision (approved):

PR #9038, opened 2023-09-25
change lora output behavior on k8s

Uxito-Ada:
## Description

As I found and user reported, finetuning output paths sometimes relate to privilege issue on version-specific k8s platform, thus I change the log and model output path to be under the /home/mpiuser to ensure a successful writing on both launcher and worker pods.

### 1. Why the change?

as above

### 2. User API changes

no need to set `outputPath` in helm values yaml file, and get finetuned model from pod 0.

### 3. Summary of the change 

as above

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [x] Application test
- [ ] Document test
- [ ] ...


xiangyuT (approved):
LGTM

PR #9050, opened 2023-09-25
separate trusted and native llm cpu finetune from lora

Uxito-Ada:
## Description

a refactor is needed to separate bigdl lora finetuning(with TDX) into two, trusted-llm and native finetuning without TDX respectively. And we aim to provide both CPU and XPU versions and now start with CPU one.

### 1. Why the change?

as above

### 2. User API changes

no

### 3. Summary of the change 

separate trusted-llm and bigdl from lora finetuning

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [x] Application test
- [ ] Document test
- [ ] ...


Uxito-Ada:
TODO: append trusted-llm k8s readme and codes

docker/llm/finetune/lora/cpu/docker/Dockerfile
```diff
@@ -0,0 +1,59 @@
+ARG HTTP_PROXY
+ARG HTTPS_PROXY
+
```
glorysdj, line 3: use  PIP_NO_CACHE_DIR=false to turn off pip cache and reduce the size of the image
https://github.com/intel-analytics/BigDL/blob/main/docker/llm/inference/cpu/docker/Dockerfile#L5

docker/llm/finetune/lora/cpu/docker/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+FROM mpioperator/intel as builder
+
+ARG HTTP_PROXY
+ARG HTTPS_PROXY
+ADD ./requirements.txt /ppml/requirements.txt
+
+RUN mkdir /ppml/data && mkdir /ppml/model && mkdir /ppml/output && \
+# install pytorch 2.0.1
+    export http_proxy=$HTTP_PROXY && \
```
glorysdj, line 12: use arg http_proxy/https_proxy and let user to choose if proxy is needed to set

xiangyuT (approved):
LGTM

PR #9055, opened 2023-09-26
Replace Llama 7b to Llama2-7b in README.md

Jasonzzt:
Need to replace the base model to Llama2-7b as we are operating on Llama2 here.


docker/llm/finetune/lora/README.md
```diff
@@ -2,15 +2,15 @@
```
Uxito-Ada, line 2: a llama 7b in the 1st line is missed, and others LGTM

Uxito-Ada (approved):
lgtm

PR #9062, opened 2023-09-26
Add bigdl llm xpu image build

Zhengjin-Wang:
- Modify Dockerfile for llm inference xpu
- Add llm inference xpu build command into workflow


glorysdj (approved):
LGTM

PR #9066, opened 2023-09-27
using bigdl-llm fused rope for llama

yangw1234:
## Description

using bigdl-llm fused rope for llama to reduce generation latency

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/models/llama.py
+++ b/python/llm/src/bigdl/llm/transformers/models/llama.py
@@ -39,6 +39,7 @@ import torch.nn.functional as F
 from bigdl.llm.utils.common import invalidInputError
 from bigdl.llm.transformers.models.utils import init_kv_cache, extend_kv_cache, append_kv_cache
 from bigdl.llm.transformers.models.utils import rotate_half, apply_rotary_pos_emb
+from bigdl.llm.transformers.models.utils import apply_rotary_pos_emb_no_cache_xpu
 
 
 def repeat_kv(hidden_states: torch.Tensor, n_rep: int) -> torch.Tensor:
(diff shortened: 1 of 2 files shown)


python/llm/src/bigdl/llm/transformers/models/llama.py
```diff
@@ -103,9 +113,15 @@ def llama_attention_forward_4_31(
...
-    cos, sin = self.rotary_emb(value_states, seq_len=kv_seq_len)
-    query_states, key_states = apply_rotary_pos_emb(query_states, key_states,
-                                                    cos, sin, position_ids, "llama")
+    # cos, sin = self.rotary_emb(value_states, seq_len=kv_seq_len)
+    # query_states, key_states = apply_rotary_pos_emb(query_states, key_states,
+    #                                                 cos, sin, position_ids, "llama")
+    q_embed = torch.empty(query_states.shape, dtype=query_states.dtype, device=query_states.device)
+    k_embed = torch.empty(key_states.shape, dtype=key_states.dtype, device=key_states.device)
+    
+    linear_q4_0.apply_rotary_embedding_half_qk(query_states, key_states, position_ids, q_embed, k_embed)
```
jason-dai, line 122: this only works on GPU, yes?

yangw1234: I think cpu ipex has a similar function.

jason-dai: > I think cpu ipex has a similar function.

We can add our GPU optimizations first

yangw1234:
I think I am going to separate rms_norm, rope and other changes into different PRs with rms_norm being the first.

yangw1234:
performance updated here https://github.com/analytics-zoo/nano/issues/606

@jason-dai would you mind reviewing again?

yangw1234:
These optimizations does not work for training.

Should we check if it is in the training mode in every place? 
Or should we can assume user should not set `optimize_model=True` in training? 

@jason-dai

yangw1234:
> These optimizations does not work for training.
> 
> Should we check if it is in the training mode in every place? Or should we can assume user should not set `optimize_model=True` in training?
> 
> @jason-dai

added checking for now

jason-dai (approved):
LGTM

PR #9072, opened 2023-09-28
LLM: support chatglm-18b convert attention forward in benchmark scripts.

lalalapotter:
## Description

This PR is to support chatglm-18b convert attention forward in benchmark scripts. The reason is that the attention implementation of chatglm-18b is similar with chatglm2-6b(not chatglm-6b), thus we should use chatglm2-6b optimization for chatglm-18b.

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/convert.py
+++ b/python/llm/src/bigdl/llm/transformers/convert.py
@@ -153,29 +153,28 @@ def optimize(model):
         # todo implement 4.28.0 ~ 4.30.2
         pass
 
-    if "chatglm" in model.config._name_or_path:
-        if model.config.hidden_size == 6144 or "chatglm2" in model.config._name_or_path:
-            # chatglm-18b or chatglm2-6b
-            modeling_module_name = model.__class__.__module__
-            module = importlib.import_module(modeling_module_name)
-            from bigdl.llm.transformers.models.chatglm2 import chatglm2_attention_forward_8eb45c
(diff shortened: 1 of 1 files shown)


python/llm/src/bigdl/llm/transformers/convert.py
```diff
@@ -153,26 +153,29 @@ def optimize(model):
...
-    elif "chatglm" in model.config._name_or_path:
-        modeling_module_name = model.__class__.__module__
-        module = importlib.import_module(modeling_module_name)
-        from bigdl.llm.transformers.models.chatglm import chatglm_attention_forward
-        convert_forward(model,
-                        module.SelfAttention,
-                        chatglm_attention_forward
-                        )
+    if "chatglm" in model.config._name_or_path:
+        if model.config.hidden_size == 6144 or "chatglm2" in model.config._name_or_path:
```
hkvision, line 157: Is it promising to use hidden_size for the check?
Why not directly check if "chatglm-18b" in model name?

python/llm/src/bigdl/llm/transformers/convert.py
```diff
@@ -166,6 +167,7 @@ def optimize(model):
                         module.CoreAttention,
                         core_attn_forward_8eb45c)
     elif "chatglm" in model.config._name_or_path:
+        # chatglm
```
hkvision, line 170: chatglm-6b?

hkvision (approved):

PR #9079, opened 2023-10-04
Qwen kv cache

jenniew:
Add kv cache to Qwen family.

related issue:https://github.com/analytics-zoo/nano/issues/572


python/llm/src/bigdl/llm/transformers/models/qwen.py
```diff
@@ -0,0 +1,219 @@
...
+        hidden_states.device
+    )
+
+    if rotary_pos_emb is not None:
+        if isinstance(rotary_pos_emb, tuple):
+            rotary_pos_emb = rotary_pos_emb
+        else:
+            rotary_pos_emb = (rotary_pos_emb,) * 2
+
+    if rotary_pos_emb is not None:
```
yangw1234, line 123: should we try to reuse some of the code in https://github.com/intel-analytics/BigDL/blob/main/python/llm/src/bigdl/llm/transformers/models/utils.py#L75-L81 ?

yangw1234: The logic seems to be similar, except there are no explicit `position_ids` here but use the last `cur_len`

Issue #9023, opened 2023-09-21
How to run bark (TTS API) on dGPU？


violet17:
Hi, I run  [synthesize_speech.py ](https://github.com/intel-analytics/BigDL/blob/main/python/llm/example/pytorch-models/bark/synthesize_speech.py) with bigdl-llm 2.4.0b20230920, and find that it can not use GPU.
And how to run this demo on Arc dGPU?

Thanks.

shane-huang:
First you need to install xpu related dependencies, follow the llama2 example install section for how to https://github.com/intel-analytics/BigDL/blob/main/python/llm/example/gpu/hf-transformers-models/llama2/README.md. 

In the original code, you can try add `model.to('xpu')` to load the model to ARC, and add `output = output.cpu()` move the generated results back to CPU, like below:

```python
    # Load model
    model.load_checkpoint(config, checkpoint_dir=model_path, eval=True)
    ## move to ARC
    model.to('xpu')
    # With only one line to enable BigDL-LLM optimization on model
    model = optimize_model(model)
    # generate results
    output_dict = model.synthesize(text, config, speaker_id="random", voice_dirs=None)
    # remove results back to CPU
     output_dict = output_dict .cpu()
```

violet17:
Hi @shane-huang  Thanks for you reply.

I put the model to arc, then get error with bigdl-llm 2.4.0b20230920 and TTS 0.17.4 :
```
====== torch.float16 torch.float32
Traceback (most recent call last):
  File "/home/a770/crystal/llm/bark_bigdl2.py", line 36, in <module>
    output_dict = model.synthesize(text, config, speaker_id="random", voice_dirs=None) # with random speaker
  File "/home/a770/miniconda3/envs/llm2/lib/python3.9/site-packages/TTS/tts/models/bark.py", line 220, in synthesize
    outputs = self.generate_audio(text, history_prompt=history_prompt, **kwargs)
  File "/home/a770/miniconda3/envs/llm2/lib/python3.9/site-packages/TTS/tts/models/bark.py", line 153, in generate_audio
    x_semantic = self.text_to_semantic(
  File "/home/a770/miniconda3/envs/llm2/lib/python3.9/site-packages/TTS/tts/models/bark.py", line 87, in text_to_semantic
    x_semantic = generate_text_semantic(
[...]

PR #9092, opened 2023-10-07
LLM: Check model device type

leonardozcm:
## Description

https://github.com/intel-analytics/BigDL/issues/9023#issuecomment-1730938421


python/llm/src/bigdl/llm/optimize.py
```diff
@@ -138,6 +138,16 @@ def load_low_bit(model, model_path):
                       "model should be a instance of "
                       f"`torch.nn.Module`, but got {type(model)} at last.")
     if low_bit:
+        # a creator
+        if is_creator:
```
plusbang, line 142: `is_creator` seems like a flag but where get it? and the following `model_or_creator` also is confusing...

leonardozcm: Yes this is a conflict, and has been fixed.

plusbang (approved):

PR #9094, opened 2023-10-07
LLM: add original model benchmark on cpu and gpu

JinBridger:
## Description

Add benchmark function for model WITHOUT bigdl optimizations. Both on CPU and GPU.

### DIFF
--- a/python/llm/dev/benchmark/all-in-one/config.yaml
+++ b/python/llm/dev/benchmark/all-in-one/config.yaml
@@ -9,11 +9,11 @@ in_out_pairs:
   - '32-32'
   - '1024-128'
 test_api:
-  - "transformer_int4"
-  - "native_int4"
-  - "optimize_model"
-  - "pytorch_autocast_bf16"
-  # - "transformer_int4_gpu"  # on arc
-  # - "optimize_model_gpu"  # on arc
-  # - "original_model_cpu"  # on cpu without optimization
-  # - "original_model_gpu"  # on gpu without optimization
+  - "transformer_int4"      # on CPU with load_in_4bit=True
+  - "transformer_int4_gpu"  # on GPU with load_in_4bit=True
+  - "native_int4"           # on CPU with BigdlNativeForCausalLM
+  - "pytorch_autocast_bf16" # on GPU with torch.bfloat16
+  - "optimize_model"        # on CPU with optimize_model
+  - "optimize_model_gpu"    # on GPU with optimize_model
+  - "original_model_cpu"    # on CPU without any bigdl-llm optimization
+  - "original_model_gpu"    # on GPU without any bigdl-llm optimization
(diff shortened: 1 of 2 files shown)


python/llm/dev/benchmark/all-in-one/run.py
```diff
@@ -258,7 +263,63 @@ def run_optimize_model(repo_id,
...
+    from transformers import AutoTokenizer, LlamaTokenizer, AutoModel, AutoModelForCausalLM
+
+    model_path = get_model_path(repo_id, local_model_hub)
+    st = time.perf_counter()
+    if repo_id in ['THUDM/chatglm-6b', 'THUDM/chatglm2-6b']:
+        model = AutoModel.from_pretrained(model_path, trust_remote_code=True, torch_dtype='auto', low_cpu_mem_usage=True).float()
+        tokenizer = AutoTokenizer.from_pretrained(model_path, trust_remote_code=True)
+    elif repo_id in ['meta-llama/Llama-2-70b-chat-hf']:
+        model = AutoModelForCausalLM.from_pretrained(model_path,
+                                                     trust_remote_code=True, optimize_model=False, low_cpu_mem_usage=True)
```
plusbang, line 290: `optimize_model` parameter of `from_pretrained` is for our `bigdl.llm.transformers.AutoModelForCausalLM`.

python/llm/dev/benchmark/all-in-one/run.py
```diff
@@ -387,6 +448,62 @@ def run_optimize_model_gpu(repo_id,
     torch.xpu.empty_cache()
     return result
 
+def run_original_model_gpu(repo_id,
+                           local_model_hub,
+                           in_out_pairs,
+                           warm_up,
+                           num_trials):
+    from transformers import AutoModel, AutoModelForCausalLM, AutoTokenizer, GPTJForCausalLM
+    import intel_extension_for_pytorch as ipex
```
plusbang, line 457: do we need to import ipex when collect baseline performance data of without using `optimize_model`?

python/llm/dev/benchmark/all-in-one/run.py
JinBridger: Got it, I'll remove line 288-291.

python/llm/dev/benchmark/all-in-one/run.py
JinBridger: Looks like it is unused in code below. I'll remove it.

JinBridger: Looks like removing it will cause `to('xpu')` stop working. I've added it now.

python/llm/dev/benchmark/all-in-one/config.yaml
```diff
@@ -14,4 +14,6 @@ test_api:
   - "optimize_model"
   - "pytorch_autocast_bf16"
   # - "transformer_int4_gpu"  # on arc
-  # - "optimize_model_gpu"  # on arc
\ No newline at end of file
+  # - "optimize_model_gpu"  # on arc
+  # - "original_model_cpu"
+  # - "original_model_gpu"
```
jason-dai, line 19: add comments to describe different options

JinBridger: Got it! I've added comments in config.yaml.

PR #9081, opened 2023-10-04
Use ipex fused rms norm for llama

yangw1234:
## Description

Use ipex fused rms norm for llama for xpu

### 4. How to test?

manually tested it end to end 
and the correctness of rms norm kernel is covered in https://github.com/intel/intel-extension-for-pytorch/blob/v2.0.110%2Bxpu/tests/gpu/examples/test_rms_norm.py#L45

The greedy output becomes different after a few tokens but I think it is due to numerical issues and is ok.

#### Main branch output XPU

> Once upon a time, there existed a little girl who liked to have adventures. She wanted to go to places and meet new people, and have fun. But her parents were always telling her to stay close to home, to be careful, and to always come back when the sun went down.
> 
> One

#### Main branch output CPU

> Once upon a time, there existed a little girl who liked to have adventures. She wanted to go to places and meet new people, and have fun. But her parents were always telling her to stay close to home, to be careful, and to not go anywhere alone.
> 
> One day, the little

#### This PR output XPU

> Once upon a time, there existed a little girl who liked to have adventures. She wanted to go to places and meet new people, and have fun. But her parents were always telling her to stay close to home, to be careful, and to always follow the rules.
> One day, the little girl

#### This PR output CPU

> Once upon a time, there existed a little girl who liked to have adventures. She wanted to go to places and meet new people, and have fun. But her parents were always telling her to stay close to home, to be careful, and to not go anywhere alone.
> 
> One day, the little

PR #9103, opened 2023-10-08
Align rope of chatglm with llama

hkvision:
This PR tries to use the same logic of llama rope for chatglm, which is the first step for use fuse rope xpu.

Since rotary_pos_emb is already indexed by position_id (i.e. cos/sin) before passing to self-attention forward, more work needs to be done followed by this PR to align with llama (where cos/sin are calculated in the kernel and then indexed by position_id instead).

Any comments on this? Not sure after such modification there would still be performance benefits. @jason-dai @yangw1234

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/models/chatglm2.py
+++ b/python/llm/src/bigdl/llm/transformers/models/chatglm2.py
@@ -20,7 +20,7 @@
 import torch
 from typing import Optional, Tuple, Union, List, Callable, Dict, Any
 import torch.nn.functional as F
-from bigdl.llm.transformers.models.utils import init_kv_cache, extend_kv_cache, append_kv_cache
+from bigdl.llm.transformers.models.utils import init_kv_cache, extend_kv_cache, append_kv_cache, apply_rotary_pos_emb
 
 
 KV_CACHE_ALLOC_BLOCK_LENGTH = 256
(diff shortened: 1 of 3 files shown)


hkvision:
Updated, the current implementation uses the existing xpu fuse rope kernel to the largest extent.

python/llm/src/bigdl/llm/transformers/models/chatglm2.py
```diff
@@ -119,19 +119,56 @@ def chatglm2_attention_forward_8eb45c(
...
+            key_layer = apply_rotary_pos_emb_chatglm(key_layer, rotary_pos_emb)
 
     if self.multi_query_attention:
         key_length = key_layer.size(0)
         query_group_size = self.num_attention_heads_per_partition // \
             self.num_multi_query_groups_per_partition
+        # if rotary_pos_emb is not None and use_fuse_rope:
+        #     key_layer = key_layer.unsqueeze(-3)
+        # else:
+        #     key_layer = key_layer.permute(1, 2, 0, 3).unsqueeze(-3)  # [bs, nh/k, sl, hn]
```
hkvision, line 168: In theory this permute can be merged with xpu kernel's output shape, but don't know why it gets the error:
```
  File "/home/arda/anaconda3/envs/kai-llm-pip/lib/python3.9/site-packages/bigdl/llm/transformers/models/chatglm2.py", line 193, in chatglm2_attention_forward_8eb45c
    key_layer = key_layer.contiguous().view((batch_size,
RuntimeError: shape '[1, 32, 1, 128]' is invalid for input of size 53248
```

hkvision:
Result update and analysis is here: https://github.com/analytics-zoo/nano/issues/608#issuecomment-1755356079

hkvision:
TODO: may need to further compare the result after rotary embedding.
Seems the generated texts are different before/after the change, but both are meaningful and in good quality.

python/llm/src/bigdl/llm/transformers/models/chatglm2.py
```diff
@@ -187,7 +275,8 @@ def chatglm2_attention_forward_8eb45c(
     # ==================================
     # core attention computation
     # ==================================
-
+    # print("----query", query_layer.shape)
```
hkvision, line 278: clear these

yangw1234:
rms_norm seems to change the result a bit for llama. See https://github.com/intel-analytics/BigDL/pull/9081

hkvision:
> rms_norm seems to change the result a bit for llama. See #9081
> 
> I guess it should be an alarm if the first a few output tokens are different.

Sure, but seems chatglm2 changes output after modifying rope now, will further check this.

hkvision:
Pending for fuse, @sgwhat will break fuse rms into a separate PR.

hkvision:
https://github.com/intel-analytics/BigDL/pull/9350 follow-up PR, closing this one.

PR #9104, opened 2023-10-09
LLM: Meta Llama2 example

leonardozcm:
## Description

In this PR we port meta's official Llama on bigdl-llm `optimize_model` API.

https://github.com/analytics-zoo/nano/issues/591


python/llm/example/pytorch-models/meta-llama/README.md
```diff
@@ -0,0 +1,94 @@
+# Llama
+
+In this directory, you will find examples on how you could apply BigDL-LLM INT4 optimizations on general pytorch models, for example Meta Llama models. For illustration purposes, we utilize the [Llama2-7b-Chat](https://ai.meta.com/llama/) as a reference Llama model.
```
jason-dai, line 3: Need to make it clear:
1) This example shows how to apply BigDL-LLM optimizations to PyTorch models which are not HF 
2) It use the official LLaMA inference code from Meta (which does not use Hugging Face

leonardozcm: OK, I've highlighted this code source.

shane-huang (approved):
LGTM

PR #9106, opened 2023-10-09
[LLM] Small update to performance tests

Oscilloscope98:
## Description

- Avoid the error when the original model is missing on the test machines. 
- Unify the dealing method of original models for arc/cpu tests


rnwang04 (approved):
LGTM

Oscilloscope98:
The performance test will passed in today's nightly tests, after today's bigdl-core-xe is released.

PR #9125, opened 2023-10-10
update falcon 7b model xpu

qiuxin2012:
## Description

fuse rope in falcon 7b model xpu

### 1. Why the change?

Speedup falcon 7b's performance on xpu.

### 2. User API changes

new falcon modelling file

### DIFF
--- a/python/llm/example/CPU/HF-Transformers-AutoModels/Model/falcon/falcon-7b-instruct/modelling_RW.py
+++ b/python/llm/example/CPU/HF-Transformers-AutoModels/Model/falcon/falcon-7b-instruct/modelling_RW.py
@@ -294,7 +294,32 @@ class Attention(nn.Module):
             _, seq_len_past, _ = layer_past[0].shape
 
             seq_len = seq_len + seq_len_past
-        query_layer, key_layer = self.maybe_rotary(query_layer, key_layer, seq_len)
+            position_ids = torch.tensor([range(seq_len_past, seq_len)], dtype=torch.long).to(query_layer.device)
+        else:
+            position_ids = torch.tensor([range(0, seq_len)], dtype=torch.long).to(query_layer.device)
+
+        use_fuse_rope = query_layer.device.type == "xpu"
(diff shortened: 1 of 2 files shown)


yangw1234:
https://github.com/intel-analytics/llm.cpp/pull/117 should support q and k of different sizes.

hkvision:
> [intel-analytics/llm.cpp#117](https://github.com/intel-analytics/llm.cpp/pull/117) should support q and k of different sizes.

The perf is not improved after supporting this?
Seems even a little bit slower, not sure if it is due to instability.

python/llm/example/CPU/HF-Transformers-AutoModels/Model/falcon/falcon-7b-instruct/modelling_RW.py
```diff
@@ -294,7 +295,13 @@ def forward(
...
 
             seq_len = seq_len + seq_len_past
-            position_ids = torch.tensor([range(seq_len_past, seq_len)], dtype=torch.long).to(query_layer.device)
+            if (seq_len_past, seq_len) in self.p_dict:
+                #print("get cache, " + str(seq_len_past) + "  " + str(seq_len))
+                position_ids = self.p_dict[(seq_len_past, seq_len)]
+            else:
+                #print("create new" + str(seq_len_past) + "  " + str(seq_len))
+                position_ids = torch.tensor([range(seq_len_past, seq_len)], dtype=torch.long).to(query_layer.device)
+                self.p_dict[(seq_len_past, seq_len)] = position_ids
```
jason-dai, line 304: do we need to store every seq_len?

qiuxin2012: Yes, store every seq_len is better than `new every time` and `cache one, and assign value`.
The best option should be add a new kernel function to accept seq_len directly.

jason-dai:
We should follow the implementation of llama2

qiuxin2012:
> We should follow the implementation of llama2

See https://github.com/intel-analytics/BigDL/pull/9242, close this one.

PR #9126, opened 2023-10-10
LLM: update benchmark_util.py for beam search

rnwang04:
## Description

update benchmark_util.py for beam search

### 4. How to test?
- [x] Unit test


liu-shaojun (approved):
LGTM

PR #9131, opened 2023-10-10
LLM: add missing local variable named result for benchmark script

WeiguangHan:
## Description
LLM: add missing local variable named result for benchmark script.


hkvision:
See: https://github.com/intel-analytics/BigDL/issues/9133#issuecomment-1758816604
native int4 currently will not return anything and this it won't need to process the results.

PR #9141, opened 2023-10-11
LLM: support num_beams in all-in-one benchmark

rnwang04:
## Description

- Add num_beams parameter in all-in-one benchmark script to support beam search benchmark
- Fix input length to `in_len`

now benchmark excel will output look like:
```bash
,model,1st token avg latency (s),2+ avg latency (s/token),encoder time (s),input/output tokens,actual input/output tokens,num_beams
0,tiiuae/falcon-7b-instruct-with-patch,xxx,xxx,0.0,32-32,32-32,1
1,tiiuae/falcon-7b-instruct-with-patch,xxx,xxx,0.0,1024-128,1024-128,1
2,tiiuae/falcon-7b-instruct-with-patch,xxx,xxx,2048-256,2048-256,1
```

### 4. How to test?
- [x] Unit test
- [x] Local test

### DIFF
--- a/python/llm/dev/benchmark/all-in-one/run.py
+++ b/python/llm/dev/benchmark/all-in-one/run.py
@@ -441,7 +441,7 @@ def run_optimize_model_gpu(repo_id,
             input_ids = tokenizer.encode(input_str, return_tensors="pt")
             input_ids = input_ids[:, :in_len]
             true_str = tokenizer.batch_decode(input_ids)[0]
-            input_ids = tokenizer.encode(true_str, return_tensors="pt").[:, :in_len].to('xpu')
+            input_ids = tokenizer.encode(true_str, return_tensors="pt")[:, :in_len].to('xpu')
             actual_in_len = input_ids.shape[1]
             result[in_out] = []
             for i in range(num_trials + warm_up):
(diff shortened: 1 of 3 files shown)


python/llm/dev/benchmark/all-in-one/run.py
```diff
@@ -159,12 +163,13 @@ def run_transformer_int4(repo_id,
             input_ids = tokenizer.encode(input_str, return_tensors="pt")
             input_ids = input_ids[:, :in_len]
             true_str = tokenizer.batch_decode(input_ids)[0]
-            input_ids = tokenizer.encode(true_str, return_tensors="pt")
+            input_ids = tokenizer.encode(true_str, return_tensors="pt")[:, :in_len]
```
plusbang, line 166: `true_str` is processed to match the input length, do we need [:, :in_len] here?

rnwang04: Yes, in my local test, when without `[:, :in_len]`, the input length maybe 1 token longer than we need.
For example, for falcon, 1024 input actually is tested with 1025 length input id.

plusbang (approved):
LGTM

Issue #9140, opened 2023-10-11
Check model more strictly before applying attention optimization


plusbang:
Refer to https://github.com/intel-analytics/BigDL/blob/main/python/llm/src/bigdl/llm/transformers/convert.py#L184, we check model through `model.config._name_or_path` before applying attention optimization, but this may cause error when loading model from local files.

If download `chatglm2-6B` model files to a local directory `/mnt/disk1/models/test-chatglm` and use transformers-style API with ` optimize_model=True` to run chatglm2, there exists error:
```python
Traceback (most recent call last):
  File "/home/arda/binbin/temp.py", line 62, in <module>
    output = model.generate(input_ids,
  File "/home/arda/anaconda3/envs/binbin-llm-cpu/lib/python3.9/site-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "/home/arda/anaconda3/envs/binbin-llm-cpu/lib/python3.9/site-packages/transformers/generation/utils.py", line 1538, in generate
    return self.greedy_search(
  File "/home/arda/anaconda3/envs/binbin-llm-cpu/lib/python3.9/site-packages/transformers/generation/utils.py", line 2362, in greedy_search
    outputs = self(
  File "/home/arda/anaconda3/envs/binbin-llm-cpu/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1501, in _call_impl
    return forward_call(*args, **kwargs)
  File "/home/arda/.cache/huggingface/modules/transformers_modules/test-chatglm/modeling_chatglm.py", line 845, in forward
    transformer_outputs = self.transformer(
  File "/home/arda/anaconda3/envs/binbin-llm-cpu/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1501, in _call_impl
    return forward_call(*args, **kwargs)
  File "/home/arda/.cache/huggingface/modules/transformers_modules/test-chatglm/modeling_chatglm.py", line 741, in forward
    hidden_states, presents, all_hidden_states, all_self_attentions = self.encoder(
  File "/home/arda/anaconda3/envs/binbin-llm-cpu/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1501, in _call_impl
    return forward_call(*args, **kwargs)
  File "/home/arda/.cache/huggingface/modules/transformers_modules/test-chatglm/modeling_chatglm.py", line 588, in forward
    hidden_states, kv_cache = layer(
[...]

PR #9149, opened 2023-10-12
LLM: fix model check before attention optimization

plusbang:
## Description

Related to https://github.com/intel-analytics/BigDL/issues/9140

Use `architectures` and `vocab_size` instead of path to distinguish between chatglm and chatglm2.
And use `model_type` instead of path to check mpt.

### 2. User API changes

N/A

### 4. How to test?
- [ ] Unit test
- [x] Local test


rnwang04 (approved):
LGTM

PR #9155, opened 2023-10-12
Optimize rope for gpt_neox on GPU

jenniew:
The original rope create cache in cpu. So every time copy data to xpu. 
Optimize rope for redpajama-7b on GPU 
Use bigdl-llm fused rope for redpajama to avoid cache on cpu issue and also fuse some kernel operations.


yangw1234 (approved):
LGTM

PR #9160, opened 2023-10-13
Add stable diffusion module to nano

NovTi:
### 1. About the change
Add the stable diffusion pipeline module to nano, currently has ipex and ov pipelines

### 2. More info and about testing 

Already tested with wheel build, now people can use the diffusion pipe by running `install bigdl-nano[diffusion]` 
and run
`NanoStableDiffusionPipeline.from_pretrained(YOUR MODEL PATH, device='CPU', precision='bfloat16')`

### 3. New dependencies

- New Python dependencies
       - diffusers==0.14.0
       - transformers
       - openvino-dev==2022.2.0


python/nano/src/bigdl/nano/diffusion/diffusers/pipelines/stable_diffusion.py
```diff
@@ -0,0 +1,237 @@
...
+    """NanoDiffusionPipeline."""
+
+    @staticmethod
+    def from_pretrained(
+            cls: DiffusionPipeline,
+            pretrained_model_name_or_path,
+            scheduler=default_scheduler,
+            device='iGPU',
+            precision='float16',
+            backend='OV',
```
shane-huang, line 134: 1. change all default backend to IPEX.
2. move `backend` parameter to the end of the parameter list, so that user can use unamed args for other params.

We can keep the OV related code there but we don't explicitly showcase `backend=` parameter in examples or README. By default we use IPEX and user doesn't have to set it.

python/nano/setup.py
```diff
@@ -144,6 +144,13 @@ def setup_package():
                         "packaging",
                         "sigfig",
                         "setuptools<66"]
+    
+    # diffusion requires
+    diffusion_requires = pytorch_requires + [
+        "diffusers==0.14.0",
+        "transformers",
```
shane-huang, line 151: add transformers lib version

python/nano/setup.py
```diff
@@ -144,6 +144,13 @@ def setup_package():
...
+        "openvino-dev==2022.2.0"
```
shane-huang, line 152: remove openvino dependency. By default install option we support only pytorch 113, linux, ipex

python/nano/src/bigdl/nano/diffusion/diffusers/api.py
```diff
@@ -0,0 +1,36 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import click
+
+from diffusers import StableDiffusionPipeline
+
+
```
shane-huang, line 21: We can remove this file diffusers folder since we don't use it right now. If we need it we can added it later.

python/nano/src/bigdl/nano/diffusion/diffusers/pipelines/stable_diffusion.py
```diff
@@ -0,0 +1,237 @@
...
+
+class NanoDiffusionPipeline:
+    """NanoDiffusionPipeline."""
+
+    @staticmethod
+    def from_pretrained(
+            cls: DiffusionPipeline,
+            pretrained_model_name_or_path,
+            scheduler=default_scheduler,
+            device='iGPU',
```
shane-huang, line 132: change default device to CPU

shane-huang: by default we support IPEX+CPU, change device, precision and backend accordingly.

shane-huang: please verify your default config can run successfully on SPR. And for those CPU without bfloat16 support, e.g. ICX, please also verify which precision configuration works.

shane-huang (approved):
Merge for now and add tests and updates later.

jason-dai:
Do we have examples for SD?

PR #9163, opened 2023-10-13
LLM: skip some model tests using certain api

WeiguangHan:
@hkvision Currently pytorch do not support bfloat16 on cpu for chatglm models. So we will skip this type of test.

### DIFF
--- a/python/llm/dev/benchmark/all-in-one/run.py
+++ b/python/llm/dev/benchmark/all-in-one/run.py
@@ -40,6 +40,7 @@ results = []
 
 def run_model(repo_id, test_api, in_out_pairs, local_model_hub=None, warm_up=1, num_trials=3, num_beams=1):
     # TODO: make a parameter
+    result= {}
     if test_api == 'transformer_int4':
         result = run_transformer_int4(repo_id, local_model_hub, in_out_pairs, warm_up, num_trials, num_beams)
     elif test_api == 'native_int4':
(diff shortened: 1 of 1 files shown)


python/llm/dev/benchmark/all-in-one/run.py
```diff
@@ -56,14 +56,15 @@ def run_model(repo_id, test_api, in_out_pairs, local_model_hub=None, warm_up=1,
...
     for in_out_pair in in_out_pairs:
-        results.append([repo_id,
-                        np.mean(result[in_out_pair], axis=0)[0],
-                        np.mean(result[in_out_pair], axis=0)[1],
-                        np.mean(result[in_out_pair], axis=0)[2],
-                        in_out_pair,
-                        f'{int(np.mean(result[in_out_pair], axis=0)[3])}' +
-                        f'-{int(np.mean(result[in_out_pair], axis=0)[4])}',
-                        num_beams])
+        if result is not None:
```
hkvision, line 59: do we need to initialize result=[] somewhere?

lalalapotter: `result` is initialized [here](https://github.com/intel-analytics/BigDL/blob/main/python/llm/dev/benchmark/all-in-one/run.py#L38)

lalalapotter (approved):
LGTM

python/llm/dev/benchmark/all-in-one/run.py
hkvision: this is results not result.
I mean we need to initialize result in case native_int4 don't return anything?

lalalapotter: That's correct, sorry for the misunderstanding.

hkvision (approved):

PR #9164, opened 2023-10-13
add ipex bf16 optimization for transformers dtype

Uxito-Ada:
## Description

enable ipex bf16 in transformers compute type

### 1. Why the change?

to speed up token computing

### 2. User API changes

append `bf16` in `load_low_bit` options

### 3. Summary of the change 

as above

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [x] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/model.py
+++ b/python/llm/src/bigdl/llm/transformers/model.py
@@ -206,8 +206,8 @@ class _BaseAutoModelClass:
                           "bf16 only supports from_pretrained, instead of load_low_bit")
 
         invalidInputError(bigdl_transformers_low_bit in ggml_tensor_qtype,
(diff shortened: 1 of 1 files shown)


Uxito-Ada:
ipex now only uses bf16 as a compute dtype, rather than store dtype, and does not have save/load bf16-quantized model support. Also, due to it is based on AVX512 rather than GGML, we cannot reuse existing sym_int4 save/load for bf16. Therefore, we do not implement `save_low_bit` and `load_low_bit` for bf16.

Uxito-Ada:
`_optimize_transformers` only works in ipex 4.21 dev now. In the master branch, there is a similar API named `ipex.transformers.optimize.optimize_transformers`, which seems to be a regular release. However, the latter one is without wheel file and cannot be accessed one. Therefore, our validation and code is currently using `_optimize_transformers`.

PR #9177, opened 2023-10-16
Update JSON-Java version to 20131013 in ppml/pom.xml

liu-shaojun:
## Description

This PR is to fix security issue: https://github.com/intel-analytics/BigDL/security/dependabot/2837


glorysdj (approved):
LGTM

PR #9183, opened 2023-10-16
LLM: add save & load usage in PyTorch API key feature page

plusbang:
## Description

Improvements in PyTorch API key feature page:
- Add save & load usage
- Update notes

### 4. How to test?
- [x] Document test (https://binbin-doc.readthedocs.io/en/improve-doc-v3/doc/LLM/Overview/KeyFeatures/optimize_model.html)

### DIFF
--- a/docs/readthedocs/source/doc/LLM/Overview/KeyFeatures/optimize_model.md
+++ b/docs/readthedocs/source/doc/LLM/Overview/KeyFeatures/optimize_model.md
@@ -2,6 +2,8 @@
 
 In general, you just need one-line `optimize_model` to easily optimize any loaded PyTorch model, regardless of the library or API you are using. With BigDL-LLM, PyTorch models (in FP16/BF16/FP32) can be optimized with low-bit quantizations (supported precisions include INT4, INT5, INT8, etc).
 
+### Optimize model
+
 First, use any PyTorch APIs you like to load your model. To help you better understand the process, here we use [Hugging Face Transformers](https://huggingface.co/docs/transformers/index) library `LlamaForCausalLM` to load a popular model [Llama-2-7b-chat-hf](https://huggingface.co/meta-llama/Llama-2-7b-chat-hf) as an example:
 
 ```python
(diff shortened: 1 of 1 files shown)


docs/readthedocs/source/doc/LLM/Overview/KeyFeatures/optimize_model.md
```diff
@@ -18,10 +20,46 @@ from bigdl.llm import optimize_model
...
+```
+
 After optimizing the model, BigDL-LLM does not require any change in the inference code. You can use any libraries to run the optimized model with very low latency.
 
+### Save & Load Optimized Model
+
+The loading process of the original model may be time-consuming and memory-intensive. For example, the [Llama-2-7b-chat-hf](https://huggingface.co/meta-llama/Llama-2-7b-chat-hf) model is stored with float16 precision, resulting in large memory usage when loaded using `LlamaForCausalLM`. To avoid high resource consumption and expedite loading process, you can use `save_low_bit` to store the model after low-bit optimization. Then, in subsequent uses, you can opt to use the `load_low_bit` API to directly load the optimized model. Besides, saving and loading operations are platform-independent, regardless of their operating systems.
+#### Save
+
+Continuing with the example of [Llama-2-7b-chat-hf](https://huggingface.co/meta-llama/Llama-2-7b-chat-hf), we can save the previously optimized model as follows:
```
Oscilloscope98, line 41: May be here it would be better to jump to the # Optimize Model section instead of jumping to the model link

plusbang: > May be here it would be better to jump to the # Optimize Model section instead of jumping to the model link

yeah, I've updated the link

docs/readthedocs/source/doc/LLM/Overview/KeyFeatures/optimize_model.md
```diff
@@ -18,10 +20,46 @@ from bigdl.llm import optimize_model
...
+```python
+saved_dir='./llama-2-bigdl-llm-4-bit'
+model.save_low_bit(saved_dir)
+```
+#### Load
+
+Use `load_low_bit` to load the optimized low-bit model as follows:
+```python
+from bigdl.llm.optimize import low_memory_init, load_low_bit
+with low_memory_init(): # Fast and low cost by loading model on meta device
```
shane-huang, line 51: Remind user to use `with low_memory_init()` statement when loading, either in notice or in descriptions.

docs/readthedocs/source/doc/LLM/Overview/KeyFeatures/optimize_model.md
```diff
@@ -18,10 +20,46 @@ from bigdl.llm import optimize_model
 model = optimize_model(model)
 ```
 
+```eval_rst
+.. note::
```
shane-huang, line 24: maybe add a new section "more precisions" for IN3/INT8, etc. instead of put it in Note. Sometime impatient readers just ignore those notes.

shane-huang (approved):

docs/readthedocs/source/doc/LLM/Overview/KeyFeatures/optimize_model.md
plusbang: > and maybe add some explanations about the difference of sym and asym

Have added descriptions according to above comments.

Issue #9185, opened 2023-10-16
NoneType Error when using TransformersEmbeddings


Kailuo-Lai:
When I try to use "sentence-transformers/all-MiniLM-L12-v2" embedding model downloaded from huggingface, I encounter a wired problem

**Code**

```python
from bigdl.llm.langchain.embeddings import TransformersEmbeddings
embeddings = TransformersEmbeddings.from_model_id(model_id="./checkpoints/all-MiniLM-L12-v2")
``` 

**Output**

```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[3], line 2
      1 llm = TransformersLLM.from_model_id_low_bit("./checkpoints/Llama-2-7b-chat-hf-INT4/")
----> 2 embeddings = TransformersEmbeddings.from_model_id(model_id="./checkpoints/all-MiniLM-L12-v2")

File ~/anaconda3/envs/llm-tutorial-new/lib/python3.9/site-packages/bigdl/llm/langchain/embeddings/transformersembeddings.py:117, in TransformersEmbeddings.from_model_id(cls, model_id, model_kwargs, **kwargs)
    114 except:
    115     tokenizer = LlamaTokenizer.from_pretrained(model_id, **_model_kwargs)
--> 117 model = AutoModel.from_pretrained(model_id, load_in_4bit=True, **_model_kwargs)
    119 if "trust_remote_code" in _model_kwargs:
    120     _model_kwargs = {
    121         k: v for k, v in _model_kwargs.items() if k != "trust_remote_code"
    122     }

File ~/anaconda3/envs/llm-tutorial-new/lib/python3.9/site-packages/bigdl/llm/transformers/model.py:95, in _BaseAutoModelClass.from_pretrained(cls, *args, **kwargs)
     93         kwargs["pretraining_tp"] = 1
     94     q_k = load_in_low_bit if load_in_low_bit else "sym_int4"
---> 95     model = cls.load_convert(q_k, optimize_model, *args, **kwargs)
     96 else:
     97     # load default
     98     model = cls.HF_Model.from_pretrained(*args, **kwargs)

File ~/anaconda3/envs/llm-tutorial-new/lib/python3.9/site-packages/bigdl/llm/transformers/model.py:125, in _BaseAutoModelClass.load_convert(cls, q_k, optimize_model, *args, **kwargs)
    123     model.config.update({"bigdl_lcmu_enabled": False})
...
[...]

PR #9190, opened 2023-10-17
LLM: fix convert of chatglm

rnwang04:
## Description

Fix convert of chatglm so that don't influence other models.

### 1. Why the change?

https://github.com/intel-analytics/BigDL/issues/9185

### 2. User API changes

No change.

### 3. Summary of the change 

Fix convert of chatglm so that don't influence other models.

### 4. How to test?
- [ ] Unit test


plusbang (approved):
LGTM

PR #9198, opened 2023-10-17
Changed the path '/ppml' into '/bigdl'

Jasonzzt:
## Description

Changed the path '/ppml' into '/bigdl' in README.md, Dockerfile and bigdl-lora-finetuing-entrypoint.sh.


Uxito-Ada:
pls also modify files under `kubernetes` directory, like [bigdl-lora-finetuning-job.yaml](https://github.com/intel-analytics/BigDL/tree/main/docker/llm/finetune/lora/cpu/kubernetes/templates).

Uxito-Ada:
By the way, change [here](https://github.com/intel-analytics/BigDL/blob/main/docker/llm/finetune/lora/cpu/kubernetes/values.yaml#L7) to llama2

Issue #9200, opened 2023-10-17
baichuan2-13B generation error when using `optimize_model=True` (by default)


shane-huang:
Using code below, Baichuan2-7B-Chat works correclty, Baichuan2-13B-Chat throws error. Load/Optimize okay, generate fails.
```python
    model = AutoModelForCausalLM.from_pretrained(model_path,
                                                 load_in_4bit=True,
                                                 trust_remote_code=True,
                                                 use_cache=True)

    # Load tokenizer
    tokenizer = AutoTokenizer.from_pretrained(model_path,
                                              trust_remote_code=True)

    # Generate predicted tokens
    with torch.inference_mode():
        prompt = BAICHUAN_PROMPT_FORMAT.format(prompt=args.prompt)
        input_ids = tokenizer.encode(prompt, return_tensors="pt")
        output = model.generate(input_ids,
                                max_new_tokens=args.n_predict)
```

Error looks like. 

```
  File "/home/cpx/.cache/huggingface/modules/transformers_modules/Baichuan2-13B-Chat/modeling_baichuan.py", line 241, in forward
    hidden_states = self.input_layernorm(hidden_states)
  File ".../lib/python3.9/site-packages/torch/nn/modules/module.py", line 1518, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File ".../lib/python3.9/site-packages/torch/nn/modules/module.py", line 1527, in _call_impl
    return forward_call(*args, **kwargs)
  File ".../lib/python3.9/site-packages/bigdl/llm/transformers/models/llama.py", line 69, in llama_rms_norm_forward
    hidden_states = hidden_states * torch.rsqrt(variance + self.variance_epsilon)
  File .../lib/python3.9/site-packages/torch/nn/modules/module.py", line 1695, in __getattr__
    raise AttributeError(f"'{type(self).__name__}' object has no attribute '{name}'")
   AttributeError: 'RMSNorm' object has no attribute 'variance_epsilon'
```

Workaround is to set `optimize_model=False`. i.e.
```python
model = AutoModelForCausalLM.from_pretrained(model_path, load_in_4bit=True, use_cache=True, trust_remote_code=True, optimize_model=False)
```

rnwang04:
Baichuan-13B/Baichuan2-13B 's RMSNorm is a little different from llama's, I have fixed in https://github.com/intel-analytics/BigDL/pull/9204

PR #9204, opened 2023-10-17
LLM: fix RMSNorm optimization of Baichuan2-13B

rnwang04:
## Description

Baichuan2-13B's RMSNorm is a little different from llama's (`self.variance_epsilon`- > `self.epsilon`), which causes error on CPU for current RMSNorm optimization. This PR support specific RMSNorm optimization for Baichuan2-13B.

### 1. Why the change?

https://github.com/intel-analytics/BigDL/issues/9200

### 2. User API changes

No change, just make `optimize_model=True` works for Baichuan2-13B on CPU.

### 3. Summary of the change 

- support specific RMSNorm optimization for Baichuan2-13B

### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/convert.py
+++ b/python/llm/src/bigdl/llm/transformers/convert.py
@@ -275,19 +275,23 @@ def optimize(model):
                             module.Attention,
                             baichuan_attention_forward_7b
                             )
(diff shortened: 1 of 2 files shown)


shane-huang (approved):
LGTM

PR #9206, opened 2023-10-17
LLM: add rms norm forward for different models.

lalalapotter:
## Description

This PR is to add rms norm forward for each model(Baichuan, Aquila, Mistral)


jason-dai:
Is this PR ready for merge?

lalalapotter:
Already fixed in https://github.com/intel-analytics/BigDL/pull/9204, temporarily close for now.

PR #9209, opened 2023-10-18
changed '/ppml' into '/bigdl' and modified llama-7b

Jasonzzt:
[changed '/ppml' into '/bigdl' and modified llama-7b]


Uxito-Ada (approved):
lgtm

PR #9213, opened 2023-10-18
Enable unit test workflow on Arc

chtanch:
## Description
Enable github workflow for llm unit test on Arc.  An inference test for transformers API is also implemented to verify that the workflow runs correctly. The inference test queries models with a simple question and checks for the correct answer.

### DIFF
--- a/.github/workflows/llm_unit_tests.yml
+++ b/.github/workflows/llm_unit_tests.yml
@@ -259,9 +259,5 @@ jobs:
         shell: bash
         run: |
           source /opt/intel/oneapi/setvars.sh
-          export USE_XETLA=OFF
-          export SYCL_PI_LEVEL_ZERO_USE_IMMEDIATE_COMMANDLISTS=1
-          export DEVICE=xpu
-
           python -m pip install expecttest
           bash python/llm/test/run-llm-inference-tests-gpu.sh
--- a/.github/workflows/llm_unit_tests.yml
+++ b/.github/workflows/llm_unit_tests.yml
@@ -186,3 +186,82 @@ jobs:
           pip install -U pandas==2.0.3
           pip install -U typing_extensions==4.5.0
           bash python/llm/test/run-llm-langchain-tests.sh
+  llm-unit-test-on-arc:
+    needs: llm-cpp-build
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.9"]
+    runs-on: [self-hosted, llm, perf]
+    env:
+      OMP_NUM_THREADS: 16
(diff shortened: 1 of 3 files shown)


.github/workflows/llm_unit_tests.yml
```diff
@@ -186,3 +186,78 @@ jobs:
...
+        with:
+          extra-dependency: "xpu"
+
+      - name: Test installed xpu version
+        shell: bash
+        run: |
+          source /opt/intel/oneapi/setvars.sh
+          bash python/llm/test/run-llm-install-tests.sh
+
+      - name: Download LLMs
```
hkvision, line 238: I don't think we need to download models here?

python/llm/test/run-llm-inference-tests-gpu.sh
```diff
@@ -0,0 +1,26 @@
...
+if [ -z "$THREAD_NUM" ]; then
+  THREAD_NUM=2
+fi
+export OMP_NUM_THREADS=$THREAD_NUM
+pytest ${LLM_INFERENCE_TEST_DIR} -v -s 
+
+now=$(date "+%s")
+time=$((now-start))
+
+echo "Bigdl-llm tests finished"
```
hkvision, line 25: Bigdl-llm gpu tests finished

python/llm/test/inference_gpu/test_transformers_api.py
```diff
@@ -0,0 +1,52 @@
...
+    import intel_extension_for_pytorch as ipex
+
+@pytest.mark.parametrize('prompt, answer', [
+    ('What is the capital of France?\n\n','Paris')
+    ])
+@pytest.mark.parametrize('Model, Tokenizer, model_path',[
+    (AutoModelForCausalLM, LlamaTokenizer, os.environ.get('LLAMA2_7B_ORIGIN_PATH')),
+    (AutoModel, AutoTokenizer, os.environ.get('CHATGLM2_6B_ORIGIN_PATH')),
+    (AutoModelForCausalLM, AutoTokenizer, os.environ.get('FALCON_7B_ORIGIN_PATH')),
+    ])
```
WeiguangHan, line 36: I think we may have missed MPT_7B_ORIGIN_PATH here?

hkvision: let's add it in another PR

hkvision (approved):

python/llm/test/inference_gpu/test_transformers_api.py
chtanch: Thanks, will include it in another PR

PR #9226, opened 2023-10-19
LLM: add Mistral support for Portable-Zip

JinBridger:
## Description

Add support for Mistral on Portable-Zip. 

- However, Mistral needs `transformers==4.34.0`, so we need to upgrade `transformers` to 4.34.0 before we run mistral.
- Also, most of our model run on `transformers==4.31.0`, so we need to downgrade `transformers` to 4.31.0 after we run mistral.

I wrote a powershell script `chat.ps1` to read `config.json` to check if the model is mistral. Most of the logic is moved into the `chat.ps1` and what `chat.bat` does is to simply run `chat.ps1`.

The usage still remains the same, which is double-click the `chat.bat`.

## Screenshot

anime

### DIFF
--- a/python/llm/portable-zip/README.md
+++ b/python/llm/portable-zip/README.md
@@ -17,6 +17,7 @@ This portable zip includes everything you need to run an LLM with BigDL-LLM opti
 - Baichuan2-7B-Chat
 - internlm-chat-7b-8k
 - Llama-2-7b-chat-hf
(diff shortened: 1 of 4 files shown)


jason-dai:
I think we should upgrade transformer to 4.34 version for bigdl-llm instead

shane-huang:
pending until we verified transformers 4.34.1

PR #9229, opened 2023-10-19
support transformers 4.34+ for llama

yangw1234:
## Description

support transformers 4.34+ for llama


rnwang04 (approved):
LGTM

rnwang04:
Maybe we should try upgrade transformers to 4.34.0 in llm's `setup.py` ? @jason-dai

jason-dai:
> Maybe we should try to upgrade transformers to 4.34.0 in llm's `setup.py` ? @jason-dai At least now Mistral and qlora training both needs `transformers>=4.34.0`.

Let's do that after the release

PR #9239, opened 2023-10-23
Add load low-bit in model-serving for reduce EPC

hzjane:
### 1. Why the change?
Add load low-bit in model-serving to reduce max EPC.


gc-fu (approved):
LGTM

PR #9242, opened 2023-10-23
add position_ids and fuse embedding for falcon

qiuxin2012:
## Description

add position_ids and fuse embedding for falcon
Fixed https://github.com/analytics-zoo/nano/issues/637

### 1. Why the change?

add position_ids for falcon, like https://github.com/huggingface/transformers/blob/v4.34.1/src/transformers/models/falcon/modeling_falcon.py
and fuse rotary embedding on xpu 

### 2. User API changes

A new falcon modelling file.


yangw1234 (approved):
LGTM

PR #9245, opened 2023-10-23
[CICD] add manually testing item for bigdl-llm-cpu-serving

gc-fu:
## Description

add manually testing item for bigdl-llm-cpu-serving


liu-shaojun (approved):
LGTM

PR #9247, opened 2023-10-23
Fix bigdl-llm-serving-cpu Dockerfile

gc-fu:
## Description

Fix bigdl-llm-serving-cpu Dockerfile by replacing `COPY` with `ADD`


liu-shaojun (approved):
LGTM

PR #9251, opened 2023-10-24
[CICD] Fix image build for bigdl-llm-serving-tdx image

gc-fu:
## Description

Fix image build for bigdl-llm-serving-tdx


hzjane (approved):
LGTM

PR #9253, opened 2023-10-24
[CICD] Refine README/CICD

gc-fu:
## Description

1. Refine README for bigdl.llm.serving
2. Fix the wrong manually_build script.


hzjane (approved):
LGTM

PR #9254, opened 2023-10-24
[LLM] Replace Embedding layer to fix it on CPU

MeouSker77:
## Description

Replace `Embedding` layer when loading model in low bit to fix it on CPU

Skip user-defined `Embedding` layer because it may have different `forward` signature with standard `Embedding` layer, such as whisper.

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/convert.py
+++ b/python/llm/src/bigdl/llm/transformers/convert.py
@@ -109,7 +109,8 @@ def _replace_with_low_bit_linear(model, qtype, modules_to_not_convert=None,
                         model._modules[name].requires_grad_(False)
 
                         module.weight = None
-        elif isinstance(module, nn.Embedding) and name not in modules_to_not_convert:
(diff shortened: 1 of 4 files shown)


rnwang04 (approved):
LGTM

python/llm/src/bigdl/llm/transformers/convert.py
```diff
@@ -107,6 +109,19 @@ def _replace_with_low_bit_linear(model, qtype, modules_to_not_convert=None,
                         model._modules[name].requires_grad_(False)
 
                         module.weight = None
+        elif type(module) == nn.Embedding and name not in modules_to_not_convert:
```
jason-dai, line 112: I don't think we want to do it unconditionally; maybe adding some config to control behavior?

MeouSker77:
use a better way to workaround ipex bug, now we can call the Embedding layer on XPU, no need to move it to CPU any more

PR #9257, opened 2023-10-24
LLM: add `low_bit` option in benchmark scripts

plusbang:
## Description

To benchmark int8/fp8, add `low_bit` option in benchmark scripts

### 2. User API changes

We could specify low_bit parameter in config.yaml
The output csv maybe as following:
```bash
,model,1st token avg latency (s),2+ avg latency (s/token),encoder time (s),input/output tokens,actual input/output tokens,num_beams,low_bit
0,THUDM/chatglm2-6b,xxxxx,xxxxxx,0.0,32-32,32-32,1,sym_int8
```

### 4. How to test?
- [x] Unit test
- [x] Local test


rnwang04 (approved):
LGTM

PR #9261, opened 2023-10-24
LLM: replace torch.float32 with auto type

JinBridger:
## Description

Replace `torch.float32` with `auto` to get lower resource needed.

Here's comparison when loading `chatglm2-6b`:

### Using `auto`

image

### Using `torch.float32` (Actually I killed the process after mem reach 90%, otherwise it will continue loading and crash my computer)

image


gc-fu (approved):
LGTM

PR #9262, opened 2023-10-24
LLM: add UI for portable-zip

JinBridger:
## Description

**Add `--chat` argument to `setup.bat` so that**
- when run `setup.bat`, it will generate portable zip without web-ui.
- when run `setup.bat --ui`, it will generate portable zip with web-ui.

**The usage of portable zip with/without web-ui is different:**
- For portable zip without web-ui, the usage still remains the same, which is double-click `chat.bat` and input path/to/model.
- However, for portable zip with web-ui, the usage is double-click `chat-ui.bat` and visit `127.0.0.1:7860` in browser.

**Add the usage of portable zip with web-ui in `README-ui.md`.**

**Update the usage of `setup.bat` in `setup.md`**

**⚠️ I didn't add gif in README-ui.md. I'll add it after I upload it in assets.** 

## Screenshots (web-ui, running chatglm2-6b)

Animation

### DIFF
--- a/python/llm/portable-zip/.gitignore
+++ b/python/llm/portable-zip/.gitignore
@@ -1,2 +1,4 @@
 python-embed
 bigdl-llm.zip
+*.log
+*.json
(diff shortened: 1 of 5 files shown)


shane-huang (approved):
Merge this PR for now and refine later if any.

PR #9263, opened 2023-10-25
add tf-int4 example for auto comment application

Ariadne330:
## Description

Add transformer int4 example for auto comment application, contain 3 files:
- auto_comment.py 
- test_codes.py
- Readme.md

Usage details refer to `Readme.md`.

### DIFF
--- a/python/llm/example/CPU/Application/auto_comment/README.md
+++ b/python/llm/example/CPU/Application/auto_comment/README.md
@@ -1,0 +1,116 @@
+# Auto Comment
+
+In this directory, you will find examples on how you could apply BigDL-LLM INT4 optimizations on Qwen models to automatically generate comments for code files. For illustration purposes, we utilize the [Qwen/Qwen-VL-7B](https://huggingface.co/Qwen/Qwen-VL-7B) as a reference Qwen model.
+
+> **Note**: If you want to download the Hugging Face *Transformers* model, please refer to [here](https://huggingface.co/docs/hub/models-downloading#using-git).
+>
+> BigDL-LLM optimizes the *Transformers* model in INT4 precision at runtime, and thus no explicit conversion is needed.
+
+## Requirements
+To run these examples with BigDL-LLM, we have some recommended requirements for your machine, please refer to [here](../README.md#recommended-requirements) for more information.
+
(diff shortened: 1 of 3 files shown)


python/llm/example/CPU/Application/auto_comment/README.md
```diff
@@ -0,0 +1,116 @@
+# Auto Comment
+
+In this directory, you will find examples on how you could apply BigDL-LLM INT4 optimizations on Qwen models to automatically generate comments for code files. For illustration purposes, we utilize the [Qwen/Qwen-VL-7B](https://huggingface.co/Qwen/Qwen-VL-7B) as a reference Qwen model.
```
shane-huang, line 3: > In this directory, you will find examples on how you could apply BigDL-LLM INT4 optimizations on Qwen models to automatically generate comments for code files.

change to

In this directory, you will find an example of using LLM to automatically generate comments for code files. This example is adapted from \[QWen Example\](add link to reference code).  We utilize the [Qwen/Qwen-VL-7B](https://huggingface.co/Qwen/Qwen-VL-7B) optimized w/ BigDL-LLM INT4 for illustration.

python/llm/example/CPU/Application/auto_comment/auto_comment.py
```diff
@@ -0,0 +1,204 @@
...
+import torch
+import time
+import argparse
+import os
+
+from bigdl.llm.transformers import AutoModelForCausalLM
+from transformers import AutoTokenizer
+from transformers.generation import GenerationConfig
+
+MaxLine = 50 # 限制单次处理最大代码行数
```
shane-huang, line 26: change the comments to English

python/llm/example/CPU/Application/auto_comment/auto_comment.py
```diff
@@ -0,0 +1,204 @@
+#
```
shane-huang, line 1: maybe we can provide a patch instead of the entire file. And in README we can tell user to download the auto_comment.py file (let's fix it to a specific revision) from QWen example, and then apply the patch, then provide the sample code and sample output for user to try.

PR #9295, opened 2023-10-27
Add test script and workflow for qlora fine-tuning

chtanch:
## Description
Add a test script to run qlora fine-tuning and merge adapter into original model. Add script to llm_unit_test workflow.

### DIFF
--- a/python/llm/test/run-llm-finetune-tests-gpu.sh
+++ b/python/llm/test/run-llm-finetune-tests-gpu.sh
@@ -7,14 +7,17 @@ set -e
 echo "# Start testing qlora fine-tuning"
 start=$(date "+%s")
 
+sed -i 's/max_steps=200/max_steps=2/; s/save_steps=100/save_steps=2/; s/logging_steps=20/logging_steps=1/' \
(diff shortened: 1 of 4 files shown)


python/llm/test/run-llm-finetune-tests-gpu.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
hkvision, line 1: rename to run-llm-example-tests-gpu.sh?

python/llm/test/run-llm-finetune-tests-gpu.sh
```diff
@@ -0,0 +1,24 @@
...
+
+export ANALYTICS_ZOO_ROOT=${ANALYTICS_ZOO_ROOT}
+
+set -e
+
+echo "# Start testing qlora fine-tuning"
+start=$(date "+%s")
+
```
hkvision, line 9: let's reduce the max steps

hkvision (approved):

python/llm/test/run-llm-finetune-tests-gpu.sh
chtanch: Reduced to two steps

PR #9299, opened 2023-10-30
LLM: improve response speed in multi-turn chat

plusbang:
## Description

In current implementation, chat history includes previous output, then 1st token latency will be quite large after multi turns.
Utilize `past_key_values` to solve this problem.
Especially, chatglm2 provide `stream_chat` native API to support `past_key_values`, so directly use it.

### 2. User API changes

N/A

### 4. How to test?
- [ ] Unit test
- [ ] Local test (Verification and collected data can be found in https://github.com/analytics-zoo/nano/issues/686)


shane-huang (approved):
Initial version LGTM. Will move some utils inside lib and clean up a bit later.

PR #9300, opened 2023-10-30
[LLM] Fix llm arc ut oom

chtanch:
## Description

https://github.com/intel-analytics/BigDL/actions/runs/6687348613/job/18167964198#step:10:417

Memory is not deallocated from GPU memory after a model's has finished. So tests with different models will use more and more memory until OOM

Moving model to cpu after it's test is done could be a workaround for this problem.

### How to test?
- [ ] Unit test

### DIFF
--- a/python/llm/test/inference_gpu/test_optimize_model.py
+++ b/python/llm/test/inference_gpu/test_optimize_model.py
@@ -41,7 +41,7 @@ def test_optimize_model(Model, Tokenizer, model_path):
                                 trust_remote_code=True)
     model = model.to(device)
     logits_base_model = (model(input_ids)).logits
-    model.to('cpu')
+    model.to('cpu')  # deallocate gpu memory
 
     model = Model.from_pretrained(model_path,
                                 load_in_4bit=True,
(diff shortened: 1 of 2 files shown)


Oscilloscope98 (approved):
LGTM

python/llm/test/inference_gpu/test_optimize_model.py
```diff
@@ -41,13 +41,15 @@ def test_optimize_model(Model, Tokenizer, model_path):
                                 trust_remote_code=True)
     model = model.to(device)
     logits_base_model = (model(input_ids)).logits
+    model.to('cpu')
```
Oscilloscope98, line 44: Could you add a short comment why we add the `model.to('cpu')` here?

chtanch: I think that this indicates to the driver that the model is no longer in use on the GPU and memory can be reallocated/deallocated.

Oscilloscope98: Could you add the comment in the code? :) So that it would be easier for others to get the idea.

PR #9310, opened 2023-10-31
Integrate vllm

gc-fu:
## Description

This PR aims to integrate vLLM into BigDL, which enables users to use vLLM on CPU platforms and use BigDL-LLM transformer-int4 models for LLM serving.

### High level ideas

The reason why we cannot use original `vLLM` on Intel Platforms is that it has some `CUDA` operations that are bound to its implementations.  These operations can be found at [here](https://github.com/vllm-project/vllm/tree/main/csrc).  Some of these operations including:

1. Page table [operations](https://github.com/vllm-project/vllm/blob/main/csrc/cache_kernels.cu), such as swap blocks etc. 
2. Model implementation operations, such as `layernorm`, `pos_encoding`, `activation`, etc.
3. Other operations, such as `quantization`.

To utilize `vLLM` on Intel platforms, we need to somehow remove these `CUDA` operations.

There are several ways to achieve this:
1. (current plan) We choose to delete the Page Table implementation in `vLLM` to avoid usage of `cache_kernels.cu` operations.  As CPU platform has support for virtual memory and tensor storage does not need to be placed at continuous physical memory, this will not be a hard problem.  However, without `PageTable` support, we would lack the ability to share KV_CACHE blocks.
For model operations that are bound to `CUDA` operations, we choose to use regular `transformers` models to replace the `vLLM` customized models. With this option, we need to maintain our own `KV_CACHE` and prepare the inputs to adapt the input format.
For quantized operations, we can utilize `bigdl-llm` for sym_int4 quantization methods.

2. (Alternative plan from other teams) We can choose to use the PageTable for CPU implementation. For CPU implementation, we can easily replace the swap_block operations.  As for the model operations, we can replace the `CUDA` implementation with native Torch operations. This require us to gradually change cuda implementations to torch implementations.  We have not yet carefully inspect this design, may need some POC if we want to use this.

We are currently choosing the first option.

### 1. Why the change?

To enable user to use the vLLM on CPU platform, and later on XPU probably.

### 2. User API changes

The main API faced by the users is this:

```
from bigdl.llm.vllm.engine.async_llm_engine import AsyncLLMEngine
```
Examples can be found at the examples folder.

### 3. Summary of the change 

The design is ported from vLLM, mainly to provide continuous batching in `cpu` environment.
Later, in `xpu` environment, finding a way to avoid padding.

### 4. TODO
- [x] Some of files can be deleted and use the official version from `vllm` to reduce code copy.
- [x] Provide install method description in readme.md
- [ ] Provide xpu-compatible codes (Maybe in other PR)
- [x] Add license into files
- [x] Fix some of the naming of the files.

### 5. New dependencies
This will not introduce any new dependencies, user should install the dependencies by themselves according to the `requirements.txt`


jason-dai:
Add a readme to describe how the user can use it

glorysdj:
design updated https://github.com/analytics-zoo/nano/issues/619#issuecomment-1813907200

python/llm/src/bigdl/llm/vllm/engine/async_llm_engine.py
```diff
@@ -0,0 +1,531 @@
...
+        self,
+        method: str,
+        *args,
+        get_all_outputs: bool=False,
+        **kwargs,
+    ) -> Any:
+        """Runs the given method on all workers."""
+        all_outputs = []
+        for worker in self.workers:
+            if self.parallel_config.worker_use_ray:
```
glorysdj, line 241: please remove ray logic

python/llm/example/CPU/vLLM-Serving/README.md
```diff
@@ -0,0 +1,75 @@
...
+This example demonstrates how to serving a llama2-7b model using BigDL-LLM 4 bits optimizations with xeon CPUs with adapted vLLM.
+
+
+## Example: Serving llama2-7b using Xeon CPU
+
+In this example, we will run Llama2-7b model using 48 cores in one socket and provide `OpenAI-compatible` interface for users.
+
+### 1. Install
+
+The original [vLLM](https://github.com/analytics-zoo/vllm) is designed to run with `CUDA` environment. To adapt vLLM into `Intel` platforms, install the dependencies like this:
```
glorysdj, line 12: wrong link? 
please declare that the vllm example and code are ported from https://github.com/vllm-project/vllm

python/llm/example/CPU/vLLM-Serving/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+ curl http://localhost:8000/v1/completions \
+         -H "Content-Type: application/json" \
+         -d '{
+                 "model": "/MODEL_PATH/Llama-2-7b-chat-hf-bigdl/",
+                 "prompt": "San Francisco is a",
+                 "max_tokens": 128,
+                 "temperature": 0
+ }' &
+```
```
glorysdj, line 75: add readme for how to support more models

python/llm/example/CPU/vLLM-Serving/README.md
gc-fu: Fixed

python/llm/example/CPU/vLLM-Serving/README.md
gc-fu: Added

glorysdj (approved):
LGTM

PR #9316, opened 2023-10-31
Use to test llm-performance on spr-perf

Romanticoseu:


liu-shaojun (approved):
LGTM

Romanticoseu:
The reason for the error reported by test-on-arc is that the model was accidentally deleted.

PR #9318, opened 2023-10-31
LLM: adjust dolly v2 GPU transformers example

JinBridger:
## Description

Adjust dolly v2 GPU transformers example README. Replace CPU settings with GPU settings.


plusbang (approved):
LGTM

PR #9322, opened 2023-10-31
add malloc and intel openmp to llm deps

### LINKED ISSUE #9304: bigdl-llm-init cannot reach the optimized performance as bigdl-nano-init
I used both methods to test Llama-2-7b-chat-hf in condition of 1k-128(beam=4) transformer-int4 api.
Next token latency using bigdl-nano-init is around 170ms, but 320ms using bigdl-llm-init.
I have made a comparison of env list between two circumstances.
https://github.com/Zhengjin-Wang/BigDL/commit/1f6d524f8967a9a020f46afe941e4fbaa6b19c01?diff=split
The left side is env list in bigdl-llm-init, and right side is in bigdl-nano-init.

qiuxin2012:
## Description

add malloc and intel openmp to llm deps

### 1. Why the change?

fixed #9304

### 2. User API changes

None

### DIFF
--- a/python/llm/setup.py
+++ b/python/llm/setup.py
@@ -126,6 +126,17 @@ linux_binarys = [
     "chatglm_C.cpython-39-x86_64-linux-gnu.so",
 ]
 
+ext_lib_urls = [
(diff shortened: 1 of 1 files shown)


MeouSker77 (approved):
LGTM

PR #9323, opened 2023-10-31
fix import ipex problem

yangw1234:
## Description

fix import ipex problem

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/models/llama.py
+++ b/python/llm/src/bigdl/llm/transformers/models/llama.py
@@ -60,6 +60,8 @@ KV_CACHE_ALLOC_BLOCK_LENGTH = 256
 
 
 _ipex_version = None
+
+
 def get_ipex_version():
 
     global _ipex_version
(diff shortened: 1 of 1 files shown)


Oscilloscope98 (approved):
LGTM

python/llm/src/bigdl/llm/transformers/models/llama.py
```diff
@@ -59,21 +59,23 @@ def repeat_kv(hidden_states: torch.Tensor, n_rep: int) -> torch.Tensor:
...
-ipex_version = get_ipex_version()
+    import intel_extension_for_pytorch as ipex
+    _ipex_version = ipex.__version__
+    return _ipex_version
 
 
 def llama_rms_norm_forward(self, hidden_states):
     if hidden_states.device.type == "xpu" and not (self.training and hidden_states.requires_grad):
-        if ipex_version == "2.0.110+xpu":
+        if get_ipex_version() == "2.0.110+xpu":
```
jason-dai, line 78: do we need to fix it for other models? @qiuxin2012

yangw1234: We only have this check in llama as an easy way to test internal IPEX on AutoTP. We might need this in other models when the new IPEX come out. (Or we can ask them to keep it backward compatible.)

PR #9335, opened 2023-11-02
Add a cpu example of  HuggingFace Transformers Agent - Audio Transcription 

yzheng124:
### 1. Why the change?

In this PR, we add a cpu example of HuggingFace Transformers Agent with audio transcription task. This example uses `vicuna-7b-v1.5` to create an agent and `whisper-tiny` to generate audio transcription. Specifically, `whisper-tiny` part is modified based on [whisper example](https://github.com/intel-analytics/BigDL/blob/main/python/llm/example/CPU/HF-Transformers-AutoModels/Model/whisper/recognize.py)

https://github.com/analytics-zoo/nano/issues/669


yzheng124:
Hi @shane-huang, please take a look. Thanks!

PR #9338, opened 2023-11-02
LLM: add CodeLlama CPU and GPU examples

JinBridger:
## Description

Add CodeLlama CPU and GPU examples, using `CodeLlama-7b-hf`.


python/llm/example/CPU/HF-Transformers-AutoModels/Model/codellama/README.md
```diff
@@ -0,0 +1,66 @@
+# CodeLlama
```
shane-huang, line 1: update BigDL/README.md and BigDL/llm/README.md

shane-huang (approved):
LGTM.

PR #9340, opened 2023-11-02
Add cpu examples of skywork

yzheng124:
### 1. Why the change?

In this PR,  we add cpu examples of Skywork using bigdl-llm's `transformer` and `optimize_model` API.

https://github.com/analytics-zoo/nano/issues/697


shane-huang (approved):
LGTM.

PR #9342, opened 2023-11-02
Fix the installation dependencies issues related to libsgx-quote-ex

Romanticoseu:
## Description

When build `bigdl-ppml-trusted-bigdl-llm-serving-tdx`, dependency issues will occur.
Refer to the p10 of [Intel_SGX_SW_Installation_Guide](https://download.01.org/intel-sgx/latest/linux-latest/docs/Intel_SGX_SW_Installation_Guide_for_Linux.pdf), we should use apt/sources.list.d for `Ubuntu 22.04`
image
image
image


gc-fu (approved):
LGTM

PR #9343, opened 2023-11-02
[LLM] Move embedding layer to CPU for iGPU inference

Oscilloscope98:
## Description

Moving embedding layer to CPU for LLM inference on iGPU.

### 1. Why the change?

Due to the large size of embedding layer, calling embedding layer on iGPU may cause some 7B models (Baichuan2, internlm, Aqulia, etc,) OOM on an 8GB iGPU. Besides, calling the embedding layer on iGPU will not theoretically bring performance improvement:  https://github.com/analytics-zoo/nano/issues/693#issuecomment-1788687024

### 2. User API changes

N/A


MeouSker77 (approved):
LGTM

PR #9347, opened 2023-11-03
[LLM] Add model correctness test on ARC for llama and falcon

sgwhat:
## Description
This PR (Pull Request) adds model correctness validation on the ARC GPU and compares the output of the last `self_attention` of the model before and after fused rope/norm optimization.

### 1. Why the change?
After optimizing the LLM model with `fused rope/norm`, the output on ARC may change. To validate the model's correctness, we need to compare the `self_attention` output layer by layer.

### 2. User API changes
N/A

### 3. Summary of the change 
- Add ut for llama model
- Remove fused norm when validating correctness

### 4. How to test?
- [x] Unit test

### DIFF
--- a/python/llm/test/inference_gpu/test_optimize_model.py
+++ b/python/llm/test/inference_gpu/test_optimize_model.py
@@ -160,7 +160,7 @@ class Test_Optimize_Gpu_Model:
         Tokenizer = AutoTokenizer
         model_path = os.environ.get('FALCON_7B_ORIGIN_PATH')
         # currently only compare the output of the last self-attention layer.
-        prev_attn = "transformer.h.31"
+        prev_attn = "transformer.h.30"
         self_attn = "transformer.h.31.self_attention"
         lower_bound = 0
 
--- a/python/llm/test/inference_gpu/test_optimize_model.py
+++ b/python/llm/test/inference_gpu/test_optimize_model.py
@@ -62,13 +62,12 @@ def test_optimize_model(Model, Tokenizer, model_path):
 
 class Test_Optimize_Gpu_Model:
     def setup(self):
-        self.lower_bound = 5e-2
-        
+
         self.layer_inputs = []
         self.layer_outputs = []
         self.pre_layer_outputs = []
 
-    def run_optimize_gpu_model(self, Model, Tokenizer, model_path, self_attn, prev_attn):
+    def run_optimize_gpu_model(self, Model, Tokenizer, model_path, self_attn, prev_attn, lower_bound):
         def forward_hook(module, input, output, layer_name):
             self.layer_outputs.append(output)
 
@@ -76,7 +75,7 @@ class Test_Optimize_Gpu_Model:
             self.pre_layer_outputs.append(output)
 
         def new_forward_hook(module, input):
-            if model_path is os.environ.get('LLAMA_ORIGIN_PATH'):
+            if model_path == os.environ.get('LLAMA_ORIGIN_PATH'):
                 repalcement_norm = model.model.norm
                 opt_model.model.norm = repalcement_norm
 
@@ -152,7 +151,7 @@ class Test_Optimize_Gpu_Model:
                         attn_output_diff.append(t3 - t4)
 
         max_diff_tensor = [torch.max(item).item() for item in attn_output_diff]
-        assert all(max_diff <= self.lower_bound for max_diff in max_diff_tensor)
+        assert all(max_diff <= lower_bound for max_diff in max_diff_tensor)
 
 
     def test_falcon_gpu_model(self):
@@ -163,8 +162,9 @@ class Test_Optimize_Gpu_Model:
         # currently only compare the output of the last self-attention layer.
         prev_attn = "transformer.h.31"
         self_attn = "transformer.h.31.self_attention"
+        lower_bound = 0
 
-        self.run_optimize_gpu_model(Model, Tokenizer, model_path, self_attn, prev_attn)
+        self.run_optimize_gpu_model(Model, Tokenizer, model_path, self_attn, prev_attn, lower_bound)
 
 
     def test_llama_gpu_model(self):
@@ -175,8 +175,9 @@ class Test_Optimize_Gpu_Model:
         # currently only compare the output of the last self-attention layer.
         prev_attn = "model.layers.30"
         self_attn = "model.layers.31.self_attn"
+        lower_bound = 5e-2
 
-        self.run_optimize_gpu_model(Model, Tokenizer, model_path, self_attn, prev_attn)
+        self.run_optimize_gpu_model(Model, Tokenizer, model_path, self_attn, prev_attn, lower_bound)
 
 
 if __name__ == '__main__':
--- a/python/llm/test/inference_gpu/test_optimize_model.py
+++ b/python/llm/test/inference_gpu/test_optimize_model.py
@@ -98,11 +98,8 @@ def test_optimize_llama_model(Model, Tokenizer, model_path):
             layer_module.register_forward_hook(
                 lambda module, output, layer_name=layer_name: forward_hook(module, 
                                                                            output, layer_name))
-
     logits_base_model = (model(input_ids)).logits
-
-    for layer_name, output_tensor in layer_outputs.items():
-        layer_tensor.append(output_tensor)
+    layer_tensor = layer_outputs.pop()
 
     del model
 
@@ -115,18 +112,15 @@ def test_optimize_llama_model(Model, Tokenizer, model_path):
     for layer_name, layer_module in opt_model.named_modules():
         if layer_name == "model.layers.31.self_attention":
             layer_module.register_forward_pre_hook(
-                lambda module, input: pre_hook(module, input))
+                lambda module, input: load_pre_hook(module, input))
             layer_module.register_forward_hook(
                 lambda module, output, layer_name=layer_name: forward_hook(module, 
                                                                            output, layer_name))
     logits_optimized_model = (opt_model(input_ids)).logits
-
-    for layer_name, output_tensor in layer_outputs.items():
-        opt_layer_tensor.append(output_tensor)
+    opt_layer_tensor = layer_outputs.pop()
 
     attn_output_diff = []
-    for i, (t1, t2) in enumerate(zip(layer_tensor[0], opt_layer_tensor[0])):
-        #if not torch.equal(t1, t2):
+    for i, (t1, t2) in enumerate(zip(layer_tensor, opt_layer_tensor)):
         if t1 is not None and t2 is not None:
             if not isinstance(t1, tuple) and not isinstance(t2, tuple):
                 attn_output_diff.append(t1 - t2)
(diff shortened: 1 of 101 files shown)


python/llm/test/inference_gpu/test_optimize_model.py
```diff
@@ -56,5 +80,115 @@ def test_optimize_model(Model, Tokenizer, model_path):
...
+                for i, (t3, t4) in enumerate(zip(t1, t2)):
+                    attn_output_diff.append(t3 - t4)
+
+    max_diff_tensor = [torch.max(item).item() for item in attn_output_diff]
+    assert all(max_diff <= lower_bound for max_diff in max_diff_tensor)
+
+
+@pytest.mark.parametrize('Model, Tokenizer, model_path',[
+    (AutoModelForCausalLM, AutoTokenizer, os.environ.get('FALCON_7B_ORIGIN_PATH'))
+])
```
hkvision, line 140: Can we use this to combine llama and falcon?
is the only diff in layer_name？

python/llm/test/inference_gpu/test_optimize_model.py
```diff
@@ -56,5 +80,115 @@ def test_optimize_model(Model, Tokenizer, model_path):
...
+
+    for layer_name, layer_module in model.named_modules():
+        if layer_name == "model.layers.31.self_attn":
+            layer_module.register_forward_pre_hook(
+                lambda module, input: pre_hook(module, input))
+            layer_module.register_forward_hook(
+                lambda module, output, layer_name=layer_name: forward_hook(module, 
+                                                                           output, layer_name))
+    logits_base_model = (model(input_ids)).logits
+    layer_tensor = layer_outputs.pop()
```
hkvision, line 105: why need to pop here?

python/llm/test/inference_gpu/test_optimize_model.py
```diff
@@ -56,5 +80,115 @@ def test_optimize_model(Model, Tokenizer, model_path):
...
+            layer_module.register_forward_hook(
+                lambda module, output, layer_name=layer_name: forward_hook(module, 
+                                                                           output, layer_name))
+    logits_optimized_model = (opt_model(input_ids)).logits
+    opt_layer_tensor = layer_outputs.pop()
+
+    attn_output_diff = []
+    for i, (t1, t2) in enumerate(zip(layer_tensor, opt_layer_tensor)):
+        if t1 is not None and t2 is not None:
+            if not isinstance(t1, tuple) and not isinstance(t2, tuple):
```
hkvision, line 128: will isinstance of torch.Tensor be clearer?

python/llm/test/inference_gpu/test_optimize_model.py
```diff
@@ -56,5 +80,115 @@ def test_optimize_model(Model, Tokenizer, model_path):
...
+                attn_output_diff.append(t1 - t2)
+            else:
+                for i, (t3, t4) in enumerate(zip(t1, t2)):
```
hkvision, line 131: under what cases it will be a tuple?

python/llm/test/inference_gpu/test_optimize_model.py
sgwhat: yes I am prepared to combine them in next commit.

python/llm/test/inference_gpu/test_optimize_model.py
sgwhat: `past_key_value` returned by attn_forward should always be a tuple.

python/llm/test/inference_gpu/test_optimize_model.py
hkvision: You need to add some doc here. It is a singleton right, you are taking the last and only item right?

sgwhat: yes, it is, will add a doc.

hkvision (approved):
Code change LGTM, need to add more docs, e.g. what is for pop, what case is a tuple, etc.

python/llm/test/inference_gpu/test_optimize_model.py
```diff
@@ -132,6 +136,7 @@ def test_falcon_gpu_model(self):
         Model = AutoModelForCausalLM
         Tokenizer = AutoTokenizer
         model_path = os.environ.get('FALCON_7B_ORIGIN_PATH')
+        # currently only need to compare the output of one self-attention layer.
```
hkvision, line 139: currently only compare the output of the last self-.... ?

python/llm/test/inference_gpu/test_optimize_model.py
```diff
@@ -56,5 +60,98 @@ def test_optimize_model(Model, Tokenizer, model_path):
...
+        for layer_name, layer_module in model.named_modules():
+            if layer_name == self_attn:
+                layer_module.register_forward_pre_hook(
+                    lambda module, input: pre_hook(module, input))
+                layer_module.register_forward_hook(
+                    lambda module, output, layer_name=layer_name: forward_hook(module, 
+                                                                               output, layer_name))
+        logits_base_model = (model(input_ids)).logits
+        # the list `layer_output` has only one element.
+        layer_tensor = self.layer_outputs.pop()
```
hkvision, line 98: use layer_outputs[0]

python/llm/test/inference_gpu/test_optimize_model.py
```diff
@@ -56,5 +60,124 @@ def test_optimize_model(Model, Tokenizer, model_path):
     assert any(diff) is False
 
 
+class Test_Optimize_Gpu_Model:
+    def setup(self):
+        self.lower_bound = 5e-2
```
hkvision, line 65: let's have each model with one value instead of all models sharing one?

python/llm/test/run-llm-inference-tests-gpu.sh
```diff
@@ -2,7 +2,7 @@
 
 export ANALYTICS_ZOO_ROOT=${ANALYTICS_ZOO_ROOT}
 export LLM_HOME=${ANALYTICS_ZOO_ROOT}/python/llm/src
-export LLM_INFERENCE_TEST_DIR=${ANALYTICS_ZOO_ROOT}/python/llm/test/inference_gpu
+export LLM_INFERENCE_TEST_DIR=${ANALYTICS_ZOO_ROOT}/python/llm/test/inference_gpu/test_optimize_model.py
```
hkvision, line 5: why change this?

python/llm/test/inference_gpu/test_optimize_model.py
```diff
@@ -27,10 +29,12 @@
 
 prompt = "Once upon a time, there existed a little girl who liked to have adventures. She wanted to go to places and meet new people, and have fun"
 
+"""
```
hkvision, line 32: after commenting out this, where is mpt?

python/llm/test/inference_gpu/test_optimize_model.py
```diff
@@ -54,6 +58,126 @@ def test_optimize_model(Model, Tokenizer, model_path):
...
+    def run_optimize_gpu_model(self, Model, Tokenizer, model_path, self_attn, prev_attn, lower_bound):
+        def forward_hook(module, input, output, layer_name):
+            self.layer_outputs.append(output)
+
+        def pre_forward_hook(module, input, output, layer_name):
+            self.pre_layer_outputs.append(output)
+
+        def new_forward_hook(module, input):
+            if model_path == os.environ.get('LLAMA2_7B_ORIGIN_PATH'):
+                repalcement_norm = model.model.norm
```
hkvision, line 79: typo replace

python/llm/test/inference_gpu/test_optimize_model.py
```diff
@@ -54,6 +58,126 @@ def test_optimize_model(Model, Tokenizer, model_path):
...
+                opt_model.model.norm = repalcement_norm
```
hkvision, line 80: I don't think it is a good practice to write code like this, since opt_model is not even defined here.

python/llm/test/inference_gpu/test_optimize_model.py
```diff
@@ -54,6 +58,126 @@ def test_optimize_model(Model, Tokenizer, model_path):
...
+
+        def pre_hook(module, input):
+            self.layer_inputs.append(input)
+
+        def replace_forward_hook(module, input, output, layer_name):
+            hidden_states, present_key_value = output
+            hidden_states = layer_states[0].detach()
```
hkvision, line 87: same as above, layer_states is not even defined at this point

python/llm/test/inference_gpu/test_optimize_model.py
```diff
@@ -54,6 +58,126 @@ def test_optimize_model(Model, Tokenizer, model_path):
...
+        def pre_forward_hook(module, input, output, layer_name):
+            self.pre_layer_outputs.append(output)
+
+        def new_forward_hook(module, input):
+            if model_path == os.environ.get('LLAMA2_7B_ORIGIN_PATH'):
+                repalcement_norm = model.model.norm
+                opt_model.model.norm = repalcement_norm
+
+        def pre_hook(module, input):
+            self.layer_inputs.append(input)
```
hkvision, line 83: where is it used?

python/llm/test/inference_gpu/test_optimize_model.py
```diff
@@ -54,6 +58,126 @@ def test_optimize_model(Model, Tokenizer, model_path):
...
+                                      optimize_model=False,
+                                      trust_remote_code=True)
+        model = model.to(device)
+
+        for layer_name, layer_module in model.named_modules():
+            if layer_name == prev_attn:
+                layer_module.register_forward_pre_hook(
+                    lambda module, input: pre_hook(module, input))
+                layer_module.register_forward_hook(
+                    lambda module, input, output, layer_name=layer_name: pre_forward_hook(module, input,
```
hkvision, line 105: what's the purpose/benefit of adding this?

python/llm/test/run-llm-inference-tests-gpu.sh
sgwhat: this commit is just to test this code for now and will revert changes after passed test.

python/llm/test/inference_gpu/test_optimize_model.py
sgwhat: it's a hook and will only be executed with forward, when it was executed, there already has a list like layer_states definded.

hkvision: I know it can work, but it is awkward to write like this. In theory, the hook should be independent of what happens outside the hook?

python/llm/test/inference_gpu/test_optimize_model.py
sgwhat: to collect `hidden_states`, which is one of the output of DecoderLayer, this could be treat as the input for the next self_attn layer.

python/llm/test/inference_gpu/test_optimize_model.py
```diff
@@ -29,7 +31,7 @@
 
 @pytest.mark.parametrize('Model, Tokenizer, model_path',[
     (AutoModelForCausalLM, AutoTokenizer, os.environ.get('MPT_7B_ORIGIN_PATH')),
-    (AutoModelForCausalLM, AutoTokenizer, os.environ.get('FALCON_7B_ORIGIN_PATH')),
+    (AutoModelForCausalLM, AutoTokenizer, os.environ.get('FALCON_7B_ORIGIN_PATH'))
```
hkvision, line 34: Do you need to remove this since you add falcon in the new place?

python/llm/test/inference_gpu/test_optimize_model.py
sgwhat: this should be `pop` to clear the list, or there will be multi values in next forward.

PR #9349, opened 2023-11-03
LLM: add more models and skip runtime error

WeiguangHan:
@hkvision Add more models to the arc performance test and skip runtime errors when the `in_out_pairs` is not applicable to some models.


python/llm/dev/benchmark/all-in-one/run.py
```diff
@@ -57,7 +57,7 @@ def run_model(repo_id, test_api, in_out_pairs, local_model_hub=None, warm_up=1,
         result = run_ipex_fp16_gpu(repo_id, local_model_hub, in_out_pairs, warm_up, num_trials, num_beams)
 
     for in_out_pair in in_out_pairs:
-        if result:
+        if result and result[in_out_pair]:
```
hkvision, line 60: is this correct? shouldn't it be `in_out_pair in result`?

WeiguangHan: Yes. The skip operation happened after `result[in_out] = []`.

.github/workflows/llm_performance_tests.yml
```diff
@@ -108,6 +108,7 @@ jobs:
           python -m pip install --upgrade einops
           python -m pip install --upgrade transformers_stream_generator
           python -m pip install --upgrade tiktoken
+          python -m pip install --upgrade transformers
```
hkvision, line 111: specify version ==4.34.0? maybe in a next PR. also add a comment 4.34.0 is for mistral

WeiguangHan: ok

hkvision (approved):

hkvision:
Arc perf test passed, merge it first.

PR #9365, opened 2023-11-07
[LLM] Use fp32 as dtype when batch_size <=8 and qtype is q4_0/q8_0/fp8

MeouSker77:
## Description

We can use XMX kernel when batch_size <=8 and qtype is q4_0/q8_0/fp8, fp16 is slower than fp32 in this case

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/low_bit_linear.py
+++ b/python/llm/src/bigdl/llm/transformers/low_bit_linear.py
@@ -441,7 +441,7 @@ class LowBitLinear(nn.Linear):
                 # sometimes fp16 cause nan and training instability
                 # disable the conversion when training
                 if self.conver_to_half and x_2d.shape[0] > 1 and x_2d.dtype == torch.float32 and \
-                    (x_2d.shape[0] > 8 or self.qtype not in [SYM_INT4, SYM_INT8, FP8]):
+                        (x_2d.shape[0] > 8 or self.qtype not in [SYM_INT4, SYM_INT8, FP8]):
                     x_2d = x_2d.half()
                     result = linear_q4_0.forward_new(x_2d, self.weight.data, self.weight.qtype,
                                                      input_seq_size)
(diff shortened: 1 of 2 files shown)


rnwang04 (approved):

python/llm/src/bigdl/llm/transformers/low_bit_linear.py
```diff
@@ -440,7 +440,8 @@ def forward(self, x: torch.Tensor):
                 # current workaround to reduce first token latency of fp32 input
                 # sometimes fp16 cause nan and training instability
                 # disable the conversion when training
-                if self.conver_to_half and x_2d.shape[0] > 1 and x_2d.dtype == torch.float32:
+                if self.conver_to_half and x_2d.shape[0] > 1 and x_2d.dtype == torch.float32 and \
```
jason-dai, line 443: `x_2d.shape[0] > 1` no longer need?

python/llm/src/bigdl/llm/transformers/low_bit_linear.py
```diff
@@ -440,7 +440,8 @@ def forward(self, x: torch.Tensor):
...
+                        (x_2d.shape[0] > 8 or self.qtype not in [SYM_INT4, SYM_INT8, FP8]):
```
jason-dai, line 444: Add a function (e.g., `using_xmx`) to check the conditions (e.g., GPU type, batch size, data types, etc.)

python/llm/src/bigdl/llm/transformers/low_bit_linear.py
MeouSker77: still needed. if we remove it, then other qtype such as `SYM_FP4` with `x_2d.shape[0] = 1` will use fp16, but all `x_2d.shape[0]=1` should use fp32

python/llm/src/bigdl/llm/transformers/low_bit_linear.py
MeouSker77: done

MeouSker77:
But I suddenly remember MTL's igpu is also recognized as Arc(TM), maybe we can't use device name to check it

PR #9372, opened 2023-11-07
LLM: optimize QLoRA by updating lora convert logic

rnwang04:
### 1. Why the change?

https://github.com/analytics-zoo/nano/issues/713#issuecomment-1798203085

### 2. User API changes

No change.

### 3. Summary of the change 

update convert logic of lora part to reduce convert time.

### 4. How to test?
- [x] Unit test
- [x] Local test

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/utils.py
+++ b/python/llm/src/bigdl/llm/transformers/utils.py
@@ -135,21 +135,17 @@ def fix_key(key):
     return key
 
 
-def check_autocast(x):
-    if x.device.type == "xpu":
-        return torch.xpu.is_autocast_xpu_enabled()
-    elif x.device.type == "cpu":
-        return torch.is_autocast_enabled()
-    else:
-        invalidInputError(False,
-                          f"Device {x.device} is not supported.")
-
-
 def get_autocast_dtype(x):
     if x.device.type == "xpu":
-        return torch.xpu.get_autocast_xpu_dtype()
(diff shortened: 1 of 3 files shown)


rnwang04:
@jason-dai  As you said, putting the autocast at the first of LoraLowBitLinear does further improve the performance.
I have refactor related code, would you mind reviewing it again?

python/llm/src/bigdl/llm/transformers/utils.py
```diff
@@ -133,3 +133,23 @@ def fix_key(key):
...
+
+
+def get_autocast_dtype(x):
+    if x.device.type == "xpu":
+        return torch.xpu.get_autocast_xpu_dtype()
+    elif x.device.type == "cpu":
+        return torch.get_autocast_cpu_dtype()
+    else:
+        invalidInputError(False,
+                          f"Device {x.device} is not supported.")
```
jason-dai, line 155: We can combine these functions into to one (e.g., `get_autocast_dtype()` returning `None` if not enabled) to reduce one conditional branch if needed

jason-dai (approved):
LGTM

python/llm/src/bigdl/llm/transformers/utils.py
rnwang04: Have revised to this way.

rnwang04:
@hzjane I will merge this PR first, please help to check whether this PR will effect CPU QLoRA tmr : )

PR #9390, opened 2023-11-09
Stream llm example for both GPU and CPU

songhappy:
## Description

Merge code of both GPU and CPU example of stream-llm together into one piece of code.

### 1. Why the change?

Less code to show the application

### 2. User API changes

None

### 3. Summary of the change 

need to find a place to put this example

### 4. How to test?
-  Run examples

### 5. New dependencies
update transformers to 4.34.0


python/llm/example/CPU/Applications/streaming-llm/run_streaming_llama.py
```diff
@@ -45,6 +45,7 @@
 import torch
```
jason-dai, line 45: this folder is for CPU example; GPU example should go to `python/llm/example/GPU`

shane-huang:
How about adding streaming llm directly into bigdl-llm instead of making it another example? It can benefit other applications /examples.

jason-dai:
> How about adding streaming llm directly into bigdl-llm instead of making it another example? It can benefit other applications /examples.

Let's do that in a separate PR (with design reviewed by the team first)

python/llm/example/CPU/Applications/streaming-llm/run_streaming_llama.py
songhappy: refactored

python/llm/example/GPU/Applications/streaming-llm/run_streaming_llama.py
```diff
@@ -0,0 +1,165 @@
...
+# SOFTWARE.
+
+import warnings
+import torch
+import argparse
+import os
+from transformers.utils import is_torch_xpu_available
+import os, sys
+CURRENT_DIR = os.path.dirname(os.path.abspath(__file__))
+stream_llm_src = CURRENT_DIR.replace("GPU", "CPU") + "/streaming_llm/"
```
shane-huang, line 51: If we change the folder structure in future, this line could break. 

If you didn't make many changes in the original stream-llm, can we just in tell user to clone the original stream-llm repo and replace some code with ours? We did something similar in the [falcon example ](https://github.com/intel-analytics/BigDL/tree/main/python/llm/example/CPU/HF-Transformers-AutoModels/Model/falcon#22-replace-modelling_rwpy)

jason-dai: I prefer to make the example self-contained (without external dependency)

jason-dai: Just make a separate copy for CPU and GPU examples, instead of sharing files

songhappy:
performance metrics are updated in issues
https://github.com/analytics-zoo/nano/issues/671
https://github.com/analytics-zoo/nano/issues/754

jason-dai (approved):

PR #9402, opened 2023-11-09
Baichuan7b leaderboard run

shaline-koh:


jason-dai:
Maybe we should use Baichuan2 instead?

PR #9421, opened 2023-11-10
Add cpu and gpu examples for Yi-6B

yzheng124:
### 1. Why the change?

In this PR, we add cpu and gpu examples of Yi-6B using bigdl-llm's `transformer` and `optimize_model` API.

https://github.com/analytics-zoo/nano/issues/729


shane-huang (approved):
LGTM.

PR #9437, opened 2023-11-13
release assembly-all

liu-shaojun:
release assembly-all without bigdl-core and zoo-core

https://github.com/intel-analytics/BigDL/pull/9399


gc-fu (approved):
LGTM

PR #9445, opened 2023-11-13
LLM: modify the script to generate html results more accurately

WeiguangHan:
@hkvision Modify the the script to generate html results of arc perf test more accurately.

### DIFF
--- a/python/llm/test/benchmark/csv_to_html.py
+++ b/python/llm/test/benchmark/csv_to_html.py
@@ -39,18 +39,54 @@ def main():
     if len(csv_files)>1:
         data2 = pd.read_csv(csv_files[1], index_col=0)
 
+        data1_df=pd.DataFrame(data1)
+        data2_df=pd.DataFrame(data2)
+
+        last1=[]
+        diff1=[]
+        last2=[]
+        diff2=[]
+
         origin_column_1='1st token avg latency (ms)'
         origin_column_2='2+ avg latency (ms/token)'
+
+        for data1_ind,data1_row in data1_df.iterrows():
+            data1_model=data1_row['model'].strip()
+            data1_input_output_pairs=data1_row['input/output tokens'].strip()
+            flag=True
+
+            for data2_ind,data2_row in data2_df.iterrows():
+                data2_model=data2_row['model'].strip()
+                data2_input_output_pairs=data2_row['input/output tokens'].strip()
+                if data1_model==data2_model and data1_input_output_pairs==data2_input_output_pairs:
+                    data1_temp=data1_row[origin_column_1]
+                    data2_temp=data2_row[origin_column_1]
+                    data3_temp=data1_row[origin_column_2]
+                    data4_temp=data2_row[origin_column_2]
+
+                    last1.append(data2_temp)
+                    diff1.append(round((data2_temp-data1_temp)*100/data2_temp,2))
+                    last2.append(data4_temp)
+                    diff2.append(round((data4_temp-data3_temp)*100/data3_temp,2))
+
+                    flag=False
+                    break
+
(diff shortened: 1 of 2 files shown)


python/llm/test/benchmark/csv_to_html.py
```diff
@@ -39,18 +39,54 @@ def main():
     if len(csv_files)>1:
         data2 = pd.read_csv(csv_files[1], index_col=0)
 
+        data1_df=pd.DataFrame(data1)
+        data2_df=pd.DataFrame(data2)
```
hkvision, line 43: do we need this? they are already dataframes, no?

python/llm/test/benchmark/csv_to_html.py
```diff
@@ -39,18 +39,54 @@ def main():
...
+            flag=True
+
+            for data2_ind,data2_row in data2_df.iterrows():
+                data2_model=data2_row['model'].strip()
+                data2_input_output_pairs=data2_row['input/output tokens'].strip()
+                if data1_model==data2_model and data1_input_output_pairs==data2_input_output_pairs:
+                    data1_temp=data1_row[origin_column_1]
+                    data2_temp=data2_row[origin_column_1]
+                    data3_temp=data1_row[origin_column_2]
+                    data4_temp=data2_row[origin_column_2]
```
hkvision, line 65: make the variable names more meaningful.

python/llm/test/benchmark/csv_to_html.py
```diff
@@ -39,18 +39,54 @@ def main():
...
+
+                    last1.append(data2_temp)
+                    diff1.append(round((data2_temp-data1_temp)*100/data2_temp,2))
+                    last2.append(data4_temp)
+                    diff2.append(round((data4_temp-data3_temp)*100/data3_temp,2))
+
+                    flag=False
+                    break
+
+            if flag:
```
hkvision, line 75: I think we don't need this flag, we can just init diff=''?

python/llm/test/benchmark/csv_to_html.py
WeiguangHan: Yes, will remove this step.

python/llm/test/benchmark/csv_to_html.py
WeiguangHan: Sure

python/llm/test/benchmark/csv_to_html.py
WeiguangHan: OK. I will try according to the advice you gave me.

python/llm/test/benchmark/csv_to_html.py
```diff
@@ -34,26 +34,52 @@ def main():
...
+                    previous_2_avg_latency=previous_csv_row[latency_2_avg]
+                    last1[latest_csv_ind]=previous_1st_token_latency
+                    diff1[latest_csv_ind]=round((previous_1st_token_latency-latest_1st_token_latency)*100/previous_1st_token_latency,2)
+                    last2[latest_csv_ind]=previous_2_avg_latency
+                    diff2[latest_csv_ind]=round((previous_2_avg_latency-latest_2_avg_latency)*100/previous_2_avg_latency,2)
+
+        added_column_last1='last1'
+        added_column_diff1='diff1(%)'
+        added_column_last2='last2'
+        added_column_diff2='diff2(%)'
```
hkvision, line 74: just put these variables in the following lines instead of defining them separately?

WeiguangHan: ok. Will do it in next commit.

hkvision (approved):

PR #9391, opened 2023-11-09
Support directly loading gptq models from huggingface

yangw1234:
## Description

Support directly loading gptq models from huggingface.

Many models are published as gptq format in huggingface. It would be nice to load them directly using bigdl-llm.

Install:

```bash
BUILD_CUDA_EXT=0 pip install git+https://github.com/PanQiWei/AutoGPTQ.git@1de9ab6
pip install optimum==0.14.0
```

Usage:
```python
from bigdl.llm.transformers import AutoModelForCausalLM
from transformers import GPTQConfig
quantization_config = GPTQConfig(
    bits=4,
    use_exllama=False,
    )

# Load model in 4 bit,
# which convert the relevant layers in the model into INT4 format
model = AutoModelForCausalLM.from_pretrained(model_path,
                                             load_in_4bit=True, # will load into asym_int4. if using load_in_low_bit, then `load_in_low_bit` must be  "asym_int4"
                                             torch_dtype=torch.float,
                                             trust_remote_code=True,
                                             quantization_config=quantization_config,)
```

Limitations:

Only works on 4bit and asc_order=False
GPU version is really slow. Investigation needed.

Perf: https://github.com/analytics-zoo/nano/issues/738

PR #9453, opened 2023-11-14
Add awq load support

cyita:
## Description

Support AWQ model load from huggingface.
based on pr #9391.

### 1. Why the change?

Support AWQ model load from huggingface.

### 2. User API changes

```python
model = AutoModelForCausalLM.from_pretrained(model_path,
                                                 load_in_4bit=True,
                                                 trust_remote_code=True)
```

### 3. Summary of the change 

Support AWQ model load from huggingface.

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

### 5. New dependencies
```
pip install autoawq==0.1.6 --no-deps
pip install transformers==4.35.0
pip install accelerate==0.24.1
```


python/llm/src/bigdl/llm/transformers/convert.py
```diff
@@ -61,18 +65,24 @@ def is_deepspeed_available():
...
     if is_auto_gptq_available() and isinstance(module, QuantLinearCudaOld):
         in_features = module.infeatures
         out_features = module.outfeatures
         mp_group = None
         result = True
-    elif isinstance(module, nn.Linear):
+    elif isinstance(module, nn.Linear) or is_awq:
         in_features = module.in_features
         out_features = module.out_features
         mp_group = None
```
jason-dai, line 88: Combine above into below?
```python
if isinstance(module, nn.Linear) or is_gptq or is_awq
```

python/llm/src/bigdl/llm/transformers/awq/act.py
```diff
@@ -0,0 +1,40 @@
...
+# https://github.com/casper-hansen/AutoAWQ/tree/main
+#
+# @article{lin2023awq,
+#   title={AWQ: Activation-aware Weight Quantization for LLM Compression and Acceleration},
+#   author={Lin, Ji and Tang, Jiaming and Tang, Haotian and Yang, Shang and Dang,
+#          Xingyu and Han, Song},
+#   journal={arXiv},
+#   year={2023}
+# }
+#
```
jason-dai, line 28: 1) Adapted or copied?
2) Give details file url
3) Add the original MIT license

python/llm/src/bigdl/llm/transformers/model.py
```diff
@@ -156,16 +179,61 @@ def load_convert(cls, q_k, optimize_model, *args, **kwargs):
...
-            model = cls.HF_Model.from_pretrained(*args, **kwargs)
-        except NotImplementedError:
-            logger.info("Failed to load models with `low_cpu_mem_usage` specified, "
-                        "will fall to traditional load method with higher memory consumption.")
-            _kwargs["low_cpu_mem_usage"] = False
-            model = cls.HF_Model.from_pretrained(*_args, **_kwargs)
-            model.config.update({"bigdl_lcmu_enabled": False})
+        awq_config = None
+
+        if quant_config and quant_config.quant_method == "awq":
```
jason-dai, line 187: is the below code copied/adapted from `transformers`? if so, add comments to show that

python/llm/src/bigdl/llm/transformers/convert.py
cyita: The names of in/out features in the gptq QuantLinearCudaOld are different from linear and awq, so we cannot combine them here.

python/llm/src/bigdl/llm/transformers/model.py
cyita: It's copied from autoawq, updated.

python/llm/src/bigdl/llm/transformers/awq/act.py
cyita: Updated.

jason-dai (approved):
LGTM

BTW, how do we test the correctness?

cyita:
> LGTM
> 
> BTW, how do we test the correctness?

Currently, I determine if the output text makes sense to determine if it is correct.
For loading awq models part, we can compare the model loaded by huggingface transformers with the one loaded using our code.
For the inference part, it's hard to test since the awq forward code is written in cuda.

cyita:
Auto-AWQ added a script to evaluate the perplexity (https://github.com/casper-hansen/AutoAWQ/blob/main/awq/utils/eval_utils.py) last week, maybe we can use this script to do the evaluation.

PR #9454, opened 2023-11-14
LLM: update qlora example

rnwang04:
### 1. Why the change?

Update QLoRA examples.

### 2. User API changes

No change

### 3. Summary of the change 

Update QLoRA examples by

- add missing dependency `datasets`
- disable default `gradient checkpointing` to speedup training
- solve loss decreasing to zero issue by changing `fp16=True` to `bf16=True` and decreasing lr

### 4. How to test?
- [x] Unit test


MeouSker77 (approved):
LGTM

PR #9456, opened 2023-11-14
set bigdl-core/zoo-core version to 2.4.0-SNAPSHOT

liu-shaojun:
We will not release bigdl-core and zoo-core for 2.4.0 and 2.5.0, we keep using 2.4.0-SNAPSHOT for bigdl-core and zoo-core.


glorysdj (approved):
LGTM

PR #9469, opened 2023-11-15
LLM : Add CPU alpaca qlora example

hzjane:
## Description
Add CPU alpaca qlora example

### 2. User API changes
Add alpaca qlora example.

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [x] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/python/llm/example/CPU/QLoRA-FineTuning/alpaca-qlora/alpaca_qlora_finetuning_cpu.py
+++ b/python/llm/example/CPU/QLoRA-FineTuning/alpaca-qlora/alpaca_qlora_finetuning_cpu.py
@@ -1,0 +1,336 @@
+#
+# Copyright 2016 The BigDL Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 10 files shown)


python/llm/example/CPU/QLoRA-FineTuning/alpaca-qlora/alpaca_qlora_finetuning_cpu.py
```diff
@@ -0,0 +1,330 @@
+#
```
Uxito-Ada, line 1: Can `alpaca_qlora_finetuning_cpu.py` and `qlora_finetuning_cpu.py` be combined? It seems they differ only in data-processing logic.
For example, maybe we can define a `datasetTokenizer` base class, use a unified map API in `qlora_finetuning_cpu.py`, and just reimplement the specific logic under the alpaca folder.
In this way, `qlora_finetuning_cpu.py` can also be ported to more datasets in the future.

python/llm/example/CPU/QLoRA-FineTuning/alpaca-qlora/utils/prompter.py
```diff
@@ -0,0 +1,81 @@
...
+
+class Prompter(object):
+    __slots__ = ("template", "_verbose")
+
+    def __init__(self, template_name: str = "", verbose: bool = False):
+        self._verbose = verbose
+        if not template_name:
+            # Enforce the default here, so the constructor can be called with '' and will not break.
+            template_name = "alpaca"
+        file_name = osp.join("templates", f"{template_name}.json")
```
Uxito-Ada, line 48: can this be changed to a absolute path? Therefore, the python code can start anywhere instead of must be under the alpaca-qlora.
For example, I want to start the application at k8s mpi console, which is at mpiuser home rather than under the alpaca-qlora.

Uxito-Ada (approved):
lgtm

PR #9477, opened 2023-11-16
[LLM] Empty cache after embedding moved to CPU

Oscilloscope98:
## Description

Empty cache after embedding moved to CPU for iGPU.

However, this may lead some native effect to the first token latency

### 1. Why the change?

https://github.com/analytics-zoo/nano/issues/693#issuecomment-1797944763


MeouSker77 (approved):
LGTM

PR #9479, opened 2023-11-16
enable new q4_1 layout

qiuxin2012:
## Description

enable new q4_1 layout

### 1. Why the change?

performance optimization on q4_1

### 2. User API changes

None


cyita (approved):
LGTM

PR #9482, opened 2023-11-16
Fix bug for embed function

Kailuo-Lai:
## Description

In the source code, the **"encode_kwargs"** parameter doesn't play the role it should play, resulting in that user cannot control the tokenizer encoding process.

According to the source code, I found that the reason for this error is that the parameters are not passed into the "embed" function normally.

## Summary of the change 

I changed four lines of code to correct the logic of function argument passing, so that the ** "encode_kwargs" ** argument does what it should.

### DIFF
--- a/python/llm/src/bigdl/llm/langchain/embeddings/transformersembeddings.py
+++ b/python/llm/src/bigdl/llm/langchain/embeddings/transformersembeddings.py
@@ -135,7 +135,7 @@ class TransformersEmbeddings(BaseModel, Embeddings):
 
         extra = Extra.forbid
     
-    def embed(self, text: str):
+    def embed(self, text: str, encode_kwargs):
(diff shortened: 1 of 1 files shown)


jason-dai:
@plusbang please take a look

python/llm/src/bigdl/llm/langchain/embeddings/transformersembeddings.py
```diff
@@ -135,7 +135,7 @@ class Config:
 
         extra = Extra.forbid
     
-    def embed(self, text: str):
+    def embed(self, text: str, encode_kwargs):
```
plusbang, line 138: `encode_kwargs` -> `**encode_kwargs`

plusbang (commented):
Hi, @Kailuo-Lai , thanks a lot for your work to fix `embed` error! But please update according to comment :）

PR #9492, opened 2023-11-20
LLM: update Alpaca QLoRA finetuning example on GPU

plusbang:
## Description

- Add test scripts on PVC 1550
- Use `mpirun` instead of `torchrun`

### 4. How to test?
- [ ] Unit test
- [x] Local test (arc, flex, pvc are verified)


python/llm/example/GPU/QLoRA-FineTuning/alpaca-qlora/finetune_llama2_7b_flex_170_3_card.sh
```diff
@@ -14,17 +14,16 @@
...
-         --micro_batch_size 2 \
-         --batch_size 128
+mpirun -n 3 \
+       python -u ./alpaca_qlora_finetuning.py \
+       --base_model "meta-llama/Llama-2-7b-hf" \
+       --data_path "yahma/alpaca-cleaned" \
+       --output_dir "./bigdl-qlora-alpaca" \
+       --gradient_checkpointing False \
+       --micro_batch_size 2 \
+       --batch_size 128 > training.log
```
rnwang04, line 29: why some sh has `> training.log` and some not?

plusbang: > why some sh has `> training.log` and some not?

During my local experiment, directly bash scripts using mpirun, loss message is not printed in terminal, so add `> training.log` additionally to save loss message.

rnwang04 (approved):
LGTM

PR #9496, opened 2023-11-20
Update all in one benchmark readme

chtanch:
## Description

In the all-in-one performance benchmark, the script run-max-gpu.sh, preloads libtcmalloc.so
```
export LD_PRELOAD=${LD_PRELOAD}:${CONDA_PREFIX}/lib/libtcmalloc.so
```
libtcmalloc.so is not part of BigDL-llm install. Without libtcmalloc.so, the script still runs successfully, but with lower performance in some test cases.

This PR adds instructions for installing gperftools package (which contains libtcmalloc.so) to all-in-one's readme.

### DIFF
--- a/python/llm/dev/benchmark/all-in-one/README.md
+++ b/python/llm/dev/benchmark/all-in-one/README.md
@@ -9,7 +9,7 @@ pip install omegaconf
 pip install pandas
 ```
 
-Script for MAX GPU uses libtcmalloc.so from gperftools package for potentially better performance.
+Install gperftools to use libtcmalloc.so for MAX GPU to get better performance:
 ```bash
 conda install -c conda-forge -y gperftools=2.10
 ```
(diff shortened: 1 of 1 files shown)


python/llm/dev/benchmark/all-in-one/README.md
```diff
@@ -9,6 +9,11 @@ pip install omegaconf
 pip install pandas
 ```
 
+Script for MAX GPU uses libtcmalloc.so from gperftools package for potentially better performance.
```
hkvision, line 12: Install gperftools to use libtcmalloc.so for MAX GPU to get better performance:

hkvision:
@rnwang04 Take a look at this PR?

rnwang04 (approved):
LGTM

hkvision (approved):

PR #9500, opened 2023-11-21
Add vllm_worker related arguments in docker serving image's entrypoint

gc-fu:
## Description

As specified in title.

### DIFF
--- a/docker/llm/serving/cpu/docker/entrypoint.sh
+++ b/docker/llm/serving/cpu/docker/entrypoint.sh
@@ -102,7 +102,7 @@ if [ "$#" == 0 ]; then
   exec /usr/bin/tini -s -- "bash"
 else
   # Parse command-line options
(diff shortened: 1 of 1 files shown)


docker/llm/serving/cpu/docker/entrypoint.sh
```diff
@@ -101,7 +102,7 @@ if [ "$#" == 0 ]; then
   exec /usr/bin/tini -s -- "bash"
 else
   # Parse command-line options
-  options=$(getopt -o "m:h" --long "mode:,help" -n "$0" -- "$@")
+  options=$(getopt -o "m:hw:" --long "mode:,help" -n "$0" -- "$@")
```
xiangyuT, line 105: Does "worker" need to be set in --long flag here?

gc-fu: fixed

xiangyuT (approved):
LGTM

PR #9505, opened 2023-11-21
LLM: enable previous models

WeiguangHan:
@hkvision Test and enable previous models


.github/workflows/llm_performance_tests.yml
```diff
@@ -84,6 +84,42 @@ jobs:
...
+          current_date=$(date +%Y-%m-%d)
+          for file in *.csv; do
+              if [ -e "$file" ] && [[ "$file" == *"$current_date"* ]]; then
+                  filename="${file%.*}"
+                  extension="${file##*.}"
+
+                  new_filename="${filename}_test1.${extension}"
+                  mv "$file" "$new_filename"
+              fi
+          done
```
hkvision, line 96: just use sed to change the file name in run.py

.github/workflows/llm_performance_tests.yml
```diff
@@ -84,6 +84,42 @@ jobs:
...
+
+      - name: switch transformers version
```
hkvision, line 98: may not need a different job, just upgrade transformers after you run once and re-run.

python/llm/test/benchmark/arc-perf-test2.yaml
```diff
@@ -0,0 +1,14 @@
+# for the models which required the version of transformers 4.34.0
```
hkvision, line 1: rename to arc-perf-transformers-434.yaml?

.github/workflows/llm_performance_tests.yml
WeiguangHan: ok

.github/workflows/llm_performance_tests.yml
WeiguangHan: ok

python/llm/test/benchmark/arc-perf-test2.yaml
WeiguangHan: ok

python/llm/test/benchmark/arc-perf-transformers-434.yaml
```diff
@@ -0,0 +1,14 @@
+# for the models which required the version of transformers 4.34.0
```
hkvision, line 1: For the models that require transformers 4.34.0

hkvision (approved):

PR #9518, opened 2023-11-23
LLM: add flash attention support

rnwang04:
### 1. Why the change?

https://github.com/analytics-zoo/nano/issues/781
Support flash attention to accelerate first token.
This PR is a starting point, and just change llama. Will update more models later.
Related test llama perf is updated here: https://github.com/analytics-zoo/nano/issues/781#issuecomment-1824067726.

### 2. User API changes

No change.

### 3. Summary of the change 

- Support flash attention for llama's first token when it is avalible. It will accelerate long sequence sdp.
- Add more type conversion to make flash attention work for fp32 dtype.

### 4. How to test?
- [x] Unit test
- [x] Local test for llama2-7b (fp32, fp16) on max 1100

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/models/llama.py
+++ b/python/llm/src/bigdl/llm/transformers/models/llama.py
@@ -105,7 +105,9 @@ def llama_attention_forward_4_31(
     **kwargs,
 ) -> Tuple[torch.Tensor, Optional[torch.Tensor], Optional[Tuple[torch.Tensor]]]:
     bsz, q_len, _ = hidden_states.size()
+    original_dtype = hidden_states.dtype()
     device = hidden_states.device
+    fsdp_flag = check_flash_attention_available(hidden_states)
 
     if self.config.pretraining_tp > 1:
         key_value_slicing = (self.num_key_value_heads * self.head_dim) // self.config.pretraining_tp
@@ -127,6 +129,9 @@ def llama_attention_forward_4_31(
         value_states = torch.cat(value_states, dim=-1)
 
     else:
+        # cast hidden_states into fp16 in advance to make use of flash attention optimization
+        if fsdp_flag and q_len > 1:
+            hidden_states = hidden_states.half()
         query_states = self.q_proj(hidden_states)
         key_states = self.k_proj(hidden_states)
         value_states = self.v_proj(hidden_states)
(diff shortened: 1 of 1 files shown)


python/llm/src/bigdl/llm/transformers/models/llama.py
```diff
@@ -194,31 +204,23 @@ def llama_attention_forward_4_31(
...
-    attn_weights = nn.functional.softmax(attn_weights, dim=-1,
-                                         dtype=torch.float32).to(query_states.dtype)
-    attn_output = torch.matmul(attn_weights, value_states)
+                                                                         dtype=attention_dtype)
+
+    if fsdp_flag and q_len > 1:
+        # now only use flash attention for first token
+        attn_output = F.scaled_dot_product_attention(query_states.to(dtype=attention_dtype),
+                                                     key_states,
+                                                     value_states,
```
jason-dai, line 215: What's the type of key_states & value_states? Does it matter?

rnwang04: in this path, it's fp16 here.

rnwang04: All Tensor input of `F.scaled_dot_product_attention` must be fp16, then we can obtain acceleration effect from flash attention.

python/llm/src/bigdl/llm/transformers/models/llama.py
```diff
@@ -106,6 +106,16 @@ def llama_attention_forward_4_31(
 ) -> Tuple[torch.Tensor, Optional[torch.Tensor], Optional[Tuple[torch.Tensor]]]:
     bsz, q_len, _ = hidden_states.size()
     device = hidden_states.device
+    # for flash attention
+    original_dtype = hidden_states.dtype
+    if not self.training and not hidden_states.requires_grad:
+        fsdp_flag = check_flash_attention_available(hidden_states)
```
jason-dai, line 112: shall we make `fsdp_flag` a global variable?

rnwang04: As `fsdp_flag` depends on input device, and it may vary during runtime, global variable seems a little dangerous?

rnwang04: I have verifed locally, if we make `fsdp_flag` global, for below test case, it will raise error of `F.scaled_dot_product_attention` at the second generate:
```python
model = model.to('xpu')
model.generate(xxx)

model = model.to('cpu')
model.generate(xxx)
```
And global variable seems not bring obvious performance change.
If we don't make totally isolated env for cpu and gpu, I prefer to keep current usage.
How do you think of it?

jason-dai: ok

jason-dai (approved):
LGTM

PR #9519, opened 2023-11-23
Replace torch.bmm with safe_bmm

leonardozcm:
## Description

This feature is only enabled for whisper on windows gpu to avoid system crash issue raised by ipex bmm(or mkl). Please do not(and no need to) set `lightweight_bmm=True` on core series.

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/model.py
+++ b/python/llm/src/bigdl/llm/transformers/model.py
@@ -100,6 +100,8 @@ class _BaseAutoModelClass:
                                        conducting model optimizations. Default to be None.
         :param replace_embedding: Whether to replace the Embedding layer, may need to set it
             to `True` when running BigDL-LLM on GPU on Windows. Default to be `False`.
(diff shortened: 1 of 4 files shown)


python/llm/src/bigdl/llm/transformers/model.py
```diff
@@ -100,6 +100,8 @@ def from_pretrained(cls,
                                        conducting model optimizations. Default to be None.
         :param replace_embedding: Whether to replace the Embedding layer, may need to set it
             to `True` when running BigDL-LLM on GPU on Windows. Default to be `False`.
+        :param safe_bmm: Whether to replace the torch.bmm ops, may need to set it
+            to `True` when running BigDL-LLM on GPU on Windows. Default to be `False`.
```
jason-dai, line 104: `cpu_embedding` and `lightweight_bmm`?

leonardozcm: Sure

PR #9520, opened 2023-11-23
LLM: support Mistral AWQ models

plusbang:
## Description

Support Mistral AWQ models

### 3. Summary of the change 

- Add mistral support in src code
- Update CPU and GPU examples

### 4. How to test?
- [x] Unit test
- [x] Local test


python/llm/example/CPU/HF-Transformers-AutoModels/Advanced-Quantizations/AWQ/mistral/generate.py
```diff
@@ -0,0 +1,65 @@
+#
```
jason-dai, line 1: what's the difference vs. the LLaMA2 example?

plusbang: > what's the difference vs. the LLaMA2 example?

The difference lies on prompt format and tokenizer class. Should we provide dedicated folder for each model, or just use one python script and list verified AWQ models in readme?

jason-dai: Maybe one python script for CPU, one for GPU

plusbang: > Maybe one python script for CPU, one for GPU

Have updated. As https://github.com/analytics-zoo/nano/issues/780#issuecomment-1824427117, the correctness is currently checked through observing the output text.

jason-dai (approved):
LGTM

PR #9529, opened 2023-11-24
LLM: fix abnormal Mistral GPU accuracy by updating rms_norm

rnwang04:
### 1. Why the change?

We locate the abnormal Mistral (and maybe llama dophin mmlu) GPU accuracy is caused by rms_norm optimization.
With further investigation, I found the issue was caused by `torch.ops.torch_ipex.rms_norm` in 2.0 don't have an eps parameter, and default value is 1e-6 in source code.
However, Mistral and llama's eps is 1e-5 during runtime, which caused the accuracy drop.
Fix it by using `torch.ops.torch_ipex.rms_norm` only for 1e-6 case.

### 2. User API changes

No change.

### 3. Summary of the change 

- fix abnormal Mistral GPU accuracy by updating rms_norm logic

### 4. How to test?
- [x] Unit test
- [x] Local test
image

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/models/baichuan2.py
+++ b/python/llm/src/bigdl/llm/transformers/models/baichuan2.py
@@ -47,18 +47,22 @@ KV_CACHE_ALLOC_BLOCK_LENGTH = 256
 
 
 def baichuan_13b_rms_norm_forward(self, hidden_states):
+    optimized_rms_norm = False
     if hidden_states.device.type == "xpu" and not (self.training and hidden_states.requires_grad):
         if get_ipex_version() <= "2.0.110+xpu":
-            hidden_states, _ = torch.ops.torch_ipex.rms_norm(hidden_states,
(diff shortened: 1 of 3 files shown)


python/llm/src/bigdl/llm/transformers/models/baichuan2.py
```diff
@@ -47,18 +47,22 @@
 
 
 def baichuan_13b_rms_norm_forward(self, hidden_states):
+    optimized_rms_norm = False
     if hidden_states.device.type == "xpu" and not (self.training and hidden_states.requires_grad):
         if get_ipex_version() <= "2.0.110+xpu":
-            hidden_states, _ = torch.ops.torch_ipex.rms_norm(hidden_states,
-                                                             [self.weight.size(0)],
-                                                             self.weight)
+            if self.epsilon == 1e-6:
```
hkvision, line 53: Can combine these two together?

hkvision (approved):

PR #9535, opened 2023-11-27
[VLLM-CPU] Remove xformers from vllm-cpu

gc-fu:
## Description

Remove xformers and its related code so that we do not need to install xformer

### 1. Why the change?

Remove xformer dependency

### 2. How to test?
- [ ] Manually test


xiangyuT (approved):
LGTM

PR #9544, opened 2023-11-28
[LLM] Support loading llama q4_0 gguf model

MeouSker77:
## Description

Support loading llama q4_0 gguf model

### 1. Why the change?

Support loading gguf model

### 2. User API changes

```python

model = AutoModelForCausalLM.from_gguf("/root/yishuo/llama-2-7b-chat.Q4_0.gguf")
```

### 3. Summary of the change 

Support loading gguf model.

It loads gguf model as the following steps:

1. read gguf file to get model config and every tensor's weight
2. dequantize gguf tensor to normal fp32/fp16 pytorch tensor
3. construct a new empty HF model and set its weights
4. call `optimize_model` to quantize HF model, using the qtype read from gguf file

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/gguf/__init__.py
+++ b/python/llm/src/bigdl/llm/transformers/gguf/__init__.py
@@ -1,0 +1,17 @@
+#
+# Copyright 2016 The BigDL Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


jason-dai (approved):
LGTM

PR #9546, opened 2023-11-29
[LLM] support loading q4_1/q5_0/q5_1/q8_0 gguf model

MeouSker77:
## Description

support loading q4_1/q5_0/q5_1/q8_0 gguf model

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/gguf/api.py
+++ b/python/llm/src/bigdl/llm/transformers/gguf/api.py
@@ -19,7 +19,11 @@ from bigdl.llm.utils.common import invalidInputError
 
 
 qtype_map = {
-    2: "sym_int4"       # q4_0
+    2: "sym_int4",      # q4_0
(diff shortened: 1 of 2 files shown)


MeouSker77:
For now, we have supported q4_0/q4_1/q5_0/q5_1/q8_0, which are supported by both llama.cpp and bigdl-llm.

While llama.cpp also has some other quantize methods which are not supported by bigdl-llm, such as q2_k, q3_k, q4_k, shall we support loading them ?

@jason-dai

jason-dai:
> While llama.cpp also has some other quantize methods which are not supported by bigdl-llm, such as q2_k, q3_k, q4_k, shall we support loading them ?
> 
> @jason-dai

We can support it later; let's work out an end-to-end example first.

MeouSker77:
> We can support it later; let's work out an end-to-end example first.

ok

rnwang04 (approved):
LGTM

PR #9550, opened 2023-11-29
LLM: fix loss error on Arc

plusbang:
## Description

Related to https://github.com/analytics-zoo/nano/issues/788

### 4. How to test?
- [ ] Unit test
- [x] Local test (Verify on both single A770, two A770 and PVC 1550)

### DIFF
--- a/python/llm/example/GPU/QLoRA-FineTuning/qlora_finetuning.py
+++ b/python/llm/example/GPU/QLoRA-FineTuning/qlora_finetuning.py
@@ -62,6 +62,7 @@ if __name__ == "__main__":
     )
     model = get_peft_model(model, config)
     cast_lora_weight(model, torch.bfloat16)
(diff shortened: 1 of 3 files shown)


python/llm/example/GPU/QLoRA-FineTuning/qlora_finetuning.py
```diff
@@ -62,6 +62,7 @@
     )
     model = get_peft_model(model, config)
     cast_lora_weight(model, torch.bfloat16)
+    torch.xpu.synchronize()
```
jason-dai, line 65: Maybe we should move these two into `get_peft_model` (if it's xpu, we simply cast the weight and synchronize)

rnwang04 (approved):
LGTM

jason-dai (approved):
LGTM

PR #9553, opened 2023-11-29
[LLM] add a llama2 gguf example

MeouSker77:
## Description

Add a example to load a gguf model and convert it to a bigdl-llm optimized model, then use it to generate


jason-dai (approved):
LGTM

PR #9572, opened 2023-11-30
use fused rms norm

qiuxin2012:
## Description

use fused rms norm when variance_epsilon != 1e-6

### 1. Why the change?

use fused rms norm when variance_epsilon != 1e-6


qiuxin2012:
excepted failed as latest bigdl-core-xe is not released

PR #9578, opened 2023-12-01
Fix setup-llm-env

zhentaocc:
## Description

seems that this action is installing packages with cache which causes incorrect version


Oscilloscope98 (approved):
LGTM

PR #9588, opened 2023-12-04
Configurable TORCH_LINEAR_THRESHOLD from env

qiyuangong:
### 3. Summary of the change 

Make `TORCH_LINEAR_THRESHOLD` through system env, default value is still 96.

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [x] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/low_bit_linear.py
+++ b/python/llm/src/bigdl/llm/transformers/low_bit_linear.py
@@ -59,7 +59,7 @@ import ctypes
 from bigdl.llm.ggml.quantize import ggml_tensor_qtype
 IS_SERVER = is_server()
 IS_SPR = is_spr()
-TORCH_LINEAR_THRESHOLD = 96
+TORCH_LINEAR_THRESHOLD = int(os.getenv("TORCH_LINEAR_THRESHOLD", "96"))
 SYM_INT4 = ggml_tensor_qtype["sym_int4"]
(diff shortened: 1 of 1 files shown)


python/llm/src/bigdl/llm/transformers/low_bit_linear.py
```diff
@@ -59,7 +59,7 @@
 from bigdl.llm.ggml.quantize import ggml_tensor_qtype
 IS_SERVER = is_server()
 IS_SPR = is_spr()
-TORCH_LINEAR_THRESHOLD = 96
+TORCH_LINEAR_THRESHOLD = int(os.getenv("TORCH_LINEAR_THRESHOLD", "96"))
```
jason-dai, line 62: Can we use a larger value @MeouSker77 ?

qiyuangong: Changed default to 256.

yangw1234: You probably should run a grid search on the target machine to determine the default value.

qiyuangong: > You probably should run a grid search on the target machine to determine the default value.

After a grid search on Icelake platform, we can set the default value to 512. :)

MeouSker77 (approved):
LGTM

PR #9597, opened 2023-12-05
gsm8k OOM workaround

zhentaocc:
## Description

A workaround for gsm8k OOM issue.

### 1. Why the change?

https://github.com/analytics-zoo/nano/issues/831

### 2. User API changes

N/A

### 3. Summary of the change 

force the datasets to inference in decreased order of seq len.

### DIFF
--- a/python/llm/dev/benchmark/harness/bigdl_llm.py
+++ b/python/llm/dev/benchmark/harness/bigdl_llm.py
@@ -22,22 +22,23 @@ from lm_eval import utils
 from functools import partial
 
 
+def force_decrease_order(Reorderer):
+    def DecreaseReorderer(arr, fn):
+        def _collate(x):
+            len, tokens = fn(x)
+            len = - abs(len)
+            return len, tokens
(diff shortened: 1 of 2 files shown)


python/llm/dev/benchmark/harness/bigdl_llm.py
```diff
@@ -13,116 +13,42 @@
...
+        # This is a workaround to avoid frequent memory allocation which may cause OOM
+        def force_decrease_order(Reorderer):
+            def DecreaseReorderer(arr, fn):
+                def _collate(x):
+                    len, tokens = fn(x)
+                    len = - abs(len)
+                    return len, tokens
+                return Reorderer(arr, _collate)
+            return DecreaseReorderer
+        utils.Reorderer = force_decrease_order(utils.Reorderer)
```
hkvision, line 39: Don't need to put this in the class init?
Just put them in the global range of this file?

hkvision (approved):

PR #9607, opened 2023-12-06
[LLM] Add traceback error output for win igpu test api in benchmark

Oscilloscope98:
## Description

Add traceback error output for win igpu test api in benchmark


liu-shaojun (approved):
LGTM

PR #9613, opened 2023-12-07
use fused rms norm in chatglm2 and baichuan

qiuxin2012:
## Description

use fused rms norm in chatglm2 and baichuan

### 1. Why the change?

Speedup


jason-dai (approved):
LGTM

PR #9614, opened 2023-12-07
Enable fused layernorm 

qiuxin2012:
## Description

Enable fused layernorm for Bloom, falcon, mpt...

### 1. Why the change?
Speedup in xpu


qiuxin2012:
The model output is the same now, on both xpu and cpu.

hkvision (approved):

PR #9616, opened 2023-12-07
[LLM] vLLM: fix memory leak in prepare_kv_cache

xiangyuT:
## Description

Revert modification in `prepare_kv_cache` to fix memory leak.


gc-fu (approved):
LGTM

PR #9619, opened 2023-12-07
[LLM] vLLM: Delete last_kv_cache before prefilling

xiangyuT:
## Description

Remove last_kv_cache before prefilling to empty memory usage.


gc-fu (approved):
LGTM

PR #9621, opened 2023-12-07
LLM: fix unlora module in qlora finetune

rnwang04:
### 1. Why the change?

https://github.com/analytics-zoo/nano/issues/813#issuecomment-1839816145
speedup effect: https://github.com/analytics-zoo/nano/issues/813#issuecomment-1844782481

### 2. User API changes

No change.

### 3. Summary of the change 

- add a `compute_dtype` for `LowBitLinear` to solve GC issue for unlora module
- add related bf16 cast logic for `LowBitLinear`

### 4. How to test?
- [x] Unit test
- [x] Local test

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/low_bit_linear.py
+++ b/python/llm/src/bigdl/llm/transformers/low_bit_linear.py
@@ -437,12 +437,13 @@ class LowBitLinear(nn.Linear):
         self.compute_dtype = None  # only for training
 
     def forward(self, x: torch.Tensor):
-        # below logic is for training
-        autocast_dtype = get_autocast_dtype(x)
-        if self.compute_dtype is not None and x.device.type == "xpu":
-            x = x.to(self.compute_dtype)  # solve GC issue for unlora module
-        elif autocast_dtype is not None:
-            x = x.to(autocast_dtype)
+        if self.training:
+            # below logic is only for training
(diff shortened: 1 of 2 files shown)


jason-dai (approved):
LGTM

yangw1234:
This PR may introduce a performance regression in first token latency.

jason-dai:
> > This PR may introduce a performance regression in first token latency.
> 
> Looks like the default value of `self.training` is `True` and `model.eval()` will set it false.
> 
> The default value of `x.requires_grad` is also `True` and `with torch.inference_mode()` will set it false.
> 
> So before this PR, user can either use `model.eval()` or `with torch.inference_mode()` to enter the inference only branch, and after this PR user will need to set `model.eval()`.

Does `with torch.inference_mode()` set any flags that we can check?

yangw1234:
> > > This PR may introduce a performance regression in first token latency.
> > 
> > 
> > Looks like the default value of `self.training` is `True` and `model.eval()` will set it false.
> > The default value of `x.requires_grad` is also `True` and `with torch.inference_mode()` will set it false.
> > So before this PR, user can either use `model.eval()` or `with torch.inference_mode()` to enter the inference only branch, and after this PR user will need to set `model.eval()`.
> 
> Does `with torch.inference_mode()` set any flags that we can check?

`torch.is_inference_mode_enabled()` might work.

PR #9631, opened 2023-12-07
[LLM] Adjust some code to adapt to version updates of bigdl.llm.transformers

Romanticoseu:
## Description

1. Import LoraConfig from bigdl.llm.transformers.qlora instead of peft.
2. Update transformers version


Uxito-Ada (approved):
lgtm

PR #9632, opened 2023-12-07
[LLM] Add exclude option in all-in-one performance test

Mingyu-Wei:
### 1. Why the change?
@hkvision @WeiguangHan 

Add exclude option in all-in-one perf test to avoid running some models that might trigger `RuntimeError: Allocation out of device memory`.

### DIFF
--- a/python/llm/dev/benchmark/all-in-one/run.py
+++ b/python/llm/dev/benchmark/all-in-one/run.py
@@ -754,7 +754,7 @@ if __name__ == '__main__':
     for api in conf.test_api:
         for model in conf.repo_id:
             in_out_pairs = conf['in_out_pairs'].copy()
-            if 'exclude' in conf:
+            if excludes:
                 for in_out in conf['in_out_pairs']:
                     model_id_input = model + ':' + in_out.split('-')[0]
                     if model_id_input in excludes:
--- a/python/llm/dev/benchmark/all-in-one/run.py
+++ b/python/llm/dev/benchmark/all-in-one/run.py
@@ -42,7 +42,7 @@ CHATGLM_IDS = ['THUDM/chatglm-6b', 'THUDM/chatglm2-6b', 'THUDM/chatglm3-6b']
 LLAVA_IDS = ['liuhaotian/llava-v1.5-7b']
 
 results = []
-excludes = []
+
 
 def run_model(repo_id, test_api, in_out_pairs, local_model_hub=None, warm_up=1, num_trials=3, num_beams=1, low_bit='sym_int4', cpu_embedding=False):
     # TODO: make a parameter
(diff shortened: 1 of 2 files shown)


python/llm/test/benchmark/arc-perf-test.yaml
```diff
@@ -27,3 +27,6 @@ in_out_pairs:
 test_api:
   - "transformer_int4_gpu"  # on Intel GPU
 cpu_embedding: False # whether put embedding to CPU (only avaiable now for gpu win related test_api)
+exclude:
+  - 'baichuan-inc/Baichuan2-13B-Chat:1024'
+  - 'baichuan-inc/Baichuan2-13B-Chat:2048'
```
hkvision, line 32: Baichuan2-13B is not here actually, you should first exclude 1k and 2k of moss?

Mingyu-Wei: Sure

python/llm/dev/benchmark/all-in-one/run.py
```diff
@@ -747,6 +757,8 @@ def run_transformer_int4_gpu_win(repo_id,
     from omegaconf import OmegaConf
     conf = OmegaConf.load(f'{current_dir}/config.yaml')
     today = date.today()
+    if 'exclude' in conf:
+        excludes = conf['exclude']
```
hkvision, line 761: is this correct? This is in `run_transformer_int4_gpu_win` not in `run_transformer_int4_gpu`?

hkvision, line 761: I think we can process exclude here?
Process exclude to be a list of tuples (model, input_len)?
And then should_exclude is not needed, you only need to check if the tuple is in the list?

python/llm/dev/benchmark/all-in-one/run.py
```diff
@@ -747,11 +747,19 @@ def run_transformer_int4_gpu_win(repo_id,
...
     
     import pandas as pd
     for api in conf.test_api:
         for model in conf.repo_id:
-            run_model(model, api, conf['in_out_pairs'], conf['local_model_hub'], conf['warm_up'], conf['num_trials'], conf['num_beams'],
+            in_out_pairs = conf['in_out_pairs'].copy()
+            if 'exclude' in conf:
```
hkvision, line 757: can use `if excludes` here?

hkvision (approved):

python/llm/test/benchmark/arc-perf-test.yaml
```diff
@@ -27,3 +27,6 @@ in_out_pairs:
 test_api:
   - "transformer_int4_gpu"  # on Intel GPU
 cpu_embedding: False # whether put embedding to CPU (only avaiable now for gpu win related test_api)
+exclude:
+  - 'fnlp/moss-moon-003-sft:1024'
+  - 'fnlp/moss-moon-003-sft:2048'
```
WeiguangHan, line 32: Need to add `'bigscience/bloomz-7b1:2048'` which causes OOM error and segment fault randomly.

PR #9633, opened 2023-12-07
LLM: integrate sdp kernel for FP16 rest token inference on GPU [DG2/ATSM]

rnwang04:
## Description

Should merge after https://github.com/intel-analytics/llm.cpp/pull/180 is merged.

### 1. Why the change?

https://github.com/analytics-zoo/nano/issues/840

### 2. User API changes

No change.

### 3. Summary of the change 

- update fp16 convert logic as some change in kernel implementation

- supports **sdp** kernel of fp16 and fp32 for **DG2 & ATS-M GPU**, now it can provide optimization for **fp16 rest token**.

As a starting point, only change llama modeling file here, and it can be easily used for other models.

### 4. How to test?
- [x] Unit test
- [x] Local test for `Llama-2-7b-chat-hf` and `Nous-Hermes-Llama2-13b` on arc04
performance update here : https://github.com/analytics-zoo/nano/issues/840#issuecomment-1849646767

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/models/llama.py
+++ b/python/llm/src/bigdl/llm/transformers/models/llama.py
@@ -286,7 +286,7 @@ def use_esimd_sdp(q_len, head_dim, query_states):
         # esimd_sdp only has optimization for FP16 now
         return False
     else:
-        device_name = torch.xpu.get_device_name(query_device.index)
+        device_name = torch.xpu.get_device_name(query_states.device.index)
         if device_name.startswith("Intel(R) Arc(TM)") or \
                 device_name.startswith("Intel(R) Data Center GPU Flex"):
             import linear_fp16_esimd
--- a/python/llm/src/bigdl/llm/transformers/models/llama.py
+++ b/python/llm/src/bigdl/llm/transformers/models/llama.py
@@ -219,8 +219,9 @@ def llama_attention_forward_4_31(
                                                      value_states,
                                                      is_causal=True)
         attn_weights = None
-    elif q_len == 1 and query_states.device.type == "xpu" and query_states.dtype == torch.float16  and \
-        flex_sdp_check(self.head_dim, query_states.device):
+    elif q_len == 1 and query_states.device.type == "xpu" and \
+            query_states.dtype == torch.float16 and \
+            flex_sdp_check(self.head_dim, query_states.device):
         # now only use flex sdp for rest token
         import linear_fp16_esimd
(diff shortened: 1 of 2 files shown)


python/llm/src/bigdl/llm/transformers/models/llama.py
```diff
@@ -219,6 +219,22 @@ def llama_attention_forward_4_31(
                                                      value_states,
                                                      is_causal=True)
         attn_weights = None
+    elif q_len == 1 and query_states.device.type == "xpu" and \
+            query_states.dtype == torch.float16 and \
+            flex_sdp_check(self.head_dim, query_states.device):
+        # now only use flex sdp for rest token
+        import linear_fp16_esimd
+        if hasattr(linear_fp16_esimd, "sdp_forward"):
```
jason-dai, line 227: Combine these condition into one method like `use_esmid_sdp`?

jason-dai (approved):
LGTM

python/llm/src/bigdl/llm/transformers/models/llama.py
rnwang04: Have combined these conditions into `use_esmid_sdp`.

python/llm/src/bigdl/llm/transformers/models/llama.py
```diff
@@ -266,6 +272,32 @@ def check_flash_attention_available(query):
...
+        return False
+    elif query_states.device.type != "xpu":
+        # esimd_sdp only support GPU now
+        return False
+    elif query_states.dtype != torch.float16:
+        # esimd_sdp only has optimization for FP16 now
+        return False
+    else:
+        device_name = torch.xpu.get_device_name(query_states.device.index)
+        if device_name.startswith("Intel(R) Arc(TM)") or \
```
MeouSker77, line 290: others LGTM, but it seems MTL doesn't support lsc function, which is widely used in esimd kernel. The esimd kernel's performance will be poor without lsc function, so we can simply think MTL doesn't support esimd kernel. The igpu device name of MTL is `Intel(R) Arc(TM) Graphics`, we should distinguish it from Arc

MeouSker77 (approved):

python/llm/src/bigdl/llm/transformers/models/llama.py
rnwang04: Use "Intel(R) Arc(TM) A" as current workaournd to distinguish MTL from Arc.
May change this part as we know how to locate MTL.

PR #9639, opened 2023-12-08
fix new lowbitlinear status

qiuxin2012:
## Description

A new LowBitLinear's training is True by default, but the model's training maybe false.
We should make them the same.

### 1. Why the change?

Fixed performance down after https://github.com/intel-analytics/BigDL/pull/9621

### 2. User API changes

None


jason-dai (approved):
LGTM

PR #9641, opened 2023-12-08
Check Arc perf test total results

WeiguangHan:
@hkvision Fix the visualization of highlighted values in html of arc perf test.

### DIFF
--- a/python/llm/test/benchmark/check_html_lines.py
+++ b/python/llm/test/benchmark/check_html_lines.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-# Python program to concat CSVs
+# Python program to check if the number of lines in html meets expectation
 
 import os
 import sys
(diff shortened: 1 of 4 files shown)


python/llm/test/benchmark/check_html_lines.py
```diff
@@ -0,0 +1,49 @@
+#
```
hkvision, line 1: rename to `check_results.py`

python/llm/test/benchmark/check_html_lines.py
```diff
@@ -0,0 +1,49 @@
...
+# Python program to check if the number of lines in html meets expectation
+
+import os
+import sys
+import argparse
+import pandas as pd
+
+def main():
+    parser = argparse.ArgumentParser(description="check if the number of lines in html meets expectation")
+    parser.add_argument("-f", "--folder_path", type=str, dest="folder_path",
```
hkvision, line 26: I think you only need to input one file?
And you can directly use the raw csv file to do so in the original folder?

python/llm/test/benchmark/check_html_lines.py
WeiguangHan: sure

python/llm/test/benchmark/check_html_lines.py
WeiguangHan: ok. Will change this according to your suggestions.

PR #9650, opened 2023-12-11
LLM: fix the issue that may generate blank html

WeiguangHan:
@hkvision Fix the issue that may generate blank html.


Oscilloscope98:
Please also fix the case that we have a single csv to generate html :) thanks

Oscilloscope98 (commented):
Please also fix the case that we have a single csv to generate html :) thanks

PR #9657, opened 2023-12-12
LLM: support `optimize_model=True` for Mixtral model

plusbang:
## Description

Support `optimize_model=True` for Mixtral model:
- [ ] attention forward
- [x] RMSNorm

### 4. How to test?
- [ ] Unit test
- [x] Local test

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/convert.py
+++ b/python/llm/src/bigdl/llm/transformers/convert.py
@@ -620,8 +620,8 @@ def _optimize_post(model, lightweight_bmm=False):
             model.config.architectures[0] == "MixtralForCausalLM":
             # For DiscoResearch/mixtral-7b-8expert
             invalidInputError(version.parse(trans_version) >= version.parse("4.36.0"),
-                            "Please upgrade transformers to 4.36.0 or higher version "
-                            "to run Mixtral models.")
+                              "Please upgrade transformers to 4.36.0 or higher version "
+                              "to run Mixtral models.")
             modeling_module_name = model.__class__.__module__
             module = importlib.import_module(modeling_module_name)
             convert_forward(model,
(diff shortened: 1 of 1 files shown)


python/llm/src/bigdl/llm/transformers/convert.py
```diff
@@ -616,6 +616,15 @@ def _optimize_post(model, lightweight_bmm=False):
         convert_forward(model,
                         module.MistralRMSNorm,
                         llama_rms_norm_forward)
+    elif model.config.model_type == "mixtral":
```
jason-dai, line 619: Also check `model.config.architectures` for mistral and mixtral (to support DiscoResearch models)?

plusbang: > Also check `model.config.architectures` for mistral and mixtral (to support DiscoResearch models)?

Have added related check.

python/llm/src/bigdl/llm/transformers/convert.py
```diff
@@ -616,6 +616,24 @@ def _optimize_post(model, lightweight_bmm=False):
...
+            convert_forward(model,
+                            module.MixtralRMSNorm,
+                            llama_rms_norm_forward)
+        elif model.config.model_type == "mistral":
+            # For DiscoResearch/mixtral-7b-8expert
+            modeling_module_name = model.__class__.__module__
+            module = importlib.import_module(modeling_module_name)
+            convert_forward(model,
+                            module.MistralRMSNorm,
+                            llama_rms_norm_forward)
```
jason-dai, line 636:
```python
if model.config.model_type == "mixtral":
    ...
elif model.config.model_type == "mistral":
   if model.config.architectures is not None and model.config.architectures[0] == "MixtralForCausalLM":
      ...
   else
      ...
```

jason-dai (approved):
LGTM

PR #9658, opened 2023-12-12
disable test_optimized_model.py temporarily due to out of memory on A730M(pr validation machine)

qiuxin2012:
## Description

- disable test_optimized_model.py temporarily due to out of memory on A730M(pr validation machine). As it may make the pr validation machine stuck for tests.
- Also disable speech_seq2seq tests for now due to wrong answers (https://github.com/intel-analytics/BigDL/actions/runs/7178356615/job/19547287595)


Oscilloscope98 (approved):
LGTM. We may fix the two problems later.

PR #9660, opened 2023-12-12
[LLM] Remove empty cache between each iteration of generation in benchmark for igpu win

Oscilloscope98:
## Description
Remove empty cache between each iteration of generation in benchmark for igpu win as it may make the first token slower.

### How to test?
- [ ] Local tests


MeouSker77 (approved):
LGTM

PR #9669, opened 2023-12-13
LLM: update moe block convert to optimize rest token latency of Mixtral

rnwang04:
### 1. Why the change?

https://github.com/analytics-zoo/nano/issues/847#issuecomment-1851572877
https://github.com/analytics-zoo/nano/issues/847#issuecomment-1853158787

### 2. User API changes

No change. Just set `optimize_model=True` for Mixtral.

### 3. Summary of the change 

- convert `MixtralSparseMoeBlock` of mixtral to simpify rest token inference logic, which greatly reduce latency

performance and accuracy validation is here: https://github.com/analytics-zoo/nano/issues/847#issuecomment-1853368957

### 4. How to test?
- [x] Local test with `Mixtral-8x7B-Instruct-v0.1`
- [x] Unit test

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/models/mixtral.py
+++ b/python/llm/src/bigdl/llm/transformers/models/mixtral.py
@@ -1,0 +1,103 @@
+#
+# Copyright 2016 The BigDL Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Some parts of this file is adapted from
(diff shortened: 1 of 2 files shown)


python/llm/src/bigdl/llm/transformers/models/mixtral.py
```diff
@@ -0,0 +1,108 @@
...
+        )
+        # One hot encode the selected experts to create an expert mask
+        # this will be used to easily index which expert is going to be sollicitated
+        expert_mask = torch.nn.functional.one_hot(selected_experts,
+                                                  num_classes=self.num_experts).permute(2, 1, 0)
+
+        # Loop over all available experts in the model and perform the computation on each expert
+        for expert_idx in range(self.num_experts):
+            expert_layer = self.experts[expert_idx]
+            idx, top_x = torch.where(expert_mask[expert_idx])
```
shane-huang, line 77: shall we also optimize this part (e.g. replace w/ c++ implementation as yang suggested?) for the 1st token speed?

shane-huang (approved):

python/llm/src/bigdl/llm/transformers/models/mixtral.py
rnwang04: for first token optimization, may need further validation, maybe we can optimize this part later if needed.

jason-dai: We also need to optimize sequence_length>1 and batch_size>1 for finetuning

PR #9678, opened 2023-12-13
[LLM] Correct prompt format of Qwen in generate.py

Jasonzzt:
## Description

The previous format was Llama's.
Actually, the recommended format is chatml.


qiyuangong:
Qwen-7B-Chat's default format is `chatml` https://huggingface.co/Qwen/Qwen-7B-Chat/blob/main/generation_config.json#L2

qiyuangong (approved):
LGTM

PR #9683, opened 2023-12-14
fix baichuan2-7b 1st token performance regression on xpu

qiuxin2012:
## Description

fix baichuan2-7b 1st token performance regression

### 1. Why the change?

Fixed https://github.com/analytics-zoo/nano/issues/849
The last Linear(lm_head)'s `training` status is true, so the condition in lowbitlinear will be wrong.
Add `and not torch.is_inference_mode_enabled()` for double checking.

### 2. User API changes

None

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/low_bit_linear.py
+++ b/python/llm/src/bigdl/llm/transformers/low_bit_linear.py
@@ -437,7 +437,10 @@ class LowBitLinear(nn.Linear):
         self.compute_dtype = None  # only for training
 
     def forward(self, x: torch.Tensor):
-        if self.training:
+        # Due to inconsistent training status in some models like Baichuan-7b-Chat,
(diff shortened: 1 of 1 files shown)


python/llm/src/bigdl/llm/transformers/low_bit_linear.py
```diff
@@ -476,7 +476,7 @@ def forward(self, x: torch.Tensor):
                 x_2d = x_2d.contiguous()
 
             input_seq_size = x_shape[1]
-            if self.training:
+            if self.training and not torch.is_inference_mode_enabled():
```
jason-dai, line 479: need to check L440 too?

qiuxin2012: fixed

rnwang04 (approved):
LGTM

PR #9688, opened 2023-12-14
handle empty fused norm result

qiuxin2012:
## Description

handle empty fused norm result when fused norm failed.

### 1. Why the change?

handle empty fused norm result when fused norm failed.

### 2. User API changes

None

### 4. How to test?
- [ ] Manual tes,t, force kernel return empty tensor

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/models/baichuan2.py
+++ b/python/llm/src/bigdl/llm/transformers/models/baichuan2.py
@@ -50,24 +50,25 @@ def baichuan_13b_rms_norm_forward(self, hidden_states):
     if hidden_states.device.type == "xpu" and not (self.training and hidden_states.requires_grad):
         if get_ipex_version() <= "2.0.110+xpu":
             import linear_q4_0
-            hidden_states = linear_q4_0.fused_rms_norm(hidden_states,
+            result = linear_q4_0.fused_rms_norm(hidden_states,
                                                        [self.weight.size(0)],
                                                        self.weight,
                                                        None,
                                                        self.epsilon)
+            # if nelement == 0, means fused norm failed, go back to python implement.
+            if result.nelement != 0:
+                return result
         else:
-            hidden_states = torch.ops.torch_ipex.fast_rms_norm(hidden_states,
+            return torch.ops.torch_ipex.fast_rms_norm(hidden_states,
                                                                [self.weight.size(0)],
                                                                self.weight,
                                                                None,
                                                                self.epsilon)
-        return hidden_states
-    else:
-        input_dtype = hidden_states.dtype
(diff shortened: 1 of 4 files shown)


rnwang04 (approved):

python/llm/src/bigdl/llm/transformers/models/baichuan2.py
```diff
@@ -50,24 +50,25 @@ def baichuan_13b_rms_norm_forward(self, hidden_states):
...
                                                        [self.weight.size(0)],
                                                        self.weight,
                                                        None,
                                                        self.epsilon)
+            # if nelement == 0, means fused norm failed, go back to python implement.
+            if result.nelement != 0:
+                return result
         else:
-            hidden_states = torch.ops.torch_ipex.fast_rms_norm(hidden_states,
+            return torch.ops.torch_ipex.fast_rms_norm(hidden_states,
```
jason-dai, line 62: We need to call `fused_rms_norm` in both cases to support `xpu:1`, yes? @lalalapotter

python/llm/src/bigdl/llm/transformers/models/baichuan2.py
```diff
@@ -50,24 +50,25 @@ def baichuan_13b_rms_norm_forward(self, hidden_states):
...
         if get_ipex_version() <= "2.0.110+xpu":
             import linear_q4_0
-            hidden_states = linear_q4_0.fused_rms_norm(hidden_states,
+            result = linear_q4_0.fused_rms_norm(hidden_states,
                                                        [self.weight.size(0)],
                                                        self.weight,
                                                        None,
                                                        self.epsilon)
+            # if nelement == 0, means fused norm failed, go back to python implement.
+            if result.nelement != 0:
```
jason-dai, line 59: do you need to check for `2.1`?

python/llm/src/bigdl/llm/transformers/models/baichuan2.py
lalalapotter: Yes, for ipex 2.1, `torch.ops.torch_ipex.fast_rms_norm` is not work normally for `xpu:1`.

python/llm/src/bigdl/llm/transformers/models/baichuan2.py
qiuxin2012: No, if can not be fused, fast_rms_norm will go to another native implement. We don't have the native implement, so we go back to python.

jason-dai: You should use fused_rms_norm for both 2.0 & 2.1

PR #9696, opened 2023-12-15
Fix path issue in install.sh & add Qwen dependencies in llm-cpu Dockerfile

Zhengjin-Wang:


glorysdj:
please also update serving docker

glorysdj (approved):
LGTM

PR #9708, opened 2023-12-18
LLM: Add QWen gguf

hzjane:
## Description
Add QWen gguf support, [model path](https://huggingface.co/Lemmih/Qwen-GGUF/tree/main)

### 1. Why the change?
Add QWen gguf

### 2. User API changes
No

### 3. Summary of the change 
Add QWen gguf

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...


jason-dai:
Is there an implementation from HuggingFace for QWen instead?

hzjane:
> Is there an implementation from HuggingFace for QWen instead?


It seems that Chinese model such as baichuan and qwen are not supported by HuggingFace API.

jason-dai:
Then maybe not add it at this moment

PR #9716, opened 2023-12-19
[LLM] Improve chatglm2/3 rest token performance with long context

MeouSker77:
## Description

When `is_causal` is False and `attenon_mask` is None.
The sdp has three ops:
1. `attn = torch.matmul(query, key * scale)`
2. `attn = torch.softmax(attn, -1)`
3. `context_layer = torch.matmul(attn, value_layer)`

IPEX's sdp used a [trick](https://tinyurl.com/sudb9s96) for numerical stability in the 1st step. But according to the issue, the numerical instability only happens when the model is very large, such as 104B.

There are several almost equivalent ways to compute the 1st step:
- `(query * scale) @ key`
- `query @ (key * scale)`
- `(query @ key) * scale`
- `(query * sqrt(scale)) @ (key * sqrt(scale))`

IPEX's trick used the last one.

The second one and the last one multiple `scale` by `key`, but when input length is 1024, the `query`'s shape is `[1, 32, 1, 128]`,  `key`'s shape is `[1, 32, 128, 1024]`, the `query @ key`'s shape is `[1, 32, 1, 1024]`. Multiple `scale` by `query` is faster than by `key`.

When input length is 1024, total `key` and `value` will take about 1GB memory, so multiple `scale` by `key` is very slow.

This PR used the first way to compute the 1st step for best performance.

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/models/chatglm2.py
+++ b/python/llm/src/bigdl/llm/transformers/models/chatglm2.py
@@ -17,6 +17,7 @@
 # https://huggingface.co/THUDM/chatglm2-6b/blob/8eb45c842594b8473f291d0f94e7bbe86ffc67d8/modeling_chatglm.py
 #
 
+import math
 import torch
 from typing import Optional, Tuple, List
 import torch.nn.functional as F
@@ -370,9 +371,13 @@ def core_attn_forward_8eb45c(self, query_layer, key_layer, value_layer, attentio
                                                                              value_layer,
                                                                              attention_mask,
                                                                              is_causal=True)
+        elif attention_mask is None:
+            scaling_factor = 1 / math.sqrt(query_layer.size(-1))
+            attn = torch.matmul(query_layer * scaling_factor, key_layer.transpose(-2, -1))
+            attn = torch.softmax(attn, -1)
+            context_layer = torch.matmul(attn, value_layer)
         else:
-            if attention_mask is not None:
-                attention_mask = ~attention_mask
+            attention_mask = ~attention_mask
             context_layer = torch.nn.functional.scaled_dot_product_attention(query_layer,
                                                                              key_layer,
                                                                              value_layer,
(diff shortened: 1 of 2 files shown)


python/llm/src/bigdl/llm/transformers/models/chatglm2.py
```diff
@@ -370,9 +371,13 @@ def core_attn_forward_8eb45c(self, query_layer, key_layer, value_layer, attentio
                                                                              value_layer,
                                                                              attention_mask,
                                                                              is_causal=True)
+        elif attention_mask is None:
+            scaling_factor = 1 / math.sqrt(query_layer.size(-1))
+            attn = torch.matmul(query_layer * scaling_factor, key_layer.transpose(-2, -1))
```
rnwang04, line 376: Does this optimization can be used for first & rest token or only for rest token?
If only for rest token, shall we add limitation such as `elif attention_mask is None and bs == 1` ?

MeouSker77: > Does this optimization can be used for first & rest token or only for rest token?

not tried yet, may try 1st token later.

> If only for rest token, shall we add limitation such as `elif attention_mask is None and bs == 1`

I think the condition `query_layer.shape[2] == key_layer.shape[2]` is enough.

rnwang04 (approved):
LGTM

jason-dai:
Does this apply to Baichuan?

MeouSker77:
> Does this apply to Baichuan?

yes for baichaun2-7b, no for baichuan2-13b. baichuan2-7b has obvious speedup, while baichaun2-13b are already using `(query @ key) * scale`.

jason-dai:
> > Does this apply to Baichuan?
> 
> yes. baichuan2-7b has obvious speedup, while baichaun2-13b are already using `(query @ key) * scale`.

OK - let's also add Baichuan2 support

MeouSker77:
> OK - let's also add Baichuan2 support

added it

MeouSker77:
qwen-7b is also already using `(query @ key) * scale`

PR #9709, opened 2023-12-18
LLM: add mlp optimization for mixtral

rnwang04:
### 1. Why the change?

apply mlp fusion optimization for mixtral model

### 2. User API changes

No change.

### 3. Summary of the change 

- apply mlp fusion optimization for mixtral model
- optimization effect and accuracy verification is here: https://github.com/analytics-zoo/nano/issues/847#issuecomment-1859832275

### 4. How to test?
- [x] Local test of `mistralai/Mixtral-8x7B-Instruct-v0.1`
- [ ] Unit test

PR #9718, opened 2023-12-19
fix mlp batch size check

MeouSker77:


gc-fu (approved):
I am doing a quick test with this PR using my workload.  It now worked normally when using `llama_mlp_forward` and not using `llama_mlp_forward`.  So LGTM.

Note: I only tested llama model.

rnwang04 (approved):
LGTM

jason-dai:
Also need to fix https://github.com/intel-analytics/BigDL/pull/9709? @rnwang04

rnwang04:
> Also need to fix #9709? @rnwang04

I think #9709 is ok for this, no need to fix it.

jason-dai:
> > Also need to fix #9709? @rnwang04
> 
> I think #9709 is ok for this, no need to fix it.

It seems you are using different conditions to check; it might be better to refactor the logic into one utility method.

PR #9724, opened 2023-12-20
mixstral fused qkv and rope

qiuxin2012:
## Description

mixstral fused qkv and rope, https://github.com/analytics-zoo/nano/issues/875

### 1. Why the change?

mixstral fused qkv and rope


sgwhat (approved):
LGTM

PR #9725, opened 2023-12-20
Load Mixtral gguf in a block-wise way

Uxito-Ada:
## Description

load in and process gguf weight in a block-wise way in the use of iteration, in order to decrease peak memory usage so that mixtral gguf can be loaded on platforms with fewer memory resources

### 1. Why the change?

as description

### 2. User API changes

no

### 3. Summary of the change 

modify gguf/model/mixtral.py to load and process gguf weight in a block-wise way

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [x] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/gguf/gguf.py
+++ b/python/llm/src/bigdl/llm/transformers/gguf/gguf.py
@@ -246,6 +246,24 @@ class GGUFTensorLoader:
                 tensor = self.convert_funcs[qtype](tensor, size, ndims, dims)
                 yield name, tensor
 
+    def load_while_process(self, process):
+        with open(self.fpath, 'rb') as f:
+            for name, ndims, dims, qtype, offset in tqdm(self.infos, desc="Loading gguf tensors"):
+                total_ne = functools.reduce(lambda x, y: x * y, dims)
+                invalidInputError(total_ne % self.block_ne[qtype] == 0,
+                                  f"wrong elements num: {dims}")
+
+                size = total_ne // self.block_ne[qtype] * self.block_size[qtype]
+                invalidInputError(size != 0, f"unsupported quantize type: {qtype}")
+
+                offset += self.base_offset
+                f.seek(offset)
+                data = f.read(size)
+                arr = numpy.frombuffer(data, dtype=numpy.uint8)
+                tensor = torch.from_numpy(arr)
+                tensor = self.convert_funcs[qtype](tensor, size, ndims, dims)
+                process(name, tensor)
+
     def convert_f32_tensor(self, tensor: torch.Tensor, size: int, ndims: int, dims: int):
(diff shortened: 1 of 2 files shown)


python/llm/src/bigdl/llm/transformers/gguf/gguf.py
```diff
@@ -246,6 +246,24 @@ def __iter__(self):
                 tensor = self.convert_funcs[qtype](tensor, size, ndims, dims)
                 yield name, tensor
 
+    def load_while_process(self, process):
```
Uxito-Ada, line 249: passing a handle to ggufFileLoader is better than applying iter as for performance test

Uxito-Ada:
The below optimizations are applied in order to save memory:
1. remove copy of numpy array in `load_while_process`
2. process in a block-wise way by passing a handle to `load_while_process` (rather than use iter of ggufFileLoader, since better performance)
3. remove repeated value.to(FP16) when loading gguf tensors, which finally to(FP16) once in `set_module_tensor_to_device`.
4. remove model.to(cpu) since it is already on cpu in `set_module_tensor_to_device`.

MeouSker77 (approved):
LGTM

qiuxin2012:
Please make sure all tests are passed, especially style check.

qiuxin2012:
You can use `./dev/test/lint-python` in llm to reproduce the sytle check result on you code base.

jason-dai:
> The below optimizations are applied in order to save memory:
> 
> 1. remove copy of numpy array in `load_while_process`
> 2. process in a block-wise way by passing a handle to `load_while_process` (rather than use iter of ggufFileLoader, since better performance)
> 3. remove repeated value.to(FP16) when loading gguf tensors, which finally to(FP16) once in `set_module_tensor_to_device`.
> 4. remove model.to(cpu) since it is already on cpu in `set_module_tensor_to_device`.

What's the current memory usage?

Uxito-Ada:
> > The below optimizations are applied in order to save memory:
> > 
> > 1. remove copy of numpy array in `load_while_process`
> > 2. process in a block-wise way by passing a handle to `load_while_process` (rather than use iter of ggufFileLoader, since better performance)
> > 3. remove repeated value.to(FP16) when loading gguf tensors, which finally to(FP16) once in `set_module_tensor_to_device`.
> > 4. remove model.to(cpu) since it is already on cpu in `set_module_tensor_to_device`.
> 
> What's the current memory usage?

Hi Jason, in my develop-and-test environment, an about 11G memory saving has been found compared with the previous version. As it was not conducted in a strictly standard environment, we are WIP to benchmark the latest version and update our internal data soon.

jason-dai:
Need to apply this to other models @hzjane ?

PR #9727, opened 2023-12-20
LLM: using separate threads to do inference

WeiguangHan:
@hkvision Using separate threads to do inference.

### DIFF
--- a/python/llm/dev/benchmark/all-in-one/run.py
+++ b/python/llm/dev/benchmark/all-in-one/run.py
@@ -45,20 +45,21 @@ LLAVA_IDS = ['liuhaotian/llava-v1.5-7b']
 results = []
 excludes = []
 
-def run_model_in_thread(model, in_out, tokenizer, result, warm_up, num_beams, input_ids, out_len, actual_in_len, i):
-    st = time.perf_counter()
-    output_ids = model.generate(input_ids, do_sample=False, max_new_tokens=out_len,
-                                num_beams=num_beams)
-    torch.xpu.synchronize()
-    end = time.perf_counter()
-    output_ids = output_ids.cpu()
-    print("model generate cost: " + str(end - st))
-    output = tokenizer.batch_decode(output_ids)
-    print(output[0])
-    actual_out_len = output_ids.shape[1] - actual_in_len
-    if i >= warm_up:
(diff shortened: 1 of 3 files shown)


python/llm/dev/benchmark/all-in-one/run.py
```diff
@@ -368,42 +383,28 @@ def run_transformer_int4_gpu(repo_id,
...
+            input_str = open(f"prompt/{test_length}.txt", 'r').read()
+            # As different tokenizer has different encodings,
+            # slice the input_ids to ensure the prompt length is required length.
+            input_ids = tokenizer.encode(input_str, return_tensors="pt")
+            input_ids = input_ids[:, :in_len]
+            true_str = tokenizer.batch_decode(input_ids)[0]
+            input_ids = tokenizer.encode(true_str, return_tensors="pt").to('xpu')
+            actual_in_len = input_ids.shape[1]
+            result[in_out] = []
+            for i in range(num_trials + warm_up):
```
hkvision, line 404: need to put the for loop in the thread as well.

python/llm/dev/benchmark/all-in-one/run.py
```diff
@@ -368,42 +383,28 @@ def run_transformer_int4_gpu(repo_id,
...
-                pass
+            in_out_len = in_out.split("-")
+            in_len = int(in_out_len[0])
+            out_len = int(in_out_len[1])
+            # As different tokenizer has different encodings,
+            # in_len.txt maybe shorter than we need,
+            # use much longer context to make sure input length
+            test_length = min(in_len*2, 8192)
+            while test_length not in [32, 256, 1024, 2048, 8192]:
+                test_length = test_length * 2
```
hkvision, line 394: need to check this logic, maybe in another PR.

python/llm/dev/benchmark/all-in-one/run.py
WeiguangHan: sure

python/llm/dev/benchmark/all-in-one/run.py
WeiguangHan: Will do it later.

hkvision (approved):

hkvision:
At this moment, we first put model generate for target number of trials in a thread.
Later if want to further refine this, we may put the entire run_model in a thread.

PR #9730, opened 2023-12-20
[LLM] IPEX auto importer set on by default

qiyuangong:
## Description

Set BIGDL_IMPORT_IPEX default as True, i.e., auto import IPEX when bigdl-xpu is installed.
Remove IPEX import from GPU example.

### 1. Why the change?

Performance tests show no impact from auto importer. 

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [x] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/python/llm/dev/benchmark/all-in-one/run-arc.sh
+++ b/python/llm/dev/benchmark/all-in-one/run-arc.sh
@@ -1,6 +1,5 @@
 source /opt/intel/oneapi/setvars.sh
 export USE_XETLA=OFF
-export BIGDL_IMPORT_IPEX="True"
 export SYCL_PI_LEVEL_ZERO_USE_IMMEDIATE_COMMANDLISTS=1
 
 python run.py # make sure config YAML file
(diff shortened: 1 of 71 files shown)


python/llm/dev/benchmark/all-in-one/run-arc.sh
```diff
@@ -1,5 +1,6 @@
 source /opt/intel/oneapi/setvars.sh
```
WeiguangHan, line 1: Should be change to `source /home/arda/intel/oneapi/setvars.sh`

python/llm/dev/benchmark/all-in-one/run-max-gpu.sh
```diff
@@ -1,5 +1,6 @@
 source /opt/intel/oneapi/setvars.sh
```
WeiguangHan, line 1: Same as above. Should be change to `source /home/arda/intel/oneapi/setvars.sh`

Oscilloscope98 (approved):
LGTM

PR #9736, opened 2023-12-20
fix arc ut test

Ricky-Ting:
### 1. Why the change?

- relax the threshold of chatglm2
- replace Falcon testcase with LLaMA2
- modify the testing standard of `test_optimize_model` to `torch.isclose`
- add `with torch.inference_mode()` to reduce gpu memory consumption 
- resume all the arc test
- Fix some environmental variables

https://github.com/analytics-zoo/nano/issues/852
https://github.com/analytics-zoo/nano/issues/874


python/llm/test/inference_gpu/test_transformers_api.py
```diff
@@ -36,38 +37,192 @@
...
+
+    def test_chatglm2_gpu_model(self):
+
+        Model = AutoModel
+        Tokenizer = AutoTokenizer
+        model_path = os.environ.get('CHATGLM2_6B_ORIGIN_PATH')
+        # currently only need to compare the output of one self-attention layer.
+        layer_norm = "transformer.encoder.layers.27.input_layernorm"
+        self_attn = "transformer.encoder.layers.27.self_attention"
+        lower_bound = 1e-3
```
hkvision, line 222: The main thing we need to figure out is why previously 5e-5 works but now it doesn't work. Maybe to relax the bound is correct, but we need to confirm the reason.

Oscilloscope98: More information of this difference can be found here: https://github.com/analytics-zoo/nano/issues/852#issuecomment-1863733041

We may have some more investigation later but I think we can merge this PR for now :)

Oscilloscope98 (approved):
LGTM

Oscilloscope98:
@Ricky-Ting @hkvision It seems that parts of the arc ut tests cannot pass for current bigdl-llm: https://github.com/intel-analytics/BigDL/actions/runs/7346047762/job/20000241792. May need to have a look on what happened

PR #9748, opened 2023-12-21
LLM: remove startcorder-15.5b

WeiguangHan:
@hkvision Remove startcorder-15.5b


hkvision (approved):

hkvision:
This model may fail at model load time, using a separate thread won't fix this issue, so temporarily remove it.

PR #9750, opened 2023-12-21
bigdl-llm stable version: let the perf test fail if the difference between perf and baseline is greater than 5%

liu-shaojun:
### 1. Why the change?

This submission lets the spr perf test fail if the difference between perf and baseline is greater than 5%


glorysdj (approved):
LGTM

PR #9740, opened 2023-12-21
LLM: bigdl-llm lora support & lora example

rnwang04:
### 1. Why the change?

Provide better support of lora on Intel GPUs, and accelerate the training speed.
benchmark and verification is here : https://github.com/analytics-zoo/nano/issues/878#issuecomment-1866091540

### 2. User API changes

See example.

### 3. Summary of the change 

- add `BF16Linear` & `LoraBF16Linear` class to accelerate bf16 lora finetuning
- add cast function to save memory and accelerate the training speed
- provide single card example of lora, and local verification
- provide multi card example of lora, and local verification
- refactor code of lora and qlora

### 4. How to test?
- [x] Unit test
- [x] Local test of llama2-7b lora finetuning on single tile PVC
- [x] Local test of llama2-7b lora finetuning on multi card PVC

PR #9755, opened 2023-12-22
LLM: fix `BF16Linear` related training & inference issue

rnwang04:
### 1. Why the change?

https://github.com/intel-analytics/BigDL/pull/9740#issuecomment-1867176404
Based on above comment, after further investigation, I found the below command's peak GPU memory is about **~26G**, which shows `torch_dtype=torch.bfloat16` doesn't take effect at this case.
```python
model = AutoModelForCausalLM.from_pretrained(model_path, use_cache=True,
                                             load_in_low_bit="bf16",
                                             torch_dtype=torch.bfloat16)
model.to('xpu')
```
At the same time, I found above model fail to do inference as two errors:

- current `BF16Linear` is missing `qtype` attribute, which will cause error in llama modeling judgement
- current flash attention will convert hidden states to fp16 (for first token), which cause bf16 inference failed.

### 2. User API changes

No change.

### 3. Summary of the change 

- Fix qtype & flash attention usage to support bf16 inference, now we can run bf16 inference by bigdl-llm which provide optimized latency
- Reduce GPU memory usage of training & inference by handling `to(torch.float32)` in `from_pretrained` and `to(torch.float32)` in `prepare_model_for_kbit_training`
- Support lora on arc and provide related script
- Add support of `torch_dtype` parameter: if user pass specific `torch_dtype`, convert non-BigDL-linear to specified type
- Update `export_merged_model.py` script to cope with new `training_mode` attribute

### 4. How to test?
- [x] Unit test
- [x]  Local test of llama2-7b lora finetuning on single tile PVC
- [x] Local test of llama2-7b lora finetuning on Arc A770
- [x] Local test of llama2-7b inference (bf16) on single tile PVC
- [x] Local test of llama2-7b inference (bf16) on Arc A770


python/llm/src/bigdl/llm/transformers/qlora.py
```diff
@@ -279,7 +280,9 @@ def prepare_model_for_kbit_training(model, use_gradient_checkpointing=True):
         # cast all non INT8 parameters to fp32
         for param in model.parameters():
             if (param.dtype == torch.float16) or (param.dtype == torch.bfloat16):
-                param.data = param.data.to(torch.float32)
+                # don't cast linear here to avoid lora OOM on arc
+                if param.dim() == 1:
```
jason-dai, line 284: what does `param.dim() == 1` mean?

rnwang04: For `LowbitLinear`, their dtype is torch.uint8 and dim is 1.
For normal `Linear` / `Embedding`, their dtype is torch.bfloat16/torch.float32 and dim is 2.
For `lateynorm`, their dtype is torch.bfloat16/torch.float32 and dim is 1.

Add `param.dim() == 1` to only cast `lateynorm` to fp32 here, otherwise original code will cast `BF16Linear` to fp32, and lora on arc will OOM at this place.

jason-dai: > For `LowbitLinear`, their dtype is `torch.uint8` and dim is 1. For normal `Linear` / `Embedding`, their dtype is `torch.bfloat16`/`torch.float32`/`torch.float16` and dim is 2. For `layernorm`, their dtype is `torch.bfloat16`/`torch.float32`/`torch.float16` and dim is 1.
> 
> Add `param.dim() == 1` to only cast `layernorm` to fp32 here, otherwise original code will cast `BF16Linear` to fp32, and lora on arc will OOM at this place.

This is too implementation specific and can be very fragile. Just check if this is a linear layer.

rnwang04: > 
> This is too implementation specific and can be very fragile. Just check if this is a linear layer.

Based on `model.parameters()`, we just got every parameter of model, we don't know the parameter from which module so seems can't check if this param is from a linear layer.
Otherwise, we change to like `for module in model.named_modules():`, but it is an iterator over all modules:
```python
model.layers.31.self_attn LlamaAttention(
  (q_proj): LowBitLinear(in_features=4096, out_features=4096, bias=False)
  (k_proj): LowBitLinear(in_features=4096, out_features=4096, bias=False)
  (v_proj): LowBitLinear(in_features=4096, out_features=4096, bias=False)
  (o_proj): LowBitLinear(in_features=4096, out_features=4096, bias=False)
  (rotary_emb): LlamaRotaryEmbedding()
)
model.layers.31.self_attn.q_proj LowBitLinear(in_features=4096, out_features=4096, bias=False)
model.layers.31.self_attn.k_proj LowBitLinear(in_features=4096, out_features=4096, bias=False)
model.layers.31.self_attn.v_proj LowBitLinear(in_features=4096, out_features=4096, bias=False)
model.layers.31.self_attn.o_proj LowBitLinear(in_features=4096, out_features=4096, bias=False)
model.layers.31.self_attn.rotary_emb LlamaRotaryEmbedding()
```
I am still looking an elegant way for this.

rnwang04: Do you think change
```python
        for param in model.parameters():
            if (param.dtype == torch.float16) or (param.dtype == torch.bfloat16):
                # don't cast linear here to avoid lora OOM on arc
                if param.dim() == 1:
                    param.data = param.data.to(torch.float32)
```
to below way :
```python
         for module in model.modules():
            if list(module.children()) == []:
                # leaf module
                if not isinstance(module, (torch.nn.Linear, LowBitLinear, torch.nn.Embedding)):
                    for param in module.parameters():
                        if (param.dtype == torch.float16) or (param.dtype == torch.bfloat16):
                            param.data = param.data.to(torch.float32)
```
would be better and acceptable?

jason-dai: Yes; BTW, LowBitLinear is still linear

rnwang04: Thanks for reminder! Have change this part.

python/llm/src/bigdl/llm/transformers/low_bit_linear.py
```diff
@@ -592,13 +592,14 @@ def forward(self, x: torch.Tensor):
...
 class BF16Linear(nn.Linear):
-    def __init__(self, input_features, output_features, bias=True,
+    def __init__(self, input_features, output_features, qtype, bias=True,
                  mp_group=None, compute_dtype=None):
         super().__init__(input_features, output_features, bias)
         self.in_len = input_features
         self.out_len = output_features
         self.weight_shape = (self.out_len, self.in_len)
         self.weight_length = self.out_len * self.in_len
+        self.qtype = qtype
```
jason-dai, line 602: self.qtype=ggml_tensor_qtype["bf16"]

python/llm/src/bigdl/llm/transformers/convert.py
```diff
@@ -364,7 +365,10 @@ def ggml_convert_low_bit(model, qtype, optimize_model=True,
             "an issue on github if you think this is a bug."
         )
     elif device == "cpu":
-        if not (getattr(model, "quantization_method", None) == "gptq"):
+        if not (getattr(model, "quantization_method", None) == "gptq") and \
+            qtype not in [ggml_tensor_qtype["bf16"],
+                          ggml_tensor_qtype["fp16"]]:
```
jason-dai, line 370: This can be potentially problematic if running on CPU (which may not support fp16/bf16)

python/llm/src/bigdl/llm/transformers/models/llama.py
```diff
@@ -138,7 +138,7 @@ def llama_attention_forward_4_31(
         fsdp_flag = check_flash_attention_available(hidden_states)
     else:
         fsdp_flag = False
-    if fsdp_flag and q_len > 1:
+    if fsdp_flag and q_len > 1 and self.q_proj.qtype != ggml_tensor_qtype["bf16"]:
```
jason-dai, line 141: we should have `BF16Linear` to deal with the input type instead

python/llm/src/bigdl/llm/transformers/low_bit_linear.py
rnwang04: Done.

python/llm/src/bigdl/llm/transformers/models/llama.py
rnwang04: Yes,  will update related logic in `BF16Linear`.
But at this code place, if we don't add some limitation, here `attention_dtype` will be set to fp16 for first token on pvc.
Even if we can cast fp16 into bf16 inside `BF16Linear`, I feel it's strange to cast q,k,v Tensor from bf16 to fp16 then use flash attention, and then cast fp16 to bf16 inside `o_proj`.
I perfer here just keep `attention_dtype` as bfloat16.
How about change
`if fsdp_flag and q_len > 1 and self.q_proj.qtype != ggml_tensor_qtype["bf16"]:`
to
`if fsdp_flag and q_len > 1 and original_dtype in [torch.float32, torch.float16]:`

rnwang04:
Current forward logic of `BF16Linear` is only supported for GPU and only cast input to bf16 when training:
```python
    def forward(self, x: torch.Tensor):
        # only work for GPU now
        invalidInputError(x.device.type == "xpu",
                          "bf16 only works for GPU now")
        is_training = self.training and not torch.is_inference_mode_enabled()
        if is_training:
            # below logic is only for training
            autocast_dtype = get_autocast_dtype(x)
            if self.compute_dtype is not None and x.device.type == "xpu":
                x = x.to(self.compute_dtype)  # solve GC issue for unlora module
            elif autocast_dtype is not None:
                x = x.to(autocast_dtype)

        if self.bias is not None and self.bias.dtype != x.dtype:
            self.bias.data = self.bias.data.to(x.dtype)
        result = F.linear(x, self.weight)
        if self.bias is not None:
            result += self.bias
```
Based on before discussion, shall I change it to below way (cast input to bf16 always, add don't check GPU esp.) ? @jason-dai 
```python
    def forward(self, x: torch.Tensor):
        x = x.to(torch.bfloat16)
        if self.weight is not None and self.weight.dtype != x.dtype:
            self.weight.data = self.weight.data.to(x.dtype)
        if self.bias is not None and self.bias.dtype != x.dtype:
            self.bias.data = self.bias.data.to(x.dtype)
        
        result = F.linear(x, self.weight)
        if self.bias is not None:
            result += self.bias
        return result.to(x.dtype)
```

python/llm/src/bigdl/llm/transformers/convert.py
rnwang04: Yes, but on CPU which not support fp16/bf16, I guess users just should not use `torch_dtype=bf16/fp16` or `load_in_low_bit=bf16/fp16`, otherwise, they will meet similar error or perf drop.
Do you think we should handle this case (maybe add warning or raise error in forward of `BF16Linear`) ? Not sure whether this make sense and can really capture errors.

python/llm/src/bigdl/llm/transformers/convert.py
```diff
@@ -364,8 +366,11 @@ def ggml_convert_low_bit(model, qtype, optimize_model=True,
             "an issue on github if you think this is a bug."
         )
     elif device == "cpu":
-        if not (getattr(model, "quantization_method", None) == "gptq"):
-            model.to(torch.float32)
+        if not (getattr(model, "quantization_method", None) == "gptq") and torch_dtype == "auto":
+            convert_bigdl_other_module(model, torch.float32)
+        elif not (getattr(model, "quantization_method", None) == "gptq") and torch_dtype != "auto":
+            # if user specify torch dtype, cast module to dtype
+            convert_bigdl_other_module(model, torch_dtype)
```
jason-dai, line 373:
```python
if not (getattr(model, "quantization_method", None) == "gptq")
    if torch_dtype == "auto":
       convert_bigdl_other_module(model, torch.float32)
    elif:
       convert_bigdl_other_module(model, torch_dtype) 
```

python/llm/src/bigdl/llm/transformers/low_bit_linear.py
```diff
@@ -592,34 +592,25 @@ def forward(self, x: torch.Tensor):
 
 
 class BF16Linear(nn.Linear):
-    def __init__(self, input_features, output_features, bias=True,
+    def __init__(self, input_features, output_features, qtype, bias=True,
```
jason-dai, line 595: no need for `qtype` here?

jason-dai:
See comments above; otherwise LGTM

MeouSker77 (approved):
LGTM

python/llm/src/bigdl/llm/transformers/convert.py
rnwang04: Have changed to this way.

python/llm/src/bigdl/llm/transformers/low_bit_linear.py
rnwang04: Have remove all unnecessary qtype passing, only keep a default qtype of `BF16Linear`.

PR #9758, opened 2023-12-22
vLLM: Apply attention optimizations for selective batching

gc-fu:
## Description

Re-apply attention optimizations from `llama_attention_forward_4_31` to `llama_attention_selective_batching_forward_4_31`.

The following three optimizations are included:
1. fuse_rope
2. cache for kv_cache
3. decoding_fast_path

Noted that for this `cache for kv_cache` optimizations, we need to create the cache for kv_cache in the first decoding path.  Otherwise, the result will be a mess for selective batching.

### 1. Why the change?

To boost the performance.

### 2. User API changes

None

### 3. Summary of the change 

Re-apply attention optimizations from `llama_attention_forward_4_31` to `llama_attention_selective_batching_forward_4_31`.

The following three optimizations are included:
1. fuse_rope
2. cache for kv_cache
3. decoding_fast_path

Noted that for this `cache for kv_cache` optimizations, we need to create the cache for kv_cache in the first decoding path.  Otherwise, the result will be a mess for selective batching.

### 4. How to test?
- [ ] Manually tested

### 5. Possibly TODO:
not batching prefill requests at all... so we can get rid of the attention_masks... We can have a try on this.


xiangyuT (approved):
LGTM

python/llm/src/bigdl/llm/transformers/models/llama.py
```diff
@@ -334,108 +334,141 @@ def llama_attention_selective_batching_forward_4_31(
...
-        key_states = self.k_proj(hidden_states)
-        value_states = self.v_proj(hidden_states)
-
-    query_states = query_states.view(bsz, q_len,
-                                     self.num_heads, self.head_dim).transpose(1, 2)
-    key_states = key_states.view(bsz, q_len,
-                                 self.num_key_value_heads, self.head_dim).transpose(1, 2)
-    value_states = value_states.view(bsz, q_len,
+        if self.config.pretraining_tp > 1:
+            invalidInputError(False, f"vLLM: config.pretraining_tp > 1 not supported yet")
```
jason-dai, line 380: why the error message is `vLLM` specific?

gc-fu: We have not yet consider the case where `self.config.pretraining_tp` greater than one.  I think the original `llama_attention_forward_4_31` handles this case correctly.  So this error message should be `vLLM` specific.

Issue #9707, opened 2023-12-18
Qwen 14B serving failed with BigDL LLM fastchat


qzheng527:
Trying to use bigdl LLM serving for model qwen 7b following the [doc](https://github.com/intel-analytics/BigDL/tree/main/python/llm/src/bigdl/llm/serving). 
1. Modify the model name from **Qwen-7B-Chat** to **bigdl-7b**.
2. Install python packages by conda, python 3.9.
```
# Install BigDL LLM
$script_dir/python-occlum/bin/pip install torch==2.1.0 --index-url https://download.pytorch.org/whl/cpu
$script_dir/python-occlum/bin/pip install --pre --upgrade bigdl-llm[all] bigdl-llm[serving]
$script_dir/python-occlum/bin/pip install einops transformers_stream_generator
```
2. Try below commands but failed.
```
./python-occlum/bin/python -m fastchat.serve.controller --host 0.0.0.0
./python-occlum/bin/python -m bigdl.llm.serving.model_worker --model-path /work/models/chatglm2-6b --device cpu --host 0.0.0.0
```

![image](https://github.com/intel-analytics/BigDL/assets/23373702/08400dd4-1524-4c8c-87e7-93520f496a4f)

Trying chatglm2 and vicuna 7b, both works fine.

qiyuangong:
Hi @qzheng527 

Fold name will not impact load, but if some files can not be found loader will raise exceptions. According to error message, loader cannot find `tokenization_qwen.py` or related files from fold. Please check files in qwen folder.

BTW, this error maybe caused by Qwen's file out of data. Qwen just changed their files (not weights) last week. Please check if these files are the same with latest Qwen files.

https://huggingface.co/Qwen/Qwen-7B-Chat/blob/main/modeling_qwen.py
https://huggingface.co/Qwen/Qwen-7B-Chat/blob/main/tokenizer_config.json
https://huggingface.co/Qwen/Qwen-7B-Chat/blob/main/generation_config.json
https://huggingface.co/Qwen/Qwen-7B-Chat/blob/main/config.json

To avoid loading exception on latest Qwen model, we updated bigdl qwen loader, but it may raise exception on legacy files.

qzheng527:
@qiyuangong The `tokenization_qwen.py` is in the model folder and the model can work fine on general chat test.
The model was released on Oct 2023.
I will update it and try again.

qzheng527:
[...]

PR #9762, opened 2023-12-22
[LLM] Add trust_remote_code for local renamed model in bigdl_llm_model.py

Jasonzzt:
## Description

Realated issue: https://github.com/intel-analytics/BigDL/issues/9707
After renaming the model, there's an error `Tokenizer class QWenTokenizer does not exist or is not currently imported when loading`.

### 1. Why the change?

Add trust_remote_code=True for local renamed model in the method from_pretrained.

### 2. User API changes

serving API

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [x] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/python/llm/src/bigdl/llm/serving/bigdl_llm_model.py
+++ b/python/llm/src/bigdl/llm/serving/bigdl_llm_model.py
@@ -243,7 +243,7 @@ class BigDLLLMAdapter(BaseModelAdapter):
     def load_model(self, model_path: str, from_pretrained_kwargs: dict):
         revision = from_pretrained_kwargs.get("revision", "main")
         tokenizer = AutoTokenizer.from_pretrained(
-            model_path, use_fast=False, revision=revision
+            model_path, use_fast=False, revision=revision, trust_remote_code=True
         )
         print("Customized bigdl-llm loader")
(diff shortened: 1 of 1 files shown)


python/llm/src/bigdl/llm/serving/bigdl_llm_model.py
```diff
@@ -243,7 +243,7 @@ def match(self, model_path: str):
     def load_model(self, model_path: str, from_pretrained_kwargs: dict):
         revision = from_pretrained_kwargs.get("revision", "main")
         tokenizer = AutoTokenizer.from_pretrained(
-            model_path, use_fast=False, revision=revision
+            model_path, use_fast=False, revision=revision, trust_remote_code=True
```
qiyuangong, line 246: Check if we can get "trust_remote_code" from from_pretrained_kwargs

qiyuangong:
Nice PR.
Please check if other loading requires `trust_remote_code=True`. 

Just like https://github.com/lm-sys/FastChat/blob/main/fastchat/model/model_adapter.py#L94

Meanwhile, we need to change the readme to avoid customers changing their model names.

qiyuangong (approved):
LGTM

PR #9769, opened 2023-12-25
[LLM] IPEX import support bigdl-core-xe-21

qiyuangong:
## Description

Add support for bigdl-core-xe-21.

### 1. Why the change?

The distribution lib used in auto ipex importer only detects bigdl-core-xe.

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [x] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/python/llm/src/bigdl/llm/utils/ipex_importer.py
+++ b/python/llm/src/bigdl/llm/utils/ipex_importer.py
@@ -35,10 +35,19 @@ class IPEXImporter:
         """
         # Check if xpu version installed
         try:
+            # Check if bigdl-core-xe is installed
             distribution('bigdl-core-xe')
             return True
         except PackageNotFoundError:
-            return False
+            # bigdl-core-xe not found
+            # Check if bigdl-core-xe-21 is installed
+            try:
+                distribution('bigdl-core-xe-21')
+                return True
+            except PackageNotFoundError:
+                # bigdl-core-xe not found
+                return False
+        return False
 
     def import_ipex(self):
         """


python/llm/src/bigdl/llm/utils/ipex_importer.py
```diff
@@ -35,10 +35,19 @@ def is_xpu_version_installed():
...
+            # Check if bigdl-core-xe is installed
             distribution('bigdl-core-xe')
             return True
         except PackageNotFoundError:
-            return False
+            # bigdl-core-xe not found
+            # Check if bigdl-core-xe-21 is installed
+            try:
+                distribution('bigdl-core-xe-21')
+                return True
```
jason-dai, line 46: Can we use `importlib.util.find_spec` instead?

Jasonzzt: `importlib.util.find_spec` is not useful here.
I tested it on arc14 with the following code
```
        if importlib.util.find_spec('bigdl-core-xe') is not None:
            return True
        else:
            print("intel_extension_for_pytorch not found")
            return False
```
`importlib.util.find_spec` doesn't find ipex
```
(ziteng-test) arda@arda-arc14:~/anaconda3/envs/ziteng-test/lib/python3.9/site-packages/bigdl/llm/utils$ python ipex_importer.py
intel_extension_for_pytorch not found
```

Oscilloscope98 (approved):
LGTM

PR #9780, opened 2023-12-25
LLM: fix wrong batch output caused by flash attention

rnwang04:
### 1. Why the change?

https://github.com/analytics-zoo/nano/issues/888

### 2. User API changes

No change.

### 3. Summary of the change 

- only use flash attention when batch_size = 1, as when batch_size > 1, flash attention seems have precision issue.

### 4. How to test?
- [x] Unit test
- [x] Local test

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/models/llama.py
+++ b/python/llm/src/bigdl/llm/transformers/models/llama.py
@@ -522,9 +522,10 @@ def check_flash_attention_available(query):
     if query.dtype not in [torch.float32, torch.float16]:
         # only use flash attention for fp32/fp16 input
         return False
-    if query.shape[0] > 1:
-        # only use flash attention for batch_size = 1
-        # as we found some precision problem with batch_size > 1
+    if query.shape[0] > 1 or query.size()[0] > 1:
+        # only use flash attention for batch_size = 1 now
+        # as flash attention doesn't support attn_mask in ipex 2.1,
+        # so it will cause output error for padded batch input
         return False
     return True
 
--- a/python/llm/src/bigdl/llm/transformers/models/llama.py
+++ b/python/llm/src/bigdl/llm/transformers/models/llama.py
@@ -137,7 +137,7 @@ def llama_attention_forward_4_31(
     # for flash attention
     original_dtype = hidden_states.dtype
     if not self.training and not hidden_states.requires_grad:
-        fsdp_flag = check_flash_attention_available(hidden_states)
+        fsdp_flag = use_flash_attention(hidden_states)
     else:
         fsdp_flag = False
     if fsdp_flag:
@@ -324,7 +324,7 @@ def llama_attention_selective_batching_forward_4_31(
     original_dtype = hidden_states.dtype
     # TODO: consider this later - flash attention
     # if not self.training and not hidden_states.requires_grad:
-    #     fsdp_flag = check_flash_attention_available(hidden_states)
+    #     fsdp_flag = use_flash_attention(hidden_states)
     # else:
     #     fsdp_flag = False
     # if fsdp_flag and q_len > 1:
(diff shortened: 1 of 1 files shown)


python/llm/src/bigdl/llm/transformers/models/llama.py
```diff
@@ -522,6 +522,10 @@ def check_flash_attention_available(query):
     if query.dtype not in [torch.float32, torch.float16]:
         # only use flash attention for fp32/fp16 input
         return False
+    if query.shape[0] > 1:
+        # only use flash attention for batch_size = 1
+        # as we found some precision problem with batch_size > 1
+        return False
     return True
```
jason-dai, line 529: Maybe change to:
```python
def use_flash_attention(query):
   if query.size()>1 || query.shape[0] > 1:
       return false;
  ...

```

rnwang04: Done.

jason-dai: Move it to the beginning of the method

jason-dai (approved):
See comments above; otherwise LGTM

python/llm/src/bigdl/llm/transformers/models/llama.py
rnwang04: Not quite understanding about "remove change q_len>1 in the forward method"?

jason-dai: You are already checking query length here, and [q_len>1](https://github.com/intel-analytics/BigDL/blob/main/python/llm/src/bigdl/llm/transformers/models/llama.py#L143) is no longer need in the forward method

rnwang04: I don't check query length here, `if query.shape[0] > 1 or query.size()[0] > 1:` only check batch size of query (query shape is [B, S, H] here).
But I can move `q_len > 1` into check_flash_attention_available(query) .

jason-dai: `bsz, q_len, _ = hidden_states.size()`, yes?

rnwang04: Yes.

python/llm/src/bigdl/llm/transformers/models/llama.py
```diff
@@ -507,7 +506,17 @@ def llama_attention_selective_batching_forward_4_31(
 
 
 def check_flash_attention_available(query):
```
jason-dai, line 508: change to `use_flash_attention`?

zhentaocc (approved):

python/llm/src/bigdl/llm/transformers/models/llama.py
rnwang04: Done, have changed to `use_flash_attention`.

python/llm/src/bigdl/llm/transformers/models/llama.py
```diff
@@ -506,8 +505,18 @@ def llama_attention_selective_batching_forward_4_31(
...
-def check_flash_attention_available(query):
+def use_flash_attention(query):
+    bsz, q_len, _ = query.size()
     # check whether ipex flash attention can be used
+    if bsz > 1:
+        # only use flash attention for batch_size = 1 now
+        # as flash attention doesn't support attn_mask in ipex 2.1,
+        # so it will cause output error for padded batch input
+        return False
+    if q_len == 1:
```
yangw1234, line 516: It seems flash attention can benefits the fp16 case even for the rest token when I experiment it with AutoTP.

rnwang04: Thanks for reminder! I will take some further more test about flash attention.

PR #9781, opened 2023-12-25
bigdl-llm stress test for stable version

liu-shaojun:
### 1. Why the change?

This submission adds stress test (500 different prompts for 1k-512, 2k-512) for stable version.


Oscilloscope98 (approved):
LGTM

PR #9784, opened 2023-12-26
[LLM] Migrate iGPU perf tests to new machine

Oscilloscope98:
## Description

- Migrate igpu perf test to new machine
- Enable all models for 1024-128 tests
- Move 1024-128 tests just after 32-32 as it is more important than others (32-256, 32-512, 512-64)
- Make sure python output encoding in utf-8 so that redirect to output txt can always be success
- Upload results to FTP server instead


MeouSker77 (approved):
LGTM

PR #9785, opened 2023-12-26
Arc stable version performance regression test

Ricky-Ting:
### 1. Why the change?
This submission lets the arc perf test fail if the difference between perf and baseline is greater than 5%

Also, empty gpu mem (in transformer_int4_gpu) between different models' benchmarks.


python/llm/test/benchmark/stable-version-arc-perf-test-fp8.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - '32-32'
+  - '1024-128'
+  - '2048-256'
+test_api:
+  - "transformer_int4_gpu"  # on Intel GPU
+cpu_embedding: False # whether put embedding to CPU (only avaiable now for gpu win related test_api)
+exclude:
+  - 'fnlp/moss-moon-003-sft:1024'
+  - 'fnlp/moss-moon-003-sft:2048'
+  - 'bigscience/bloomz-7b1:2048'
```
hkvision, line 22: remove these?

.github/workflows/llm_performance_tests_stable_version.yml
```diff
@@ -69,87 +69,92 @@ jobs:
...
-
-      - name: Download llm binary
-        uses: ./.github/actions/llm/download-llm-binary
-
-      - name: Run LLM install (all) test
-        uses: ./.github/actions/llm/setup-llm-env
+          python csv_to_html.py -f $CSV_SAVE_PATH/int4 -b $CSV_SAVE_PATH/int4/transformer_int4_gpu-results-1baseline.csv -t 5.0
 
-      - name: Test on cpu
+      - name: Test on xpu (fp8)
```
hkvision, line 89: Do we need to put each precision as a separate job? Why not running the bash for multiple times in a single job?

hkvision: Can leave this for now, seems quite clear already.

hkvision (approved):

PR #9786, opened 2023-12-26
[LLM] Correct prompt format of Yi, Llama2 and Qwen in generate.py

Jasonzzt:
## Description

The prompt format for model Yi and model Llama2 in generate.py is wrong, which result in worse output.
Yi's prompt format refers to https://huggingface.co/01-ai/Yi-6B-Chat#31-use-the-chat-model
Llama2's prompt format refers to https://huggingface.co/TheBloke/Llama-2-70B-Chat-GGML#prompt-template-llama-2-chat
Qwen's prompt format refers tohttps://huggingface.co/Qwen/Qwen-1_8B-Chat/blob/main/generation_config.json#L2

The prompt format of Llama2 in GPU example and Qwen in CPU example are correct, so only correct prompt format of Llama2 in CPU example, prompt format of Yi in CPU/GPU examples and prompt format of Qwen in GPU example.

### 1. Why the change?

prompt format for model Yi, Llama2 and Qwen in generate.py

### 2. User API changes

Model example

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [x] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/python/llm/example/CPU/HF-Transformers-AutoModels/Model/llama2/generate.py
+++ b/python/llm/example/CPU/HF-Transformers-AutoModels/Model/llama2/generate.py
@@ -22,11 +22,12 @@ from bigdl.llm.transformers import AutoModelForCausalLM
 from transformers import LlamaTokenizer
 
 # you could tune the prompt based on your own model,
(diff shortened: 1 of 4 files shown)


qiyuangong (approved):
LGTM

PR #9791, opened 2023-12-27
LLM: better FP16 support for Intel GPUs

rnwang04:
### 1. Why the change?

https://github.com/analytics-zoo/nano/issues/890#issuecomment-1869437620
https://github.com/analytics-zoo/nano/issues/890#issuecomment-1869920497

### 2. User API changes

- Optimized fp16 usage

```python
model = AutoModelForCausalLM.from_pretrained(model_path, 
                                             torch_dtype=torch.float16,
                                             load_in_low_bit="fp16", 
                                             trust_remote_code=True,
                                             use_cache=True)
model = model.to('xpu')
```
- You can also do save & load (**must under same ipex version**), although it is maybe unnecessary ：)
```
model.save_low_bit(dir_path)
model = AutoModelForCausalLM.load_low_bit((dir_path)
```
### 3. Summary of the change 

- better FP16 support for Intel GPUs

### 4. How to test?
- [x] Unit test
- [x] Local test of llama2-7b on pvc (ipex 2.1)
- [x] Local test of llama2-7b on pvc (ipex 2.0)
- [x] Local test of llama2-7b on arc (ipex 2.1)
- [x] Local test of llama2-7b on arc (ipex 2.0)
- [x] Local test of mistral-7b on pvc (ipex 2.1)
- [x] Local test of mistral-7b on pvc (ipex 2.0)
- [x] Local test of mistral-7b on arc (ipex 2.1)
- [x] Local test of mistral-7b on arc (ipex 2.0)
- [x] save & load verification on pvc
- [x] save & load verification on arc

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/convert.py
+++ b/python/llm/src/bigdl/llm/transformers/convert.py
@@ -201,7 +201,7 @@ def _replace_with_low_bit_linear(model, qtype, modules_to_not_convert=None,
                             mp_group=mp_group,
                         )
                         device = module.qweight.data.device
-                        invalidInputError(device != "meta",
+                        invalidInputError(device.type != "meta",
                                           "converting from meta device is not supported")
                         # Copy the weights
                         paramsLowBit = FP4Params(data=convert_gptq(module, awq=is_awq),
--- a/python/llm/src/bigdl/llm/transformers/convert.py
+++ b/python/llm/src/bigdl/llm/transformers/convert.py
@@ -236,31 +236,26 @@ def _replace_with_low_bit_linear(model, qtype, modules_to_not_convert=None,
                             new_linear._parameters['bias'] = nn.Parameter(module.bias.data)\
                                 .to(device_type)
                     elif qtype == ggml_tensor_qtype["fp16"]:
-                        #  only support two size now
-                        #  may generalize to other sizes
-                        if module.in_features in [4096, 11008]:
-                            # esimd fp16 path
-                            new_linear = FP16Linear(
-                                in_features,
-                                out_features,
-                                qtype,
-                                module.bias is not None,
-                                mp_group=mp_group,
-                            )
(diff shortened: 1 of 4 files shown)


rnwang04:
After further confirmation, current code can normally save & load **under same ipex version**.
If not, there will be error.
Do you think we can fix this issue in later PR or you prefer we fix it in this PR directly ? @jason-dai

rnwang04:
Performance verification is update here: https://github.com/analytics-zoo/nano/issues/890#issuecomment-1870136847

jason-dai:
> After further confirmation, current code can normally save & load **under same ipex version**. If not, there will be error. Do you think we can fix this issue in later PR or you prefer we fix it in this PR directly ? @jason-dai

I think this is expected, as long as the user understand what the error is

python/llm/src/bigdl/llm/transformers/convert.py
```diff
@@ -236,31 +236,26 @@ def _replace_with_low_bit_linear(model, qtype, modules_to_not_convert=None,
...
-                                new_linear._parameters['bias'] = nn.Parameter(module.bias.data)\
-                                    .to(device_type)
+                        module.to(torch.float16)
+                        new_linear = FP16Linear(
+                            in_features,
+                            out_features,
+                            module.bias is not None,
+                            mp_group=mp_group,
+                        )
+                        device_type = module.weight.data.device.type
```
MeouSker77, line 246: may change all device type to device, there is no reason to use device type instead of device here

MeouSker77 (approved):
others LGTM

python/llm/src/bigdl/llm/transformers/convert.py
rnwang04: Done

rnwang04:
Plan to merge this PR first for further test, if there is any comment for this PR, please let me know, and I can fix it in next PR.

PR #9797, opened 2023-12-27
Add cpu and gpu examples of Mamba

yzheng124:
### 1. Why the change?

In this PR, we add cpu and gpu examples of Mamba using bigdl-llm's `optimize_model` API.

https://github.com/analytics-zoo/nano/issues/880


shane-huang:
need to clarify how to use `model.py`

shane-huang (approved):
`model.py` is used in `generate.py`.  `generate.py` can directly run without further modifications to Mamba model files. May add a note later to this example README.

PR #9799, opened 2023-12-28
update warm_up and num_trials for stable version cpu stress test

liu-shaojun:
### 1. Why the change?

update warm_up to 3 and num_trials to 50 for stable version cpu stress test


glorysdj (approved):
LGTM

PR #9802, opened 2023-12-28
fix csv_to_html

Ricky-Ting:
### 1. Why the change?

When running llm-perf-regression-test-on-spr, this error occurs:
```log
  File "/actions-runner/_work/BigDL/BigDL/python/llm/test/benchmark/csv_to_html.py", line 122, in main
    print("The diffs are outside the normal range: %" + highlight_threshold)
TypeError: can only concatenate str (not "float") to str
```


liu-shaojun (approved):
LGTM

PR #9805, opened 2023-12-28
Update BigDL-LLM-inference image

gc-fu:
### 1. Why the change?
Bump oneapi version to `2024.0` so that user can choose to use torch 2.1, ipex 2.1


glorysdj (approved):
LGTM

PR #9811, opened 2023-12-29
Try to add baichuan2-13b back to arc perf

hkvision:


hkvision:
In the run_transformers_int4_gpu implementation, 1k input is using 2048.txt as the input file and Baichuan2-13B can generate 98 tokens for this input, which causes OOM [if without empty cache]. (Don't know why it passes once for the PR validation http://10.239.44.136:8888/pr_perf_gpu/transformer_int4_gpu-results-2023-12-27.html)

For 1024.txt, it will generate 3 tokens and can run successfully.

PR #9812, opened 2023-12-29
[LLM] support quantize kv cache to fp8

MeouSker77:
## Description

quantize qwen's kv cache to fp8 to accelerate long context performance, and also fix <https://github.com/analytics-zoo/nano/issues/922>

## Note:

todo in the following PRs:

- [ ] CPU has no fp8 matmul implementation so it will dequantize key cache before matmul for now
- [ ] It only supports single input token in next token stage for now. For those methods which have multiple input tokens in next token stage such as Medusa and parallel-decoding, it will also dequantize key cache before matmul. (But note, it supports accelerating batch forward and beam search. batch forward and beam search are different from multiple input tokens in attention)

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/models/qwen.py
+++ b/python/llm/src/bigdl/llm/transformers/models/qwen.py
@@ -63,6 +63,7 @@ def apply_rotary_pos_emb(t, freqs):
     t_ = (t_ * cos) + (rotate_half(t_) * sin)
     return torch.cat((t_, t_pass_), dim=-1).type_as(t)
 
+
 def qwen_attention_forward(
     self,
     hidden_states: Optional[Tuple[torch.FloatTensor]],
@@ -186,7 +187,7 @@ def qwen_attention_forward(
                 cache_v = new_cache_v
 
             key_states, value_states = append_kv_cache(cache_k, cache_v,
-                                                    key.transpose(1, 2), value.transpose(1, 2))
+                                                       key.transpose(1, 2), value.transpose(1, 2))
             key = key_states
             value = value_states
         elif use_cache:
--- a/python/llm/src/bigdl/llm/transformers/models/qwen.py
+++ b/python/llm/src/bigdl/llm/transformers/models/qwen.py
@@ -77,7 +77,7 @@ def qwen_attention_forward(
     output_attentions: Optional[bool] = False,
     use_cache: Optional[bool] = False,
 ):
-    if quantize_kv_cache(hidden_states):
+    if quantize_kv_cache(self.c_attn, hidden_states):
         qwen_quantized_attention_forward(
             self=self,
             hidden_states=hidden_states,
--- a/python/llm/src/bigdl/llm/transformers/models/qwen.py
+++ b/python/llm/src/bigdl/llm/transformers/models/qwen.py
@@ -78,7 +78,7 @@ def qwen_attention_forward(
     use_cache: Optional[bool] = False,
 ):
     if quantize_kv_cache(self.c_attn, hidden_states):
-        qwen_quantized_attention_forward(
+        return qwen_quantized_attention_forward(
             self=self,
             hidden_states=hidden_states,
             rotary_pos_emb_list=rotary_pos_emb_list,
@@ -91,7 +91,7 @@ def qwen_attention_forward(
             use_cache=use_cache
         )
     else:
-        qwen_original_attention_forward(
+        return qwen_original_attention_forward(
             self=self,
             hidden_states=hidden_states,
             rotary_pos_emb_list=rotary_pos_emb_list,
--- a/python/llm/src/bigdl/llm/transformers/models/qwen.py
+++ b/python/llm/src/bigdl/llm/transformers/models/qwen.py
@@ -39,8 +39,7 @@ except ImportError:
 from bigdl.llm.transformers.models.utils import extend_kv_cache, init_kv_cache, append_kv_cache
 from bigdl.llm.transformers.models.utils import init_fp8_kv_cache, extend_fp8_kv_cache, \
     append_fp8_kv_cache, restore_fp8_kv_cache
-from bigdl.llm.transformers.models.utils import rotate_half
-from bigdl.llm.transformers.utils import quantize_kv_cache
+from bigdl.llm.transformers.models.utils import rotate_half, quantize_kv_cache
 from bigdl.llm.utils.common import invalidInputError, invalidOperationError
 from bigdl.llm.ggml.quantize import ggml_tensor_qtype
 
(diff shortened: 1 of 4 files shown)


leonardozcm (approved):
LGTM

jason-dai:
Will need to run some C-Eval test @hkvision ?

hkvision:
> Will need to run some C-Eval test @hkvision ?

Sure, we will add back the C-Eval test very soon.

jason-dai:
> * [ ]  CPU has no fp8 matmul implementation so it will dequantize key cache before matmul for now

We can separate the CPU path and XPU path, and only apply quantization to the XPU path, yes? Similarly, we may also need a flag (config or ENV) to enable/disable quantization 

> * [ ]  It only supports single input token in next token stage for now. For those methods which have multiple input tokens in next token stage such as Medusa and parallel-decoding, it will also dequantize key cache before matmul. (But note, it supports accelerating batch forward and beam search. batch forward and beam search are different from multiple input tokens in attention)

MeouSker77:
> We can separate the CPU path and XPU path, and only apply quantization to the XPU path, yes? Similarly, we may also need a flag (config or ENV) to enable/disable quantization

Add a parameter named `quantize_kv_cache` to control whether to use this, default to be false

python/llm/src/bigdl/llm/optimize.py
```diff
@@ -208,6 +208,8 @@ def optimize_model(model, low_bit='sym_int4', optimize_llm=True, modules_to_not_
         to `True` when running BigDL-LLM on GPU on Windows. Default to be `False`.
     :param lightweight_bmm: Whether to replace the torch.bmm ops, may need to set it
         to `True` when running BigDL-LLM on GPU on Windows. Default to be `False`.
+    :param quantize_kv_cache: Whether to quantize kv cache, it can be only used on GPU.
+        Default to be `False`.
```
jason-dai, line 212: Instead of having the user to manually specify it, I think we can automatically control its behavior as follows:
```python
def quantize_kv_cache:
    if ENV BIGDL_QUANTIZE_KV is defined
       return BIGDL_QUANTIZE_KV 
    elif it's running on CPU or qtype is FP16/BF16
       return fasle
    elif
       return true

def qwen_attention_forward:
    if quantize_kv_cache()
       qwen_quantized_attention_forward()
    else
       qwen_unquantized_attention_forward()
```

MeouSker77: a small question, if user use `from_pretrained` API, then this convert will happen in a very early stage, at this time, we don't know whether user will put this model to XPU and whether user will convert it to other dtype

MeouSker77: so we can only check these condition inside the `qwen_attention_forward` function, shall we do that?

jason-dai: > so we can only check these condition inside the `qwen_attention_forward` function, shall we do that?

Yes, we need to check inside `qwen_attention_forward`

python/llm/src/bigdl/llm/transformers/utils.py
```diff
@@ -175,3 +175,10 @@ def get_xpu_device_type(x):
         return "pvc"
     else:
         return "others"
+
+
+def quantize_kv_cache(x: torch.Tensor) -> bool:
+    if os.environ.get("BIGDL_QUANTIZE_KV_CACHE", None) is not None:
+        return os.environ["BIGDL_QUANTIZE_KV_CACHE"] == 1
+    else:
+        return x.device.type == 'xpu' and x.dtype != torch.half and x.dtype != torch.bfloat16
```
jason-dai, line 184: How's the dtype of `hidden_states` determined?

MeouSker77: `hidden_states` is the input tensor of this layer,  its dtype is determined by `model.dtype`

jason-dai: > `hidden_states` is the input tensor of this layer, its dtype is determined by `model.dtype`

`model.dtype` is different from `qtype` (i.e., the `low_in_low_bit` parameter); we probably should not quantize here if the user specifies `fp16` or `bf16` as `low_in_low_bit` when loading the model

MeouSker77: changed to qtype

python/llm/src/bigdl/llm/transformers/utils.py
```diff
@@ -182,5 +182,5 @@ def quantize_kv_cache(linear: nn.Module, x: torch.Tensor) -> bool:
     if os.environ.get("BIGDL_QUANTIZE_KV_CACHE", None) is not None:
         return os.environ["BIGDL_QUANTIZE_KV_CACHE"] == 1
     else:
-        return x.device.type == 'xpu' and linear.qtype != ggml_tensor_qtype["fp16"] and \
-            linear.qtype != ggml_tensor_qtype["bf16"]
+        return x.device.type == 'xpu' and hasattr(linear, "qtype") and \
+            linear.qtype != ggml_tensor_qtype["fp16"] and linear.qtype != ggml_tensor_qtype["bf16"]
```
jason-dai, line 186: Does it work? If there are unconverted layers, we will quantize for some layers and not quantize for some other layers.

I think we should record the a static model `qtype` value, instead of checking each layer in a dynamic fashion.

python/llm/src/bigdl/llm/transformers/models/qwen.py
```diff
@@ -73,6 +76,46 @@ def qwen_attention_forward(
     encoder_attention_mask: Optional[torch.FloatTensor] = None,
     output_attentions: Optional[bool] = False,
     use_cache: Optional[bool] = False,
+):
+    if quantize_kv_cache(self.c_attn, hidden_states):
+        qwen_quantized_attention_forward(
```
lalalapotter, line 81: Should use `return qwen_quantized_attention_forward(...)`.

python/llm/src/bigdl/llm/transformers/models/qwen.py
```diff
@@ -73,6 +76,46 @@ def qwen_attention_forward(
...
+            layer_past=layer_past,
+            attention_mask=attention_mask,
+            head_mask=head_mask,
+            encoder_hidden_states=encoder_hidden_states,
+            encoder_attention_mask=encoder_attention_mask,
+            output_attentions=output_attentions,
+            use_cache=use_cache
+        )
+    else:
+        qwen_original_attention_forward(
```
lalalapotter, line 94: Should use `return qwen_original_attention_forward(...)`.

python/llm/src/bigdl/llm/transformers/utils.py
```diff
@@ -175,3 +176,11 @@ def get_xpu_device_type(x):
         return "pvc"
     else:
         return "others"
+
+
+def quantize_kv_cache(linear: nn.Module, x: torch.Tensor) -> bool:
+    if os.environ.get("BIGDL_QUANTIZE_KV_CACHE", None) is not None:
+        return os.environ["BIGDL_QUANTIZE_KV_CACHE"] == 1
```
lalalapotter, line 183: `os.environ["BIGDL_QUANTIZE_KV_CACHE"]` returns a string maybe use `os.environ["BIGDL_QUANTIZE_KV_CACHE"]=='1'`

python/llm/src/bigdl/llm/transformers/utils.py
MeouSker77: we can use a model level `qtype` to avoid check `qtype` dynamically, but we still need to check if is running on XPU dynamically. I think only checking XPU or checking both has no much difference ?

Issue #9768, opened 2023-12-23
RuntimeError: PyTorch is not linked with support for xpu devices


openvino-book:
RuntimeError: PyTorch is not linked with support for xpu devices

Install BigDL GPU version on Windows 11 as https://bigdl.readthedocs.io/en/latest/doc/LLM/Overview/install_gpu.html
![image](https://github.com/intel-analytics/BigDL/assets/128127436/6fb1c829-f716-409e-9eb7-5c5cf9594131)

when execute the code as below, the model is chatglm3-6b

```python
import torch
import time
import argparse
import numpy as np

from bigdl.llm.transformers import AutoModel
from transformers import AutoTokenizer

# you could tune the prompt based on your own model,
# here the prompt tuning refers to https://github.com/THUDM/ChatGLM3/blob/main/PROMPT.md
CHATGLM_V3_PROMPT_FORMAT = "<|user|>\n{prompt}\n<|assistant|>"

if __name__ == '__main__':
    parser = argparse.ArgumentParser(description='Predict Tokens using `generate()` API for ChatGLM3 model')
    parser.add_argument('--repo-id-or-model-path', type=str, default="d:/chatglm3-6b",
                        help='The huggingface repo id for the ChatGLM3 model to be downloaded'
                             ', or the path to the huggingface checkpoint folder')
    parser.add_argument('--prompt', type=str, default="AI是什么？",
                        help='Prompt to infer')
    parser.add_argument('--n-predict', type=int, default=32,
                        help='Max tokens to predict')

    args = parser.parse_args()
    model_path = args.repo_id_or_model_path

    # Load model in 4 bit,
    # which convert the relevant layers in the model into INT4 format
    model = AutoModel.from_pretrained(model_path,
                                      load_in_4bit=True,
                                      trust_remote_code=True)
    
    model.save_low_bit("bigdl_chatglm3-6b-q4_0.bin")
    #run the optimized model on Intel GPU
    model = model.to('xpu')

    # Load tokenizer
    tokenizer = AutoTokenizer.from_pretrained(model_path,
                                              trust_remote_code=True)
    
    # Generate predicted tokens
    with torch.inference_mode():
        prompt = CHATGLM_V3_PROMPT_FORMAT.format(prompt=args.prompt)
        input_ids = tokenizer.encode(prompt, return_tensors="pt")
        st = time.time()
        # if your selected model is capable of utilizing previous key/value attentions
[...]

PR #9820, opened 2024-01-03
LLM: fix chatglm3 issue

rnwang04:
### 1. Why the change?

https://github.com/analytics-zoo/nano/issues/905
https://github.com/intel-analytics/BigDL/issues/9768#issuecomment-1869890853

### 2. User API changes

No change.

### 3. Summary of the change 

- code refactor
- add usage limitation for `torch.nn.functional.scaled_dot_product_attention`, only `use_flash_attention` is True then use this function.

### 4. How to test?
- [ ] Unit test
- [x] Local test for chatglm2 / llama2-7b on arc
- [x] Local test for chatglm2 / llama2-7b on pvc

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/models/utils.py
+++ b/python/llm/src/bigdl/llm/transformers/models/utils.py
@@ -16,6 +16,7 @@
 
 import torch
 from bigdl.llm.utils.common import invalidInputError
+from bigdl.llm.transformers.utils import get_ipex_version
 
 
 def init_kv_cache(batch_size, num_heads, head_dim, current_length, max_length, dtype, device):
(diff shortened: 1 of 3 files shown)


python/llm/src/bigdl/llm/transformers/models/utils.py
```diff
@@ -119,3 +120,61 @@ def is_enough_kv_cache_room_4_31(past_key_value):
...
+        return False
+    ipex_version = get_ipex_version()
+    if ipex_version <= "2.0.110+xpu":
+        # ipex flash attention is supported from ipex 2.1
+        return False
+    if not torch.xpu.has_xetla():
+        # ipex flash attention is only supported for xetla
+        # may update this later
+        return False
+    if query.dtype not in [torch.float32, torch.float16]:
```
jason-dai, line 151: How's `query.dtype` determined?

rnwang04: It's determined by output from the previous layer .
In most cases, `query.dtype` is torch.float32.
Unless user pass `torch_dtype=fp16/bf16` or uses call `model.half().to('xpu')`, then here query.dtype will be fp16 or bf16.

MeouSker77 (approved):

PR #9825, opened 2024-01-03
LLM: Temporarily disable Test_Optimize_Gpu_Model for arc ut

Mingyu-Wei:
### 1. Why the change?

Disable Test_Optimize_Gpu_Model for falcon/llama/chatglm2. Will enable them again once we fix the failures in arc unit tests


Oscilloscope98:
After UT passed, we may merge for now. Need to investigate on why the failure happened and enable these tests again.

Oscilloscope98 (approved):

PR #9832, opened 2024-01-04
[LLM] IPEX auto importer set on by default

Jasonzzt:
## Description

Set BIGDL_IMPORT_IPEX default as True, i.e., auto import IPEX when bigdl-xpu is installed.
Remove IPEX import from GPU example.

### 1. Why the change?

Performance tests show no impact from auto importer.

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [x] Application test
- [ ] Document test
- [ ] ...


qiyuangong (approved):
LGTM

PR #9834, opened 2024-01-04
[LLM] Fix dtype mismatch in Baichuan2-13b

xiangyuT:
## Description

Add `.to(dtype=value_states.dtype)` to fix dtype mismatch when running Baichuan2-13b with BigDL BF16.


rnwang04 (approved):
LGTM

PR #9835, opened 2024-01-04
LLM: update qlora alpaca example to change lora usage

rnwang04:
### 1. Why the change?

As lora just use bf16, not 4bit here, so update related example to not use bnb config for lora finetuning.

### 2. User API changes

None, usage change see example.

### 3. Summary of the change 

- update qlora alpaca example to change lora usage

### 4. How to test?
- [ ] Unit test
- [x] Local test
- [x] Local perf verification


jason-dai:
LGTM

cyita (approved):

PR #9837, opened 2024-01-04
LLM: arc perf test using ipex2.1

WeiguangHan:
@hkvision Do arc perf test using ipex2.1


Oscilloscope98 (approved):
LGTM. Please revert changes to trigger pr tests

PR #9841, opened 2024-01-05
LLM: fix first token judgement of flash attention

rnwang04:
### 1. Why the change?

Fix flash attention error when past_past_key_values is not None.
Now speculative sampling can works normally.

### 2. User API changes

No change.

### 3. Summary of the change 

- change first token judgement of flash attention to `q_len == k_len` other than `q_len > 1`

### 4. How to test?
- [ ] Unit test
- [x] Local test on pvc for llama2-7b & chatglm3

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/models/llama.py
+++ b/python/llm/src/bigdl/llm/transformers/models/llama.py
@@ -265,7 +265,8 @@ def llama_attention_forward_4_31(
     value_states = repeat_kv(value_states, self.num_key_value_groups).to(device,
                                                                          dtype=attention_dtype)
 
-    if fsdp_flag:
+    if fsdp_flag and query_states.shape[2] == key_states.shape[2]:
+        # by comparing query and key's length, to make sure it's first token
         attn_output = F.scaled_dot_product_attention(query_states.to(dtype=attention_dtype),
                                                      key_states,
                                                      value_states,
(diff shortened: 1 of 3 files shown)


python/llm/src/bigdl/llm/transformers/models/llama.py
```diff
@@ -265,7 +265,8 @@ def llama_attention_forward_4_31(
     value_states = repeat_kv(value_states, self.num_key_value_groups).to(device,
                                                                          dtype=attention_dtype)
 
-    if fsdp_flag:
+    if fsdp_flag and query_states.shape[2] == key_states.shape[2]:
```
jason-dai, line 268: add this check to `use_flash_attention`?

cyita (approved):

python/llm/src/bigdl/llm/transformers/models/llama.py
rnwang04: In our current code, we put `use_flash_attention` at the beginning of `llama_attention_forward_4_31`, usage likes `use_flash_attention(hidden_states)`, which means at this time, it only has hidden_states, don't know query_states / key_states .
If we want to put this check into `use_flash_attention`, maybe we can acheive it by delaying the check to line 261.
Do you prefer this ?

rnwang04: Have refactored this part, would you mind taking a look again?

jason-dai (approved):
LGTM

PR #9849, opened 2024-01-05
Support llm-awq backend

cyita:
## Description

Support llm-awq backend. Based on https://github.com/intel-analytics/BigDL/pull/9833.

Different from autoawq:
1. Transpose zeros, scales, and weight
2. Remove padding

Sample output of https://huggingface.co/ybelkada/vicuna-7b-1.5-awq
```
-------------------- Prompt --------------------
### HUMAN:
What is AI?

### RESPONSE:

-------------------- Output --------------------
### HUMAN:
What is AI?

### RESPONSE:

AI, or artificial intelligence, refers to the development of computer systems that can perform tasks that typically require human intelligence, such as visual perception, speech recognition
```

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

### 5. New dependencies

- [ ] New Python dependencies
       - Dependency1 
       - Dependency2
       - ...
- [ ] New Java/Scala dependencies and their license
       - Dependency1 and license1
       - Dependency2 and license2
       - ...

PR #9856, opened 2024-01-08
Support llm-awq backend

Romanticoseu:
## Description

Support llm-awq backend. Based on #9849 

### 3. Summary of the change 
Support llm-awq backend-->

### 4. How to test?
- [x] Application test

### 5. New dependencies
None

### DIFF
--- a/python/llm/example/CPU/HF-Transformers-AutoModels/Advanced-Quantizations/AWQ/README.md
+++ b/python/llm/example/CPU/HF-Transformers-AutoModels/Advanced-Quantizations/AWQ/README.md
@@ -3,7 +3,7 @@
 This example shows how to directly run 4-bit AWQ models using BigDL-LLM on Intel CPU.
 
 ## Verified Models
-
+### Auto-AWQ Backend
 - [Llama-2-7B-Chat-AWQ](https://huggingface.co/TheBloke/Llama-2-7B-Chat-AWQ)
 - [CodeLlama-7B-AWQ](https://huggingface.co/TheBloke/CodeLlama-7B-AWQ)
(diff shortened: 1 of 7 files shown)


Romanticoseu:
## Application test:
Sample output of https://huggingface.co/ybelkada/vicuna-7b-1.5-awq#9856 
``` bash
### HUMAN:
Write a python code to find Fibonacci

### RESPONSE:

### HUMAN:
Write a python code to find Fibonacci

### RESPONSE:

Here's a Python code to find the Fibonacci sequence:
```python
def fibonacci(n):
    if n <= 1:
        return n
    else:
        return fibonacci(n-1) + fibonacci(n-2)

for i in range(10):
    print(fibonacci(i))
```

python/llm/example/CPU/HF-Transformers-AutoModels/Advanced-Quantizations/AWQ/README.md
```diff
@@ -37,6 +41,17 @@ pip install accelerate==0.24.1
 pip install einops
 ```
 
+**(Optional) Support for LLM-AWQ Backend**
+
+BigDL-LLM uses `autoawq` as its default awq backend, and also automatically supports `llmawq` if your model is orginally quantized by llmawq. You can directly input an llmawq model in the same way as autoawq, after installing the following dependencies:
+
+```bash
+git clone https://github.com/mit-han-lab/llm-awq
```
Uxito-Ada, line 49: Do we really require users to install llm-awq? It seems nothing is imported from llm-awq in codes.

Romanticoseu: You are correct, we have integrated the `llm-awq` code.

cyita (approved):
LGTM

PR #9862, opened 2024-01-09
[LLM] Only use quantize kv cache on MTL

MeouSker77:
## Description

For now, quantize kv cache has some bugs on ARC, and seems it will cause worse performance on ARC, so only use it on MTL


rnwang04 (approved):
LGTM

PR #9867, opened 2024-01-09
Add CPU and GPU examples for DeciLM-7B

Zhangky11:
## Description
In this PR, we add CPU and GPU examples of DeciLM-7B using bigdl-llm's transformer and optimize_model API.
This is the issue page: https://github.com/analytics-zoo/nano/issues/936

### DIFF
--- a/python/llm/example/CPU/HF-Transformers-AutoModels/Model/deciLM-7b/README.md
+++ b/python/llm/example/CPU/HF-Transformers-AutoModels/Model/deciLM-7b/README.md
@@ -1,0 +1,69 @@
+# DeciLM-7B
+In this directory, you will find examples on how you could apply BigDL-LLM INT4 optimizations on DeciLM-7B models. For illustration purposes, we utilize the [Deci/DeciLM-7B-instruct](https://huggingface.co/Deci/DeciLM-7B-instruct) as a reference DeciLM-7B model.
+
+## 0. Requirements
+To run these examples with BigDL-LLM, we have some recommended requirements for your machine, please refer to [here](../README.md#recommended-requirements) for more information.
+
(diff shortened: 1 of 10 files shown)


shane-huang (approved):
Need to update the table in https://github.com/intel-analytics/BigDL/tree/main/python/llm#verified-models.

python/llm/example/CPU/HF-Transformers-AutoModels/Model/deciLM-7b/README.md
```diff
@@ -0,0 +1,69 @@
...
+## Example: Predict Tokens using `generate()` API
+In the example [generate.py](./generate.py), we show a basic use case for a DeciLM-7B model to predict the next N tokens using `generate()` API, with BigDL-LLM INT4 optimizations.
+### 1. Install
+We suggest using conda to manage environment:
+```bash
+conda create -n llm python=3.9
+conda activate llm
+
+pip install --pre --upgrade bigdl-llm[all] # install the latest bigdl-llm nightly build with 'all' option
+pip install transformers==4.35.2 # required by DeciLM-7B
```
shane-huang, line 16: maybe we should bump to transformers 4.36 as mixtral? @jason-dai

Oscilloscope98:

Zhangky11:
> Please also updated the readme for Windows GPU supports as we have done here: #9992
> 
> It would be better to have some verification on Windows GPU.

Sure, I'll do it.

Zhangky11:
I've updated README file and finished verifications on windows GPU.

Zhangky11:
It can work when I set cpu_embedding=True for both HF-transformer API and Pytorch API.

PR #9869, opened 2024-01-09
Add SOLAR-10.7B to README

Alisia0182:
## Description
Add SOLAR-10.7B to README
https://github.com/analytics-zoo/nano/issues/877


shane-huang (commented):
We usually don't include the sizes in the model folder, i.e..  solar-10.7b -> solar. A Model arch may have different sizes and they will be put in the same folder to avoid too many folders.

plusbang (approved):

PR #9876, opened 2024-01-10
Add plot function to all-in-one

Zhangky11:
## Description

Add an additional file "run-plot.py" to run performance test and draw a line graph of memory usage at the same time.

The current code can only be run on Linux. The 'top' tool is used to obtain memory usage in Linux. However, due to different Linux distributions and versions of the top tool, the output format may vary. Therefore, you need to manually check if the memory usage obtained correctly corresponds to the right location in the original output. After running the code, instructions will appear; you just need to follow the instructions.


python/llm/dev/benchmark/all-in-one/run.py
```diff
@@ -930,21 +990,62 @@ def run_transformer_autocast_bf16( repo_id,
...
+                    for in_out in conf['in_out_pairs']:
+                        model_id_input = model + ':' + in_out.split('-')[0]
+                        if model_id_input in excludes:
+                            in_out_pairs.remove(in_out)
+                if conf['plot_memory_vs_time_graph'] == True:
+                    run_model(model, api, in_out_pairs, conf['local_model_hub'], conf['warm_up'], conf['num_trials'], conf['num_beams'],
+                            conf['low_bit'], conf['cpu_embedding'], conf['batch_size'], index=index, plot_memory_vs_time_graph=True)
+                else:
+                    run_model(model, api, in_out_pairs, conf['local_model_hub'], conf['warm_up'], conf['num_trials'], conf['num_beams'],
+                            conf['low_bit'], conf['cpu_embedding'], conf['batch_size'], plot_memory_vs_time_graph=False)
```
leonardozcm, line 1045: better passing a default value here instead of discussion
```python
                    run_model(model, api, in_out_pairs, conf['local_model_hub'], conf['warm_up'], conf['num_trials'], conf['num_beams'],
                            conf['low_bit'], conf['cpu_embedding'], conf['batch_size'], index=index, plot_memory_vs_time_graph=conf['plot_memory_vs_time_graph'])
```

python/llm/dev/benchmark/all-in-one/run.py
```diff
@@ -930,21 +990,62 @@ def run_transformer_autocast_bf16( repo_id,
...
         df = pd.DataFrame(results, columns=['model', '1st token avg latency (ms)', '2+ avg latency (ms/token)', 'encoder time (ms)',
                                             'input/output tokens', 'actual input/output tokens', 'num_beams', 'low_bit', 'cpu_embedding', 
                                             'peak mem (GB)'])
 
-        df.to_csv(f'{current_dir}/{api}-results-{today}.csv')
+        df.to_csv(f'{current_dir}/output/{api}-results-{today}.csv')
```
leonardozcm, line 1050: hi @Oscilloscope98 , this will change the location of our nightly test results right?

python/llm/dev/benchmark/all-in-one/run.py
```diff
@@ -930,21 +990,62 @@ def run_transformer_autocast_bf16( repo_id,
...
-                      conf['low_bit'], conf['cpu_embedding'], conf['batch_size'])
+            in_out_pairs_ = conf['in_out_pairs'].copy()
+            for in_out_pairs in in_out_pairs_:
+                in_out_pairs = [in_out_pairs]
+                if excludes:
+                    for in_out in conf['in_out_pairs']:
+                        model_id_input = model + ':' + in_out.split('-')[0]
+                        if model_id_input in excludes:
+                            in_out_pairs.remove(in_out)
+                if conf['plot_memory_vs_time_graph'] == True:
```
leonardozcm, line 1040: conf.get('plot_memory_vs_time_graph', False)

In case there is someone's config.yaml being staled.

python/llm/dev/benchmark/all-in-one/run.py
```diff
@@ -33,6 +34,10 @@
 from benchmark_util import BenchmarkWrapper
 from bigdl.llm.utils.common.log4Error import invalidInputError
 
+import subprocess
+import threading
```
leonardozcm, line 38: duplicated with line 24

python/llm/dev/benchmark/all-in-one/run.py
```diff
@@ -16,6 +16,7 @@
 
 
 # this code is copied from llama2 example test, and added performance test
+import argparse
```
leonardozcm, line 19: where is `argparse` used?

python/llm/dev/benchmark/all-in-one/run.py
```diff
@@ -930,21 +990,62 @@ def run_transformer_autocast_bf16( repo_id,
     today = date.today()
     if 'exclude' in conf:
         excludes = conf['exclude']
-    
+    if conf['plot_memory_vs_time_graph'] == True:
+        process = subprocess.Popen(['top', '-b', '-n', '1'], stdout=subprocess.PIPE, text=True)
```
Oscilloscope98, line 994: This may not work on Windows

python/llm/dev/benchmark/all-in-one/run.py
Oscilloscope98: Yes. And this may cause some problems

python/llm/dev/benchmark/all-in-one/run.py
Zhangky11: Yes, it only works on Linux now. And I mention it in README file and set the defalut parameter to False.

PR #9879, opened 2024-01-10
add mistral and chatglm support to vllm

Zhengjin-Wang:


python/llm/src/bigdl/llm/vllm/model_executor/model_loader.py
```diff
@@ -61,8 +61,9 @@
     # "GPTNeoXForCausalLM": GPTNeoXForCausalLM,
     # "InternLMForCausalLM": InternLMForCausalLM,
     "LlamaForCausalLM": BigDLLlamaForCausalLM,
+    "ChatGLMModel": BigDLChatGLMForCausalLM,
```
xiangyuT, line 64: not imported?

xiangyuT (approved):
LGTM

PR #9888, opened 2024-01-11
[LLM] support chatglm3 with bigdl-bf16

Jasonzzt:
## Description

During benchmarking BigDL BF16 on SPR, observed high first token latency (>3s) for 1024-128 request on Chatglm3-6B model and worse latency (>60s)
Comparing the two trace diagrams, found that the implementation of bigdl's BF16Linear is different from that of pytorch.

### 1. Why the change?

Modify method F.linear's params

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [x] Application test
- [ ] Document test
- [ ] ...


rnwang04 (approved):
LGTM

PR #9890, opened 2024-01-11
[LLM] Add bigdl-bf16 into all-in-one

Jasonzzt:
## Description

For benchmarking BigDL BF16 on SPR

### 1. Why the change?

Add test-api `bigdl-bf16` into all-in-one
Modify config and readme

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [x] Application test
- [x] Document test
- [ ] ...

### DIFF
--- a/python/llm/dev/benchmark/all-in-one/config.yaml
+++ b/python/llm/dev/benchmark/all-in-one/config.yaml
@@ -16,6 +16,7 @@ test_api:
   - "native_int4"
   - "optimize_model"
   - "pytorch_autocast_bf16"
+  # - "bigdl_bf16"
   # - "ipex_fp16_gpu" # on Intel GPU
   # - "transformer_int4_gpu"  # on Intel GPU
   # - "optimize_model_gpu"  # on Intel GPU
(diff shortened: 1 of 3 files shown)


python/llm/dev/benchmark/all-in-one/config.yaml
```diff
@@ -16,6 +16,7 @@ test_api:
   - "native_int4"
   - "optimize_model"
   - "pytorch_autocast_bf16"
+  # - "bigdl_bf16"
```
qiyuangong, line 19: maybe change to `"transformer_bf16`

python/llm/dev/benchmark/all-in-one/run.py
```diff
@@ -759,6 +761,71 @@ def run_transformer_int4_gpu_win(repo_id,
...
+        model = AutoModelForCausalLM.from_pretrained(model_path, load_in_low_bit='bf16', trust_remote_code=True, torch_dtype=torch.bfloat16,
+                                                     use_cache=True).eval()
+        tokenizer = AutoTokenizer.from_pretrained(model_path, trust_remote_code=True)
+    end = time.perf_counter()
+    print(">> loading of model costs {}s".format(end - st))
+
+    model = BenchmarkWrapper(model)
+
+    result = {}
+    with torch.inference_mode():
```
qiyuangong, line 795: Maybe we should use autocast cpu bf16. Because some ops only support FP32

qiyuangong (approved):
LGTM

PR #9896, opened 2024-01-12
change xmx condition

MeouSker77:
## Description

change to latest xmx condition: pvc use xmx when 1 < input_length <=8, arc/flex use xmx when dtype=fp32 and 1 < input_length <=64 or dtype=others and 1 < input_length <= 8


lalalapotter (approved):
LGTM

PR #9906, opened 2024-01-16
[LLM] Add pv for llm-serving k8s deployment

Romanticoseu:
## Description
Increase `pv` to address the issue of accessing model files across nodes.


glorysdj (approved):
LGTM

PR #9907, opened 2024-01-16
move reserved memory to benchmark_utils.py

qiuxin2012:
## Description

1. move reserved memory to benchmark_utils.py
2. Empty cache after each generation

### DIFF
--- a/python/llm/dev/benchmark/all-in-one/run.py
+++ b/python/llm/dev/benchmark/all-in-one/run.py
@@ -742,7 +742,6 @@ def run_transformer_int4_gpu_win(repo_id,
                     print("model generate cost: " + str(end - st))
                     output = tokenizer.batch_decode(output_ids)
                     print(output[0])
-                    torch.xpu.empty_cache()
                     actual_out_len = output_ids.shape[1] - actual_in_len
                     if i >= warm_up:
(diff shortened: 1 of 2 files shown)


python/llm/dev/benchmark/all-in-one/run.py
```diff
@@ -743,16 +738,15 @@ def run_transformer_int4_gpu_win(repo_id,
                                                 num_beams=num_beams)
                     torch.xpu.synchronize()
                     end = time.perf_counter()
-                    reserved_mem_list.append(torch.xpu.memory.memory_reserved()/(1024**3))
-                    gpu_peak_mem = max(reserved_mem_list) # always keep the peak gpu mem at current stage
                     output_ids = output_ids.cpu()
                     print("model generate cost: " + str(end - st))
                     output = tokenizer.batch_decode(output_ids)
                     print(output[0])
+                    torch.xpu.empty_cache()
```
Oscilloscope98, line 745: `torch.xpu.empty_cache()` Previously we have some tests and found that this may make first token slower on iGPU

qiuxin2012: It's real, removed. But on A770, the empty_cache doesn't affect the performance.

Oscilloscope98 (approved):
LGTM

PR #9909, opened 2024-01-16
copy fused rms norm's reuslt to avoid <unk>

qiuxin2012:
## Description

Copy fused rms norm's result to avoid <unk> by unknown reason on Arc GPUs.
It looks like some conflicts with xmx linear forward.

### 1. Why the change?

Fixed https://github.com/analytics-zoo/nano/issues/955


MeouSker77 (approved):
LGTM

PR #9913, opened 2024-01-16
GGUF load memory optimization

liu-shaojun:
### 1. Why the change?

This submission optimizes the memory of loading llama gguf model through converting q4_0 to fp16 to int4 layer by layer, in this way the peak memory of loading llama2-7b gguf model decreased from 13G to 5.8G.


glorysdj (approved):
LGTM

MeouSker77 (approved):
LGTM

PR #9916, opened 2024-01-17
Update BigDL-LLM-XPU image

liu-shaojun:
### 1. Why the change?

Install related pypi packages into docker image.


glorysdj (approved):
LGTM

PR #9924, opened 2024-01-17
[LLM]Solve the problem of calling bmm operator in BF16Linear

Jasonzzt:
## Description

During benchmarking BigDL BF16 on SPR, observed high first token latency (>60s) for 1024-128 request on Chatglm3-6B model.
Comparing the two trace diagrams, found that the implementation of bigdl's BF16Linear is different from that of pytorch.

Since the shape of context_layer is 3D, the nn.functional.linear method calls the bmm operator, resulting in a decrease in first token performance.

### 1. Why the change?

When the shape of context_layer is greater than 2D, reshape its shape to 2D

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [x] Application test
- [ ] Document test
- [ ] ...


rnwang04 (approved):
LGTM, please merge it after you can pass style check and ut test.

PR #9926, opened 2024-01-17
LLM: Fix rope of chatglm3 to support speculative decoding on CPU

rnwang04:
### 1. Why the change?

To support speculative decoding for chatglm3 on CPU.

### 2. User API changes

No change.

### 4. How to test?
- [x] Unit test
- [x] Local test


qiyuangong (approved):
LGTM

PR #9932, opened 2024-01-18
LLM: add mlp & qkv fusion for FP16 Llama-7B

rnwang04:
### 1. Why the change?

To further accelerate fp16 performance on PVC.

### 2. User API changes

No change.

### 3. Summary of the change 

- add possible FP16 kernel fusion into llama

### 4. How to test?
- [x] Unit test
- [x] Local test on PVC with transformers==4.31.0, llama2-7b FP16 / INT4
- [x] Local test on PVC with transformers==4.31.0, llama2-13b FP16
- [x] Local test on PVC with transformers==4.36.0, llama2-7b FP16 / INT4
- [x] Local test on PVC with transformers==4.36.0, llama2-13b FP16

Performance verification is update here: https://github.com/analytics-zoo/nano/issues/890#issuecomment-1898496735

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/models/utils.py
+++ b/python/llm/src/bigdl/llm/transformers/models/utils.py
@@ -312,3 +312,19 @@ def use_fused_layer_norm(x: torch.Tensor, training: bool):
             or x.numel() // x.size(-1) == 1
         )
     )
+
+
+def fp16_fusion_check(proj, x, training):
+    # only use fp16 fusion on PVC inference
+    if proj.qtype != ggml_tensor_qtype["fp16"]:
+        return False
+    if proj.weight_type != 2:
+        return False
+    if training:
+        return False
+    if x.requires_grad:
+        return False
+    device_type = get_xpu_device_type(x)
+    if device_type != "pvc":
+        return False
+    return True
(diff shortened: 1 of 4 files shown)


python/llm/src/bigdl/llm/transformers/models/utils.py
```diff
@@ -295,3 +295,19 @@ def use_fused_layer_norm(x: torch.Tensor, training: bool):
...
+    # only use fp16 fusion on PVC inference
+    if proj.qtype != ggml_tensor_qtype["fp16"]:
+        return False
+    if proj.weight_type != 2:
+        return False
+    if training:
+        return False
+    if x.requires_grad:
+        return False
+    device_type = get_xpu_device_type(x)
```
MeouSker77, line 310: may check device type is 'xpu' first, if calling this function with a 'cpu' tensor, it will throw an error. or change this function and make it return 'cpu' when the tensor is not on 'xpu'

MeouSker77 (approved):
others LGTM

python/llm/src/bigdl/llm/transformers/models/utils.py
rnwang04: fix it

PR #9935, opened 2024-01-18
update transformer int4 xpu ut

qiuxin2012:
## Description

update transformer int4 xpu ut

### 1. Why the change?

test_optimize_model's assert is always true.


MeouSker77 (approved):

Oscilloscope98:
We may need wait for the UT results :)

PR #9936, opened 2024-01-18
Add trust_remote_code=True for run_optimize_model in run.py

NovTi:
## Description

Add `trust_remote_code=True` in `BigDL/python/llm/dev/benchmark/all-in-one/run.py` `AutoModelForCausalLM.from_pretrained` at [line 309](https://github.com/intel-analytics/BigDL/blob/main/python/llm/dev/benchmark/all-in-one/run.py#L309) and `AutoTokenizer.from_pretrained` at [line 311](https://github.com/intel-analytics/BigDL/blob/main/python/llm/dev/benchmark/all-in-one/run.py#L311C26-L311C26)


qiuxin2012:
Do you test other models？

PR #9939, opened 2024-01-19
GGUF load memory optimization for Mixtral

liu-shaojun:
### 1. Why the change?

This submission optimizes the memory of loading Mixtral gguf model through converting q4_0 to fp16 to int4 layer by layer.


glorysdj (approved):
LGTM

PR #9945, opened 2024-01-19
LLM: fix outlier value

WeiguangHan:
@hkvision Fix outlier value such as rest token latency is 0.0.

### DIFF
--- a/python/llm/test/benchmark/csv_to_html.py
+++ b/python/llm/test/benchmark/csv_to_html.py
@@ -30,6 +30,10 @@ def highlight_vals(val, max=3.0, color1='red', color2='green'):
     else:
         return ''
 
+def first_nonzero_min(lst):
+    non_zero_lst = [num for num in lst if num != 0.0]
+    return min(non_zero_lst) if non_zero_lst else None
+
 def is_diffs_within_normal_range(diff1, diff2, threshold=5.0):
     return not any(diff < (-threshold) for diff in diff1 + diff2 if isinstance(diff, float))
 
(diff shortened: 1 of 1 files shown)


python/llm/test/benchmark/csv_to_html.py
```diff
@@ -30,6 +30,10 @@ def highlight_vals(val, max=3.0, color1='red', color2='green'):
     else:
         return ''
 
+def first_nonzero_min(lst):
```
hkvision, line 33: just nonzero_min?

python/llm/test/benchmark/csv_to_html.py
```diff
@@ -30,6 +30,10 @@ def highlight_vals(val, max=3.0, color1='red', color2='green'):
...
+    non_zero_lst = [num for num in lst if num != 0.0]
```
hkvision, line 34: if num <= 0.0?

python/llm/test/benchmark/csv_to_html.py
```diff
@@ -40,8 +44,8 @@ def add_to_dict(dict, key, value):
 
 def best_in_dict(dict, key, value):
     if key in dict:
-        best_value=min(dict[key])
-        if best_value < value:
+        best_value = first_nonzero_min(dict[key])
+        if best_value < value or value == 0.0:
```
hkvision, line 48: value <=0.0?

python/llm/test/benchmark/csv_to_html.py
WeiguangHan: ok

python/llm/test/benchmark/csv_to_html.py
WeiguangHan: ok

python/llm/test/benchmark/csv_to_html.py
WeiguangHan: ok

hkvision (approved):

PR #9949, opened 2024-01-22
fix error for benchmark_util.py running on cpu

qiuxin2012:
## Description

fix error for benchmark_util.py running on cpu

### 1. Why the change?

MicrosoftTeams-image (2)


hzjane (approved):
LGTM

PR #9964, opened 2024-01-23
LLM: Convert draft_model kv_cache from bf16 to fp32

hzjane:
## Description
To solve the issue of draft_model slower, maybe We have three ways to test.
1. Convert draft_model kv_cache from bf16 to fp32 (every iter, when draft_step== 0, draft_model generate time will become  slower causing the average draft_time more slower)
2. (This way) Convert the same kv_cache from bf16 to fp32(only first init), and then the target_model will use fp32 and generate fp32 cache. (The speed of qwen's target_model has not changed, but llama has become slower)
3. Draft_model maintains kv_cache itself, and append new matched kv_cache (Maybe the accept rate will become lower，need to test)
On way1 and way3, the draft_time on step == 0 always more slower than normal_time (need to find the reason)

The reason is the draft_kv_cache convert from bf16 (malloc new memory) will call extend_kv_cache when  draft_step== 0, So we can init the needed kv_cache and reuse it each iter.

With the same chinest promt:
spr01 | Llama-2-13b-chat-hf | chatglm3-6b| Qwen-14B
-- | -- | -- | -- |
before draft_time | 74.752 | 34.252 | 61.053
before target_time | 200.961 | 101.003| 186.477
after draft_time | 64.221 | 29.139 | 53.270
after target_time | 201.692 | 102.653| 188.057 |

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [x] Application test
- [ ] Document test
- [ ] ...

### 5. New dependencies

- [ ] New Python dependencies
       - Dependency1 
       - Dependency2
       - ...
       - Dependency1 and license1
       - Dependency2 and license2
       - ...

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/speculative.py
+++ b/python/llm/src/bigdl/llm/transformers/speculative.py
@@ -370,30 +370,30 @@ def speculative_generate(self,
                          v[:, :, :-(max_of_max_matched - max_matched)]) for k, v in past_key_values
                     ]
 
-                # Each iter assign new_matched kv_cache to past_key_values1
-                if self.device.type == 'cpu':
-                    for i in range(len(past_key_values)):
-                        if self.config.model_type == "qwen":
-                            size = tmp_past_key_values[i][0].size(1)
-                            size1 = past_key_values[i][0].size(1)
-                            past_key_values1[i][0][:, size:size1, :, :] = (
-                                past_key_values[i][0][:, size:size1, :, :].to(torch.float32))
-                            past_key_values1[i][1][:, size:size1, :, :] = (
-                                past_key_values[i][1][:, size:size1, :, :].to(torch.float32))
-                        elif self.config.model_type == "chatglm":
-                            size = tmp_past_key_values[i][0].size(0)
-                            size1 = past_key_values[i][0].size(0)
-                            past_key_values1[i][0][size:size1, :, :, :] = (
-                                past_key_values[i][0][size:size1, :, :].to(torch.float32))
-                            past_key_values1[i][1][size:size1, :, :, :] = (
-                                past_key_values[i][1][size:size1, :, :].to(torch.float32))
-                        else:
-                            size = tmp_past_key_values[i][0].size(2)
-                            size1 = past_key_values[i][0].size(2)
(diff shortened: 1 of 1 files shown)


python/llm/src/bigdl/llm/transformers/speculative.py
```diff
@@ -195,6 +195,10 @@ def speculative_generate(self,
             # Draft model auto-regressively generate k tokens
             # Early stop when prob less then th_stop_draft
             for step_draft in range(max_step_draft):
+                if self.device.type == 'cpu':
+                    past_key_values = [
```
Uxito-Ada, line 199: Is this repeated every step, which may cause reductant latency and memory copy?

Uxito-Ada: could it be done once e.g. when init, and only process the additive part in the following?

jason-dai: > could it be done once e.g. when init, and only process the additive part in the following?

Yes, we should try that.

jason-dai: Is this implemented in the same when during our previous testing?

glorysdj (approved):
LGTM

python/llm/src/bigdl/llm/transformers/speculative.py
```diff
@@ -189,7 +190,71 @@ def speculative_generate(self,
...
+                                        dtype=torch.float32)
+                        v0 = torch.ones(len0, len1, len2 + max_new_tokens, len3,
+                                        dtype=torch.float32)
+                        past_key_values1.append((k0, v0))
+                        past_key_values1[i][0][:, :, :len2, :] = past_key_values[i][0].to(
+                            torch.float32)
+                        past_key_values1[i][1][:, :, :len2, :] = past_key_values[i][1].to(
+                            torch.float32)
+
+            # each iter cut off cur_len kv_cache from past_key_values1
```
cyita, line 236: Maybe move these two lines to `if self.device.type == 'cpu':`?

cyita:
This PR won't affect the GPU performance.

PR #9970, opened 2024-01-23
fix qwen fp16 model generate a lot of !!!!

qiuxin2012:
## Description

fix qwen fp16 model generate a lot of !!!

### 1. Why the change?

sin and cos are floats, we need to convert them to query's type.

before:
```
Once upon a time, there was a young girl named Samantha who lived with her parents in a small town. Samantha had always dreamed of traveling the world and experiencing!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```
After this PR:
```
Once upon a time, there was a young girl named Samantha who lived with her parents in a small town. Samantha had always dreamed of traveling the world and experiencing new cultures, but her parents were hesitant to let her go on such an adventure.

One day, Samantha stumbled upon a mysterious map that led to a hidden treasure
```


rnwang04 (approved):
LGTM

PR #9976, opened 2024-01-23
LLM: add mistral speculative example

rnwang04:
### 1. Why the change?

add speculative GPU example of mistral.

### 4. How to test?
- [ ] Unit test
- [ ] Local test


python/llm/example/GPU/Speculative-Decoding/mistral/speculative.py
```diff
@@ -0,0 +1,110 @@
...
+        st = time.perf_counter()
+        output = model.generate(input_ids,
+                                max_new_tokens=args.n_predict,
+                                do_sample=False,
+                                th_stop_draft=0.6)
+        output_str = tokenizer.decode(output[0], skip_special_tokens=True)
+        torch.xpu.synchronize()
+        end = time.perf_counter()
+
+        print(output_str)
```
cyita, line 94: Remove the redundant infos?

cyita (approved):

python/llm/example/GPU/Speculative-Decoding/mistral/speculative.py
rnwang04: done

PR #9982, opened 2024-01-24
fix unexpected keyword argument 'device' 

zhentaocc:
## Description

Fix nightly test failure
[](https://github.com/intel-analytics/BigDL/actions/runs/7626141255/job/20793066834)

This is caused by fixing chatglm3 #9941 , we added `device="meta"` to model init funtion while other models may not accept this argument. So here, I added an if sentence to make it work only for chatglm.
### 1. Why the change?
Fix nightly test failure
[](https://github.com/intel-analytics/BigDL/actions/runs/7626141255/job/20793066834)

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/model.py
+++ b/python/llm/src/bigdl/llm/transformers/model.py
@@ -507,6 +507,11 @@ class _BaseAutoModelClass:
             with ContextManagers(init_contexts):
                 if config.architectures is not None \
                     and config.architectures[0] in ["ChatGLMModel", "ChatGLMForConditionalGeneration"]:
+                    """
+                    ChatGLMModel uses skip_init by default, which will force modules placed on cpu 
+                    if the device is not specified. This will further cause replaced linear allocating
+                    memory on cpu.
+                    """ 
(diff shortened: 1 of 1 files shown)


zhentaocc:
tested on 0118 + this pr changes. results:
image

zhentaocc:
seems that other changes causes memory issue after 0118. need to track the root cause further.

python/llm/src/bigdl/llm/transformers/model.py
```diff
@@ -505,7 +505,9 @@ def load_low_bit(cls,
 
         if bigdl_lcmu_enabled:
             with ContextManagers(init_contexts):
-                kwargs["device"] = "meta"
+                if config.architectures is not None \
```
leonardozcm, line 508: how about adding comments on why chatglm behaves differently? and It is better to make sure the failed nightly test passed on your local machine.

leonardozcm (approved):

zhentaocc:
rerun nightly tests https://github.com/intel-analytics/BigDL/actions/runs/7635883294

PR #9991, opened 2024-01-25
Add Deepseek-6.7B

Zhangky11:
## Description
Add Deepseek-6.7B CPU and GPU examples


Oscilloscope98:

Zhangky11:
I've updated README file and finished verifications on windows GPU.

Zhangky11:
It can work when I set cpu_embedding=True for both HF-transformer API and Pytorch API.

shane-huang:
please resolve conflicts.

shane-huang:
> It can work when I set cpu_embedding=True for both HF-transformer API and Pytorch API.

Does `cpu_embedding=True` has negative impact for Arc? If not, maybe we can add this option in example code (so that iGPU can work well and Arc does not have negative impact)

shane-huang (approved):
LGTM

PR #9996, opened 2024-01-25
Add quantize kv cache support for chaglm2/3

MeouSker77:
## Description

Add quantize kv cache support for chaglm2/3: significant performance improvements when context is long


Oscilloscope98 (approved):
LGTM

PR #9988, opened 2024-01-25
[LLM] Use IPEX Optimization for BF16 Model

xiangyuT:
## Description

Use IPEX Optimization for BF16 Model. Controlled by env `ENABLE_IPEX`.

PR #9997, opened 2024-01-25
[LLM] Use IPEX Optimization for Self Speculative Decoding

xiangyuT:
## Description

Use IPEX Optimization for BF16 Model. Controlled by env ENABLE_IPEX.

Depend on #9988


python/llm/src/bigdl/llm/transformers/model.py
```diff
@@ -248,6 +248,7 @@ def from_pretrained(cls,
             if speculative:
                 from .speculative import speculative_generate, clear_benchmarks
                 # load a sym_int4 model as draft model
+                kwargs["torch_dtype"] = torch.float32
```
hzjane, line 251: This change will affect xpu performance. Have fixed it refer to this [pr](https://github.com/intel-analytics/BigDL/pull/10005)

python/llm/src/bigdl/llm/transformers/speculative.py
```diff
@@ -271,29 +276,59 @@ def speculative_generate(self,
...
+                    draft_past_key_values_storage = torch.zeros(s0, s1, s2 + 256, s3, dtype=tmp_draft_past_key_values.dtype, device=tmp_draft_past_key_values.device)
+                    draft_past_key_values_storage[:, :, :s2, :] = tmp_draft_past_key_values
+                    draft_past_key_values = draft_past_key_values_storage[:, :, :s2, :]
+                    return draft_past_key_values
+
+                tmp_draft_past_key_values = [
+                    (k.to(torch.float32), v.to(torch.float32)) for k, v in tmp_draft_past_key_values
+                ]
+                draft_past_key_values = [
+                    [_expand(pkv[0]), _expand(pkv[1])] for pkv in tmp_draft_past_key_values
```
hzjane, line 297: Each draft_step == 0 always call _expand method will causing preformance decrease. And need to solve conflict with [this pr](https://github.com/intel-analytics/BigDL/pull/9964)

python/llm/src/bigdl/llm/transformers/model.py
jason-dai: why do we need this here？

hzjane: On CPU draft_model need to set `kwargs["torch_dtype"] = torch.float32` to make the query.dtype is the same as key.dtype. Or only can use `(query.to(key.dtype)` to avoid error.

python/llm/src/bigdl/llm/transformers/convert.py
```diff
@@ -560,6 +568,287 @@ def replace_func(m, target_m, func_name, new_func):
...
+
+    next_cache = next_decoder_cache if use_cache else None
+    if not return_dict:
+        return tuple(v for v in [hidden_states, next_cache, all_hidden_states, all_self_attns] if v is not None)  # noqa
+    return BaseModelOutputWithPast(
+        last_hidden_state=hidden_states,
+        past_key_values=next_cache,
+        hidden_states=all_hidden_states,
+        attentions=all_self_attns,
+    )
```
jason-dai, line 832: Move above code to a separate file such as `optimize_ipex.py`

python/llm/src/bigdl/llm/transformers/model.py
jason-dai: > On CPU draft_model need to set `kwargs["torch_dtype"] = torch.float32` to make the query.dtype is the same as key.dtype. Or only can use `(query.to(key.dtype)` to avoid error.

Why does it only affect speculative then?

hzjane: I think the reason is that when only using bigdl-llm int4, user will not set the config `torch_dtype=torch.bfloat16`, but target_model will use it like [this](https://github.com/intel-analytics/BigDL/blob/main/python/llm/example/CPU/Speculative-Decoding/qwen/speculative.py#L65C49-L65C77). And the draft_model is converted from target_model, so it will cause the problem.

hzjane (approved):
LGTM

python/llm/src/bigdl/llm/transformers/speculative.py
```diff
@@ -275,11 +281,36 @@ def speculative_generate(self,
 
             # init draft_self_past_key_values:past_key_values1 and assign initial fp32 value
             if self.device.type == 'cpu' and step == 1:
+                if _enable_ipex:
+                    ipex_past_key_values = []
+                    cur_len = past_key_values[0][0].size(1)
+                    if self.config.model_type == "qwen":
+                        invalidInputError(False, "QWEN with IPEX BF16 is not supported now.")
+                    elif self.config.model_type == "chatglm":
+                        invalidInputError(False, "ChatGLM with IPEX BF16 is not supported now.")
```
jason-dai, line 290: only check what you can currently support here

glorysdj (approved):
LGTM

python/llm/src/bigdl/llm/transformers/speculative.py
```diff
@@ -275,11 +281,36 @@ def speculative_generate(self,
...
+                    if not _enable_ipex:
+                        len0 = past_key_values[i][0].size(0)
+                        len1 = past_key_values[i][0].size(1)
+                        len2 = past_key_values[i][0].size(2)
+                        len3 = past_key_values[i][0].size(3)
+                    else:
+                        if self.config.model_type == "qwen":
+                            invalidInputError(False, "QWEN with IPEX BF16 is not supported now.")
+                        elif self.config.model_type == "chatglm":
+                            invalidInputError(False, "ChatGLM with IPEX BF16 is not supported now.")
```
jason-dai, line 308: only check what you can currently support here

python/llm/src/bigdl/llm/transformers/speculative.py
```diff
@@ -432,55 +505,83 @@ def speculative_generate(self,
...
-                        size = tmp_past_key_values[i][0].size(2)
-                        size1 = past_key_values[i][0].size(2)
-                        past_key_values1[i][0][:, :, size:size1, :] = \
-                            past_key_values[i][0][:, :, size:size1, :].to(torch.float32)
-                        past_key_values1[i][1][:, :, size:size1, :] = \
-                            past_key_values[i][1][:, :, size:size1, :].to(torch.float32)
+                        if self.config.model_type == "qwen":
+                            invalidInputError(False, "QWEN with IPEX BF16 is not supported now.")
+                        elif self.config.model_type == "chatglm":
+                            invalidInputError(False, "ChatGLM with IPEX BF16 is not supported now.")
```
jason-dai, line 572: Only check what you can currently support

python/llm/src/bigdl/llm/transformers/speculative.py
```diff
@@ -388,27 +433,56 @@ def speculative_generate(self,
...
             else:
-                output = self(input_ids=drafted_input_ids,
-                              past_key_values=past_key_values,
-                              return_dict=True,
-                              use_cache=True)
-            logits = output['logits']
+                appended_len = drafted_input_ids.size(1) + step - 1
+                ones_to_append = torch.ones(attention_mask.size(0), appended_len)
+                cur_attention_mask = torch.cat((attention_mask, ones_to_append), dim=1)
+            if hasattr(self, "trace_graph"):
```
jason-dai, line 442: is this for CPU only?

python/llm/src/bigdl/llm/transformers/speculative.py
```diff
@@ -275,11 +281,36 @@ def speculative_generate(self,
 
             # init draft_self_past_key_values:past_key_values1 and assign initial fp32 value
             if self.device.type == 'cpu' and step == 1:
```
jason-dai, line 283: please refactor CPU only logic into utility functions

python/llm/src/bigdl/llm/transformers/speculative.py
```diff
@@ -231,6 +231,10 @@ def speculative_generate(self,
 
     _enable_ipex = os.getenv("BIGDL_OPT_IPEX")
     _enable_ipex = (_enable_ipex is not None) and (_enable_ipex.lower() == "true")
+    if _enable_ipex:
+        if not (self.config.model_type == 'baichuan' or 'llama' in self.config.model_type):
+            invalidInputError(False, "BigDL Speculative Decoding with IPEX BF16 only supports \
+                                      Llama and Baichuan2-13b models currently.")
```
jason-dai, line 237: how about baichuan1 and baichuan2-7b?

xiangyuT: Currently baichuan1 and baichuan2-7b are not supported yet.

PR #10001, opened 2024-01-25
[LLM] Add UTs of load_low_bit for transformers-style API

Oscilloscope98:
## Description

- Add UTs of `load_low_bit` for `transformers-style` API.
  - CPU (chatglm2-6b. mistral-7b)
  - GPU (chatglm2-6b, llama-2-7b)
- Reorganize existing CPU UTs

Currently, simple completion test are added.

TODO: 
- add more tests e.g. regarding `optimize_model=True`/`False`
- add more models
- add UTs for `optimize_model` API with `load_low_bit`

### How to test?
- [x] Unit test


python/llm/test/inference/test_transformers_api.py
```diff
@@ -109,5 +111,60 @@ def test_transformers_chatglm_for_causallm(self):
         res = 'Paris' in output_str        
         self.assertTrue(res)
 
+@pytest.mark.parametrize('prompt, answer', [
+    ('What is the capital of France?\n\n', 'Paris')
+    ])
+@pytest.mark.parametrize('Model, Tokenizer, model_path',[
+    (AutoModel, AutoTokenizer, os.environ.get('ORIGINAL_CHATGLM2_6B_PATH')),
```
rnwang04, line 118: shall we also add `load_low_bit` UT for Bloom / Llama ?

rnwang04 (approved):
LGTM as a starting point for `load_low_bit` UTs.

python/llm/test/inference/test_transformers_api.py
Oscilloscope98: Thanks for the comments. We could add them in further PRs.

PR #10003, opened 2024-01-25
Baichuan2 CPU example of speculative decoding

Uxito-Ada:
## Description

support and verify baichuan2 speculative decoding on CPU with bigdl-bf16 target model

### 1. Why the change?

as above

### 2. User API changes

None

### 3. Summary of the change 

None

### 4. How to test?
bigdl-llm enables baichuan2 of both 7B and 13B to run speculative decoding on transformers 4.31.0 and 4.36.2


glorysdj (approved):
LGTM

PR #10006, opened 2024-01-26
[LLM] Avoid KV Cache OOM when seq len is larger than 1

qiyuangong:
## Description

Avoid KV Cache OOM when seq len is larger than 1

### 1. Why the change?

Existing KV cache extend only handle seq=1, when seq > 1, KV cache storage may OOM.

### 2. User API changes

* For llama like kv cache, i.e., [bs, n_head, seq_len, head_dim], use is_enough_kv_cache_room_4_31.
* Other format need to check kv_len.

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [x] Application test
- [ ] Document test
- [ ] ...

### 5. New dependencies

- [ ] New Python dependencies
       - Dependency1 
       - Dependency2
       - ...
       - Dependency1 and license1
       - Dependency2 and license2
       - ...

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/models/utils.py
+++ b/python/llm/src/bigdl/llm/transformers/models/utils.py
@@ -194,17 +194,19 @@ def apply_rotary_pos_emb_cache_freq_xpu(q, k, sin, cos, model_family):
 
 def is_enough_kv_cache_room_4_36(past_key_value, idx, seq_len=1):
     # to determinate if is enough kv cache room in transformers==4.36
+    # seq_len for current seq
     return past_key_value is not None and len(past_key_value.key_cache) > idx and \
         past_key_value.key_cache[idx].stride()[1] > \
         (past_key_value.key_cache[idx].size(2) + seq_len - 1) * \
         past_key_value.key_cache[idx].size(3)
 
 
-def is_enough_kv_cache_room_4_31(past_key_value, seq_len=1):
+def is_enough_kv_cache_room_4_31(past_key_value, seq_dim=2, seq_len=1):
     # to determinate if is enough kv cache room in transformers between 4.31 and 4.35
+    # seq_len for current seq
     return past_key_value is not None and \
         past_key_value[0].stride()[1] > \
-        (past_key_value[0].size(2) + seq_len - 1) * past_key_value[0].size(3)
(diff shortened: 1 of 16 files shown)


cyita:
Please also add support for llama transformers>4.36

python/llm/src/bigdl/llm/transformers/models/utils.py
```diff
@@ -202,6 +204,8 @@ def is_enough_kv_cache_room_4_36(past_key_value, idx, seq_len=1):
 
 def is_enough_kv_cache_room_4_31(past_key_value, seq_len=1):
     # to determinate if is enough kv cache room in transformers between 4.31 and 4.35
+    # seq_len for current seq len
+    # For llama like kv cache, i.e., [bs, n_head, seq_len, head_dim]
     return past_key_value is not None and \
         past_key_value[0].stride()[1] > \
```
MeouSker77, line 210: may change to `stride(1) >= (size(2) + seq_len) * size(3)`

python/llm/src/bigdl/llm/transformers/models/utils.py
```diff
@@ -194,6 +194,8 @@ def apply_rotary_pos_emb_cache_freq_xpu(q, k, sin, cos, model_family):
 
 def is_enough_kv_cache_room_4_36(past_key_value, idx, seq_len=1):
     # to determinate if is enough kv cache room in transformers==4.36
+    # seq_len for current seq len
+    # For llama like kv cache, i.e., [bs, n_head, seq_len, head_dim]
     return past_key_value is not None and len(past_key_value.key_cache) > idx and \
         past_key_value.key_cache[idx].stride()[1] > \
```
MeouSker77, line 200: the same

MeouSker77 (approved):
others LGTM

python/llm/src/bigdl/llm/transformers/models/utils.py
qiyuangong: Yes. That's more readable.

python/llm/src/bigdl/llm/transformers/models/utils.py
qiyuangong: Done

PR #10011, opened 2024-01-26
simplify quantize kv cache api

MeouSker77:
## Description

simplify quantize kv cache api (no performance change)


rnwang04 (approved):
LGTM

PR #10022, opened 2024-01-29
Fix "local variable 'model' referenced before assignment" for GGUF Mixtral

liu-shaojun:
### 1. Why the change?

This submission is to fix the error when loading Mixtral-8x7B-Instruct-v0.1_q4_0.gguf 
`UnboundLocalError: local variable 'model' referenced before assignment`

After this fix, the output is 
image


Uxito-Ada (approved):
lgtm

PR #9992, opened 2024-01-25
LLM: GPU Example Updates for Windows

JinBridger:
Updates all examples under HF-Transformers-AutoModels and PyTorch Model.

PR #10026, opened 2024-01-29
Add InternLM2 and Phixtral example. Add phixtral.py and modify convert.py correspondingly. 

ivy-lv11:
1. Add InternLM2 and Phixtral example. Performance is listed in issue https://github.com/analytics-zoo/nano/issues/981 and https://github.com/analytics-zoo/nano/issues/988, with both CPU and GPU results. 
2. Add phixtral.py and modify convert.py to speed up Phixtral like what is done in mixtral.py. 
3. Modify convert.py to align with the definition of InternLM2 in huggingface. 
4. Update the README.


Oscilloscope98:

PR #10027, opened 2024-01-29
[LLM] Change nightly perf to install from pypi

Oscilloscope98:
## Description

https://github.com/analytics-zoo/nano/issues/1006 change nighty perf to install bigdl-llm from pypi so that the test is completely separated from llm-cpp-build.

### Summary of the change
- Change perf test from installing from source to installing from pypi for bigdl-llm
- Change perf test time to 00:30, on which the nightly bigdl-llm has already been released
- Add an check for the correct bigdl-llm version
- Make sure the date in the result file name is the same as the tested bigdl-llm version

### How to test?
- [x] Perf test for Arc workflow
- [x] Perf test for iGPU workflow

### DIFF
--- a/.github/workflows/llm_performance_tests.yml
+++ b/.github/workflows/llm_performance_tests.yml
@@ -86,6 +86,9 @@ jobs:
       - name: Test on xpu
         shell: bash
         run: |
+          date_for_test_version=$(date -d yesterday +%Y-%m-%d)
+          sed -i "s/date.today()/\"$date_for_test_version\"/g" python/llm/dev/benchmark/all-in-one/run.py
+
           source /opt/intel/oneapi/setvars.sh
           export USE_XETLA=OFF
           export SYCL_PI_LEVEL_ZERO_USE_IMMEDIATE_COMMANDLISTS=1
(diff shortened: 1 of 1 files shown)


.github/workflows/llm_performance_tests.yml
```diff
@@ -59,13 +60,22 @@ jobs:
...
+      #   uses: ./.github/actions/llm/setup-llm-env
+
+      - name: Install BigDL-LLM from Pypi
+        shell: bash
+        run: |
+          pip install --pre --upgrade bigdl-llm[xpu] -f https://developer.intel.com/ipex-whl-stable-xpu
+          test_version_data=`date -d 'yesterday' '+%Y%m%d'`
+          if ! pip show bigdl-llm | grep $test_version_data; then
+            echo "Did not install bigdl-llm with excepted version $test_version_data"
+            exit 1
```
hkvision, line 77: Maybe we put this as the last step of the test to make the job fail? If we fail to release on some day, it is still reasonable to run the version of the previous day?

Oscilloscope98: I think if we change to install from pypi, we may want to test bigdl-llm for each release version. If so, if the release job failed, the performance test of that release should also be meaningless? And it would be more clear for reproducing the performance results.

hkvision (approved):

PR #10032, opened 2024-01-30
add rwkv time shift optimization

MeouSker77:
## Description

add rwkv time shift optimization, optimization rest token performance


rnwang04 (approved):
LGTM

PR #10036, opened 2024-01-30
LLM: Fix ppl test errors

Mingyu-Wei:
1. Fix the argument inconsistency in README
2. Release XPU allocation when running with multiple precisions to avoid OOM errors
3. Use `copy()` to avoid always setting `torch_dtype=float16` in `model_kwargs`

### DIFF
--- a/python/llm/dev/benchmark/perplexity/run.py
+++ b/python/llm/dev/benchmark/perplexity/run.py
@@ -44,7 +44,7 @@ def main():
     additional_model_kwargs = parse_kwargs(args.model_kwargs)
     summary = {}
     for precision in args.precisions:
-        model_kwargs = additional_model_kwargs
+        model_kwargs = additional_model_kwargs.copy()
         if precision in ggml_tensor_qtype.keys():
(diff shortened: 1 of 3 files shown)


python/llm/dev/benchmark/perplexity/run.py
```diff
@@ -44,7 +44,7 @@ def main():
     additional_model_kwargs = parse_kwargs(args.model_kwargs)
     summary = {}
     for precision in args.precisions:
-        model_kwargs = additional_model_kwargs
+        model_kwargs = additional_model_kwargs.copy()
```
hkvision, line 47: does this change impact anything?

Mingyu-Wei: Without `copy()`, when the precisions include `float16` at the beginning, all following precisions will run with`float16` since this line actually modifies the original `additional_model_kwargs`

hkvision: make sense.

PR #10067, opened 2024-02-01
LLM: Support gpt-j in speculative decoding

cyita:
## Description

Support gpt-j in speculative decoding.

### 3. Summary of the change 

1. Fix gpt-j int4 fp16 error (optimize_model=True)
2. Fix position_ids for gpt-j
3. Fix kvcache update for gpt-j
4. GPU example

### 4. How to test?
- [x] local test, transformers 4.31
- [x] local test, transformers 4.36.2


rnwang04 (approved):
LGTM

cyita:
The pure int4 (gpt-j) inference will raise an error when transformers `4.36`. But Spec with Transformers 4.36 works well.

jason-dai:
> The pure int4 (gpt-j) inference will raise an error when transformers `4.36`. But Spec with Transformers 4.36 works well.

Why do we need to use 4.36?

cyita:
> > The pure int4 (gpt-j) inference will raise an error when transformers `4.36`. But Spec with Transformers 4.36 works well.
> 
> Why do we need to use 4.36?

We can use transformers 4.31, just test if we can support 4.36.

PR #10075, opened 2024-02-02
LLM: fix mpt load_low_bit issue

JinBridger:
https://github.com/analytics-zoo/nano/issues/948

Enable `tie_word_embeddings` when save mpt model. Otherwise, mpt that saved with `save_low_bit` cannot run normally.

This may because [MPT model requires tie_word_embeddings=True](https://huggingface.co/mosaicml/mpt-7b-chat/blob/main/modeling_mpt.py#L232) while we [manually save this variable as False when saving low bit models](https://github.com/intel-analytics/BigDL/pull/8977/files).


leonardozcm (approved):
LGTM， the last layer of mpt is a customized normal layer instead of linear which is usually seen in other llm models.

PR #10078, opened 2024-02-04
LLM: make finetuning examples more common for other models

plusbang:
## Description

Make our finetuning examples more common for other models.
Currently we directly use `LlamaTokenizer` in our finetuning examples, then we have to provide users with a new python script when users want to try other model, just like in https://github.com/intel-analytics/BigDL/issues/9883

### 3. Summary of the change 

- Use `AutoTokenizer` instead of `LlamaTokenizer`
- Add `trust_remote_code=True`

### 4. How to test?
- [x] Unit test
- [x] Local test


rnwang04 (approved):
LGTM

PR #10081, opened 2024-02-04
[LLM] Make sure python 310-311 tests only happen for nightly tests

Oscilloscope98:
## Description

Make sure python 310-311 tests only happen for nightly tests


plusbang (approved):
LGTM

PR #10088, opened 2024-02-05
UX: Script to print env info

ivy-lv11:
Run the script to print the environment-related information to debug. 
### Linux
Run the command
```
bash env-check.sh
```
### Windows
Run the command
```
env-check.bat
```
Env-related issues could be found https://github.com/analytics-zoo/nano/issues/990 .


python/llm/scripts/README.md
```diff
@@ -0,0 +1,25 @@
+# README
+
+## Linux
+### CPU
+Use `python linux-env-check.py` to print the environment-related information. 
+```bash
+python linux-env-check.py
+```
+
+### XPU
```
shane-huang, line 10: - use a single script on different OS, detect OS and choose different paths
- Print if hardware (e.g. XPU) is available, and if available, the detailed information. Don't use options '-xpu'

shane-huang: add some examples outputs of the script.

shane-huang: add try catch to ensure when error happens, it will print the error and proceed to print following sections.

shane-huang:
Merge for now and submit another PR to revise the instructions.

shane-huang (approved):

PR #10093, opened 2024-02-05
Unit test on final logits and the logits of the last attention layer

Zhangky11:
## Description
I refactored the test code under ` python/llm/test/inference_gpu/ `.
Related to: https://github.com/analytics-zoo/nano/issues/1012
* ` test_transformers_api.py `: modified. We have separated the testing of logits, keeping only the tests for completion in this file.
* ` test_transformers_api_final_logits.py `: newly added. Final logits before and after optimization are tested here when XMX is disabled. Currently, ` MPT-7B ` and ` Falcon-6B ` have passed the test successfully. 
* ` test_transformers_attention.py `: newly added. The logits of the last attention layer before and after optimization are tested here when XMX is disabled. For those who fails the final logits test, we consider testing the logits of the last attention layer. For those who passes the final logits test, they are most likely to pass this test too. Currently, ` Llama2-7B ` and ` ChatGLM2-6B `  have passed the test successfully.
* ` test_transformer_api_disable_xmx `: deleted. We move the functions in this file to ` test_transformers_api_final_logits.py `.


python/llm/test/inference_gpu/test_transformers_api_attention.py
```diff
@@ -0,0 +1,157 @@
...
+            self.Falcon_7B_gpu_model(Name, Model, Tokenizer, model_path)
+        elif Name == "ChatGLM2-6B":
+            self.Chatglm2_gpu_model(Name, Model, Tokenizer, model_path)
+
+    
+    def MPT_7B_gpu_model(self, Name, Model, Tokenizer, model_path):
+        # currently only need to compare the output of one self-attention layer.
+        layer_norm = "transformer.blocks.31.norm_1"
+        self_attn = "transformer.blocks.31.attn"
+        lower_bound = 1e-3
```
hkvision, line 135: why mpt is that large?

hkvision (approved):

PR #10108, opened 2024-02-06
Update threshold for cpu stable version tests

liu-shaojun:
### 1. Why the change?

This submission updates the threshold from 5% to 10% for CPU stable version tests


glorysdj (approved):
LGTM

PR #10112, opened 2024-02-06
IPEX Speculative Support for Baichuan2 7B

Uxito-Ada:
## Description

Enable IPEX (and transformers >= 4.35.2) speculative on Baichuan2 7B by modifying model files.

**Why need to modify model files?** Because Baichuan2 uses static (hardcoded) methods in modeling.py, which can not be override at runtime, and BigDL does not optimize tokenizer, thus no way to override tokenizer's method. These lead to that we have to make extra modifications to model files.

### 1. Why the change?

To accelerate Baichuan2 7B's speculative decoding

### 2. User API changes

Need to run `convert_baichuan2_7b_with_bigdl_ipex.sh` script before applying IPEX on Baichuan2 7B's speculative decoding. This is not needed for 13B.

### 3. Summary of the change 

as above

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [x] Document test
- [ ] ...


glorysdj (approved):
LGTM

PR #10115, opened 2024-02-07
[LLM] Fix log condition for BIGDL_OPT_IPEX

xiangyuT:
## Description

Only logger.info on cpu and for bf16 model.


glorysdj (approved):
LGTM

PR #10116, opened 2024-02-07
Chronos: fix PSIRT issues

plusbang:
## Description

To fix PSIRT issues:

1 - The problem is in this file:https://github.com/intel-analytics/BigDL/blob/main/python/chronos/src/bigdl/chronos/data/utils/public_dataset.py
It will try to download some data from:http://clusterdata2018pubcn.oss-cn-beijing.aliyuncs.com/machine_usage.tar.gz
Because this is a http not https, so attacker can perform man-in-the-middle attack.

2 - In line 137 will try to extract the tar.gz file:
tar = tarfile.open(file_path, 'r:gz')
The problem is for tar file, python do not check filename in tar contains slash and backslash, so attacker can replace a malicious tar.gz file.

### 2. User API changes

N/A

### 4. How to test?
- [ ] Unit test


glorysdj (approved):
LGTM

PR #10121, opened 2024-02-07
fix OpenSSF Token-Permissions issues

liu-shaojun:
### 1. Why the change?

This submission is to fix [OpenSSF Token-Permissions](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) issues


glorysdj (approved):
LGTM

Issue #10103, opened 2024-02-06
llama2-7b inference with fp16 precision issue about text generation


Fred-cell:
i installed bigdl-llm=2.5.0b20240205, and inference with fp16 precision, the generation has some issue as below:
model = AutoModelForCausalLM.from_pretrained(model_path,
                                                 load_in_low_bit='fp16',
                                                 device_map = "xpu",
                                                 optimize_model=True,
                                                 trust_remote_code=True,
                                                 use_cache=True)
the output is less than torch_dtype=torch.float16:
![image](https://github.com/intel-analytics/BigDL/assets/99886928/af3ba5f8-e306-4d73-8642-5ffb6d114948)

rnwang04:
Based on below code, the output is normal:
```python
from bigdl.llm.transformers import AutoModelForCausalLM
model_path = 'meta/Llama-2-7b-chat-hf'
model = AutoModelForCausalLM.from_pretrained(model_path,
                                                 load_in_low_bit="fp16",
                                                 torch_dtype=torch.float16,
                                                 optimize_model=True,
                                                 trust_remote_code=True,
                                                 use_cache=True)
```

```bash
[INST] <<SYS>>

<</SYS>>

What is AI? [/INST]  Artificial intelligence (AI) refers to the development of computer systems that can perform tasks that typically require human intelligence, such as learning, problem-sol
```

Hi @Fred-cell , what's your model is and could you please test it again based on above usage ?
If you still meet this issue, could you please share detailed code so that we can reproduce it ?

Fred-cell:
please add guide in each model about precision usage.

Fred-cell:
> Llama-2-7b-

I used llama2-7b-chat-hf:
python generate.py --repo-id-or-model-path ~/LLM/llama-2-7b-hf/ --n-predict 512 --prompt "It is done..."

hkvision:
> please add guide in each model about precision usage.

For the non-chat model, you are recommended to remove the prompt template and directly input the sentence.

hkvision:
Let's add default dtype when load_in_low_bit=fp16. @lalalapotter

hkvision:
Synced with @Fred-cell the problem gets resolved after using the chat model.
And we have add better implementation for fp16. Closing this issue.

PR #10124, opened 2024-02-07
LLM: add default torch_dtype for fp16.

lalalapotter:
## Description

This PR is to add default torch_dtype for fp16, and solve issue #10103 .

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/model.py
+++ b/python/llm/src/bigdl/llm/transformers/model.py
@@ -250,7 +250,7 @@ class _BaseAutoModelClass:
                 else:
                     kwargs["torch_dtype"] = torch.float16
             else:
-                kwargs["torch_dtype"] = kwargs.get("torch_dtype", 'auto')
+                kwargs["torch_dtype"] = torch_dtype if torch_dtype else "auto"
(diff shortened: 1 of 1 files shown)


hkvision (approved):

python/llm/src/bigdl/llm/transformers/model.py
```diff
@@ -237,8 +238,19 @@ def from_pretrained(cls,
...
+                if torch_dtype is not None and torch_dtype != torch.float16:
+                    invalidInputError(
+                        False,
+                        f"Please use torch_dtype=torch.float16"
+                        f" when setting load_in_low_bit='fp16'."
+                    )
+                else:
+                    kwargs["torch_dtype"] = torch.float16
+            else:
+                kwargs["torch_dtype"] = kwargs.get("torch_dtype", 'auto')
```
rnwang04, line 253: `torch_dtype` seems has been pop in line 145.
Should use `torch_dtype` instead of `kwargs.get("torch_dtype")` here ?

lalalapotter: Sure.

rnwang04 (approved):
LGTM

PR #10134, opened 2024-02-08
add quantize kv cache support for qwen2

MeouSker77:
## Description

add quantize kv cache support for qwen2 and fix style check

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/kv.py
+++ b/python/llm/src/bigdl/llm/transformers/kv.py
@@ -1,0 +1,39 @@
+import torch
+
+from .models.utils import init_fp8_kv_cache, append_fp8_kv_cache
+from typing import Optional, Dict, Tuple, Any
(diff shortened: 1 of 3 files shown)


python/llm/src/bigdl/llm/transformers/kv.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+import torch
+
+from .models.utils import init_fp8_kv_cache, append_fp8_kv_cache
+from typing import Optional, Dict, Tuple, Any
+from transformers.cache_utils import DynamicCache
+
+
+class DynamicFp8Cache(DynamicCache):
```
MeouSker77, line 25: if we use this 4.36's new kv cache API like this, maybe we can implement normal kv cache without changing any attention code

rnwang04 (approved):
LGTM

PR #10136, opened 2024-02-08
Separate llmcpp build of linux and windows

zhentaocc:
## Description

Linux and windows build of llmcpp sometimes is time-consuming to wait for available runners, and sometimes fails. To accelerate it, separate two platforms and build target platform only to reduce building time a little bit.

### DIFF
--- a/.github/actions/llm/download-llm-binary/action.yml
+++ b/.github/actions/llm/download-llm-binary/action.yml
@@ -11,14 +11,14 @@ runs:
       run: |
         rm -rf python/llm/llm-binary || true
         mkdir -p python/llm/llm-binary
-        mv linux-avx2/* python/llm/llm-binary/
-        mv linux-avx512/* python/llm/llm-binary/
-        mv linux-avxvnni/* python/llm/llm-binary/
-        mv linux-avx/* python/llm/llm-binary/
-        mv linux-amx/* python/llm/llm-binary/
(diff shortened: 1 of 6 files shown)


Oscilloscope98 (commented):
It seems that we also need to change the logic here: https://github.com/intel-analytics/BigDL/blob/main/.github/actions/llm/download-llm-binary/action.yml

.github/actions/llm/download-llm-binary/action.yml
```diff
@@ -11,14 +16,18 @@ runs:
...
-        mv windows-avx/* python/llm/llm-binary/
+        if ${{contains(inputs.platform, 'Linux')}}; then
+          mv linux-avx2/* python/llm/llm-binary/ 
+          mv linux-avx512/* python/llm/llm-binary/
+          mv linux-avxvnni/* python/llm/llm-binary/
+          mv linux-avx/* python/llm/llm-binary/
+          mv linux-amx/* python/llm/llm-binary/
+        fi
+        if ${{contains(inputs.platform, 'Windows')}}; then
+          mv windows-avx2/* python/llm/llm-binary/ || true
```
Oscilloscope98, line 27: Should we remove the `|| true`?

Oscilloscope98 (approved):
LGTM

PR #10154, opened 2024-02-18
LLM: fix device setting during saving optimized model

plusbang:
## Description

During `save_low_bit`, model is directly moved to `cpu`. Therefore, exist `found at least two devices, cpu and xpu` error when running the following code:
```python
model = model.to('xpu')
model.save_low_bit(save_path)

input_ids = tokenizer.encode(prompt, return_tensors="pt").to('xpu')
output = model.generate(input_ids, max_new_tokens=32)
```

### 3. Summary of the change 

- Fix device setting of `save_low_bit` 

### 4. How to test?
- [x] Unit test
- [x] Local test


leonardozcm (approved):
LGTM

PR #10155, opened 2024-02-18
Add DeepSeek-MoE-16B-Chat

ivy-lv11:
Add DeepSeek-MoE-16B-Chat and the performance could be found in the issue https://github.com/analytics-zoo/nano/issues/987.


shane-huang (approved):
LGTM

PR #10163, opened 2024-02-19
enable fused rmsnorm and rope qwen2

qiuxin2012:
## Description

enable fused rmsnorm and rope qwen2


Oscilloscope98 (commented):
Should we also remove the qwen2 here?
image

qiuxin2012:
> Should we also remove the qwen2 here? image

No, it's right here.

Oscilloscope98 (approved):

PR #10167, opened 2024-02-19
[LLM] Add support for `low_low_bit` benchmark on Windows GPU

Oscilloscope98:
## Description

Add support for `low_low_bit` benchmark on Windows GPU:
- add a save.py file for saving models in low bit according to config file
- add a new test api to test for `load_low_bit` with transformers-style API for Windows GPU tests


leonardozcm (approved):

Oscilloscope98:
@jason-dai We will merge this PR for `load_low_bit` benchmark with “load+Inference twice and then recording the results of the second attempt” method for now. And conducted further tests regarding 2048-256, etc.

Please let us know for any comment :)

PR #10170, opened 2024-02-19
LLM : Support embedding quantization (only q2k now)

rnwang04:
### 1. Why the change?

To further support 2bit quantization.

### 4. How to test?
- [ ] Local test
- [ ] Unit test

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/convert.py
+++ b/python/llm/src/bigdl/llm/transformers/convert.py
@@ -191,7 +191,7 @@ def convert_gptq(module, awq=False, llm_awq=False):
 def _replace_with_low_bit_linear(model, qtype, modules_to_not_convert=None,
                                  current_key_name=None, convert_shape_only=False,
                                  cpu_embedding=False, prefix_name='',
-                                 imatrix_data=None):
+                                 imatrix_data=None, embedding_qtype=None):
     from bigdl.llm.transformers.low_bit_linear import LowBitLinear, FP4Params, \
         FP16Linear, BF16Linear
     from bigdl.llm.transformers.embedding import LLMEmbedding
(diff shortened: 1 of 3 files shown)


python/llm/src/bigdl/llm/transformers/utils.py
```diff
@@ -224,37 +224,49 @@ def load_imatrix_data(imatrix_file):
...
-                cur_qtype = ggml_tensor_qtype['sym_int4']
         else:
+            # if no imatrix is available, use fp8 for lm_head
             cur_imatrix = None
-            # custom mixed quantization strategy
-            if cur_module == 'v' or (cur_module == 'down' and int(layer) in [0, 1, 10, 11]) \
-                    or new_module_name == 'lm_head':
-                cur_qtype = ggml_tensor_qtype['sym_int4']
+            if new_module_name == 'lm_head':
+                cur_qtype = ggml_tensor_qtype['fp8']
```
jason-dai, line 256: shall we use int8?

rnwang04: yeah we can use int8, I have verified locally that int8 can produce similar output as fp8.

rnwang04:
Plan to merge this PR first to make 2-bit quantization function (iq2_xxs) complete, if there is any comment, please let me know and I will update in next PR.

rnwang04:
## model size comparison

| model | precision | size | 
| -- | -- | -- |
| Llama-2-7b-chat-hf | iq2_xxs | 1.8G |
| Llama-2-7b-chat-hf | q4_0 | 1.8G | 3.8G | 
| Llama-2-7b-chat-hf | fp16 | 1.8G | 12.6G |

PR #10187, opened 2024-02-21
qwen2 cpu fix

qiuxin2012:
## Description

qwen2 will throw exception when using cpu.
### 1. Why the change?

qwen2 will throw exception when using cpu.


Oscilloscope98 (approved):
LGTM

PR #10194, opened 2024-02-21
Update bigdl-llm-finetune-qlora-xpu Docker Image

liu-shaojun:
### 1. Why the change?

- Bump oneapi version to 2024.0 since the default ipex is 2.1.
- Install deps


Uxito-Ada (approved):
lgtm

PR #10209, opened 2024-02-22
Move harness nightly related files to llm/test folder

zhentaocc:
## Description

some nightly files are not supposed to be in `dev/benchmark` foler.


hkvision:
Also need to modify fp16 path in harness yml?

hkvision (approved):

PR #10224, opened 2024-02-23
add gemma example

qiuxin2012:
## Description

add gemma gpu and cpu example


python/llm/example/CPU/HF-Transformers-AutoModels/Model/gemma/README.md
```diff
@@ -0,0 +1,138 @@
+# Mistral
```
jason-dai, line 1: Gemma

sgwhat (approved):

jason-dai:
Please update the table here: https://github.com/intel-analytics/bigdl?tab=readme-ov-file#verified-models

PR #10225, opened 2024-02-23
[LLM] Enable kv_cache and forward_qkv optimizations for yuan2

sgwhat:
## Description
Add kv_cache and forward_qkv in yuan2 model to improve performance.

### 1. Why the change?

The generation performance of the Yuan2 model on GPU needs to be optimized.

### 2. User API changes
N/A

### 3. Summary of the change 
- [x] kv_cache
- [x] forward_qkv

### 4. How to test?
- [x] Application test
- [ ] Perf on Arc and MTL iGPU


python/llm/src/bigdl/llm/transformers/models/yuan.py
```diff
@@ -62,46 +91,95 @@ def yuan_attention_forward(
...
+            new_key_states[:] = key_states
+            new_value_states[:] = value_states
+            key_states = new_key_states
+            value_states = new_value_states
 
     past_key_value = \
         (key_states, value_states, inference_hidden_states_memory) if use_cache else None
 
     if self.use_flash_attention:
+        from flash_attn import flash_attn_varlen_func as flash_attn_unpadded_func
```
rnwang04, line 182: i think we don't have `flash_attn`. maybe remove it for now or use our flash attention here.

python/llm/src/bigdl/llm/transformers/models/yuan.py
```diff
@@ -20,13 +20,32 @@
...
+
+KV_CACHE_ALLOC_BLOCK_LENGTH = 256
+
+
+def use_decoding_fast_path(q_type, use_fuse_rope, enough_kv_room, bs):
+    return q_type in [SYM_INT4, FP8E5] and \
+        use_fuse_rope and enough_kv_room and bs == 1
+
+
+def should_use_fuse_rope(self, hidden_states, position_ids):
```
rnwang04, line 44: maybe we can put `use_decoding_fast_path` and `should_use_fuse_rope` into `utils` later,.

python/llm/src/bigdl/llm/transformers/models/yuan.py
```diff
@@ -62,46 +91,95 @@ def yuan_attention_forward(
...
+            hidden_states = self.lf_gate(hidden_states, before_hidden_states)
+            query_states = self.q_proj(hidden_states)
+            key_states = self.k_proj(hidden_states)
+            qk_states = torch.cat([query_states, key_states], dim=-1)
+            qk_states = qk_states.view(bsz, q_len,
+                                    self.num_heads,
+                                    int(qk_states.shape[-1]//self.num_heads))
+            (query_states, key_states) = torch.chunk(qk_states, 2, dim=-1)
+            query_states = query_states.transpose(1, 2)
+            key_states = key_states.transpose(1, 2)
```
rnwang04, line 132: these cat / transpose logic seems a little complex... Is it original implementation look likes this ?

rnwang04 (approved):
see comments inline, others LGTM

python/llm/src/bigdl/llm/transformers/models/yuan.py
jason-dai: just remove `if self.use_flash_attention`

sgwhat:
`forward_qkv` causes abnormal output on MTL and will be supported in a subsequent PR.

jason-dai:
> `forward_qkv` causes abnormal output on MTL and will be supported in a subsequent PR.

Maybe the activation of this model is different from others? @qiuxin2012 - please take a look

qiuxin2012:
> > `forward_qkv` causes abnormal output on MTL and will be supported in a subsequent PR.
> 
> Maybe the activation of this model is different from others? @qiuxin2012 - please take a look

Looks like it's caused by rotary_pos_emb's sin and cos. Run the logits comparing test with yuwen, if `apply_rotary_pos_emb_cache_freq_xpu`'s difference is much smaller than `apply_rotary_pos_emb_no_cache_xpu`.

PR #10247, opened 2024-02-26
LLM: add patching function for llm finetuning

plusbang:
## Description

Background: https://github.com/analytics-zoo/nano/issues/1069

### 1. Why the change?

Provide patching function for qlora/lora finetuning examples to run on Intel GPUs easily.

### 2. User API changes

Only need to add the following code at the beginning of their CUDA code of training
```python
from bigdl.llm import llm_patch, llm_unpatch
llm_patch(train=True)
```

### 4. How to test?
- [x] Unit test
- [x] Local test


python/llm/src/bigdl/llm/transformers/model.py
```diff
@@ -210,6 +210,20 @@ def from_pretrained(cls,
...
+            if is_bigdl_patched:
+                # To handle finetuning patch of `load_in_8bit=True` etc
+                if kwargs.get('load_in_8bit', False):
+                    kwargs.pop('load_in_8bit')
+                if patched_training_mode == "lora":
+                    load_in_low_bit = "bf16"
+                elif patched_training_mode == "qalora":
+                    load_in_low_bit = "sym_int4"
+                elif patched_training_mode == "qlora" or patched_training_mode == "relora":
+                    load_in_low_bit = "nf4"
```
jason-dai, line 224: Not needed - only need to support QLoRA or LoRA (based on BnB config)

plusbang: > Not needed - only need to support QLoRA or LoRA (based on BnB config)

I think LoRA fails to use BnB config. Current implementation of considering `patched_training_mode` would better handle LoraConfig setting and model convert.

For example, according to my local experiment, if I add `llm_patch(training_mode='lora')` at the beginning of https://github.com/tloen/alpaca-lora/blob/main/finetune.py, lora finetuning is in progress on GPU. And if I add `llm_patch(training_mode='qlora')` at the beginning, qlora finetuning will be in progress.

jason-dai: > > Not needed - only need to support QLoRA or LoRA (based on BnB config)
> 
> I think LoRA fails to use BnB config. Current implementation of considering `patched_training_mode` would better handle LoraConfig setting and model convert.
> 
> For example, according to my local experiment, if I add `llm_patch(training_mode='lora')` at the beginning of https://github.com/tloen/alpaca-lora/blob/main/finetune.py, lora finetuning is in progress on GPU. And if I add `llm_patch(training_mode='qlora')` at the beginning, qlora finetuning will be in progress.

I think you may use various configs (BnB, load-in-4-bit, etc.) to choose LoRA vs. QLoRA

jason-dai: > > I think you may use various configs (BnB, load-in-4-bit, etc.) to choose LoRA vs. QLoRA
> 
> OK, I'll modify the conditions of choosing training mode. BTW, we only need to consider patch case of LoRA and QLoRA?

Yes - existing peft code does not have qa-lora or relora

plusbang:
Have updated and related validation (https://github.com/analytics-zoo/nano/issues/1069#issuecomment-1975796867)

python/llm/src/bigdl/llm/transformers/llm_patching.py
```diff
@@ -0,0 +1,77 @@
...
+attrs = []
+
+
+def replace_attr(obj, name: str, value):
+    original_attr = getattr(obj, name)
+    setattr(obj, name, value)
+    attrs.append((obj, name, original_attr))
+
+
+def llm_patch():
```
jason-dai, line 32: `llm_patch(train=fasle)`

python/llm/src/bigdl/llm/llm_patching.py
```diff
@@ -20,6 +20,7 @@
 from bigdl.llm.utils.common import invalidInputError
 
 is_bigdl_patched = False
+is_train_patched = False  # Specifically to check finetuning patch
```
jason-dai: I think you can combine these two variables into one

plusbang: > I think you can combine these two variables into one

If we need to add inference patch later, we still need different variables to handle. I think maybe we could keep it also. Currently, if `train=False`, nothing is done during patching.

jason-dai: You can have one variable "bigdl_patched" as enum (None, Train, Inference).

plusbang: > You can have one variable "bigdl_patched" as enum (None, Train, Inference).

OK, I'll use this instead.

plusbang:
`bigdl_patched` is used and the value could be None or 'Train' or 'Inference'. Please let me know if any more comment / concern? if not, I think we may merge it first. @jason-dai

python/llm/example/GPU/LLM-Finetuning/patch_usage/finetune.py
```diff
@@ -0,0 +1,89 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# The only necessary code change to finetune with BigDL-LLM optimizations on Intel GPU
+from bigdl.llm import llm_patch
+llm_patch(train=True)
+
+# The following could be CUDA code of LLM finetuning application of PEFT
```
jason-dai, line 21: # The following is the original LLM finetuning code using PEFT (without BigDL-LLM)

python/llm/example/GPU/LLM-Finetuning/README.md
```diff
@@ -8,6 +8,7 @@ This folder contains examples of running different training mode with BigDL-LLM
 - [ReLora](ReLora): examples of running ReLora finetuning
 - [DPO](DPO): examples of running DPO finetuning
 - [common](common): common templates and utility classes in finetuning examples
+- [patch usage](patch_usage): examples of how to easily run finetuning on Intel GPU with BigDL-LLM patch
```
jason-dai, line 11: `[HF-PEFT](PEFT): run finetuning on Intel GPU using Hugging Face PEFT code without modification`

plusbang: > [HF-PEFT](PEFT): run finetuning on Intel GPU using Hugging Face PEFT code without modification

Have modified example folder name and comment. Besides, considering https://github.com/analytics-zoo/nano/issues/1112, we'd better not include english_quotes dataset in our examples, so I use alpaca lora example to demonstrate usage instead.

python/llm/example/GPU/LLM-Finetuning/README.md
```diff
@@ -8,6 +8,7 @@ This folder contains examples of running different training mode with BigDL-LLM
 - [ReLora](ReLora): examples of running ReLora finetuning
 - [DPO](DPO): examples of running DPO finetuning
 - [common](common): common templates and utility classes in finetuning examples
+- [HF-PEFT](PEFT): run finetuning on Intel GPU using Hugging Face PEFT code without modification
```
jason-dai, line 11: change folder name to HF-PEFT too

python/llm/src/bigdl/llm/llm_patching.py
```diff
@@ -0,0 +1,79 @@
...
+# limitations under the License.
+#
+
+import transformers
+import importlib
+import sys
+from bigdl.llm.utils.common import invalidInputError
+from enum import Enum
+
+bigdl_patched = None  # None or 'Train' or 'Inference'
```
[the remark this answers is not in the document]
plusbang: > do you plan to use enum or string?

I'm using string now. `bigdl_patched` would be None or 'Train' or 'Inference'.

plusbang:
Please take a look @jason-dai . I feel we could merge it first if no more any concern.

jason-dai (approved):
LGTM

PR #10268, opened 2024-02-28
Fix Arc StarCoder wrong query_shape when input is long

Uxito-Ada:
## Description

Referenced transformers 4.38.1 to refactor the forward code of gpt_bigcode, where layer_past is processed even not multi query.

### 1. Why the change?

perf test found wrong query shape when 1024 and 2048 long inputs are used

### 2. User API changes

no

### 3. Summary of the change 

Fix Arc StarCoder wrong query_shape when input is long

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [x] Application test
- [ ] Document test
- [ ] ...


Uxito-Ada:
Passed local test of all-in-one 1024 with starcoder 15.5B on Arc 10:
image

WeiguangHan (approved):

PR #10293, opened 2024-03-01
add warmup advice in quickstart

shane-huang:
add warmup advice in quickstart

### DIFF
--- a/docs/readthedocs/source/doc/LLM/Quickstart/install_windows_gpu.md
+++ b/docs/readthedocs/source/doc/LLM/Quickstart/install_windows_gpu.md
@@ -1,10 +1,10 @@
-# Install BigDL-LLM on Windows for Intel GPU
+# Install BigDL-LLM on Windows with Intel GPU
 
 This guide demonstrates how to install BigDL-LLM on Windows with Intel GPUs. 
 
 It applies to Intel Core Ultra and Core 12 - 14 gen integrated GPUs (iGPUs), as well as Intel Arc Series GPU.
 
-## Install GPU driver
+## Install GPU Driver
 
 * Download and Install Visual Studio 2022 Community Edition from the [official Microsoft Visual Studio website](https://visualstudio.microsoft.com/downloads/). Ensure you select the **Desktop development with C++ workload** during the installation process.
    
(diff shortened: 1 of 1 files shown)


Oscilloscope98 (approved):
LGTM

docs/readthedocs/source/doc/LLM/Quickstart/install_windows_gpu.md
```diff
@@ -121,3 +121,8 @@ Now let's play with a real LLM. We'll be using the [phi-1.5](https://huggingface
    Answer: AI stands for Artificial Intelligence, which is the simulation of human intelligence in machines.
    ```
 
+## Tips & Troubleshooting
+
+### Warmup for optimial performance on first run
+When running LLMs on GPU for the first time, you might notice the performance is lower than expected, with delays up to several minutes before the first token is generated. This delay occurs because the GPU kernels require compilation and initialization, which varies across different GPU models. To achieve optimal and consistent performance, we recommend a one-time warm-up by running `model.generate(...)` an additional time before starting your actual generation tasks. If you're developing an application, you can incorporate this warmup step into start-up or loading routine to enhance the user experience.
+
```
Oscilloscope98, line 128: Do we need include the warmup directly in the example code above?

MeouSker77 (approved):
LGTM

docs/readthedocs/source/doc/LLM/Quickstart/install_windows_gpu.md
```diff
@@ -101,6 +101,8 @@ Now let's play with a real LLM. We'll be using the [phi-1.5](https://huggingface
    # Generate predicted tokens
    with torch.inference_mode():
        input_ids = tokenizer.encode(prompt, return_tensors="pt").to('xpu')
+       # warm up one more time before the actual generation task for the first run, see details in `Tips & Troubleshooting`
+       # output = model.generate(input_ids, do_sample=False, max_new_tokens=32, generation_config = generation_config)
```
shane-huang, line 105: added a warmup code in quick start, but put it in comment to avoid confusion. @Oscilloscope98

jason-dai:
Maybe also update the WebUI quickstart

shane-huang:
> Maybe also update the WebUI quickstart

@hkvision would you please add this into quickstart enhancement?

Also we are adding the warmup step in code (during model loading step).

PR #10307, opened 2024-03-04
LLM: add script to build portable zip

JinBridger:


JinBridger:
Ready for review. PTAL. @glorysdj @liu-shaojun

PR #10314, opened 2024-03-04
Add llamaindex gpu example

Ricky-Ting:
### 1. Why the change?

port llamaindex rag example to GPU

https://github.com/analytics-zoo/nano/issues/1090


shane-huang (approved):
Merge for now and add UT and update README in following PR.

PR #10323, opened 2024-03-05
[FastChat-integration] Add initial implementation for loader

gc-fu:
## Description

Used for provide the related interface for FastChat

In the main method, we provide a function to test model loading, example result:
```bash
(fastchat-integrate) spr@spr-15:~/guancheng/fastchat-integrate/BigDL/python/llm/src/bigdl/llm/transformers$ column -t -s , loader-results-2024-03-05.csv
   model                           device  low_bit   result
0  meta-llama/Llama-2-7b-chat-hf   cpu     sym_int4  True
1  meta-llama/Llama-2-7b-chat-hf   cpu     bf16      True
2  baichuan-inc/Baichuan2-7B-Chat  cpu     sym_int4  False
3  baichuan-inc/Baichuan2-7B-Chat  cpu     bf16      False
4  Qwen/Qwen-7B-Chat               cpu     sym_int4  True
5  Qwen/Qwen-7B-Chat               cpu     bf16      True
```

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/__init__.py
+++ b/python/llm/src/bigdl/llm/transformers/__init__.py
@@ -15,7 +15,7 @@
 #
 
 
-from .convert import ggml_convert_low_bit
+from .convert import ggml_convert_low_bit, get_enable_ipex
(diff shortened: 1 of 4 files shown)


glorysdj (approved):
LGTM

PR #10325, opened 2024-03-05
Fix fschat DEP version error

Romanticoseu:
## Description

Old versions of gradio_web_server.py and openai_api_server.py will conflict with the new versions of gradio and pydantic. 
We should use the latest version of `fschat`.

### 2. User API changes
None

### 3. Summary of the change 

In this line:
https://github.com/intel-analytics/BigDL/blob/90be0f947b6a6c4b5925a96cc6bf9fdc610832f6/docker/llm/serving/cpu/docker/Dockerfile#L15
We install `fschat` with:
https://github.com/intel-analytics/BigDL/blob/90be0f947b6a6c4b5925a96cc6bf9fdc610832f6/python/llm/setup.py#L56
So update with latest version.

### DIFF
--- a/docker/llm/serving/cpu/docker/model_adapter.py.patch
+++ b/docker/llm/serving/cpu/docker/model_adapter.py.patch
@@ -1,11 +1,11 @@
---- model_adapter.py.old	2024-01-24 01:56:23.903144335 +0000
-+++ model_adapter.py	2024-01-24 01:59:22.605062765 +0000
-@@ -1346,15 +1346,17 @@
+--- model_adapter.py.old        2024-03-05 15:08:47.169275336 +0800
++++ model_adapter.py    2024-03-05 15:10:13.434703674 +0800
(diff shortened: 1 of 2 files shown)


gc-fu:
What is the version of `gradio`?

Romanticoseu:
> What is the version of `gradio`?

>=4.10
https://github.com/lm-sys/FastChat/blob/d04ce6453ae016d9e03626b679c07aa1388dcbee/pyproject.toml#L23
In fact, it was:
image

gc-fu:
Have you tested this in CPU and XPU platforms?

gc-fu (approved):
LGTM

Romanticoseu:
> Have you tested this in CPU and XPU platforms?

In XPU: if use `bigdl-llm` version after `2.5.0b20240229`, `bigdl.llm.serving.model_worker` can not work.

PR #10340, opened 2024-03-07
Fix device_map bug by raise an error when using device_map=xpu

Zhangky11:
## Description
Currently, when using the `AutoModelForCausalLM.from_pretrained(device_map='xpu')` option, it leads to significantly higher memory consumption compared to calling the `model.to('xpu')` statement. The issue is located at https://github.com/intel-analytics/BigDL/blob/main/python/llm/src/bigdl/llm/transformers/model.py#L411.

The current workaround is to prohibit users from passing the `device_map='xpu'` parameter during `from_pretrained`.

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/model.py
+++ b/python/llm/src/bigdl/llm/transformers/model.py
@@ -43,7 +43,7 @@ from .utils import extract_local_archive_file, \
     load_state_dict, \
     get_local_shard_files, load_imatrix_data
 from bigdl.llm.ggml.quantize import ggml_tensor_qtype
-from bigdl.llm.utils.common import invalidInputError, invalidOperationError
+from bigdl.llm.utils.common import invalidInputError
 from bigdl.llm.transformers.gguf.api import load_gguf_model
 import torch
 import warnings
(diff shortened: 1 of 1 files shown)


shane-huang:
also fix the misuses in examples and UTs.

python/llm/src/bigdl/llm/transformers/model.py
```diff
@@ -409,6 +414,7 @@ def load_convert(cls, q_k, optimize_model, *args, **kwargs):
             _load_pre()
             try:
                 model = cls.HF_Model.from_pretrained(*args, **kwargs)
+
```
hkvision, line 417: revert this?

Zhangky11: OK

python/llm/src/bigdl/llm/transformers/model.py
```diff
@@ -145,6 +145,11 @@ def from_pretrained(cls,
         invalidInputError(model_hub in ["huggingface", "modelscope"],
                           "The parameter `model_hub` is supposed to be `huggingface` or "
                           f"`modelscope`, but got {model_hub}.")
+        if 'device_map' in kwargs and 'xpu' in kwargs['device_map']:
+            invalidOperationError(condition=0,
```
hkvision, line 149: why invalidOperationError not invalidInputError?

hkvision (approved):

PR #10348, opened 2024-03-08
Langchain readme

shane-huang:
The current langchain CPU readme still uses native_int4 instructions which is quite confusing for new users. 

Changes:
- Updates the README and move the native_int4 REAME into legacy section.
- Update the install and example run command sections (will verify)


shane-huang:
merge for now for verification and will revise in following PRs.

PR #10351, opened 2024-03-08
langchain: fix core dump issue caused by missing tokenizer when loading from low bit model

ivy-lv11:
Fix langchain. 
Details could be found in the issue https://github.com/analytics-zoo/nano/issues/1126. 
Use `model_id` and `tokenizer_id` to distinguish.


shane-huang:
Also update README to describe this example.

shane-huang:
Does this mean user have to keep both the "original model folder" and "quantized model folder" at the same time? 
Some users might want to keep only the quantized model to save disk space.  
We need to write some instructions in README to tell user how to use the quantized model folder only  (e.g. copy the tokenizers, etc. to quantized model folder, and no need to set tokenizer_id). Please also verify it works.

ivy-lv11:
> Does this mean user have to keep both the "original model folder" and "quantized model folder" at the same time? Some users might want to keep only the quantized model to save disk space. We need to write some instructions in README to tell user how to use the quantized model folder only (e.g. copy the tokenizers, etc. to quantized model folder, and no need to set tokenizer_id). Please also verify it works.

I have verified that it works well if users copy the tokenizers into quantized model folder and update the README.

PR #10358, opened 2024-03-11
LLM serving xpu memory opt

cyita:
## Description

The current implementation tends to result in out-of-memory errors. Add `empty cache` to free up memory.

### 4. How to test?
- [x] local test


xiangyuT (approved):
LGTM

PR #10364, opened 2024-03-11
LLM: fix error of 'AI-ModelScope/phi-2' hosted by ModelScope hub

plusbang:
## Description

In [ModelScope](https://modelscope.cn/models/AI-ModelScope/phi-2/file/view/master/config.json), `model_type` of `AI-ModelScope/phi-2` is `phi-msft`.
Then optimization for `phixtral` is applied to `phi-2` and will cause error. More details are in https://github.com/analytics-zoo/nano/issues/1140#issuecomment-1987534096

### 2. User API changes

N/A

### 4. How to test?
- [ ] Unit test
- [ ] Local test


rnwang04 (approved):
LGTM, please merge it after style ut is passed.
May also add simple comments here to illustrate why we need to limit this condition.

PR #10368, opened 2024-03-11
Qwen fused qkv

qiuxin2012:
## Description

Optimize qwen using fused qkv+rope.
Quantized & fused qkv+rope is not enabled because not benefit on MTL.


python/llm/src/bigdl/llm/transformers/models/qwen.py
```diff
@@ -193,30 +222,201 @@ def qwen_attention_forward(
...
+            bsz,
+            self.num_heads,
+            self.head_dim,
+            0,
+            1,
+            dtype=hidden_states.dtype,
+            device=device
+        )
+
+        position_ids = self.position_ids[self.kv_seq_len].to('xpu')
```
lalalapotter, line 299: maybe use `to(device)` ?

lalalapotter (approved):
others LGTM.

PR #10382, opened 2024-03-12
LLM: add low bit option in deepspeed autotp example

plusbang:
## Description

Add low bit option in deepspeed autotp example to make users easily run with other low bit optimization.

### 4. How to test?
- [x] Local test


plusbang:
> lgtm
> 
> btw, has it been applied on cpu?

It seems not applied on cpu yet, maybe we could add it in another PR.

rnwang04 (approved):
LGTM

PR #10384, opened 2024-03-12
use new rms norm

MeouSker77:
## Description

use new rms norm

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/models/baichuan2.py
+++ b/python/llm/src/bigdl/llm/transformers/models/baichuan2.py
@@ -50,16 +50,12 @@ KV_CACHE_ALLOC_BLOCK_LENGTH = 256
 def baichuan_13b_rms_norm_forward(self, hidden_states):
     if hidden_states.device.type == "xpu" and not (self.training and hidden_states.requires_grad):
         import linear_q4_0
-        result = linear_q4_0.fused_rms_norm(hidden_states,
-                                            [self.weight.size(0)],
-                                            self.weight,
-                                            None,
-                                            self.epsilon)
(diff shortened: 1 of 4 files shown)


rnwang04 (approved):
LGTM

hkvision:
This PR doesn't pass the unit tests: https://github.com/intel-analytics/BigDL/actions/runs/8246492638/job/22552952279
Please make sure all the tests are passed before you merge your PR.

MeouSker77:
action failed because it used a new API from bigdl-core-xe-21, but action will use latest nightly bigdl-core-xe-21, not main branch codeof bigdl-core-xe-21, there is no way to avoid such problems unless we change BigDL's action to build bigdl-core-xe-21 in action instead of using nightly bigdl-core-xe-21

liu-shaojun:
@MeouSker77 
So tonight's test with today's nightly bigdl-core-xe-21 will run successfully because it includes the new API, right?

MeouSker77:
yes, this action failed because it has no new API. 

yesterday's nightly test already had new API, it failed because qwen's rms norm error became larger.

hkvision:
> action failed because it used a new API from bigdl-core-xe-21, but action will use latest nightly bigdl-core-xe-21, not main branch code of bigdl-core-xe-21, there is no way to avoid such problems unless we change BigDL's action to build bigdl-core-xe-21 in action instead of using nightly bigdl-core-xe-21

To avoid this problem, if the PR in BigDL is not that urgent, can wait for the next morning to merge after the new core-xe releases?

MeouSker77:
> To avoid this problem, if the PR in BigDL is not that urgent, can wait for the next morning to merge after the new core-xe releases?

yes, this is a solution

Issue #10354, opened 2024-03-08
Qwen1.5-7B wrong outputs with 1024 prompts


Uxito-Ada:
**code:** all-in-one benchmark, where prmopt/2048.txt is replaced with the below Chinese ones
**in-out pair:** 1024-128 (2048 prompts are truncated to 1024)
**model:** Qwen1.5-7B-Chat
**machine:** SPR01

[红楼梦](https://github.com/analytics-zoo/speculative-decoding/blob/main/prompts/xft_prompt.json#L65)
INT4/INT8/BF16 all repeat like:
```bash
空空道人便骑着驴往往去，往去往去，往去往去，往去往去，往去往去，往去往去,往去往去,往去往去,往去往去,往去往去,往去往去,往去往去,往去往去,往去往去,往去往去,往去往去,往去往去,往去往去,往去往去,往去往去,往去往去,往去往去,往去往去,往去往去,往
```

[患者](https://github.com/analytics-zoo/speculative-decoding/blob/main/prompts/ccg_ch_prompt.json#L7)
INT4 repeats like below, while BF16 and INT8 give no answer:
```bash
临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断临床实验室实验室诊断
```

Uxito-Ada:
@ivy-lv11 pls take a look at this

Uxito-Ada:
transformers: 4.38.1/4.37.0
torch: 2.2.0+cpu
ipex: 2.2.0

jason-dai:
> transformers: 4.38.1/4.37.0 torch: 2.2.0+cpu ipex: 2.2.0

[...]

PR #10386, opened 2024-03-12
Fix convert_forward recursive submodule issue

Uxito-Ada:
## Description

Not every model needs recursive convert_forward on its submodule, otherwise strange behavior of model appears due to repeated setattr.

Add a parameter to specify whether submodule needs to recursively converted in conver_forward, default to True. Qwen1.5 (qwen2) is an example of False.

### 1. Why the change?

#10354 

### 2. User API changes

no

### 3. Summary of the change 

as above

### 4. How to test?
- [ ] N/A
- [x] Unit test
- [x] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/convert.py
+++ b/python/llm/src/bigdl/llm/transformers/convert.py
@@ -669,7 +669,7 @@ def convert_bigdl_other_module(model, dtype):
                 module.to(dtype)
 
 
-def convert_forward(m, target_m, new_forward, replace_sub_module = True):
+def convert_forward(m, target_m, new_forward, replace_sub_module=True):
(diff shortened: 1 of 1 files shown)


Uxito-Ada:
With this, Qwen1.5 7b can generate normal output on 1024 prompt, which was not available previously.

image

Uxito-Ada:
Also, inserted print point and saw that all the Qwen2Attention's forward had been converted as expected in this way.

plusbang (approved):
Please also confirm the behavior of qwen2 on other platforms (such as ARC). Besides, with this fix, is there any influence on performance?

PR #10396, opened 2024-03-13
use fp8 sdp in llama

MeouSker77:
## Description

use fp8 sdp in llama with transformers 4.31.

Note: current fp8 sdp and fp16 sdp both don't support attention_mask, we assume masks are all zeros


rnwang04 (approved):
LGTM

PR #10402, opened 2024-03-13
[Fix] LLM: Fix condition check error for speculative decoding on CPU

xiangyuT:
## Description

Fix condition check error for speculative decoding on CPU.


gc-fu (approved):
LGTM

PR #10404, opened 2024-03-13
support new baichuan model

MeouSker77:
## Description

support new baichuan model

### Usage

```python
import torch

model_path = "<model_path>"

model = AutoModelForCausalLM.from_pretrained(model_path, trust_remote_code=True,
                                             load_in_low_bit='sym_int8', optimize_model=True).eval()
model = model.to('xpu')
```


jason-dai:
`import intel_extension_for_pytorch as ipex` is not needed as we have auto imported it?

MeouSker77:
> `import intel_extension_for_pytorch as ipex` is not needed as we have auto imported it?

yes, removed it

rnwang04 (approved):

PR #10406, opened 2024-03-13
Add baichuan2-13b 1k to arc nightly perf

hkvision:
After empty cache, can now run baichuan2-13b for 1k input.


WeiguangHan (approved):
LGTM

PR #10416, opened 2024-03-14
add original example for llm_patch

zhentaocc:
## Description

an original example to show llm_patch usage


python/llm/example/GPU/LLM-Patch-Inference/hf_generate.py
```diff
@@ -0,0 +1,384 @@
...
+import warnings
+from argparse import ArgumentParser, ArgumentTypeError, Namespace
+from contextlib import nullcontext
+from typing import Dict, Union
+
+import numpy as np
+import torch
+from transformers import AutoConfig, AutoModelForCausalLM, AutoTokenizer
+from bigdl.llm.llm_patching import llm_patch
+llm_patch()
```
hkvision, line 34: Can we put this at the very beginning of the file?

zhentaocc: the order won't influence the result

hkvision: I mean putting it at the very beginning is more friendly to users and easier to find/read.

zhentaocc: seems that this function must be called before bigdl import

PR #10420, opened 2024-03-14
update deepseek example readme

ivy-lv11:
Update readme for running the Deepseek-moe-16b example. 
Details could be found in the issue https://github.com/analytics-zoo/nano/issues/1151.

### DIFF
--- a/python/llm/example/CPU/HF-Transformers-AutoModels/Model/deepseek-moe/README.md
+++ b/python/llm/example/CPU/HF-Transformers-AutoModels/Model/deepseek-moe/README.md
@@ -30,6 +30,14 @@ After setting up the Python environment, you could run the example by following 
 >
 > Please select the appropriate size of the DeepSeek-MoE model based on the capabilities of your machine.
 
+> **Note**: Turn off flash-attention before running the example. 
+> Some modifications are made in the `modeling_deepseek.py` under the model folder after download the model from https://huggingface.co/deepseek-ai/deepseek-moe-16b-chat. 
+> * comment `is_flash_attn_2_available` in L45 and also comment corresponding definition in L54-L56
+> * comment `is_flash_attn_greater_or_equal_2_10` in L46
+> * set ` _supports_flash_attn_2 = False` in the definition of DeepseekPreTrainedModel in L999
+> * set `self._use_flash_attention_2 = False` in L1107 in the initialization of DeepseekModel. 
+> 
+
 #### 2.1 Client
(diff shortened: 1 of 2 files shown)


shane-huang:
There're so many modifications and notes may be ignored by users. 
- also upload the modified `modeling_deepseek.py` and specify the path of the file which user need to replace, i.e. https://huggingface.co/deepseek-ai/deepseek-moe-16b-chat/blob/main/modeling_deepseek.py
- in readme, add an extra step to replace the  `modeling_deepseek.py`  with command. And you can add a note to tell user what has been replaced and why the changes are needed (i.e. to disable flash attention)

python/llm/example/CPU/HF-Transformers-AutoModels/Model/deepseek-moe/README.md
```diff
@@ -30,6 +30,17 @@ After setting up the Python environment, you could run the example by following
 >
 > Please select the appropriate size of the DeepSeek-MoE model based on the capabilities of your machine.
 
+> **Note**: Turn off flash-attention before running the example.
```
shane-huang, line 33: #### 2.0 disable flash attention 
You need to disable flash attention to run this model. To do this, simply replace the file `deepseek-moe-16b-chat/modeling_deepseek.py`(the original file can be found [here](https://huggingface.co/deepseek-ai/deepseek-moe-16b-chat/blob/main/modeling_deepseek.py)) in the downloaded model path with the file [`./modeling_deepseek.py`](./modeling_deepseek.py) we provided in the current directory. 

>Note: The modifications we made in `modeling_deepseek.py` include: 
> * comment `is_flash_attn_2_available` in L45 and also comment corresponding definition in L54-L56
> * comment `is_flash_attn_greater_or_equal_2_10` in L46
> * set ` _supports_flash_attn_2 = False` in the definition of DeepseekPreTrainedModel in L999
> * set `self._use_flash_attention_2 = False` in L1107 in the initialization of DeepseekModel. 

If you encounter `transformers.cache_utils` related problems, please refer to [this discussion thread](https://stackoverflow.com/questions/77920734/modulenotfounderror-no-module-named-transformers-utils ) for possible solutions.

jason-dai: This is too complex; just provide a modified file and ask the user to use that file

shane-huang (approved):
LGTM.

PR #10421, opened 2024-03-14
Add quickstart for install bigdl-llm in docker on windows with Intel GPU

hxsz1997:
Add quickstart for install bigdl-llm in docker on windows with Intel GPU.


shane-huang (approved):
Merge the initial version and revise in future PR.

PR #10425, opened 2024-03-14
Fix pad_token_id issue in Phi-2 example

jenniew:
In latest Phi-2 model, it add eos_token_id , but doesn't add pad_token_id. Therefore it would cause error: https://github.com/analytics-zoo/nano/issues/1139

Fix the problem by `model.generation_config.pad_token_id = model.generation_config.eos_token_id`


Oscilloscope98 (approved):
LGTM

PR #10430, opened 2024-03-15
Add FastChat iGPU example

Romanticoseu:
## Description

Add FastChat iGPU example

### 4. How to test?
- [ ] Application test


Uxito-Ada:
Could u provide a request/response example? And also tell users how to verify Web UI?

PR #10433, opened 2024-03-15
add mask support for llama/chatglm fp8 sdp

MeouSker77:
## Description

add mask support for llama/chatglm fp8 sdp


rnwang04 (approved):
LGTM

PR #10441, opened 2024-03-18
LLM: Fix log condition for BIGDL_OPT_IPEX

xiangyuT:
## Description

Use torch.__version__ to check whether BigDL is running on CPU or XPU.


glorysdj (approved):
LGTM

PR #10454, opened 2024-03-19
[Dependabot]: upgrade org.apache.zookeeper:zookeeper

liu-shaojun:
to 3.9.2

### 1. Why the change?

Upgrade org.apache.zookeeper:zookeeper to 3.9.2 to fix Dependabot issues.
https://github.com/intel-analytics/BigDL/security/dependabot/2892
https://github.com/intel-analytics/BigDL/security/dependabot/2893


glorysdj (approved):
LGTM

PR #10456, opened 2024-03-19
LLM: add esimd sdp for pvc

rnwang04:
### 1. Why the change?
https://github.com/analytics-zoo/nano/issues/1176#issuecomment-2005541797

### 2. User API changes

No change.

### 3. Summary of the change 

- add esimd sdp for pvc , only update llama now

### 4. How to test?
- [ ] Unit test

### DIFF
--- a/python/llm/src/bigdl/llm/transformers/models/utils.py
+++ b/python/llm/src/bigdl/llm/transformers/models/utils.py
@@ -325,7 +325,7 @@ def use_esimd_sdp(q_len, k_len, head_dim, query_states, attention_mask=None):
                 device_name.startswith("Intel(R) Data Center GPU Flex") or \
                 device_name.startswith("Intel(R) Data Center GPU Max"):
             import linear_fp16_esimd
-            if not hasattr(linear_fp16_esimd, "sdp_forward"):
+            if hasattr(linear_fp16_esimd, "sdp_forward"):
                 return True
             else:
                 return False
--- a/python/llm/src/bigdl/llm/transformers/models/utils.py
+++ b/python/llm/src/bigdl/llm/transformers/models/utils.py
@@ -299,7 +299,7 @@ def use_flash_attention(query, key, attention_mask=None):
     return True
 
 
-def use_esimd_sdp(q_len, k_len, head_dim, query_states):
+def use_esimd_sdp(q_len, k_len, head_dim, query_states, attention_mask=None):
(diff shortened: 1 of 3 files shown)


python/llm/src/bigdl/llm/transformers/models/utils.py
```diff
@@ -315,10 +315,15 @@ def use_esimd_sdp(q_len, k_len, head_dim, query_states):
     elif query_states.dtype != torch.float16:
         # esimd_sdp only has optimization for FP16 now
         return False
+    elif query_states.shape[0] > 1 and attention_mask is not None:
+        # for batched input, can't accept attention_mask
```
hkvision, line 319: For batched input and all zero attention mask, it has no return and will be None?

rnwang04: Oh yes... Thanks for pointing that out!
Maybe should change `use_esimd_sdp` into below way ?
```python
def use_esimd_sdp(q_len, k_len, head_dim, query_states):
    if head_dim != 128:
        # esimd_sdp only support head_dim = 128 now
        return False
    elif q_len != 1:
        # esimd_sdp only support rest token now
        return False
    elif k_len < 8:
        # esimd_sdp will cause wrong output when k_len < 8
        return False
    elif query_states.device.type != "xpu":
        # esimd_sdp only support GPU now
        return False
    elif query_states.dtype != torch.float16:
        # esimd_sdp only has optimization for FP16 now
        return False
    elif query_states.shape[0] > 1 and attention_mask is not None:
        # for batched input, can't accept attention_mask
        if len(torch.nonzero(attention_mask, as_tuple=True)) > 1:
            return False

    device_name = torch.xpu.get_device_name(query_states.device.index)
    if device_name.startswith("Intel(R) Arc(TM) A") or \
              device_name.startswith("Intel(R) Data Center GPU Flex") or \
              device_name.startswith("Intel(R) Data Center GPU Max"):
        import linear_fp16_esimd
        if not hasattr(linear_fp16_esimd, "sdp_forward"):
            return False
    else:
        return False

    return True
```

hkvision: Seems good!

rnwang04:
new pr is raised here: https://github.com/intel-analytics/ipex-llm/pull/10543

PR #10466, opened 2024-03-19
gemma fp16 sdp

qiuxin2012:
## Description

gemma fp16 sdp


qiuxin2012:
Could not used on Arc as head_dim is 256, not 128.

PR #10473, opened 2024-03-20
LLM: fix whiper model missing config.

lalalapotter:
## Description

This PR is to fix whisper model missing config attributes.


liu-shaojun (approved):
LGTM

PR #10474, opened 2024-03-20
fix rwkv v5 fp16

MeouSker77:
## Description

fix rwkv v5 fp16


rnwang04 (approved):

MeouSker77:
the action fail is not caused by this PR, so I'll merge this PR

PR #10483, opened 2024-03-20
Fix `modules_not_to_convert` argument

MeouSker77:
## Description

Fix `modules_not_to_convert` argument in `from_pretrained` and `load_low_bit`


rnwang04 (approved):
LGTM

Issue #10476, opened 2024-03-20
LLM QLoRA script issue


JeNi0310:
Hi Habana team,

I met a issue when I ran the script "qlora_finetune_llama2_70b_pvc_1550_4_card.sh" and deepspeed parameters are used.
When running the code, errors occur whenever a checkpoint step is encountered (for quick verification, I set the save checkpoint to 10 steps).
Please refer error log as below.
`Traceback (most recent call last):
  File "/home/pvc/BigDL/python/llm/example/GPU/LLM-Finetuning/QLoRA/alpaca-qlora/./alpaca_qlora_finetuning_70B.py", line 291, in <module>
    fire.Fire(train)
  File "/home/pvc/anaconda3/envs/llm/lib/python3.9/site-packages/fire/core.py", line 143, in Fire
    component_trace = _Fire(component, args, parsed_flag_args, context, name)
  File "/home/pvc/anaconda3/envs/llm/lib/python3.9/site-packages/fire/core.py", line 477, in _Fire
    component, remaining_args = _CallAndUpdateTrace(
  File "/home/pvc/anaconda3/envs/llm/lib/python3.9/site-packages/fire/core.py", line 693, in _CallAndUpdateTrace
    component = fn(*varargs, **kwargs)
  File "/home/pvc/BigDL/python/llm/example/GPU/LLM-Finetuning/QLoRA/alpaca-qlora/./alpaca_qlora_finetuning_70B.py", line 281, in train
    trainer.train(resume_from_checkpoint=resume_from_checkpoint)
  File "/home/pvc/anaconda3/envs/llm/lib/python3.9/site-packages/transformers/trainer.py", line 1591, in train
    return inner_training_loop(
  File "/home/pvc/anaconda3/envs/llm/lib/python3.9/site-packages/transformers/trainer.py", line 1984, in _inner_training_loop
    self._maybe_log_save_evaluate(tr_loss, model, trial, epoch, ignore_keys_for_eval)
  File "/home/pvc/anaconda3/envs/llm/lib/python3.9/site-packages/transformers/trainer.py", line 2339, in _maybe_log_save_evaluate
    self._save_checkpoint(model, trial, metrics=metrics)
  File "/home/pvc/anaconda3/envs/llm/lib/python3.9/site-packages/transformers/trainer.py", line 2400, in _save_checkpoint
    self.model_wrapped.save_checkpoint(output_dir)
[...]

PR #10484, opened 2024-03-20
LLM: fix deepspeed error of finetuning on xpu

rnwang04:
### 1. Why the change?

User issue: https://github.com/intel-analytics/BigDL/issues/10476
related internal issue: https://github.com/analytics-zoo/nano/issues/1183

### 2. User API changes

No change.

### 3. Summary of the change 

- Fix  deepspeed error of finetuning on xpu

### 4. How to test?
- [ ] Local test
- [ ] Unit test

### DIFF
--- a/python/llm/example/GPU/LLM-Finetuning/QLoRA/alpaca-qlora/alpaca_qlora_finetuning.py
+++ b/python/llm/example/GPU/LLM-Finetuning/QLoRA/alpaca-qlora/alpaca_qlora_finetuning.py
@@ -32,7 +32,7 @@
 
 import os
 from typing import List
-
+os.environ["ACCELERATE_USE_XPU"] = "true"
 import fire
 import torch
 import transformers
--- a/python/llm/example/GPU/LLM-Finetuning/QLoRA/alpaca-qlora/deepspeed_zero2.json
+++ b/python/llm/example/GPU/LLM-Finetuning/QLoRA/alpaca-qlora/deepspeed_zero2.json
@@ -7,10 +7,10 @@
       "contiguous_gradients": true,
       "overlap_comm": true
     },  
-    "bp16": {
+    "bf16": {
       "enabled": true
     },
     "train_micro_batch_size_per_gpu": "auto",
     "gradient_accumulation_steps": "auto"
-  }
+}
   


plusbang (approved):
LGTM

yangw1234 (approved):

yangw1234:
The root cause of this problem is when we do not set "ACCELERATE_USE_XPU", the [deepspeed code branch](https://github.com/intel-analytics/BigDL/blob/main/python/llm/src/bigdl/llm/transformers/training_patch.py#L115-L120) will be skipped because the code goes to the [xpu branch](https://github.com/intel-analytics/BigDL/blob/main/python/llm/src/bigdl/llm/transformers/training_patch.py#L106-L111).

This problem is fixed in `transformers==4.37.0` along with the device issues in the xpu code branch, (which is reason why we have this `training_patch.py`)


So, I think a cleaner solution is to upgrade `transformers` to 4.37.0 or above, and remove the [setup_devices patch](https://github.com/intel-analytics/BigDL/blob/main/python/llm/src/bigdl/llm/transformers/training_patch.py#L197-L198)

qiuxin2012:
Please delete branch after you merge PR, because this fix_ds is on upstream.

rnwang04:
> This problem is fixed in `transformers==4.37.0` along with the device issues in the xpu code branch, (which is reason why we have this `training_patch.py`)
> 
> So, I think a cleaner solution is to upgrade `transformers` to 4.37.0 or above, and remove the [setup_devices patch](https://github.com/intel-analytics/BigDL/blob/main/python/llm/src/bigdl/llm/transformers/training_patch.py#L197-L198)

Thanks for this solution ! I will merge this PR first and raise another PR to upgrade to transformers 4.37.0 .
Maybe we should keep setup_devices patch to be compatible with 4.34.0 ? Just add a version check for it.

rnwang04:
> Please delete branch after you merge PR, because this fix_ds is on upstream.

Sorry for my mistake. will delete it once I merge.

yangw1234:
> > This problem is fixed in `transformers==4.37.0` along with the device issues in the xpu code branch, (which is reason why we have this `training_patch.py`)
> > So, I think a cleaner solution is to upgrade `transformers` to 4.37.0 or above, and remove the [setup_devices patch](https://github.com/intel-analytics/BigDL/blob/main/python/llm/src/bigdl/llm/transformers/training_patch.py#L197-L198)
> 
> Thanks for this solution ! I will merge this PR first and raise another PR to upgrade to transformers 4.37.0 . Maybe we should keep setup_devices patch to be compatible with 4.34.0 ? Just add a version check for it.

I personally prefer deleting it and only support 4.37.0+ to keep low maintenance cost.

rnwang04:
> > Thanks for this solution ! I will merge this PR first and raise another PR to upgrade to transformers 4.37.0 . Maybe we should keep setup_devices patch to be compatible with 4.34.0 ? Just add a version check for it.
> 
> I personally prefer deleting it and only support 4.37.0+ to keep low maintenance cost.

@jason-dai I wonder do you think shall we directly upgrade to 4.37.0 ?

jason-dai:
> > > Thanks for this solution ! I will merge this PR first and raise another PR to upgrade to transformers 4.37.0 . Maybe we should keep setup_devices patch to be compatible with 4.34.0 ? Just add a version check for it.
> > 
> > 
> > I personally prefer deleting it and only support 4.37.0+ to keep low maintenance cost.
> 
> @jason-dai I wonder do you think shall we directly upgrade to 4.37.0 ?

yes, but probably later

PR #10488, opened 2024-03-20
LLM: add empty cache in deepspeed autotp benchmark script

plusbang:
## Description

Add `torch.xpu.empty_cache()` during every generation

### 4. How to test?
- [x] Unit test
- [x] Local test


rnwang04 (approved):
LGTM

PR #10490, opened 2024-03-20
lm_head empty_cache for more models

hkvision:
As tested here for llama: https://github.com/analytics-zoo/nano/issues/1131#issuecomment-2009534203
empty cache seems won't impact performance for models besides baichuan/qwen (may need to test for other models)

Relax the constraint to include llama, this will decrease peak memory and make memory usage better for 4k input.


lalalapotter (approved):
LGTM.

hkvision:
image
The previous commit passed the checks. The other commit fixes the style. Merge it first.

PR #10496, opened 2024-03-21
[LLM] Add nightly igpu perf test for INT4+FP16 1024-128

Oscilloscope98:
## Description

Add nightly igpu perf test for INT4+FP16 1024-128
Skip rwkv models due to https://github.com/analytics-zoo/nano/issues/1180#issuecomment-2008576711

### How to test?
- [x] Perf test

TODO: add an action function for different in-out and test_api


rnwang04 (approved):
LGTM

PR #10498, opened 2024-03-21
LLM: don't export env variables when linux kernel is 6.5

WeiguangHan:
@hkvision Don't export env variables when linux kernel is 6.5


hkvision (approved):

hkvision:
Also need to update run-max-gpu.sh?

WeiguangHan:
> Also need to update run-max-gpu.sh?

sure

python/llm/dev/benchmark/all-in-one/run-arc.sh
```diff
@@ -1,5 +1,9 @@
 source /opt/intel/oneapi/setvars.sh
+
 export USE_XETLA=OFF
-export SYCL_PI_LEVEL_ZERO_USE_IMMEDIATE_COMMANDLISTS=1
+KERNEL_VERSION=$(uname -r)
+if [[ $KERNEL_VERSION != *"6.5"* ]]; then
```
qiuxin2012, line 5: what if 6.5 and higher?

hkvision: we don't set `SYCL_PI_LEVEL_ZERO_USE_IMMEDIATE_COMMANDLISTS` for 6.5 or higher versions, namely do nothing.

PR #10529, opened 2024-03-25
LLM: Add length check for IPEX-CPU speculative decoding 

xiangyuT:
## Description

Add length check for IPEX-CPU speculative decoding. For length < 256 prompts will use original_generate() method.

### DIFF
--- a/python/llm/src/ipex_llm/transformers/speculative.py
+++ b/python/llm/src/ipex_llm/transformers/speculative.py
@@ -64,6 +64,7 @@ def generate(
                         'auto_th_stop_draft', 'auto_parameters', 'min_step_draft',
                         'th_batch_num']:
                 value = kwargs.pop(var, None)
+            del self.draft_model
             return original_generate(self,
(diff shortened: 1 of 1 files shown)


python/llm/src/ipex_llm/transformers/speculative.py
```diff
@@ -53,6 +53,28 @@ def generate(
     **kwargs,
 ):
     if hasattr(self, "draft_model"):
+        from ipex_llm.llm.transformers.convert import get_enable_ipex
+        _enable_ipex = get_enable_ipex()
```
qiyuangong, line 57: Add CPU device check.

qiyuangong: get_enable_ipex is CPU only. Others LGTM

qiyuangong: @jason-dai

qiyuangong (approved):
LGTM

PR #10540, opened 2024-03-26
fix chatglm

MeouSker77:
## Description

The shape of chatglm's hidden_states is `[seq_len, batch_size, hidden_size]`, which is different from other models' `[batch_size, seq_len, hidden_size]`.

When `2 <= seq_len <= 8`, it will use quantize kv in first token, and not use quantize kv in rest token, which causes this error, so fix it.

https://github.com/intel-analytics/ipex-llm/issues/10515


sgwhat (approved):

Oscilloscope98 (approved):

Oscilloscope98:
@JinBridger has conducted some tests on Arc A770, and the bug has been resolved. :)

PR #10544, opened 2024-03-26
Remove ipex-llm dependency in readme

JinBridger:
We remove this as this is for native int4 and may be confused for users.


qiyuangong (approved):
LGTM

qiyuangong:
@jason-dai 
Can we remove this legacy glibc table?

jason-dai:
> @jason-dai Can we remove this legacy glibc table?

Yes

PR #10545, opened 2024-03-26
replace bigdl-llm with ipex-llm

liu-shaojun:
### 1. Why the change?

This submission replaces bigdl-llm with ipex-llm in workflows.


glorysdj (approved):
LGTM

PR #10555, opened 2024-03-27
support fp8 in xetla

yangw1234:
### 1. Why the change?

Support fp8 in xetla path and fixed several bugs.

### 2. User API changes

No

### 3. Summary of the change 
1. adapt latest changes in core-xe to rename `mm_int4` to `mm_xetla`
2. support moving back to cpu
3. support fp8
4. add fuse qkv to mistral

### 4. How to test?
- [x] Unit test
- [x] manually tested

### 5. New dependencies

No


yangw1234:
The llama index error is irrelevant.

PR #10559, opened 2024-03-27
Create scorecard.yml

liu-shaojun:
### 1. Why the change?

This submission creates scorecard.yml to scan and generate OpenSSF score.


glorysdj (approved):
LGTM

PR #10560, opened 2024-03-27
fix -1 top_k

gc-fu:
## Description

fix top_k error


Oscilloscope98 (approved):
LGTM

We need to be more careful regarding the default `top_p`, `top_k` values, etc., after supporting of sampling, beam search, etc.

PR #10566, opened 2024-03-27
LLM: set different envs based on different Linux kernels

WeiguangHan:
@hkvision Set different envs based on different Linux kernels

### DIFF
--- a/python/llm/dev/benchmark/all-in-one/run-arc.sh
+++ b/python/llm/dev/benchmark/all-in-one/run-arc.sh
@@ -1,5 +1,10 @@
 source /opt/intel/oneapi/setvars.sh
+ 
 export USE_XETLA=OFF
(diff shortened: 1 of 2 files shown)


hkvision (approved):

python/llm/dev/benchmark/all-in-one/run-arc.sh
```diff
@@ -1,5 +1,10 @@
 source /opt/intel/oneapi/setvars.sh
+ 
 export USE_XETLA=OFF
-export SYCL_PI_LEVEL_ZERO_USE_IMMEDIATE_COMMANDLISTS=1
-
-python run.py # make sure config YAML file
+export SYCL_CACHE_PERSISTENT=1
```
hkvision, line 4: @Oscilloscope98 We have added this as default.

PR #10582, opened 2024-03-28
Fix Qwen-VL example problem

jenniew:
Fix Qwen-VL example issue: https://github.com/analytics-zoo/nano/issues/1095
Also the issue: AttributeError: 'QWenAttention' object has no attribute 'c_attn'
Apply fuse qkv to Qwen-VL.

Already tested on GPU and CPU


python/llm/src/ipex_llm/transformers/models/qwen_vl.py
```diff
@@ -139,28 +174,42 @@ def qwen_attention_forward_vl(
...
 
+    if decoding_fast_path:
+        query = query.transpose(1, 2) # change to (bsz, q_len, num_heads, head_dim)
+
     if self.use_logn_attn and not self.training:
         if self.logn_tensor.device != query.device or self.logn_tensor.dtype != query.dtype:
             self.logn_tensor = self.logn_tensor.to(query.device).type_as(query)
-        seq_start = key.size(1) - query.size(1)
-        seq_end = key.size(1)
+        seq_start = key_size - key_size
```
lalalapotter, line 191: please check if the `seq_start` logic is correct.

lalalapotter (approved):
Others LGTM.

python/llm/src/ipex_llm/transformers/models/qwen_vl.py
jenniew: already tested this logic

PR #10585, opened 2024-03-29
nightly build docker images

liu-shaojun:
### 1. Why the change?

This submission builds ipex-llm docker images every night


glorysdj (approved):
LGTM

PR #10591, opened 2024-03-29
fix rwkv with pip installer

MeouSker77:
## Description

remove rwkv bf16 support to fix issues when using pip installer


rnwang04 (approved):
LGTM

PR #10592, opened 2024-03-29
LLM: add memory optimization for llama.

lalalapotter:
## Description

This PR is to add some of the memory optimization methods mentioned [here](https://github.com/analytics-zoo/nano/issues/1071#issuecomment-1997370459) for llama.

### DIFF
--- a/python/llm/src/ipex_llm/transformers/models/llama.py
+++ b/python/llm/src/ipex_llm/transformers/models/llama.py
@@ -83,7 +83,7 @@ def repeat_kv(hidden_states: torch.Tensor, n_rep: int) -> torch.Tensor:
                                                            n_rep, slen, head_dim)
     return hidden_states.reshape(batch, num_key_value_heads * n_rep, slen, head_dim)
 
-KV_CACHE_ALLOC_BLOCK_LENGTH = 256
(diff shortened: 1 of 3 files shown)


python/llm/src/ipex_llm/transformers/models/llama.py
```diff
@@ -186,7 +186,14 @@ def llama_mlp_forward(
             hidden_states = attn_output.view(x.shape)
         return hidden_states
     else:
-        out = self.down_proj(self.act_fn(self.gate_proj(x)) * self.up_proj(x))
+        if os.environ.get("IPEX_LLM_LOW_MEM", None) == "1":
```
hkvision, line 189: This may not need to environment control?
Splitting the implementation won't impact the performance?

lalalapotter: Sure, test bigdl_fp16_gpu API on ARC, results:
before split:
0,meta-llama/Llama-2-7b-chat-hf,492.33,36.25,0.0,1024-128,1,1025-128,1,sym_int4,N/A,30.71,13.837890625,N/A
after split:
0,meta-llama/Llama-2-7b-chat-hf,493.1,35.49,0.0,1024-128,1,1025-128,1,sym_int4,N/A,7.57,13.6640625,N/A
This split should not influence the performance, will remove the environment variable check.

hkvision (approved):

PR #10593, opened 2024-03-29
refine and verify ipex-llm-xpu docker document

liu-shaojun:
### 1. Why the change?

This submission refines and verifies ipex-llm-xpu docker document


glorysdj (approved):
LGTM

PR #10609, opened 2024-04-01
Fix dtype mismatch error

ivy-lv11:
Fix dtype mismatch error when `load_in_low_bit='bf16'` 
CPU and GPU both: RuntimeError: expected m1 and m2 to have the same dtype, but got: float != c10::BFloat16 and details could be found in the issue https://github.com/analytics-zoo/nano/issues/1111

### DIFF
--- a/python/llm/src/ipex_llm/transformers/models/llama.py
+++ b/python/llm/src/ipex_llm/transformers/models/llama.py
@@ -1324,6 +1324,8 @@ def native_sdp(query, key, value, attention_mask,
 
     # at inference time, for memory considerations, may not need to upcast attention to fp32
     attn_weights = nn.functional.softmax(attn_weights, dim=-1)
+    if attn_weights.dtype != value.dtype:
+        attn_weights = attn_weights.to(value.dtype)
     attn_output = torch.matmul(attn_weights, value)
     return attn_output, attn_weights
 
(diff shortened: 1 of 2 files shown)


hkvision:
@rnwang04 Please take a look at it?

python/llm/src/ipex_llm/transformers/models/llama.py
```diff
@@ -1324,6 +1324,10 @@ def native_sdp(query, key, value, attention_mask,
 
     # at inference time, for memory considerations, may not need to upcast attention to fp32
     attn_weights = nn.functional.softmax(attn_weights, dim=-1)
+
+    if attn_weights.dtype != value.dtype:
+        attn_weights = attn_weights.to(value.dtype)
```
jason-dai, line 1329: Just do it unconditionally?

rnwang04:
This PR can fix llama, but will other model meet similar issues ? 
Shall we add `torch_dtype=torch.bfloat16` for  `load_in_low_bit='bf16'` as we did for fp16 too ?

ivy-lv11:
> This PR can fix llama, but will other model meet similar issues ? Shall we add `torch_dtype=torch.bfloat16` for `load_in_low_bit='bf16'` as we did for fp16 too ?

Yes

rnwang04 (approved):
LGTM

PR #10611, opened 2024-04-01
LLM: fix llama2 FP16 & bs>1 & autotp on PVC and ARC

plusbang:
## Description

Background: https://github.com/analytics-zoo/nano/issues/1228
Fix bugs of  fp16 & bs>1 using AutoTP on both arc and pvc

### 3. Summary of the change 

- For PVC, fix shape error
- For ARC, also disable `esimd_sdp` when bs>1

### 4. How to test?
- [x] Unit test
- [x] Local test (on both ARC and PVC)

### DIFF
--- a/python/llm/src/ipex_llm/transformers/models/utils.py
+++ b/python/llm/src/ipex_llm/transformers/models/utils.py
@@ -341,6 +341,9 @@ def use_esimd_sdp(q_len, k_len, head_dim, query_states, attention_mask=None):
     if query_states.shape[0] > 1 and device_name.startswith("Intel(R) Data Center GPU Max"):
         # esimd_sdp not support PVC GPU when batch size > 1 for now
         return False
+    if query_states.shape[0] > 1 and device_name.startswith("Intel(R) Arc(TM) A"):
+        # esimd_sdp not support ARC GPU when batch size > 1 for now
+        return False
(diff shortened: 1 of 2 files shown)


python/llm/src/ipex_llm/transformers/models/utils.py
```diff
@@ -332,6 +332,9 @@ def use_esimd_sdp(q_len, k_len, head_dim, query_states, attention_mask=None):
     if query_states.shape[0] > 1 and device_name.startswith("Intel(R) Data Center GPU Max"):
         # esimd_sdp not support PVC GPU when batch size > 1 for now
         return False
+    if query_states.shape[0] > 1 and device_name.startswith("Intel(R) Arc(TM) A"):
+        # esimd_sdp not support ARC GPU when batch size > 1 for now
+        return False
```
jason-dai, line 337: Shall we return false when query_states.shape[0] > 1?

plusbang: > Shall we return false when query_states.shape[0] > 1?

But this works and have performance optimization for bs>1 single A770 inference. I added the limitation now, only using deepspeed and bs>1 on ARC will disable.

MeouSker77 (approved):

rnwang04 (approved):

PR #10612, opened 2024-04-01
Fix starcoder first token perf

hkvision:
Fix https://github.com/analytics-zoo/nano/issues/1203

          (c_fc): LowBitLinear(in_features=6144, out_features=24576, bias=True)
          (c_proj): LowBitLinear(in_features=24576, out_features=6144, bias=True)

For starcoder mlp, it has large in/out features as well, `24576*6144>32000*4096`, larger than llama's lm_head. empty cache is executed within each transformer block and thus the first token latency is hugely impacted.
Add a condition for bias to exclude these layers. For lm_head seems it won't have bias.


lalalapotter (approved):
LGTM.

PR #10620, opened 2024-04-02
add python style check

liu-shaojun:
### 1. Why the change?

This submission adds python style check workflow and fix checks.


Oscilloscope98 (approved):
Let's wait for all UT passed

PR #10629, opened 2024-04-02
Add seq len check for llama softmax upcast to fp32

hkvision:
Fix https://github.com/intel-analytics/ipex-llm/issues/10598


jason-dai (approved):
LGTM

zhentaocc (approved):
1k input next token latency is improved from 22.54ms to 16.9ms with this change. (Xeon+kernel6.5+A770)

rnwang04 (approved):

PR #10633, opened 2024-04-02
Fix qwen-vl style

jenniew:
Fix qwen-vl style


jenniew:
The failed test is not related with this PR.

PR #10635, opened 2024-04-03
fix TUF invalid key bug in openssf check

liu-shaojun:
### 1. Why the change?

update ossf/scorecard-action to fix TUF invalid key bug: https://github.com/ossf/scorecard-action/issues/998


glorysdj (approved):
LGTM

PR #10640, opened 2024-04-03
Add Deepspeed TP Example of FLEX Mistral

Uxito-Ada:
## Description

This helps users to run deepspeed TP on FLEX platforms.

### 1. Why the change?

as above

### 2. User API changes

none

### 3. Summary of the change 

Add Deepspeed TP Example of FLEX Mistral

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [x] Application test
- [x] Document test
- [ ] ...


qiyuangong (approved):
LGTM

PR #10644, opened 2024-04-03
update spr perf test runner

liu-shaojun:
### 1. Why the change?

This submission 
- update nightly build to 23:00 China
- update the runner label of spr perf test
- update model path since runner change to spr14


glorysdj (approved):
LGTM

PR #10645, opened 2024-04-03
fix wrong cpu core num seen by qlora

Uxito-Ada:
## Description

After **ipex-llm-init**, nproc used to see 96 cores, for example, and only can see 48 which is equal to OMP_NUM_THREADS set by source oneAPI.
This may be a bug of ipex-llm-init, and for qlora, we fix it here by nproc get cpu core num before ipex-llm-init.

### 1. Why the change?

as above

### 2. User API changes

no

### 3. Summary of the change 

fix wrong cpu core num seen by qlora

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [x] Application test
- [x] Document test
- [ ] ...


liu-shaojun (approved):
LGTM

PR #10664, opened 2024-04-04
Fix vllm print error message issue

jenniew:
Fix issue: https://github.com/intel-analytics/ipex-llm/issues/10661
Also correct README.
Also fix all print error message issue.


yangw1234 (approved):

jenniew:
The test failure is not related with this PR.

PR #10681, opened 2024-04-07
Modify the .md link in chatchat readthedoc

Zhangky11:
Due to a certain regular matching mechanism in Sphinx, links with ".md" in the file are stripped of the ".md" extension, causing some links to fail to work properly. 

To solve the above problem, "#" is added to the end of links with ".md", so that the extension is not removed, and this method can successfully achieve redirection.


shane-huang (approved):
LGTM.

PR #10702, opened 2024-04-09
Fix low memory generation example issue in transformers 4.36

jenniew:
Fix low memory generation example issue in transformers 4.36.
Related issue: https://github.com/analytics-zoo/nano/issues/1157
Can support all transformers versions of 4.31+.
Test it under transformers 4.31 and 4.36.


yzheng124 (approved):
LGTM.

PR #10704, opened 2024-04-09
Support Mistral fp16 sdp when transformers >=4.36

cyita:
## Description

Support Mistral fp16 sdp when transformers >=4.36

### 1. Why the change?

Improve the performance of mistral half on GPU when transformers >=4.36

### 4. How to test?
- [x] local test


plusbang (approved):
LGTM

PR #10709, opened 2024-04-09
Update Env check Script

NovTi:
## Description

Update Env check Script


NovTi:
This is the current env-check.sh result


env-check2
env-check1

hkvision (approved):

PR #10730, opened 2024-04-10
Edit ppl update_HTML_parent_folder

MargarettMao:
ppl test and harness test are using the same update_in_parent_folder file. To reduce the repetition, change the ppl update HTML file to the same one under the harness folder and delete the HTML file under the ppl folder.


hkvision (approved):

hxsz1997 (approved):
LGTM

PR #10733, opened 2024-04-11
Fix transformers version dependabot issues

liu-shaojun:
### 1. Why the change?

https://github.com/intel-analytics/ipex-llm/security/dependabot/2897


xiangyuT (approved):
LGTM

PR #10738, opened 2024-04-11
[WIP] vLLM: Add ipex-llm entrypoint

xiangyuT:
## Description

Provide ipex_llm.serving.vllm.entrypoints to run vLLM. Need to install `vllm` first.

### DIFF
--- a/python/llm/src/ipex_llm/serving/vllm/cpu_executor.py
+++ b/python/llm/src/ipex_llm/serving/vllm/cpu_executor.py
@@ -1,0 +1,42 @@
+#
+# Copyright 2016 The BigDL Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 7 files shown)


python/llm/src/ipex_llm/optimize.py
```diff
@@ -230,9 +230,9 @@ def optimize_model(model, low_bit='sym_int4', optimize_llm=True, modules_to_not_
     invalidInputError(isinstance(model, torch.nn.Module),
                       "model should be an instance of "
                       f"`torch.nn.Module`, but got {type(model)} at last.")
-    invalidInputError(model.device.type in ('cpu', 'meta'),
-                      "Expect model on device `cpu` or `meta`, "
-                      f"but got device type {model.device.type}")
+    # invalidInputError(model.device.type in ('cpu', 'meta'),
```
gc-fu, line 233: Consider change to this?
```python
    if hasattr(model, 'device'):
        invalidInputError(model.device.type in ('cpu', 'meta'),
                        "Expect model on device `cpu` or `meta`, "
                        f"but got device type {model.device.type}")
```

PR #10742, opened 2024-04-11
Fix wrong output For Llama-2-7b-chat-hf on CPU

gc-fu:
## Description

Fix wrong output for Llama-2-7b-chat-hf

Related issues: https://github.com/intel-analytics/ipex-llm/issues/10724

Result on generate.py:

```

[INST] <<SYS>>
You are a helpful assistant.
<</SYS>>
What is AI?[/INST]

[INST] <<SYS>>
You are a helpful assistant.
<</SYS>>
What is AI?[/INST]
Ah, a great question! AI, or Artificial Intelligence, refers to the development of computer systems that can perform tasks that typically require human intelligence

```

### DIFF
--- a/python/llm/src/ipex_llm/transformers/models/llama.py
+++ b/python/llm/src/ipex_llm/transformers/models/llama.py
@@ -1327,10 +1327,22 @@ def llama_attention_forward_4_36_original(
         attn_weights = None
     else:
         # otherwise, use native attention
-        attn_output, attn_weights = native_sdp(query_states, key_states, value_states,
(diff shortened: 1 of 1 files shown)


jenniew:
For my new findings, let's take a closer look at what's the exact reason.

jenniew:
Please test if output_attentions=True, the result is correct too.

jenniew (approved):
LGTM

gc-fu:
Tested with `output_attentions=True` and `output_attentions=False`, both are correct

PR #10746, opened 2024-04-11
Fix baichuan-13b issue on portable zip under transformers 4.36

jenniew:
Related issue: https://github.com/analytics-zoo/nano/issues/1083
* For Baichuan-13B-Chat, cannot support stream chat with past_key_values, change back to use stream chat with chat history(add history to prompt).
* When create StartRecentKVCache, set bigger recent_size instead of 512(default value) to store more previous key_values for multi-run chat. 
* Also updated Readme by add more supported models.

Test on CPU.


python/llm/portable-zip/chat.py
```diff
@@ -276,31 +321,43 @@ def auto_select_model(model_name):
...
 
     model_path = args.model_path
     start_size = args.start_size
+    recent_size = args.recent_size
 
     model = auto_select_model(model_path)
     model = optimize_model(model)
 
     tokenizer = AutoTokenizer.from_pretrained(model_path, trust_remote_code=True)
+    print("model.config.architectures[0] is: ", model.config.architectures[0])
```
plusbang, line 335: I feel these print messages could be removed?

python/llm/portable-zip/chat.py
```diff
@@ -276,31 +321,43 @@ def auto_select_model(model_name):
...
     elif model.config.architectures is not None and model.config.architectures[0] == "LlamaForCausalLM":
-        kv_cache = StartRecentKVCache(start_size=start_size)
+        kv_cache = StartRecentKVCache(start_size=start_size, recent_size=recent_size)
         if "yi" in model_path.lower():
             stop_words = get_stop_words_ids("Yi", tokenizer=tokenizer)
             yi_stream_chat(model=model, tokenizer=tokenizer, kv_cache=kv_cache, stop_words=stop_words)
         else:
             llama_stream_chat(model=model, tokenizer=tokenizer, kv_cache=kv_cache)
+    elif model.config.architectures[0] == "BaichuanForCausalLM" and model.config.vocab_size == 64000:
+        # Baichuan-13B-Chat
```
plusbang, line 356: Only one concern, by using the previous chat history mode, the 1st token latency will by much longer after several turns.

jenniew: Yes, but Baichuan-13B cannot use our past_key_values method.

jenniew: Other models will use our kv_cache method

python/llm/portable-zip/chat.py
jenniew: fixed

plusbang (approved):

PR #10747, opened 2024-04-11
Fix internlm-chat-7b-8k repo name in examples

jenniew:
The repo "internlm/internlm-chat-7b-8k" does not exist now. So do the corresponding changes from "internlm/internlm-chat-7b-8k" to "internlm/internlm-chat-7b".

Verify the gpu/cpu examples of repo "internlm/internlm-chat-7b".


hzjane (approved):
LGTM

PR #10766, opened 2024-04-15
LLM: fix qwen AutoTP

plusbang:
## Description

Background: https://github.com/intel-analytics/ipex-llm/issues/10763

There exists shape error when run Qwen-7b or Qwen-14b using AutoTP.
The root cause is our `split_qkv_proj_func` not compatible with deepspeed splitting (https://github.com/microsoft/DeepSpeed/pull/4902/files).

Take Qwen-14b for example, after deepspeed splitting, the shape of `c_attn_weight` is [7680, 5120], but still use `projection_size=5120` to split q/k/v.

So change the `projection_size` to fix.

### 4. How to test?
- [x] Unit test
- [x] Local test


rnwang04 (approved):
LGTM

PR #10777, opened 2024-04-17
Fix spec error caused by lookup pr

cyita:
## Description

Fix spec error caused by lookup pr(https://github.com/intel-analytics/ipex-llm/pull/10768)
If the user passes attn mask to generate, will cause an error.

### 4. How to test?
- [x] local test


Romanticoseu:
LGTM

rnwang04 (approved):
LGTM

PR #10784, opened 2024-04-17
Transformers ppl evaluation on wikitext

hkvision:
python run_wikitext.py --model_path /mnt/disk1/Llama-2-7b-chat-hf/ --dataset path=wikitext,name=wikitext-2-raw-v1 --precision sym_int4

Datasets version 2.14.6


glorysdj (approved):
LGTM

PR #10787, opened 2024-04-18
use config rope_theta

yangw1234:
## Description
use config rope_theta in llama


MeouSker77 (approved):

yangw1234:
The failed case seems irrelevant. I'll merge it first.

PR #10841, opened 2024-04-22
PPO Finetuning Example

Jasonzzt:
## Description

This is an example of TRL PPO, a kind of RLHF, where uses dataset HuggingFaceH4/helpful_instructions that on our whitelist.

### 1. Why the change?

- Enable PPO on Intel GPU with IPEX LLM.
- Add class AutoModelForCausalLMWithValueHead from trl to IPEX-LLM
- fix readme of lisa finetuning

### 3. Summary of the change 

PPO Finetuning Example

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [x] Application test
- [x] Document test
- [ ] ...

### 5. New dependencies

- [x] New Python dependencies
       - trl

### DIFF
--- a/python/llm/example/GPU/LLM-Finetuning/LISA/README.md
+++ b/python/llm/example/GPU/LLM-Finetuning/LISA/README.md
@@ -38,11 +38,11 @@ python ./lisa_finetuning.py \
 
 Optional parameters for `lisa_finetuning.py`:
 
-**--repo-id-or-model-path** : default to `meta-llama/Llama-2-7b-hf`, and you can also specify your local model path.
+**--base_model** : default to `meta-llama/Llama-2-7b-hf`, and you can also specify your local model path.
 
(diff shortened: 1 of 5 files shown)


Uxito-Ada:
@Jasonzzt From the log, it is found that PPO also applies PEFT LoRA.
Therefore, like QLoRA, rather than `from_pretrained` a peft model with lora config, we should first load the base model, and then use `get_peft_model`, `prepare_model_for_kbit_training` etc. methods  in [qlora.py](https://github.com/intel-analytics/ipex-llm/blob/main/python/llm/src/ipex_llm/transformers/qlora.py) to create a peft model. Such a model is built on top of layers with supported operators like [here](https://github.com/intel-analytics/ipex-llm/blob/main/python/llm/src/ipex_llm/transformers/qlora.py#L119).

PR #10843, opened 2024-04-22
add phi-2 optimization

MeouSker77:
## Description

add phi-2 optimization: merge_qkv, fuse rope and normal kv cache. sdp and quantize kv are not supported because its head_dim is 80


rnwang04 (approved):
LGTM

MeouSker77:
this model requires to compute sdp with fp32, so fp32+int4 is faster than fp16+int4 (if use fp16, we must convert q and k to fp32 dynamically before computing sdp)

PR #10855, opened 2024-04-23
Fix speculative decoding bug

Romanticoseu:
## Description

IPEX_CPU optimized models have issues for speculative decoding with short prompts, and in warmup, `draf_model` will be deleted, so we should remove these attributes in `else

### 4. How to test?
- [x] Application test

### DIFF
--- a/python/llm/src/ipex_llm/transformers/speculative.py
+++ b/python/llm/src/ipex_llm/transformers/speculative.py
@@ -100,6 +100,12 @@ def generate(
                                          draft_model=self.draft_model,
                                          **new_speculative_kwargs)
     else:
+        # When `draft_model` is false, these attributes
+        # ralted to speculative decoding should be removed
+        for var in ['max_step_draft', 'th_stop_draft', 'hf_adjust',
+                    'auto_th_stop_draft', 'auto_parameters', 'min_step_draft',
+                    'th_batch_num']:
+            value = kwargs.pop(var, None)
         return original_generate(self,
                                  inputs=inputs,
                                  generation_config=generation_config,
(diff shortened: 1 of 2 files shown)


python/llm/src/ipex_llm/transformers/speculative.py
```diff
@@ -100,6 +100,12 @@ def generate(
                                          draft_model=self.draft_model,
                                          **new_speculative_kwargs)
     else:
+        # When `draft_model` is false, these attributes
+        # ralted to speculative decoding should be removed
```
qiyuangong, line 104: related

python/llm/src/ipex_llm/transformers/speculative.py
```diff
@@ -100,6 +100,12 @@ def generate(
...
+        for var in ['max_step_draft', 'th_stop_draft', 'hf_adjust',
+                    'auto_th_stop_draft', 'auto_parameters', 'min_step_draft',
+                    'th_batch_num']:
+            value = kwargs.pop(var, None)
```
qiyuangong, line 108: If value is not used, remove value

python/llm/src/ipex_llm/transformers/speculative.py
```diff
@@ -74,8 +74,7 @@ def generate(
             for var in ['max_step_draft', 'th_stop_draft', 'hf_adjust',
                         'auto_th_stop_draft', 'auto_parameters', 'min_step_draft',
                         'th_batch_num']:
-                value = kwargs.pop(var, None)
-            del self.draft_model
```
glorysdj, line 78: self.draft_model should keep for next round inference.

glorysdj (approved):
LGTM

PR #10859, opened 2024-04-23
Add patch for fastchat

JinBridger:
Needed by https://github.com/intel-analytics/Langchain-Chatchat/pull/27


python/llm/src/ipex_llm/serving/fastchat/patch_fastchat.py
```diff
@@ -0,0 +1,204 @@
+#
```
Oscilloscope98, line 1: Do not forget the original bigdl lisence :)

JinBridger:
Closed because this problem is caused by model itself. Use latest model to fix this issue.

PR #10865, opened 2024-04-23
Add phi-2 to igpu performance test

Oscilloscope98:
## Description

Did not add it for fp16+int4 test as it has worse performance


liu-shaojun (approved):
LGTM

PR #10868, opened 2024-04-24
LLM: Install CPU version torch with extras [all]

xiangyuT:
## Description

Set torch version in `all_requirements` to install cpu version torch instead of cuda version on CPU.


Oscilloscope98 (commented):
Could the `torch==2.0.1+cpu` be successfully installed without `--index-url https://download.pytorch.org/whl/cpu`? If not, we may also update the UT, documentation, examples, etc.

Besides, is there any difference between Windows and Linux?

Oscilloscope98:
Seems like the UT failed after this change: https://github.com/intel-analytics/ipex-llm/actions/runs/8809956275/job/24181761869?pr=10868

xiangyuT:
> Could the `torch==2.0.1+cpu` be successfully installed without `--index-url https://download.pytorch.org/whl/cpu`? If not, we may also update the UT, documentation, examples, etc.
> 
> Besides, is there any difference between Windows and Linux?

No, it needs to add `--extra-index-url https://download.pytorch.org/whl/cpu` during installation. And the UT's failure seems the same reason: the install action is missing `--extra-index-url` setting. (refer to [here](https://github.com/intel-analytics/ipex-llm/blob/main/.github/actions/llm/setup-llm-env/action.yml#L41))

The installation on Windows is not tested yet.

python/llm/setup.py
```diff
@@ -50,8 +50,10 @@
 llm_home = os.path.join(os.path.dirname(os.path.abspath(__file__)), "src")
 github_artifact_dir = os.path.join(llm_home, '../llm-binary')
 libs_dir = os.path.join(llm_home, "ipex_llm", "libs")
+
+all_torch_version = 'torch == 2.0.1+cpu'
```
qiyuangong, line 54: all torch version is not a good name for this variable. Maybe change to CPU torch version

qiyuangong:
Can we add `-extra-index-url` into `setup.py` or `requirements.txt`?

python/llm/setup.py
```diff
@@ -50,8 +50,13 @@
 llm_home = os.path.join(os.path.dirname(os.path.abspath(__file__)), "src")
 github_artifact_dir = os.path.join(llm_home, '../llm-binary')
 libs_dir = os.path.join(llm_home, "ipex_llm", "libs")
+
+cpu_torch_version_linux = 'torch == 2.1.2+cpu'
+cpu_torch_version_win = 'torch == 2.1.2'
+global cpu_torch_version
+cpu_torch_version = cpu_torch_version_linux
```
Oscilloscope98, line 57: Should we use
["torch==2.1.2+cpu;platform_system=='Linux'", "torch==2.1.2;platform_system=='Windows'" instead]

qiyuangong (approved):
LGTM

Oscilloscope98 (approved):

PR #10872, opened 2024-04-24
LLM: update split tensor conditions.

lalalapotter:
## Description

This PR is to update split tensor conditions:
1. enable when "IPEX_LLM_SPLIT_QKV" is enabled.
2. enable when intermedia `attn_weights` size larger than block memory limitation.
3. update quantize kv cache conditions.

### DIFF
--- a/python/llm/src/ipex_llm/transformers/models/chatglm2.py
+++ b/python/llm/src/ipex_llm/transformers/models/chatglm2.py
@@ -97,13 +97,13 @@ def repeat_kv(key: torch.Tensor, value: torch.Tensor, n_head: int) -> (torch.Ten
 
 
 def should_split_qkv_tensor(query_layer, bsz, n_head, seq_len):
-    if os.environ.get("IPEX_LLM_LOW_MEM", None) == "1":
-        return True
-    if query_layer.dtype == torch.float16 and query_layer.shape[2] >= 5000:
+    if os.environ.get("IPEX_LLM_SPLIT_QKV", None) is not None:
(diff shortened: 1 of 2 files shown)


python/llm/src/ipex_llm/transformers/models/chatglm2.py
```diff
@@ -96,6 +96,19 @@ def repeat_kv(key: torch.Tensor, value: torch.Tensor, n_head: int) -> (torch.Ten
...
 
 
+def should_split_qkv_tensor(query_layer, bsz, n_head, seq_len):
+    if os.environ.get("IPEX_LLM_LOW_MEM", None) == "1":
+        return True
+    if query_layer.dtype == torch.float16 and query_layer.shape[2] >= 5000:
+        # split tensor for memory block limitation
+        # support fp16 and set input length threshold at 5000 for now
+        return True
+    if query_layer.element_size()*bsz*n_head*seq_len*seq_len >= 4*1024*1024*1024:
```
hkvision, line 106: will this check hurt performance? query_layer.element_size(), should be no?

lalalapotter: Seems not:
use condition with element_size():
0,meta-llama/Llama-2-7b-chat-hf,4839.14,27.14,0.0,8192-512,1,8192-512,1,fp8,False,4.49,11.966796875,N/A,N/A
use condition with shape:
0,meta-llama/Llama-2-7b-chat-hf,4837.71,27.13,0.0,8192-512,1,8192-512,1,fp8,False,4.48,11.966796875,N/A,N/A

hkvision (approved):

PR #10876, opened 2024-04-24
LLM: Enable batch generate (world_size>1) in Deepspeed-AutoTP-FastAPI example

xiangyuT:
## Description

Enable batch generate (world_size>1) in Deepspeed-AutoTP-FastAPI example.
Use `request_queue` and `result_dict` to store and return requests' result.


glorysdj (approved):
LGTM

PR #10879, opened 2024-04-25
Add tensor parallel for vLLM

gc-fu:
## Description

This PR adds tp support for vLLM in `ipex_llm` side.

## Verified low_bit format

- [x] sym_int4
- [x] fp8
- [x] fp16


python/llm/src/ipex_llm/transformers/convert.py
```diff
@@ -49,6 +49,7 @@
 import os
 from ipex_llm.utils.common import invalidInputError
 from typing import List, Optional, Tuple, Union
+import ray
```
xiangyuT, line 52: Is ray always needed for ipex_llm?

gc-fu: Good catch, will delete

glorysdj (approved):
LGTM

PR #10882, opened 2024-04-25
Use new fp16 sdp in Qwen and modify the constraint

cyita:
## Description

Use new fp16 sdp in Qwen and modify the constraint.

### 1. Why the change?

It shows that the new fp16 SDP won't cause an error on 2 card arcs when inference using deepspeed.

### 4. How to test?
- [x] Local test


MeouSker77 (approved):

plusbang (approved):
LGTM

PR #10889, opened 2024-04-26
Unpin pydantic version

liu-shaojun:
### 1. Why the change?

https://github.com/intel-analytics/ipex-llm/security/dependabot/2914


glorysdj (approved):
LGTM

PR #10891, opened 2024-04-26
Fix apt install oneapi scripts

liu-shaojun:
### 1. Why the change?

This commit sets the apt installer as the default way to install OneAPI on Linux, for two reasons: 
- Firstly, we originally wanted users to install it via pip install because it's simpler. However, pip install only satisfies the runtime needs, not the compilation requirements. Some of our features, such as vLLM, require a compilation environment, which cannot be achieved through pip install. 
- Secondly, most of our internal colleagues require a compilation environment for development. Therefore, we need to maintain environments for both apt and pip installation methods. Providing a compilation environment only for certain machines would lead to resource contention and insufficient machine availability. Moreover, maintaining both methods simultaneously can cause confusion. 

Therefore, for the sake of simplicity for users and colleagues, we have changed the default recommended method to apt installation.

Some users previously reported issues with apt installation method: 
https://github.com/intel-analytics/ipex-llm/issues/10825 
https://github.com/intel-analytics/ipex-llm/issues/10550 
https://github.com/analytics-zoo/nano/issues/1230. 
This commit also addresses the missing packages to resolve the issue of package absence.

### DIFF
--- a/docs/readthedocs/source/doc/LLM/Overview/install_gpu.md
+++ b/docs/readthedocs/source/doc/LLM/Overview/install_gpu.md
@@ -240,8 +240,11 @@ IPEX-LLM GPU support on Linux has been verified on:
                   intel-oneapi-mpi=2021.11.0-49493 \
                   intel-oneapi-mpi-devel=2021.11.0-49493 \
                   intel-oneapi-dal=2024.0.1-25 \
+                  intel-oneapi-dal-devel=2024.0.1-25 \
                   intel-oneapi-ippcp=2021.9.1-5 \
+                  intel-oneapi-ippcp-devel=2021.9.1-5 \
                   intel-oneapi-ipp=2021.10.1-13 \
+                  intel-oneapi-ipp-devel=2021.10.1-13 \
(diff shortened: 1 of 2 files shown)


docs/readthedocs/source/doc/LLM/Overview/install_gpu.md
```diff
@@ -215,6 +215,47 @@ IPEX-LLM GPU support on Linux has been verified on:
...
+                  intel-oneapi-dnnl=2024.0.0-49521 \
+                  intel-oneapi-tcm-1.0=1.0.0-435
+
+            .. note::
+
+               You can uninstall the package by running the following command:
+
+               .. code-block:: bash
+               
+                  sudo apt autoremove intel-oneapi-common-vars
```
hkvision, line 257: will this remove everything?

hkvision (approved):

docs/readthedocs/source/doc/LLM/Overview/install_gpu.md
liu-shaojun: yes

PR #10899, opened 2024-04-28
Fix nightly perf

Oscilloscope98:
Fix nightly perf by adding default value in benchmark for `use_fp16_torch_dtype` and `n_gpu`


plusbang (approved):
LGTM

PR #10907, opened 2024-04-29
Fix loader issue when loading fp16

gc-fu:
## Description

This fixes the loader issue when using low_bit fp16.

Related issue: https://github.com/intel-analytics/ipex-llm/issues/10905


hzjane (approved):
LGTM

PR #1096, opened 2017-06-26
Add markdown documentation

psyyz10:
## What changes were proposed in this pull request?

Add markdown documentation

PR #10917, opened 2024-04-30
LLM: Add Pipeline-Parallel-FastAPI example

xiangyuT:
## Description

Implement refer to [#1096](https://github.com/analytics-zoo/nano/issues/1096#issuecomment-1995756893)


python/llm/example/GPU/Pipeline-Parallel-FastAPI/run.sh
```diff
@@ -0,0 +1,13 @@
...
+export FI_PROVIDER=tcp
+export OMP_NUM_THREADS=8
+
+export USE_XETLA=OFF
+export SYCL_PI_LEVEL_ZERO_USE_IMMEDIATE_COMMANDLISTS=2
+export TORCH_LLM_ALLREDUCE=0
+export CCL_DG2_ALLREDUCE=1 # For internal CCL
+
+export MODEL_PATH=1
+CCL_ROOT=./1ccl_dg2_allreduce_20240308  LD_LIBRARY_PATH=./1ccl_dg2_allreduce_20240308/src:./1ccl_dg2_allreduce_20240308/deps/mpi/lib:/opt/intel/oneapi/2024.0/lib \
```
yangw1234, line 12: These are not public available packages right?

xiangyuT: Yes, will remove

python/llm/example/GPU/Pipeline-Parallel-FastAPI/pipeline_models.py
```diff
@@ -0,0 +1,304 @@
...
+    
+                
+    def generate(self, input_ids=None, max_tokens=5, attention_mask=None):
+        times = []
+        with torch.no_grad():
+            _input_ids = None
+            _past_key_values = None
+            bs = input_ids.shape[0]
+            output_ids = input_ids.clone()
+            for i in range(max_tokens):
```
yangw1234, line 263: I don't think this is truly pipeline parallel. 

Let's take the first half of the model for example assuming PP=2, the model takes `input_ids` and run forward on line 268, and it then blocking waits on line 282 for the next token while the 2nd part of the model is running to generate the next token. 

When `generate` finished generating all the tokens, it starts to work on the next request, so the GPU is idle half of the time.

yangw1234:
Maybe we should wrap the model implementation in IPEX-LLM package and only expose an API for users to call in the example.

e.g. in the example code:

for running offline batch processing task in pipeline parallel
```python
from ipex-llm.pipeline.models import LlamaForCausalLM
from ipex-llm.pipeline import PP_Config
config = PP_Config(...)
model = LlamaForCausalLM.from_pretrained(pp_config=config, ...)
prompts =["abc"] * batch_size
batches = [prompts] * num_of_batches
output = model.generate(batches)
```

for running online iterative task
```python
from ipex-llm.pipeline.models import LlamaForCausalLM
from ipex-llm.pipeline import PP_Config
config = PP_Config(...)
model = LlamaForCausalLM.from_pretrained(pp_config=config, ...)

while true:
    if model.should_get_new_batch(queue)
        new_batch = queue.get()
        output = model.step(new_batch)
    else:
       output = model.step()
    yield output
```

glorysdj (approved):
LGTM

plusbang:
Maybe we could merge this first and refactor the model implementation to source code later.

PR #10920, opened 2024-04-30
Fix Langchain upstream ut

Zhangky11:
Fix Langchain upstream ut for Langchain


shane-huang (commented):
LGTM.

hkvision:
Follow-up work in https://github.com/intel-analytics/ipex-llm/pull/10985

PR #10928, opened 2024-05-06
improve ipex-llm-init for Linux

liu-shaojun:
### 1. Why the change?

Issue: https://github.com/analytics-zoo/nano/issues/1291

This submission updates ipex-llm-init for Linux so that user can use ipex-llm-init to set related ENVs for different devices and print info.

### DIFF
--- a/python/llm/scripts/ipex-llm-init
+++ b/python/llm/scripts/ipex-llm-init
@@ -47,7 +47,6 @@ function unset_gpu_envs {
     unset SYCL_CACHE_PERSISTENT
     unset BIGDL_LLM_XMX_DISABLED
     unset SYCL_PI_LEVEL_ZERO_USE_IMMEDIATE_COMMANDLISTS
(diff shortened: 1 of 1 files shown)


glorysdj (approved):
LGTM

python/llm/scripts/ipex-llm-init
```diff
@@ -88,94 +103,68 @@ disable_tcmalloc
...
 done
 
-shift $((OPTIND -1))
-
 # Find ipex-llm-init dir
 if [ ! -z $BASH_SOURCE ]; then
     # using bash
     if [ "$BASH_SOURCE" = "$0" ]; then
         echo "Error: Incorrect usage: ipex-llm-init must be sourced."
-        exit 1
```
MeouSker77, line 165: `return` can only be used in a function or sourced script.

This code is checking whether user is sourcing script or directly executing script, if use `return` and user is executing, it will report an error like `return: can only return from a function or sourced script`, not exits with error code 1, so we should use `exit 1` here.

others LGTM

MeouSker77 (approved):

Oscilloscope98 (approved):

PR #10953, opened 2024-05-07
optimize phi3 again: use quantize kv if possible

MeouSker77:
## Description

optimize phi3 again: use quantize kv if possible

### DIFF
--- a/python/llm/src/ipex_llm/transformers/models/phi3.py
+++ b/python/llm/src/ipex_llm/transformers/models/phi3.py
@@ -39,8 +39,10 @@ from ipex_llm.transformers.models.utils import (
     rotate_half, should_use_fuse_rope,
     apply_rotary_pos_emb_cache_freq_xpu
 )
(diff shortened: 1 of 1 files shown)


python/llm/src/ipex_llm/transformers/models/phi3.py
```diff
@@ -179,8 +189,12 @@ def model_forward(
     ):
         # IPEX-LLM OPT: kv cache but no sdp (its head_dim 96, cannot use sdp)
         use_cache = use_cache if use_cache is not None else self.config.use_cache
+        use_quantize_kv = (use_quantize_kv_cache(self.layers[0].mlp.down_proj, input_ids) and
+                           self.config.hidden_size // self.config.num_attention_heads in [64, 128])
```
rnwang04, line 193: maybe put `self.config.hidden_size // self.config.num_attention_heads in [64, 128` this check into `use_quantize_kv_cache`  ？

MeouSker77: yes, we'll move it later

rnwang04 (approved):

PR #10972, opened 2024-05-09
Empty cache after phi first attention to support 4k input

hkvision:
Fix https://github.com/analytics-zoo/nano/issues/1348

### DIFF
--- a/python/llm/src/ipex_llm/transformers/convert.py
+++ b/python/llm/src/ipex_llm/transformers/convert.py
@@ -194,6 +194,7 @@ def empty_cache_post(module, input, output):
     except:  # cpu
         pass
 
+
 def convert_gptq(module, awq=False, llm_awq=False, act_order=False):
     from ipex_llm.transformers.low_bit_linear import get_block_size
     Q4_1 = get_block_size("asym_int4")
--- a/python/llm/src/ipex_llm/transformers/convert.py
+++ b/python/llm/src/ipex_llm/transformers/convert.py
@@ -188,6 +188,12 @@ def is_linear_module(module):
     return result, (in_features, out_features, mp_group)
 
 
+def empty_cache_post(module, input, output):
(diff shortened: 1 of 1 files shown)


hkvision:
Test result is here: https://github.com/analytics-zoo/nano/issues/1348#issuecomment-2102086271
Performance is not impacted. If there are issues, may add this to low memory mode. But at this moment, add this by default for all cases.

python/llm/src/ipex_llm/transformers/convert.py
```diff
@@ -188,6 +188,12 @@ def is_linear_module(module):
     return result, (in_features, out_features, mp_group)
 
 
+def empty_cache_post(module, input, output):
+    try:
```
jason-dai, line 192: why not check if it's xpu here?

hkvision: will it hurt performance for the xpu check? I will test it.

hkvision: Seems the impact is not obvious, https://github.com/analytics-zoo/nano/issues/1348#issuecomment-2102086271
using xpu check may be slightly slower than using this way, >5ms for 4k first token latency.

Do I need to change to xpu check?

MeouSker77 (approved):
LGTM after fixing style check

lalalapotter (approved):
LGTM.

hkvision:
Merge this first. Can further modify if necessary later.

hkvision:
http://10.239.44.136:8888/nightly_perf_gpu/transformer_int4_gpu-results-2024-05-09.html Yesterday's performance drop should not relate to this PR, the same behavior already happens on 0505, need further observation.
If there are issues, can restrict the empty cache to first token only, but from my test result, the impact for rest is ignorable.

PR #10985, opened 2024-05-11
Fix Langchain upstream ut

hkvision:
Follow this one: https://github.com/intel-analytics/ipex-llm/pull/10920
Fix the git checkout environment variable issue. Can not directly use the variable set in the previous step, instead need to use github output from the previous step, reference: https://stackoverflow.com/questions/59191913/how-do-i-get-the-output-of-a-specific-step-in-github-actions
https://stackoverflow.com/questions/67957150/using-env-variable-as-ref-in-actions-checkoutv2


liu-shaojun (approved):
LGTM

PR #10987, opened 2024-05-11
Add chatchat and text-gen web-ui on xpu-serving images

hzjane:
## Description
Add chatchat and text-gen web-ui on xpu-serving images

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [ ] Document test
- [ ] ...

### 5. New dependencies

- [ ] New Python dependencies
       - Dependency1 
       - Dependency2
       - ...
       - Dependency1 and license1
       - Dependency2 and license2
       - ...


jason-dai:
For serving, I think we only need FastAPI/FastChat/vLLM, maybe also AutoTP support; I don't think we need langchain-chatchat and text-gen-webui

glorysdj:
> For serving, I think we only need FastAPI/FastChat/vLLM, maybe also AutoTP support; I don't think we need langchain-chatchat and text-gen-webui

OK, get it.

PR #11007, opened 2024-05-14
use new sdp and fp32 sdp

MeouSker77:
## Description

use new sdp (supports fp16 and fp32) instead of old one (only supports fp16), baichuan2 and chatglm2 may use bool attention_mask, need to verify, will change them in next PR if possible


rnwang04 (approved):
LGTM

PR #11010, opened 2024-05-14
Update igpu performance test to use pypi installed oneAPI

Oscilloscope98:
## Description

Update igpu performance test to use pypi installed oneAPI, which should not bring results change in performance tests

We need to do this as we have included oneAPI related pypi packages in `[xpu]` option on Windows for `ipex-llm` version >= `2.1.0b20240514`


plusbang (approved):
LGTM

PR #11011, opened 2024-05-14
[LLM]: Add llama3 finetuning example based on alpaca qlora example

Jasonzzt:
## Description

Add a simple example for llama3 finetuning based on alpaca qlora example, and the finetuning process during testing is normal.

### 1. Why the change?

- add script for llama3 finetuning on single Arc A770
- modify readme

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [x] Application test
- [x] Document test
- [ ] ...


qiyuangong (approved):
LGTM

PR #11018, opened 2024-05-14
Update environment variable setting in AutoTP with arc

plusbang:
## Description

According to benchmark experiments on multi ARC with different type CPU, `SYCL_PI_LEVEL_ZERO_USE_IMMEDIATE_COMMANDLISTS` setting have different infulence.

### 4. How to test?
- [x] Local test

### DIFF
--- a/python/llm/dev/benchmark/all-in-one/run-deepspeed-arc.sh
+++ b/python/llm/dev/benchmark/all-in-one/run-deepspeed-arc.sh
@@ -10,7 +10,11 @@ source $basekit_root/ccl/latest/env/vars.sh --force
 
 NUM_GPUS=2 # number of used GPU
 export USE_XETLA=OFF
-export SYCL_PI_LEVEL_ZERO_USE_IMMEDIATE_COMMANDLISTS=2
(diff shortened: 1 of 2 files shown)


python/llm/dev/benchmark/all-in-one/run-deepspeed-arc.sh
```diff
@@ -10,7 +10,11 @@ source $basekit_root/ccl/latest/env/vars.sh --force
 
 NUM_GPUS=2 # number of used GPU
 export USE_XETLA=OFF
-export SYCL_PI_LEVEL_ZERO_USE_IMMEDIATE_COMMANDLISTS=2
+if grep -q "Xeon" /proc/cpuinfo; then
+    export SYCL_PI_LEVEL_ZERO_USE_IMMEDIATE_COMMANDLISTS=0
```
hkvision, line 14: Can we just don't set it in this case?

plusbang: > Can we just don't set it in this case?

Have changed to only set it to be 2 when Core CPU.

hkvision (approved):

PR #11025, opened 2024-05-15
Use new sdp again

MeouSker77:
## Description

change chatglm2/3 and baichuan2 to use new sdp, and check `head_dim` in `use_sdp_causal`


python/llm/src/ipex_llm/transformers/models/baichuan2.py
```diff
@@ -166,9 +167,8 @@ def baichuan_attention_forward_7b_quantized(
 
     past_key_value = (key_states, value_states) if use_cache else None
 
-    if attention_mask is not None:
-        if attention_mask.dtype == torch.bool:
-            attention_mask.masked_fill_(attention_mask.logical_not(), float("-inf"))
+    invalidInputError(attention_mask is None or attention_mask.dtype != torch.bool,
```
rnwang04, line 170: I think we can manually turn attention_mask from bool into float ?

rnwang04 (approved):

python/llm/src/ipex_llm/transformers/models/baichuan2.py
MeouSker77: baichuan2-7b is using float mask, it doesn't use `bool` mask

PR #11032, opened 2024-05-15
add fp6 mlp fusion

rnwang04:
## Description

https://github.com/intel-analytics/llm.cpp/pull/403

### 1. Why the change?

add mlp fusion for fp6.


cyita (approved):
LGTM

PR #11051, opened 2024-05-16
Make install ut for cpu exactly the same as what we want for users

Oscilloscope98:
## Description

Make install ut for cpu exactly the same as what we want for users

### How to test?
- [x] Unit test


xiangyuT (approved):
LGTM

PR #11055, opened 2024-05-16
Align ppl with llama.cpp

hkvision:
Modifications:
- Directly read data from raw file
- Add the no stride version like llama.cpp, and at the same time keep the stride version same as before.


sgwhat (approved):
LGTM

PR #11059, opened 2024-05-17
disable mlp fusion of fp6 on mtl

rnwang04:
## Description

disable mlp fusion of fp6 on mtl as it causes performance regression.

### 1. Why the change?
https://github.com/intel-analytics/llm.cpp/pull/407


cyita (approved):
LGTM

PR #11060, opened 2024-05-17
Update for cpu install option in performance tests

Oscilloscope98:
## Description

Update for cpu install option in performance tests


liu-shaojun (approved):
LGTM

PR #11061, opened 2024-05-17
Update docker image tags in Docker Quickstart

liu-shaojun:
### 1. Why the change?

- Update docker image tags from `2.1.0-SNAPSHOT` to `latest` to ensure consistency in tags
- Add a note to recommend Windows users with MTL iGPU use Anaconda Prompt + pip install solution
image


glorysdj (approved):
LGTM

PR #11074, opened 2024-05-20
refactor qwen

MeouSker77:
## Description

refactor qwen: use sdp_causal and new rope: 1st token perf improvement on Arc and MTL, 2nd+ token perf improvement on MTL


MeouSker77:
will fix qwen logits error in next PR, for now, just use a larger error bound

PR #11083, opened 2024-05-21
LLM: Add CPU vLLM entrypoint

xiangyuT:
## Description

Add CPU vLLM entrypoint and update CPU vLLM serving example.


glorysdj (approved):
LGTM

PR #11096, opened 2024-05-22
Fix Arc UT

Oscilloscope98:
image

`setuptools<70.0.0` will resolve this import problem

TODO: We may add related troubleshooting (if default setuptools installed by conda may also be >=70.0.0)


liu-shaojun (approved):
LGTM

PR #11098, opened 2024-05-22
Update linux igpu run script

leonardozcm:


python/llm/dev/benchmark/all-in-one/run-igpu.sh
```diff
@@ -0,0 +1,10 @@
+source /opt/intel/oneapi/setvars.sh
+ 
+export USE_XETLA=OFF
+export SYCL_CACHE_PERSISTENT=1
+KERNEL_VERSION=$(uname -r)
+if [[ $KERNEL_VERSION != *"6.5"* ]]; then
+    export SYCL_PI_LEVEL_ZERO_USE_IMMEDIATE_COMMANDLISTS=1
```
Oscilloscope98, line 7: I am not sure whether this `SYCL_PI_LEVEL_ZERO_USE_IMMEDIATE_COMMANDLISTS=1` is useful for igpu on Linux. @MeouSker77 What do you think? :)

MeouSker77: not sure too

Oscilloscope98 (approved):
Others LGTM

python/llm/dev/benchmark/all-in-one/run-igpu.sh
```diff
@@ -0,0 +1,6 @@
+source /opt/intel/oneapi/setvars.sh
+ 
+export USE_XETLA=OFF
+export SYCL_CACHE_PERSISTENT=1
```
Oscilloscope98, line 4: Seems like we should have
```
export SYCL_CACHE_PERSISTENT=1
export BIGDL_LLM_XMX_DISABLED=1
```

image

PR #11107, opened 2024-05-23
LLM: unify baichuan2-13b alibi mask dtype with model dtype.

lalalapotter:
## Description

This PR is to unify baichuan2-13b alibi mask dtype with model dtype.


python/llm/src/ipex_llm/transformers/models/baichuan2.py
```diff
@@ -260,14 +260,15 @@ def _buffered_future_mask(tensor, maxpos, alibi, attn_heads):
 
 def baichuan_13b_gen_alibi_mask(tensor, n_head, max_pos):
     # May use fp16 for alibi mask to further reduce memory
-    slopes = torch.Tensor(_get_interleave(n_head))  # .half()
```
hkvision, line 263: remove this and .half()?

hkvision (approved):

PR #11111, opened 2024-05-23
support running internlm-xcomposer2 on cpu

MeouSker77:
## Description

support running internlm-xcomposer2 on cpu, xpu support and more optimization are in progress


rnwang04 (approved):
LGTM

PR #11112, opened 2024-05-23
FIX: APT Repository not working (signatures invalid)

liu-shaojun:
### 1. Why the change?

Fix https://github.com/intel-analytics/ipex-llm/actions/runs/9193745920/job/25285547131

```
W: GPG error: https://repositories.intel.com/gpu/ubuntu jammy InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 28DA432DAAC8BAEA
E: The repository 'https://repositories.intel.com/gpu/ubuntu jammy InRelease' is not signed.
```

Using chmod 644 /usr/share/keyrings/xxx.gpg is to ensure that the public key file has the appropriate permissions set for security purposes.


hzjane (approved):
LGTM

PR #11117, opened 2024-05-23
Fix type mismatch in eval for Baichuan2 QLora example

qiyuangong:
## Description

During evaluation, Baichuan2 will raise type mismatch when training with bfloat16. Fix this issue by modifying modeling_baichuan.py

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [ ] Application test
- [x] Document test
- [ ] ...

### 5. New dependencies

- [ ] New Python dependencies
       - Dependency1 
       - Dependency2
       - ...
       - Dependency1 and license1
       - Dependency2 and license2
       - ...

### DIFF
--- a/python/llm/example/GPU/LLM-Finetuning/QLoRA/alpaca-qlora/README.md
+++ b/python/llm/example/GPU/LLM-Finetuning/QLoRA/alpaca-qlora/README.md
@@ -175,6 +175,23 @@ bash qlora_finetune_qwen15_7b_arc_1_card.sh
 
 ##### Finetuning Baichuan2-7B examples on single Arc A770
 
+Modify `modeling_baichuan.py` in model dir. Add following 2 lines into Line 234. This changes fix [Baichuan2 type mismatch issue](https://github.com/baichuan-inc/Baichuan2/issues/291).
+
+```python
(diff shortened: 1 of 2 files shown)


python/llm/example/GPU/LLM-Finetuning/QLoRA/alpaca-qlora/README.md
```diff
@@ -175,6 +175,23 @@ bash qlora_finetune_qwen15_7b_arc_1_card.sh
 
 ##### Finetuning Baichuan2-7B examples on single Arc A770
 
+Modify `modeling_baichuan.py` in model dir. Add following 2 lines into Line 234. This change fixes [Baichuan2 type mismatch issue](https://github.com/baichuan-inc/Baichuan2/issues/291).
```
plusbang, line 178: This solution requires users to download model first and modify local model file. Maybe we need to add related description? Besides, maybe we also need to modify `--base_model` in `qlora_finetune_baichuan2_7b_arc_1_card.sh` to be local path?

qiyuangong: Done

plusbang (approved):

PR #11128, opened 2024-05-24
Fix qwen 14b run into register attention fwd

leonardozcm:
## Description

The max_position_embeddings of qwen 14b is also 8192, should be excluded.


liu-shaojun (approved):
LGTM

PR #11150, opened 2024-05-28
Fix concurrent issue in autoTP streming.

Romanticoseu:
## Description

Fix concurrent issue in autoTP streming.


python/llm/example/GPU/Deepspeed-AutoTP-FastAPI/README.md
```diff
@@ -159,4 +159,22 @@ Please change the test url accordingly.
...
+```
+
+## Using the `benchmark.py` Script
+
+The `benchmark.py` script is designed to evaluate the performance of a streaming service by measuring response times and other relevant metrics. Below are the details on how to use the script effectively:
+
+### Command Line Arguments
+- `--prompt_length`: Specifies the length of the prompt used in the test. Acceptable values are `32`, `1024`, and `2048`.
+- `--max_concurrent_requests`: Defines the levels of concurrency for the requests. You can specify multiple values to test different levels of concurrency in one run.
+- `MAX_TOKENS`: Set this value directly in the script at line 238 to define the maximum number of new tokens generated per request.
```
xiangyuT, line 171: Should we also set `MAX_TOKENS` by args, such as `--output_length`?

Romanticoseu: Ok, I add `max_new_tokens`

xiangyuT (approved):
LGTM

glorysdj (approved):
LGTM

PR #11153, opened 2024-05-28
fix first token sdp with batch

MeouSker77:
## Description

fix first token sdp with batch: pass `attention_mask` to `sdp_causal`, and if bsz > 1, it will compute attention according to  `attention_mask`.

As for qwen, seems no way to get correct `attention_mask`, so just use `None` instead of `attention_mask`

### DIFF
--- a/python/llm/src/ipex_llm/transformers/models/qwen.py
+++ b/python/llm/src/ipex_llm/transformers/models/qwen.py
@@ -127,9 +127,9 @@ def qwen_attention_forward(
     elif use_sdp_causal(q_len, kv_seq_len, self.head_dim, query_states, self.training):
         import xe_addons
         if use_quantize_kv:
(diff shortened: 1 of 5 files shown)


python/llm/src/ipex_llm/transformers/models/qwen.py
```diff
@@ -127,9 +127,9 @@ def qwen_attention_forward(
     elif use_sdp_causal(q_len, kv_seq_len, self.head_dim, query_states, self.training):
         import xe_addons
         if use_quantize_kv:
-            attn_output = xe_addons.sdp_fp8_causal(query_states, key_states, value_states)
+            attn_output = xe_addons.sdp_fp8_causal(query_states, key_states, value_states, None)
```
rnwang04, line 130: maybe add a comment for why here use None.

rnwang04 (approved):
others LGTM.

PR #11168, opened 2024-05-29
LLM: Refine Pipeline Parallel FastAPI example

xiangyuT:
## Description

Add transformers version for Llama2 models.
Add NUM_GPUS in run.sh


gc-fu (approved):
LGTM

PR #11185, opened 2024-06-03
LLM: fix input length condition in deepspeed all-in-one benchmark.

lalalapotter:
## Description

fix input length condition in deepspeed all-in-one benchmark to prevent infinite loop.

### DIFF
--- a/python/llm/dev/benchmark/all-in-one/run.py
+++ b/python/llm/dev/benchmark/all-in-one/run.py
@@ -1518,7 +1518,7 @@ def run_deepspeed_optimize_model_gpu(repo_id,
             # in_len.txt maybe shorter than we need,
             # use much longer context to make sure input length
             test_length = min(in_len*2, 8192)
-            while test_length not in [32, 256, 1024, 2048, 8192]:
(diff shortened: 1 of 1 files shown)


python/llm/dev/benchmark/all-in-one/run.py
```diff
@@ -1518,7 +1518,7 @@ def get_int_from_env(env_keys, default):
             # in_len.txt maybe shorter than we need,
             # use much longer context to make sure input length
             test_length = min(in_len*2, 8192)
-            while test_length not in [32, 256, 1024, 2048, 8192]:
+            while test_length not in [32, 256, 1024, 2048, 8192] and test_length < 8192:
```
plusbang, line 1521: shall we also add this to other test_api ?

plusbang (approved):

python/llm/dev/benchmark/all-in-one/run.py
lalalapotter: merge this PR first, will add in other PR.

PR #11190, opened 2024-06-03
Fix Starcoder issue on CPU on transformers 4.36+

jenniew:
Fix Starcoder issue on transformers 4.36
Related issue: https://github.com/analytics-zoo/nano/issues/1408

when transformers==4.36+,  starcoder will use GPTBigCodeSdpaAttention instead of GPTBigCodeAttention,
 but our optimize would replace GPTBigCodeAttention.forward with gptbigcode_attention_forward. 
Since GPTBigCodeSdpaAttention is subclass of GPTBigCodeAttention, and our condition expression uses isinstance(sub_m, target_m) which will also return true if  sub_m is the instance of subclass of target_m, then GPTBigCodeSdpaAttention.forward is also replaced by gptbigcode_attention_forward that caused the error.

Change condition expression to if sub_m.__class__ == target_m will not apply replacement for subclass. 

Also Add optimization of GPTBigCodeSdpaAttention.forward for transformers 4.36+ on CPU

Tested on CPU with 4.31, 4.36 and 4.39

### DIFF
--- a/python/llm/src/ipex_llm/transformers/convert.py
+++ b/python/llm/src/ipex_llm/transformers/convert.py
@@ -836,7 +836,7 @@ def convert_bigdl_other_module(model, dtype):
 
 def convert_forward(m, target_m, new_forward):
     for _, sub_m in m.named_children():
-        if isinstance(sub_m, target_m):
+        if sub_m.__class__ == target_m:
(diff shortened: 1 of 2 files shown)


Uxito-Ada (approved):
lgtm

PR #11191, opened 2024-06-03
fix mistral for transformers>=4.39

songhappy:
## Description

fix https://github.com/analytics-zoo/nano/issues/1384 mistral model on transformers>=4.39.

### DIFF
--- a/python/llm/src/ipex_llm/transformers/models/mistral.py
+++ b/python/llm/src/ipex_llm/transformers/models/mistral.py
@@ -1074,3 +1074,232 @@ def mistral_attention_forward_4_36_original(
         attn_weights = None
 
     return attn_output.to(original_dtype), attn_weights, past_key_value
+
+
+def mistral_attention_forward_4_38(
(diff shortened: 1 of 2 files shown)


python/llm/src/ipex_llm/transformers/models/mistral.py
```diff
@@ -1074,3 +1074,232 @@ def mistral_attention_forward_4_36_original(
         attn_weights = None
 
     return attn_output.to(original_dtype), attn_weights, past_key_value
+
+
+def mistral_attention_forward_4_38(
```
yangw1234, line 1079: do you mean `4_39` here? `4_38` is not used anywhere

jason-dai:
> updated and tested on both CPU and Arc, transformers==4.39.2 and 4.41.2

Please also verify on MTL

songhappy:
tested on MTL(Xiaoxin04)

yangw1234 (approved):

PR #11193, opened 2024-06-04
[WIP] Add look up table in 1st token stage

leonardozcm:
## Description

We save the overhead of each unfold to do the draft by maintaining a lookup table of `input ids`


cyita (approved):
LGTM

PR #11197, opened 2024-06-04
Add extra warmup for chatglm3-6b in igpu-performance test

Oscilloscope98:
## Description

The performance of ChatGLM3-6B in nightly performance test is very unstable, especially for 32-32 (int4+fp32). Thus, extra warmup (load + inference once) is added to ChatGLM3-6B to record more stable performance.


liu-shaojun (approved):
LGTM

PR #11203, opened 2024-06-04
Remove software-properties-common Module to Eliminate GPL/LGPL Dependency

liu-shaojun:
### 1. Why the change?

This submission aims to remove the `software-properties-common` module, which introduced the GPL/LGPL license, to meet the requirements of OSPDT. 

`software-properties-common` is a package used for managing software repositories in Ubuntu and related distributions. It provides utilities for adding, removing, and managing apt repositories. The `add-apt-repository` command, which is included in this package, is used to add new apt repositories to your system.  Since the default apt sources already include Python 3.11 related packages, we no longer need it.


glorysdj (approved):
LGTM

PR #11206, opened 2024-06-04
LLM: Fix vLLM CPU version error

xiangyuT:
## Description

Add version check for vllm to import tp related methods only with xpu,
Fix model runner on CPU.


gc-fu (approved):
LGTM

PR #11207, opened 2024-06-04
Improve ipex-llm-init.bat

ATMxsp01:
## improvement to ipex-llm-init.bat
issue:
https://github.com/analytics-zoo/nano/issues/1362#issuecomment-2147033835

---

This pr only include the GPU mode part. type ipex-llm-init.bat without option will enable gpu mode.


python/llm/scripts/ipex-llm-init-support.py
```diff
@@ -0,0 +1,122 @@
+import os
+import sys
```
Oscilloscope98, line 2: This python scripts may also need to be put into here: https://github.com/intel-analytics/ipex-llm/blob/main/python/llm/setup.py#L337 to make sure users can directly use it through ipex-llm-init after installing ipex-llm

PR #11211, opened 2024-06-05
[Fix] always cleanup conda env after linux-build-avx512

liu-shaojun:
### 1. Why the change?

This submission is to fix https://github.com/intel-analytics/ipex-llm/actions/runs/9367630068/job/25816285425


glorysdj (approved):
LGTM

PR #11231, opened 2024-06-05
Add check before access model.device in load_low_bit

jenniew:
In load_low_bit(), we'll check whether model.device.type in ('cpu', 'meta'). Since some models do not have 'device' attribute, there would get error when access model.device.type. Add `hasattr(model, 'device')` before access model.device.

### DIFF
--- a/python/llm/src/ipex_llm/optimize.py
+++ b/python/llm/src/ipex_llm/optimize.py
@@ -170,9 +170,10 @@ def load_low_bit(model, model_path):
         invalidInputError(isinstance(model, torch.nn.Module),
                           "model should be an instance of `torch.nn.Module`, "
                           f"but got {type(model)} at last.")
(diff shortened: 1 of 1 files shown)


leonardozcm (approved):

python/llm/src/ipex_llm/optimize.py
```diff
@@ -170,9 +170,10 @@ def load_low_bit(model, model_path):
         invalidInputError(isinstance(model, torch.nn.Module),
                           "model should be an instance of `torch.nn.Module`, "
                           f"but got {type(model)} at last.")
-        invalidInputError(model.device.type in ('cpu', 'meta'),
```
leonardozcm, line 173: Then how do you ensure that model is on cpu and meta? how about check the status of model.parameters()?

PR #11237, opened 2024-06-06
Fix SDLe CT222 Vulnerabilities

liu-shaojun:
### 1. Why the change?

This submission is to fix CT222 Vulnerabilities:
image
image
image

This submission won't fix 
image
Notes: In the Docker container, we need to use the root user to access the Intel XPU; therefore, this vulnerability cannot be bypassed.

image
Notes: This is more of a best practice than a vulnerability. We set up workdir in our dockerfile, but many commands need to be executed in different directories.
 

### 2. Test
Build docker images: https://github.com/intel-analytics/ipex-llm/actions/runs/9397406169

### DIFF
--- a/docker/llm/inference/cpu/docker/Dockerfile
+++ b/docker/llm/inference/cpu/docker/Dockerfile
@@ -73,4 +73,4 @@ RUN env DEBIAN_FRONTEND=noninteractive apt-get update && \
     pip install transformers==4.36.2
     
 USER myuser
-ENTRYPOINT ["sudo", "/bin/bash"]
+ENTRYPOINT ["/bin/bash"]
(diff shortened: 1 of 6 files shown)


docker/llm/inference/cpu/docker/Dockerfile
```diff
@@ -66,4 +72,5 @@ RUN env DEBIAN_FRONTEND=noninteractive apt-get update && \
     pip install oneccl_bind_pt==2.2.0 --extra-index-url https://pytorch-extension.intel.com/release-whl/stable/cpu/us/ && \
     pip install transformers==4.36.2
     
+USER myuser
```
glorysdj, line 75: change to USER ipex-llm

qiyuangong (approved):
LGTM

PR #11238, opened 2024-06-06
disable quantize kv in specific qwen model

MeouSker77:
## Description

disable quantize kv in specific qwen model


Oscilloscope98 (approved):
LGTM

PR #11244, opened 2024-06-06
Refactor qwen2 moe

MeouSker77:
## Description

Refactor qwen2_moe, qwen2_moe use same Attention as qwen2, so use qwen2's attention directly


plusbang (approved):
LGTM

PR #11254, opened 2024-06-07
LLM: Fix vLLM CPU model convert mismatch

xiangyuT:
## Description

Update Qwen2Attention forward and BaichuanAttention forward.
Add model check before ipex-llm convert.


glorysdj (approved):
LGTM

PR #11257, opened 2024-06-07
Update sample output for HF Qwen2 GPU and CPU

lzivan:
## Description
Fix sample output under README for HF Qwen2 GPU and CPU


Oscilloscope98 (approved):
LGTM

Issue #11246, opened 2024-06-06
Add support for StableLM-2-12B on GPUs


aahouzi:
## Type of Change

Stability.ai open sourced StableLM-2-12B, which has a different architecture than its 1.6B & 3B counterparts. This issue is to ask for adding support for these models: stabilityai/stablelm-2-12b & stabilityai/stablelm-2-12b-chat to IPEX-LLM.

## Description

- Uses GQA instead of MHA + Parallel MLP layer + per-head qk_normalization
- Model description: [StableLM-2-12B](https://huggingface.co/stabilityai/stablelm-2-12b)

PR #11265, opened 2024-06-07
support stablm2 12b

MeouSker77:
## Description

Add qk layernorm in stablelm2 to support stablelm2-12b. #11246 

Its performance is not good because its head_dim is 160, which is not supported by sdp kernel for now, and its qk norm is too slow

### DIFF
--- a/python/llm/src/ipex_llm/transformers/models/stablelm.py
+++ b/python/llm/src/ipex_llm/transformers/models/stablelm.py
@@ -131,7 +131,7 @@ def stablelm_attention_forward(
     query_states, key_states, value_states = qkv.split([self.num_heads,
                                                         self.num_key_value_heads,
                                                         self.num_key_value_heads], dim=1)
-    # For stablelm-2-12b's qk per norm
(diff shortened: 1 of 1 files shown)


Oscilloscope98 (approved):

aahouzi:
@MeouSker77 Few questions for you here:
1. You said head_dim of 160 is not supported in sdp kernel for now: Can u provide more explanation and when will it be supported then ?
2. Any optimizations we can apply to make the per-head qk_normalization faster ?

MeouSker77:
> 1. You said head_dim of 160 is not supported in sdp kernel for now: Can u provide more explanation and when will it be supported then ?

we haven't written sdp kernel for head_dim=160, write new kernel is not easy. 

we'll support it later if no other more important tasks

> 2. Any optimizations we can apply to make the per-head qk_normalization faster ?

group_norm may help, I'll try it

aahouzi:
image

@MeouSker77 Can u add group_norm as u mentioned in the comment above ? I did the same when enabling it on Neural Speed, combining the per-head normalization might a good thing to do

PR #11272, opened 2024-06-11
Remove manual importing ipex in all-in-one benchmark

Oscilloscope98:
## Description

Remove manual importing ipex in all-in-one benchmark

Related commit: https://github.com/intel-analytics/ipex-llm/pull/11227


liu-shaojun (approved):
LGTM

PR #11288, opened 2024-06-12
Expose timeout for streamer for fastchat worker

Oscilloscope98:
## Description

Expose timeout for streamer for fastchat worker; through env variable `FASTCHAT_WORKER_API_TIMEOUT`

Reference: https://github.com/lm-sys/FastChat/blob/74aa8ae36659092542a24bf03df1c35ce94ba494/fastchat/constants.py#L45


gc-fu (approved):
LGTM

PR #11292, opened 2024-06-12
Fix `FP16Linear`

Oscilloscope98:
## Description

`torch.ops.torch_ipex.matmul_bias_out` does not work on MTL iGPU for Linux with short input (with oneAPI 2024.0).

As `torch.ops.torch_ipex.matmul_bias_out` is proved to have benefits for FP16Linear forward on pvc, flex and arc, limit the usage to these devices for now.

### How to test?
- [x] Unit test


rnwang04 (approved):
LGTM

PR #11315, opened 2024-06-14
exclude dolly-v2-12b for arc perf test

liu-shaojun:
### 1. Why the change?

During the nightly Arc performance test, the Arc11 runner was OOM-killed while running Dolly-v2-12b. We will exclude this model for now and investigate the issue later.

Nightly Test: https://github.com/intel-analytics/ipex-llm/actions/runs/9503383526/job/26213157929


lalalapotter (approved):
LGTM

PR #11318, opened 2024-06-14
Fix pipeline parallel inference past_key_value error in Baichuan

sgwhat:
## Description
This PR mainly supports `past_key_values` with `tuple` type in pipeline parallel inference.

### 1. Why the change?
Currently we only support `past_key_values` with `DynamicCache`.

### 2. User API changes

N/A

### 3. Summary of the change 
- Add dummy past_key_value for dummy layer.

### 4. How to test?
- [x] Local test
- [x] Unit test


plusbang (approved):
Please also update verified model list in README.

PR #11328, opened 2024-06-17
Fix igpu performance test regarding html generation

Oscilloscope98:
## Description

Fix igpu performance test regarding html generation: https://github.com/intel-analytics/ipex-llm/actions/runs/9537420911/job/26291985511


liu-shaojun (approved):
LGTM

PR #11334, opened 2024-06-17
Add phi-3 model support for pipeline parallel inference

sgwhat:
## Description
Support phi3 model to run with pipeline parallel inference.

### 1. Why the change?
Previously implementation cannot support to run phi3 model.

### 2. User API changes
N/A

### 3. How to test?
- [x] Local test
- [x] Unit test


plusbang (approved):
LGTM

PR #11338, opened 2024-06-17
LLM: Add RAGFlow with Ollama Example QuickStart

lzivan:
## Description

-Add RAGFlow with Ollama Example QuickStart


shane-huang (approved):
LGTM. 

Merge for now and will do some revisions in following PR.

PR #11344, opened 2024-06-18
add config and default value

hxsz1997:
- Add config `lookahead`, `max_matching_ngram_size`, `task`  in `python\llm\test\benchmark\arc-perf-test.yaml`
- Add default value of `lookahead`, `max_matching_ngram_size`, `task`


python/llm/test/benchmark/arc-perf-test-batch2.yaml
```diff
@@ -35,4 +35,9 @@ exclude:
   - 'bigcode/starcoder-15.5b-4bit:2048'
```
Oscilloscope98, line 35: I don't think it is necessary to change the yaml files for performance tests as all of these 5 arguments have default value for now?

liu-shaojun (approved):
LGTM

hxsz1997:
> need to update readme (add task and remove lookahead)

Thanks for reminder. I will update readme in another PR soon.

PR #11346, opened 2024-06-18
Remove zero3 context manager from LoRA

Uxito-Ada:
### 1. Why the change?

#11167
This is an [unsolved issue](https://github.com/microsoft/DeepSpeed/issues/4703) in Deepspeed community, that zero3 context manager (which is needed by IPEX-LLM NF4 QLoRA) fails to find dist meta data e.g. world_size, and tensor position.

As LoRA BF16 is actually native huggingface transformers, zero3 context manager can be removed, which is only required by IPEX-LLM NF4 model.

### 2. User API changes

no

### 3. Summary of the change 

Remove zero3 context manager from LoRA

### 4. How to test?
- [ ] N/A
- [ ] Unit test
- [x] Application test
- [ ] Document test
- [ ] ...

### DIFF
--- a/python/llm/example/GPU/LLM-Finetuning/LoRA/alpaca_lora_finetuning.py
+++ b/python/llm/example/GPU/LLM-Finetuning/LoRA/alpaca_lora_finetuning.py
@@ -158,51 +158,28 @@ def train(
     # Check if parameter passed or if set within environ
     use_wandb = wandb_check(wandb_project, wandb_watch, wandb_log_model)
 
-    if deepspeed_zero3:
-        deepspeed = deepspeed if deepspeed is not None else "./deepspeed_zero3_config.json"
-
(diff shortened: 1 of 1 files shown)


python/llm/example/GPU/LLM-Finetuning/LoRA/alpaca_lora_finetuning.py
qiyuangong, line 1: Do we need to remove `deepspeed_zero3` in parameter?

Uxito-Ada: Yes, to judge whether it is needed to premove model to XPU.

qiyuangong (approved):
Other LGTM

plusbang (approved):
LGTM

PR #11347, opened 2024-06-18
add initial support for intel npu acceleration library

MeouSker77:
## Description

Add initial support for intel npu acceleration library, just keep ipex-llm's API, calling `intel_npu_acceleration_library`'s `compile` function directly

### 2. User API changes

```python
from ipex_llm.transformers.npu_model import AutoModelForCausalLM
model = AutoModelForCausalLM.from_pretrained(model_path, trust_remote_code=True, load_in_low_bit='fp16')
```

`ipex_llm.transformers.npu_model.AutoModelForCausalLM` receives all arguments of `ipex_llm.transformers.AutoModelForCausalLM`, but it doesn't support most of them, so it just ignores them.

`load_in_low_bit` only supports `sym_int4`, `sym_int8` and `fp32`

**Note: import ipex will cause `intel_npu_acceleration_library` cannot find NPU, so we must disable ipex auto-import when use it by setting `set BIGDL_IMPORT_IPEX=0`**

### DIFF
--- a/python/llm/src/ipex_llm/transformers/npu_model.py
+++ b/python/llm/src/ipex_llm/transformers/npu_model.py
@@ -1,0 +1,141 @@
+#
+# Copyright 2016 The BigDL Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import warnings
+import torch
+import transformers
+from typing import List
+from unittest.mock import patch
+from transformers.dynamic_module_utils import get_imports
+
+import intel_npu_acceleration_library as npu_lib
+from intel_npu_acceleration_library.dtypes import int8, int4
+
+from ipex_llm.utils.common.log4Error import invalidInputError
+
+
+def patch_flash_attn_import(filename: str) -> List[str]:
+    """Work around for https://huggingface.co/microsoft/phi-1_5/discussions/72."""
+    imports = get_imports(filename)
+    if "flash_attn" in imports:
+        imports.remove("flash_attn")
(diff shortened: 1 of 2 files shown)


python/llm/src/ipex_llm/transformers/npu_model.py
```diff
@@ -0,0 +1,141 @@
...
+        """
+        if kwargs.get('device_map', None) not in [None, 'cpu', 'auto']:
+            warnings.warn("`device_map` will be ignored")
+        kwargs['device_map'] = 'cpu'
+
+        low_bit = kwargs.pop('load_in_low_bit', None)
+        low_bit_to_dtype_map = {
+            'sym_int4': int4,
+            'sym_int8': int8,
+            'fp32': torch.float,
```
rnwang04, line 71: how about fp16?

rnwang04 (approved):
others LGTM as an initial PR.

python/llm/src/ipex_llm/transformers/npu_model.py
MeouSker77: it only moves Linear Computation to NPU, other OPs are still run on CPU, CPU doesn't support most fp16 operations

python/llm/src/ipex_llm/transformers/npu_model.py
```diff
@@ -0,0 +1,141 @@
...
+                        *args,
+                        **kwargs):
+        """
+        Load a model from a directory or the HF Hub. Use load_in_low_bit parameter to convert
+        model to low-bit format, like int4 and int8.
+        The loaded model will run supported OPs on NPU, then run other OPs on CPU.
+
+        Three new arguments are added to extend Hugging Face's from_pretrained method as follows:
+        :param load_in_low_bit: str value, options are ``'sym_int4'``, ``'sym_int8'``,
+                                ``'fp16'`` and ``'fp32'``.
```
rnwang04, line 59: So we should not declare `fp16` support here ?

jason-dai:
> **Note: import ipex will cause `intel_npu_acceleration_library` cannot find NPU, so we must disable ipex auto-import when use it by setting `set BIGDL_IMPORT_IPEX=0`**

Maybe add another ENV "BIGDL_USE_NPU" to disable auto import

MeouSker77:
> Maybe add another ENV "BIGDL_USE_NPU" to disable auto import

added it

jason-dai (approved):
LGTM

PR #11348, opened 2024-06-18
fix vLLM/docker issues

gc-fu:
## Description

1. Add more benchmark options in the vLLM benchmark script.
2. Fix segfault in serving image.


glorysdj (approved):
LGTM

PR #11350, opened 2024-06-18
Upgrade scikit-learn to 1.5.0 to fix dependabot issue

liu-shaojun:
### 1. Why the change?

Upgrade scikit-learn to 1.5.0 to fix dependabot issue: https://github.com/intel-analytics/ipex-llm/security/dependabot/2939


qiyuangong (approved):
LGTM

PR #11352, opened 2024-06-18
add fp16 NPU Linear support and fix intel_npu_acceleration_library version 1.0 support

MeouSker77:
## Description

add fp16 NPU Linear support and fix intel_npu_acceleration_library version 1.0 support.

its Linear can use fp16 as long as its whole model's dtype is still fp32


rnwang04 (approved):
LGTM

PR #11359, opened 2024-06-19
Add initial save/load low bit support for NPU

MeouSker77:
## Description

Add initial save/load low bit support for NPU to accelerate model loading


MeouSker77:
For now, only fp16 can be saved and loaded successfully, 

int8 and int4 can be saved and loaded, but after that, they will generate wrong output

rnwang04 (approved):

PR #11361, opened 2024-06-19
fix a bug on run.py

RyuKosei:
edited the function run_ipex_fp16_gpu() so that this api can work correctly.


python/llm/dev/benchmark/all-in-one/run.py
```diff
@@ -698,7 +698,7 @@ def run_ipex_fp16_gpu(repo_id,
                 print(output[0])
                 if i >= warm_up:
                     result[in_out].append([model.first_cost, model.rest_cost_mean, model.encoder_time,
-                                           actual_in_len, actual_out_len, load_time])
+                                           actual_in_len, actual_out_len, load_time,model.peak_memory])
```
hkvision, line 701: space after `,`

hkvision (approved):

PR #11375, opened 2024-06-20
Support PP inference for chatglm3

plusbang:
## Description

Support chatglm3 and update related PP example

### 2. User API changes
N/A

### 3. Summary of the change 
- Add split operations for chatglm model structure
- Add `num_layers` config and other compatible processing
- Add related example script

### 4. How to test?
- [x] Unit test
- [x] Local test
image


sgwhat (commented):
Maybe we can use `try/except` instead of creating a new `glm_generate.py`.

plusbang:
> Maybe we can use `try/except` instead of creating a new `glm_generate.py`.

Have updated.

sgwhat (approved):
LGTM

PR #11403, opened 2024-06-24
optimize qwen1.5/2 memory usage when running long input with fp16

MeouSker77:
## Description

1. remove converting `logits` to fp32 to optimize qwen1.5/2 memory usage when running long input with fp16
2. current `convert_forward` cannot replace `forward` method of the outermost model, so fix it


rnwang04 (approved):
LGTM

PR #11432, opened 2024-06-26
FIX: Qwen1.5-GPTQ-Int4 inference error

liu-shaojun:
### 1. Why the change?

Fix user issue: https://github.com/intel-analytics/ipex-llm/issues/11413

### 2. User API changes

No user api change. 

### 3. Summary of the change 

The weights of gptq have already been quantized. During the merge_qkv process in qwen2, we encounter an `AttributeError: 'QuantLinear' object has no attribute 'weight'. Did you mean: 'qweight'?` This PR skips merge_qkv for the qwen2 gptq model as a quick fix. We will submit another PR later to thoroughly resolve this issue.

### 4. How to test?
manually tested it

### 5. New dependencies

No

### DIFF
--- a/python/llm/src/ipex_llm/transformers/convert.py
+++ b/python/llm/src/ipex_llm/transformers/convert.py
@@ -733,17 +733,17 @@ def _optimize_pre(model):
         model.apply(split_mlp)
     # for qwen2
     if model.config.model_type == "qwen2":
-        from ipex_llm.transformers.models.qwen2 import merge_qkv
-        # Skip merge_qkv if quant_method is 'gptq'
+        # Skip merge_qkv and padding_mlp if quant_method is 'gptq'
         should_apply_merge_qkv = (
             not hasattr(model.config, "quantization_config") or
             not hasattr(model.config.quantization_config, "quant_method") or
             model.config.quantization_config.quant_method != "gptq"
         )
         if should_apply_merge_qkv:
+            from ipex_llm.transformers.models.qwen2 import merge_qkv
             model.apply(merge_qkv)
-        from ipex_llm.transformers.models.qwen2 import padding_mlp
-        model.apply(padding_mlp)
+            from ipex_llm.transformers.models.qwen2 import padding_mlp
+            model.apply(padding_mlp)
     if model.config.model_type == "qwen2_moe":
         from ipex_llm.transformers.models.qwen2_moe import merge_qkv
         model.apply(merge_qkv)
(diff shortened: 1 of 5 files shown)


python/llm/src/ipex_llm/transformers/convert.py
```diff
@@ -734,7 +734,14 @@ def _optimize_pre(model):
...
         from ipex_llm.transformers.models.qwen2 import merge_qkv
-        model.apply(merge_qkv)
+        # Skip merge_qkv if quant_method is 'gptq'
+        should_apply_merge_qkv = (
+            not hasattr(model.config, "quantization_config") or
+            not hasattr(model.config.quantization_config, "quant_method") or
+            model.config.quantization_config.quant_method != "gptq"
+        )
+        if should_apply_merge_qkv:
+            model.apply(merge_qkv)
```
jason-dai, line 744: Is gptq loaded as q4_1? If so, why can't we use `merge_qkv`? @qiuxin2012

qiuxin2012: We need a new merge_qkv for gptq here. Just convert q k v to one LowBitLinear.

jason-dai: > We need a new merge_qkv for gptq here. Just convert q k v to one LowBitLinear.
It's different from q4_1?

liu-shaojun: gptq is loaded as q4_1. However, gptq's weights have already been quantized. If we want to use merge_qkv, we need to dequantize from the gptq format to the normal format first, perform merge_qkv, and then quantize back into the LowBitLinear. 

I will try this method.

jason-dai: > gptq is loaded as q4_1. However, gptq's weights have already been quantized. If we want to use merge_qkv, we need to dequantize from the gptq format to the normal format first, perform merge_qkv, and then quantize back into the LowBitLinear.
> 
> I will try this method.

Instead of dequantization, I think we can just rearrange the quantized qkv tensors into a combined one? Anyway I think we may fix it later in a separate PR if needed.

liu-shaojun: I'll merge this PR first so users can start using it, and then I'll submit another PR to address further fixes.

yangw1234 (approved):

python/llm/src/ipex_llm/transformers/convert.py
yangw1234: I guess we probably cannot merge qkv if `quantization_config.desc_act==True`.

PR #11434, opened 2024-06-26
Fix error while using pipeline parallism

gc-fu:
## Description

Fix vllm enabling error while doing pipeline parallism


xiangyuT (approved):
LGTM

PR #11435, opened 2024-06-26
optimize qwen2 gpu memory usage again

MeouSker77:
## Description

pin rope cos and sin table on CPU to save gpu memory.

gpu can use fuse rope kernel to compute cos and sin, so gpu doesn't need them.


rnwang04 (approved):
LGTM

PR #11463, opened 2024-06-28
Support pipeline parallel for glm-4-9b-chat

plusbang:
## Description

Support pipeline parallel inference & serving for glm-4-9b-chat

### 2. User API changes

N/A

### 4. How to test?
- [x] Unit test (https://github.com/intel-analytics/ipex-llm-workflow/actions/runs/9772330989)
- [x] Local test (both serving and inference of glm4 is verified)

### DIFF
--- a/python/llm/src/ipex_llm/transformers/models/chatglm4.py
+++ b/python/llm/src/ipex_llm/transformers/models/chatglm4.py
@@ -133,7 +133,7 @@ def chatglm4_attention_forward(
     # hidden_states: [b, sq, h]
     # torch.set_printoptions(edgeitems=hidden_states.numel())
     # print(f'On my rank {torch.distributed.get_rank()}, INPUT IS {hidden_states}, AND {attention_mask}, AND {rotary_pos_emb}, AND {kv_cache}')
-    print(f'ON ATTENTION FORWARD, INPUT IS {hidden_states}, AND {attention_mask}, AND {rotary_pos_emb}, AND {kv_cache}')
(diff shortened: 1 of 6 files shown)


python/llm/src/ipex_llm/transformers/models/chatglm4.py
```diff
@@ -46,10 +46,11 @@ def chatglm4_model_forward(
     use_cache = use_cache if use_cache is not None else self.config.use_cache
     return_dict = return_dict if return_dict is not None else self.config.use_return_dict
 
-    batch_size, seq_length = input_ids.shape
-
     if inputs_embeds is None:
+        batch_size, seq_length = input_ids.shape
         inputs_embeds = self.embedding(input_ids)
+    else:
+        batch_size, seq_length, _ = inputs_embeds.shape
```
xiangyuT, line 53: If attention_mask is given a not-None and not-All value, `input_ids` is needed in line 58 `self.get_masks()` and it will raise error if `input_ids` is still None. Maybe add an empty tensor here? 
```python
input_ids = torch.empty((batch_size, seq_length), device=inputs_embeds.device)
```

xiangyuT: Relative code in modeling_chatglm:
```python
    def get_masks(self, input_ids, past_key_values, padding_mask=None):
        batch_size, seq_length = input_ids.shape
```

plusbang: > If attention_mask is given a not-None and not-All value, `input_ids` is needed in line 58 `self.get_masks()` and it will raise error if `input_ids` is still None. Maybe add an empty tensor here?

Have updated in `chatglm2.py` and `chatglm4.py` :)

xiangyuT (approved):
LGTM

PR #11475, opened 2024-07-01
Enable LLM UT workflow to be called in another repo

Oscilloscope98:
## Description

Enable LLM UT workflow to be called in another repo


liu-shaojun (approved):
LGTM

PR #11479, opened 2024-07-01
Pin compute runtime version for xpu images

gc-fu:
## Description

Pin compute runtime version for xpu images.

For 6.5.0-35-generic kernel.


hzjane (approved):
LGTM

PR #11499, opened 2024-07-03
Wang jun/ipex llm workflow

ACupofAir:
## Description
change `github.event.input` to `input`

### 1. Why the change?
* `github.event.input` will change any type to string which will lead to template not valid


liu-shaojun (approved):
LGTM

PR #11516, opened 2024-07-05
Fix setting of `use_quantize_kv_cache` on different GPU in pipeline parallel

plusbang:
## Description

When run llama2-13b & bs=2 using PP, exist following error:
```bash
attn_output = xe_addons.sdp_fp8(query_states, key_states, value_states, new_attn_mask)

                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

RuntimeError: expected scalar type Byte but found Half
```
The root cause is dummylayer don't have `qtype` attribute, then different value of `use_quantize_kv_cache` is obtained on `xpu0` and `xpu1` if env variable is not set.

Fix it in this PR. 

### 2. User API changes

N/A

### 4. How to test?
https://github.com/intel-analytics/ipex-llm-workflow/actions/runs/9805773757
- [x] Local test

### DIFF
--- a/python/llm/src/ipex_llm/transformers/pipeline_parallel.py
+++ b/python/llm/src/ipex_llm/transformers/pipeline_parallel.py
@@ -107,7 +107,7 @@ def init_pipeline_parallel():
     dist.init_process_group('ccl')
 
 
-def _check_quantize_kv_cache(model, idx, batch_size) -> bool:
+def _check_quantize_kv_cache(model, idx, batch_size):
     # align use_quantize_kv_cache setting for different GPU in pipeline parallel
     pp_quantize_kv_cache = (os.environ.get("BIGDL_QUANTIZE_KV_CACHE", None) == "1") or \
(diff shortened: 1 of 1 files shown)


python/llm/src/ipex_llm/transformers/pipeline_parallel.py
```diff
@@ -106,6 +107,29 @@ def init_pipeline_parallel():
     dist.init_process_group('ccl')
 
 
+def _check_quantize_kv_cache(model, idx, batch_size) -> bool:
```
xiangyuT, line 110: Does this function return a bool value?

plusbang: > Does this function return a bool value?

Fix it. No need to return bool value here :)

xiangyuT (approved):
LGTM

PR #11520, opened 2024-07-05
optimize qwen2 memory usage again

MeouSker77:
## Description

`attention_mask` is not required when using sdp_causal and batch_size=1 or generating 2nd+ token, so don't generate it in these cases to save memory


rnwang04 (approved):

MeouSker77:
PR validation: <https://github.com/intel-analytics/ipex-llm-workflow/actions/runs/9805767952/job/27076276807>

PR #11567, opened 2024-07-12
LLM: Add XPU Memory Optimizations for Pipeline Parallel

xiangyuT:
### 3. Summary of the change 

Add model forward convert for llama, chatglm3/4 when `IPEX_LLM_LOW_MEM` is set.
### 4. How to test?
- [x] Local test
https://github.com/intel-analytics/ipex-llm-workflow/actions/runs/9900998189

### DIFF
--- a/python/llm/src/ipex_llm/transformers/pipeline_parallel.py
+++ b/python/llm/src/ipex_llm/transformers/pipeline_parallel.py
@@ -119,22 +119,23 @@ def low_mem_convert(model):
             model,
             transformers.models.llama.modeling_llama.LlamaForCausalLM,
             llama_causallm_forward_4_37_lowmem)
-    elif model.config.model_type == "chatglm" and model.config.num_layers == 40:
-        # for glm4-9b
-        modeling_module_name = model.__class__.__module__
-        module = importlib.import_module(modeling_module_name)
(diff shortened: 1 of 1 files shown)


hkvision (approved):

python/llm/src/ipex_llm/transformers/pipeline_parallel.py
```diff
@@ -186,6 +218,11 @@ def pipeline_parallel(model, pipeline_parallel_stages):
             model._modules['model'].norm = DummyLayer()
             model._modules['lm_head'] = DummyLayer()
 
+    _enable_lowmem = os.getenv('IPEX_LLM_LOW_MEM')
+    _enable_lowmem = (_enable_lowmem is not None) and (_enable_lowmem.lower() == "1")
+    if _enable_lowmem:
+        model = low_mem_convert(model)
```
plusbang, line 224: Not sure this memory optimization could also be added in single GPU inference case? Other LGTM.

hkvision: It could, but for single GPU current existing optimizations may be enough for the demands, so we don't override the CausalLM forward previously.

PR #11571, opened 2024-07-12
Add npu benchmark all-in-one script

leonardozcm:
## Description

Add npu benchmark script

### DIFF
--- a/python/llm/dev/benchmark/all-in-one/run.py
+++ b/python/llm/dev/benchmark/all-in-one/run.py
@@ -636,6 +636,8 @@ def run_transformer_win_npu_int4(repo_id,
                 if i >= warm_up:
                     result[in_out].append([model.first_cost, model.rest_cost_mean, model.encoder_time,
                                            actual_in_len, actual_out_len, load_time])
+    del model
+    gc.collect()
     return result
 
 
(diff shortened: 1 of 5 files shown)


python/llm/dev/benchmark/all-in-one/run.py
```diff
@@ -161,6 +161,8 @@ def run_model(repo_id, test_api, in_out_pairs, local_model_hub=None, warm_up=1,
         result = run_speculative_gpu(repo_id, local_model_hub, in_out_pairs, warm_up, num_trials, num_beams, batch_size)
     elif test_api == 'pipeline_parallel_gpu':
         result = run_pipeline_parallel_gpu(repo_id, local_model_hub, in_out_pairs, warm_up, num_trials, num_beams, low_bit, batch_size, cpu_embedding, fp16=use_fp16_torch_dtype)
+    elif test_api == 'transformer_win_npu_int4':
```
Oscilloscope98, line 164: Should we call it `'transformers_int4_npu_win'`? To make the `test_api`'s name be consistent with others

Oscilloscope98 (commented):
We may also update here: https://github.com/intel-analytics/ipex-llm/blob/main/python/llm/dev/benchmark/all-in-one/config.yaml#L14 and https://github.com/intel-analytics/ipex-llm/blob/main/python/llm/dev/benchmark/all-in-one/README.md#config here? And may also add in comments that some further dependency is needed for this test API? :)

Oscilloscope98 (approved):

PR #11616, opened 2024-07-18
Add check in iGPU perf workflow for results integrity

Oscilloscope98:
## Description

Add check in iGPU perf workflow for results integrity


Oscilloscope98:
https://github.com/intel-analytics/ipex-llm-workflow/actions/runs/9984927225

PR #11647, opened 2024-07-24
Update oneccl used

gc-fu:
## Description

Update oneccl used.


docker/llm/serving/xpu/docker/start-lightweight_serving-service.sh
```diff
@@ -1,4 +1,5 @@
 cd /llm/lightweight_serving
 model_path="/llm/models/Llama-2-7b-chat-hf"
 low_bit="sym_int4"
+source /opt/intel/1ccl-wks/setvars.sh
```
glorysdj, line 4: do we need 1ccl in lightweight serving?

glorysdj (approved):
LGTM

PR #11654, opened 2024-07-25
LLM: Fix ParallelLMHead convert in vLLM cpu

xiangyuT:
### 3. Summary of the change 

Current `ParallelLMHead` convert is only set on XPU and will crush on CPU. Remove `ParallelLMHead` from VLLM_LINEAR_LIST for CPU.

### 4. How to test?
- [ ] Local test


gc-fu (approved):
LGTM

PR #11665, opened 2024-07-26
empty cache only for 1st token but rest token to speed up

Uxito-Ada:
## Description

Work with @qiyuangong 

No need to empty cache for rest token, which also introduces much latency overhead in llama's low-memory forward

### 1. Why the change?

as above

### 2. User API changes

no

### 3. Summary of the change 

empty cache only for 1st token but rest token to speed up

### 4. How to test?
- [ ] N/A
- [x] Application test
- [ ] Document test
- [ ] ...


qiyuangong (approved):
LGTM

PR #11682, opened 2024-07-30
Fix import vllm condition

gc-fu:
## Description

In python script, the first '' will be converted to current directory.  The possible option is to remove the first one.

The path added by modifying PYTHONPATH will be second entry.


xiangyuT (approved):
LGTM

PR #11713, opened 2024-08-05
add new benchmark_util.py

lzivan:
## Description
- Add new benchmark_util.py for transformers>=4.43.1. The old one renamed to benchmark_util_prev.py.
- The version control is done in utils.py: `from ipex_llm.utils.utils import BenchmarkWrapper`


python/llm/src/ipex_llm/utils/utils.py
```diff
@@ -17,7 +17,14 @@
 import sys
 import pathlib
 from ipex_llm.utils.common import invalidInputError, invalidOperationError
+import transformers
 
+trans_version = transformers.__version__
```
qiuxin2012, line 22: You can move this to `__init__.py`, and import lib by `from ipex_llm.utils import BenchmarkWrapper`

qiuxin2012:
rename benchmark_util_prev.py to benchmark_util_4_29.py, and benchmark_util.py to benchmark_util_4_43.py

python/llm/src/ipex_llm/utils/utils.py
```diff
@@ -17,7 +17,14 @@
...
+
+if trans_version >= "4.43.1":
```
qiuxin2012, line 24: 4.43.0?

lzivan: 4.43.0 won't work for llama3.1

python/llm/src/ipex_llm/utils/benchmark_util_4_43.py
```diff
@@ -0,0 +1,4570 @@
+# coding=utf-8
```
qiuxin2012, line 1: Use BIGDL license here, and add a reference to huggingface's utils.py, like 4_29.

PR #11717, opened 2024-08-06
Update Eagle example to Eagle2+ipex-llm integration

jenniew:
Update Eagle example to Eagle2+ipex-llm integration.
With EAGLE-2+IPEX-LLM, the inference speed on Arc with fp16 can increase.


rnwang04 (approved):
LGTM

PR #11765, opened 2024-08-12
Set mistral fuse rope to false except fp6 & fp16

ATMxsp01:
This pr set the fuse rope of `mistral` model to `False`

PR-11747 fixed the `mistral-7B-instruct-v0.2` `FP16` results difference between ipex-llm/ipex/NV in Longbench. However, this caused some other precisions like `sym_int4` output garbled characters. This pr disabled the fuse rope optimization as a workaround to solve this problem (temporarily) .

> [!NOTE]
> Since `fuse qkv` is also disabled in this case, it will cause a decrease to the model performance


cyita (approved):
LGTM

cyita:
PR validation: https://github.com/intel-analytics/ipex-llm-workflow/actions/runs/10348113758

PR #11772, opened 2024-08-13
Fix stdout in all-in-one benchmark to utf-8

Oscilloscope98:
## Description

Fix stdout in all-in-one benchmark to utf-8. This should fix bug in workflow https://github.com/intel-analytics/ipex-llm-workflow/actions/runs/10355795172/job/28682779918


liu-shaojun (approved):
LGTM

PR #11785, opened 2024-08-13
follow up on experimental support of fused decoder layer for llama2

yangw1234:
## Description

1. running two 8-layer decoders instead of one 16-layer in one process
2. support value cache in transposed layout

Test: https://github.com/intel-analytics/ipex-llm-workflow/actions/runs/10378148146


plusbang (approved):
LGTM

PR #11817, opened 2024-08-15
Fix vLLM not convert issues

gc-fu:
## Description

This is used to fix the issue when the layer is not converted in some cases where the in_features % 64 != 0.

This usually happens when TP is used.


gc-fu:
https://github.com/intel-analytics/ipex-llm-workflow/actions/runs/10401547413

liu-shaojun (approved):

PR #11845, opened 2024-08-19
delete transformers version requirement

JinheTang:
It is tested that transformers==4.38 doesn't work on the latest internlm2-chat-7b model. So the specification of 4.38 is deleted.


rnwang04 (approved):
LGTM

PR #11860, opened 2024-08-20
Pytorch models transformers version update

JinheTang:
9 models:
- yi
- yuan2
- internlm2
- mistral
- codellama
- codegeex
- solar
- decilm-7b
- replit

yi's huggingface example default prompt in readme(chinese) is inconsistent with the default prompt in code(english). Synced it to chinese prompt by default


rnwang04:
Please also check whether these model can run with fp16+int4.
If so, let's change them to fp16+int4 under PyTorch and HF.

rnwang04 (approved):
LGTM, merge it first.

PR #11869, opened 2024-08-20
Update performance test regarding updated default `transformers==4.37.0`

Oscilloscope98:
## Description

Update performance test regarding updated default `transformers==4.37.0`


rnwang04 (approved):
LGTM

PR #11894, opened 2024-08-22
change 5 pytorch/huggingface models to fp16

JinheTang:
updated 5 models:
 codegeex2-6b / CodeLlama-7b-hf / Mistral-7B-v0.1 / internlm2-chat-7b-old / SOLAR-10.7B-Instruct-v1.0
to fp16+int4

among those, mistral in huggingface and codegeex2 in both huggingface and pytorch models are already in fp16


rnwang04 (approved):
LGTM

PR #11896, opened 2024-08-22
Fix non-stop at eos token problem for lookup generation

Oscilloscope98:
## Description

Fix non-stop at eos token problem for lookup generation: https://github.com/analytics-zoo/nano/issues/1589#issue-2477776061
(question from gm8ks)
- Before this PR:
  image
- After this PR:
  image

### DIFF
--- a/python/llm/src/ipex_llm/transformers/lookup.py
+++ b/python/llm/src/ipex_llm/transformers/lookup.py
@@ -382,9 +382,17 @@ def lookup_generate(self,
 
         # Stop on eos and remove content after eos
         output_ids_list = output_ids[0].tolist()
-        if generation_config.eos_token_id in output_ids_list:
-            idx = output_ids_list.index(generation_config.eos_token_id)
-            step -= (len(output_ids_list) - idx - 1)
+        if isinstance(generation_config.eos_token_id, list):
+            eos_token_id_set = set(generation_config.eos_token_id)
+        else:
+            eos_token_id_set = [generation_config.eos_token_id]
+        first_eos_idx = -1
+        for out_idx, out_id in enumerate(output_ids_list):
+            if out_id in eos_token_id_set:
+                first_eos_idx = out_idx
+                break
+        if first_eos_idx > -1:
+            step -= (len(output_ids_list) - first_eos_idx - 1)
             break
 
     step = min(step, max_new_tokens)
--- a/python/llm/src/ipex_llm/transformers/lookup.py
+++ b/python/llm/src/ipex_llm/transformers/lookup.py
@@ -381,19 +381,20 @@ def lookup_generate(self,
             self.post_time.append(pot-mot)
 
         # Stop on eos and remove content after eos
-        output_ids_list = output_ids[0].tolist()
-        if isinstance(generation_config.eos_token_id, list):
-            eos_token_id_set = set(generation_config.eos_token_id)
-        else:
-            eos_token_id_set = set([generation_config.eos_token_id])
-        first_eos_idx = -1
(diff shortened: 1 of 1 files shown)


Oscilloscope98:
PR Validation: https://github.com/intel-analytics/ipex-llm-workflow/actions/runs/10505344033

python/llm/src/ipex_llm/transformers/lookup.py
```diff
@@ -381,11 +381,20 @@ def lookup_generate(self,
...
-        if generation_config.eos_token_id in output_ids_list:
-            idx = output_ids_list.index(generation_config.eos_token_id)
-            step -= (len(output_ids_list) - idx - 1)
-            break
+        if generation_config.eos_token_id is not None:
+            output_ids_list = output_ids[0].tolist()
+            if isinstance(generation_config.eos_token_id, list):
+                eos_token_id_set = set(generation_config.eos_token_id)
+            else:
+                eos_token_id_set = set([generation_config.eos_token_id])
```
jason-dai, line 389: This can be initialized just once

jason-dai:
What's the performance impact?

python/llm/src/ipex_llm/transformers/lookup.py
Oscilloscope98: Thank you for the comments. Have move the initialization before the loop.

cyita (approved):
LGTM

Oscilloscope98:
> What's the performance impact?

It should depend on the generated texts after the eos token. If it would be duplication of the eos token (such as the previous Qwen2 case), the performance may drop as we stop earlier with less duplication.

For other cases like self-questioning-and-answering, we will refresh the tested results again on GSM8k dataset, etc. based on today's nightly released version with this commit and update the performance impact in this issue: https://github.com/analytics-zoo/nano/issues/1589

Please let us know for any comments :) @jason-dai

python/llm/src/ipex_llm/transformers/lookup.py
```diff
@@ -381,11 +388,16 @@ def lookup_generate(self,
...
+        if eos_token_id_set is not None:
+            output_ids_list = output_ids[0].tolist()
+            first_eos_idx = -1
+            for out_idx, out_id in enumerate(output_ids_list):
+                if out_id in eos_token_id_set:
+                    first_eos_idx = out_idx
+                    break
+            if first_eos_idx > -1:
+                step -= (len(output_ids_list) - first_eos_idx - 1)
+                break
```
jason-dai, line 400: Is it better to do something like
```python
       for eos_token_id in eos_token_id_set
          if eos_token_id  in output_ids_list:
             idx = output_ids_list.index(eos_token_id)
             step -= (len(output_ids_list) - idx - 1)
             break
```

Oscilloscope98: We do this as we need to find the first token in `output_ids_list` which is in `eos_token_id` :)

jason-dai: OK; then maybe
```python
      for out_idx, out_id in enumerate(output_ids_list):
           if out_id in eos_token_id_set:
                step -= (len(output_ids_list) - out_idx - 1)
                break
```

Oscilloscope98: Thank you for the comment :) But we need a flag here (i.e., `first_eos_idx `) to jump out of the token generation loop. By this way, we could only break the `output_ids_list` iteration loop, which may make eos token ineffective.

jason-dai: ok

PR #11912, opened 2024-08-23
Support qwen2-7b with fused decoderlayer optimization on NPU

plusbang:
## Description

Add qwen2-7B NPU support with fused decoderlayer and multi process optimization.
Details: https://github.com/analytics-zoo/nano/issues/1576#issuecomment-2314969138

Different from other model support, we use QuantizedLinear instead of FusedQwenLowBitDecoderlayer during prefill process.

### 2. User API changes

N/A

### 4. How to test?
- [x] Application test

### DIFF
--- a/python/llm/src/ipex_llm/transformers/npu_models/convert_mp.py
+++ b/python/llm/src/ipex_llm/transformers/npu_models/convert_mp.py
@@ -13,8 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
 import torch
 import importlib
+from ipex_llm.transformers.low_bit_linear import LowBitLinear, FP4Params
 
 
 def convert_forward(m, target_m, new_forward):
@@ -25,7 +27,7 @@ def convert_forward(m, target_m, new_forward):
         convert_forward(sub_m, target_m, new_forward)
 
 
-def optimize_llm_pre(model: torch.nn.Module):
+def optimize_llm_pre(model: torch.nn.Module, qtype):
     if model.config.model_type == "baichuan":
         # process NormHead module in Baichuan2 7B
         if hasattr(model, 'lm_head') and model.lm_head is not None:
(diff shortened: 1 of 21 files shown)


python/llm/src/ipex_llm/transformers/npu_models/convert_mp.py
```diff
@@ -57,8 +57,8 @@ def optimize_llm(
         from transformers.models.llama.modeling_llama import LlamaForCausalLM
         from ipex_llm.transformers.npu_models.llama_mp import llama2_casullm_forward
         convert_forward(model, LlamaForCausalLM, llama2_casullm_forward)
-    elif model.config.model_type == "qwen2" and model.config.intermediate_size == 8960:
-        # for qwen2-1.5B
+    elif model.config.model_type == "qwen2":
+        # for qwen2-1.5B and qwen2-7B
```
jason-dai, line 61: How about Qwen2-0.5B or 72B?

plusbang: > How about Qwen2-0.5B or 72B? Need to add check

Sure, have added related check : )

python/llm/src/ipex_llm/transformers/npu_models/mp_models_base.py
```diff
@@ -229,9 +229,15 @@ def mlp(self, hidden_states):
             hidden_states, self.intermediate_size, self.hidden_size, bias=False, wt_dtype=self.dtype
         )  # type: ignore[attr-defined]
         mm1 = self.eltwise_mul(self.swish(mm1), mm2)  # type: ignore[attr-defined]
-        hidden_states = self.linear(
-            mm1, self.hidden_size, self.intermediate_size, bias=False, wt_dtype=self.dtype
-        )
+        if self.intermediate_size == 18944:
```
yangw1234, line 232: maybe you can just override `self.mlp` in the sub-class

python/llm/src/ipex_llm/transformers/npu_models/mp_models_base.py
```diff
@@ -383,7 +383,7 @@ def update_cache(self, past_key_value, indexes):
             self.load_cache_async()
 
     def load_cache_async(self):
-        self.load_wt_fn(len(self.input_ops), self._mm, self.kv_cache_c_handle)
+        self.load_wt_fn(len(self.input_ops), self._mm, self.kv_cache_c_handle, verify_size=True)
```
yangw1234, line 386: Not sure how much runtime overhead of `verify_size`, if it is large, maybe we can disable it at runtime and only use it when debugging?

yangw1234 (approved):

python/llm/src/ipex_llm/transformers/npu_models/mp_models_base.py
plusbang: > Not sure how much runtime overhead of `verify_size`, if it is large, maybe we can disable it at runtime and only use it when debugging?

According to my experiment, the overhead seems quite small. BTW have removed it : )

PR #11921, opened 2024-08-26
add gptq option for ppl test

cranechu0131:
add gptq option for ppl test


python/llm/dev/benchmark/perplexity/run_wikitext.py
```diff
@@ -106,6 +118,5 @@ def parse_kwargs(kwstr):
     prev_end_loc = end_loc
     if end_loc == seq_len:
         break
-
```
hkvision, line 109: revert this

hkvision (approved):

PR #11932, opened 2024-08-27
Update benchmark script for NPU

plusbang:
## Description

Update all-in-one benchmark script, we could also use all-in-one benchmark npu models with fused decoderlayer optimization.

usage:
set `optimize_model` as True in config.yaml

### 4. How to test?
- [x] Application test

### DIFF
--- a/python/llm/dev/benchmark/all-in-one/run.py
+++ b/python/llm/dev/benchmark/all-in-one/run.py
@@ -136,7 +136,7 @@ def preprocess_prompt(tokenizer, in_len, task):
         input_ids = tokenizer.encode(input_str, return_tensors="pt")    
     return input_ids
 
-def run_model(repo_id, test_api, in_out_pairs, local_model_hub=None, warm_up=1, num_trials=3, num_beams=1, low_bit='sym_int4', cpu_embedding=False, batch_size=1, streaming=False, use_fp16_torch_dtype=False, lookahead=False, task='continuation'):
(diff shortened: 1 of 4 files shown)


sgwhat (approved):

python/llm/dev/benchmark/all-in-one/run.py
```diff
@@ -603,24 +603,30 @@ def transformers_int4_npu_win(repo_id,
...
     # which convert the relevant layers in the model into INT4 format
     st = time.perf_counter()
     if repo_id in CHATGLM_IDS:
         model = AutoModel.from_pretrained(model_path, load_in_low_bit=low_bit, trust_remote_code=True,
+                                          optimize_model=optimize_model, max_output_len=max_output_len, max_prompt_len=int(in_out_len[0]), transpose_value_cache=True,
                                           torch_dtype='auto', attn_implementation="eager").eval()
         tokenizer = AutoTokenizer.from_pretrained(model_path, trust_remote_code=True)
     elif repo_id in LLAMA_IDS:
         model = AutoModelForCausalLM.from_pretrained(model_path, load_in_low_bit=low_bit, trust_remote_code=True,
+                                                     optimize_model=optimize_model, max_output_len=max_output_len, max_prompt_len=int(in_out_len[0]), transpose_value_cache=True,
```
cyita, line 624: Should we add `torch_dtype=torch.float16` as suggested in our example?

plusbang: > Should we add `torch_dtype=torch.float16` as suggested in our example?

Have updated.

PR #11948, opened 2024-08-28
added accelerate upgrade in ollama UI quickstart

JinheTang:
accelerate version specified in `bigdl-core-cpp` is fallen behind and will result in incompatibility with transformers when running `open-webui` for ollama, therefore requires an upgrade before running `start_windows.bat`

### DIFF
--- a/docs/mddocs/Quickstart/open_webui_with_ollama_quickstart.md
+++ b/docs/mddocs/Quickstart/open_webui_with_ollama_quickstart.md
@@ -75,6 +75,7 @@ You may run below commands to install Open WebUI dependencies:
   # Install Dependencies
   cd ./backend
   pip install -r requirements.txt -U
+  pip install --pre --upgrade accelerate
   ```
 
 - For **Windows users**:
@@ -90,6 +91,7 @@ You may run below commands to install Open WebUI dependencies:
   :: Install Dependencies
   cd .\backend
   pip install -r requirements.txt -U
+  pip install --pre --upgrade accelerate
   ```
 
 ### 3. Start the Open-WebUI 


JinheTang:
@rnwang04 pls kindly review :)

rnwang04 (approved):
LGTM

jason-dai:
1) Is it specific to Open WebUI, or shall we upgrade [cpp] dependency?
2) Instead of using the latest version, we should probably specify the particular versions that we have tested
3) We should add these versions into `requirements.txt`?

rnwang04:
> 1. Is it specific to Open WebUI, or shall we upgrade [cpp] dependency?
> 2. Instead of using the latest version, we should probably specify the particular versions that we have tested
> 3. We should add these versions into `requirements.txt`?

OK, we will do more test with transformers / accelerate version, and evaluate if we can directly upgrade and specify their version in [cpp] dependency.

jason-dai:
> > 1. Is it specific to Open WebUI, or shall we upgrade [cpp] dependency?
> > 2. Instead of using the latest version, we should probably specify the particular versions that we have tested
> > 3. We should add these versions into `requirements.txt`?
> 
> OK, we will do more test with transformers / accelerate version, and evaluate if we can directly upgrade and specify their version in [cpp] dependency.

You should always specify the particular versions that you have tested; using latest versions will break the compatibility in future.

rnwang04:
> > > 1. Is it specific to Open WebUI, or shall we upgrade [cpp] dependency?
> > > 2. Instead of using the latest version, we should probably specify the particular versions that we have tested
> > > 3. We should add these versions into `requirements.txt`?
> > 
> > 
> > OK, we will do more test with transformers / accelerate version, and evaluate if we can directly upgrade and specify their version in [cpp] dependency.
> 
> You should always specify the particular versions that you have tested; using latest versions will break the compatibility in future.

Got it, we will pay attention to it later.

PR #11953, opened 2024-08-28
use sdp_causal to reduce internvl2-4b memory usage if set environment variable

MeouSker77:
## Description

use sdp_causal to reduce internvl2-4b memory usage if set environment variable


rnwang04 (approved):
LGTM

PR #11980, opened 2024-08-30
[WIP] fuse lm head on npu

plusbang:
## Description

**Still working on it**

Add model.norm and lm_head in graph but blocking at how to process the lm_head forward result

### 4. How to test?
- [ ] N/A
- [ ] Application test
- [ ] Document test
- [ ] ...

### 5. New dependencies

- [ ] New Python dependencies
       - Dependency1 
       - Dependency2
       - ...
       - Dependency1 and license1
       - Dependency2 and license2
       - ...


python/llm/src/ipex_llm/transformers/npu_models/llama_mp.py
```diff
@@ -171,13 +172,33 @@ def __init__(
...
+            model_norm_weight = self.constant(model_norm_weight)
+
+            # norm forward
+            # result = hidden_states
+            result = self.layer_norm(result, model_norm_weight)
+            
+            # lm_head forward
+            result = self.linear(
+                result, 32000, self.hidden_size, bias=False, wt_dtype=self.dtype
+            )
```
jason-dai, line 187: Maybe move this into `build_decoder` and return linear output instead of `hidden_states`?

plusbang: > Move this into `build_decoder` and return linear output instead of `hidden_states`? Then the output from the decoder layer is the same as before.


yeah, I also tried this but the tensor shape of lm_head output is different from decoder layer, still working on the related send/recv logic to support it.

PR #11992, opened 2024-09-02
Revert prefill logic of qwen2-7b

plusbang:
## Description

We don't need to use prefill workaround to avoid abnormal output on LNL, revert prefill logic for better 1st token perf and lower cpu utilization.

### 4. How to test?
- [x] Application test


rnwang04 (approved):
LGTM

PR #12027, opened 2024-09-05
added minicpm cpu examples

JinheTang:
added minicpm-v-2 cpu example. 2_5 is yet to commit.
@rnwang04


rnwang04:
Please also add link to here: https://github.com/intel-analytics/ipex-llm?tab=readme-ov-file#verified-models

rnwang04 (approved):

PR #12052, opened 2024-09-10
optimize minicpm3 again

MeouSker77:
## Description

padding minicpm3's v cache to use sdp optimization

### DIFF
--- a/python/llm/src/ipex_llm/transformers/models/minicpm3.py
+++ b/python/llm/src/ipex_llm/transformers/models/minicpm3.py
@@ -2,12 +2,15 @@ import torch
 import warnings
 
 from torch import nn
-from typing import Optional, Tuple
+from typing import Optional, Tuple, List
 from transformers.cache_utils import Cache
 
 from ipex_llm.utils.common.log4Error import invalidInputError
 from ipex_llm.transformers.models.utils import should_use_fuse_rope
 from ipex_llm.transformers.models.utils import rotate_half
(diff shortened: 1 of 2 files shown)


python/llm/src/ipex_llm/transformers/models/minicpm3.py
```diff
@@ -20,6 +23,72 @@ def pre_compute_inv_freq(module: torch.nn.Module):
...
+        inputs_embeds: Optional[torch.FloatTensor] = None,
+        use_cache: Optional[bool] = None,
+        output_attentions: Optional[bool] = None,
+        output_hidden_states: Optional[bool] = None,
+        return_dict: Optional[bool] = None,
+    ):
+        # IPEX-LLM OPT: kv cache and quantize kv cache and sdp
+        inputs = input_ids if input_ids is not None else inputs_embeds
+        use_cache = use_cache if use_cache is not None else self.config.use_cache
+        use_cache = True if inputs.device.type == "xpu" else use_cache
```
rnwang04, line 68: For me, it's strange that force `use_cache` to True for xpu, it broke the original usage of this parameter.

rnwang04 (approved):
others LGTM.

python/llm/src/ipex_llm/transformers/models/minicpm3.py
MeouSker77: if `use_cache=False`, `sdp_causal` will get wrong result, so set it to `True` automatically if using `sdp_causal`

PR #12115, opened 2024-09-24
use english prompt in vllm_online_benchmark.py

liu-shaojun:
### 1. Why the change?

We found that when running Qwen2.5-7B-Instruct with batch=4, there is an error when using Chinese prompts, but no issue when switching to English prompts. Therefore, for now, this PR will be modified to use English prompts to collect performance data, and @hzjane will continue investigating the cause of the error with Chinese prompts.

### 4. How to test?
- [ ] N/A
- [ ] Application test
- [ ] Document test
- [ ] ...

### 5. New dependencies

- [ ] New Python dependencies
       - Dependency1 
       - Dependency2
       - ...
       - Dependency1 and license1
       - Dependency2 and license2
       - ...


glorysdj (approved):
LGTM

PR #12165, opened 2024-10-08
add llama3.2-vision Pytorch example

lzivan:
create chat.py for llama3.2-vision

- add chat.py
- readme pending
- readme added


plusbang (approved):
LGTM

PR #12251, opened 2024-10-23
Support IR and blob format for llama level0 pipeline

plusbang:
## Description

Background: https://github.com/analytics-zoo/nano/issues/1687
Necessary modification for llama IR and blob generation format in level0.

### 2. User API changes
N/A

### 4. How to test?
- [x] Application test


rnwang04 (approved):
LGTM

PR #12256, opened 2024-10-23
Fix npu group size setting of optimize_model=False

plusbang:
## Description

Fix npu group size setting of optimize_model=False

### 4. How to test?
- [x] Application test


cyita (approved):
LGTM

PR #12271, opened 2024-10-25
[fix] vllm-online-benchmark first token latency error

ACupofAir:
## Description
* Problem: When benchmark the output length is 1, the first token latency will ben none and lead to error. And when the output length is 2, the first token latency will be ok, but next token latency will be nan. The reason is the `token_time` will not be computed for the choice in json don't have `first_token_time` and `rest_token_time` which is required in the code.
image
* Solution: When the `text` in choice, and `first_token_time` is None, we should compute it.
* Verified result:
image


gc-fu (approved):
LGTM

hzjane (approved):
LGTM

xiangyuT (approved):
LGTM

PR #12275, opened 2024-10-25
Add llama3 level0 example

plusbang:
## Description

Background: https://github.com/analytics-zoo/nano/issues/1706#issuecomment-2437244193
Add llama3 example first.
llama3 use another prompt format, so separate to different script.

### 4. How to test?
- [x] Application test


rnwang04:
Shall we add `openvino` into ipex-llm[npu] ? May need to update setup.py ?

plusbang:
> Shall we add `openvino` into ipex-llm[npu] ? May need to update setup.py ?

I think we don't need to add, [npu] option is enough.

rnwang04 (approved):
LGTM

PR #12277, opened 2024-10-25
Update FunASRAutoModel Class Impl

sgwhat:
### 1. Why the change?

To avoid redundant import `funasr` in other subclasses.

### 2. User API changes

N/A

### 3. Summary of the change 

Add `get_cls_model` method in `FunASRAutoModel` class


cyita (approved):
LGTM

PR #12309, opened 2024-10-31
Qwen layernorm as input

hkvision:
Seems layernorm as input is faster for qwen-cw


cyita:
Also add `parser.add_argument("--quantization_group_size", type=int, default=0)` in this PR?

hkvision:
> Also add `parser.add_argument("--quantization_group_size", type=int, default=0)` and `quantization_group_size=args.quantization_group_size,` in this PR?

Sure.

cyita (approved):
LGTM

PR #12319, opened 2024-11-04
Fix performance tests regarding `trl` version

Oscilloscope98:
## Description

Fix performance tests regarding `trl` version


liu-shaojun (approved):
LGTM

PR #12322, opened 2024-11-04
[NPU L0] Add layernorm weight as const / input setting

plusbang:
## Description

We could specify layernorm weight as const / input through environment variable `IPEX_LLM_LAYERNORM_CONST`

### 2. User API changes
layernorm weight is set as const by default, we could change it to input by `set IPEX_LLM_LAYERNORM_CONST=0`

### 4. How to test?
- [ ] Application test

### DIFF
--- a/python/llm/src/ipex_llm/transformers/npu_pipeline_model/convert_pipeline.py
+++ b/python/llm/src/ipex_llm/transformers/npu_pipeline_model/convert_pipeline.py
@@ -189,6 +189,8 @@ def convert_llm(model: torch.nn.Module,
                 max_prompt_len: int,
                 transpose_value_cache: bool,
                 group_size: int):
+    # whether to set layernorm weight as const
+    layernorm_const = os.environ.get("IPEX_LLM_LAYERNORM_CONST", "1") == "1"
     if group_size == 0:
         n_splits_linear = 1
         n_splits_down_proj = 2 if model.config.intermediate_size == 18944 else 1
@@ -230,7 +232,7 @@ def convert_llm(model: torch.nn.Module,
                 res = InitLLMPipeline("llama", kv_len, model.num_head, model.head_dim, layer_num,
                                       model.vocab_size, weight_dir, "model",
                                       first_blob_path, last_blob_path,
-                                      os.path.join(temp_dir, "decoder_layer"))
+                                      os.path.join(temp_dir, "decoder_layer"), layernorm_const)
             except:
                 invalidInputError(False,
                                   "False to InitLLMPipeline.")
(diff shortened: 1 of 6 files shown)


python/llm/src/ipex_llm/transformers/npu_pipeline_model/convert_pipeline.py
```diff
@@ -189,6 +189,8 @@ def convert_llm(model: torch.nn.Module,
                 max_prompt_len: int,
                 transpose_value_cache: bool,
                 group_size: int):
+    # whether to set layernorm weight as const
+    layernorm_const = os.environ.get("IPEX_LLM_LAYERNORM_CONST", "1") == "1"
```
rnwang04, line 193: I feel it would be better to make it as a input parameter instead of env var ?

python/llm/src/ipex_llm/transformers/npu_pipeline_model/minicpm.py
```diff
@@ -258,9 +259,20 @@ def convert_minicpm_layer(model, layer_idx, n_splits_linear, n_splits_down_proj,
                                                         f"decoder_layer_{layer_idx}",
                                                         temp_dir)
 
-    for idx, (weight, scale) in enumerate(weights):
-        bin_file = os.path.join(weight_dir, f"model_{layer_idx}_input_{5+idx*2}.bin")
-        weight.numpy().tofile(bin_file)
-        bin_file = os.path.join(weight_dir, f"model_{layer_idx}_input_{5+idx*2+1}.bin")
-        scale.numpy().tofile(bin_file)
+    if layernorm_const:
+        for idx, (weight, scale) in enumerate(weights):
```
cyita, line 263: Maybe add a start_idx and merge the duplicated parts?

python/llm/src/ipex_llm/transformers/npu_pipeline_model/convert_pipeline.py
rnwang04: @jason-dai Not sure how you do think of this ?

jason-dai: Which method will set the parameter then? And how does the user specify it?

hkvision: I would suggest each model has a default value for this, which is based on our test result with better performance? Normally users won't know this value.

rnwang04: For now, binbin use `set IPEX_LLM_LAYERNORM_CONST` to  control it.
If we change to input var, we can update this parameter in 
`model = AutoModelForCausalLM.from_pretrained(model_path, const_layernorm=True, ...`, and it's dafault value can be None.

python/llm/src/ipex_llm/transformers/npu_pipeline_model/minicpm.py
plusbang: > Maybe add a start_idx and merge the duplicated parts or make it a utility function?

Have added `st_idx` and remove duplicated part.

plusbang:
As we discussed offline, currently first use environment variable to control layernorm weight setting. Will update default value for each model according to performance comparison.

cyita (approved):

PR #12329, opened 2024-11-04
Replace gradio_web_server.patch to adjust webui

ATMxsp01:
### Replace gradio_web_server.patch to adjust webui
4 adjust added:
- pull the `Parameters` on top of the `Performance Metrics`
- set `Parameters` default to be hidden
- change `Performance Metrics` info from Textbox to Label
- round latency info to the second decimal point

---

Now the webui looks like:
image


liu-shaojun:
I triggered a Docker build for this change: https://github.com/intel-analytics/ipex-llm-workflow/actions/runs/11675849804. Please verify the Docker image to ensure everything is functioning correctly.

liu-shaojun (approved):
LGTM

PR #12361, opened 2024-11-07
[NPU L0] Support llama3.2 in L0 pipeline

plusbang:
## Description

Support Llama-3.2-1B-Instruct & Llama-3.2-3B-Instruct in L0 pipeline

### 3. Summary of the change 

- create `Llama32Embedding` to involve rotary_emb module as first blob
- change sin and cos to fp32 input node
- update related ir and blob generation logic for llama3.2
- update readme

### 4. How to test?
- [x] Application test
Verify llama2 & llama3.2 L0 & npu-acc-lib example
https://github.com/analytics-zoo/nano/issues/1706#issuecomment-2461756923

### DIFF
--- a/python/llm/src/ipex_llm/transformers/npu_models/llama_mp.py
+++ b/python/llm/src/ipex_llm/transformers/npu_models/llama_mp.py
@@ -124,19 +124,12 @@ class LowBitLlamaMultiDecoderlayer(LLMBaseNNFactory):
         if self.cached_cos is None:
             if mode == "prefill":
                 position_ids = self.create_input_op((self.batch_size, self.seq_len), dtype=np.int64)
-                cos = self.create_input_op((self.batch_size, self.cos_len, self.head_dim),
-                                           dtype=np.float32)
-                self.cos = self.convert_to_fp16(cos)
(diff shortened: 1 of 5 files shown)


python/llm/src/ipex_llm/transformers/npu_models/llama_mp.py
```diff
@@ -124,11 +124,19 @@ def __init__(
         if self.cached_cos is None:
             if mode == "prefill":
                 position_ids = self.create_input_op((self.batch_size, self.seq_len), dtype=np.int64)
-                self.cos = self.create_input_op((self.batch_size, self.cos_len, self.head_dim))
-                self.sin = self.create_input_op((self.batch_size, self.cos_len, self.head_dim))
+                cos = self.create_input_op((self.batch_size, self.cos_len, self.head_dim),
```
cyita, line 127: It seems we can merge this if/else?

cyita (approved):

python/llm/src/ipex_llm/transformers/npu_models/llama_mp.py
plusbang: > It seems we can merge this if/else?

Sure, have updated.

PR #12416, opened 2024-11-19
Fix speech_paraformer issue with unexpected changes

sgwhat:
## Description

Hotfix for Paraformer model issue due to unexpected changes from version `2.2.0b20241118`.

### 4. How to test?
- [x] Local Test


jason-dai:
Do we need to specify which Paraformer version that we support?

rnwang04 (approved):
LGTM

sgwhat:
> Do we need to specify which Paraformer version that we support?

Sure, added related version requirement in our document.

PR #12446, opened 2024-11-26
[NPU] Fix abnormal output for Qwen2-7B when CW `sym_int8`

Oscilloscope98:
## Description

Fix abnormal output for Qwen2-7B when CW `sym_int8`, by not splitting mlp down_proj

https://github.com/analytics-zoo/nano/issues/1741#issuecomment-2499806664


cyita (approved):
LGTM

PR #12461, opened 2024-11-28
Integrate NPU C++ imple into ipex-llm

plusbang:
## Description

With https://github.com/intel-analytics/llm.cpp/pull/690, integrate NPU C++ implementation and add qwen2.5-7B support first.

### 2. User API changes

For qwen2.5-7B model, `save_directory` parameter is necessary when load model with `optimize_model=True`

### 4. How to test?
- [x] Application test
(qwen2.5-7b CW & GW)


jason-dai (approved):
LGTM

PR #12473, opened 2024-12-02
[NPU] update save-load API usage

plusbang:
## Description

Update save-load API usage.

- save: For all cases with `optimize_model=True`, specify `save_directory` is required during the first time to load model, ckpt is saved to `save_directory` after converting.
- load: still use `load_low_bit` API to load, and add python cpp backend support

### 2. User API changes
```python
# first time load, and ckpt is saved to `save_directory` after converting
model = AutoModelForCausalLM.from_pretrained(
    model_path,
    torch_dtype=torch.float16,
    trust_remote_code=True,
    attn_implementation="eager",
    optimize_model=True,
    max_context_len=1024,
    max_prompt_len=960,
    mixed_precision=True,
    quantization_group_size=0,
    save_directory=save_directory  # required and has related check
)

# load converted model
model = AutoModelForCausalLM.load_low_bit(
    save_directory,
    attn_implementation="eager",
    torch_dtype=torch.float16,
    optimize_model=True,
    max_context_len=1024,
    max_prompt_len=960,
    trust_remote_code=True,
)
```
example and benchmark scripts have updated.

### 4. How to test?
- [x] Application test

### DIFF
--- a/python/llm/src/ipex_llm/transformers/npu_model.py
+++ b/python/llm/src/ipex_llm/transformers/npu_model.py
@@ -406,6 +406,7 @@ class _BaseAutoModelClass:
         mixed_precision = config_dict.pop("mixed_precision", False)
         quantization_group_size = config_dict.pop("group_size", 0)
         optimize_model = config_dict.pop("optimize_model", False)
+        enable_cpp_backend = "weight_idx" in config_dict
 
         invalidInputError(
             qtype,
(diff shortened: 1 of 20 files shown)


python/llm/src/ipex_llm/transformers/npu_model.py
```diff
@@ -45,6 +45,9 @@ def ignore_argument(kwargs: dict, key: "str"):
 
 
 def save_low_bit(self, model_dir: str, *args, **kwargs):
+    if hasattr(self, "save_directory"):
+        logger.info(f"Model has already saved to {self.save_directory}.")
+        return 1
```
jason-dai, line 50: What if "save_directory" is different from "model_dir"?

python/llm/src/ipex_llm/transformers/npu_model.py
```diff
@@ -267,6 +270,9 @@ def optimize_npu_model(cls, *args, **kwargs):
 
         if not pipeline:
             if model.config.model_type in ["qwen2", "llama", "minicpm"]:
+                invalidInputError(save_directory is not None,
+                                  "Please provide the path to save converted model "
+                                  "through `save_directory`.")
```
jason-dai, line 275: Can we use temp dir in this case?

python/llm/src/ipex_llm/transformers/npu_model.py
plusbang: > What if "save_directory" is different from "model_dir"?

Have updated the logic of save.

For all cases with `optimize_model=True`, specify `save_directory` is required during the first time to load model, ckpt is saved to `save_directory` after converting.

python/llm/src/ipex_llm/transformers/npu_model.py
```diff
@@ -45,6 +45,8 @@ def ignore_argument(kwargs: dict, key: "str"):
 
 
 def save_low_bit(self, model_dir: str, *args, **kwargs):
+    if hasattr(self, "save_directory"):
```
jason-dai, line 48: Report a warning that the model is already saved at "save_directory"

rnwang04 (approved):
LGTM

plusbang:
Merge it first and will add tokenizer save-load processing in examples later.

PR #12477, opened 2024-12-02
Support hf generate

hkvision:
original generate -> simple_generate
current generate -> hf generate

### DIFF
--- a/python/llm/src/ipex_llm/transformers/npu_models/convert.py
+++ b/python/llm/src/ipex_llm/transformers/npu_models/convert.py
@@ -313,9 +313,9 @@ def generate(
     streamer: Optional["BaseStreamer"] = None,
     negative_prompt_ids: Optional[torch.Tensor] = None,
     negative_prompt_attention_mask: Optional[torch.Tensor] = None,
-    simple: bool = True,
     **kwargs,
 ):
+    simple = kwargs.pop("simple", True)
     if simple:
         return simple_generate(self, inputs=inputs, streamer=streamer, **kwargs)
     else:
--- a/python/llm/src/ipex_llm/transformers/npu_models/convert.py
+++ b/python/llm/src/ipex_llm/transformers/npu_models/convert.py
@@ -462,9 +462,11 @@ def causal_lm_forward(
         input_list = input_ids[0]
     input_length = len(input_list)
     if input_length > 1:
-        logits = run_prefill_with_logits(self.model_ptr, input_list, self.logits_buffer, self.vocab_size)
+        logits = run_prefill_with_logits(self.model_ptr, input_list,
+                                         self.logits_buffer, self.vocab_size)
     else:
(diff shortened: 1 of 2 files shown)


hkvision:
> Please also test other models.

Sure.

python/llm/src/ipex_llm/transformers/npu_models/convert.py
```diff
@@ -390,10 +399,65 @@ def optimize_llm_single_process(
...
+) -> Union[Tuple, CausalLMOutputWithPast]:
+    start = time.perf_counter()
+    from .npu_llm_cpp import run_decode, run_prefill, get_logits
+    if isinstance(input_ids[0], torch.Tensor):
+        input_list = input_ids[0].flatten().tolist()
+    else:
+        input_list = input_ids[0]
+    input_length = len(input_list)
+    if input_length > 1:
+        run_prefill(self.model_ptr, input_list, self.vocab_size)
```
hkvision, line 449: modify to call _lib.run_prefill

hkvision:
Current usage:
- Default to use `simple=True`, if want to benchmark, need to add `do_print=True`
- If change to `simple=False`, will change to hf generate. In this way, can't be used together with BenchmarkWrapper, because BenchmarkWrapper's generate doesn't have the `simple` argument, will have the following error:
```
File "C:\Users\arda\miniforge3\envs\kai-acc-lib\lib\site-packages\ipex_llm\utils\benchmark_util_4_29.py", line 1180, in _validate_model_kwargs
raise ValueError(
ValueError: The following model_kwargs are not used by the model: ['simple'] (note: typos in the generate arguments will also show up in this list)
```
- If want to benchmark hf generate performance, just add BenchmarkWrapper without specifying simple=True when generate.

@jason-dai

jason-dai:
> * If change to `simple=False`, will change to hf generate. In this way, can't be used together with BenchmarkWrapper, because BenchmarkWrapper's generate doesn't have the `simple` argument, will have the following error:

Can we make `simple` a kwarg instead?

hkvision:
Won't impact current all-in-one, in all-in-one, will default always to test simple generate.
If want to test hf generate, may need to manually remove the if condition here: https://github.com/intel-analytics/ipex-llm/blob/main/python/llm/dev/benchmark/all-in-one/run.py#L650C5-L651C40 or add a new option in config.yaml.
Pending to support this if needed in the future.

hkvision:
> > * If change to `simple=False`, will change to hf generate. In this way, can't be used together with BenchmarkWrapper, because BenchmarkWrapper's generate doesn't have the `simple` argument, will have the following error:
> 
> Can we make `simple` a kwarg instead?

Updated.

Still `simple` can't work together with BenchmarkWrapper, the code looks better after making it a kwarg.

jason-dai (approved):
LGTM

hkvision:
Merge it first. Will fix llama3.2 issue in a next PR very soon.

PR #12485, opened 2024-12-03
Fix NPU LLM example save/load tokenizer

JinBridger:


plusbang (commented):
Please also modify https://github.com/intel-analytics/ipex-llm/tree/main/python/llm/example/NPU/HF-Transformers-AutoModels/LLM/Pipeline-Models

plusbang (approved):
LGTM

PR #12496, opened 2024-12-04
optimize minicpm

MeouSker77:
## Description

apply scale to o_proj and down_proj to avoid scaling residual during inference


rnwang04 (approved):
LGTM

PR #12497, opened 2024-12-04
Fix hf generate for llama3.2

hkvision:
```
<|begin_of_text|><|begin_of_text|><|start_header_id|>system<|end_header_id|>

    <|eot_id|><|start_header_id|>user<|end_header_id|>

What is AI?<|eot_id|><|start_header_id|>assistant<|end_header_id|>

Artificial Intelligence (AI) is a broad term that refers to the development of computer systems that can perform tasks that would typically require human intelligence, such as:
```

When using transformers 4.45.0, past_key_values is not None initially, but DynamicCache, the previous condition is wrong and thus not running the prefill. https://github.com/huggingface/transformers/blob/329f5dbf97a5cb2473914c88c05aa3dcb242e19a/src/transformers/generation/utils.py#L1809

@jason-dai At this moment only default hf generate settings are tested. Likely that other advanced options would have issues (e.g. some other methods need to be overridden as well).

### DIFF
--- a/python/llm/src/ipex_llm/transformers/npu_models/convert.py
+++ b/python/llm/src/ipex_llm/transformers/npu_models/convert.py
@@ -455,7 +455,7 @@ def optimize_llm_single_process(
 
 def prepare_input_ids(
         self, input_ids, past_key_values=None, attention_mask=None, inputs_embeds=None, **kwargs):
-    if isinstance(past_key_values, bool) and past_key_values:  # kvcache
+    if past_key_values and isinstance(past_key_values, bool):  # kvcache
         input_ids = input_ids[:, -1]
     else:  # prefill, reset the model here
         from .npu_llm_cpp import reset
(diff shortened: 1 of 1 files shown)


plusbang (approved):
LGTM

python/llm/src/ipex_llm/transformers/npu_models/convert.py
```diff
@@ -455,7 +455,7 @@ def optimize_llm_single_process(
 
 def prepare_input_ids(
         self, input_ids, past_key_values=None, attention_mask=None, inputs_embeds=None, **kwargs):
-    if past_key_values is not None:  # kvcache
+    if isinstance(past_key_values, bool) and past_key_values:  # kvcache
```
jason-dai, line 458: `if past_key_values and isinstance(past_key_values, bool)`?

hkvision: Updated.

PR #12501, opened 2024-12-04
[NPU] Small fix for NPU Python cpp simple generate regarding eos tokens

Oscilloscope98:
## Description

Small fix for NPU Python cpp simple generate regarding eos tokens


plusbang (approved):
LGTM

PR #12509, opened 2024-12-05
[NPU] Fix hf generate with save/load generation config for Python (cpp backend)

Oscilloscope98:
## Description

https://github.com/analytics-zoo/nano/issues/1741#issuecomment-2519458756

Fix hf generate with save/load generation config for Python (cpp backend)

### DIFF
--- a/python/llm/src/ipex_llm/transformers/npu_model.py
+++ b/python/llm/src/ipex_llm/transformers/npu_model.py
@@ -439,11 +439,14 @@ class _BaseAutoModelClass:
                 model.vocab_size = config_dict['vocab_size']
                 model.logits_buffer = torch.empty(1, 1, model.vocab_size, dtype=torch.float32)
                 if model.can_generate():
-                    model.generation_config = GenerationConfig.from_pretrained(
-                        pretrained_model_name_or_path,
-                        subfolder=subfolder,
-                        **kwargs,
-                    )
+                    try:
+                        model.generation_config = GenerationConfig.from_pretrained(
+                            pretrained_model_name_or_path,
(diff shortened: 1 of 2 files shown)


jason-dai:
Do we need to save it for other backends (or even GPU)?

Oscilloscope98:
> Do we need to save it for other backends (or even GPU)?

We already have logic for saving/loading generation config for other backends (NPU Python (acc_lib) and GPU) :)

python/llm/src/ipex_llm/transformers/npu_model.py
```diff
@@ -438,6 +438,12 @@ def load_low_bit(cls, pretrained_model_name_or_path: str, *model_args, **kwargs)
                 model.kv_len = config_dict['kv_len']
                 model.vocab_size = config_dict['vocab_size']
                 model.logits_buffer = torch.empty(1, 1, model.vocab_size, dtype=torch.float32)
+                if model.can_generate():
+                    model.generation_config = GenerationConfig.from_pretrained(
```
plusbang, line 442: We use try except [here](https://github.com/intel-analytics/ipex-llm/blob/b39d0587555892467831fa8624c4d4c7a68f526f/python/llm/src/ipex_llm/transformers/npu_model.py#L622C9-L630C20), should we use similar operation here?

plusbang (approved):
Other LGTM

python/llm/src/ipex_llm/transformers/npu_model.py
Oscilloscope98: Have updated :)

PR #12526, opened 2024-12-11
All-in-one Benchmark run.py: Ignore error if import BenchmarkWrapper …

ATMxsp01:
Ignore error if import BenchmarkWrapper failed.

### DIFF
--- a/python/llm/dev/benchmark/all-in-one/run.py
+++ b/python/llm/dev/benchmark/all-in-one/run.py
@@ -32,7 +32,10 @@ current_dir = os.path.dirname(os.path.realpath(__file__))
 import sys
 sys.stdout.reconfigure(encoding='utf-8')
 
-from ipex_llm.utils import BenchmarkWrapper
+try:
+    from ipex_llm.utils import BenchmarkWrapper
+except:
+    pass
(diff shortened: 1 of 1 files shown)


python/llm/dev/benchmark/all-in-one/run.py
```diff
@@ -32,7 +32,10 @@
 import sys
 sys.stdout.reconfigure(encoding='utf-8')
 
-from ipex_llm.utils import BenchmarkWrapper
+try:
+    from ipex_llm.utils import BenchmarkWrapper
+except:
+    pass
```
Oscilloscope98, line 38: Should we fix the `except` logic for npu-related tesg api?

plusbang:
Currently, no benchmarkwrapper is compatible for transformers==4.47.0. Need to support it later : )

Oscilloscope98 (approved):
Others LGTM

python/llm/dev/benchmark/all-in-one/run.py
plusbang: > Should we only use the `except` logic for npu-related test_api?

But it's not quite convenient to add test API related check here. This is only a work around, we could remove it after support.

PR #12541, opened 2024-12-13
[NPU] Fix generate example

plusbang:
## Description

Fix generate example caused by tokenizer save logic.

### 4. How to test?
- [x] Application test


rnwang04 (approved):
LGTM

PR #12543, opened 2024-12-13
[NPU] Add `IPEX_LLM_NPU_MTL` to enable support on mtl

plusbang:
## Description

Background: https://github.com/analytics-zoo/nano/issues/1785

### 2. User API changes

To run optimized models on MTL NPU, need to set following variable:
```bash
set IPEX_LLM_NPU_MTL=1
```

No change for LNL.

### 4. How to test?
- [x] Application test

### DIFF
--- a/docs/mddocs/Quickstart/npu_quickstart.md
+++ b/docs/mddocs/Quickstart/npu_quickstart.md
@@ -90,6 +90,9 @@ For `ipex-llm` NPU support, set the following environment variable with active `
 
 ```cmd
 set BIGDL_USE_NPU=1
+
+:: [optional] for MTL support
+set IPEX_LLM_NPU_MTL=1
 ```
 
 ## Python API
--- a/python/llm/src/ipex_llm/transformers/npu_models/convert_mp.py
+++ b/python/llm/src/ipex_llm/transformers/npu_models/convert_mp.py
@@ -32,6 +32,11 @@ def convert_forward(m, target_m, new_forward):
 
 def optimize_llm_pre(model: torch.nn.Module, qtype, mixed_precision,
                      quantization_group_size=0, load=False, max_prompt_len=512):
+    if os.environ.get("IPEX_LLM_NPU_MTL", "0") != "0":
+        # For MTL support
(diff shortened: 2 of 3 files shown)


docs/mddocs/Quickstart/npu_quickstart.md
```diff
@@ -90,6 +90,9 @@ For `ipex-llm` NPU support, set the following environment variable with active `
 
 ```cmd
 set BIGDL_USE_NPU=1
```
rnwang04, line 92: Maybe we can rename to `IPEX_LLM_USE_NPU` later ? Now it's a little strange with other env vars.

python/llm/src/ipex_llm/transformers/npu_models/convert_mp.py
```diff
@@ -32,6 +32,11 @@ def convert_forward(m, target_m, new_forward):
 
 def optimize_llm_pre(model: torch.nn.Module, qtype, mixed_precision,
                      quantization_group_size=0, load=False, max_prompt_len=512):
+    if os.environ.get("IPEX_LLM_NPU_MTL", "0") != "0":
```
rnwang04, line 35: if os.environ.get("IPEX_LLM_NPU_MTL", "0") == "1":

rnwang04 (approved):
others LGTM.

docs/mddocs/Quickstart/npu_quickstart.md
plusbang: > ?

Good catch! I think we could also verify whether it's still necessary later : )

plusbang: > Maybe we can rename to `IPEX_LLM_USE_NPU` later ? Now it's a little strange with other env vars.

Good catch! I think we could also verify whether this is still necessary later : )

python/llm/src/ipex_llm/transformers/npu_models/convert_mp.py
plusbang: > if os.environ.get("IPEX_LLM_NPU_MTL", "0") == "1":

Have updated.

PR #12575, opened 2024-12-18
Replace runner name to ipex_llm in ollama quickstart

sgwhat:
## Description

Ipex-llm ollama has already switched server runner from `[cpu, cpu_avx, cpu_avx2]` to `[ipex_llm] `.


rnwang04 (approved):
LGTM

PR #12619, opened 2024-12-26
support passing attn_scale to sdpa to fix head_dim padding

MeouSker77:
## Description

support passing attn_scale to sdpa to fix head_dim padding


rnwang04 (approved):
LGTM

PR #12699, opened 2025-01-10
Update ollama v0.5.1 document

sgwhat:
### 1. Why the change?

Update ollama version to v0.5.1 and one additional troubleshooting. 

Also include updating outdating ollama serve image to v0.5.1.


yizhangliu:
The update is too slow, hope to keep up with the ollama version in time. So many new models...

PR #12726, opened 2025-01-21
Intercept ocl environment variables before running

leonardozcm:
## Description

https://github.com/analytics-zoo/nano/issues/1814#issuecomment-2601435547

### DIFF
--- a/python/llm/setup.py
+++ b/python/llm/setup.py
@@ -211,6 +211,30 @@ def download_libs(url: str, change_permission=False):
     if change_permission:
         os.chmod(libso_file, 0o775)
 
+from setuptools.command.install import install
+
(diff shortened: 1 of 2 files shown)


python/llm/setup.py
```diff
@@ -211,6 +211,30 @@ def download_libs(url: str, change_permission=False):
...
 
+from setuptools.command.install import install
+
+class IPEXLLMInstallCommand(install):
+    def run(self):
+        install.run(self)
+        self.copy_activate_script()
+
+    def copy_activate_script(self):
+        conda_prefix = os.environ.get('CONDA_PREFIX')
```
Oscilloscope98, line 222: What if for Windows?

PR #12755, opened 2025-01-26
Add troubleshooting section for bmg quickstart

liu-shaojun:
## Description

This PR adds a troubleshooting section to the bmg documentation, which provides a solution for the `RuntimeError: could not create an engine` issue encountered on Linux systems. The issue can be resolved by unsetting the `OCL_ICD_VENDORS` environment variable using the command:

```bash
unset OCL_ICD_VENDORS
```

This fix is specific to Linux and does not affect Windows environments.


glorysdj (approved):
LGTM

PR #12786, opened 2025-02-07
[NPU] Update deepseek support in python examples and quickstart

plusbang:
## Description

Update deepseek support in python examples and quickstart.
For python examples, could directly reuse qwen2.py to run.

### 4. How to test?
- [x] Application test


rnwang04 (approved):
LGTM

PR #12788, opened 2025-02-07
[NPU] Support qwen models with `cos_sin_input=True`

plusbang:
## Description

With https://github.com/intel-analytics/llm.cpp/pull/826
- When transformers>=4.45.0, we must support `cos_sin_input=True`.
- To reduce IR size

### 2. User API changes

N/A, but now could run qwen models with `transformers==4.45.0`

### 4. How to test?
- [x] Application test (verified qwen2-1.5b with both transformers==4.40.0 and 4.45.0, verified llama3.2)

### DIFF
--- a/python/llm/src/ipex_llm/transformers/npu_pipeline_model/qwen.py
+++ b/python/llm/src/ipex_llm/transformers/npu_pipeline_model/qwen.py
@@ -18,8 +18,9 @@
 import torch
 import numpy as np
 import os
-from .common import update_names_of_IR_and_export_blob, LLMEmbedding, LowBitLLMLMHead, \
-    obtain_weight_from_single_layer, obtain_qkv_bias_from_single_layer
+from .common import update_names_of_IR_and_export_blob, LowBitLLMLMHead, \
+    obtain_weight_from_single_layer, obtain_qkv_bias_from_single_layer, \
+    obtain_embedding_from_model
 from ipex_llm.transformers.npu_models.lm_head import SlicedLMHead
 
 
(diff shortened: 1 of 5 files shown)


python/llm/src/ipex_llm/transformers/npu_pipeline_model/qwen.py
```diff
@@ -107,24 +107,63 @@ def convert_lm_head_and_embedding(model, temp_dir, weight_dir,
...
+                                                                 temp_dir, keep_ir=keep_ir,
+                                                                 compile_blob=compile_blob)
+            os.remove(os.path.join(temp_dir, "embedding.bin"))
     else:
-        first_blob_path = update_names_of_IR_and_export_blob(new_embedding, f"embedding",
-                                                             temp_dir, keep_ir=keep_ir,
-                                                             compile_blob=compile_blob)
-        os.remove(os.path.join(temp_dir, "embedding.bin"))
+        # transformers >= 4.45.0
+        from .llama import Llama32Embedding, Llama32PostEmbedding
```
rnwang04, line 131: Shall we move Llama32Embedding, Llama32PostEmbedding to common and  maybe rename it ?

plusbang: > Shall we move Llama32Embedding, Llama32PostEmbedding to common and maybe rename it ?

yeah, have moved. And for llama & qwen models , could use the same logic of embedding, so also move this to common.

rnwang04 (approved):
LGTM

PR #12805, opened 2025-02-11
Fix ipex-llm CPU linear dtype not match

xiangyuT:
## Description

Fix ipex-llm CPU linear dtype not match on ICX, etc.

https://github.com/intel-analytics/ipex-llm-workflow/actions/runs/13254227081


gc-fu (approved):
LGTM

PR #12808, opened 2025-02-11
add basic support for Baichuan-M1-14B-Instruct

MeouSker77:
## Description

Add basic support and optimization for Baichuan-M1-14B-Instruct

Note:
1. batch inference may be not supported
2. sliding window is not support, so the output may be wrong when `input_length+output_length > 8192`
3. cpu inference is not supported, only xpu is supported
4. only test with transformers 4.45

### 2. User API changes

1. modify downloaded `modeling_baichuan.py`
```
from transformers.modeling_flash_attention_utils import _flash_supports_window_size, \
    _upad_input
```

->

```
from transformers.modeling_flash_attention_utils import _upad_input
```

```
        self.self_attn = Baichuan_ATTENTION_CLASSES['flash_attention_2'](config, layer_idx)
```

->

```
        self.self_attn = Baichuan_ATTENTION_CLASSES['eager'](config, layer_idx)
```

2. use torch 2.6 (torch 2.3 may be ok, but not test yet), torch 2.1 is not supported

3. load model
```python
from ipex_llm.transformers import AutoModelForCausalLM
model = AutoModelForCausalLM.from_pretrained(
    model_name, trust_remote_code=True,
    torch_dtype = torch.half, load_in_low_bit='sym_int4'
).eval()
model = model.to('xpu')
```

### DIFF
--- a/python/llm/src/ipex_llm/transformers/convert.py
+++ b/python/llm/src/ipex_llm/transformers/convert.py
@@ -1062,6 +1062,11 @@ def _optimize_pre(model, qtype=None):
         from ipex_llm.transformers.models.glm import merge_qkv, split_mlp
         model.apply(merge_qkv)
         model.apply(split_mlp)
+    elif model.config.model_type == "baichuan_m1":
+        from ipex_llm.transformers.models.baichuan_m1 import pre_register_inv_freq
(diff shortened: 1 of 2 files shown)


Oscilloscope98 (approved):
LGTM

PR #12866, opened 2025-02-20
Simplify EOS token handling in tokenize function

DreadPirate09:
Let the tokenizer handle EOS token

### DIFF
--- a/docker/llm/finetune/lora/cpu/docker/lora_finetune.py
+++ b/docker/llm/finetune/lora/cpu/docker/lora_finetune.py
@@ -137,26 +137,18 @@ def train(
     tokenizer.padding_side = "left"  # Allow batched inference
 
     def tokenize(prompt, add_eos_token=True):
-        # there's probably a way to do this with the tokenizer settings
-        # but again, gotta move fast
         result = tokenizer(
             prompt,
             truncation=True,
             max_length=cutoff_len,
             padding=False,
             return_tensors=None,
-        )
-        if (
-            result["input_ids"][-1] != tokenizer.eos_token_id
-            and len(result["input_ids"]) < cutoff_len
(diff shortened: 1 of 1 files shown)


DreadPirate09:
@Oscilloscope98 can you review this ? Thanks!

Uxito-Ada:
Hi @DreadPirate09 ,

Thanks for your contribution. The code previously manually appends EOS token because `add_special_token=True` will add both a BOS and EOS at the beginning and end respectively, as illustrated [here](https://github.com/huggingface/transformers/issues/22935). Therefore, I think `add_special_token` does not exactly match `add_eos_token`'s semantics.

docker/llm/finetune/lora/cpu/docker/lora_finetune.py
```diff
@@ -137,26 +137,18 @@ def train(
...
             return_tensors=None,
-        )
-        if (
-            result["input_ids"][-1] != tokenizer.eos_token_id
-            and len(result["input_ids"]) < cutoff_len
-            and add_eos_token
-        ):
-            result["input_ids"].append(tokenizer.eos_token_id)
-            result["attention_mask"].append(1)
+            add_special_tokens=add_eos_token,
```
Uxito-Ada, line 146: Hi @DreadPirate09 ,

Thanks for your contribution. The code previously manually appends EOS token because `add_special_token=True` will add both a BOS and EOS at the beginning and end respectively, as illustrated [here](https://github.com/huggingface/transformers/issues/22935). Therefore, I think `add_special_token` does not exactly match `add_eos_token`'s semantics.

DreadPirate09: Ok ... I understand now

PR #12886, opened 2025-02-24
optimize attention part of moonlight-14B-A3B

MeouSker77:
### 1. Why the change?

most time is cost by MOE part during decoding, so the speedup of this optimization maybe not obvious


rnwang04 (approved):
LGTM

PR #13007, opened 2025-03-25
Fix vllm glm edge model

gc-fu:
## Description

ipex-llm does not obey `skip_bias_add` argument.

It always add the bias if bias is not None.

But we can safely expand the check condition based on this rule to support glm-edge-4B-Chat model


xiangyuT (approved):
LGTM

PR #13031, opened 2025-04-01
Fix: pin cmake version in cpu serving dockerfile to resolve build issue 

xiangyuT:
## Description

Pin cmake version to 3.31.6 in cpu serving dockerfile to resolve vLLM build issue.


liu-shaojun (commented):
lgtm

liu-shaojun (approved):

PR #13148, opened 2025-05-12
Update Dockerfile

gc-fu:
## Description

Re-enable sym_int4 in Dockerfile. As it may induce problems in BMG platform


liu-shaojun (approved):
LGTM

PR #13177, opened 2025-05-22
vLLM: Fix conver_to_half condition 

xiangyuT:
## Description

Found by Jian. Only need to check `self.conver_to_half` flag when x_2d.dtype is not half.


hzjane (approved):
LGTM

xiangyuT:
https://github.com/intel-analytics/ipex-llm-workflow/actions/runs/15180089434

PR #13305, opened 2025-09-11
Bug Fix - Dependency Confusion in oneccl_bind_pt package

liu-shaojun:
## Description

This PR addresses **BUG - Dependency Confusion in `oneccl_bind_pt` package** (see [AO8-417](https://jira.devtools.intel.com/browse/AO8-417)).

The issue arises because we currently use `--extra-index-url` in multiple places together with the `oneccl_bind_pt` dependency. Since `oneccl_bind_pt` is **not** published on [pypi.org](https://pypi.org/), `pip` will first attempt to fetch from PyPI, and if a malicious actor publishes a fake `oneccl_bind_pt` package there, our scripts would install it.

**Fix:**

* Replace all occurrences of `--extra-index-url` with `--index-url`.
* This ensures that `pip` only pulls from the trusted internal index and does **not** fallback to `pypi.org`.


xiangyuT (approved):
LGTM

glorysdj (approved):
LGTM

